PR #1515, opened 2022-10-05
feat(quotas): Enforce rate limits on metrics buckets [INGEST-1654]

jjbayer:
Apply rate limits to metrics buckets from the `transactions` namespace.

# Background 

Dynamic Sampling introduces a new type of quota, `transactions_processed`, which defines how many transactions should be "processed", i.e. whether or not we should extract metrics from them. This quota is always >= the `transactions` quota, which defines how many transactions should be stored & indexed.

The new data category should not only apply to incoming transaction payloads (see https://github.com/getsentry/relay/pull/1517), but also to metrics buckets, which may have been extracted by an upstream Relay.

Moreover, processing Relays must _count_ the number of transactions for which metrics have been dropped, and create outcomes for them, in the same way that we create accepted outcomes (see https://github.com/getsentry/sentry/pull/39236).

# Data Flow

Processing Relays check rate limits against Redis. Currently the only actor that has access to this rate limiter is the `EnvelopeProcessor`, and we want to make use of its thread pool to make the blocking calls to Redis. This makes it necessary to redirect metrics buckets to the processor before sending them to the `EnvelopeManager` for publishing:

## Old Flow (still applies under certain conditions)

```
 ┌──────────┐    ┌────────────┐ ┌───────────────┐
 │Aggregator│    │ProjectCache│ │EnvelopeManager│
 └────┬─────┘    └─────┬──────┘ └──────┬────────┘
      │                │               │
      │  FlushBuckets  │               │
      ├───────────────►│               │
      │                │  SendMetrics  │
      │                ├──────────────►│
      │                │               │
```

## New Flow

```
 ┌──────────┐  ┌────────────┐       ┌─────────────────┐ ┌─────┐ ┌───────────────┐
 │Aggregator│  │ProjectCache│       │EnvelopeProcessor│ │Redis│ │EnvelopeManager│
 └────┬─────┘  └─────┬──────┘       └────────┬────────┘ └──┬──┘ └──────┬────────┘
      │              │                       │             │           │
      │ FlushBuckets │                       │             │           │
      ├─────────────►│                       │             │           │
      │              │ RateLimitFlushBuckets │             │           │
      │              ├──────────────────────►│             │           │
      │              │                       ├────────────►│           │
      │              │                       │             │           │
      │              │                       │◄────────────┤           │
      │              │                       │             │           │
      │              │                       │             │           │
      │              │                       │       SendMetrics       │
      │              │                       ├─────────────┬──────────►│
      │              │                       │             │           │
```

# Business Logic

1. ProjectCache received a `FlushBuckets` from the metrics aggregator.
2. Count the number of transactions that contributed to these buckets (see below).
2. Check if cached rate limits apply.
3. If so, drop transaction-related buckets, generate outcomes, and send the buckets to `EnvelopeManager`.
4. Otherwise, if processing mode is enabled, send buckets to the `EnvelopeProcessor`.
5. The processor communicates the transaction count to redis and applies the rate limit if reached.
6. The processor sends a message to the project cache to update the cached rate limits.
7. Drop transaction-related buckets, generate outcomes, and send the buckets to `EnvelopeManager`.

## Counting processed transactions

* For `d:transactions/duration@millisecond`, increment the counter in redis by the length of the bucket's value. This is an accurate count of the number of transactions that contributed to the bucket.
* For any other metric, do not increment the counter in redis, but still enforce the rate limit if the quota is currently exceeded.

# Not in this PR
* Enforcement of rate limits on the fast path.
* Enforcements of rate limits on metrics buckets in the `sessions` namespace.

.


relay-server/src/actors/processor.rs
```diff
@@ -515,13 +517,31 @@ impl EncodeEnvelope {
...
+pub struct CheckRateLimits {
+    /// Quotas to be checked.
+    pub quotas: Vec<Quota>,
+    /// Category to check against.
+    pub category: DataCategory,
+    /// Scoping information for the project.
+    pub scoping: Scoping,
+    /// By how much the quota usage is incremented.
+    pub quantity: usize,
+}
```
jjbayer, line 527: The business logic of checking rate limits lives in `ProjectCache`, but checking redis rate limits happens in the processor, such that we don't block the caller thread.

relay-server/src/actors/project.rs
```diff
@@ -610,6 +616,7 @@ impl Project {
     ///
     /// The metrics will be keyed underneath this project key.
     pub fn insert_metrics(&mut self, metrics: Vec<Metric>) {
+        // TODO: rate limits
```
jjbayer, line 619: @jan-auer I remember adding these TODOs together, but isn't this actually too late to rate limit incoming metrics? `InsertMetrics` is sent by the processor, don't we want to apply rate limits already on the fast path?

relay-server/src/actors/project_cache.rs
```diff
@@ -40,6 +49,151 @@ pub enum ProjectError {
...
+        ));
+    }
+
+    Ok(rate_limits)
+}
+
+/// Remove rate limited metrics buckets and track outcomes for them.
+///
+/// Returns any buckets that were *not* rate limited.
+async fn rate_limit_buckets(
```
jjbayer: This stand-alone function should be called from wherever we want to rate limit metrics buckets.

tests/integration/test_store.py
```diff
@@ -470,6 +471,194 @@ def test_processing_quotas(
...
+    mini_sentry,
+    relay_with_processing,
+    metrics_consumer,
+    outcomes_consumer,
+    violating_bucket,
+):
+    """
+    param violating_bucket is parametrized so we cover both cases:
+        1. the quota is matched exactly
+        2. quota is exceeded by one
```
jjbayer, line 485: The second parametrization will fail until https://github.com/getsentry/relay/pull/1521 gets merged.

relay-server/src/actors/project_cache.rs
```diff
@@ -623,10 +797,33 @@ impl Handler<FlushBuckets> for ProjectCache {
...
+        let quotas = project_state
+            .config
+            .quotas
+            .iter()
+            .filter(|q| q.categories.contains(&DataCategory::TransactionProcessed))
+            .map(Quota::clone)
+            .collect::<Vec<_>>();
+        let processing_enabled = config.processing_enabled();
+        let rate_limits = project.rate_limits().clone();
+        use futures::{FutureExt, TryFutureExt};
```
olksdr: Do we actually want to remove the import here and move it to the top of the file?

relay-server/src/actors/project_cache.rs
```diff
@@ -4,30 +4,39 @@ use std::time::Instant;
 use actix::prelude::*;
 use actix_web::ResponseError;
 use failure::Fail;
+// use futures::{FutureExt, TryFuture, TryFutureExt};
```
olksdr: I think we can uncomment this.

relay-server/src/actors/project_cache.rs
```diff
@@ -40,6 +49,171 @@ pub enum ProjectError {
...
+            remote_addr: None,
+            category: DataCategory::TransactionProcessed,
+            quantity: transaction_count as u32,
+        });
+    }
+}
+
+/// Remove rate limited metrics buckets and track outcomes for them.
+///
+/// Returns any buckets that were *not* rate limited.
```
olksdr:
```suggestion
/// Returns any buckets that were *NOT* rate limited.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -40,6 +49,171 @@ pub enum ProjectError {
...
+async fn rate_limit_buckets(
+    buckets: Vec<Bucket>,
+    quotas: Vec<Quota>,
+    cached_rate_limits: RateLimits,
+    scoping: Scoping,
+    check_redis: bool,
+) -> Result<Vec<Bucket>, ()> {
+    // Collect bucket information (bucket, transaction_count)
```
olksdr:
```suggestion
    // Collect bucket information (bucket, transaction_count).
```

relay-server/src/actors/project_cache.rs
```diff
@@ -40,6 +49,171 @@ pub enum ProjectError {
...
+                    &mut annotated_buckets,
+                    transaction_count,
+                    outcome,
+                    *item_scoping,
+                );
+            }
+        };
+    }
+
+    // Restore vector of buckets
```
olksdr:
```suggestion
    // Restore vector of buckets.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -40,6 +49,171 @@ pub enum ProjectError {
...
+            quotas,
+            item_scoping,
+            check_redis,
+        )
+        .await;
+
+        match rate_limits {
+            Ok(rate_limits) => {
+                // If a rate limit is active, discard transaction buckets.
+                if let Some(limit) = rate_limits.limit_for(DataCategory::TransactionProcessed) {
```
jjbayer: Is it even necessary to filter by data category here, given the `item_scoping`? Should we use `rate_limits.longest()` instead?

relay-server/src/actors/project.rs
```diff
@@ -601,6 +606,7 @@ impl Project {
     ///
     /// The buckets will be keyed underneath this project key.
     pub fn merge_buckets(&mut self, buckets: Vec<Bucket>) {
+        // TODO: rate limits
```
jjbayer, line 609: We will add a call to `BucketLimiter::enforce_limits()` to `metrics_allowed()` in a follow-up PR.

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    // Drop transaction-related buckets and create outcomes for any active rate limits.
+    //
+    // If rate limits could not be checked for some reason, pass an `Err` to this function.
+    // In this case, transaction-related metrics buckets will also be dropped, and an "internal"
+    // outcome is generated.
+    //
+    // Returns true if any buckets were dropped.
+    pub fn enforce_limits(&mut self, rate_limits: Result<&RateLimits, ()>) -> bool {
```
olksdr, line 136: To accept the result into the function seems a little bit strange and we kind of ignore the `Err` anyway.
Could we just accept `Option<&RateLimits>` instead? In my opinion it will make the interface a little bit cleaner.

olksdr (approved):
I have only one question/suggestion, otherwise lgtm!

relay-server/src/utils/metrics_rate_limits.rs
jjbayer: I see your point, given that the `Err` variant is empty anyway. But I feel that using a `Result` instead of an `Option` makes it clear to the reader that this is an error case.

olksdr: When I look into this, first comes to my mind that all the errors should have been handled before calling this function, and it should get here only if we are ok with enforcing the limits (whatever it means in this case).  

If we have `Option`, in case of `None` it says to me - there is no limits to enforce, just emit outcome, otherwise enforce what you can 😄 once there is `Some(_)`

PR #1517, opened 2022-10-05
feat(quotas): Require TransactionProcessed quota

flub:
If metrics have not yet been extracted, a transaction event needs to
have `DataCategory::TransactionProcessed` quota available for ingestion.

Otherwise, we are in the quota enforcement part of the Relay event
processing pipeline and should check and count transaction indexing
quota. Transaction processing quota however is only counted in the
metrics part and not in event ingestion.

# Background

Rate limiting and quota enforcement has to apply to _legacy_ plans that
did not have processed transaction functionality, and to _new_ plans
that introduce them.

### Organization on Legacy Plan

There is a single quota for indexed transactions. We need to extract
metrics for indexed transactions so that there is backfilled history at
the time the customer switches to the new plan. This requires to extract
metrics for all accepted transactions.

### Organization on New Plan

This plan has a dedicated quota for processed transactions. Indexing
transactions requires quota for processed transactions.

- Validation and inbound data filters do not consume processing quota.
  There are no metrics extracted for such dropped events.
- If there is processing quota available, extract and ingest metrics for
  this transaction. Otherwise, drop the transaction event and any
  transaction metrics bucket.
- Dynamic sampling may drop (previously processed) transaction events.
  Extracted metrics and the processed transaction quotas are not
  impacted by this.
- If there is indexing (and processing) quota available, store the
  sampled transaction event. Otherwise, drop the transaction event.

# Algorithm

**Terminology:**

- `tx`: Existing quota for indexed transactions.
- `txp`: New quota for processed transactions.

**Preconditions:**

- On legacy plans, the upstream sets `txp` := `tx`

**Flow:**

1. `CheckEnvelope` in the fast-path
   1. If `txp` is limited or `txp==0`: drop; else: keep
2. Filtering / Validation
   1. Potentially drop the event here
3. Metrics extraction
4. Dynamic Sampling
   1. legacy plans never have sampling rules, so skip sampling
   2. if sampling yields “drop”, go to step 6a
5. Quota enforcement on the transaction event item
   1. Run `tx += 1`, `txp += 0`
   2. Drop event if **either** `tx` or `txp` are limited.
   3. Drop metrics if `txp` is limited.
6. Store what is left
   1. Insert metrics into aggregator
   2. Place event into envelope and store

# Dependencies

- This requires `TransactionProcessing` quota to be set for all legacy
  plans.


relay-server/src/utils/rate_limits.rs
```diff
@@ -132,6 +132,9 @@ pub struct EnvelopeSummary {
 
     /// Indicates that the envelope contains regular attachments that do not create event payloads.
     pub has_plain_attachments: bool,
+
+    /// Whether the envelope contains a transaction event which already had the metrics extracted.
+    pub metrics_extracted: bool,
```
jjbayer: Let's call this `transaction_metrics_extracted`, since session items have their metrics extracted, too.

relay-server/src/utils/rate_limits.rs
```diff
@@ -152,6 +155,10 @@ impl EnvelopeSummary {
                 summary.has_plain_attachments = true;
             }
 
+            if item.metrics_extracted() {
+                summary.metrics_extracted = true;
```
jjbayer: Both sessions and transactions use this item header. Even if they are never sent in the same envelope, I would be defensive here and check the item type.

relay-server/src/utils/rate_limits.rs
```diff
@@ -372,18 +387,53 @@ where
...
+                if summary.metrics_extracted {
+                    // Check and increment indexing quota.
+                    let index_limits = (self.check)(scoping.item(DataCategory::Transaction), 1)?;
+                    let longest = index_limits.longest();
+                    enforcement.event = CategoryLimit::new(DataCategory::Transaction, 1, longest);
+                    rate_limits.merge(index_limits);
+
+                    // Check processing quota, but DO NOT increment.
+                    let processing_limits =
+                        (self.check)(scoping.item(DataCategory::TransactionProcessed), 0)?;
```
jjbayer: It looks like `is_rate_limited` does not return any limits when the quantity is 0, see investigation here: https://github.com/getsentry/relay/pull/1519.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - Remove long-running futures from metrics flush. ([#1492](https://github.com/getsentry/relay/pull/1492))
 - Migrate to 2021 Rust edition. ([#1510](https://github.com/getsentry/relay/pull/1510))
 - Make the profiling frame object compatible with the stacktrace frame object from event. ([#1512](https://github.com/getsentry/relay/pull/1512))
+- Fix quota DataCategory::TransactionProcessed serialisation to match that of the CAPI. ([#1514](hptts://github.com/getsentry/relay/pull/1514))
```
iker-barriocanal: Please merge master.

relay-server/src/utils/rate_limits.rs
```diff
@@ -372,18 +387,53 @@ where
...
+                        CategoryLimit::new(DataCategory::TransactionProcessed, 1, longest);
+                    rate_limits.merge(processing_limits);
+                } else {
+                    let processing_limits =
+                        (self.check)(scoping.item(DataCategory::TransactionProcessed), 0)?;
+                    let longest = processing_limits.longest();
+
+                    // Reject the entire transaction event as we do not want to index it nor
+                    // extract metrics from it.
+                    enforcement.event = CategoryLimit::new(DataCategory::Transaction, 1, longest);
```
iker-barriocanal: I don't understand why the `longest` rate limit for `Transaction` is coming from `TransactionProcessed`. Couldn't it be lower?

relay-server/src/utils/rate_limits.rs
```diff
@@ -372,18 +387,53 @@ where
...
+                if summary.transaction_metrics_extracted {
+                    // Check and increment indexing quota.
+                    let index_limits = (self.check)(scoping.item(DataCategory::Transaction), 1)?;
+                    let longest = index_limits.longest();
+                    enforcement.event = CategoryLimit::new(DataCategory::Transaction, 1, longest);
+                    rate_limits.merge(index_limits);
+
+                    // Check processing quota, but DO NOT increment.
+                    let processing_limits =
+                        (self.check)(scoping.item(DataCategory::TransactionProcessed), 0)?;
```
iker-barriocanal, line 400: I don't understand the logic behind checking the processed quota with a quantity of 0.

iker-barriocanal: Why do we check `TransactionProcessed` quota here? If metrics have already been extracted, the transaction has been processed. One reason is that this function is called again later in the pipeline, but in that case we shouldn't increment the limit below?

flub: This branch is **if** metrics were already extracted, i.e. the 2nd time in the pipeline (or the first time if an earlier relay extracted the metrics).  The reason for this check is that we do not want to index a transaction which we're not also extracting metrics from.  So even though we might still have indexing quota we should not index this if we're out of processing quota.

But writing this makes me realise this code is incorrect.  If this quota is exhausted we should **also** mark the event as enforced.  This it doesn't currently do.

relay-server/src/utils/rate_limits.rs
flub: I'm not sure I follow your question.  This marks the event as enforced since if we don't have processing quota we also do not want to index the transaction, so we enforce both.  Since only processing quota now matters we take the rate limit (duration till when we're not accepting transactions anymore) from the processed rate limiter call.

Does that answer the question or did I get it entirely wrong?

flub:
I believe this is ready, please review.

It does not yet have integration tests, I think for this it needs to be based on top of https://github.com/getsentry/relay/pull/1521 since it needs this to increment the used quotas of the metric buckets.

relay-server/src/utils/rate_limits.rs
```diff
@@ -237,15 +244,17 @@ impl Default for CategoryLimit {
 #[derive(Default, Debug)]
 pub struct Enforcement {
     /// The event item rate limit.
-    event: CategoryLimit,
+    pub event: CategoryLimit,
```
jan-auer: Could we replace the public fields with the accessor methods we need?

relay-server/src/utils/rate_limits.rs
```diff
@@ -378,18 +387,59 @@ where
...
+
+                    rate_limits.merge(processing_limits);
+                }
+            } else {
+                let event_limits = (self.check)(scoping.item(category), 1)?;
+                let longest = event_limits.longest();
+                enforcement.event = CategoryLimit::new(category, 1, longest);
+
+                // Record the same reason for attachments, if there are any.
+                enforcement.attachments = CategoryLimit::new(
```
jan-auer, line 433: Transactions can also have attachments, but this branch does not run on attachments. If we're dropping a transaction, then we should also record the same limit for attachments.

relay-server/src/utils/rate_limits.rs
```diff
@@ -378,18 +387,59 @@ where
...
+                    if !enforcement.event.is_active() {
+                        // Check and increment indexing quota.
+                        let index_limits =
+                            (self.check)(scoping.item(DataCategory::Transaction), 1)?;
+                        let longest = index_limits.longest();
+                        enforcement.event =
+                            CategoryLimit::new(DataCategory::Transaction, 1, longest);
+                        rate_limits.merge(index_limits);
+                    }
+                } else {
```
jan-auer: Unless I'm missing something, both branches look identical with the one difference that you run another check for the event. If that's the case, then this could be rewritten with less code duplication.

relay-server/src/utils/rate_limits.rs
```diff
@@ -378,18 +387,59 @@ where
...
-            // Record the same reason for attachments, if there are any.
-            enforcement.attachments = CategoryLimit::new(
-                DataCategory::Attachment,
-                summary.attachment_quantity,
-                longest,
-            );
-
-            rate_limits.merge(event_limits);
+            if category == DataCategory::Transaction {
+                if summary.transaction_metrics_extracted {
```
jan-auer: I do not follow the reasoning behind this branch here, could you elaborate? Even if there is a transaction item without metrics, then we still need to invoke the rate limiter for the (indexed) transaction category.

tests/integration/test_store.py
```diff
@@ -366,7 +366,7 @@ def test_processing(
 @pytest.mark.parametrize(
     "window,max_rate_limit", [(86400, 2 * 86400), (2 * 86400, 86400)]
 )
-@pytest.mark.parametrize("event_type", ["default", "error", "transaction"])
+@pytest.mark.parametrize("event_type", ["default", "error"])
```
jan-auer, line 370: We still need to run this integration test for transactions, this is still valid behavior.

relay-server/src/utils/rate_limits.rs
```diff
@@ -850,4 +900,69 @@ mod tests {
...
+        let limiter = EnvelopeLimiter::new(|s, q| mock.check(s, q));
+        let (enforcement, limits) = limiter.enforce(&mut envelope, &scoping()).unwrap();
+
+        assert!(!limits.is_limited());
+        assert!(!enforcement.extracted_transaction_metrics.is_active());
+
+        // Note we never check for the DataCategory::Transaction so we are not setting this
+        // to enforced.  After metrics extraction will check the rate limiter again and then
+        // this will actually do something.
+        assert!(!enforcement.event.is_active());
```
jan-auer, line 947: I believe this is an invalid assumption. Even if metrics extraction is switched off, we will still have to check for this quota. This check always needs to run.

relay-server/src/utils/rate_limits.rs
iker-barriocanal: > if we don't have processing quota we also do not want to index the transaction, so we enforce both

That's correct, but unless I'm missing something the code does a different thing. To know the processing quota is exhausted, we check the processing quota; but to know whether the indexing quota is exhausted, shouldn't we check the indexing quota? So this line (423) currently enforces the indexed transaction based on the quota left for processed transactions, ignoring the two quotas are different. The other part of the branch follows the same approach, which means this case isn't "fixed" at a later point in the pipeline.

So I wonder why we don't perform two different checks (i.e. calling `(self.check)(...Transaction...)` for the two quotas, and only do one check. I'm not sure, however, if I've missed something again.

relay-server/src/utils/rate_limits.rs
flub: Sure, but out of curiosity, what is the tradeoff here?  For public APIs from libraries I understand that the accessor methods are easier to provide backwards compatibility for, but for something internal I'm not sure what advantage this has. What am I missing?

relay-server/src/utils/rate_limits.rs
flub: If transactions are extracted then we're at the `enforce_quotas()` part of the flow and we need to execute that logic.  If not we're at the `check_envelope` part.  There is the case where an earlier relay has already extracted the metrics and we go straight to the enforce_quotas behaviour even in check_envelope.

relay-server/src/utils/rate_limits.rs
flub: yup, I kept the control flow literally as in the design doc to start with.  Was expecting maybe tighten this down during review once we're sure it's the right logic.

relay-server/src/utils/rate_limits.rs
flub: This branch handles the case where metrics are not yet extracted.  This can only happen during the check_envelope stage.  At this point we check for processing quota.  If have exhausted this quota we do not want to ingest this event, at all.  So we enforce the event so that it will be removed and not further processed at all, not for indexing nor extracting metrics

We never do the second check for indexing quota since we only do that check *after* we have extracted the metrics (the branch above).  We should not check it since we will check it later anyway, and when we check it it will increment the count.

relay-server/src/utils/rate_limits.rs
flub: I'm confused, isn't that exactly what this code does?  It is missing from the branch above though.

relay-server/src/utils/rate_limits.rs
```diff
@@ -850,4 +900,69 @@ mod tests {
...
+        let mut mock = MockLimiter::default().deny(DataCategory::Transaction);
+        let limiter = EnvelopeLimiter::new(|s, q| mock.check(s, q));
+        let (enforcement, limits) = limiter.enforce(&mut envelope, &scoping()).unwrap();
+
+        assert!(!limits.is_limited());
+        assert!(!enforcement.extracted_transaction_metrics.is_active());
+
+        // Note we never check for the DataCategory::Transaction so we are not setting this
+        // to enforced.  After metrics extraction will check the rate limiter again and then
+        // this will actually do something.
```
iker-barriocanal, line 946: I'd like to have a test covering this, no opinions on the type of the test. At this point, I don't feel confident in applying changes to how this whole logic works and expect it to work, so the more coverage we have for these changes the better.

iker-barriocanal (commented):
The logic looks good to me. There are a few improvements to reduce the duplicated code, the test comment I added below, and making CI green. Otherwise, ready to merge IMO.

relay-server/src/utils/rate_limits.rs
iker-barriocanal: Should the indexed transaction quota check run at a specific point in the pipeline? Quotas are checked twice throughout the processing of the envelope, and the indexed transaction quota is only checked the second time. I believe the behavior is correct.

relay-server/src/utils/rate_limits.rs
jan-auer: Sorry, that's what I meant.

relay-server/src/utils/rate_limits.rs
jan-auer: Across modules, we should build all APIs as public APIs - mostly to maintain high code quality.

relay-server/src/utils/rate_limits.rs
jan-auer: Makes sense, thank you.

relay-server/src/utils/rate_limits.rs
flub: Did this in https://github.com/getsentry/relay/pull/1517/commits/817c70355665ad93cd2bfd373f192835a349b946 but honestly, I don't think this is better from a code quality point of view:

- This is now no longer symmetric: there is only one field with an accessor function.  This is because the linter does not like unused methods (good!).  So now the user has to guess why only one of them is made available.  Weird.
- The caller now has `field.accessor().method()`, this is entirely subjective, but `field.field.method()` is faster to parse for a human because really what's written is `field.method().method()` and you have to mentally translate this to `field.accessor().method()` to understand what's going on.

But I don't mind, happy to leave it in.

relay-server/src/utils/rate_limits.rs
flub: We are testing this is the very next test though: `test_enforce_transaction_metrics_extracted_no_indexing_quota`

relay-server/src/utils/rate_limits.rs
flub: I still believe this is a correct assertion: When we have not yet extracted metrics we are in `check_envelope`.  At this point our design said to only ever check the `TransactionProcessed` quota and we never need to check `Transaction` quota.  So the fact that `Transaction` quota is out should not result in any enforcement.

tests/integration/test_store.py
flub: yep, adding this as a new test because modifying this one became too complicated.

relay-server/src/utils/rate_limits.rs
jan-auer: Updated the description to include the full algorithm. Viewed in isolation, the assertion here is correct, but this case should never be triggered due to the precondition: There must always be processing transaction quota that's also checked for legacy plans. In the implementation, so far we've not updated tests to reflect this. 

@flub and I have had an offline conversation about this and will circle around to this.

PR #1519, opened 2022-10-06
fix(quotas): Make redis rate limiter work with quantity 0 [INGEST-1656]

jjbayer:
For #1515, it is required to check for a required quota of another data category without incrementing it. This PR updates the Redis LUA script to support a rate limiting quantity of `0`, which checks for existing rate limits without incrementing internal counters.

The rate limiter gains a new explicit branch to check whether the quantity is `0`. We could have indirectly achieved the same effect with `max(quantity, 1)`.

### DIFF
--- a/relay-quotas/src/is_rate_limited.lua
+++ b/relay-quotas/src/is_rate_limited.lua
@@ -8,7 +8,7 @@
 -- ``ARGV`` (3 per quota):
 --  * [number] Quota limit. Can be ``-1`` for unlimited quotas.
 --  * [number] Absolute Expiration time as Unix timestamp (secs since 1.1.1970 ) for the key.
---  * [number] Quantity to increment the quota by.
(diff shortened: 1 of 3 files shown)


relay-quotas/src/is_rate_limited.lua
```diff
@@ -57,8 +63,10 @@ if not failed then
         local k = i * 2 + 1
         local v = i * 3 + 1
 
-        redis.call('INCRBY', KEYS[k], ARGV[v + 2])
-        redis.call('EXPIREAT', KEYS[k], ARGV[v + 1])
+        if tonumber(ARGV[v + 2]) > 0 then
```
jan-auer, line 66: Consider this an optimization. I'm assuming that the check is quite a bit faster than executing the two commands below.

flub (approved):

PR #1520, opened 2022-10-09
Add support for multi-extensions config

### LINKED ISSUE #1094: .yaml support
Please add `config.yaml` besides `config.yml` to config recognition and other YAML files. This is gonna save hundreds of hours debugging.

boozec:
Hi all :wave: 
with this PR config file with `.yaml` extension will be valid. If a `config.yaml` file exists, try to open that file. If not, try `config.yml`.

Why try `.yaml` instead of `.yml` first? Because if none exists and you run `credentials generate` the newly created credentials will be placed in `.yml`.

This resolves #1094

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -154,7 +154,7 @@ enum ConfigFormat {
 impl ConfigFormat {
     pub fn extensions(&self) -> Vec<&'static str> {
         match self {
-            ConfigFormat::Yaml => vec!["yml", "yaml"],
+            ConfigFormat::Yaml => vec!["yaml", "yml"],
             ConfigFormat::Json => vec!["json"],
         }
     }
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -168,8 +168,8 @@ trait ConfigObject: DeserializeOwned + Serialize {
     /// The basename of the config file.
     fn name() -> &'static str;
 
-    /// Check the full filename of the config file if it exists in the file storage. If none of
-    /// the defined files exist, it returns the last one even if it does not exist
+    /// The full filename of the config file, including the file extension.
+    /// The first matching file is returned, falling back to the last extension if none match.
     fn path(base: &Path) -> PathBuf {
         let extensions = Self::format().extensions();
 
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -152,10 +152,10 @@ enum ConfigFormat {
 }
 
 impl ConfigFormat {
-    pub fn extension(&self) -> &'static str {
+    pub fn extensions(&self) -> Vec<&'static str> {
         match self {
-            ConfigFormat::Yaml => "yml",
-            ConfigFormat::Json => "json",
+            ConfigFormat::Yaml => vec!["yml", "yaml"],
+            ConfigFormat::Json => vec!["json"],
         }
     }
 }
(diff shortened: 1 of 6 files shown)


relay-config/src/config.rs
```diff
@@ -152,10 +152,10 @@ enum ConfigFormat {
 }
 
 impl ConfigFormat {
-    pub fn extension(&self) -> &'static str {
+    pub fn extensions(&self) -> Vec<&'static str> {
         match self {
-            ConfigFormat::Yaml => "yml",
-            ConfigFormat::Json => "json",
+            ConfigFormat::Yaml => vec!["yaml", "yml"],
+            ConfigFormat::Json => vec!["json"],
```
iker-barriocanal: The size of these is small and won't change often, can we use `smallvec` instead? See https://docs.rs/smallvec/latest/smallvec/

Here's an example of its usage in the codebase:
https://github.com/dcariotti/relay/blob/926827a0f42ab9532faf521bb9818284389a3fca/relay-general/src/pii/attachments.rs#L28

relay-config/src/config.rs
```diff
@@ -169,7 +169,18 @@ trait ConfigObject: DeserializeOwned + Serialize {
 
     /// The full filename of the config file, including the file extension.
     fn path(base: &Path) -> PathBuf {
-        base.join(format!("{}.{}", Self::name(), Self::format().extension()))
+        let extensions = Self::format().extensions();
+
+        for extension in &extensions[..extensions.len() - 1] {
```
iker-barriocanal, line 176:
```suggestion
        for extension in extensions {
```

relay-config/src/config.rs
```diff
@@ -169,7 +169,18 @@ trait ConfigObject: DeserializeOwned + Serialize {
...
+            let path = base.join(format!("{}.{}", Self::name(), extension));
+
+            if path.exists() {
+                return path;
+            }
+        }
+
+        // It must returns something, so returns the last extension even if it does not exist
```
```suggestion
```
This should be documented in the function's docstrings instead.

relay-config/src/config.rs
```diff
@@ -169,7 +169,18 @@ trait ConfigObject: DeserializeOwned + Serialize {
 
     /// The full filename of the config file, including the file extension.
```
iker-barriocanal, line 171: Please update the documentation here too. It should reflect that the function returns the path to the first matching file, and falls back to the last extension if no files exist.

iker-barriocanal (changes requested):
Thanks for contributing!

Could you add tests to ensure the change works as intended, and a new entry to the changelog (see [instructions](https://github.com/getsentry/relay#instructions-for-changelogs))?

relay-config/src/config.rs
boozec: But in this case the last element couldn't be checked twice?

iker-barriocanal: You're right, there's no need to check the last element since the function falls back to it.

relay-config/src/config.rs
```diff
@@ -167,9 +168,20 @@ trait ConfigObject: DeserializeOwned + Serialize {
     /// The basename of the config file.
     fn name() -> &'static str;
 
-    /// The full filename of the config file, including the file extension.
+    /// Check the full filename of the config file if it exists in the file storage. If none of
+    /// the defined files exist, it returns the last one even if it does not exist
```
iker-barriocanal: By reading this comment it's not clear to me what the behavior is if a file exists (files are only checked?). What do you think about the following?
```suggestion
    /// The full filename of the config file, including the file extension. The first matching file is returned, falling back to the last extension if none match.
```

iker-barriocanal (commented):
Running CI, please fix any issues if a job fails.

boozec:
> Running CI, please fix any issues if a job fails.

Fixed black @iker-barriocanal :)

iker-barriocanal:
Hi @dcariotti, sorry for the iterations requesting changes but I've checked with the team and we won't be proceeding with the support of `.yaml`. We don't see a reason to support `yaml` while `yml` is the primary one. Please use `yml` instead, and thanks for your contributions.

PR #1521, opened 2022-10-10
feat(quotas): Optionally exceed quota once [INGEST-1654]

jjbayer:
For checks with `quantity=0`, we need the quota stored in redis to _exceed_ the rate limit to be executed. For this, it's necessary to _accept_ the first call of the rate limiter that exceeds the limit.

- [x] Test that describes desired behavior
- [x] Add an optional argument to lua script to over-accept.

```lua
 if quantity == 0 or over_accept_once then
      rejected = consumed >= limit
  else
      rejected = consumed + quantity > limit
  end
  ``` 

#skip-changelog

### DIFF
--- a/relay-quotas/src/is_rate_limited.lua
+++ b/relay-quotas/src/is_rate_limited.lua
@@ -17,7 +17,7 @@
 -- Send these values:
 --
 --     KEYS = {"foo", "foo_refund", "bar", "bar_refund"}
---     ARGV = {10, 600 + now(), 5, 20, 600 + now(), 1}
+--     ARGV = {10, 600 + now(), 5, false, 20, 600 + now(), 1, true}
 --
 -- The script applies the following logic:
 --  * If all checks pass, the item is accepted and the counters for all quotas
@@ -28,24 +28,36 @@
 -- The result is a Lua table/array (Redis multi bulk reply) that specifies
 -- whether or not the item was *rejected* based on the provided limit.
 assert(#KEYS % 2 == 0, "there must be 2 keys per quota")
-assert(#ARGV % 3 == 0, "there must be 3 args per quota")
-assert(#KEYS / 2 == #ARGV / 3, "incorrect number of keys and arguments provided")
+assert(#ARGV % 4 == 0, "there must be 4 args per quota")
+assert(#KEYS / 2 == #ARGV / 4, "incorrect number of keys and arguments provided")
+
+-- parase the incoming string into boolean, returns `nil` if could not parse which is also considered `false`
+local function parse_boolean(v)
+    if v == '1' or v == 'true' or v == 'TRUE' then
+        return true
+    elseif v == '0' or v == 'false' or v == 'FALSE' then
+        return false
+    else
+        return nil
+    end
+end
 
 local results = {}
 local failed = false
 local num_quotas = #KEYS / 2
 for i=0, num_quotas - 1 do
     local k = i * 2 + 1
-    local v = i * 3 + 1
+    local v = i * 4 + 1
 
     local limit = tonumber(ARGV[v])
     local quantity = tonumber(ARGV[v+2])
+    local over_accept_once = parse_boolean(ARGV[v+3])
     local rejected = false
     -- limit=-1 means "no limit"
     if limit >= 0 then
         local consumed = (redis.call('GET', KEYS[k]) or 0) - (redis.call('GET', KEYS[k + 1]) or 0)
         -- we never increment past the limit. if quantity is 0, check instead if we reached limit.
-        if quantity == 0 then
+        if quantity == 0 or over_accept_once then
             rejected = consumed >= limit
         else
             rejected = consumed + quantity > limit
@@ -61,7 +73,7 @@ end
 if not failed then
     for i=0, num_quotas - 1 do
         local k = i * 2 + 1
-        local v = i * 3 + 1
+        local v = i * 4 + 1
 
         if tonumber(ARGV[v + 2]) > 0 then
             redis.call('INCRBY', KEYS[k], ARGV[v + 2])
--- a/relay-quotas/src/is_rate_limited.lua
+++ b/relay-quotas/src/is_rate_limited.lua
@@ -5,7 +5,7 @@
 --  * [string] Key of the counter.
 --  * [string] Key of the refund counter.
 --
--- ``ARGV`` (3 per quota):
+-- ``ARGV`` (4 per quota):
 --  * [number]  Quota limit. Can be ``-1`` for unlimited quotas.
 --  * [number]  Absolute Expiration time as Unix timestamp (secs since 1.1.1970 ) for the key.
 --  * [number]  Quantity to increment the quota by, or ``0`` to check without incrementing.
@@ -34,9 +34,9 @@ assert(#KEYS / 2 == #ARGV / 4, "incorrect number of keys and arguments provided"
 
 -- parse the incoming string into boolean, returns `nil` if could not parse which is also considered `false`
 local function parse_boolean(v)
-    if v == '1' or v == 'true' or v == 'TRUE' then
+    if v == 'true' then
         return true
-    elseif v == '0' or v == 'false' or v == 'FALSE' then
+    elseif v == 'false' then
         return false
     else
         return nil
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: relay-quotas/src/is_rate_limited.lua [hunk 1]
```diff
@@ -17,7 +17,7 @@
 -- Send these values:
 --
 --     KEYS = {"foo", "foo_refund", "bar", "bar_refund"}
---     ARGV = {10, 600 + now(), 5, 20, 600 + now(), 1}
+--     ARGV = {10, 600 + now(), 5, false, 20, 600 + now(), 1, true}
```
jjbayer, line 21: Let's also add the new arg to the description starting on line 8.

relay-quotas/src/is_rate_limited.lua
```diff
@@ -28,24 +28,36 @@
 -- The result is a Lua table/array (Redis multi bulk reply) that specifies
 -- whether or not the item was *rejected* based on the provided limit.
 assert(#KEYS % 2 == 0, "there must be 2 keys per quota")
-assert(#ARGV % 3 == 0, "there must be 3 args per quota")
-assert(#KEYS / 2 == #ARGV / 3, "incorrect number of keys and arguments provided")
+assert(#ARGV % 4 == 0, "there must be 4 args per quota")
+assert(#KEYS / 2 == #ARGV / 4, "incorrect number of keys and arguments provided")
+
+-- parase the incoming string into boolean, returns `nil` if could not parse which is also considered `false`
```
jjbayer:
```suggestion
-- parse the incoming string into boolean, returns `nil` if could not parse which is also considered `false`
```

relay-quotas/src/is_rate_limited.lua
```diff
@@ -28,24 +28,36 @@
...
+    local v = i * 4 + 1
 
     local limit = tonumber(ARGV[v])
     local quantity = tonumber(ARGV[v+2])
+    local over_accept_once = parse_boolean(ARGV[v+3])
     local rejected = false
     -- limit=-1 means "no limit"
     if limit >= 0 then
         local consumed = (redis.call('GET', KEYS[k]) or 0) - (redis.call('GET', KEYS[k + 1]) or 0)
         -- we never increment past the limit. if quantity is 0, check instead if we reached limit.
```
jjbayer:
```suggestion
        -- Without over_accept_once, we never increment past the limit. if quantity is 0, check instead if we reached limit.
        -- With over_accept_once, we only reject if the previous update already reached the limit. This way, we ensure that we increment to or past the limit at some point, such that subsequent checks with quantity=0 are actually rejected.
```

jjbayer (commented):
Some suggestions to improve documentation, other than that, this looks good!

relay-quotas/src/is_rate_limited.lua
olksdr: done in [6875902](https://github.com/getsentry/relay/pull/1521/commits/6875902fb62f52263758c3711f8e898777003bcc)

relay-quotas/src/is_rate_limited.lua
olksdr: done in [6875902](https://github.com/getsentry/relay/pull/1521/commits/6875902fb62f52263758c3711f8e898777003bcc)

relay-quotas/src/is_rate_limited.lua
olksdr: done in [6875902](https://github.com/getsentry/relay/pull/1521/commits/6875902fb62f52263758c3711f8e898777003bcc)

relay-quotas/src/is_rate_limited.lua
```diff
@@ -6,9 +6,10 @@
 --  * [string] Key of the refund counter.
 --
 -- ``ARGV`` (3 per quota):
```
iker-barriocanal:
```suggestion
-- ``ARGV`` (4 per quota):
```

relay-quotas/src/is_rate_limited.lua
```diff
@@ -28,24 +29,39 @@
 -- The result is a Lua table/array (Redis multi bulk reply) that specifies
 -- whether or not the item was *rejected* based on the provided limit.
 assert(#KEYS % 2 == 0, "there must be 2 keys per quota")
-assert(#ARGV % 3 == 0, "there must be 3 args per quota")
-assert(#KEYS / 2 == #ARGV / 3, "incorrect number of keys and arguments provided")
+assert(#ARGV % 4 == 0, "there must be 4 args per quota")
+assert(#KEYS / 2 == #ARGV / 4, "incorrect number of keys and arguments provided")
+
+-- parse the incoming string into boolean, returns `nil` if could not parse which is also considered `false`
+local function parse_boolean(v)
```
iker-barriocanal: Can we agree to a protocol and not do these operations? They run quite often and it's expensive. For example, we could use `t` as truthy value and consider it falsy otherwise.

relay-quotas/src/redis.rs
```diff
@@ -176,6 +180,7 @@ impl RedisRateLimiter {
         quotas: &[Quota],
         item_scoping: ItemScoping<'_>,
         quantity: usize,
+        over_accept_once: bool,
```
iker-barriocanal, line 183: nit: using a boolean makes reading the code calling this function harder to read, could we use an enum instead? An example is available in the `processor.rs` file in this diff. Example:

```rust
enum Overaccept {
    Once,
    No,
}
```

relay-quotas/src/redis.rs
```diff
@@ -394,29 +400,82 @@ mod tests {
...
+                organization_id: 42,
+                project_id: ProjectId::new(43),
+                project_key: ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap(),
+                key_id: Some(44),
+            },
+        };
+
+        let rate_limiter = build_rate_limiter();
+
+        // limit is 1, so first call not rate limited
```
iker-barriocanal:
```suggestion
        // limit is 2, so first call not rate limited
```

relay-quotas/src/is_rate_limited.lua
jjbayer: Good idea, we could check for the string `"true"` instead (assuming that is what the rust redis library converts the argument to).

relay-quotas/src/redis.rs
jjbayer: I get your point, but you could apply this reasoning to any boolean function arg. I'd rather keep it a bool and rely on IDEs to annotate the arguments you're passing.

flub: this is totally my doing.  and yes, i'd do this to any boolean arg :).

relay-quotas/src/is_rate_limited.lua
olksdr: I reduced this only to `"true"`  and `"false"` but it does not really matter, we still have to have comparison to the string to get the boolean from that, either we assign this to the variable or do it in the if, there still will be that additional operation.  IMHO  having function and variable is much cleaner.

relay-quotas/src/redis.rs
olksdr: Agreed with @jjbayer , this is overkill in here.  If we had 3 and more values we want to accept that we could use enum for that, right now it's only 1 function with one marker flag.

relay-quotas/src/is_rate_limited.lua
```diff
@@ -28,24 +29,39 @@
...
+    if v == 'true' then
+        return true
+    elseif v == 'false' then
+        return false
+    else
+        return nil
```
flub: why does this nil branch exist?  it's used directly in a boolean expression which i assume will be treated like false.  Might as well simplify to `if ... true else false`

relay-quotas/src/is_rate_limited.lua
flub: go lisp-style and use `t` for faster `strcmp` results :)

i'm not sure what the overheads are of a function in lua, but it is used only once to assign to a variable (which i agree is clearer to read) so could as well write the if statement right there.

relay-quotas/src/redis.rs
flub: one day you too will despise boolean args.  but i'm not going to die on this hill today :)

flub (approved):

PR #1522, opened 2022-10-11
meta(vscode): Configure rust import preferences

jan-auer:
Configures Rust Analyzer to follow the preferred import style when organizing
imports, mostly during auto-import:

 - Group imports by module path, emitting one line per module.
 - Prefix imports from the local crate with `crate::` instead of using `super`.

Note that Rust Analyzer is not configured to rewrite existing imports, and it
will try to match the import style of the files being edited.

This also enables clippy as check command so that clippy warnings show up 
while editing. This was likely already enabled in most local development
environments.

#skip-changelog

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,6 +14,11 @@
   "python.linting.enabled": true,
   "python.formatting.provider": "black",
 
+  // Rust Analyzer
+  "rust-analyzer.checkOnSave.command": "clippy",
+  "rust-analyzer.imports.granularity.group": "module",
+  "rust-analyzer.imports.prefix": "crate",
+
   // Language-specific overrides
   "[markdown]": {
     "editor.rulers": [80],


olksdr (approved):
This is very nice :shipit: !

It also would be great to set the `rustfmt.toml` config for `rustfmt`

flub (approved):
This probably doesn't work for all those who don't use vscode, but at least there's a snippet to point at.  nice!

jan-auer:
In the [Development](https://github.com/getsentry/relay#development) section we mention VSCode as the primary editor and encourage regular contributors to use that, so configuration in this repo will be biased towards that. 

`group_imports` in rustfmt isn't stable yet, unfortunately, so that feature comes directly from Rust Analyzer during auto-import. For rustfmt, we use all defaults without overrides. We can consider adding rustfmt.yml when this feature is stable.

PR #1525, opened 2022-10-13
ref(protocol): Split contexts.rs into separate files

jjbayer:
Create a separate file for every `Context`, such that they are easier to navigate. No functional changes in this PR.


olksdr (approved):

jan-auer (approved):
🍪

PR #1526, opened 2022-10-13
ref: Use wildcard re-exports where applicable

jan-auer:
In some cases, we export the entire public API of a module via the containing
parent module. The internal module structure is supposed to be an implementation
detail for code organization but does not matter to the public API. An example
for this is the Event protocol.

In most cases, we use wildcard exports `pub use module::*` for this. The
`relay-general` crate did not do this. With modern tooling and code search, it
is now equally easy to locate types. The benefit of using a wildcard export is
that we cannot forget to re-export types that would be then missing from the
public signature.

#skip-changelog


relay-metrics/src/protocol.rs
```diff
@@ -5,11 +5,10 @@ use std::iter::FusedIterator;
 use hash32::{FnvHasher, Hasher};
 use serde::{Deserialize, Serialize};
 
-pub use relay_common::UnixTimestamp;
-
 #[doc(inline)]
 pub use relay_common::{
     CustomUnit, DurationUnit, FractionUnit, InformationUnit, MetricUnit, ParseMetricUnitError,
+    UnixTimestamp,
```
jjbayer, line 11: Was there a reason this was not under `#[doc(inline)]` before?

jjbayer (approved):
LGTM!

olksdr (approved):
:shipit:

relay-metrics/src/protocol.rs
jan-auer: I don't think so, looks like an oversight.

Issue #1501, opened 2022-09-27
Data scrubbing should filter sensitive headers such as `Set-Cookie`


marandaneto:
[OkHttp defines](https://github.com/square/okhttp/blob/58ee1ce170c2aea75a113012956260873e808dad/okhttp/src/commonMain/kotlin/okhttp3/internal/-UtilCommon.kt#L194-L200) its own sensitive headers.
Some of them are filtered by Relay already, but not all of them.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie

Not all HTTP clients respect that `Set-Cookie` should not be accessible.

`Cookie` is just removed from the headers list instead of being `[Filtered]`, most likely because there's a high-level property `Request#cookies` already.

<img width="655" alt="Screenshot 2022-09-27 at 10 46 02" src="https://user-images.githubusercontent.com/5731772/192478851-207e7d81-df89-4ba4-b118-035afa925f35.png">

untitaker:
the request interface of an event is for http request headers, set-cookie is a response header. generally we are filtering header names iirc

marandaneto:
We're trying to use the `request` property for all the merged headers (request, response), but since there's this limitation plus a few other ones, I will reconsider that.
I'll leave that up to you if you still wanna run scrubbing or not, there's also the inconsistency of `Cookie` being removed instead of filtered.

jjbayer:
I suggest we put this hold until we've completed the RFC on where to put response data in the event payload.

untitaker:
>We're trying to use the request property for all the merged headers (request, response), but since there's this limitation plus a few other ones, I will reconsider that.

I would definitely not do that. Which RFC are we talking about?

marandaneto:
@untitaker I'm going to write an RFC that adds `Event.response` or `Event.contexts.response`, data structure similar to `event.request` but with additional fields such as `status_code`, etc...
Data scrubbing would need to give a pass in the `headers` field, etc.

untitaker:
@marandaneto sounds good. there is potential for a lot of pii there so i think we should think about designing new default scrubbing rules in the same vein.

marandaneto:
> @marandaneto sounds good. there is potential for a lot of pii there so i think we should think about designing new default scrubbing rules in the same vein.

https://github.com/getsentry/rfcs/pull/22

marandaneto:
[...]

PR #1529, opened 2022-10-14
feat(protocol): Implement response context schema [INGEST-1661]

### LINKED ISSUE #1530: Implement Response protocol
https://github.com/getsentry/rfcs/blob/main/text/0022-response-context.md

Add an additional field to `Request` protocol as well, `bodySize`.

There's a JIRA issue already, this is just a meta issue for the Mobile SDKs, we only use GH issues.

olksdr:
Add `Response` context schema and add `body_size` field to the request. 

closes: https://github.com/getsentry/relay/issues/1530


relay-general/src/protocol/contexts/response.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[metastructure(skip_serialization = "empty")]
+    pub headers: Annotated<Headers>,
+
+    /// HTTP status code.
+    pub status_code: Annotated<u64>,
+
+    /// HTTP response body size.
+    pub body_size: Annotated<u64>,
+
+    /// Additional arbitrary fields for forwards compatibility.
```
jjbayer, line 29: Just adding some context here:

```suggestion
    /// Additional arbitrary fields for forwards compatibility.
    ///
    /// These fields are retained (`retain = "true"`) to keep supporting the format that the Dio integration sends:
    /// https://github.com/getsentry/sentry-dart/blob/7011abe27ac69bd160bdc6ecf3314974b8340b97/dart/lib/src/protocol/sentry_response.dart#L4-L8
```

jjbayer (commented):
Nice, thanks for picking this up! Can we add a test case to verify that the following headers are successfully stripped by pii scrubbing?

```
Authorization
Proxy-Authorization
Set-Cookie
```

For more context, see https://github.com/getsentry/relay/issues/1501. That issue is closed but the headers mentioned are still relevant, `Set-Cookie` in particular.

relay-general/src/protocol/contexts/response.rs
```diff
@@ -0,0 +1,42 @@
+use crate::protocol::{Cookies, Headers};
+use crate::types::{Annotated, Object, Value};
+
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct ResponseContext {
+    /// Type `response`.
+    #[metastructure(field = "type")]
+    pub ty: Annotated<String>,
```
jan-auer: The type is injected by the `Contexts` wrapper, so this should not be added as a field here.

relay-general/src/protocol/contexts/response.rs
```diff
@@ -0,0 +1,42 @@
+use crate::protocol::{Cookies, Headers};
+use crate::types::{Annotated, Object, Value};
+
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct ResponseContext {
```
jan-auer, line 7: Please document this type.

relay-general/src/pii/regexes.rs
```diff
@@ -269,6 +269,6 @@ static US_SSN_REGEX: Lazy<Regex> = Lazy::new(|| {
 
 static PASSWORD_KEY_REGEX: Lazy<Regex> = Lazy::new(|| {
     Regex::new(
-        r"(?i)(password|secret|passwd|api_key|apikey|auth|credentials|mysql_pwd|privatekey|private_key|.*token.*)"
+        r"(?i)(password|secret|passwd|api_key|apikey|auth|credentials|mysql_pwd|privatekey|private_key|set-cookie|.*token.*)"
```
jjbayer: Something I missed on the previous PR: If `auth` matches `Proxy-Authorization`, then `token` matches `whatever_token_123` ... so I think we can omit the `.*`
```suggestion
        r"(?i)(password|secret|passwd|api_key|apikey|auth|credentials|mysql_pwd|privatekey|private_key|set-cookie|token)"
```

jjbayer (approved):

relay-general/src/pii/regexes.rs
olksdr: good point! done in [8e37424](https://github.com/getsentry/relay/pull/1529/commits/8e37424ff871ea3e8dfc5ffd50c6b8dd86ce7a63)

relay-general/src/store/normalize/contexts.rs
```diff
@@ -177,10 +177,33 @@ fn normalize_os_context(os: &mut OsContext) {
...
 }
 
+fn normalize_response(response: &mut ResponseContext) {
+    let headers = match response.headers.value_mut() {
+        Some(headers) => headers,
+        None => return,
+    };
+
+    if response.cookies.value().is_some() {
+        headers.remove("Cookie");
```
jjbayer: Since `Cookie` is not a response header, I think we should use `Set-Cookie` here.

@marandaneto was the intention of the `cookies` field to contain the parsed contents of the `Set-Cookie` header?

marandaneto: Not really, I just noticed that they were not scrubbed and opened the issue, but they were response headers and I was looking into the request headers, so it made sense.
Your call if it should be removed or not due to PII concerns.

olksdr: @jjbayer  pointed that `Cookie` is the request header and is not expected in the response.  We can just follow pattern and do what was proposed above, remove `set-cookie` header and convert those values to `cookies` when needed.

olksdr:
@jjbayer  if you wanna take another look just to double check if everything is good to go here.

jjbayer (approved):
Perfect, thanks!

marandaneto:
@olksdr is this going to be automatically reflected in https://develop.sentry.dev/sdk/event-payloads/types/#event ?

untitaker:
> @olksdr is this going to be automatically reflected in [develop.sentry.dev/sdk/event-payloads/types/#event](https://develop.sentry.dev/sdk/event-payloads/types/#event) ?

yes

untitaker:
corresponding commit in https://github.com/getsentry/develop/commit/6a7e1025cfdb846cb6fdbddcc1f485c817fba5ff

PR #1531, opened 2022-10-17
test(auth): Unflake re_auth_failure

jjbayer:
It seems that in `test_re_auth_failure`, Relay was able to process a new event before it changed its `auth_state` from `Registered` to `Registering`. The integration test waits for the sentry side to handle the failing auth request, but not for the Relay side to handle the failing auth _response_.

#skip-changelog

### DIFF
--- a/tests/integration/test_store.py
+++ b/tests/integration/test_store.py
@@ -891,7 +891,7 @@ def test_re_auth_failure(relay, mini_sentry):
     assert auth_count_1 < auth_count_2
 
     # Give Relay some time to process the auth response and mark itself as not ready
-    sleep(1)
+    sleep(0.1)
 
     # send a message, it should not come through while the authentication has failed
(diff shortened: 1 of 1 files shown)


tests/integration/test_store.py
```diff
@@ -906,6 +909,9 @@ def counted_check_challenge(*args, **kwargs):
     auth_count_3 = counter[0]
     assert auth_count_2 < auth_count_3
 
+    # Give Relay some time to process the auth response and mark itself as ready
+    sleep(0.1)
```
jjbayer, line 913: Ran this 10 times on a different branch, `0.1` seems to work fine:

https://github.com/getsentry/relay/actions/runs/3318797256/jobs/5483154174#step:10:662

iker-barriocanal (approved):

PR #1535, opened 2022-10-20
feat: New data category for indexed transactions [INGEST-1535]

jjbayer:
Add a new data category for indexed transactions, that is, transactions that are stored.

This implements [Option 2](https://www.notion.so/sentry/Dynamic-Sampling-Packaging-v3-ec32aaeb6c054fe6930a8e6f519f634f#d69f3cea151c4c288917f45756708f3c) of the Dynamic Sampling Packaging V3 DACI.

The mapping from old to new looks as follows:

* `transaction` -> `transaction_indexed`
* `transaction_processed` -> `transaction`

PR #1537, opened 2022-10-20
feat(quotas): Separate processing and indexing quotas

jan-auer:
Once metrics have been extracted, transaction events need to have the
new `DataCategory::TransactionIndexed` quota available to ingest and
store the payload.

The new indexing quota replaces the previous `DataCategory::Transaction`
for transaction payloads. Transaction quotas now cover all ingested
transactions that have metrics extracted. Transaction quota however is
only counted in the metrics part and not in event ingestion.

# Background

Rate limiting and quota enforcement has to apply to _legacy_ plans that
did not have processed transaction functionality, and to _new_ plans
that introduce them.

### Organization on Legacy Plan

There is a single quota for total transactions. We need to extract
metrics for indexed transactions so that there is backfilled history at
the time the customer switches to the new plan. This requires to extract
metrics for all accepted transactions.

### Organization on New Plan

This plan has a dedicated additional quota for indexed transactions.

- Validation and inbound data filters do not consume any quota.
  There are no metrics extracted for such dropped events.
- If there is transaction quota available, extract and ingest metrics for
  this transaction. Otherwise, drop the transaction event and any
  transaction metrics bucket.
- Dynamic sampling may drop (previously processed) transaction events.
  Extracted metrics and the transaction quotas are not
  impacted by this.
- If there is additional indexing quota available, store the
  sampled transaction event payload. Otherwise, drop the transaction event.

# Algorithm

**Flow:**

1. Check if metrics transaction is enabled. 
   1. If disabled, rate limit the transaction item with the `transaction` data category and exit.
   2. Otherwise, continue with step 2.
2. Check for transaction quota but do not increment it.
   1. If quota is exceeded, drop the item and all metrics. 
   2. Otherwise, continue with step 3.
3. Check if metrics have been extracted for this transaction yet.
   1. If not, skip to step X
   2. Otherwise, keep the item. Metrics can be extracted now.
4. Metrics have been extracted. Check for indexing quota.
   1. If quota is exceeded, drop the item but keep metrics.
   2. Otherwise, keep the item. The transaction should be stored.

Transaction quota is incremented by the extracted metrics.

# References

- Requires https://github.com/getsentry/relay/pull/1535
- Requires https://github.com/getsentry/getsentry/pull/8642
- Replaces https://github.com/getsentry/relay/pull/1517


relay-server/src/utils/rate_limits.rs
```diff
@@ -246,9 +255,23 @@ pub struct Enforcement {
     profiles: CategoryLimit,
     /// The combined replay item rate limit.
     replays: CategoryLimit,
+    /// Metrics extraction from a transaction is rate limited.
+    event_metrics: CategoryLimit,
 }
 
 impl Enforcement {
+    /// Returns true if the event should be rate limited.
```
```suggestion
    /// Returns `true` if the event should be rate limited.
```

to keep it consistent

relay-server/src/utils/rate_limits.rs
```diff
@@ -152,6 +157,10 @@ impl EnvelopeSummary {
                 summary.has_plain_attachments = true;
             }
 
+            if *item.ty() == ItemType::Transaction && item.metrics_extracted() {
```
jjbayer, line 160: nit: As the field is now called `event_metrics_extracted`, should we allow item type `Event` here as well?

relay-server/src/utils/rate_limits.rs
```diff
@@ -246,9 +255,23 @@ pub struct Enforcement {
...
 impl Enforcement {
+    /// Returns true if the event should be rate limited.
+    pub fn event_active(&self) -> bool {
+        self.event.is_active()
+    }
+
+    /// Returns `true` if metrics extracted from the event should be rate limited.
+    #[cfg(feature = "processing")]
+    pub fn event_metrics_active(&self) -> bool {
+        // TODO: Rename this
```
jjbayer:
```suggestion
```

relay-server/src/utils/rate_limits.rs
```diff
@@ -378,9 +425,29 @@ where
...
-            enforcement.event = CategoryLimit::new(category, 1, longest);
+            let mut longest;
+            let mut event_limits;
+
+            if let Some(index_category) = self.index_category(category) {
+                // Check for rate limits on the main category (e.g. transaction) but do not consume
+                // quota. Quota will be consumed by metrics in the metrics aggregator instead.
+                event_limits = (self.check)(scoping.item(category), 0)?;
+                longest = event_limits.longest();
+                enforcement.event_metrics = CategoryLimit::new(category, 1, longest);
```
jjbayer, line 436: This `1` is only here such that `is_active()` evaluates to `true`, right? It might as well be `2` or `999`.

jjbayer (approved):

relay-cabi/include/relay.h
```diff
@@ -49,9 +49,17 @@ enum RelayDataCategory {
...
-   * A transaction that was processed but not stored.
+   * DEPRECATED: A transaction for which metrics were extracted.
+   * This category is now obsolete because the `Transaction` variant will represent
+   * processed transactions from now on.
    */
   RELAY_DATA_CATEGORY_TRANSACTION_PROCESSED = 8,
+  /**
+   * Indexed transaction events.
+   * This is the category for transaction payloads that were stored, that is, they were not
+   * filtered out by Dynamic Sampling.
```
iker-barriocanal:
```suggestion
   * Indexed transaction events.
   * This is the category for transaction payloads that were stored, that is, they were not
   * filtered out by filters such as Dynamic Sampling.
```
Dynamic sampling is not the only mechanism to avoid storing transactions.

relay-common/src/constants.rs
```diff
@@ -113,8 +113,14 @@ pub enum DataCategory {
...
     /// Session Replays
     Replay = 7,
-    /// A transaction that was processed but not stored.
+    /// DEPRECATED: A transaction for which metrics were extracted.
+    /// This category is now obsolete because the `Transaction` variant will represent
+    /// processed transactions from now on.
     TransactionProcessed = 8,
+    /// Indexed transaction events.
+    /// This is the category for transaction payloads that were stored, that is, they were not
+    /// filtered out by Dynamic Sampling.
```
iker-barriocanal: Same suggestion as the previous.

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -51,13 +51,13 @@ impl BucketLimiter {
                 }
 
                 if mri.name == "duration" {
-                    // The "duration" metric is extracted exactly once for every processed transaction,
-                    // so we can use it to count the number of transactions.
+                    // The "duration" metric is extracted exactly once for every processed
+                    // transaction, so we can use it to count the number of transactions.
```
iker-barriocanal, line 51: Since this PR also implements the renaming dance of transactions, indexed transactions, and processed transactions:
```suggestion
                    // The "duration" metric is extracted exactly once for every
                    // transaction, so we can use it to count the number of transactions.
```

relay-server/src/utils/rate_limits.rs
```diff
@@ -246,9 +255,23 @@ pub struct Enforcement {
...
 
 impl Enforcement {
+    /// Returns true if the event should be rate limited.
+    pub fn event_active(&self) -> bool {
+        self.event.is_active()
+    }
+
+    /// Returns `true` if metrics extracted from the event should be rate limited.
+    #[cfg(feature = "processing")]
+    pub fn event_metrics_active(&self) -> bool {
```
iker-barriocanal, line 271: The reason to enable the processing flag here is for optimization reasons, right? I see it's only used in `enforce_quotas`.

relay-server/src/utils/rate_limits.rs
iker-barriocanal: IIRC, no; the `1` is because there's only one transaction. There's no case in which we use any other number than `1`, but `0` in this PR. You could argue that there's no point in having the API designed that way, and that's a valid point.

relay-server/src/utils/rate_limits.rs
```diff
@@ -378,9 +425,29 @@ where
...
+            if let Some(index_category) = self.index_category(category) {
+                // Check for rate limits on the main category (e.g. transaction) but do not consume
+                // quota. Quota will be consumed by metrics in the metrics aggregator instead.
+                event_limits = (self.check)(scoping.item(category), 0)?;
+                longest = event_limits.longest();
+                enforcement.event_metrics = CategoryLimit::new(category, 1, longest);
+
+                // If the main category is rate limited, we drop both the event and metrics. If
+                // there's no rate limit, check for specific indexing quota and drop just the event.
+                if summary.event_metrics_extracted && longest.is_none() {
```
iker-barriocanal, line 440: I want to double-check if I understand this correctly. The first time we get here is from the "fast path", and metrics aren't extracted at that point; which means we don't care about indexing transactions for now. Then, we process the envelope, which includes the metric extraction. The second time we get here, in `enforce_quotas` iirc, the envelope have almost been fully processed but the transaction isn't indexed yet, so we do care about extracting metrics and we'd get inside this `if` branch.

Is this correct?

relay-server/src/utils/rate_limits.rs
```diff
@@ -850,4 +936,110 @@ mod tests {
...
+    fn test_enforce_transaction_no_indexing_quota() {
+        let mut envelope = envelope![Transaction];
+        let config = config_with_tx_metrics();
+
+        let mut mock = MockLimiter::default().deny(DataCategory::TransactionIndexed);
+        let limiter = EnvelopeLimiter::new(Some(&config), |s, q| mock.check(s, q));
+        let (enforcement, limits) = limiter.enforce(&mut envelope, &scoping()).unwrap();
+
+        // NOTE: Since metrics have not been extracted on this item, we do not check the indexing
+        // quota. Basic processing quota is not denied, so the item must pass rate limiting.
```
iker-barriocanal: Is it worth noting that the transaction is rate limited at a later point? Your call.

relay-server/src/utils/rate_limits.rs
```diff
@@ -850,4 +936,110 @@ mod tests {
...
+        let (enforcement, _limits) = limiter.enforce(&mut envelope, &scoping()).unwrap();
+
+        assert!(enforcement.event.is_active());
+        assert!(enforcement.attachments.is_active());
+        mock.assert_call(DataCategory::Transaction, Some(0));
+        mock.assert_call(DataCategory::Attachment, None);
+    }
+
+    #[test]
+    fn test_enforce_transaction_attachment_enforced_metrics_extracted() {
```
iker-barriocanal:
```suggestion
    fn test_enforce_transaction_attachment_enforced_metrics_extracted_indexing_quota() {
```
not sure? To be consistent with other test names.

relay-server/src/utils/rate_limits.rs
```diff
@@ -850,4 +936,110 @@ mod tests {
...
+
+    #[test]
+    fn test_enforce_transaction_attachment_enforced() {
+        let mut envelope = envelope![Transaction, Attachment];
+        let config = config_with_tx_metrics();
+
+        let mut mock = MockLimiter::default().deny(DataCategory::Transaction);
+        let limiter = EnvelopeLimiter::new(Some(&config), |s, q| mock.check(s, q));
+
+        let (enforcement, _limits) = limiter.enforce(&mut envelope, &scoping()).unwrap();
```
iker-barriocanal, line 1021: Why don't we check the limits in this test and in the one below?

tests/integration/test_store.py
```diff
@@ -649,10 +649,70 @@ def send_buckets(buckets):
...
+
+    metrics_consumer = metrics_consumer()
+    tx_consumer = transactions_consumer()
+
+    project_id = 42
+    projectconfig = mini_sentry.add_full_project_config(project_id)
+    key_id = mini_sentry.get_dsn_public_key_configs(project_id)[0]["numericId"]
+    projectconfig["config"]["quotas"] = [
+        {
+            "id": "test_rate_limiting_{}".format(uuid.uuid4().hex),
```
iker-barriocanal: nit: f-strings
```suggestion
            "id": f"test_rate_limiting_{uuid.uuid4().hex}",
```

tests/integration/test_store.py
```diff
@@ -649,10 +649,70 @@ def send_buckets(buckets):
...
+            "scope": "key",
+            "scopeId": six.text_type(key_id),
+            "categories": ["transaction_indexed"],
+            "limit": 1,
+            "window": 300,
+            "reasonCode": "get_lost",
+        },
+        {
+            "id": "test_rate_limiting_{}".format(uuid.uuid4().hex),
```
iker-barriocanal: nit: f-strings
```suggestion
            "id": f"test_rate_limiting_{uuid.uuid4().hex},
```

relay-server/src/utils/rate_limits.rs
jan-auer: We would probably also have to do this for `Security` and all other event-like categories. I'm not sure exactly how we'd want those to behave yet, which is why I would stick with just `Transaction` for the detection.

relay-cabi/include/relay.h
jan-auer: Let me actually rewrite this. Dynamic sampling isn't really a filter -- and inbound filters would affect both data categories alike. The new text is going to be:

```
    /// Indexed transaction events.
    ///
    /// This is the category for transaction payloads that were accepted and stored in full. In
    /// contrast, `transaction` only guarantees that metrics have been accepted for the transaction.
```

relay-server/src/utils/rate_limits.rs
jan-auer: The compiler would also complain that this function is dead code.

relay-server/src/utils/rate_limits.rs
jan-auer: This test is mostly about asserting correct calls to the inner rate limiter and there are already tests that then check the response behavior. We can keep this test to a single concern with fewer assertions.

relay-server/src/utils/rate_limits.rs
jan-auer: This is basically it. The original idea was to create a non-optional type that we assign unconditionally, and that carries all information about the enforced rate limit. The only exposed method ended being `is_active`, which checks if we have a non-zero quantity. It's factually correct here, since we are in fact limiting `1` transaction.

olksdr (approved):
still lgtm

relay-server/src/utils/rate_limits.rs
iker-barriocanal: Hmmm, smart compiler.

iker-barriocanal (approved):

PR #1538, opened 2022-10-21
chore(redis): Update redis and r2d2 (pooling) crates

olksdr:
Updating dependencies to the latest ones:
* r2d2 -> `0.8.10`
* redis -> `0.22.1` 

#skip-changelog


relay-quotas/src/is_rate_limited.lua
```diff
@@ -50,7 +51,7 @@ for i=0, num_quotas - 1 do
         -- With over_accept_once, we only reject if the previous update already reached the limit. 
         -- This way, we ensure that we increment to or past the limit at some point,
         -- such that subsequent checks with quantity=0 are actually rejected.
-        if quantity == 0 or over_accept_once == 'true' then
```
iker-barriocanal, line 53: Why is this changing?

relay-redis/Cargo.toml
```diff
@@ -11,8 +11,8 @@ publish = false
 
 [dependencies]
 failure = "0.1.8"
-r2d2 = { version = "0.8.8", optional = true }
-redis = { version = "0.15.1", optional = true, features = ["cluster", "r2d2"] }
+r2d2 = { version = "0.8.10", optional = true }
+redis = { version = "0.22.1", optional = true, features = ["cluster", "r2d2"] }
```
iker-barriocanal, line 15: I assume no breaking changes.

relay-quotas/src/is_rate_limited.lua
olksdr: Because new version of the lib set the boolean differently `true` -> `'1'` in this case.  in was changed in version [`0.18.0`](https://github.com/redis-rs/redis-rs/blob/main/redis/CHANGELOG.md#0180-2020-12-03)

That's why I wanted to add [`parse` a function](https://github.com/getsentry/relay/pull/1521/commits/3253c7201d34c06b5af3c3acdea1748e0810f7ea#diff-d341d4f91000b4fd4d74af358f091d3331b9a143a39cd00a7177a7108c2c15d5R35)  in my previous PR to make this a bit more robust.

relay-redis/Cargo.toml
olksdr: There are some breaking changes, yes. This is quite big jump between `0.15.1` -> `0.22.1`, but looks like there is nothing which could concern us. 

But there are lots of bug fixes as well.

relay-quotas/src/is_rate_limited.lua
jjbayer: Is this the only place where we pass parameters into a Lua script?

olksdr: As far as I see, all the invocations are done in `relay-quotas/src/redis.rs`  where we prepare and pass all the parameters to this script

jjbayer (approved):
I wonder if the integration tests in CI timed out because some redis calls now take longer or even block? If not, I think we can risk merging this.

relay-quotas/src/is_rate_limited.lua
iker-barriocanal: I think it's beneficial to add a comment to note that.

iker-barriocanal (approved):
LGTM. I'd wait to merge (and deploy and monitor) this PR until Monday, in case there are any last-minute relay deployments.

olksdr:
@jjbayer it's actually cache takes too long 
image

usually it takes seconds to restore and but here it takes like 5 - 6 min 🤷‍♂️  and there is no time for running the tests

relay-quotas/src/is_rate_limited.lua
olksdr: @iker-barriocanal  added a note in [4d0ca6b](https://github.com/getsentry/relay/pull/1538/commits/4d0ca6b2d8d586d2cbad897f9f5471f80043ca7f)

PR #1540, opened 2022-10-24
feat(quotas): Rate limit metrics before aggregator [INGEST-1655]

jjbayer:
### Background

In https://github.com/getsentry/relay/pull/1515 we implemented rate limiting functionality for metrics buckets, which are applied after flushing them from the metrics aggregator by checking redis.

### This PR

When a quota is already exhausted, and that information has already been cached on the project state, there is no need to go through the aggregator at all. Instead, we apply the rate limiting logic _before_ sending metrics or buckets into the aggregator.

### Implementation details

The utility type `BucketLimiter` was converted to a generic `MetricsLimiter<T>`, where `T` can be either `Metric` or `Bucket`.

### Notes

Will be merged into https://github.com/getsentry/relay/pull/1537 & deployed together.

#skip-changelog (there's already a line about rate limiting metrics buckets)


olksdr (approved):
lgtm

PR #1542, opened 2022-10-24
meta(vscode): Enable all features in Rust-Analyzer

jan-auer:
Configures rust-analyzer to enable all features by default. During regular
deployment, we particularly require the `processing` feature to be enabled.
There are CI jobs that lint and test without all features.

#skip-changelog


olksdr (approved):
:shipit:

PR #1543, opened 2022-10-24
feat(profiling): Add measurements

viglia:
This will let us ingest new measurements (such as dropped frames) along with the profile.

#skip-changelog

### DIFF
--- a/relay-profiling/src/android.rs
+++ b/relay-profiling/src/android.rs
@@ -7,6 +7,7 @@ use serde::{Deserialize, Serialize};
 
 use relay_general::protocol::EventId;
 
+use crate::measurements::Measurement;
 use crate::transaction_metadata::TransactionMetadata;
 use crate::utils::{deserialize_number_from_string, is_zero};
 use crate::ProfileError;
@@ -58,6 +59,9 @@ struct AndroidProfile {
 
     #[serde(default = "AndroidProfile::default")]
     profile: AndroidTraceLog,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    measurements: Option<HashMap<String, Measurement>>,
 }
 
 impl AndroidProfile {
--- a/relay-profiling/src/android.rs
+++ b/relay-profiling/src/android.rs
@@ -61,7 +61,7 @@ struct AndroidProfile {
     profile: AndroidTraceLog,
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    measurements: Option<HashMap<String, Measurement>>,
+    measurements: Option<HashMap<String, Vec<Measurement>>>,
 }
 
 impl AndroidProfile {
--- a/relay-profiling/src/android.rs
+++ b/relay-profiling/src/android.rs
@@ -61,7 +61,7 @@ struct AndroidProfile {
     profile: AndroidTraceLog,
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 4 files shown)


relay-profiling/src/measurements.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Measurement {
+    // nanoseconds elapsed since the start of the profile (wall clock)
+    elapsed_since_start_ns: i64,
```
Zylphrex: I think this should align with the type on the frame struct which is a `u64` and deserializes from string

viglia: good point on using `u64`, but why deserializing from string if we're expecting a number?

relay-profiling/src/measurements.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Measurement {
+    // nanoseconds elapsed since the start of the profile (wall clock)
+    elapsed_since_start_ns: u64,
+    unit: String,
+    value: f64,
+}
```
iker-barriocanal: How is this planned to interact with built-in measurements? See
https://github.com/getsentry/relay/blob/d77c489292123e53831e10281bd310c6a85c63cc/relay-general/src/protocol/measurements.rs#L55

viglia: I don't think it's necessarily planned to interact with the built-in one.

It's more internal to profile. 

Also, for us it's not a map `[String --> Measurement]` but rather a map `[String --> Vec<Measurement>]` (measurements tracked over time).

We also have a specific field  like `elapsed_since_start_ns`

relay-profiling/src/measurements.rs
phacops: JSON doesn't have an unsigned 64 bits type, so typically the correct way to pass such a number is as a string. Now, we might not see all SDKs do this so we'd need to support both inputs.

relay-profiling/src/android.rs
```diff
@@ -58,6 +59,9 @@ struct AndroidProfile {
 
     #[serde(default = "AndroidProfile::default")]
     profile: AndroidTraceLog,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    measurements: Option<HashMap<String, Vec<Measurement>>>,
```
phacops: We don't need to make this an `Option`, we can just `default` and call `HashMap::is_empty` if we don't want to serialize it.

phacops (approved):

relay-profiling/src/measurements.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Measurement {
+    // nanoseconds elapsed since the start of the profile (wall clock)
+    elapsed_since_start_ns: u64,
+    unit: String,
```
Zylphrex: having the unit on every measurement seems like overkill since each key in the `measurements` should have exactly 1 unit unless you're imagining a scenario where the a SDK can send a measurement like `foo` and uses different units at different times.

Zylphrex (approved):
LGTM

PR #1544, opened 2022-10-24
feat(dyn): Add support for decaying rules [INGEST-1670]

iker-barriocanal:
Decaying rules are a new set of rules to be applied in dynamic sampling. They are temporary rules (i.e. rules with a start and end time) to alter the sampling for some time. A simple use case is increasing the sampling rate of a project for some time after a release.

Implementation-wise, they are basically normal rules. `timeRange` is the new and optional parameter on every rule, and its existence determines whether a rule is a decaying rule or not. `start` and `end` subparameters are also optional, and they determine when the range starts (inclusive) and ends (exclusive). When a limiter is not provided relay doesn't restrict that end; for example, if `start` is provided but `end` isn't, relay will apply the rule indefinitely after `start` time. See the following rule example, taken from a unit test in this PR:

```json
{
    "condition":{
        "op":"and",
        "inner": [
                { "op" : "glob", "name": "releases", "value":["1.1.1", "1.1.2"]}
        ]
    },
    "sampleRate": 0.7,
    "type": "trace",
    "id": 1,
    "timeRange": {
        "start": "2022-10-10T10:10:10.101010Z",
        "end": "2022-10-20T20:20:20.202020Z"
    }
}
```

For an easy understanding of how the logic is intended to work, see the unit test in https://github.com/getsentry/relay/pull/1544/commits/d2d5f6a5e7c00eaca2546ca25f27ffd6873e77f8.


relay-sampling/Cargo.toml
```diff
@@ -19,6 +19,7 @@ relay-filter = { path = "../relay-filter" }
 rand = "0.6.5"
 rand_pcg = "0.1.2"
 unicase = "2.6.0"
+chrono = "0.4.11"
```
iker-barriocanal, line 22: This is the version used in other packages (`relay-auth`, `relay-cabi`, `relay-common`...).

relay-sampling/src/lib.rs
```diff
@@ -324,12 +341,59 @@ pub struct SamplingRule {
     #[serde(rename = "type")]
     pub ty: RuleType,
     pub id: RuleId,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub time_range: Option<TimeRange>,
```
iker-barriocanal: Is it worth skipping serialization in cases the rule has expired?

relay-sampling/src/lib.rs
```diff
@@ -1738,6 +1814,40 @@ mod tests {
...
+            }
+        }"#;
+        let rule: Result<SamplingRule, _> = serde_json::from_str(serialized_rule);
+        let rule = rule.unwrap();
+        let time_range = rule.time_range.unwrap();
+        assert_eq!(
+            time_range.start,
+            DateTime::parse_from_rfc3339("2022-10-10T10:10:10.101010Z")
+                .unwrap()
+                .with_timezone(&Utc)
```
iker-barriocanal: Feel free to make suggestions to make this piece a little less ugly.

relay-sampling/src/lib.rs
```diff
@@ -2019,6 +2139,116 @@ mod tests {
...
 
+    #[test]
+    fn test_decaying_rule_precedence() {
+        let rules = SamplingConfig {
+            rules: vec![
+                // Expired
+                SamplingRule {
+                    condition: eq("trace.release", &["1.1.1"], true),
+                    sample_rate: 0.2,
+                    ty: RuleType::Trace,
```
iker-barriocanal: This unit test only verifies rules match the context. Do we want to add another test to ensure it works for transactions? I can either duplicate the test + some minor changes, or create a macro for the tests and do some parametrization.

relay-sampling/src/lib.rs
```diff
@@ -324,12 +341,64 @@ pub struct SamplingRule {
...
+
+    /// Returns whether the sampling rule is a decaying rule.
+    ///
+    /// A decaying rule is a sampling rule that only applies in a closed time range.
+    fn is_decaying_rule(&self) -> bool {
+        match self.time_range {
+            None => false,
+            Some(_) => true,
+        }
+    }
```
```suggestion
    fn is_decaying_rule(&self) -> bool {
        self.time_range.is_some() 
    }
```

relay-sampling/src/lib.rs
```diff
@@ -315,6 +316,22 @@ impl Display for RuleId {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TimeRange {
```
jjbayer, line 327: Let's add a doc comment for this type. Specifically, is this range right-inclusive `[start, end]` or right-exclusive `[start, end)`? Looking at `has_finished`, it's `[start, end]`.

relay-sampling/src/lib.rs
```diff
@@ -324,12 +341,64 @@ pub struct SamplingRule {
     #[serde(rename = "type")]
     pub ty: RuleType,
     pub id: RuleId,
+    /// The time range the rule should be applicable in.
+    ///
+    /// If no time range is provided, relay will always apply the rule. If a range is
+    /// provided the rule becomes a decayng rule, and relay will only apply it inside
```
```suggestion
    /// provided the rule becomes a decaying rule, and relay will only apply it inside
```

relay-sampling/src/lib.rs
jjbayer: It's a good idea, I would not have thought of that! But also, not strictly necessary.

relay-sampling/src/lib.rs
```diff
@@ -324,12 +341,64 @@ pub struct SamplingRule {
...
+    /// neither idle nor expired.
+    fn is_active(&self) -> bool {
+        if !self.is_decaying_rule() {
+            return true;
+        }
+        if !self.is_idle() && !self.is_expired() {
+            return true;
+        }
+        false
+    }
```
jjbayer: I feel like we could simplify these helper methods. What if we just give `TimeRange` an `is_active` function (instead of `has_started` and `has_finished`), and give `SamplingRule` only one function that does

```rust
fn is_active() -> bool {
    match self.time_range { None => true, Some(range) => range.is_active() }
}
```

relay-sampling/src/lib.rs
```diff
@@ -324,12 +341,64 @@ pub struct SamplingRule {
...
+    /// Returns whether the sampling rule is an idle decaying rule.
+    ///
+    /// An idle decaying rule is a decaying rule that is not operative yet, but it will
+    /// in the future.
+    fn is_idle(&self) -> bool {
+        match self.time_range.as_ref() {
+            None => false,
+            Some(time_range) => !time_range.has_started(),
+        }
+    }
```
olksdr: It seems a little bit too verbose. Maybe something like this, in functional style, would be better for readability? 
```suggestion
    fn is_idle(&self) -> bool {
        !self
            .time_range
            .as_ref()
            .map_or(true, TimeRange::has_started)
    }
```

relay-sampling/src/lib.rs
```diff
@@ -324,12 +341,64 @@ pub struct SamplingRule {
...
+
+    /// Returns whether the sampling rule is an expired decaying rule.
+    ///
+    /// An expired decaying rule is a decaying rule that is no longer operative.
+    fn is_expired(&self) -> bool {
+        match self.time_range.as_ref() {
+            None => false,
+            Some(time_range) => time_range.has_finished(),
+        }
+    }
```
```suggestion
    fn is_expired(&self) -> bool {
         self.time_range
             .as_ref()
             .map_or(false, TimeRange::has_finished)
    }
```

relay-sampling/src/lib.rs
jjbayer: Would be nice to test both, how about a parametrized helper function instead of a macro?

relay-sampling/src/lib.rs
```diff
@@ -324,12 +341,64 @@ pub struct SamplingRule {
     #[serde(rename = "type")]
     pub ty: RuleType,
     pub id: RuleId,
+    /// The time range the rule should be applicable in.
+    ///
+    /// If no time range is provided, relay will always apply the rule. If a range is
+    /// provided the rule becomes a decayng rule, and relay will only apply it inside
+    /// the time ranges.
```
```suggestion
    /// the time range.
```

jjbayer (commented):
This looks good, just some minor comments.

relay-sampling/src/lib.rs
```diff
@@ -373,32 +377,11 @@ impl SamplingRule {
         if !self.is_decaying_rule() {
             return true;
         }
-        if !self.is_idle() && !self.is_expired() {
-            return true;
+        if let Some(time_range) = self.time_range.as_ref() {
```
jjbayer: This check is still redundant with `self.is_decaying_rule()`. Is `is_decaying_rule` used anywhere else? If not, I would remove it and just do a `match` statement here.

jjbayer (approved):

relay-sampling/src/lib.rs
iker-barriocanal: I won't be pursuing this for now. I don't want to put the time into correctly setting the serialization skipping function, plus testing it. Ideally, there won't be many expired rules; and if the amount of non-active rules becomes an issue, we probably need to take more action than just skipping the serialization here.

Unless someone disagrees with this and considers it a blocker to landing this PR, I'll keep this as a to-do for some time in the future.

relay-sampling/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 )]
 #![allow(clippy::derive_partial_eq_without_eq)]
 
+use chrono::{DateTime, Utc};
```
jan-auer: This import should be ordered into the next group above `rand`.

relay-sampling/src/lib.rs
```diff
@@ -315,6 +316,29 @@ impl Display for RuleId {
...
 
+/// A range of time.
+///
+/// The time range should be applicable between the start and end times, both inclusive
+/// (i.e. [start, end]). There aren't any explicit checks to ensure the end time is equal
+/// to or greater than the start time; the time range isn't valid in such cases.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TimeRange {
+    start: DateTime<Utc>,
+    end: DateTime<Utc>,
```
jan-auer: I wonder if we should explicitly support half-open ranges right away, for example:
 - a rule that has no defined start but ends at a certain point.
 - a rule that starts at a point and is valid indefinitely.

As such, we could make the time range non-optional, and build it with two options internally.

relay-sampling/src/lib.rs
jan-auer: I'm +1 on skipping serialization eventually, since that is also Sentry's behavior and will make this more compatible with old Relays.

iker-barriocanal: Done in 98e15ae0137d22404454f5ca0b60ea92fc1b1cf1. That commit also supports open ranges, so serialization is only skipped when both `start` and `end` are empty.

relay-sampling/src/lib.rs
iker-barriocanal: Done in 98e15ae0137d22404454f5ca0b60ea92fc1b1cf1.

relay-sampling/src/lib.rs
```diff
@@ -315,6 +317,36 @@ impl Display for RuleId {
...
+/// time is equal to or greater than the start time; the time range isn't valid
+/// in such cases.
+#[derive(Debug, Clone, Copy, Default, Serialize, Deserialize)]
+pub struct TimeRange {
+    start: Option<DateTime<Utc>>,
+    end: Option<DateTime<Utc>>,
+}
+
+impl TimeRange {
+    pub fn is_empty(&self) -> bool {
```
jan-auer, line 334: _nit_: Add a doc comment for this method.

jan-auer (approved):

PR #1545, opened 2022-10-24
feat(replays): Add PII scrubber for replay recordings

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/141


relay-server/src/actors/processor.rs
```diff
@@ -1079,12 +1079,22 @@ impl EnvelopeProcessorService {
...
 
+        // TODO: single function combines pii configs into one.
+        let pii_config1 = state.project_state.config().pii_config.as_ref();
+        let pii_config2 = state
+            .project_state
+            .config()
+            .datascrubbing_settings
+            .pii_config()
+            .map_err(|e| ProcessingError::PiiConfigError(e.clone()))?
+            .as_ref();
```
jjbayer: Do we even want replay scrubbing to use the same config as data scrubbing for events? Is there a use case for making this user-configurable? If not, we might as well use a simple hard coded config here.

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,656 @@
...
+        }
+    }
+
+    fn strip_pii(&mut self, value: &mut String) -> String {
+        match self.pii_processor.process_string(
+            value,
+            &mut Meta::default(),
+            ProcessingState::root(),
+        ) {
+            Ok(_) => value.to_string(),
```
jjbayer: I think this unnecessarily clones a String. Could we change this signature of this function to either accept a `&mut String` and return nothing, or accept a `String` and return the modified string?

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,656 @@
...
+            CustomEventDataVariant::PerformanceSpan(_) => {}
+        }
+    }
+
+    fn recurse_element(&mut self, element: &mut ElementNode) {
+        match element.tag_name.as_str() {
+            "script" | "style" => {}
+            "img" | "source" => {
+                let attrs = &mut element.attributes;
+                attrs.insert("src".to_string(), "#".to_string());
```
jjbayer, line 182: This overwrites `src` if it's already set. Is that intentional?

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,656 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+struct PerformanceSpan {
+    tag: String,
+    payload: PerformanceSpanPayload,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct PerformanceSpanPayload {
+    op: String,
+    description: String, // TODO: needs to be pii stripped (uri params)
```
jjbayer: Is this TODO still relevant?

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,656 @@
...
+    #[serde(rename = "startTimestamp")]
+    start_timestamp: f64,
+    #[serde(rename = "endTimestamp")]
+    end_timestamp: f64,
```
jjbayer: You can use [`serde(rename_all="camelCase")`](https://serde.rs/container-attrs.html#rename_all) on all structs and enums, especially when there's multiple fields to rename.

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,656 @@
...
+
+#[derive(Debug, Deserialize, Serialize)]
+struct Node {
+    #[serde(rename = "rootId", skip_serializing_if = "Option::is_none")]
+    root_id: Option<u32>,
+    #[serde(rename = "isShadowHost", skip_serializing_if = "Option::is_none")]
+    is_shadow_host: Option<bool>,
+    #[serde(rename = "isShadow", skip_serializing_if = "Option::is_none")]
+    is_shadow: Option<bool>,
+    #[serde(rename = "compatMode", skip_serializing_if = "Option::is_none")]
```
jjbayer: See note about `rename_all`.

relay-server/src/actors/processor.rs
cmanallen: Advice on potential configuration is needed here.

relay-replays/src/recording.rs
cmanallen: Yes.

relay-server/src/actors/processor.rs
jjbayer: This would give you some sensible defaults

```rust
DataScrubbingConfig {
    scrub_data: true,
    scrub_defaults: true,
    ..Default::default()
}.pii_config()
```

equivalent to

image

cmanallen: I didn't use `..Default::default()`.  I was getting an error that `pii_config` is private.  So I used:

```rust
let mut scrub_config = DataScrubbingConfig::default();
scrub_config.scrub_data = true;
scrub_config.scrub_defaults = true;
```

Is this equivalent?

relay-replays/src/recording.rs
cmanallen: It is but I think it needs more thought before we commit to scrubbing uri params.

cmanallen: Removed the todo in favor of an issue.

relay-server/src/actors/processor.rs
jjbayer: My bad, didn't think of that! Yes, it's equivalent.

cmanallen:
@jjbayer I want to enable this change for the Sentry org only.  Do I have access to the org id inside relay-server/actors/processor.rs?

jjbayer:
> @jjbayer I want to enable this change for the Sentry org only. Do I have access to the org id inside relay-server/actors/processor.rs?

You have access to the project state (via `ProcessEnvelopeState`), which contains the organization ID but also the project config sent down from sentry. So what you could do is introduce a feature flag in sentry, add it to [`EXPOSABLE_FEATURES` here](https://github.com/getsentry/sentry/blob/de4fda7603dc83ad51f9b2be3b5d28db72257788/src/sentry/relay/config/__init__.py#L44) and check [`has_feature`](https://github.com/getsentry/relay/blob/c5ff3c349f3ade03f03e582ab3061e9e205d9610/relay-server/src/actors/project.rs#L473) from your code.

https://github.com/getsentry/relay/blob/c5ff3c349f3ade03f03e582ab3061e9e205d9610/relay-server/src/actors/processor.rs#L240

cmanallen:
@jjbayer I think hard coding is fine.  This is a temporary measure to limit impact to alpha users.  We're just trying to evaluate the impact of this change with no plans to gate the roll out to customers.  If you have strong opinions on the feature flag approach let me know and I can make the change.

relay-server/src/actors/processor.rs
```diff
@@ -1079,7 +1079,7 @@ impl EnvelopeProcessorService {
     }
 
     /// Remove replays if the feature flag is not enabled
-    fn process_replays(&self, state: &mut ProcessEnvelopeState) {
+    fn process_replays(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
```
jjbayer: I think we can revert the signature change here now, right?

```suggestion
    fn process_replays(&self, state: &mut ProcessEnvelopeState) {
```

relay-server/src/actors/processor.rs
```diff
@@ -1109,9 +1109,30 @@ impl EnvelopeProcessorService {
                     }
                 }
             }
-            ItemType::ReplayRecording => replays_enabled,
+            ItemType::ReplayRecording => {
+                // XXX: Temporarily, only the Sentry org will be allowed to parse replays while
+                // we measure the impact of this change.
+                if replays_enabled && state.project_state.organization_id == Some(1) {
```
jjbayer, line 1115: I think this is OK as long as it's temporary. Note that this also applies to internal orgs on self-hosted (not sure about single-tenant). In the mid to long term, we should either remove the condition or feature-gate it as outlined in the PR comments.

relay-replays/tests/fixtures/rrweb-pii.json
```diff
@@ -0,0 +1,33 @@
...
+                            }
+                        ],
+                        "id": 283
+                    }
+                }
+            ]
+        },
+        "timestamp": 1665063926945
+    }
+]
```
jjbayer, line 33: Is this fixture supposed to contain PII? I noticed there is no test that actually loads this file.

relay-server/src/actors/processor.rs
cmanallen: 👍

relay-replays/tests/fixtures/rrweb-pii.json
cmanallen: Good spot.  I added coverage but it doesn't seem to be stripping the PII.  This condition never returns true: https://github.com/getsentry/relay/blob/master/relay-general/src/pii/processor.rs#L42

Any idea what needs to be done?

jjbayer: Seems that the following condition was the problem, i.e. the matcher never matched because we're plugging in a root element:

https://github.com/getsentry/relay/blob/68526d4fe86e5597250124d64b1fef1174077948/relay-general/src/processor/attrs.rs#L598

I checked out this branch and added 

```rust
let processing_state = processing_state.enter_borrowed(
      "",
      processing_state.inner_attrs(),
      Some(ValueType::String),
);
```

This makes the unit test pass, but it feels quite hacky, maybe @jan-auer has a better idea? At this point it might actually be easier to use `Annotated`.

cmanallen: @jan-auer ping

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,662 @@
...
+            self.recurse_snapshot_node(node)?
+        }
+
+        Ok(())
+    }
+
+    fn strip_pii(&mut self, value: &mut String) -> Result<(), ProcessingAction> {
+        let field_attrs = Cow::Owned(FieldAttrs::new().pii(Pii::True));
+        let processing_state =
+            ProcessingState::new_root(Some(field_attrs), Some(ValueType::String));
```
jjbayer:
```suggestion
        let processing_state =
            ProcessingState::root().enter_static("", Some(field_attrs), Some(ValueType::String));
```

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,662 @@
...
+        recording::strip_pii(&mut events).unwrap();
+
+        let aa = events.pop().unwrap();
+        if let recording::Event::T3(bb) = aa {
+            if let recording::IncrementalSourceDataVariant::Mutation(mut cc) = bb.data {
+                let dd = cc.adds.pop().unwrap();
+                if let recording::NodeVariant::T2(mut ee) = dd.node.variant {
+                    let ff = ee.child_nodes.pop().unwrap();
+                    if let recording::NodeVariant::Rest(gg) = ff.variant {
+                        assert!(gg.text_content != "4917-4845-8989-7107".to_string());
```
jjbayer:
```suggestion
                        assert!(gg.text_content.as_str() == "[Filtered]");
```

jjbayer (commented):
@cmanallen I spoke briefly with Jan, I think the best way to unblock you now is to apply the patch I posted in the comments. In the long run, if PII processing for recordings becomes more elaborate, we might consider defining the entire schema as `Annotated`.

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,662 @@
...
+    let result = encoder.finish()?;
+    Ok(result)
+}
+
+fn strip_pii(events: &mut Vec<Event>) -> Result<(), ProcessingAction> {
+    let mut scrub_config = DataScrubbingConfig::default();
+    scrub_config.scrub_data = true;
+    scrub_config.scrub_defaults = true;
+
+    let pii_config = scrub_config.pii_config().unwrap().as_ref().unwrap().clone();
```
jjbayer:
```suggestion
    let pii_config = scrub_config.pii_config().unwrap().as_ref().unwrap();
```

relay-replays/src/recording.rs
```diff
@@ -0,0 +1,662 @@
...
+    // Processing.
+    strip_pii(&mut events).map_err(RecordingParseError::ProcessingAction)?;
+
+    // Serialization.
+    let out_bytes = dumps(events)?;
+    Ok([header, vec![b'\n'], out_bytes].concat())
+}
+
+fn loads(zipped_input: Vec<u8>) -> Result<Vec<Event>, RecordingParseError> {
+    let mut decoder = ZlibDecoder::new(zipped_input.as_slice());
```
jjbayer:
```suggestion
fn loads(zipped_input: &[u8]) -> Result<Vec<Event>, RecordingParseError> {
    let mut decoder = ZlibDecoder::new(zipped_input);
```

I think we can make this work without allocating a vector for `body`. `bytes.split(|b| b == &b'\n')` should give you an iterator of `&[u8]`, which you could pass directly to this function.

`header` would then also be a `&[u8]` so you'd have to pass `header.into()` into `concat()`.

jjbayer (approved):

PR #1438, opened 2022-08-26
feat: Add crossbuild for AARCH64 on X86_64 for relay-cabi lib

olksdr:
These changes basically add the possibility to cross compile `relay-cabi` library with proper environment, correct version of libc (the one which is in `manylinux2014_aarch64` - **2.17**) and other required dependencies.

The main challenge is to assemble the correct env, old enough to have very old libc version and the required libs. That's why as the base for building this env `centos:7` image was chosen. The newer versions of the images (e.g. recent `ubuntu`) do not have such challenges, or we could just use [`cross`](https://github.com/cross-rs/cross) which is included in [`action-rs/cargo`](https://github.com/actions-rs/cargo#cross-compilation) which we already use.

 
-----

#### New release build

image

#### Old release build
image

As you can see from 1h 20 min build we are down to 12 min. Quite some time savings. Which still can be slimmed down if we pre-build the containers and re-use them for future builds.  

#skip-changelog

PR #1547, opened 2022-10-25
build(craft): Fix manylinux artifact name

jjbayer:
In the process of https://github.com/getsentry/relay/pull/1438, we bumped everything to manylinux2014, but the craft file was not updated.

#skip-changelog


olksdr (approved):
:shipit:

PR #1555, opened 2022-10-27
fix(quotas): Correct reporting of rate limits for transactions

jan-auer:
With #1537 Relay started to under-report outcomes for rate limited transactions.
The root cause is that `Enforcement` was not updated to emit outcomes for
metrics.

This reveals that enforcements for metrics should be recorded only if metrics
have not been extracted from a transaction item yet. In this case, metrics can
still be considered part of the transaction item and as such an outcome is
needed. In contrast, once metrics have been extracted, responsibility for
recording outcomes is moved to the recipient of metrics.

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -1760,10 +1760,6 @@ impl EnvelopeProcessorService {
                 .map_err(ProcessingError::QuotasFailed)?
         });
 
-        if enforcement.event_metrics_active() {
-            state.extracted_metrics.clear();
-        }
-
         if limits.is_limited() {
             ProjectCache::from_registry()
                 .do_send(UpdateRateLimits::new(scoping.project_key, limits));
--- a/relay-server/src/utils/rate_limits.rs
+++ b/relay-server/src/utils/rate_limits.rs
@@ -266,18 +266,20 @@ impl Enforcement {
         self.event.is_active()
     }
 
-    /// Returns `true` if metrics extracted from the event should be rate limited.
-    #[cfg(feature = "processing")]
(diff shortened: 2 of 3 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -1760,10 +1760,6 @@ impl EnvelopeProcessorService {
                 .map_err(ProcessingError::QuotasFailed)?
         });
 
-        if enforcement.event_metrics_active() {
-            state.extracted_metrics.clear();
```
jan-auer, line 1764: This was introduced in #1537, but it is now dead code. Metrics will always be extracted at this point, and as such `event_metrics_active` can never be true. 

This is not a problem, since the extracted metrics will be sent to the project cache and be dropped there. Over there, we have proper outcome logging for metrics implemented. It's an unnecessary indirection, but is the smallest change to get this consistent and working.

iker-barriocanal (approved):
🚀

relay-server/src/utils/rate_limits.rs
```diff
@@ -970,7 +977,7 @@ mod tests {
         let (enforcement, limits) = limiter.enforce(&mut envelope, &scoping()).unwrap();
 
         assert!(limits.is_limited());
-        assert!(enforcement.event_metrics.is_active());
+        assert!(!enforcement.event_metrics.is_active());
```
jan-auer, line 980: This change is intentional. We don't want an enforcement -- and not record outcomes -- if metrics had been extracted already. The outcome would be logged somewhere else.

olksdr (approved):
:shipit:

PR #1556, opened 2022-10-28
ref(event): Add correct limit and validation to distribution name [INGEST-1615]

olksdr:
Even though we had the `max_chars` limit set to 200 on the `dist` we never validated that length while trying to normalize the event.

This change adds the correct limit (64 chars) which is set on the Sentry site (field in the DB is varchar(64)) and also adds the validation with proper error generation if the value is longer than the length limit.

Also related: https://github.com/getsentry/sentry/pull/40640


relay-general/src/store/normalize.rs
```diff
@@ -90,20 +90,25 @@ pub fn is_valid_platform(platform: &str) -> bool {
...
+            *dist = None;
+        } else {
+            let trimmed = val.trim();
+            // Validate the length of the distribution name and make sure to attach the error in
+            // case of the issues.
+            if bytecount::num_chars(trimmed.as_bytes()) > MaxChars::Distribution.limit() {
+                *distribution = Annotated::from_error(
+                    Error::new(ErrorKind::ValueTooLong),
+                    Some(Value::String(val.to_owned())),
+                )
```
iker-barriocanal: Instead of creating a new `Annotated` object, does it make sense to just add the error to the meta? Something like the following:
```suggestion
                distribution.meta_mut().add_error(...);
```
In this case, you'd still have the initial string there which is too long, so I'm not sure if this suggestion makes sense.

relay-general/src/store/normalize.rs
```diff
@@ -90,20 +90,25 @@ pub fn is_valid_platform(platform: &str) -> bool {
...
+            } else if trimmed != val {
+                *val = trimmed.to_string()
```
iker-barriocanal: I'm curious, why do you assign the new value this way using references, instead of something like `distribution.set_value(...)`? Same question for `*dist = None` above.

relay-general/src/store/normalize.rs
```diff
@@ -90,20 +90,25 @@ pub fn is_valid_platform(platform: &str) -> bool {
     VALID_PLATFORMS.contains(&platform)
 }
 
-pub fn normalize_dist(dist: &mut Option<String>) {
-    let mut erase = false;
+pub fn normalize_dist(distribution: &mut Annotated<String>) {
```
iker-barriocanal: I like this change.

relay-general/src/store/normalize.rs
```diff
@@ -1964,30 +1995,30 @@ mod tests {
...
+        assert_eq!(dist.value(), Some(&"foo".to_string()));
     }
 
     #[test]
     fn test_normalize_dist_whitespace() {
-        let mut dist = Some(" ".to_owned());
+        let mut dist = Annotated::new(" ".to_owned());
         normalize_dist(&mut dist);
-        assert_eq!(dist.unwrap(), ""); // Not sure if this is what we want
+        assert_eq!(dist.value(), Some(&"".to_string())); // Not sure if this is what we want
```
iker-barriocanal: Normalization should be idempotent. If we make another normalization call, `dist` is then `None`. This situation must not happen -- we're running multiple layers of relays and debugging can be complicated.

iker-barriocanal: I'd also take the chance and make a decision with regard to that comment. If the resulting string after trimming is empty, `dist` should be `None`.

iker-barriocanal (commented):
This PR is actually a bugfix, please add the entry to the changelog.

relay-general/src/store/normalize.rs
olksdr: I've rewritten this using `apply` method

relay-general/src/store/normalize.rs
olksdr: agreed and fixed this

relay-general/src/store/normalize.rs
```diff
@@ -90,20 +90,19 @@ pub fn is_valid_platform(platform: &str) -> bool {
...
-        if trimmed != val {
-            *val = trimmed.to_string()
+pub fn normalize_dist(distribution: &mut Annotated<String>) -> ProcessingResult {
+    distribution.apply(|dist, meta| {
+        let trimmed = dist.trim();
+        if trimmed.is_empty() {
+            return Err(ProcessingAction::DeleteValueHard);
+        } else if bytecount::num_chars(trimmed.as_bytes()) > MaxChars::Distribution.limit() {
+            meta.add_error(Error::new(ErrorKind::ValueTooLong));
+            return Err(ProcessingAction::DeleteValueSoft);
```
olksdr, line 100: This is the same behaviour as in case of `release` field.

jjbayer (commented):
It seems that in sentry we trim tags instead of deleting them, so I wonder which way we should go here: https://github.com/getsentry/sentry/blob/3a7c72d622b1fe64fa26ef568e20493be7bbee9e/src/sentry/event_manager.py#L178

jjbayer:
> It seems that in sentry we trim tags instead of deleting them, so I wonder which way we should go here: https://github.com/getsentry/sentry/blob/3a7c72d622b1fe64fa26ef568e20493be7bbee9e/src/sentry/event_manager.py#L178

Especially because https://github.com/getsentry/sentry/pull/40640 truncates instead of deleting, so I feel like we shouldn't delete it in Relay?

olksdr:
> But from the users' point of few - is truncated information better than no information on distribution at all ? 

This is a good point. Deleting it while annotating the error gives the user explicit information about what went wrong, which is probably better than silently truncating. Another option would be to truncate _and_ annotate, like the `TrimmingProcessor` does. But I'm fine with the change as-is.

jjbayer (approved):

PR #1563, opened 2022-11-03
ref(kafka): Move code into separate crate [INGEST-1627]

olksdr:
Create `relay-kafka` crate with its own configuration and producer functionality.

Having dedicated crate will help to keep all the related code together and it also simplifies the dependency tree.


relay-server/src/actors/store.rs
```diff
@@ -224,35 +83,69 @@ impl Producers {
...
         let mut reused_producers = BTreeMap::new();
         let producers = Producers {
-            attachments: make_producer(&**config, &mut reused_producers, KafkaTopic::Attachments)?,
-            events: make_producer(&**config, &mut reused_producers, KafkaTopic::Events)?,
-            transactions: make_producer(
-                &**config,
+            attachments: Producer::create(
+                &config
+                    .kafka_config(KafkaTopic::Attachments)
+                    .map_err(|_| ServerErrorKind::KafkaError)?,
```
jjbayer: I'm not sure how `make_producer`s `create_with_context` worked, but are we now erasing information from the original error that used to be preserved? Or was it always erased and mapped to a `ServerErrorKind::KafkaError`?

olksdr: The `create_with_context` call did not go anywhere, it  was moved into the kafka crate and still used but in `Producer::create` function - which captures and reports the errors. 

But we do erase the error here and replace it with `ServerErrorKind::KafkaError` which means that something wrong with kafka configuration and the producer could not be initialized. 

Do you think we wanna keep the error around here as well?  I can add required changes

jjbayer: This is a nit by the way, but yes, I would give `KafkaError` a `#[cause] ProducerError` to propagate the information. Would that work?

relay-server/src/actors/store.rs
```diff
@@ -224,27 +83,69 @@ impl Producers {
...
+                &mut reused_producers,
+            )
+            .map_err(|_| ServerErrorKind::KafkaError)?,
+            replay_events: Producer::create(
+                &config
+                    .kafka_config(KafkaTopic::ReplayEvents)
+                    .map_err(|_| ServerErrorKind::KafkaError)?,
+                &mut reused_producers,
+            )
+            .map_err(|_| ServerErrorKind::KafkaError)?,
```
jjbayer: Another nit: These expressions all follow the same pattern, should we replace them with a helper function that just takes a `KafkaTopic` as argument?

jjbayer (approved):

relay-kafka/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+serde = { version = "1.0.114", features = ["derive"] }
+serde_json = { version = "1.0.55", optional = true }
+
+[dev-dependencies]
+serde_yaml = "0.8.13"
+
+
+[features]
+default = []
+processing = [
```
jan-auer: This feature effectively gives us producer implementations, should we name it as such?

relay-kafka/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    InvalidMsgPack(#[cause] rmp_serde::encode::Error),
+    #[fail(display = "failed to serialize json message")]
+    InvalidJson(#[cause] serde_json::Error),
+    #[fail(display = "invalid kafka shard")]
+    InvalidShard,
+}
+
+/// Kafka configuration errors.
+#[derive(Fail, Debug)]
+pub enum ConfigError {
```
jan-auer: Is there a reason this is not in `config`? This causes some strange imports now where most types get imported from `Config` except for the related error.

relay-kafka/src/lib.rs
```diff
@@ -0,0 +1,54 @@
+pub mod config;
```
jan-auer: It would be great if we could use this as a chance to improve documentation. Particularly, the root doc comment of the crate is valuable because it can explain the purpose of the crate and point to the various entry-points.

relay-kafka/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+#[cfg(feature = "processing")]
+mod producer;
+#[cfg(feature = "processing")]
+mod utils;
+
+use failure::Fail;
+
+#[cfg(feature = "processing")]
+pub use producer::*;
```
jan-auer, line 45: Since we already create a config module, we could also make the producer module public and bundle all the types there.

There's not yet a consistent paradigm for how to structure such crates in Relay. I've mostly used this heuristic:
 - If it's something that you almost always import when using a crate/module, export it top-level.
 - If it's something you rarely use or do not need to import directly (e.g. because it's a return type), move it into submodules.
 - If there's several concerns that would get messy in a single namespace, put everything behind modules.

relay-kafka/src/producer.rs
```diff
@@ -0,0 +1,170 @@
+use std::{collections::BTreeMap, sync::Arc};
```
jan-auer: Please un-merge those imports. Same with the `crate` imports below.

relay-kafka/src/producer.rs
```diff
@@ -0,0 +1,170 @@
...
+
+impl Producer {
+    /// Creates the kafka [`Producer`] based on the provided configuration.
+    ///
+    /// # Errors
+    /// Returns [`ProducerError::InvalidConfig`] error if the provided configuration is wrong and
+    /// the producer could not be created.
+    pub fn create<'a>(
+        config: &'a KafkaConfig,
+        reused_producers: &mut ReusedProducersMap<'a>,
```
jan-auer: This was fine while this was still in a private module, but this mixes several concerns for a public signature:
- It creates a new producer instance
- It memoizes them and determines the data structure for memoization

I realize that this increases the effort for this refactor, but it would be good to clean this up and separate those concerns. Ideally the memo-map is either generic, or becomes a higher-level utility that uses `create` internally.

In the end, I think the abstraction here is entirely wrong. The real producer is the `Arc`'ed threaded producer inside. The `Producer` type here then gets one or more of them and wraps them together with a topic name.

relay-kafka/src/producer.rs
```diff
@@ -0,0 +1,170 @@
...
+    config::{KafkaConfig, KafkaParams},
+    utils::{CaptureErrorContext, KafkaHistograms, ThreadedProducer},
+    Message, ProducerError,
+};
+
+/// Temporary map used to deduplicate kafka producers
+type ReusedProducersMap<'a> = BTreeMap<Option<&'a str>, Arc<ThreadedProducer>>;
+
+/// This object containes the Kafka producer variants for single and sharded configurations.
+pub enum Producer {
```
jan-auer: _nit_: The internals of the enum are now part of the public signature. Do we want this explicitly?

- If yes, then let's also introduce a type for the `Single` variant. This allows us to move specific functionality to a method on that type rather than having large functions covering all variants here.
- If not, let's make this a `struct` with a hidden `inner` field that is then a private enum. I'd still propose the refactor from the previous point, though.

relay-kafka/src/lib.rs
```diff
@@ -0,0 +1,54 @@
+pub mod config;
+#[cfg(feature = "processing")]
+mod producer;
+#[cfg(feature = "processing")]
+mod utils;
```
jan-auer: Since this module just contains producer utilities, why don't we merge them into the `producer` module? That would also be quite nice if we rename the feature to `producer`.

relay-kafka/src/utils.rs
```diff
@@ -0,0 +1,62 @@
...
+            metric!(counter(KafkaCounters::ProcessingProduceError) += 1);
+        }
+    }
+}
+
+/// The wrapper type around the kafka [`rdkafka::producer::ThreadedProducer`] with our own
+/// [`CaptureErrorContext`] context.
+pub type ThreadedProducer = rdkafka::producer::ThreadedProducer<CaptureErrorContext>;
+
+enum KafkaCounters {
```
jan-auer: In order to document these metrics, we need to list them in the documentation CI job here:
https://github.com/getsentry/relay/blob/453a5d054c5e06c50e83b7732ea1944f9db31a33/.github/workflows/deploy.yml#L134

_nit_: I would propose to move these into a `statsd` module so the path stays stable.

relay-kafka/src/lib.rs
jan-auer: Please add this to the preamble:

```rust
#![warn(missing_docs)]
#![warn(missing_debug_implementations)]
#![doc(
)]
```

Particularly, note that this will require `Debug` to be implemented on most types. Let's ensure this is the case.

jan-auer (commented):
Please also list the `relay-kafka` crate in the doc comment of the main `relay` crate.

relay-kafka/Cargo.toml
olksdr: Yep, sounds good.

relay-kafka/src/lib.rs
olksdr: 👍  I will create some initial docs here.

relay-kafka/src/lib.rs
olksdr: yeah, I will reorganize the modules and make it simpler and uniform.

relay-kafka/src/producer.rs
olksdr: I will revisit this and make some changes into this refactor.

relay-kafka/src/utils.rs
olksdr: Sounds good 👍

.github/workflows/deploy.yml
```diff
@@ -131,7 +131,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: run
-          args: -p document-metrics -- -o relay_metrics.json relay-server/src/statsd.rs relay-metrics/src/statsd.rs
+          args: -p document-metrics -- -o relay_metrics.json relay-server/src/statsd.rs relay-metrics/src/statsd.rs relay-kafka/src/statsd.rs
```
jjbayer, line 134: This is easy to miss when adding a crate, would a glob pattern work here? That is, `relay-*/src/statsd.rs`

relay-kafka/src/producer/mod.rs
```diff
@@ -0,0 +1,331 @@
...
+            );
+            ClientError::InvalidTopicName
+        })?;
+        producer.send(organization_id, key, variant, payload)
+    }
+}
+
+/// Helper structures responsable for building the actual [`KafkaClient`].
+#[derive(Default)]
+pub struct KafkaClientBuilder {
```
jjbayer, line 172: Nice!

relay-server/src/actors/outcome.rs
```diff
@@ -693,30 +687,19 @@ impl KafkaOutcomesProducer {
...
-        &self.billing
+        let mut client_builder = KafkaClient::builder();
+
+        for topic in &[KafkaTopic::Outcomes, KafkaTopic::OutcomesBilling] {
+            let kafka_config = &config
+                .kafka_config(*topic)
+                .map_err(|_| ServerErrorKind::KafkaError)?;
+            client_builder = client_builder
+                .add_kafka_topic_config(kafka_config)
+                .map_err(|_| ServerErrorKind::KafkaError)?;
```
jjbayer, line 698: This nit still stands, but it's not a blocker: 

> I would give KafkaError a #[cause] ProducerError to propagate the information.

https://github.com/getsentry/relay/pull/1563/files/3050fca26247f6d2e61d77c2b74842a93a721465#r1013800430

jjbayer (approved):

relay-server/src/actors/outcome.rs
olksdr: @jjbayer so, the reason that the error neither here nor in any other case when the `ServerErrorKind` is used is that, the enum contains all the `"empty"` variants and also implement `Copy` and `Clone` and also `Hash` where most of the error do not do that, see:
https://github.com/getsentry/relay/blob/1e2bbf96936adfd170dfdef2c7f4656eaaf30ea2/relay-server/src/service.rs#L41-L75

In case we want to propagate the errors - we would need  either to implement `Hash` trait on kafka `ClientError`  and remove `Clone` and `Copy`,  or if we add only additional formatted message inside the variant - we would have to at least remove `Copy` and `Clone`  and return reference in https://github.com/getsentry/relay/blob/1e2bbf96936adfd170dfdef2c7f4656eaaf30ea2/relay-server/src/service.rs#L96

What do you think would be better?

jjbayer: I see. Ignore my comment then, given that this error only happens on initialization I think we can live without the extra information being bubbled up.

.github/workflows/ci.yml
```diff
@@ -103,7 +103,7 @@ jobs:
           args: --workspace --all-targets --no-deps -- -D warnings
 
   test:
-    timeout-minutes: 15
+    timeout-minutes: 20
```
jan-auer: Is that timeout change still required?

relay-kafka/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "relay-kafka"
+authors = ["Sentry <oss@sentry.io>"]
+description = "Kafka related functionality for Relay"
+homepage = "https://getsentry.github.io/relay/"
+repository = "https://github.com/getsentry/relay"
+version = "22.9.0"
```
jan-auer: The version is outdated.

```suggestion
version = "22.10.0"
```

Please also note that there's a self-hosted release today, so ideally we merge before that.

relay-kafka/src/config.rs
```diff
@@ -0,0 +1,390 @@
...
+    /// ReplayEvents, breadcrumb + session updates for replays
+    ReplayEvents,
+    /// ReplayRecordings, large blobs sent by the replay sdk
+    ReplayRecordings,
+}
+
+impl KafkaTopic {
+    /// Returns iterator over the variants of [`KafkaTopic`].
+    /// It will have to be adjusted if the new variants are added.
+    pub fn iter() -> std::slice::Iter<'static, KafkaTopic> {
```
jan-auer:
```suggestion
    pub fn iter() -> std::slice::Iter<'static, Self> {
```

_nit_: This iterates references of `KafkaTopic`. It would be slightly more ergonomic to iterate owned references, but it's likely not worth the added complexity. There would be [`std::array::IntoIter`](https://doc.rust-lang.org/stable/std/array/struct.IntoIter.html), but that will internally own the array.

relay-kafka/src/config.rs
```diff
@@ -0,0 +1,390 @@
...
+            ReplayRecordings,
+        ];
+        TOPICS.iter()
+    }
+}
+
+/// Configuration for topics.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(default)]
+pub struct TopicAssignments {
```
jan-auer, line 69: To separate concerns, it would be beneficial if the `relay-kafka` crate doesn't know about the topic assignments. For instance, we could leave the `TopicAssignment` type here, but move `TopicAssignments` and `KafkaTopic` back into `relay-config` or wherever it's best defined. That way, we don't have to rebuild it every time we add a new topic. 

I'm happy to merge the PR as-is though, and follow up with this later.

relay-kafka/src/config.rs
```diff
@@ -0,0 +1,390 @@
...
+    /// The list of the Kafka configs with related shard configs.
+    Sharded {
+        /// The maximum number of logical shards for this set of configs.
+        shards: u64,
+        /// The list of the sharded Kafka configs.
+        configs: BTreeMap<u64, KafkaParams<'a>>,
+    },
+}
+
+/// Sharded Kafka config
```
jan-auer:
```suggestion
/// Sharded Kafka config.
```

relay-kafka/src/lib.rs
```diff
@@ -0,0 +1,44 @@
+//! Kafka-related functionality.
+//!
+//! By default, this library only provides the support for `relay` configuration. With `producer`
+//! feature enabled, the [`KafkaClient`] is provided and can be used to send messages to the Kafka
+//! broker.
+//!
+//! # Usage
+//!
+//! ```compile_fail
```
jan-auer, line 9: It's unfortunate that we cannot even compile this because of missing features. ~Maybe we move the documentation of the producer part into the `producer` module so it can compile?~ (that won't work since the module is not public, just its contents)

Regardless, we should at least insert the correct imports here.

jan-auer (approved):
Excellent documentation, thank you for taking the time to write this!

This is good to merge from my end, see minor and optional comments below.

.github/workflows/ci.yml
olksdr: I think it can be removed now. Added before rust cache fix.

relay-kafka/src/config.rs
olksdr: Agreed! I will followup on this

PR #1564, opened 2022-11-03
ci: lower timouts for tests

olksdr:
It reverts increased timeouts done in https://github.com/getsentry/relay/pull/1538 to previous levels.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
           args: --workspace --all-targets --no-deps -- -D warnings
 
   test:
-    timeout-minutes: 20
+    timeout-minutes: 15
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -228,7 +228,7 @@ jobs:
   test_integration:
     name: Integration Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 25
+    timeout-minutes: 15
```
jjbayer: Was this 15 or 20?

olksdr: `15`

jjbayer (approved):

.github/workflows/ci.yml
olksdr: Bumped this back to `20`,  since after running this just and I noticed that if we had a clean build it would take up to 17 min. So let's stay on the safe side and have some room in case we will have to re-build the tree from scratch.

PR #1565, opened 2022-11-03
fix(tools): Add light normalization to process-event

jjbayer:
The tool `process-event` should apply full store normalization according to its docs:

```
        --store      
            Apply full store normalization
```

Full store normalization now also includes light normalization, so add that to the command.

Note: I would be in favor of putting light normalization back into regular store normalization, and control what parts it actually runs through a flag, similar to `is_renormalize`. But that is out of scope for this PR.

#skip-changelog


olksdr (approved):
:shipit:

tools/process-event/src/main.rs
```diff
@@ -83,6 +85,7 @@ impl Cli {
         }
 
         if self.store {
+            light_normalize_event(&mut event, &LightNormalizationConfig::default()).unwrap();
```
olksdr: On the other hand, do you really want to just `unwrap` it here ?  Maybe do same like other errors, add context and use `?` to bubble up the error?

jjbayer: Good call, will map the error as the `process_value` calls do.

PR #1554, opened 2022-10-27
test(integration): Skip flaky test

iker-barriocanal:
Skip the marked flaky test.

1. [This PR](https://github.com/getsentry/relay/pull/1544) passes CI and is merged.
2. CI on master fails [attempt 1](https://github.com/getsentry/relay/actions/runs/3338360468/jobs/5525929842#step:10:784).
3. CI on master passes [attempt 2](https://github.com/getsentry/relay/actions/runs/3338360468/jobs/5526346187).

I have not spent time figuring out how to fix it, or what is really failing. I'm disabling the test until we put in the time and figure it out.

#skip-changelog

PR #1568, opened 2022-11-04
fix(tests): Reduce rate limiting test flakyness [INGEST-1680]

jan-auer:
Rate limiting tests use real time. With small rate limiting windows, chances are
higher that the window changes between two event submissions. To counter this,
we've so far increased the window of the test quotas to one day, which makes it
extremely unlikely.

A proper solution for this problem would be to set explicit timestamps and move
such tests to unit-test level.

Closes https://github.com/getsentry/relay/pull/1554
#skip-changelog


jjbayer (approved):

olksdr (approved):
:shipit:

PR #1569, opened 2022-11-04
fix(normalize): Drop custom measurements with invalid names

jjbayer:
### Legal Boilerplate

### DIFF
--- a/relay-common/src/constants.rs
+++ b/relay-common/src/constants.rs
@@ -665,9 +665,70 @@ impl schemars::JsonSchema for MetricUnit {
     }
 }
 
+/// An error parsing a [`MetricName`] or one of its variants.
+#[derive(Clone, Copy, Debug)]
+pub struct ParseMetricNameError(());
+
+/// A valid metric name. This is the statsd name, i.e. without namespace, type or unit.
+///
+/// Metric names cannot be empty, must begin with a letter and can consist of ASCII alphanumerics,
+/// underscores, slashes and periods.
+#[derive(Debug)]
+pub struct MetricName<'a>(&'a str);
+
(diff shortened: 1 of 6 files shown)


relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -414,9 +415,18 @@ fn extract_transaction_metrics_inner(
                 tags_for_measurement.insert("measurement_rating".to_owned(), rating);
             }
 
+            let metric_name = format!("measurements.{}", name);
```
jjbayer, line 418: This is the line which makes this PR necessary in the first place: The metric name is generated on-the-fly from whatever measurement name the event contains, so it might be invalid.

relay-common/src/constants.rs
```diff
@@ -665,9 +665,70 @@ impl schemars::JsonSchema for MetricUnit {
...
+/// An error parsing a [`MetricName`] or one of its variants.
+#[derive(Clone, Copy, Debug)]
+pub struct ParseMetricNameError(());
+
+/// A valid metric name. This is the statsd name, i.e. without namespace, type or unit.
+///
+/// Metric names cannot be empty, must begin with a letter and can consist of ASCII alphanumerics,
+/// underscores, slashes and periods.
+#[derive(Debug)]
+pub struct MetricName<'a>(&'a str);
```
jjbayer, line 677: Not sure if this belongs in `constants.rs`, but `MetricUnit` is also part of this module. Maybe we can move both types to a new file like `relay-common/src/metrics.rs`.

jjbayer:
Closing in favor of https://github.com/getsentry/relay/pull/1571.

PR #1571, opened 2022-11-07
fix(protocol): Disallow `-` in measurement and breakdown names

jjbayer:
Names of custom measurements are converted into MRIs without any validation. This means that in metrics extraction, we are able to construct MRIs that are less strict than what the metrics protocol demands. Subsequent attempts to parse the MRI string into a `MetricsResourceIdentifier` object will fail.

This popped up in rate limiting, where we have to parse the MRI to know what namespace and name we are dealing with. Because of this bug, custom measurements with invalid names are currently not rate limited at all.

Fixes [POP-RELAY-2N8](https://sentry.my.sentry.io/organizations/sentry/issues/315035).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Features**:
 
 - Support decaying rules. Decaying rules are regular sampling rules, but they are only applicable in a specific time range. ([#1544](https://github.com/getsentry/relay/pull/1544))
+- Disallow `-` in measurement and breakdown names. These items are converted to metrics, which do not allow `-` in their name. ([#1571](https://github.com/getsentry/relay/pull/1571))
 
 **Bug Fixes**:
 
--- a/relay-general/src/protocol/breakdowns.rs
+++ b/relay-general/src/protocol/breakdowns.rs
@@ -16,7 +16,7 @@ impl Breakdowns {
             && name.starts_with(|c| matches!(c, 'a'..='z' | 'A'..='Z'))
             && name
                 .chars()
-                .all(|c| matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_' | '.'))
+                .all(|c| matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '.'))
     }
 }
 
@@ -34,7 +34,7 @@ impl FromValue for Breakdowns {
                         return Some((name.into(), object));
                     } else {
                         processing_errors.push(Error::invalid(format!(
-                            "breakdown name '{}' can contain only characters a-z0-9.-_",
+                            "breakdown name '{}' can contain only characters a-z0-9._",
                             name
                         )));
                     }
--- a/relay-general/src/protocol/measurements.rs
+++ b/relay-general/src/protocol/measurements.rs
@@ -101,7 +101,7 @@ impl FromValue for Measurements {
                         return Some((name.to_lowercase(), object));
                     } else {
                         processing_errors.push(Error::invalid(format!(
-                            "measurement name '{}' can contain only characters a-z0-9.-_",
+                            "measurement name '{}' can contain only characters a-z0-9._",
                             name
                         )));
                     }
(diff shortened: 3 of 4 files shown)


relay-common/src/utils.rs
```diff
@@ -145,6 +145,20 @@ impl<T: Clone> GlobMatcher<T> {
...
+/// underscores and periods.
+pub fn is_valid_metric_name(name: &str) -> bool {
+    let mut iter = name.as_bytes().iter();
+    if let Some(first_byte) = iter.next() {
+        if first_byte.is_ascii_alphabetic() {
+            return iter.all(|b| b.is_ascii_alphanumeric() || matches!(b, b'.' | b'_'));
+        }
+    }
+    false
+}
```
jjbayer: I had to move this function from `relay-metrics` to `relay-common` so it can be used from `relay-general`.

jjbayer: I had to move this function from `relay-metrics` to `relay-common` so it can be used from `relay-general`. Note that this version no longer allows `/`.

relay-metrics/src/protocol.rs
```diff
@@ -516,10 +510,7 @@ impl Metric {
 
         let name_value_str = components.next()?;
         let ty = components.next().and_then(|s| s.parse().ok())?;
-        let (name_and_namespace, unit, value) = parse_name_unit_value(name_value_str, ty)?;
-        let (raw_namespace, name) = name_and_namespace
-            .split_once('/')
-            .unwrap_or(("custom", string));
+        let (raw_namespace, name, unit, value) = parse_name_unit_value(name_value_str, ty)?;
```
jjbayer: I had to refactor these parser functions because `is_valid_metric_name` no longer allows `/` (because it is also called from normalization). Instead of validate-then-split, we know do split-then-validate.

olksdr (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Features**:
 
 - Support decaying rules. Decaying rules are regular sampling rules, but they are only applicable in a specific time range. ([#1544](https://github.com/getsentry/relay/pull/1544))
+- Restrict custom measurement names to alphanumericals, '.' and '_'. Invalid measurements are dropped. ([#1571](https://github.com/getsentry/relay/pull/1571))
```
iker-barriocanal: nit:
```suggestion
- Restrict custom measurement names to alphanumericals, `.` and `_`. Invalid measurements are dropped. ([#1571](https://github.com/getsentry/relay/pull/1571))
```

relay-common/src/utils.rs
iker-barriocanal: Can disallowing the usage of `/` cause issues in any of the places where it's currently used?

relay-general/src/store/normalize.rs
```diff
@@ -2130,6 +2136,44 @@ mod tests {
...
+    #[test]
+    fn test_filter_invalid_measurement_name() {
+        let json = r#"
+        {
+            "type": "transaction",
+            "timestamp": "2021-04-26T08:00:05+0100",
+            "start_timestamp": "2021-04-26T08:00:00+0100",
+            "measurements": {
+                "my_custom-measurement_1": {"value": 123},
+                "my_custom_measurement_2": {"value": 789}
```
iker-barriocanal: nit: considering the parsing fails if the name is not correct and we've modified the accepted domain, I'd add a third measurement with a `.` (valid), and maybe one with a `/` (invalid).

iker-barriocanal (approved):

relay-common/src/utils.rs
jjbayer: The only other usage is [here](https://github.com/getsentry/relay/blob/4dcaf5d8f0e87da67877e050535f9d6504b09550/relay-metrics/src/protocol.rs#L281), and I refactored that part to split by `/` before validating, so all cases should be covered.

relay-general/src/store/normalize.rs
jjbayer: This was great advice! By adding some more cases I realized that there is already protocol validation in `protocol.rs`, which actually adds `errors` for invalid names, so I reverted all changes in this PR and just fixed the validation there.

olksdr (approved):

iker-barriocanal (approved):
🚀

dashed:
Sorry I missed this. But this looks good to me 👍

PR #1573, opened 2022-11-09
fix(server): Track session envelopes as accepted

jan-auer:
Relay has a flag to disable ingestion of raw session payloads. When this flag is enabled, it converts sessions into metrics, passes them to the metrics aggregator, and removes the session item from the envelope.

This process may leave an empty envelope behind. Relay internally logs this as a rejection, since before an empty envelope was always the result of rate limiting. This PR changes this, so that an empty envelope counts as accepted if metrics were extracted for one of its items.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -2056,23 +2056,27 @@ impl EnvelopeProcessorService {
             || {
                 match self.process_state(&mut state) {
                     Ok(()) => {
-                        if !state.extracted_metrics.is_empty() {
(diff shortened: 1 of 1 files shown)


olksdr (approved):

jjbayer (commented):
Looks good to me. Should we add a test?

jan-auer:
I truly hate to say it, but this code isn't well testable right now. I'll try to refactor it and add some tests around it, but would prefer to get this merged sooner.

PR #1575, opened 2022-11-10
ref(store): Remove unit mismatch error [INGEST-1701]

jan-auer:
Relay logged internal diagnostics when SDKs sent well-known transactions with an
unexpected unit. All these differences have been fixed in SDKs, so we can remove
this error now.

#skip-changelog


olksdr (approved):
:shipit:

jjbayer (approved):
❤️

PR #1577, opened 2022-11-11
fix(normalization): Use correct meta object for logentry in light normalization [INGEST-1679]

olksdr:
This PR fixes bug for log entry light normalization, where we used wrong level meta object for errors annotations.


jjbayer (approved):
Looks good to me! On the `invalid-transaction.json` payload, is the diff with the version before light normalization now empty?

olksdr:
> Looks good to me! On the `invalid-transaction.json` payload, is the diff with the version before light normalization now empty?

So now we have again the `errors` populated and `_meta` in the correct place, like it was before. 

image


See screenshot ,  left side it the old before light normalization, and on the right side is the latest revision with this fix.

PR #1578, opened 2022-11-11
feat(protocol): Add app.in_foreground and thread.main flag

markushi:
Extends the protocol for contexts
 * `in_foreground` for the app context event payload
 * `main` for the thread context

This is especially useful for analysing events for mobile platforms, as it provides extra context for e.g. crashes.

Related PRs:
develop: https://github.com/getsentry/develop/pull/728

Related Issues:
https://github.com/getsentry/team-mobile/issues/36
https://github.com/getsentry/rfcs/pull/36


relay-general/src/protocol/contexts/app.rs
```diff
@@ -36,6 +36,9 @@ pub struct AppContext {
     /// Amount of memory used by the application in bytes.
     pub app_memory: Annotated<u64>,
 
+    /// A flag indicating whether the app is in foreground or not. An app is in foreground when it's visible to the user.
+    pub app_in_foreground: Annotated<bool>,
```
jjbayer: nit: I realize it's consistent with the other fields in this context, but the `app_` prefix seems redundant to me.

jjbayer (approved):
This looks good to me. Let's not merge until https://github.com/getsentry/develop/pull/728 is approved and we know the naming is final.

relay-general/src/protocol/contexts/app.rs
markushi: Yeah, I prefer to not have the `app_` prefix as well. It might not be consistent within the app context, but looking at other contexts, it seems to be a better fit without the prefix. Let me remove the prefix 👍

iker-barriocanal: Update the changelog too to reflect this name change, please.

iker-barriocanal (approved):

PR #1579, opened 2022-11-11
ref(profiling): Enforce and standardize profile measurement units

viglia:
This PR, as the title suggests, aims at enforcing the unit type checking.

New units will be added as needed when we'll start to support new kind of measurements.

#skip-changelog

### DIFF
--- a/relay-profiling/src/measurements.rs
+++ b/relay-profiling/src/measurements.rs
@@ -1,11 +1,10 @@
-use serde::{de, Deserialize, Serialize};
+use serde::{Deserialize, Serialize};
 
 use crate::utils::deserialize_number_from_string;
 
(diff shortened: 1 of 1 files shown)


jjbayer (commented):
Instead of writing a customer deserializer, you can use an enum:

```rust
#[serde(rename_all = "lowercase")]
#[derive(Clone, Debug, Serialize, Deserialize)]
pub enum MeasurementUnit {
    #[serde(alias = "ns")]
    Nanosecond,
    #[serde(alias = "hz")]
    Hertz,
}

#[derive(Debug, Serialize, Deserialize, Clone)]
pub struct Measurement {
    unit: MeasurementUnit,
    values: Vec<MeasurementValue>,
}
```

phacops (approved):

PR #1581, opened 2022-11-15
feat(sampling): Allow boosting individual transactions [INGEST-1706]

jjbayer:
## Background

We want to boost sample rates for key transactions. Previously, a combination of trace and event sampling rules would lead to a multiplication of sample rates. For example, a base rate of `0.1` and a boost rule of `0.5` would lead to an effective sampling rate of

```
0.1 * 0.5 == 0.05
```

The reason for this is that trace rules use a different random number seed as transaction rules (trace-ID vs. event-ID), so we effectively rolled the dice twice.

## Solution

* As soon as a rule matches, make a sampling decision (roll the dice) and exit (to prevent double sampling).
* When confronted with two different rule sets (trace rules and event rules), let event rules take precedence. This allows us to define boost rules for key transactions as well as reducing rules for health checks.

## Drawbacks

With the new behavior, it is not possible to make a trace rule (e.g. a boost rule for recent releases) take precedence over an active transaction rule.

### DIFF
--- a/relay-server/src/utils/dynamic_sampling.rs
+++ b/relay-server/src/utils/dynamic_sampling.rs
@@ -320,4 +320,81 @@ mod tests {
         );
         assert_eq!(result, SamplingResult::Drop(RuleId(1)));
     }
+
+    #[test]
+    /// The first matching rule applies, even if it is a transaction rule.
+    fn test_first_rule_applies() {
+        let sampling_config = serde_json::json!(
+            {
+                "rules": [
+                    {
+                        "sampleRate": 1,
+                        "type": "transaction",
+                        "condition": {
+                            "op": "or",
+                            "inner": [
+                            {
+                                "op": "glob",
+                                "name": "event.transaction",
+                                "value": [
+                                "my-important-transaction",
+                                ],
+                                "options": {
+                                "ignoreCase": true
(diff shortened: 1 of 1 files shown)


relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -119,6 +124,8 @@ pub fn should_keep_event(
 
         if random_number >= adjusted_sample_rate {
             return SamplingResult::Drop(rule.id);
+        } else {
+            return SamplingResult::Keep;
```
jjbayer, line 128: With this change, the sampling decision is now taken by a single rule (the first matching one).

olksdr (approved):

andriisoldatenko (approved):

iker-barriocanal (approved):
If we add a "preference" option to the rules we can avoid the drawback, but I don't know how much effort that is. The preference could look something like the following: transaction preference > trace preference > transaction > trace. It's also possible that this approach requires enough work to think about a proper solution instead.

jjbayer:
> If we add a "preference" option to the rules we can avoid the drawback

@iker I think this is a good idea, but let's not implement it until we have a product requirement for that use case.

PR #1582, opened 2022-11-15
feat(replays): Use Annotated struct definition for replay-event parsing

cmanallen:
Removes direct dependency on Serde and replaces it with Annotated<T>.  The motivation to move to Annotated<T> was to maintain strong compatibility with wider Sentry ecosystem and to streamline the implementation of PII scrubbing.

closes: https://github.com/getsentry/replay-backend/issues/183


cmanallen:
@jan-auer or @jjbayer Having a tough time finishing this one.  Trying to figure out the best place to put things and also how all these things fit together.

Goal: normalize the replay event payload and provide data scrubbing functionality that is configurable from sentry.io.

I have a `process_replay` method defined on the impl for `NormalizeProcessor`.  I'm doing platform normalization there.  In actors/processor.rs I'm normalizing the user-agent and the user's ip-address.  The ip-address comes from `envelope.meta().client_addr();`.  I'm doing this normalization process in two locations and I'm thinking they should be merged into a single method.

Questions: How should `NormalizeProcessor.process_replay` be called and should I call it in the `process_replays` method in actors/processors.rs?  How do I access client_addr() inside the normalization processor?  Should I even be using the normalization processor?

jjbayer:
> Should I even be using the normalization processor?

@cmanallen haven't reviewed in detail, but I think it might make more sense to write a separate `ReplayProcessor` (or similar) that implements `Processor` and normalizes Replay events. It might be possible to reuse the existing `PiiProcessor` for Pii processing though, by adding a `process_replay` method to it.

To pass the client IP, you could give your processor a config similar to `StoreConfig`, which contains client information and is passed into the `StoreProcessor`:

https://github.com/getsentry/relay/blob/4fc80841346a088c40bb4cd69165e80b5e8862f7/relay-general/src/store/mod.rs#L33-L35

cmanallen:
@jjbayer Updated with PII scrubbing.  Question: the `PiiProcessor` is removing the user's ip-address (as I expect) but its not removing a credit-card in the tags.  Is this unsupported or have I failed to define the correct behavior for this?

Edit: also I did not follow this advice:

> It might be possible to reuse the existing PiiProcessor for Pii processing though, by adding a process_replay method to it.

I did not know how process_replay would be called and wasn't really sure what to populate it with.  Thankfully process_value with the unmodified PiiProcessor seems to partially work.  I wonder what else I need to tweak to get to 100%?

cmanallen:
@jjbayer Also the linter is freaking out because it says my imports are unnecessary but if I remove them the tests break.

cmanallen:
@jjbayer Updated the PiiProcessor with a `process_replay` function and it gets triggered when I call `process_value` (yay).  But I lose user PII processing even with the `process_user` call.  Any ideas?

jjbayer:
> the linter is freaking out because it says my imports are unnecessary but if I remove them the tests break.

We should put a `cfg(test)` attribute on `mod test`, such that it is only compiled when tests are being executed:

https://github.com/getsentry/relay/blob/6ae0a4635d13d5971558479a5ffb5e5f8568ae8e/relay-general/src/protocol/replay.rs#L185-L186

Note that the test module is usually called `tests` (plural):

https://github.com/getsentry/relay/blob/6ae0a4635d13d5971558479a5ffb5e5f8568ae8e/relay-auth/src/lib.rs#L686-L687

> @jjbayer Updated the PiiProcessor with a `process_replay` function and it gets triggered when I call `process_value` (yay). But I lose user PII processing even with the `process_user` call. Any ideas?

Have you tried calling `process_child_values` from `process_replay`, instead of calling `process_user` explicitly? Like for example

https://github.com/getsentry/relay/blob/6ae0a4635d13d5971558479a5ffb5e5f8568ae8e/relay-general/src/pii/processor.rs#L141-L142

Let me know if there's a failing test case you can point me to, with exactly the fields you would like to have scrubbed. That might make it easier to debug.

cmanallen:
@jjbayer Updated with a failing test.  Specifically `test_scrub_pii_from_annotated_replay` in protocol/replay.  Both asserts fail.

Looking for the user.ip_address and tags["credit-card"] to be scrubbed.

relay-general/src/pii/processor.rs
```diff
@@ -162,6 +162,19 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+    fn process_replay(
+        &mut self,
+        replay: &mut Replay,
+        _meta: &mut Meta,
+        state: &ProcessingState<'_>,
+    ) -> ProcessingResult {
+        if let Some(user) = replay.user.value_mut() {
+            user.process_child_values(self, state)?;
+        }
+        Ok(())
```
jjbayer: I may have given misleading advice here earlier. We want all children of replay to be processed, not just the user field:
```suggestion
        replay.process_child_values(self, state)
```

This should fix the assertion on the credit card number.

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,412 @@
...
+
+    #[test]
+    fn test_scrub_pii_from_annotated_replay() {
+        let mut scrub_config = DataScrubbingConfig::default();
+        scrub_config.scrub_data = true;
+        scrub_config.scrub_defaults = true;
+        scrub_config.scrub_ip_addresses = true;
+
+        let pii_config = scrub_config.pii_config().unwrap().as_ref().unwrap();
+        let mut pii_processor = PiiProcessor::new(pii_config.compiled());
```
jjbayer: Concerning the IP address, I think we have the same problem as in one of the previous PRs: `scrub_ip_addresses` only clears specific fields. In order to scrub IP addresses (and other common identifiers) in any field, use

```
        let mut pii_config = PiiConfig::default();
        pii_config.applications =
            BTreeMap::from([(SelectorSpec::And(vec![]), vec!["@common".to_string()])]);

**Review by jjbayer — COMMENTED** [review]
Added two suggestions that should fix the test in question.

File: relay-general/src/pii/processor.rs (hunk 1 above)
Comment (cmanallen) [reply relay-general/src/pii/processor.rs]: This fixes the assertion on the user PII but not on the tags PII.  Progress!

File: relay-general/src/protocol/replay.rs (hunk 2 above)
Comment (cmanallen) [reply relay-general/src/protocol/replay.rs]: This config does not scrub the user's ip-address.

Comment (jjbayer) [reply relay-general/src/protocol/replay.rs]: Opened a draft PR where both asserts pass: https://github.com/getsentry/relay/pull/1705

Comment (cmanallen) [reply relay-general/src/protocol/replay.rs]: Ah, my test case was just bad... Thank you!

File: relay-general/src/protocol/replay.rs [hunk 3]
```diff
@@ -0,0 +1,409 @@
...
+use crate::store::user_agent::normalize_user_agent_generic;
+use crate::types::{Annotated, Array};
+use crate::user_agent;
+use std::net::IpAddr as RealIPAddr;
+
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+#[metastructure(process_func = "process_replay", value_type = "Replay")]
+pub struct Replay {
+    pub event_id: Annotated<String>,
```
jjbayer: Should we type this more strictly to `EventId` or `Uuid`?

https://github.com/getsentry/relay/blob/4cb923f0667da94801039ff2cbd99635221dd97b/relay-general/src/protocol/event.rs#L193

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,409 @@
...
+    pub replay_id: Annotated<String>,
```
jjbayer: `replay_id` is typed as an `EventId` in other places. Should we use the same type here? https://github.com/getsentry/relay/blob/4cb923f0667da94801039ff2cbd99635221dd97b/relay-server/src/actors/store.rs#L569

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    #[metastructure(skip_serialization = "empty")]
+    pub user: Annotated<User>,
+
+    #[metastructure(skip_serialization = "empty")]
+    pub request: Annotated<Request>,
+
+    #[metastructure(field = "sdk")]
+    #[metastructure(skip_serialization = "empty")]
+    pub sdk: Annotated<ClientSdkInfo>,
```
jjbayer, line 238: Let's put doc comments on top of every `pub` member with a short explanation of what the field is for, like [here](https://github.com/getsentry/relay/blob/4cb923f0667da94801039ff2cbd99635221dd97b/relay-general/src/protocol/event.rs#L174).

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,409 @@
...
+        replay_value.normalize(None, None);
+
+        let loaded_browser_context = replay_value
+            .contexts
+            .value_mut()
+            .as_mut()
+            .unwrap()
+            .get("browser")
+            .unwrap()
+            .clone();
```
jjbayer: Not sure if it helps you here, but for tests, we have a `get_path!` macro, for example https://github.com/getsentry/relay/blob/4cb923f0667da94801039ff2cbd99635221dd97b/relay-general/src/store/normalize.rs#L1604

jjbayer (commented):
@cmanallen is this PR ready for final review? Or are you still making changes?

cmanallen:
@jjbayer Yes ready for final review!  Question: the PII scrubber removes values and replaces them with `[email]` (for example).  On the Snuba side I assume I would want to nullify these values?  Do you know what pattern Sentry uses for managing these scrubbed values?  I looked at some of the snuba processors and it seems like they don't acknowledge these sentinel values exist.  Are there additional steps I need to take to normalize this payload?

jjbayer:
> Do you know what pattern Sentry uses for managing these scrubbed values?

As far as I know, Sentry & Snuba do not treat scrubbed values differently from regular values, i.e. they are still strings, just with a different value. If you'd rather have these values set to `null` instead of being overwritten, you could look into using the [`Remove`](https://github.com/getsentry/relay/blob/4cb923f0667da94801039ff2cbd99635221dd97b/relay-general/src/pii/redactions.rs#L41-L42) redaction instead of `Replace`. E.g. by configuring `@ip:remove` instead of `@ip:replace` (which is part of `@common`):

https://github.com/getsentry/relay/blob/4cb923f0667da94801039ff2cbd99635221dd97b/relay-general/src/pii/builtin.rs#L103-L106

cmanallen:
@jjbayer Currently I'm getting the `PiiConfig` from `ProjectConfig`?  Would using `ip@remove` require implementing a static PII config?

jjbayer:
> @jjbayer Currently I'm getting the `PiiConfig` from `ProjectConfig`? Would using `ip@remove` require implementing a static PII config?

@cmanallen the short answer is yes.

Do I understand correctly that in this version, replay events are scrubbed with `PiiConfig` from project config but replay recordings are still scrubbed with a static config? Was it a deliberate product decision to use the one from project configs here? If not, I would apply a static config to both, especially when we need to erase fields to `null` rather than doing the default scrubbing.

cmanallen:
@jjbayer We're happy with the configured behavior (`[ip]` instead of null).  Is this ready to be approved?

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,567 @@
...
+
+        let loaded_os_context = replay_value
+            .contexts
+            .value_mut()
+            .as_mut()
+            .unwrap()
+            .get("client_os")
+            .unwrap()
+            .clone();
```
jjbayer: No need to take a mutable reference or clone here, please simplify to 

```suggestion
            .value()
            .unwrap()
            .get("client_os")
            .unwrap();
```

and then compare to a reference in the assert (i.e. `&os_context`).

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,567 @@
...
+    #[test]
+    fn test_loose_type_requirements() {
+        let payload = include_str!("../../tests/fixtures/replays/replay_failure_22_08_31.json");
+
+        let mut replay: Annotated<Replay> = Annotated::from_json(payload).unwrap();
+        let replay_value = replay.value_mut().as_mut().unwrap();
+        replay_value.normalize(None, None);
+
+        let user = replay_value.user.value_mut().as_mut().unwrap();
+        assert!(user.ip_address.value_mut().as_mut().unwrap().as_str() == "127.1.1.1");
```
jjbayer:
```suggestion
        let user = replay_value.user.value().unwrap();
        assert!(user.ip_address.value().unwrap().as_str() == "127.1.1.1");
```

etc.

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,567 @@
...
+        let maybe_ip_address = replay
+            .value()
+            .unwrap()
+            .user
+            .value()
+            .unwrap()
+            .ip_address
+            .value();
+
+        assert_eq!(maybe_ip_address.unwrap().as_str(), "[ip]");
```
jjbayer:
```suggestion
        assert_eq!(get_value!(replay.user.ip_address!).as_str(), "[ip]");
```

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,567 @@
...
+    fn normalize_user_agent(&mut self, default_user_agent: Option<&str>) {
+        let user_agent = match user_agent::get_user_agent_generic(&self.request) {
+            Some(ua) => ua,
+            None => match default_user_agent {
+                Some(dua) => dua,
+                None => return,
+            },
+        };
+
+        if let Some(contexts) = self.contexts.value_mut() {
```
jjbayer: Why not just use `get_or_insert_with`, as we do here?

https://github.com/getsentry/relay/blob/c0d39b3fad33d97a28d89f5cd9b05d7b040cb3f3/relay-general/src/store/normalize/user_agent.rs#L30

If you don't want an empty `contexts` to show up in the resulting payload, you could use `#[metastructure(skip_serialization = "empty")]`.

relay-general/src/user_agent.rs
```diff
@@ -53,7 +54,11 @@ pub fn init_parser() {
 /// Returns `Some` if the event's request interface contains a `user-agent` header. Returns `None`
 /// otherwise.
 pub fn get_user_agent(event: &Event) -> Option<&str> {
-    let request = event.request.value()?;
+    get_user_agent_generic(&event.request)
+}
+
+pub fn get_user_agent_generic(request: &Annotated<Request>) -> Option<&str> {
```
jjbayer: Let's just remove the wrapper function and call this one `get_user_agent`.

relay-server/src/actors/processor.rs
```diff
@@ -1009,32 +1011,66 @@ impl EnvelopeProcessorService {
...
+                    &item.payload(),
+                    &state.project_state.config,
+                    client_addr,
+                    user_agent,
+                );
+
+                match result {
                     Ok(replay) => {
-                        item.set_payload(ContentType::Json, &replay[..]);
+                        item.set_payload(ContentType::Json, replay.to_json().unwrap().as_bytes());
```
jjbayer: Instead of unwrapping, can we log an error here and return `false`? See for example https://github.com/getsentry/relay/blob/e87b92e261ffe4f563c8e5519c12c7c1e756984d/relay-server/src/actors/processor.rs#L657-L665.

relay-server/src/actors/processor.rs
```diff
@@ -1009,32 +1011,66 @@ impl EnvelopeProcessorService {
     }
 
     /// Remove replays if the feature flag is not enabled
-    fn process_replays(&self, state: &mut ProcessEnvelopeState) {
+    fn process_replays(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
```
jjbayer: It looks to me like this function never returns an `Err`, so can we make it infallible by not returning anything?

relay-general/src/store/normalize/user_agent.rs
```diff
@@ -8,70 +8,102 @@ use std::fmt::Write;
...
+            contexts.add(Context::Device(Box::new(device_context)));
+        }
     }
 
-    // avoid conflicts with OS-context sent by a serverside SDK by using `contexts.client_os`
-    // instead of `contexts.os`. This is then preferred by the UI to show alongside device and
-    // browser context.
-    //
-    // Why not move the existing `contexts.os` into a different key on conflicts? Because we still
-    // want to index (derive tags from) the SDK-sent context.
```
jjbayer, line 55: What happened to this comment? Can we restore it?

relay-general/src/protocol/replay.rs
```diff
@@ -0,0 +1,567 @@
...
+        assert!(user.username.value_mut().is_none());
+        assert!(user.email.value_mut().as_mut().unwrap().as_str() == "email@sentry.io");
+        assert!(user.id.value_mut().as_mut().unwrap().as_str() == "1");
+    }
+
+    #[test]
+    fn test_scrub_pii_from_annotated_replay() {
+        let mut pii_config = PiiConfig::default();
+        pii_config.applications =
+            BTreeMap::from([(SelectorSpec::And(vec![]), vec!["@common".to_string()])]);
```
jjbayer: When we fixed this test, I did not realize you were planning to read PII config from project configs instead of using a hard-coded one. So it might make sense to change this test back to a pii config which looks like the most common datascrubbing settings coming from sentry:

https://github.com/getsentry/relay/blob/e87b92e261ffe4f563c8e5519c12c7c1e756984d/relay-general/src/pii/convert.rs#L196-L203

But with that, even the credit card scrubbing seems to fail and I'm not sure why. I can investigate later this week.

jjbayer (approved):
Looks good to me! Could you please add a PR description including a motivation for this change?

cmanallen:
@jjbayer Updated description.  Thanks for helping me through this marathon of a PR!  :D

PR #1583, opened 2022-11-15
feat(system): Log service back pressure as metric [INGEST-1630]

jan-auer:
This adds central back pressure monitoring via statsd metrics to all
services that use the new tokio-based service framework. Channels track
the number of pending messages and emit a gauge metric with that number.

The metric is `service.back_pressure` and tagged with a `service` tag to
identify the specific service. The metric is debounced once per second.
Between these intervals, the back pressure number is not checked, so
this metric is not suitable to detect very short-lived spikes.
Additionally, the current implementation requires that the service makes
progress by polling `recv`.

The purpose of this metric is to detect lasting increases in backlogs
that indicate bottlenecks or degraded system behavior.

### DIFF
--- a/relay-statsd/src/lib.rs
+++ b/relay-statsd/src/lib.rs
@@ -62,8 +62,7 @@ use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 
 use cadence::{
-    BufferedUdpMetricSink, Metric, MetricBuilder, QueuingMetricSink, SpyMetricSink, StatsdClient,
-    UdpMetricSink,
+    BufferedUdpMetricSink, Metric, MetricBuilder, QueuingMetricSink, StatsdClient, UdpMetricSink,
 };
 use parking_lot::RwLock;
 use rand::distributions::{Distribution, Uniform};
@@ -164,8 +163,9 @@ pub fn set_client(client: MetricsClient) {
 }
 
 /// Set a test client for the period of the called function (only affects the current thread).
+#[cfg(feature = "test")]
 pub fn with_capturing_test_client(f: impl FnOnce()) -> Vec<String> {
-    let (rx, sink) = SpyMetricSink::new();
+    let (rx, sink) = cadence::SpyMetricSink::new();
     let test_client = MetricsClient {
         statsd_client: StatsdClient::from_sink("", sink),
         default_tags: Default::default(),
@@ -573,7 +573,6 @@ macro_rules! metric {
 
 #[cfg(test)]
 mod tests {
-
     use cadence::{NopMetricSink, StatsdClient};
 
     use crate::{set_client, with_capturing_test_client, with_client, GaugeMetric, MetricsClient};
@@ -608,7 +607,7 @@ mod tests {
         });
 
         assert_eq!(
-            captures.as_slice(),
+            captures,
             [
                 "foo:123|g|#server:server1,host:host1",
                 "bar:456|g|#server:server2,host:host2"
--- a/relay-system/src/service.rs
+++ b/relay-system/src/service.rs
@@ -540,8 +540,9 @@ mod tests {
 
         fn spawn_handler(self, mut rx: Receiver<Self::Interface>) {
             tokio::spawn(async move {
-                tokio::time::sleep(BACKLOG_INTERVAL * 1.5).await;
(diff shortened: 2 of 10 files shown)


jan-auer:
I'll add unit tests that assert emitted metrics; otherwise this is ready for review.

relay-system/src/service.rs
```diff
@@ -352,15 +365,78 @@ impl<I: Interface> Addr<I> {
...
+    ///
+    /// If there are no messages in the channel's buffer, but the channel has
+    /// not yet been closed, this method will sleep until a message is sent or
+    /// the channel is closed.
+    pub async fn recv(&mut self) -> Option<I> {
+        loop {
+            tokio::select! {
+                biased;
+
+                _ = self.interval.tick() => {
```
olksdr, line 391: I'm not sure about this select now. We will be missing some metrics, since once we get the message we exit this receive method. 

I'm thinking maybe it will be better to spawn the `tokio::task` and just report the metrics from there.

olksdr (commented):
Overall it looks great to me. Just left a small thought

relay-system/src/service.rs
jan-auer: `recv` is called in a loop by the service, so this will run continuously. All that this inner loop and `select!` does is inject into the existing external loop and slide some metrics in.

What this means is that the metric cannot be logged if the service does not make any progress. On the other hand, as long as the service makes slow progress, we'll get updates.

relay-statsd/src/lib.rs
```diff
@@ -164,8 +163,9 @@ pub fn set_client(client: MetricsClient) {
 }
 
 /// Set a test client for the period of the called function (only affects the current thread).
+#[cfg(feature = "test")]
```
jjbayer, line 166: This is nice, I remember removing `#[cfg(test)]` here because it did not work across crates.

relay-system/src/service.rs
```diff
@@ -305,12 +314,14 @@ pub trait FromMessage<M>: Interface {
 /// long as the service is running. It can be freely cloned.
 pub struct Addr<I: Interface> {
     tx: mpsc::UnboundedSender<I>,
+    backlog: Arc<AtomicU64>,
```
jjbayer: nit: To me, "backlog" refers to the actual list/queue of backlogged items, while this is merely a counter, so I would name it something like `backlog_size` or `queue_size`.

relay-system/src/service.rs
```diff
@@ -352,15 +365,78 @@ impl<I: Interface> Addr<I> {
...
     let (tx, rx) = mpsc::unbounded_channel();
-    (Addr { tx }, rx)
+
+    let addr = Addr {
+        tx,
+        backlog: backlog.clone(),
+    };
+
+    let mut interval = tokio::time::interval(BACKLOG_INTERVAL);
+    interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
```
jjbayer, line 430: TIL

relay-system/src/service.rs
```diff
@@ -426,8 +502,105 @@ pub trait Service: Sized {
...
+
+    #[test]
+    fn test_backpressure_metrics() {
+        let rt = tokio::runtime::Builder::new_current_thread()
+            .enable_time()
+            .build()
+            .unwrap();
+
+        let _guard = rt.enter();
+        tokio::time::pause();
```
jjbayer, line 561: How does this work? Is the time still paused after calling `tokio::time::sleep`? That is, calling `sleep(X)` increases the paused `now` time by `X`?

jjbayer (approved):

olksdr (approved):

relay-system/src/service.rs
jan-auer: The relevant information is in the ["Auto-advance" section](https://docs.rs/tokio/latest/tokio/time/fn.pause.html#auto-advance) on the docs of pause: The runtime skips ahead to the next waiting timer, in our case a sleep. We can think about it like skipping idle time. Also, the timer does not advance in between, so multiple calls to `Instant::now()` will always give the same instant back.

There is also `tokio::time::advance()`, but that does not run all paused timers, which we want in this case.

PR #1585, opened 2022-11-16
[DO NOT MERGE] Add metric on envelope process errors

iker-barriocanal:
**DO NOT MERGE**

The error logged in `relay-server/src/actors/processor.rs` has not reached Sentry in a long time, and it would have been critical to identifying an incident earlier. We don't know what the cause of the error not popping in Sentry is, so I'm adding a metric to see if this code path is ever reached in the first place.

The goal is to deploy this in a canary and let it run for some time until we collect data to determine if that piece of code was actually reached.

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -2139,6 +2139,10 @@ impl EnvelopeProcessorService {
                 };
             }
             Err(error) => {
+                relay_statsd::metric!(
+                    counter(RelayCounters::HandleProcessEnvelopeError) += 1,
+                    internal_error = error.is_internal()
+                );
                 // Errors are only logged for what we consider infrastructure or implementation
                 // bugs. In other cases, we "expect" errors and log them as debug level.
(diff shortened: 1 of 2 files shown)


jjbayer (approved):

relay-server/src/actors/processor.rs
```diff
@@ -2139,6 +2139,10 @@ impl EnvelopeProcessorService {
                 };
             }
             Err(error) => {
+                relay_statsd::metric!(
+                    counter(RelayCounters::HandleProcessEnvelopeError) += 1,
+                    internal_error = if error.is_internal() { "true" } else { "false" }
```
```suggestion
                    internal_error = &error.is_internal().to_string()
```

would something like this work?  This `if` just looks strange here.

iker-barriocanal: Good suggestion 👍. I misread the documentation and understood there was no `ToString` or `FromBool` implemented.

iker-barriocanal:
End of experiment.

PR #1586, opened 2022-11-16
chore(kafka): Log payload size with extras

olksdr:
make sure to record payload length but use `extra`s for this 

#skip-changelog


iker-barriocanal (approved):
LGTM.

I see in the [docs](https://docs.sentry.io/platforms/rust/enriching-events/context/#additional-data) `set_extra` is deprecated and we should use `set_context` instead, but it adds more friction and it's not related to this PR so that's ok for now.

PR #1590, opened 2022-11-17
fix(pii): scrub PII in span description [INGEST-1709]

olksdr:
Filter description if it contains PII

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Internal**:
 
 - Emit a `service.back_pressure` metric that measures internal back pressure by service. ([#1583](https://github.com/getsentry/relay/pull/1583))
+- Scrub PII in the span description. ([#1590](https://github.com/getsentry/relay/pull/1590))
 
 ## 22.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 **Internal**:
 
 - Emit a `service.back_pressure` metric that measures internal back pressure by service. ([#1583](https://github.com/getsentry/relay/pull/1583))
-- Scrub PII in the span description. ([#1590](https://github.com/getsentry/relay/pull/1590))
+- Always scrub PII in the span description and data. So far, the user had to configure this by hand. ([#1590](https://github.com/getsentry/relay/pull/1590))
 
 ## 22.11.0
 
--- a/relay-general/src/pii/processor.rs
+++ b/relay-general/src/pii/processor.rs
@@ -351,11 +351,13 @@ fn insert_replacement_chunks(rule: &RuleRef, text: &str, output: &mut Vec<Chunk<
 
 #[cfg(test)]
 mod tests {
-    use crate::pii::{PiiConfig, ReplaceRedaction};
+    use chrono::{TimeZone, Utc};
+
+    use crate::pii::{DataScrubbingConfig, PiiConfig, ReplaceRedaction};
     use crate::processor::process_value;
     use crate::protocol::{
         Addr, DebugImage, DebugMeta, Event, ExtraValue, Headers, LogEntry, NativeDebugImage,
-        Request, TagEntry, Tags,
+        Request, Span, SpanId, SpanStatus, TagEntry, Tags, TraceId,
     };
     use crate::testutils::assert_annotated_snapshot;
     use crate::types::{Annotated, Object, Value};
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Internal**:
 
 - Emit a `service.back_pressure` metric that measures internal back pressure by service. ([#1583](https://github.com/getsentry/relay/pull/1583))
+- Scrub PII in the span description. ([#1590](https://github.com/getsentry/relay/pull/1590))
```
```suggestion
- Always scrub PII in the span description. So far, the user had to configure this by hand. ([#1590](https://github.com/getsentry/relay/pull/1590))
```

relay-general/src/protocol/span.rs
```diff
@@ -18,7 +18,7 @@ pub struct Span {
     pub exclusive_time: Annotated<f64>,
 
     /// Human readable description of a span (e.g. method URL).
-    #[metastructure(pii = "maybe")]
+    #[metastructure(pii = "true")]
```
marandaneto, line 21: `data` field is also `maybe` and some SDKs also set the URL in there.
See this https://github.com/getsentry/sentry/blob/fef9c695a1a7d3384fb3ce7ec6c264632e77061d/static/app/components/events/interfaces/spans/spanTreeModel.spec.tsx#L47
and https://github.com/getsentry/sentry/blob/fef9c695a1a7d3384fb3ce7ec6c264632e77061d/src/sentry/interfaces/spans.py#L33-L38

CHANGELOG.md
jjbayer: I'm not very consistent with this myself, but let's also add this to `py/CHANGELOG.md`. See https://github.com/getsentry/relay#instructions-for-changelogs.

jjbayer (approved):
LGTM. I would add a bit more explanation to the PR description (that this change was motivated by an incident, etc.).

marandaneto:
Can we add a test for breadcrumbs as well?
https://github.com/getsentry/relay/blob/87181c0935ab56093d1db5925ffe7a44cbdefe1e/relay-general/src/protocol/breadcrumb.rs#L104

SDKs sets the `data.url` as well https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/

marandaneto:
`request.url` is also `maybe`https://github.com/getsentry/relay/blob/87181c0935ab56093d1db5925ffe7a44cbdefe1e/relay-general/src/protocol/request.rs#L440
Should we change that too? Same for `query_string`.

olksdr:
> Can we add a test for breadcrumbs as well?
> 
> https://github.com/getsentry/relay/blob/87181c0935ab56093d1db5925ffe7a44cbdefe1e/relay-general/src/protocol/breadcrumb.rs#L104
> 
> SDKs sets the `data.url` as well https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/

this already is tested in https://github.com/getsentry/relay/blob/87181c0935ab56093d1db5925ffe7a44cbdefe1e/relay-general/src/pii/convert.rs#L1197

relay-general/src/protocol/span.rs
olksdr: yep I  will add this

CHANGELOG.md
olksdr: Will do

olksdr:
> `request.url` is also `maybe`
> 
> https://github.com/getsentry/relay/blob/87181c0935ab56093d1db5925ffe7a44cbdefe1e/relay-general/src/protocol/request.rs#L440
> 
> 
> Should we change that too? Same for `query_string`.

we can add, but it can be a lot of false positives, if anything is in the URL , domain name which will contain the key words we look for we will strip entire URL. Is that ok?

relay-general/src/pii/processor.rs
```diff
@@ -426,6 +428,36 @@ mod tests {
...
+            ..Default::default()
+        });
+        let mut event = Annotated::new(Event {
+            spans: Annotated::new(vec![span]),
+            ..Default::default()
+        });
+
+        let mut processor = PiiProcessor::new(config.compiled());
+        process_value(&mut event, &mut processor, ProcessingState::root()).unwrap();
+        assert_annotated_snapshot!(event);
```
iker-barriocanal, line 458: nit: we're only interested in ensuring the description is `[Filtered]`, so it may be easier to read that assertion over a snapshot?

relay-general/src/pii/processor.rs
```diff
@@ -426,6 +428,36 @@ mod tests {
         assert_annotated_snapshot!(event);
     }
 
+    #[test]
+    fn test_span_default_rules() {
+        let ds_config = DataScrubbingConfig {
+            scrub_data: true,
+            scrub_defaults: true,
```
iker-barriocanal, line 435: Question: what fields are the default fields?

relay-general/src/pii/processor.rs
```diff
@@ -426,6 +428,36 @@ mod tests {
         assert_annotated_snapshot!(event);
     }
 
+    #[test]
+    fn test_span_default_rules() {
+        let ds_config = DataScrubbingConfig {
```
iker-barriocanal, line 433: Is `exclude_fields` the vector of safe fields? Should we also test that?

iker-barriocanal (approved):

marandaneto:
> > `request.url` is also `maybe`
> > https://github.com/getsentry/relay/blob/87181c0935ab56093d1db5925ffe7a44cbdefe1e/relay-general/src/protocol/request.rs#L440
> > 
> > Should we change that too? Same for `query_string`.
> 
> we can add, but it can be a lot of false positives, if anything is in the URL , domain name which will contain the key words we look for we will strip entire URL. Is that ok?

mm I am unsure, this will definatelly degrade the feature (filtering the whole URL instead of the only key words), Mobile already scrub out the query parameters (by not sending them), if you think this is ok, we are good already.

olksdr:
> mm I am unsure, this will definatelly degrade the feature (filtering the whole URL instead of the only key words), Mobile already scrub out the query parameters (by not sending them), if you think this is ok, we are good already.

@marandaneto ,  checked this with @jjbayer - we already normalize the URL and push query string from `url` field to `query_string` which has `pii=true` and will be scrubbed

relay-general/src/pii/processor.rs
olksdr: we test this in other tests already, I don't think we have to add another test for this here

relay-general/src/pii/processor.rs
olksdr: sorry, I don't really understand the question

relay-general/src/pii/processor.rs
olksdr: The test is to ensure that the things are working as expected now, and it will show the problem if we have the regression. But I do not really care who this implemented.

relay-general/src/pii/processor.rs
iker-barriocanal: Sounds good then 👍

relay-general/src/pii/processor.rs
iker-barriocanal: What is the difference between `scrub_data` and `scrub_defaults`? What fields does `scrub_defaults` refer to?

olksdr: ah, so first is enabling the data scrubbing and the second is enabling default data scrubbers provided by relay, e.g. remove tokens, passwords, auth, ssh keys etc.

PR #1591, opened 2022-11-17
ref(sampling): Switch default sampling mode to "received" [INGEST-1715]

jan-auer:
SDKs with client side sampling reduce the number of items sent to Relay, where dynamic sampling
occurs. The sampling mode controlls whether the sample rate is relative to the original
population of items before client-side sampling, or relative to the number received by Relay
after client-side sampling.

There are two modes:
 - `received` (default): Server-side dynamic sampling occurs on top of potential client-side sampling in the SDK. For
    example, if the SDK samples at 50% and the server sampling rate is set at 10%, the resulting
    effective sample rate is 5%.
 - `total`: Server-side sampling compensates potential client-side sampling in the SDK. For example, if
   the SDK samples at 50% and the server sampling rate is set at 10%, the resulting effective
   sample rate is 10%.
   In this mode, the server sampling rate is capped by the client's sampling rate. Rules with a
   higher sample rate than what the client is sending are effectively inactive.

For beta customers of dynamic sampling that expect the former behavior, this
will be switched to `total` temporarily.

### DIFF
--- a/relay-server/src/utils/dynamic_sampling.rs
+++ b/relay-server/src/utils/dynamic_sampling.rs
@@ -46,7 +46,7 @@ fn check_unsupported_rules(
     Ok(())
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug, PartialEq)]
 struct SamplingSpec {
     sample_rate: f64,
     rule_id: RuleId,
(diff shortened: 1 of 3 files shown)


olksdr (approved):

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -62,23 +69,42 @@ fn get_trace_sampling_rule<'a>(
...
+    let sample_rate = match (sampling_context, sampling_config.mode) {
+        (Some(ctx), SamplingMode::Total) => ctx.adjusted_sample_rate(rule.sample_rate),
+        _ => rule.sample_rate,
+    };
+
+    Ok(Some(SamplingSpec {
+        sample_rate,
+        rule_id: rule.id,
+        seed: event_id.0,
+    }))
```
jjbayer, line 107: Can we add tests for these functions? I would even volunteer to write some :)

jan-auer: You read my mind - was already in the works. See 5081542700579e0267beec56bd01ce77acb25193

jjbayer (approved):

PR #1600, opened 2022-11-18
feat(replays): Scrub ip-addresses from recording

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/211

#skip-changelog

### DIFF
--- a/relay-replays/src/recording.rs
+++ b/relay-replays/src/recording.rs
@@ -1,10 +1,12 @@
 use std::borrow::Cow;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::fmt::Display;
 use std::io::{Read, Write};
 
-use relay_general::pii::{DataScrubbingConfig, PiiProcessor};
-use relay_general::processor::{FieldAttrs, Pii, ProcessingState, Processor, ValueType};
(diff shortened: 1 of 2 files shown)


cmanallen:
@jan-auer @jjbayer `ip-address` scrubbing isn't working with this config for some reason.  Do you have any thoughts?  No rush, I won't be working on this until next week.

jjbayer:
> `ip-address` scrubbing isn't working with this config for some reason. Do you have any thoughts? No rush, I won't be working on this until next week.

@cmanallen Our default IP filter looks for specific fields in the request headers / event payload. Since we're hard-coding config anyway, maybe we should construct a `PiiConfig` directly instead of going trough `DatascrubbingConfig`:

```rust
    let mut pii_config = PiiConfig::default();
    pii_config.applications =
        BTreeMap::from([(SelectorSpec::And(vec![]), vec!["@common".to_string()])]);
```

this will replace your strings with `[creditcard]` and `[ip]` respectively. Background: The `@common` rule set includes the following builtin rules:

https://github.com/getsentry/relay/blob/8ccb5d821fc775b2131cd124cda7f7a4af5dee1e/relay-general/src/pii/builtin.rs#L39-L46

jjbayer (approved):

PR #1616, opened 2022-11-21
fix(logging): Log error for unexpected event drops

### LINKED ISSUE #1609: [INC-237] Missing alert for dropped envelopes
There should be a Relay error on [sentry.my.sentry.io](http://sentry.my.sentry.io/) + alert whenever envelopes get dropped, in https://getsentry.atlassian.net/browse/INC-237 it did not trigger.

[Alert](https://sentry.my.sentry.io/organizations/sentry/alerts/rules/details/17/) looks for “error processing envelope”, but that did not occur since Aug 8 (INC-192), see [Discover](https://sentry.my.sentry.io/organizations/sentry/discover/results/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=9&project=4&query=event.type%3Adefault+%22error+processing+envelope%22&sort=-timestamp&statsPeriod=90d&yAxis=count%28%29).

That error [still exists in Relay code](https://github.com/getsentry/relay/blob/07eff5a6a1942829f34307b9593961480961e284/relay-server/src/actors/processor.rs#L2123), but maybe that code path is never reached?

We did get a lot of “error fetching project state, deadline exceeded” errors ([Discover](https://sentry.my.sentry.io/organizations/sentry/discover/results/?display=top5&end=2022-10-19T16%3A17%3A29.000&field=project&field=title&field=count%28%29&name=All+Events&project=9&project=4&query=event.type%3Adefault+%21message%3A%22unit+mismatch%22&sort=-count&start=2022-10-19T15%3A45%3A51.000&yAxis=count%28%29))

jjbayer:
If we drop an envelope because we cannot load its project state, we should log an error.

After we extend the alert to include the new log message, we can close https://github.com/getsentry/relay/issues/1609.

### DIFF
--- a/relay-server/src/actors/outcome.rs
+++ b/relay-server/src/actors/outcome.rs
@@ -208,6 +208,22 @@ impl Outcome {
             Outcome::Abuse => None,
         }
     }
+
+    /// Returns true if there is a bug or an infrastructure problem causing event loss.
+    ///
+    /// This can happen when we introduce bugs or during incidents.
+    ///
+    /// During healthy operation, this should always return false.
+    pub fn is_unexpected(&self) -> bool {
+        matches!(
+            self,
+            Outcome::Invalid(
+                DiscardReason::Internal
+                    | DiscardReason::ProjectState
+                    | DiscardReason::ProjectStatePii,
+            )
+        )
+    }
 }
 
 impl fmt::Display for Outcome {
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -160,8 +160,9 @@ impl ProcessingError {
         }
     }
 
-    fn is_internal(&self) -> bool {
-        self.to_outcome() == Some(Outcome::Invalid(DiscardReason::Internal))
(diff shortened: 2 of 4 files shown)


olksdr (approved):
lgtm

relay-server/src/actors/outcome.rs
```diff
@@ -208,6 +208,22 @@ impl Outcome {
...
+    /// This can happen when we introduce bugs or during incidents.
+    ///
+    /// During healthy operation, this should always return false.
+    pub fn is_unexpected(&self) -> bool {
+        matches!(
+            self,
+            Outcome::Invalid(
+                DiscardReason::Internal
+                    | DiscardReason::ProjectState
+                    | DiscardReason::ProjectStatePii,
```
jjbayer, line 223: These are only emitted when the project config cannot be parsed:

https://github.com/getsentry/relay/blob/a9297f9cecb00b2caa09033a3c12438cc8c23b76/relay-server/src/actors/outcome.rs#L330-L335

Issue #1615, opened 2022-11-21
Add OpenTelemetry metrics to Relay


AbhiPrasad:
To get some high level insights into OpenTelemetry usage, we want to track OpenTelemetry projects via Relay statsd metrics. This will also help us make decisions as we move forward with server-side OpenTelemetry logic.

- [x] Add OpenTelemetry Context to schema https://github.com/getsentry/relay/pull/1617
  - See: https://develop.sentry.dev/sdk/performance/opentelemetry/#opentelemetry-context
- [x] Add counter that represents event volume of OpenTelemetry Events to Relay. https://github.com/getsentry/relay/pull/1618

PR #1617, opened 2022-11-21
feat(protocol): Add OpenTelemetry Context

AbhiPrasad:
ref: https://github.com/getsentry/relay/issues/1615

Add a new Context field that is sent with OpenTelemetry Events. Documented here: https://develop.sentry.dev/sdk/performance/opentelemetry/#opentelemetry-context.

### DIFF
--- a/relay-general/src/protocol/contexts/otel.rs
+++ b/relay-general/src/protocol/contexts/otel.rs
@@ -8,13 +8,13 @@ use crate::types::{Annotated, Object, Value};
 pub struct OtelContext {
     /// Attributes of the OpenTelemetry span that maps to a Sentry event.
     ///
-    /// https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/trace/v1/trace.proto#L174-L186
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):

relay-general/src/protocol/contexts/otel.rs
```diff
@@ -0,0 +1,31 @@
+use crate::types::{Annotated, Object, Value};
+
+/// OpenTelemetry Context
+///
+/// If an event has this context, it was generated from a OpenTelemetry signal (trace, metric, log).
```
AbhiPrasad:
```suggestion
/// If an event has this context, it was generated from an OpenTelemetry signal (trace, metric, log).
```

jjbayer (commented):
Looks good to me, shall we add a roundtrip test to the new module? See for example https://github.com/getsentry/relay/blob/a9297f9cecb00b2caa09033a3c12438cc8c23b76/relay-general/src/protocol/contexts/trace.rs#L198.

AbhiPrasad:
> Looks good to me, shall we add a roundtrip test to the new module? See for example 

Done in [87fac38](https://github.com/getsentry/relay/pull/1617/commits/87fac3816273c19cc62819c0f9c07eee333d73d8)!

jjbayer (approved):

PR #1618, opened 2022-11-21
feat(server): Track metrics for OpenTelemetry events

### LINKED ISSUE #1615: Add OpenTelemetry metrics to Relay

AbhiPrasad:
fixes https://github.com/getsentry/relay/issues/1615

This PR adds a statsd metric for OpenTelemetry events. It allows us to see at a high level the volume of OpenTelemetry and what SDKs they are being sent from.

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -1639,6 +1639,20 @@ impl EnvelopeProcessorService {
                     sdk = envelope.meta().client_name().unwrap_or("proprietary"),
                     platform = event.platform.as_str().unwrap_or("other"),
                 );
+
+                let otel_context = event
+                    .contexts
+                    .value()
+                    .and_then(|contexts| contexts.get("otel"))
+                    .and_then(Annotated::value);
+
+                if otel_context.is_some() {
+                    metric!(
+                        counter(RelayCounters::OpenTelemetryEvent) += 1,
+                        sdk = envelope.meta().client_name().unwrap_or("proprietary"),
+                        platform = event.platform.as_str().unwrap_or("other"),
+                    );
+                }
             }
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 **Internal**:
 
 - Emit a `service.back_pressure` metric that measures internal back pressure by service. ([#1583](https://github.com/getsentry/relay/pull/1583))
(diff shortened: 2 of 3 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -1639,6 +1639,20 @@ impl EnvelopeProcessorService {
...
                 );
+
+                let otel_context = event
+                    .contexts
+                    .value()
+                    .and_then(|contexts| contexts.get("otel"))
+                    .and_then(Annotated::value);
+
+                if otel_context.is_some() {
+                    metric!(
```
iker-barriocanal, line 1650: Collecting the metrics in `finalize_event` means you'll get metrics for events relay drops as part of light normalization), is this your intention?

jjbayer (approved):

relay-server/src/actors/processor.rs
iker-barriocanal, line 1650: How many OTel events (and metrics) are we expecting, and how big is the cardinality of these counters (are most events coming from the same SDK) are in the statsd client in relay? If the number is not large it's ok. My concern is that we've been in the maximum threshold before, and the additional metrics are dropped.

iker-barriocanal (commented):
A couple of questions to better understand the intention, not blocking.

relay-server/src/actors/processor.rs
AbhiPrasad: We should be consuming at the same volume as transaction source information, which is set above in `RelayCounters::EventTransactionSource` with the same cardinality, and we've been able to manage that fine.

Most events should come from 4 SDKs atm (Node, Java, Python, Ruby), so it's even less than transaction source at the moment.

AbhiPrasad: Yes this was my intention, the same as transaction source metrics we are collecting.

This will be the estimate we use to understand the consequences of doing more complicated processing of opentelemetry events in Relay, details of which we will share with the Ingest team very soon.

Issue #1602, opened 2022-11-21
Update ProjectCache Actor


HazAT:
- [x] https://github.com/getsentry/relay/pull/1619
- [x] https://github.com/getsentry/relay/pull/1622
- [x] https://github.com/getsentry/relay/pull/1635

PR #1485, opened 2022-09-20
ref(actix): Migrate the RelayCache actor

jan-auer:
Updates the Relay cache to run as Tokio service. The upstream query is moved to a background task with its own `mpsc`. All other operations, such as updating the internal map and retrieving relay information, run sequentially in the service.

The service operates exactly as before, including an pre-existing race condition when a relay is queried while a fetch is running.

#skip-changelog

PR #1619, opened 2022-11-22
feat(system): Add broadcast behavior for debounced services

jan-auer:
The `ProjectCache` and `RelayCache` actors both follow a common pattern: They
maintain a map of cached objects that they populate asynchronously. There is a
message in their public interface to retrieve data from this cache.

Since the `RelayCache` was ported in #1485, it keeps a list of senders for each
waiting caller of that message. Once the cache entry has been resolved, it loops
through all senders and passed a clone of the value on. This has the downside
that it doesn't only require resources on the requesting side, but also in the
cache actor itself.

A better approach is to break this into a two-step process:
 1. The service creates a shared channel that will be populated once
 2. Recipients attach to the shared channel and wait for the response

The recipient therefore first has to await the receive end of the shared
channel, and then await the value from that channel. Of course, if the value is
already in the cache, this can be skipped, and the value can be sent directly
to the recipient.

# Implementation Details

This PR implements this with two nested `oneshot` channels. The outer channel
resolves an enum, which is either the result value, or a shared inner channel
that resolves the value.

We implement this as service response behavior, so that the building blocks for
this (senders, channels, ...) can be reused between multiple services. This can
be used for some of the messages, while other messages have regular async
behavior. The response future type is opaque, so users of such a service do not
notice the two-step behavior; they simply resolve the final value.

This PR updates the `RelayCacheService` to use this new response behavior.
Internally, the service still needs to hold two maps:
 - A map for the actual raw data that it pulls responses from. This data can be
   different or more elaborate than what's exposed via public messages.
 - A map for open channels. Entries in this map can be removed as soon as the
   cache value is resolved, because after that the service can short-circuit and
	 respond with the value directly.

# Alternatives

Tokio provides [`watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html)
channels that would be a suitable building block for this. However, since the
service currently controls when it considers cached values valid and when to
expose them, we cannot expose watch channels directly.

# References

Tracked by getsentry/relay#1602
Requires getsentry/relay#1622
#skip-changelog

### DIFF
--- a/relay-system/src/service.rs
+++ b/relay-system/src/service.rs
@@ -239,24 +239,31 @@ impl MessageResponse for NoResponse {
     }
 }
 
-// pub struct SharedRequest<T>(oneshot::Receiver<T>);
+// TODO(ja): Implement as concrete future type if possible
 pub struct SharedRequest<T>(Pin<Box<dyn Future<Output = Result<T, SendError>> + Send + Sync>>);
 
 #[derive(Debug)]
-pub struct SharedChannel<T>(oneshot::Sender<T>, Shared<oneshot::Receiver<T>>);
+pub struct SharedChannel<T> {
+    tx: oneshot::Sender<T>,
+    rx: Shared<oneshot::Receiver<T>>,
+}
 
 impl<T: Clone> SharedChannel<T> {
     pub fn new() -> Self {
         let (tx, rx) = oneshot::channel();
-        Self(tx, rx.shared())
+        Self {
+            tx,
+            rx: rx.shared(),
+        }
     }
 
     pub fn send(self, value: T) {
-        self.0.send(value).ok();
+        self.tx.send(value).ok();
     }
 
     pub fn attach(&self, sender: SharedSender<T>) {
-        sender.0.send(Foo::Shared(self.1.clone())).ok();
+        // TODO(ja): Keep this or the other one?
+        sender.0.send(Foo::Shared(self.rx.clone())).ok();
     }
 }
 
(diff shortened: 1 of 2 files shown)


relay-system/src/service.rs
```diff
@@ -237,6 +239,232 @@ impl MessageResponse for NoResponse {
...
+    /// # Example
+    ///
+    /// ```
+    /// todo!()
+    /// ```
+    pub fn attach(&mut self, sender: BroadcastSender<T>) {
+        sender.0.send(InitialResponse::Poll(self.rx.clone())).ok();
+    }
+
+    /// Send a value to all attached senders and close the channel.
```
olksdr:
```suggestion
    /// Sends a value to all attached senders and close the channel.
```

If we want to keep the docs consistent.

olksdr (approved):
lgtm!

relay-system/src/service.rs
```diff
@@ -237,6 +243,316 @@ impl MessageResponse for NoResponse {
...
+/// The request when sending an asynchronous message to a service.
+///
+/// This is returned from [`Addr::send`] when the message responds asynchronously through
+/// [`BroadcastResponse`]. It is a future that should be awaited. The message still runs to
+/// completion if this future is dropped.
+///
+/// # Panics
+///
+/// This future is not fused and panics if it is polled again after it has resolved.
+pub struct BroadcastRequest<T>(BroadcastState<T>)
```
jjbayer, line 275: For my understanding: Could we merge `BroadcastRequest` and `BroadcastState` into a single enum here?

jan-auer: We could, but then the internal states would become part of the public interface, and I wanted to prevent that.

PR #1622, opened 2022-11-23
ref(system): Introduce a generic recipient type

jan-auer:
`Recipient` is a type similar to `Addr`, but it is independent of interfaces. This makes it suitable for inversion of control, where a service would like to send messages to other services without knowing their interfaces.

This may not be the final implementation of the IoC pattern. This initial implementation only supports a single message and not entire interfaces (or extending interfaces). It is modeled after [`actix::Addr`](https://docs.rs/actix/0.13/actix/struct.Recipient.html). In the future, Recipient may be refactored to support entire interfaces. 

Tracked by https://github.com/getsentry/relay/issues/1602
#skip-changelog


olksdr (approved):
:shipit:

PR #1623, opened 2022-11-24
chore(toolchain): Fix lint and test for beta toolchain

olksdr:
Make sure we also prepare for the next Rust release. 

This PR fixes lint for beta toolchain, and make sure all the tests are still passing.

#skip-changelog

### DIFF
--- a/relay-general/src/types/meta.rs
+++ b/relay-general/src/types/meta.rs
@@ -474,7 +474,7 @@ impl Meta {
     }
 
     fn upsert(&mut self) -> &mut MetaInner {
-        self.0.get_or_insert_with(|| Box::<MetaInner>::default())
+        self.0.get_or_insert_with(|| Box::default())
     }
 
     /// The original length of this field, if applicable.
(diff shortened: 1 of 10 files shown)


jjbayer (approved):
Nice! Do we also want to run this in CI, as a non-required check?

olksdr:
> Nice! Do we also want to run this in CI, as a non-required check?

We do have this in CI https://github.com/getsentry/relay/actions/workflows/beta.yml and it runs periodically.  

Maybe adding this to our common CI runs will bring more benefits.  But also sometimes `beta` has something which `stable` does not support yet, and it is just not possible to fix it on both at the same time, so we will be having 1 red job.

relay-general/src/types/meta.rs
```diff
@@ -474,7 +474,7 @@ impl Meta {
     }
 
     fn upsert(&mut self) -> &mut MetaInner {
-        self.0.get_or_insert_with(|| Box::new(MetaInner::default()))
+        self.0.get_or_insert_with(|| Box::<MetaInner>::default())
```
jan-auer: This should work without type hints.

```suggestion
        self.0.get_or_insert_with(|| Box::default())
```

jan-auer (approved):

relay-general/src/types/meta.rs
olksdr: I haven't tried it, just used what clippy suggested. Will try it now and make a change.

PR #1635, opened 2022-11-24
ref(server): Migrate the project cache to Tokio

jan-auer:
Migrates the `ProjectCache` to a tokio-based Service. This includes changes to the inner workings of `Project` and `ProjectCache`:

- In addition to caching project states, the actor also dispatched requests for new states to the local-, Redis- and upstream sources. This message still exists, but the entire dispatch is now factored into the `ProjectSource` utility. Each lookup runs in its own task and does not hold (mutable) access of the cache. In a follow-up, the source utility can be further factored out of the project cache service.
- The `GetProjectState` created futures that resolves when the state has fetched. If projects cannot be resolved for longer periods of time, this accumulated many futures that slowed down the actor. The service now uses recently introduced `BroadcastChannels` and immediately end processing the incoming message. This means guaranteed responsiveness of the service independent of how many `GetProjectState` messages are sent in.
- Since the project fetch is now decoupled from the main service, there is a new internal channel to send resolved states back into the cache. The former message to trigger an update has been renamed to `RequestUpdate` for clarity.
- `FlushBuckets` was implemented directly in the project cache, while all other message handlers call through to a method on `Project`. The functionality is now moved to `Project`.
- The public `Project::get_or_fetch_state` function was used to probe for cached state, wait for loaded state, or trigger a prefetch for later access. These are now separate methods to make the intent more clear. Internally, they still rely on a common stub for simplicity.

Additionally, related utilities are updated:

- The metrics aggregator provided a hook to subscribe to flushed buckets. This was still implemented as `actix` recipient. Now that the project cache is migrated, it was possible and required to move the remaining bit of the aggregator to tokio as well. It works in the same way, just that this is now a `relay-system::Recipient`.
- To move metrics unit tests away from `actix`, the aggregator now uses `tokio::time::Instant`. This allows to [auto-advance](https://docs.rs/tokio/latest/tokio/time/fn.pause.html#auto-advance) time and use mocked time during tests.
- Similar to other services, the project cache subscribes to shutdown notifications before it accepts messages. This complicated tests, since the `Controller` internally still requires `actix`. For this reason, the subscribe mechanism was changed so that it always returns a `ShutdownHandle`, regardless of whether the controller runs or not.

Tracked by #1602 
Requires #1619 
Requires #1622 

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregation.rs
+++ b/relay-metrics/src/aggregation.rs
@@ -1,4 +1,3 @@
-use std::cmp::max;
 use std::collections::{btree_map, hash_map::Entry, BTreeMap, BTreeSet, HashMap};
 use std::fmt;
 use std::hash::Hasher;
@@ -1836,7 +1835,7 @@ impl AggregatorService {
             None => {
                 return BTreeMap::from([(None, buckets.into_iter().map(|x| x.bucket).collect())]);
             }
-            Some(x) => max(1, x), // handle 0,
+            Some(x) => x.max(1), // handle 0,
         };
         let mut partitions = BTreeMap::<_, Vec<Bucket>>::new();
         for bucket in buckets {
--- a/relay-metrics/src/aggregation.rs
+++ b/relay-metrics/src/aggregation.rs
@@ -1956,11 +1956,9 @@ impl AggregatorService {
         }
     }
 
-    fn handle_shutdown(&mut self, message: &Option<Shutdown>) {
-        if let Some(message) = message {
-            if message.timeout.is_some() {
-                self.state = AggregatorState::ShuttingDown;
-            }
+    fn handle_shutdown(&mut self, message: Shutdown) {
+        if message.timeout.is_some() {
+            self.state = AggregatorState::ShuttingDown;
         }
     }
 }
@@ -1981,7 +1979,7 @@ impl Service for AggregatorService {
     fn spawn_handler(mut self, mut rx: relay_system::Receiver<Self::Interface>) {
         tokio::spawn(async move {
             let mut ticker = tokio::time::interval(FLUSH_INTERVAL);
-            let mut shutdown = Controller::subscribe_v2().await;
+            let mut shutdown = Controller::shutdown_handle();
             relay_log::info!("aggregator started");
 
             // Note that currently this loop never exists and will run till the tokio runtime shuts
@@ -1992,7 +1990,7 @@ impl Service for AggregatorService {
 
                     _ = ticker.tick() => self.try_flush(),
                     Some(message) = rx.recv() => self.handle_message(message),
-                    _ = shutdown.changed() => self.handle_shutdown(&shutdown.borrow_and_update()),
+                    shutdown = shutdown.notified() => self.handle_shutdown(shutdown),
 
                     else => break,
                 }
--- a/relay-metrics/src/aggregation.rs
+++ b/relay-metrics/src/aggregation.rs
@@ -4,19 +4,18 @@ use std::fmt;
 use std::hash::Hasher;
 use std::iter::{FromIterator, FusedIterator};
 use std::mem;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
-use actix::{Message, Recipient};
-use failure::Fail;
 use float_ord::FloatOrd;
 use fnv::FnvHasher;
 use serde::{Deserialize, Serialize};
-use tokio::time;
+use thiserror::Error;
+use tokio::time::Instant;
 
 use relay_common::{MonotonicResult, ProjectKey, UnixTimestamp};
 use relay_log::LogError;
 use relay_system::{
-    compat, AsyncResponse, Controller, FromMessage, Interface, NoResponse, Sender, Service,
+    AsyncResponse, Controller, FromMessage, Interface, NoResponse, Recipient, Sender, Service,
     Shutdown,
 };
 
@@ -640,9 +639,9 @@ fn tags_cost(tags: &BTreeMap<String, String>) -> usize {
 }
 
 /// Error returned when parsing or serializing a [`Bucket`].
-#[derive(Debug, Fail)]
-#[fail(display = "failed to parse metric bucket")]
-pub struct ParseBucketError(#[cause] serde_json::Error);
+#[derive(Debug, Error)]
+#[error("failed to parse metric bucket")]
+pub struct ParseBucketError(#[source] serde_json::Error);
 
 /// An aggregation of metric values by the [`Aggregator`].
 ///
@@ -854,8 +853,8 @@ impl MetricsContainer for Bucket {
 }
 
 /// Any error that may occur during aggregation.
-#[derive(Debug, Fail, PartialEq)]
-#[fail(display = "failed to aggregate metrics: {}", kind)]
+#[derive(Debug, Error, PartialEq)]
+#[error("failed to aggregate metrics: {}", .kind)]
 pub struct AggregateMetricsError {
     kind: AggregateMetricsErrorKind,
 }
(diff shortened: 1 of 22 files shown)


relay-server/src/actors/project.rs
```diff
@@ -512,13 +513,18 @@ impl StateChannel {
         self
     }
 
-    pub fn receiver(&self) -> Shared<oneshot::Receiver<Arc<ProjectState>>> {
-        self.receiver.clone()
-    }
+    // pub fn receiver(&self) -> Shared<oneshot::Receiver<Arc<ProjectState>>> {
+    //     self.inner.clone()
+    // }
```
olksdr: Commented code to remove.

relay-server/src/actors/project.rs
```diff
@@ -512,13 +513,18 @@ impl StateChannel {
...
 
-    pub fn send(self, state: Arc<ProjectState>) {
-        self.sender.send(state).ok();
-    }
+    // pub fn send(self, state: Arc<ProjectState>) {
+    //     self.sender.send(state).ok();
+    // }
```
olksdr: same here

relay-metrics/src/aggregation.rs
```diff
@@ -640,9 +638,9 @@ fn tags_cost(tags: &BTreeMap<String, String>) -> usize {
 }
 
 /// Error returned when parsing or serializing a [`Bucket`].
-#[derive(Debug, Fail)]
-#[fail(display = "failed to parse metric bucket")]
-pub struct ParseBucketError(#[cause] serde_json::Error);
+#[derive(Debug, Error)]
```
jan-auer, line 641: Changes to errors are unrelated. I initially had some troubles with `Fail` and chose to keep the change in.

relay-metrics/src/aggregation.rs
```diff
@@ -1916,13 +1902,12 @@ impl AggregatorService {
             let partitioned_buckets = self.partition_buckets(project_buckets, num_partitions);
             for (partition_key, buckets) in partitioned_buckets {
                 self.process_batches(buckets, |batch| {
-                    if let Some(receiver) = self.receiver.clone() {
-                        let flush_buckets = FlushBuckets {
+                    if let Some(ref receiver) = self.receiver {
+                        receiver.send(FlushBuckets {
```
jan-auer, line 1906: This is the actual change to `MetricsAggregator`. It now uses a `relay_system::Recipient` instead of the `actix` one, since the Project Cache (only consumer) is now a `Service`.

relay-metrics/src/aggregation.rs
```diff
@@ -2763,77 +2754,76 @@ mod tests {
         assert!(iter.next().is_none());
     }
 
-    #[test]
-    fn test_flush_bucket() {
+    #[tokio::test]
+    async fn test_flush_bucket() {
```
jan-auer, line 2758: Thanks to the changed shutdown, we no longer need an actix runtime for these tests.

relay-metrics/src/aggregation.rs
```diff
@@ -1,22 +1,20 @@
...
+use std::time::Duration;
 
-use actix::{Message, Recipient};
-use failure::Fail;
 use float_ord::FloatOrd;
 use fnv::FnvHasher;
 use serde::{Deserialize, Serialize};
-use tokio::time;
+use thiserror::Error;
+use tokio::time::Instant;
```
jan-auer, line 12: Note that we replace `std::time::Instant` with `tokio::time::Instant`. In production, it is a simple wrapper, but this allows us to speed up tests with `tokio::time::pause()`.

relay-server/src/actors/health_check.rs
```diff
@@ -117,21 +117,20 @@ impl Service for HealthCheckService {
...
-            let mut shutdown_rx = Controller::subscribe_v2().await;
-
-            while shutdown_rx.changed().await.is_ok() {
-                if shutdown_rx.borrow_and_update().is_some() {
-                    service.is_shutting_down.store(true, Ordering::Relaxed);
+            let mut shutdown = Controller::shutdown_handle();
+
+            loop {
+                tokio::select! {
+                    biased;
```
jan-auer, line 125: Newer services all use a single loop with `select!`. Since I was updating the health check service already, I aligned the implementation. Please let me know what you think of this change: It makes the code slightly harder to read, but also aligns it with all other services.

Note that we no longer need an atomic bool for the flag anymore, but I didn't want to introduce even more changes.

olksdr (approved):
:shipit: 

Let's deploy and test it

relay-server/src/actors/project.rs
```diff
@@ -650,24 +648,27 @@ impl Project {
...
-    ///  - [`Response::Future`] if the state was expired or `no_cache` was specified. This future
-    ///    may fail if the state repeatedly cannot be fetched. The future does not have to be
-    ///    awaited for the update to pass.
-    pub fn get_or_fetch_state(
-        &mut self,
-        mut no_cache: bool,
-    ) -> Response<Arc<ProjectState>, ProjectError> {
+    /// If the state is already being updated in the background, this method checks if the request
+    /// needs to be upgraded with the `no_cache` flag to ensure a more recent update.
+    fn fetch_state(&mut self, no_cache: bool) -> &mut StateChannel {
```
jan-auer, line 655: The former `get_or_fetch_state` served multiple purposes and was thus split up into dedicated methods. 

- The new `get_or_fetch_state` runs the actual logic of whether the state should be refreshed and either returns the cached state or a handle to the broadcast channel.
- `fetch_state` now additionally contains deduplication of requests if there's already an active channel and `no_cache` flag.
- `get_cached_state` returns the state if it's cached and valid. As such, this is similar to previously calling `get_or_fetch_state` and matching `Response::Reply`.
- `get_state` waits for an updated cache, but uses the broadcast sender instead of returning a future.
- `prefetch` just triggers the fetch but returns no future. This method wouldn't have been necessary, strictly speaking, but it shows clear intent in what it's supposed to do. Before, we used to call `get_or_fetch_state` and then had to describe intent with comments.

relay-server/src/actors/project.rs
```diff
@@ -904,6 +925,63 @@ impl Project {
             rate_limits,
         })
     }
+
+    pub fn flush_buckets(&mut self, partition_key: Option<u64>, buckets: Vec<Bucket>) {
```
jan-auer, line 929: This was moved from the `ProjectCache` so that all project-related logic is contained here.

relay-server/src/actors/project_cache.rs
```diff
@@ -209,15 +47,15 @@ impl Message for FetchOptionalProjectState {
 /// Requests to the upstream are performed via `UpstreamProjectSource`, which internally batches
 /// individual requests.
 #[derive(Clone)]
-pub struct UpdateProjectState {
+pub struct RequestUpdate {
```
jan-auer, line 50: This name was changed for clarity. Note that there is now a new `UpdateProjectState` that performs the actual update in the cache.

relay-server/src/actors/project_cache.rs
```diff
@@ -542,121 +194,467 @@ impl UpdateRateLimits {
...
+    fn handle_check_envelope(
+        &mut self,
+        message: CheckEnvelope,
+    ) -> Result<CheckedEnvelope, DiscardReason> {
+        let project = self.get_or_create_project(message.envelope.meta().public_key());
+
+        // Preload the project cache so that it arrives a little earlier in processing. However,
+        // do not pass `no_cache`. In case the project is rate limited, we do not want to force
+        // a full reload. Fetching must not block the store request.
+        project.prefetch(false);
```
jan-auer, line 546: Since this always runs the prefetch, we could move this inside of `check_envelope`. However, there's also `validate_envelope` which enqueues this differently, so I propose to keep it as-is.

relay-server/src/endpoints/project_configs.rs
```diff
@@ -172,12 +176,9 @@ fn get_project_configs(
                         );
                     };
                 }
-                Ok(None) => {
+                None => {
                     pending.push(project_key);
                 }
-                Err(_) => {
-                    configs.insert(project_key, None);
```
jan-auer, line 179: This error has been eliminated completely from the interface. The project cache should populate with an errored state, and we should go through `Some` above.

relay-system/src/controller.rs
```diff
@@ -4,16 +4,45 @@ use std::time::Duration;
...
 
 use crate::compat;
 
+type ShutdownChannel = (
+    watch::Sender<Option<Shutdown>>,
+    watch::Receiver<Option<Shutdown>>,
+);
+
+/// Global [`ShutdownChannel`] for all services.
+static SHUTDOWN: OnceCell<ShutdownChannel> = OnceCell::new();
```
jan-auer, line 23: This refactor made it possible to test the metrics aggregator without an actix runtime since we don't actually require shutdown logic. There are two changes:

 1. The public interface is almost identical, but the handle for shutdown signals is now an opaque type instead of a `watch::Receiver` directly. We still need an object instead of just an async function to retain cancellation safety for the future.
 2. To avoid talking to the actor, the shutdown channel is now moved to a lazy global. The static `Controller::shutdown_handle()` function already was a singleton, so having this global is not strictly worse. In a future refactor, this should be wired up using `Recipient` instead.

relay-system/src/service.rs
```diff
@@ -654,7 +654,7 @@ pub trait FromMessage<M>: Interface {
 }
 
 /// Abstraction over address types for service channels.
-trait SendDispatch<M> {
+trait SendDispatch<M>: Send + Sync {
```
jan-auer, line 657: This was an oversight in the original PR that marked `Recipient: !Send` and `!Sync`.

PR #1637, opened 2022-11-24
fix(profiling): Parse string as number to handle a release bug

phacops:
Android SDK 6.8.0 went out sending a string in the `value` field. This is technically a bug but since the release is out, we'd like to accommodate it instead of rejecting profiles.


Zylphrex (approved):

jan-auer (approved):
Please consider leaving a comment about this in the code to distinguish this from cases where fields intentionally support both.

viglia (approved):

PR #1246, opened 2022-04-27
fix(server): Accept and forward unknown envelope items

jan-auer:
Relay has a policy to accept and forward unknown data on near every level, from individual fields in the event protocol to entire Envelope items. However, unknown envelope item types failed envelope parsing which lead to dropping the full Envelope with all its items.

This patch introduces a new `ItemType::Unknown` that contains the original, case sensitive, item type. Based on a new config flag `routing.accept_unknown_items`, this item is either retained and forwarded without modification, or is dropped immediately after parsing the envelope. The default is `false` for processing Relays (to drop unknown items), and `true` for all other configurations.

Unknown items are treated as event-independent. This means, they are split away from an event into a separate envelope. This is the most likely case for future event payloads. As with all other feature updates, customers and users are still recommended to update Relay when adopting new major features on their clients.

Alongside this change, empty envelopes no longer cause a `400 Bad Request` response. Most importantly, this allows clients to send envelopes containing all new items without receiving an error. As a side effect, this effectively makes sending empty envelopes legal in the ingestion API. To diagnose SDK bugs in the future, we can start to add internal diagnostics.

See also:
- https://github.com/getsentry/develop/pull/564
- https://github.com/getsentry/self-hosted/issues/1432

PR #1638, opened 2022-11-25
fix(server): Forward-compatible AttachmentType

jjbayer:
As we did in https://github.com/getsentry/relay/pull/1246 for `ItemType`, add an `Unknown(String)` variant to `AttachmentType`. When `accept_unknown_items` is set to false, items with this attachment type will be dropped.

This will allow outdated external Relays to forward new attachment types instead of dropping the entire envelope.

This defect was discovered in https://github.com/getsentry/rfcs/pull/33, which introduces a new attachment type.

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -1471,11 +1471,11 @@ impl EnvelopeProcessorService {
         let raw_security_item = envelope.take_item_by(|item| item.ty() == &ItemType::RawSecurity);
         let form_item = envelope.take_item_by(|item| item.ty() == &ItemType::FormData);
         let attachment_item = envelope
-            .take_item_by(|item| item.attachment_type() == Some(AttachmentType::EventPayload));
+            .take_item_by(|item| item.attachment_type() == Some(&AttachmentType::EventPayload));
         let breadcrumbs1 = envelope
-            .take_item_by(|item| item.attachment_type() == Some(AttachmentType::Breadcrumbs));
+            .take_item_by(|item| item.attachment_type() == Some(&AttachmentType::Breadcrumbs));
         let breadcrumbs2 = envelope
-            .take_item_by(|item| item.attachment_type() == Some(AttachmentType::Breadcrumbs));
+            .take_item_by(|item| item.attachment_type() == Some(&AttachmentType::Breadcrumbs));
 
         // Event items can never occur twice in an envelope.
         if let Some(duplicate) = envelope.get_item_by(|item| self.is_duplicate(item)) {
@@ -1550,9 +1550,9 @@ impl EnvelopeProcessorService {
         let envelope = &mut state.envelope;
 
         let minidump_attachment =
-            envelope.get_item_by(|item| item.attachment_type() == Some(AttachmentType::Minidump));
+            envelope.get_item_by(|item| item.attachment_type() == Some(&AttachmentType::Minidump));
         let apple_crash_report_attachment = envelope
-            .get_item_by(|item| item.attachment_type() == Some(AttachmentType::AppleCrashReport));
+            .get_item_by(|item| item.attachment_type() == Some(&AttachmentType::AppleCrashReport));
 
         if let Some(item) = minidump_attachment {
             let event = state.event.get_or_insert_with(Event::default);
@@ -1608,7 +1608,7 @@ impl EnvelopeProcessorService {
 
             let attachment_size = envelope
                 .items()
-                .filter(|item| item.attachment_type() == Some(AttachmentType::Attachment))
+                .filter(|item| item.attachment_type() == Some(&AttachmentType::Attachment))
                 .map(|item| item.len() as u64)
                 .sum::<u64>();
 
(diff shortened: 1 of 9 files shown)


relay-server/src/envelope.rs
```diff
@@ -365,8 +359,11 @@ pub enum AttachmentType {
     /// [`symbolic_unreal::Unreal4LogEntry`].
     ///
     /// [`symbolic_unreal::Unreal4LogEntry`]: https://docs.rs/symbolic/*/symbolic/unreal/struct.Unreal4LogEntry.html
-    #[serde(rename = "unreal.logs")]
     UnrealLogs,
+
+    /// Unknown attachment type, forwarded for compatibility.
+    /// Attachments with this type will be dropped if `accept_unknown_items` is set to false.
+    Unknown(String),
```
olksdr, line 366: just curious what the benefit to implement `FromStr` instead of just using [`#serde(other)`](https://serde.rs/variant-attrs.html#other) here?

jjbayer: `#serde(other)` does not work on non-unit enum variants, unfortunately. See https://github.com/serde-rs/serde/issues/1973.

relay-server/src/actors/processor.rs
```diff
@@ -1471,11 +1471,11 @@ impl EnvelopeProcessorService {
...
         let form_item = envelope.take_item_by(|item| item.ty() == &ItemType::FormData);
         let attachment_item = envelope
-            .take_item_by(|item| item.attachment_type() == Some(AttachmentType::EventPayload));
+            .take_item_by(|item| item.attachment_type() == Some(&AttachmentType::EventPayload));
         let breadcrumbs1 = envelope
-            .take_item_by(|item| item.attachment_type() == Some(AttachmentType::Breadcrumbs));
+            .take_item_by(|item| item.attachment_type() == Some(&AttachmentType::Breadcrumbs));
         let breadcrumbs2 = envelope
-            .take_item_by(|item| item.attachment_type() == Some(AttachmentType::Breadcrumbs));
+            .take_item_by(|item| item.attachment_type() == Some(&AttachmentType::Breadcrumbs));
```
jjbayer, line 1478: `AttachmentType` is not `Copy` anymore, because one of its variants now contains a `String`.

relay-server/src/envelope.rs
```diff
@@ -365,8 +359,11 @@ pub enum AttachmentType {
     /// [`symbolic_unreal::Unreal4LogEntry`].
     ///
     /// [`symbolic_unreal::Unreal4LogEntry`]: https://docs.rs/symbolic/*/symbolic/unreal/struct.Unreal4LogEntry.html
-    #[serde(rename = "unreal.logs")]
```
iker-barriocanal, line 368: Why are we removing these renames?

iker-barriocanal (commented):
Why are we adding an `Unknown(String)` variant instead of a specific variant for the view hierarchy? Having a fallback unknown variant seems ok to me, but I wonder if it's the best solution for that specific case we're now trying to solve.

olksdr (approved):

relay-server/src/envelope.rs
jan-auer: This is now manually implemented below.

relay-server/src/actors/store.rs
```diff
@@ -316,7 +316,10 @@ impl StoreService {
             content_type: item
                 .content_type()
                 .map(|content_type| content_type.as_str().to_owned()),
-            attachment_type: item.attachment_type().unwrap_or_default(),
+            attachment_type: item
+                .attachment_type()
+                .unwrap_or(&AttachmentType::default())
+                .clone(),
```
jan-auer:
```suggestion
                .cloned()
                .unwrap_or_default(),
```

relay-server/src/envelope.rs
```diff
@@ -375,6 +372,43 @@ impl Default for AttachmentType {
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(match s {
+            "event.attachment" => AttachmentType::Attachment,
+            "event.minidump" => AttachmentType::Minidump,
+            "event.applecrashreport" => AttachmentType::AppleCrashReport,
+            "event.payload" => AttachmentType::EventPayload,
+            "event.breadcrumbs" => AttachmentType::Breadcrumbs,
+            "unreal.context" => AttachmentType::UnrealContext,
+            "unreal.logs" => AttachmentType::UnrealLogs,
+            other => AttachmentType::Unknown(other.to_owned()), // FIXME
```
jan-auer: What's the FIXME here?

relay-server/src/envelope.rs
```diff
@@ -607,15 +641,21 @@ impl Item {
...
-                | AttachmentType::Minidump
-                | AttachmentType::EventPayload
-                | AttachmentType::Breadcrumbs => true,
-                AttachmentType::Attachment
-                | AttachmentType::UnrealContext
-                | AttachmentType::UnrealLogs => false,
-            },
+            ItemType::Attachment => {
+                match self.attachment_type().unwrap_or(&AttachmentType::default()) {
+                    &AttachmentType::AppleCrashReport
```
jan-auer: I believe with auto-borrowing, it should be possible to match without the `&` in patterns:

```suggestion
                    AttachmentType::AppleCrashReport
```

jan-auer (changes requested):

jjbayer:
> Why are we adding an `Unknown(String)` variant instead of a specific variant for the view hierarchy?

@iker-barriocanal we will add a specific variant for the view hierarchy in a different PR, so this PR actually won't help with the rollout of that new type. It only ensures that when we add attachment types in the future, envelopes will not be dropped by external Relays.

jan-auer (approved):

Issue #1597, opened 2022-11-18
Add hardcoded transaction name parsing rules to relay


HazAT:
The goal is to ship something from which we can already get value and learn from it. The rule spec will probably look different than this initial iterations; that’s expected and they will be adapted to match the spec.

**Details**

* Replace UUIDs, git SHAs and numerical IDs in transaction names by placeholders. 

**Acceptance Criteria**

* The `transaction` tag on transaction metrics is transformed.
* The `transaction` field of the Dynamic Sampling Context (and its FieldValueProvider) is transformed.
* The output of the FieldValueProvider for events is transformed.
* The `transaction` field of the event payload is transformed, but the original transaction name is also written to the event (for example in _meta).
  * The original transaction name does not have to be searchable (yet)
  * Optional: Mark in the string what section was replaced and put that information in `_meta`.
* Feature flag and ship for sentry

PRs:
* https://github.com/getsentry/relay/pull/1621
* https://github.com/getsentry/sentry/pull/41642

jjbayer:
Initial proposal for detecting identifiers (internal docs): https://www.notion.so/sentry/Prevent-high-cardinality-for-transaction-names-09fc121e1a5743f980a8ccae1b543e6c#e895b742080944d28762f985a77b5803

Regex we use to detect identifiers in issue grouping: https://github.com/getsentry/sentry/blob/6ba59023a78bfe033e48ea4e035b64710a905c6b/src/sentry/grouping/strategies/message.py#L16

jjbayer:
This will not provide value in the product yet, because erasing identifiers does not give us enough confidence to declare the transaction name "low cardinality", and with that, it will still be changed to "unparameterized" in metrics extraction.

olksdr:
Changes deployed and enabled for sentry. We could watch the changes to transaction names as the result of linked PRs.

PR #1640, opened 2022-11-25
chore(transaction): Improve regex to group URL segments

olksdr:
follow-up for #1597 

Group the entire segment in the pattern is found inside it. 
E.g. for url path which looks like `/user/public-19-profile/edit` we  want make sure to fold second segment and get the following result ~> `/user/*/edit`

#skip-changelog

### DIFF
--- a/relay-general/src/store/transactions.rs
+++ b/relay-general/src/store/transactions.rs
@@ -1439,6 +1439,7 @@ mod tests {
                 "#,
                     $input
                 );
+
                 let mut event = Annotated::<Event>::from_json(&json).unwrap();
 
                 process_value(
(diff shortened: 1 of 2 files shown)


iker-barriocanal (approved):

olksdr:
@jjbayer added the additional check in [b7f8c17](https://github.com/getsentry/relay/pull/1640/commits/b7f8c1780d92f542575083a54fe82d1fd0f23e42), to avoid replacing the entire string. PTAL!

relay-general/src/store/transactions.rs
```diff
@@ -1411,4 +1414,78 @@ mod tests {
...
+        test_transaction_name_normalize_hex,
+        "/u/0x3707344A4093822299F31D008/profile/123123213",
+        "/u/*/profile/*"
+    );
+    transaction_name_test!(
+        test_transaction_name_normalize_windows_path,
+        r#"C:\\\\Program Files\\1234\\Files"#,
+        r#"C:\\Program Files\*\Files"#
+    );
+    transaction_name_test!(test_transaction_name_skip_replace_all, "12345", "12345");
```
jjbayer, line 1490:
```suggestion
    transaction_name_test!(test_transaction_name_skip_replace_all, "12345", "12345");
    transaction_name_test!(test_transaction_name_skip_replace_all2, "open-12345-close", "open-12345-close");
```

jjbayer (approved):

jjbayer (approved):

PR #1641, opened 2022-11-28
fix(sampling): Forward compatibility for config

jjbayer:
Add missing variants to enums in sampling config for forward compatibility.

#skip-changelog

### DIFF
--- a/relay-sampling/src/lib.rs
+++ b/relay-sampling/src/lib.rs
@@ -33,6 +33,10 @@ pub enum RuleType {
     /// A non transaction rule applies to Errors, Security events...every type of event that
     /// is not a Transaction
     Error,
+
+    /// If the sampling config contains new rule types, do not sample at all.
+    #[serde(other)]
+    Unsupported,
 }
 
 /// A condition that checks the values using the equality operator.
@@ -372,7 +376,7 @@ pub struct SamplingRule {
 
 impl SamplingRule {
     fn supported(&self) -> bool {
-        self.condition.supported()
+        self.condition.supported() && self.ty != RuleType::Unsupported
     }
 
     /// Returns whether the sampling rule is active.
@@ -688,6 +692,10 @@ pub enum SamplingMode {
     /// In this mode, the server sampling rate is capped by the client's sampling rate. Rules with a
     /// higher sample rate than what the client is sending are effectively inactive.
     Total,
+
+    /// Catch-all variant for forward compatibility.
+    #[serde(other)]
+    Unsupported,
 }
 
 impl Default for SamplingMode {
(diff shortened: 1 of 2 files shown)


relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -438,6 +444,20 @@ mod tests {
         assert_eq!(spec.unwrap().unwrap().sample_rate, 0.2);
     }
 
+    #[test]
+    fn test_trace_rule_unsupported() {
+        let project_state = state_with_rule(Some(0.1), RuleType::Trace, SamplingMode::Unsupported);
+        let sampling_context = create_sampling_context(Some(0.5));
+        let spec = get_trace_sampling_rule(
+            true, // irrelevant, just skips unsupported rules
```
```suggestion
            true,
```

olksdr (approved):

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -72,6 +72,12 @@ fn get_trace_sampling_rule(
     let sample_rate = match sampling_config.mode {
         SamplingMode::Received => rule.sample_rate,
         SamplingMode::Total => sampling_context.adjusted_sample_rate(rule.sample_rate),
+        SamplingMode::Unsupported => {
+            if processing_enabled {
+                relay_log::error!("Found unsupported sampling mode even as processing Relay, keep");
+            }
+            return Err(SamplingResult::Keep);
```
iker-barriocanal, line 79: I'm not sure if I like this, but I don't have better suggestions.

iker-barriocanal (approved):

relay-sampling/src/lib.rs
```diff
@@ -372,7 +376,7 @@ pub struct SamplingRule {
 
 impl SamplingRule {
     fn supported(&self) -> bool {
-        self.condition.supported()
+        self.condition.supported() && self.ty != RuleType::Unsupported
```
jan-auer, line 379: _nit_: Introduce a `supported` method on rules types, too?

relay-server/src/utils/dynamic_sampling.rs
jan-auer: We could alternatively log an error for during project state deserialization (since that's where the error really happens) and fail silently here instead. 

This behavior is fine since it's the same as if sampling were not implemented. That matches behavior in other cases, where Relay ignores all configuration it doesn't know about.

jan-auer (approved):

relay-server/src/utils/dynamic_sampling.rs
iker-barriocanal: > We could alternatively log an error for during project state deserialization (since that's where the error really happens)

I'd rather log the error during deserialization.

jjbayer: I tried to be consistent with `check_unsupported_rules` here, but maybe it's a bad pattern anyway.

https://github.com/getsentry/relay/blob/ca275954015fb23d5f9c96bc55f3db29435f660a/relay-server/src/utils/dynamic_sampling.rs#L33-L47

jjbayer: As discussed offline, going to merge as-is.

PR #1642, opened 2022-11-29
feat(server): Add support for ViewHierarchy attachment_type

markushi:
Add support for the upcoming [View Hierarchy](https://github.com/getsentry/rfcs/pull/33) feature.

If there's more work that needs to be done for relay, please let me know or kindly take over the PR 🙂 

I mapped the new attachment type to the String `"event.viewhierarchy"`, which matches with the existing `"event.applecrashreport"` - but I guess technically `"event.view_hierarchy"` would be more correct here.

### DIFF
--- a/relay-server/src/envelope.rs
+++ b/relay-server/src/envelope.rs
@@ -361,6 +361,9 @@ pub enum AttachmentType {
     /// [`symbolic_unreal::Unreal4LogEntry`]: https://docs.rs/symbolic/*/symbolic/unreal/struct.Unreal4LogEntry.html
     UnrealLogs,
 
+    /// An application ui view hierarchy (json payload)
+    ViewHierarchy,
+
     /// Unknown attachment type, forwarded for compatibility.
     /// Attachments with this type will be dropped if `accept_unknown_items` is set to false.
     Unknown(String),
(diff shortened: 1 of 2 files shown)


relay-server/src/envelope.rs
```diff
@@ -361,6 +361,9 @@ pub enum AttachmentType {
     /// [`symbolic_unreal::Unreal4LogEntry`]: https://docs.rs/symbolic/*/symbolic/unreal/struct.Unreal4LogEntry.html
     UnrealLogs,
 
+    /// An application ui view hierarchy (json payload)
```
iker-barriocanal:
```suggestion
    /// An application UI view hierarchy (json payload).
```

relay-server/src/envelope.rs
```diff
@@ -382,6 +385,7 @@ impl fmt::Display for AttachmentType {
             AttachmentType::Breadcrumbs => write!(f, "event.breadcrumbs"),
             AttachmentType::UnrealContext => write!(f, "unreal.context"),
             AttachmentType::UnrealLogs => write!(f, "unreal.logs"),
+            AttachmentType::ViewHierarchy => write!(f, "event.viewhierarchy"),
```
iker-barriocanal: For the name, I'm leaning towards `"event.view_hierarchy"` but not strongly opinionated for the same reason you describe in the PR description. I'd match it with the type coming in the envelope though.

iker-barriocanal (approved):

jjbayer (approved):

relay-server/src/envelope.rs
jjbayer: This was already discussed here: https://github.com/getsentry/rfcs/pull/33#discussion_r1032075538. But there seems to be a strong preference for `view_hierarchy` so go for it.

markushi: Ah I knew it was mentioned somewhere, but couldn't find it - thanks for linking!

PR #1643, opened 2022-11-29
chore(kafka): switch to another error crate: thiserror

olksdr:
The crate we use for error handling `failure` is deprecated and outdated. `thiserror` is a good, near drop-in replacement for `#[derive(Fail)]`.

#skip-changelog

### DIFF
--- a/relay-kafka/src/lib.rs
+++ b/relay-kafka/src/lib.rs
@@ -16,7 +16,7 @@
 //!     builder = builder.add_kafka_topic_config(KafkaTopic::Events, &kafka_config_events).unwrap();
 //!
 //!     // And add potentially another topic config.
-//!     builder.add_kafka_topic_config(&kafka_config_metrics)?;
+//!     builder = builder.add_kafka_topic_config(KafkaTopic::Transactions, &kafka_config_metrics).unwrap();
 //!
 //!     // build the client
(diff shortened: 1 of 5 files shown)


relay-kafka/src/lib.rs
```diff
@@ -16,7 +16,7 @@
 //!     builder = builder.add_kafka_topic_config(KafkaTopic::Events, &kafka_config_events).unwrap();
 //!
 //!     // And add potentially another topic config.
-//!     builder.add_kafka_topic_config(&kafka_config_metrics)?;
+//!     builder = builder.add_kafka_topic_config(KafkaTopic::Transactions, &kafka_config_metrics).unwrap();
```
iker-barriocanal, line 19: Is this change related to the PR?

iker-barriocanal (approved):
`thiserror` has the same [licensing](https://github.com/dtolnay/thiserror) (MIT or Apache) as [`serde_json`](https://github.com/serde-rs/json), so there shouldn't be any issues regarding that.

relay-kafka/src/lib.rs
olksdr: This is just to make the doc consistent, and use the proper params to to show the use-case.

jan-auer (approved):
:shipit:

PR #1644, opened 2022-11-29
fix(server): Use compat::send for actix messages

jan-auer:
Fix a panic in the newly converted project cache by using `compat::send` when dispatching messages from the tokio runtime to legacy actors.

Follow-up to #1635
Relates to #1602
Fixes [RELAY-CRM](https://sentry.my.sentry.io/organizations/sentry/issues/316000/)

#skip-changelog


olksdr (approved):
:shipit:

PR #1651, opened 2022-11-29
ref: Migrate main crates away from failure

jan-auer:
Unfortunately, this became a rather large PR since many of the error types go across error boundaries. With this PR, all crates other than the FFI/C-ABI are fully converted to `Error`. 

For application-level errors in `relay-server`, `relay-config`, and `relay`, we use `anyhow`. There is a single case where down-casting the contained config error is required.

#skip-changelog


relay/src/cli.rs
```diff
@@ -34,13 +32,14 @@ pub fn execute() -> Result<(), Error> {
...
-        }
-        Err(e) => return Err(e.into()),
+        // TODO(ja): Bring this back
+        // Err(e)
+        //     if matches.value_of("config").is_none()
+        //         && e.kind() == ConfigErrorKind::CouldNotOpenFile =>
+        // {
+        //     Config::default()
+        // }
+        Err(e) => return Err(e),
```
olksdr: So this is how I did it:
```rust
 Err(e) => match e.root_cause().downcast_ref::<ConfigError>() {
     Some(ce)
         if matches.value_of("config").is_none()
             && ce.kind() == ConfigErrorKind::CouldNotOpenFile =>
     {
         Config::default()
     }
     _ => bail!(e),
 },

```

which theoretically should do what it did before, since we know that we are expecting `ConfigError` here.

relay-log/src/utils.rs
```diff
@@ -49,22 +48,19 @@ pub fn ensure_error<E: failure::AsFail>(error: &E) {
...
+impl<'a, E: Error + ?Sized> fmt::Display for LogError<'a, E> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let fail = self.0.as_fail();
+        write!(f, "{}", self.0)?;
 
-        write!(f, "{}", fail)?;
-        for cause in fail.iter_causes() {
-            write!(f, "\n  caused by: {}", cause)?;
+        let mut source = self.0.source();
+        while let Some(s) = source {
```
olksdr, line 50: heh, I was thinking about exactly same implementation 👍

olksdr (approved):
Few TODOs to cleanup but overall 👍

relay/src/cli.rs
jan-auer: That's maybe even better than what I ended up doing. I opted to move it into a separate function and then return from the innermost branch. 

Note that we can't use `root_cause()` here since that would be some `io` error or something out of our control. Instead, we want to match on the top-most error.

relay-log/src/sentry_failure.rs
```diff
@@ -1,55 +0,0 @@
-use failure::Fail;
-use sentry::integrations::backtrace::parse_stacktrace;
-use sentry::parse_type_from_debug;
-use sentry::protocol::Exception;
-use sentry::{ClientOptions, Integration};
-
-/// The Sentry Failure Integration.
-#[derive(Debug, Default)]
-pub struct FailureIntegration;
```
jan-auer, line 9: The failure integration was removed entirely since we have backtraces disabled. If desired, we can leave it in the codebase, though.

relay-log/src/sentry_failure.rs
```diff
@@ -1,55 +0,0 @@
...
-}
-
-/// This converts a single fail instance into an exception.
-///
-/// This is typically not very useful as the `event_from_error` and
-/// `event_from_fail` methods will assemble an entire event with all the
-/// causes of a failure, however for certain more complex situations where
-/// fails are contained within a non fail error type that might also carry
-/// useful information it can be useful to call this method instead.
-pub fn exception_from_single_fail<F: Fail + ?Sized>(
```
jan-auer, line 41: This was moved directly into the vendored actix web integration into `relay-server` to remove the dependency on `failure` here.

relay-server/src/actors/outcome.rs
```diff
@@ -554,7 +554,7 @@ impl HttpOutcomeProducer {
             match compat::send(UpstreamRelay::from_registry(), SendQuery(request)).await {
                 Ok(_) => relay_log::trace!("outcome batch sent."),
                 Err(error) => {
-                    relay_log::error!("outcome batch sending failed with: {}", LogError(&error))
+                    relay_log::error!("outcome batch sending failed with: {}", error)
```
jan-auer, line 557: This `LogError` was removed since the underlying `MailboxError` doesn't implement `Error` and it also doesn't contain nested sources. This can be reverted when converting `UpstreamRelay` to a service.

relay-server/src/endpoints/common.rs
```diff
@@ -30,59 +29,65 @@ use crate::utils::{
...
 
-    #[fail(display = "failed to queue envelope")]
-    QueueFailed(#[cause] BufferError),
+    #[error("failed to queue envelope")]
+    QueueFailed(#[from] BufferError),
 
-    #[fail(display = "failed to read request body")]
-    PayloadError(#[cause] PayloadError),
+    #[error("failed to read request body")]
+    PayloadError(#[source] failure::Compat<PayloadError>),
```
jan-auer, line 71: `failure::Compat` is required until we upgrade or replace `actix-web`.

relay-server/src/service.rs
```diff
@@ -32,85 +32,41 @@ use crate::{endpoints, utils};
...
-#[derive(Debug)]
-pub struct ServerError {
-    inner: Context<ServerErrorKind>,
-}
-
 /// Indicates the type of failure of the server.
-#[derive(Debug, Fail, Copy, Clone, PartialEq, Eq, Hash)]
-pub enum ServerErrorKind {
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, thiserror::Error)]
+pub enum ServerError {
```
jan-auer, line 37: The `ServerError` type is now replaced by an opaque `anyhow::Error`, and the former `kind` just there to create the context message. I chose to rename it for brevity.

An alternative to this design is that we inline the error messages directly like so:

```rust
  bind(4711).context("bind to interface failed")?
```

relay-server/src/utils/api.rs
```diff
@@ -77,3 +76,5 @@ impl fmt::Display for ApiErrorResponse {
         }
     }
 }
+
+impl Error for ApiErrorResponse {}
```
jan-auer, line 80: This cannot have a source because the raw `String` is not an error.

relay-server/src/utils/error_boundary.rs
```diff
@@ -1,45 +1,12 @@
...
-    fn as_cloned_fail(&self) -> ClonedFail {
-        ClonedFail(self.to_string())
-    }
-}
-
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum ErrorBoundary<T> {
-    Err(Box<dyn Fail>),
+    Err(Arc<dyn Error + Send + Sync + 'static>),
```
jan-auer, line 9: Turns out we can clone the error if we put the error behind an `Arc`. Since the error boundary is used in some types that are sent between services or used in async functions, I additionally constrained to `Send + Sync`.

PR #1653, opened 2022-11-29
feat(replays): Add non-chunked recording message type

cmanallen:
Optimization to allow greater parallelization in our consumer.  A single message containing the payload does not require ordering guarantees and can be processed in a process pool or thread pool.


jjbayer:
> Where should test coverage for this exist?

@cmanallen there is a failing integration test, that seems to be a good spot to add test coverage.

jjbayer:
@cmanallen could you add some motivation for this change to the PR description? Is it an optimization to reduce the number of messages?

cmanallen:
@jjbayer Updated.

relay-server/src/actors/store.rs
```diff
@@ -914,8 +974,9 @@ enum KafkaMessage {
     Metric(MetricKafkaMessage),
     Profile(ProfileKafkaMessage),
     ReplayEvent(ReplayEventKafkaMessage),
-    ReplayRecording(ReplayRecordingKafkaMessage),
-    ReplayRecordingChunk(ReplayRecordingChunkKafkaMessage),
+    ReplayRecordingNotChunked(ReplayRecordingKafkaMessage),
+    ReplayRecording(ChunkedReplayRecordingKafkaMessage),
```
jjbayer: Personally I would flip the variant names to match the struct names, e.g. `ReplayRecording(...)`, `ChunkedReplayRecording(...)`.

relay-server/src/actors/store.rs
```diff
@@ -846,6 +895,17 @@ struct ReplayRecordingKafkaMessage {
...
 
+#[derive(Debug, Serialize)]
+struct ReplayRecordingKafkaMessage {
+    replay_id: EventId,
+    key_id: Option<u64>,
+    org_id: u64,
+    project_id: ProjectId,
+    received: u64,
+    retention_days: u16,
+    payload: Bytes,
```
jjbayer, line 908: It might simplify the code if we only defined a single type `ReplayRecordingKafkaMessage`, with

```rust
payload: Option<Bytes>,

replay_recording: Option<ChunkedReplayRecording>,
```

I realize this would in theory allow the creation of a message with both fields set, but it would allow you to simplify `produce_replay_recording` to something like

```rust
let mut message = ...;
if item.payload().len() < max_payload_size {
    message.payload = ...;
} else {
    message.replay_recording = ...;
}
metric!(...);
self.produce(...);
```

relay-server/src/actors/store.rs
cmanallen: I initially had the naming reversed.  Unfortunately, it translates the types to snake case and the types are flipped on the consumer because of legacy reasons.

relay-server/src/actors/store.rs
cmanallen: I can see the value in that.  But I prefer to keep impossible states impossible.  I don't want to have to debug weird states in my messages.

relay-server/src/actors/store.rs
cmanallen: Let me think about a migration strategy.  Could be worth it.

cmanallen: I misread this earlier I think.  I've updated the naming to be more consistent and I don't think I will migrate the message types.

relay-server/src/actors/store.rs
jjbayer: I get that. It would be nice to use an enum here to represent the two possible states, but I would not know how to make that work with serde if the field is supposed to have a different name in each case.

jjbayer: Maybe we can at least declare the fields of the two structs in the same order, so it's easy to spot that they are largely the same (e.g. `project_id` is at different positions).

jjbayer (approved):

PR #1658, opened 2022-12-01
build(docker): Build an image for sentry CI

asottile-sentry:
we're moving sentry's CI from depending on both gcr.io and ghcr.io to only ghcr.io

#skip-changelog


Dockerfile
```diff
@@ -1,23 +1,14 @@
-ARG DOCKER_ARCH=amd64
-
```
asottile-sentry, line 2: I deleted all the `ARCH` stuff here to make the dockerfile easier to build -- it now depend on the `--platform` argument to `docker build` instead of specifying the architecture as an `--arg`

.github/workflows/image.yml
```diff
@@ -0,0 +1,43 @@
...
+      matrix:
+        # the arm64 build takes too long, so disable for now
+        arch: [amd64]
+    runs-on: ubuntu-latest
+    env:
+      IMG_CACHE: ghcr.io/getsentry/relay:${{ matrix.arch }}-latest
+      IMG_VERSIONED: ghcr.io/getsentry/relay:${{ matrix.arch }}-${{ github.sha }}
+    steps:
+    - uses: actions/checkout@v3
+    - run: docker run --rm --privileged tonistiigi/binfmt --install arm64
```
mattgauntseo-sentry: Optional: Should this be removed for now as this is just amd64?

.github/workflows/image.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: build
+      run: |
+        set -euxo pipefail
+        args=()
+        if docker pull -q "$IMG_CACHE"; then
+          args+=(--cache-from "$IMG_CACHE")
+        fi
+        docker buildx build \
+            "${args[@]}" \
+            --build-arg BUILDKIT_INLINE_CACHE=1 \
```
mattgauntseo-sentry: Is this argument needed in relay?

mattgauntseo-sentry (approved):

.github/workflows/image.yml
asottile-sentry: this is so you get image caching

.github/workflows/image.yml
asottile-sentry: I figure it's simpler to leave it, and if we want arm64 we can just adjust the matrix

iker-barriocanal (dismissed):

olksdr (commented):
This looks good. 
@asottile-sentry could you, please, merge `master` back and resolve the conflicts, so we could give another look and finally merge it in.

Dockerfile
```diff
@@ -33,7 +25,8 @@ ENV RUSTUP_HOME=/usr/local/rustup \
     PATH=/usr/local/cargo/bin:$PATH
 
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
-    | sh -s -- -y --profile minimal --default-toolchain=${RUST_TOOLCHAIN_VERSION}
+    | sh -s -- -y --profile minimal --default-toolchain=${RUST_TOOLCHAIN_VERSION} \
+    && echo -e "[net]\ngit-fetch-with-cli = true" > $CARGO_HOME/config
```
iker-barriocanal, line 28: TIL `git-fetch-with-cli`

iker-barriocanal (approved):

scripts/docker-build-linux.sh
```diff
@@ -15,11 +15,11 @@ BUILD_IMAGE="us.gcr.io/sentryio/relay:deps"
```
jan-auer, line 15: We should change the `BUILD_IMAGE`:
 - Use the GHCR or DockerHub instead of GCR
 - Use caches, but also build that image as part of this script

Makefile
```diff
@@ -30,6 +30,17 @@ build-linux-release: setup-git ## build linux release of the relay
 	objcopy --add-gnu-debuglink target/${TARGET}/release/relay{.debug,}
 .PHONY: build-linux-release
 
+collect-source-bundle: setup-git ## copy the built relay binary to current folder and collects debug bundles
```
jan-auer, line 33: Let's move this to a script. This allows us to:

 - Call this more naturally from other scripts
 - Use it in gocd deployments
 - Publish it to GH releases

.github/workflows/ci.yml
```diff
@@ -13,6 +13,130 @@ env:
...
+        with:
+          key: ${{ github.job }}-${{ matrix.arch }}
+
+      - run: docker run --rm --privileged tonistiigi/binfmt --install arm64
+        if: matrix.arch == 'arm64'
+
+      - name: ghcr.io login
+        run: docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
+
+      - name: build-deps
```
jan-auer: Idea: These steps could be moved to a script that we can also use locally to:

- Build / update the builder image
- Build the binary to a local folder (should we introduce a `build/` directory?)
- Create the release docker container w/ the binary inside

In a close follow-up, we can then replace `docker-build-linux.sh` in the `build_binary` workflow with our new script.

jan-auer:
After merging, we need to 

 1. Disable the GCB repository trigger
 2. Update required status checks in this repo
 3. Update freight checks by removing cloudbuild and adding required jobs:

```json
"checks": [
  {
    "type": "github-apps",
    "config": {
      "repo": "getsentry/relay",
      "contexts": [
        "Integration Tests",
        "Test (macos-latest)",
        "Test (windows-latest)",
        "Test All Features (ubuntu-latest)",
        "Push GCR Docker Image"
      ]
    }
  }
]
```

scripts/build-docker-image.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Set the correct build target and update the arch if required.
+if [[ "$ARCH" = "amd64" ]]; then
+  BUILD_TARGET="x86_64-unknown-linux-gnu"
+elif [[ "$ARCH" = "arm64" ]]; then
+  BUILD_TARGET="aarch64-unknown-linux-gnu"
+elif [[ "$ARCH" = "aarch64" ]]; then
+  BUILD_TARGET="aarch64-unknown-linux-gnu"
+  ARCH="arm64"
+fi
```
jan-auer: Let's add an `else` branch that errors if no supported architecture is set, for instance:

```suggestion
if [[ "$ARCH" == "aarch64 ]]; then
  ARCH="arm64"
fi

if [[ "$ARCH" = "amd64" ]]; then
  BUILD_TARGET="x86_64-unknown-linux-gnu"
elif [[ "$ARCH" = "arm64" ]]; then
  BUILD_TARGET="aarch64-unknown-linux-gnu"
else
  echo "ERROR Unsupported architecture"
  exit 1
fi
```

jan-auer (approved):

.github/workflows/ci.yml
```diff
@@ -225,6 +225,100 @@ jobs:
...
+      
+    # Skip redundant checks for library releases
+    if: "!startsWith(github.ref, 'refs/heads/release-library/')"
+
+    env:
+      IMG_BASE: ghcr.io/getsentry/relay
+      IMG_DEPS: ghcr.io/getsentry/relay-deps:${{ matrix.arch }}
+      # GITHUB_SHA in pull requests points to the merge commit
+      IMG_VERSIONED: ghcr.io/getsentry/relay:${{ github.event.pull_request.head.sha || github.sha }}
+      ARCH: ${{ matrix.arch }}
```
asottile-sentry, line 246: at a glance this looks unused, but it's actually used inside `build-docker-image` -- I would probably make it an argument to the script instead

.github/workflows/ci.yml
```diff
@@ -225,6 +225,100 @@ jobs:
...
+      - run: docker run --rm --privileged tonistiigi/binfmt --install arm64
+        if: matrix.arch == 'arm64'
+
+      - name: build
+        run: ./scripts/build-docker-image.sh
+
+      - name: Push to ghcr.io
+        run: |
+          set -euxo pipefail
+          docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
```
asottile-sentry, line 266: this will fail for external contributors -- you'll probably want to gate this somehow on that

Dockerfile.builder
```diff
@@ -0,0 +1,29 @@
...
+
+RUN yum -y update \
+    && yum -y install centos-release-scl epel-release \
+    # install a modern compiler toolchain
+    && yum -y install cmake3 devtoolset-10 git \
+    # below required for sentry-native
+    llvm-toolset-7.0-clang-devel \
+    && yum clean all \
+    && rm -rf /var/cache/yum \
+    && ln -s /usr/bin/cmake3 /usr/bin/cmake
```
asottile-sentry, line 16: typically `/usr/bin` is managed by the package manager, this should be a symlink to `/usr/local/bin/cmake` probably

Dockerfile.release
```diff
@@ -0,0 +1,29 @@
...
+ENV \
+    RELAY_UID=10001 \
+    RELAY_GID=10001
+
+# Create a new user and group with fixed uid/gid
+RUN groupadd --system relay --gid $RELAY_GID \
+    && useradd --system --gid relay --uid $RELAY_UID relay
+
+RUN mkdir /work /etc/relay \
+    && chown relay:relay /work /etc/relay
```
asottile-sentry, line 17: not necessary for this PR, but this is much easier to do with `USER` (and then not needing `gosu` at all)

scripts/build-docker-image.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Images to use and build.
+IMG_DEPS=${IMG_DEPS:-"ghcr.io/getsentry/relay-deps:$ARCH"}
+IMG_VERSIONED=${IMG_VERSIONED:-"relay:latest"}
+
+# Build a builder image with all the depdendencies.
+args=(--progress auto)
```
asottile-sentry, line 27: shouldn't be necessary?

scripts/build-docker-image.sh
```diff
@@ -0,0 +1,55 @@
...
+    --cache-to type=inline \
+    --platform "linux/$ARCH" \
+    --tag "$IMG_DEPS" \
+    --target relay-deps \
+    --file Dockerfile.builder \
+    .
+
+# Build the binary inside of the builder image.
+docker run \
+    -v "$(pwd):/work" \
```
asottile-sentry: `$(pwd)` => `$PWD` (var lookup is free compared to a subshell and subprocess)

I also like to be explicit and use `--volume` and state whether it's `:ro` or `:rw` (I don't even remember the default)

scripts/build-docker-image.sh
```diff
@@ -0,0 +1,55 @@
...
+    --platform "linux/$ARCH" \
+    -e TARGET="$BUILD_TARGET" \
+    "$IMG_DEPS" \
+    scl enable devtoolset-10 llvm-toolset-7.0 -- make build-release-with-bundles RELAY_FEATURES="ssl,processing,crash-handler"
+
+# Fix permissions for shared directories
+USER_ID=$(id -u)
+GROUP_ID=$(id -g)
+sudo chown -R "${USER_ID}:${GROUP_ID}" target/
```
asottile-sentry: better: run the container with `--user` and then you don't need to do this.  or utilize `docker cp`

scripts/build-docker-image.sh
olksdr: @jan-auer  changed to switch this one

scripts/build-docker-image.sh
olksdr: it fails locally for me, when the `args` is empty 
```
./scripts/build-docker-image.sh: line 32: args[@]: unbound variable
```

asottile-sentry: ah, ancient macos bash -- you might be able to use `declare -a args`

asottile-sentry: or put the real args into the array

PR #1659, opened 2022-12-01
ref: use apt-get instead of apt

asottile-sentry:
apt is not intended for scripting and does not have a stable commandline interface

#skip-changelog


mattgauntseo-sentry (approved):

olksdr (approved):
lgtm

PR #1660, opened 2022-12-01
build(deps): Use internal pypi for python dependencies

asottile-sentry:
this should speed up a few jobs which are currently building python deps every time

this changes the tested python to 3.8 -- we'll need to update the required checks for that (though maybe we should take this opportunity to remove it from the job id so that it doesn't need to be bumped every time?)

#skip-changelog


jan-auer:
We had the python version in there on purpose since at the time python 3.7 was the required minimum support we had to provide for Sentry. We can now bump this to 3.8, but I'd prefer if we keep it explicit given that it doesn't change that often. 

I'll make that change once approved 👍

requirements-dev.txt
```diff
@@ -1,17 +1,19 @@
-black==19.10b0
+--index-url https://pypi.devinfra.sentry.io/simple
```
jan-auer, line 1: Let's leave a comment here referring to the README for how to add new dependencies.

jan-auer (approved):
We should add a changelog entry that we're requiring python 3.8 now and bump to `0.9.0`.

olksdr (approved):

PR #1661, opened 2022-12-01
feat(profiling): Expand the discard reasons

phacops:
We only have a limited number of discard reasons for profiling and as the number of invalid profiles grow, it's difficult to understand why and make sure it's not due to a bug.

This PR aims to add more discard reasons and log an error in case we have a JSON deserialization error so we can investigate and fix it.

### DIFF
--- a/relay-profiling/src/error.rs
+++ b/relay-profiling/src/error.rs
@@ -2,8 +2,8 @@ use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum ProfileError {
-    #[error("invalid json in profile")]
-    InvalidJson(#[source] serde_json::Error),
+    #[error(transparent)]
+    InvalidJSON(#[from] serde_json::Error),
     #[error("invalid base64 value")]
     InvalidBase64Value,
     #[error("invalid sampled profile")]
@@ -20,4 +20,8 @@ pub enum ProfileError {
     InvalidTransactionMetadata,
     #[error("missing profile metadata")]
     MissingProfileMetadata,
+    #[error("malformed stacks")]
+    MalformedStacks,
+    #[error("malformed samples")]
+    MalformedSamples,
 }
--- a/relay-profiling/src/error.rs
+++ b/relay-profiling/src/error.rs
@@ -3,7 +3,7 @@ use thiserror::Error;
 #[derive(Debug, Error)]
 pub enum ProfileError {
     #[error(transparent)]
-    InvalidJSON(#[from] serde_json::Error),
+    InvalidJson(#[from] serde_json::Error),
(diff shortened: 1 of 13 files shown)


relay-profiling/src/sample.rs
```diff
@@ -233,6 +261,10 @@ pub fn expand_sample_profile(payload: &[u8]) -> Result<Vec<Vec<u8>>, ProfileErro
             }
         });
 
+        if new_profile.profile.samples.is_empty() {
+            return Err(ProfileError::NotEnoughSamples);
+        }
```
Zylphrex: This is going to drop the entire profile when one of the transactions in it has no samples right?

relay-profiling/src/outcomes.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub fn discard_reason(err: ProfileError) -> DiscardReason {
+    match err {
+        ProfileError::CannotSerializePayload => DiscardReason::FailedSerialization,
+        ProfileError::NotEnoughSamples => DiscardReason::NotEnoughSamples,
+        ProfileError::NoTransactionAssociated => DiscardReason::NoTransactionAssociated,
+        ProfileError::InvalidTransactionMetadata => DiscardReason::InvalidTransactionMetadata,
+        ProfileError::MissingProfileMetadata => DiscardReason::InvalidProfileMetadata,
+        ProfileError::MalformedStacks => DiscardReason::MalformedStacks,
+        ProfileError::MalformedSamples => DiscardReason::MalformedSamples,
```
jjbayer: Given that this is a 1:1 mapping, why not just implement a direct mapping from `ProfileError -> &'static str`?

relay-profiling/src/error.rs
```diff
@@ -2,8 +2,8 @@ use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum ProfileError {
-    #[error("invalid json in profile")]
-    InvalidJson(#[source] serde_json::Error),
+    #[error(transparent)]
+    InvalidJSON(#[from] serde_json::Error),
```
jjbayer: We usually keep acronyms camel case, see 

https://github.com/getsentry/relay/blob/4fc80841346a088c40bb4cd69165e80b5e8862f7/relay-kafka/src/producer/mod.rs#L38

and 

https://rust-lang.github.io/api-guidelines/naming.html

jjbayer (approved):
Two comments, apart from those this looks good to me!

relay-profiling/src/sample.rs
phacops: Not anymore.

Zylphrex (approved):

Issue #1630, opened 2022-11-24
Restrict transactions to 5 days into the past


jan-auer:
Currently we allow transactions to be back-dated 30 days into the past, but transaction metrics are restricted to 5 days, a restriction which our storage relies on.

It would make sense to apply the same 5 day restriction to transactions themselves now as well.

From an internal conversation:

> The system relies on that 5 days invariant. Places it relies on:
>  * The raw table has 6 days retention. I am not even sure clickhouse will materialize data older than 5 days.
>  * Inserts are not idempotent in the DB. So any migration requires us to ingest for a duration of time equal to the max backdating. If we increase that we will have a hard time migrating data.
>  * The raw table has daily partitions. If we start having concurrent writes for 30 days we can blow up the DB

jan-auer:
Transactions are now accepted with up to 5 days delay and otherwise dropped.

Note that we still observe internal errors about outdated metric inserts and buckets. Those are most likely coming from sessions, which will need a similar fix to #1663.

PR #1663, opened 2022-12-02
fix(server): Drop transactions outside of the metrics time range

jan-auer:
Relay used to accept transactions back-dated up to 30 days into the
past, but transaction metrics are restricted to 5 days, a restriction
which our storage relies on. With this PR, metric extraction now
validates the transaction timestamps and does not extract metrics if the
timestamp is outside the allowed range for metrics.

Additionally, Relay also needs to drop the transaction event for
consistency between metrics and events. Otherwise, a transaction may be
indexed that is not included in the "total" metrics dataset. This would
lead to inconsistency between the datasets and violate an invariant that
metrics contain a superset of indexed transactions.

Ref https://github.com/getsentry/relay/issues/1630

### DIFF
--- a/relay-server/src/metrics_extraction/transactions.rs
+++ b/relay-server/src/metrics_extraction/transactions.rs
@@ -15,6 +15,21 @@ use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt;
 
+/// Error returned from [`extract_transaction_metrics`].
+#[derive(Clone, Copy, Debug, Eq, PartialEq, thiserror::Error)]
+pub enum ExtractMetricsError {
+    /// The start or end timestamps are missing from the event payload.
+    #[error("no valid timestamp could be found in the event")]
+    MissingTimestamp,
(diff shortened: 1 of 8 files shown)


relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -9,11 +9,27 @@ use relay_general::protocol::{
...
+pub enum ExtractMetricsError {
+    /// The start or end timestamps are missing from the event payload.
+    #[error("no valid timestamp could be found in the event")]
+    MissingTimestamp,
+    /// The event timestamp is outside the supported range.
+    ///
+    /// The supported range is derived from the
+    /// [`max_secs_in_past`](AggregatorConfig::max_secs_in_past) and
+    /// [`max_secs_in_future`](AggregatorConfig::max_secs_in_future) configuration options.
+    #[error("timestamp too old or too much in the future")]
```
olksdr:
```suggestion
    #[error("timestamp too old or too far in the future")]
```

olksdr (approved):
lgtm!

PR #1664, opened 2022-12-02
chore(changelog): Replace dangerbot for changelog checks

### LINKED ISSUE #1654: Make Dangerbot work on external PRs
The Github token issued in forks for external PRs does not have sufficient permissions to read pull request information. This fails the Dependabot job with:

```
Request failed [403]: https://api.github.com/repos/getsentry/relay/issues/1645/comments
Response: {
  "message": "Resource not accessible by integration",
  "documentation_url": "https://docs.github.com/rest/reference/issues#create-an-issue-comment"
}
```

PRs that are blocked on this:

- https://github.com/getsentry/relay/pull/1307
- https://github.com/getsentry/relay/pull/1558
- https://github.com/getsentry/relay/pull/1645

olksdr:
Instead of using `dangerbot`,  we can just simply use our own script to for `CHANGELOG` checks. When check fails it will annotate the `CHANGELOG` file and ask to add the entry: 

image

Also give a tip in the failed check what to do:

image

with duplicating information in the log of the action:

image

#skip-changelog
fix: #1654


package.json
```diff
@@ -0,0 +1,36 @@
+{
```
jan-auer: Can we move this somewhere into `.github/workflows` or `.github/actions`? I'm not sure what the right pattern for this is, but given that we have more files now it would be better to have everything contained there. Same for `changelog.js`.

package.json
```diff
@@ -0,0 +1,36 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/getsentry/relay/issues"
+  },
+  "homepage": "https://github.com/getsentry/relay#readme",
+  "dependencies": {
+    "@actions/core": "^1.10.0",
+    "@actions/github": "^5.1.1"
+  },
+  "devDependencies": {
```
jan-auer: Do we need all of these?

package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "relay",
+  "version": "1.0.0",
+  "description": "Changelog checker.",
+  "main": "changelog.js",
+  "directories": {
+    "test": "tests"
```
jan-auer: This does not exit.

dangerfile.js
```diff
@@ -1,73 +0,0 @@
...
-  title = title.split(": ").slice(-1)[0].trim();
-  // remove links to JIRA tickets, i.e. a suffix like [ISSUE-123]
-  title = title.split("[")[0].trim();
-  // remove trailing dots
-  title = title.replace(/\.+$/, "");
-
-  return title;
-}
-
-function getChangelogDetails() {
```
jan-auer, line 17: Please bring back these instructions and write them to a [job summary](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/). Without such a description, it's going to be harder to get it right.

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  env: {
+    browser: true,
+    commonjs: true,
+    es2021: true
+  },
+  extends: 'standard',
```
jan-auer: Could we stick with something closer to what we use at Sentry? See https://github.com/getsentry/sentry/blob/d92c8f42227ab9ff4e834b1e063bdabdce28e8b4/prettier.config.js

changelog.js
```diff
@@ -0,0 +1,77 @@
...
+
+  if (core.isDebug()) {
+    core.debug(`Checking file: ${path}.`)
+    core.debug(`${fileContent}`)
+  }
+
+  return fileContent.includes(prLink)
+}
+
+async function main () {
```
jan-auer: This is just a nitpick, but if you compare this new function with the old implementation, then it is much harder to read now. Could we get this closer to the old implementation, especially since this tries to do exactly the same as before? There are a few more debug outputs now, but I'm not sure if we really need them.

jan-auer (commented):
Thank you. It makes sense that we remove danger and create GitHub Actions output directly. 

I've temporarily removed `#skip-changelog` from the PR description so we can see how it looks. My main concern right now is that it's not very straight forward to get to a description of what needs to be done to resolve or silence the warning.

jan-auer (approved):
Thanks!

iker-barriocanal (approved):
I haven't looked at the code, but the error message in CI looks good to me. Please update the PR description if it's not accurate anymore.

jjbayer (approved):
As Iker said, would be nice to update the screenshot in the PR description.

olksdr:
@jjbayer @iker-barriocanal  done

PR #1665, opened 2022-12-02
feat(metrics): Add an `abnormal_mechanism` field to Session Update protocol

shruthilayaj:
Added an optional `abnormal_mechanism` tag to the
sessions protocol to calculate and display ANR (AppNotResponding)
rates in sentry.io

We expect the SDKs to send one of two values `'anr_foreground'` or
`'anr_background'` with the `abnormal_mechanism` tag. 
Values for the `abnormal_mechanism` field is normalized as follows:
- enums `AbnormalMechanism::AnrForeground` or `AbnormalMechanism::AnrBackground` if a valid value is sent
- or to None if it isn't set or if any invalid values are sent
The valid values are extracted and stored as strings (`'anr_foreground'` or
`'anr_background'`)

Metrics extraction on the `abnormal_mechanism` tag only
happens on the `user` count metric and only when `session.status` is `abnormal`.
Tagging `abnormal_mechanism` will be supported on `SessionUpdate`
only (not `SessionAggregates`) 

The EXTRACT_VERSION is also incremented in this PR to allow feature
flagging the abnormal_mechanism extraction.

Corresponding metrics indexer change: https://github.com/getsentry/sentry/pull/41970


relay-general/src/protocol/session.rs
```diff
@@ -151,6 +182,9 @@ pub struct SessionUpdate {
     /// The session event attributes.
     #[serde(rename = "attrs")]
     pub attributes: SessionAttributes,
+    /// The abnormal mechanism.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub abnormal_mechanism: Option<AbnormalMechanism>,
```
jjbayer: Looking at the `abnormal_mechanism()` helper method, there seems to be no semantic difference between the two `None`s, so I think we could simplify this to:

```suggestion
    #[serde(default, skip_serializing_if = "AbnormalMechanism::is_none")]
    pub abnormal_mechanism: AbnormalMechanism,
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -10,7 +10,8 @@ use super::utils::with_tag;
 const METRIC_NAMESPACE: MetricNamespace = MetricNamespace::Sessions;
 
 /// Current version of metrics extraction.
-const EXTRACT_VERSION: u16 = 1;
+const EXTRACT_VERSION: u16 = 2;
```
jjbayer, line 15: Is the purpose of this version bump to stop extracting session metrics in external Relays, such that we get a quicker adoption of the new tag?

relay-server/src/actors/store.rs
```diff
@@ -476,6 +479,7 @@ impl StoreService {
                 started: protocol::datetime_to_timestamp(session.started),
                 duration: session.duration,
                 status: session.status,
+                abnormal_mechanism: session.abnormal_mechanism,
```
jjbayer: Have you been in touch with SnS about storing this new field? It might not be relevant for SaaS, but it is for self-hosted and single-tenant.

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -152,13 +158,26 @@ pub fn extract_session_metrics<T: SessionLike>(
...
                 "user",
                 MetricUnit::None,
                 MetricValue::set_from_str(distinct_id),
                 timestamp,
-                with_tag(&tags, "session.status", SessionStatus::Abnormal),
+                with_tag(
+                    &tags_for_abnormal_session,
+                    "session.status",
+                    SessionStatus::Abnormal,
+                ),
```
jjbayer: nit: `with_tag` also clones the tag map internally, so we can shave off one clone by calling `tags_for_abnormal_session.insert("session.status", SessionStatus::Abnormal)` instead.

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -152,13 +158,26 @@ pub fn extract_session_metrics<T: SessionLike>(
         ));
 
         if let Some(distinct_id) = nil_to_none(session.distinct_id()) {
+            let mut tags_for_abnormal_session = tags.clone();
+            if extract_abnormal_mechanism {
+                if let Some(ref abnormal_mechanism) = session.abnormal_mechanism() {
+                    tags_for_abnormal_session.insert(
+                        "abnormal_mechanism".to_owned(),
+                        abnormal_mechanism.to_string(),
+                    );
```
jjbayer: Let's make sure SnS is aware that we're adding a new tag to this metric.

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -152,13 +158,26 @@ pub fn extract_session_metrics<T: SessionLike>(
         ));
 
         if let Some(distinct_id) = nil_to_none(session.distinct_id()) {
+            let mut tags_for_abnormal_session = tags.clone();
```
jjbayer: Out of curiosity, why do we only need this tag on the `user` metric, and not on the session count?

relay-server/src/actors/store.rs
shruthilayaj: I've been in touch with SnS about storing this new field but we primarily talked about the resources needed to accomadate this for SaaS but self-hosted/single-tenat never came up. I guess it would be similar resource concerns on for SH/ST as well? How is this typically handled - what would I need to ask SnS to do for these particular cases?

relay-server/src/metrics_extraction/sessions.rs
shruthilayaj: This is to be able to feature flag the extraction of the abnormal mechanism tag. We want to do a slow roll out because our current hypothesis is that if we stop ingesting session duration data, our CH cluster will have enough space to handle the new tag and wouldn't require SnS to scale out the cluster. We want to be able to control the % extraction of the new tag/stop extracting if necessary

relay-server/src/metrics_extraction/sessions.rs
shruthilayaj: This is going to be used to calculate anr rate which is defined as:
total unique users who have experienced an anr / total unique users 
this is an android vital/definition and that's all we want to support in sentry

relay-server/src/metrics_extraction/sessions.rs
shruthilayaj: Yup, they are aware 👍

relay-general/src/protocol/session.rs
shruthilayaj: Addressed in https://github.com/getsentry/relay/pull/1665/commits/5c5a504b8fa4a5966909cf2fd341c22cca315e4b

relay-server/src/metrics_extraction/sessions.rs
shruthilayaj: Addressed in https://github.com/getsentry/relay/pull/1665/commits/5c5a504b8fa4a5966909cf2fd341c22cca315e4b

relay-general/src/protocol/session.rs
```diff
@@ -208,6 +248,13 @@ impl SessionLike for SessionUpdate {
             None
         }
     }
+
+    fn abnormal_mechanism(&self) -> Option<AbnormalMechanism> {
```
jjbayer: We could also change this signature to return `AbnormalMechanism` now, then the user of this function only has to handle two possible states instead of three.

relay-server/src/actors/store.rs
jjbayer: SaaS and SH/ST are currently opposites as far as session storage is concerned: SaaS stores session _metrics_, but does not populate the legacy [`sessions` entity](https://github.com/getsentry/snuba/blob/8e2d5028d57e52e9871de2d96aa57f188d0627fb/snuba/datasets/entities/sessions.py#L262), while SH/ST _only_ populates the sessions entity (which is fed by the kafka producer in this file) AFAIK.

So I guess we have two options here:
1. Implement the new feature only for metrics. In that case, we do not need to add `abnormal_mechanism` to `SessionKafkaMessage`.
2. Implement the new feature for both storages. In that case, I assume we need to change the db schema for sessions in snuba, and update the corresponding [processor](https://github.com/getsentry/snuba/blob/9683fc1beb5632265fe0e06afe21f45b280c5e52/snuba/datasets/processors/sessions_processor.py#L31).

relay-server/src/metrics_extraction/sessions.rs
jjbayer: We built this version mechanism mainly to disable metrics extraction in external relays in case there is a bug in their version. For feature flagging, I would prefer if we could pass the feature flag explicitly by

1. adding the feature to [`EXPOSABLE_FEATURES`](https://cs.github.com/getsentry/sentry/blob/75a049207fb9ab51ce422c8f18bb025226c91ed8/src/sentry/relay/config/__init__.py#L46) on the sentry side,
2. adding the feature flag in Relay and passing it into metrics extraction: https://github.com/getsentry/relay/blob/1edaf631c61968066bd2f905a0b8dcfa22dc7558/relay-server/src/actors/project.rs#L71

jjbayer: Discussed this with @jan-auer: we can go with the incremented EXTRACT_VERSION as you implemented it. As long as we do a slow rollout using sample_modulo on the sentry side, the increase in traffic to our own Relays should be manageable.

relay-general/src/protocol/session.rs
shruthilayaj: ah yup, updated in 87684f80d1a8f520e11f2bb71e09ea6f62a37001

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -377,13 +447,15 @@ mod tests {
...
+            assert_eq!(session_metric.tags["session.status"], "abnormal");
+            assert_eq!(session_metric.tags.len(), 2);
 
             let user_metric = &metrics[3];
             assert_eq!(user_metric.timestamp, started());
             assert_eq!(user_metric.name, "s:sessions/user@none");
             assert!(matches!(user_metric.value, MetricValue::Set(_)));
-            assert_eq!(user_metric.tags["session.status"], status.to_string());
+            assert_eq!(user_metric.tags["session.status"], "abnormal");
+            assert_eq!(user_metric.tags.len(), expected_user_metric_tags);
```
jjbayer: Don't we want to assert on `user_metric.tags["abnormal_mechanism"]` here?

relay-general/src/protocol/session.rs
```diff
@@ -421,4 +470,21 @@ mod tests {
...
         assert_eq!(update.attributes.ip_address, Some(IpAddr::auto()));
     }
+
+    #[test]
+    fn test_session_invalid_abnormal_mechanism() {
+        let json = r#"{
+  "sid": "8333339f-5675-4f89-a9a0-1c935255ab58",
+  "started": "2020-02-07T14:16:00Z",
+  "status": "abnormal",
+  "abnormal_mechanism": "invalid_mechanism",
```
jjbayer, line 496: Let's also add tests for parsing valid abnormal mechanisms, just to be sure.

jjbayer (commented):
This looks good, just two comments about test coverage.

relay-general/src/protocol/session.rs
shruthilayaj: added! 1a9c225d026c3ff6267866ef2423d22bf1803af4

relay-server/src/metrics_extraction/sessions.rs
shruthilayaj: done 1a9c225d026c3ff6267866ef2423d22bf1803af4

jjbayer (approved):
Thanks for contributing!

PR #1667, opened 2022-12-03
fix(sampling): Sample transactions without DSC

jan-auer:
Dynamic sampling relies on the Dynamic Sampling Context (DSC). The envelope header contains information about the trace, the origin project, and attributes for trace sampling. 

Originally, when the DSC is not specified in an envelope, dynamic sampling did not run on the envelope. Transaction events in this envelope are effectively sampled at 100% percent. However, there are more cases where the DSC is not available or not valid:

 - Old or incompatible SDKs do not send a DSC, but the transaction needs to be sampled at least with the uniform sample rate.
 - The DSN public key in the DSC could point to a nonexistent project or a project from another organization.
 - There could be an error resolving the project state for the specified project.

This PR makes the following changes: 

- Uses the event to generate an ad-hoc DSC for trace sampling if there's no DSC in the event or the sampling project state cannot be retrieved for any reason.
- Ensures that only rules from the same organization are used for trace sampling, falling back to the local project.

### DIFF
--- a/relay-server/src/actors/project.rs
+++ b/relay-server/src/actors/project.rs
@@ -767,7 +767,7 @@ impl Project {
     ) {
         if let Ok(checked) = self.check_envelope(envelope, envelope_context) {
             if let Some((envelope, envelope_context)) = checked.envelope {
-                let process = ProcessEnvelope {
+                let mut process = ProcessEnvelope {
                     envelope,
                     envelope_context,
                     project_state,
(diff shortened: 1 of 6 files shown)


relay-server/src/actors/project.rs
```diff
@@ -800,8 +810,14 @@ impl Project {
 
     /// Adds the project state for dynamic sampling and submits the Envelope for processing.
     fn flush_sampling(&self, mut message: ProcessEnvelope) {
-        // Intentionally ignore all errors and leave the envelope unsampled.
-        message.sampling_project_state = self.valid_state();
+        // Intentionally ignore all errors. Fallback sampling behavior applies in this case.
+        if let Some(state) = self.valid_state() {
+            // Never use rules from another organization.
+            if state.organization_id == message.project_state.organization_id {
+                message.sampling_project_state = Some(state);
```
jjbayer, line 817: Should we log a statsd metric to know how often this happens?

jjbayer (approved):
This looks good to me, do we need to add a test to cover the new behavior?

relay-server/src/actors/project.rs
jan-auer: I was thinking about this. Are we interested in how often this happens or how many projects this happens with? I haven't found a good way to log something truly useful.

jjbayer: Hmm yeah maybe add the project ID / org ID to a set metric?

jjbayer (approved):

PR #1466, opened 2022-09-13
feat(protocol): Track transaction name changes

jan-auer:
Introduces `propagations` and `changes` that provides more insight into how a
transaction name was changed during the lifetime of a transaction span. Since
the transaction name is propagated to downstream SDKs as part of the Dynamic
Sampling Context (DSC), this is a relevant quality metric for sampling traces by
transaction name.

All traces where propagations occur before the final transaction name is set
may not be sampled consistently if there are rules matching on the transaction
name.

Ref https://github.com/getsentry/sentry-javascript/issues/5679

PR #1671, opened 2022-12-05
ref(server): Remove internal dynamic sampling metrics

jan-auer:
The metrics introduced in #1466 are no longer needed and can be removed. We keep
the protocol to ingest information about propagations, however.

#skip-changelog


olksdr (approved):
love seeing red PRs :shipit:

jjbayer (approved):

PR #1677, opened 2022-12-06
feat: Unreal Crash Reporter event level

bitsandfoxes:
The Unreal Crash Reporter allows for sending events triggered not only through `crashes` but [`asserts`
 and `ensures`](https://docs.unrealengine.com/4.26/en-US/ProgrammingAndScripting/ProgrammingWithCPP/Assertions/) as well.  Right now, everything reported through the Unreal Crash Reporter ends up with the event level `fatal`.

Proposal:
- Asserts (Check): Halt the execution and will be reported as `Error`
- Ensure: Do not halt execution and will be reported as `Warning`


jjbayer (commented):
This looks good to me, would be nice to add some test coverage though.

jjbayer (approved):
Thanks for adding tests!

PR #1679, opened 2022-12-07
ci: Use Google Cloud Build image for self-hosted end-to-end tests

emmatyping:
Instead of building the image all over again. This should make this action a bit faster, and once the image building has moved to Github Actions, it can pull from the cached image there, which should be nice and fast.

_#skip-changelog_


olksdr (approved):

.github/workflows/ci.yml
```diff
@@ -362,39 +362,30 @@ jobs:
...
+          fi
+          echo "Polling for $RELAY_TEST_IMAGE"
+          until docker pull "$RELAY_TEST_IMAGE" 2>/dev/null; do
+              sleep 10
+          done
+          echo "relay-test-image-name=$RELAY_TEST_IMAGE" >> "$GITHUB_OUTPUT"
       # TODO: push the image here
       - name: Run Sentry self-hosted e2e CI
-        uses: getsentry/action-self-hosted-e2e-tests@5bbd3d6725bbe4bf5cc12f8a3291ce87df6a891f
+        uses: getsentry/action-self-hosted-e2e-tests@fa5b8240848f0e645ac2918c530e60ec8f50e4b8
```
chadwhitacre, line 387: Sidecar version bump? Or necessary for this to work?

chadwhitacre: Remind me, did we try using a branch name here and were asked to SHA pin?

chadwhitacre (approved):

.github/workflows/ci.yml
emmatyping: It is a needed change, since it removes the requirement to checkout the project in the action. 

I am using SHA versions until we are comfortable to call the action stable and make a release. I think it is OK to use a branch name since it is our action, but maybe @asottile-sentry has an opinion?

asottile-sentry: for `sentry` / `getsentry` we've standardized on freezing these -- but I don't have super strong opinions elsewhere

PR #1682, opened 2022-12-07
bug(replays): Use all bytes after the first newline split as the body value

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/227

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Make `attachment_type` on envelope items forward compatible by adding fallback variant. ([#1638](https://github.com/getsentry/relay/pull/1638))
 - Relay no longer accepts transaction events older than 5 days. Previously the event was accepted and stored, but since metrics for such old transactions are not supported it did not show up in parts of Sentry such as the Performance landing page. ([#1663](https://github.com/getsentry/relay/pull/1663))
 - Apply dynamic sampling to transactions from older SDKs and even in case Relay cannot load project information. This avoids accidentally storing 100% of transactions. ([#1667](https://github.com/getsentry/relay/pull/1667))
-- Replay recording parser now uses the entire body rather than a subset.  ([#1682](https://github.com/getsentry/relay/pull/1682))
+- Replay recording parser now uses the entire body rather than a subset. ([#1682](https://github.com/getsentry/relay/pull/1682))
 
 **Internal**:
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Make `attachment_type` on envelope items forward compatible by adding fallback variant. ([#1638](https://github.com/getsentry/relay/pull/1638))
 - Relay no longer accepts transaction events older than 5 days. Previously the event was accepted and stored, but since metrics for such old transactions are not supported it did not show up in parts of Sentry such as the Performance landing page. ([#1663](https://github.com/getsentry/relay/pull/1663))
 - Apply dynamic sampling to transactions from older SDKs and even in case Relay cannot load project information. This avoids accidentally storing 100% of transactions. ([#1667](https://github.com/getsentry/relay/pull/1667))
+- Replay recording parser now uses the entire body rather than a subset.  ([#1682](https://github.com/getsentry/relay/pull/1682))
```
```suggestion
- Replay recording parser now uses the entire body rather than a subset. ([#1682](https://github.com/getsentry/relay/pull/1682))
```

relay-replays/src/recording.rs
```diff
@@ -17,17 +17,30 @@ use serde::{Deserialize, Serialize};
...
+    // Find the body value.
+    let mut body: Vec<u8> = vec![];
+    let mut cursor = io::Cursor::new(bytes);
+    cursor.set_position((header.len() + 1).try_into().unwrap());
+    cursor.read_to_end(&mut body)?;
+
+    // Check for null body condition.
+    if body.is_empty() {
+        return Err(RecordingParseError::Message("no body found.".to_string()));
+    }
```
```suggestion
    if  if bytes.len() <= header.len() + 1 {
        return Err(RecordingParseError::Message("no body found.".to_string()));
    }
```

relay-replays/src/recording.rs
```diff
@@ -17,17 +17,30 @@ use serde::{Deserialize, Serialize};
...
 
     // Deserialization.
-    let mut events = loads(body)?;
+    let mut events = loads(body.as_slice())?;
```
```suggestion
    // From the check above we know that we have more data in `bytes`.
    let mut events = loads(&bytes[header.len() + 1..])?;
```

olksdr (commented):
I left few suggestions. 
My questions is just why don't use the bytes slice directly and avoid wrapping it in the cursor and introducing another `Vec`, where we again just copying data. 

I've tested the proposed changes locally and looks like all the tests are still passing. 

Please, have a look and let me know what you think.

cmanallen:
> My questions is just why don't use the bytes slice directly and avoid wrapping it in the cursor and introducing another Vec, where we again just copying data.

@olksdr The answer to these questions and more is that I do not know rust :D.  Strict code reviews are appreciated and encouraged.

olksdr (approved):

PR #1683, opened 2022-12-07
feat(replays): Filter recordings which could not be parsed

cmanallen:
Drop replay recording payloads if they fail processing, which happens either when they contain invalid JSON or when data scrubbing fails. If replay processing is disabled, Relay does not inspect the payloads and unconditionally retains them.

Closes https://github.com/getsentry/replay-backend/issues/229


iker-barriocanal (approved):

iker-barriocanal:
Is this PR still relevant?

cmanallen:
@iker-barriocanal Still relevant.  If PII scrubbing fails we'll want to reject.  I'll update this pull as we get closer to that point.

jjbayer (approved):

PR #1685, opened 2022-12-08
fix(profiling): Warn on invalid JSON errors

phacops:
We'd like the error to be logged but not sent to Sentry as it's high cardinality and not very useful in Sentry.

Changing it to a warn level log seems to log in Cloud Logging but not send anything to Sentry, which is the behavior we'd like to have.

### DIFF
--- a/relay-profiling/src/outcomes.rs
+++ b/relay-profiling/src/outcomes.rs
@@ -3,6 +3,9 @@ use crate::ProfileError;
 pub fn discard_reason(err: ProfileError) -> &'static str {
     match err {
         ProfileError::CannotSerializePayload => "profiling_failed_serialization",
+        ProfileError::InvalidBase64Value => "profiling_invalid_base64_value",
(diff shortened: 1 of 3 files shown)


Zylphrex (approved):

relay-profiling/src/outcomes.rs
```diff
@@ -3,13 +3,15 @@ use crate::ProfileError;
 pub fn discard_reason(err: ProfileError) -> &'static str {
     match err {
         ProfileError::CannotSerializePayload => "profiling_failed_serialization",
+        ProfileError::InvalidBase64Value => "profiling_invalid_base64_value",
+        ProfileError::InvalidJson(_) => "profiling_invalid_json",
```
jan-auer, line 7: We should use the existing discard reasons for this, such as `DiscardReason::InvalidJson` to keep cardinality of the column overall as low as possible. However, we can also follow up about this.

jan-auer (approved):

PR #1686, opened 2022-12-08
ci: Publish Docker images on DockerHub via Github Action

emmatyping:
This PR simply adds the docker login information (to be stored in a secret) so that the self-hosted e2e test action can push to DockerHub. For now the action pushes to a different tag, as not to conflict with Google Cloud Build pushing to the same repo at the same time. This will be changed once we are confident this action is stable and we can remove Google Cloud Build outright.

_#skip-changelog_


.github/workflows/ci.yml
```diff
@@ -389,3 +389,4 @@ jobs:
           project_name: relay
           local_image: ${{ steps.image_pull.outputs.relay-test-image-name }}
           docker_repo: docker.io/getsentry/relay
+          docker_password: ${{ secrets.DOCKERHUB_PASSWORD }}
```
hubertdeng123: should the action-self-hosted-e2e-tests@5bbd3d6725bbe4bf5cc12f8a3291ce87df6a891f be the newest commit?

emmatyping: Yes, I should update it to 05814e09c6cc1d9d1ac1c03a22216b7fbed1756a actually.

olksdr (approved):

chadwhitacre (approved):

hubertdeng123 (approved):

Issue #1244, opened 2022-04-27
Accept transactions containing unfinished spans


bruno-garcia:
https://getsentry.atlassian.net/browse/INGEST-1109

Currently, Relay will drop transactions that contain Spans that don't have an end timestamp.

The code in question:

https://github.com/getsentry/relay/blob/569678659bac656d6c4a6edcbdf3381677b5eaa2/relay-general/src/store/transactions.rs#L110-L130

https://github.com/getsentry/relay/blob/569678659bac656d6c4a6edcbdf3381677b5eaa2/relay-general/src/store/transactions.rs#L110-L130

In practice, it's common in SDKs that a transaction is finished while an asynchronous task didn't complete yet. This in turn results in confusion, particularly when SDKs are being developed as the engineer doesn't see any transactions showing up.

Different SDKs chose different approaches on how to handle it. But it's mainly:

- Exclude Spans that are not finished before sending the transaction out
- Set the end timestamp to be the same as the transaction and status `timeline_exceeded`
    - Dart → [https://github.com/getsentry/sentry-dart/blob/75bd585dbe9a7553f60f751614e6506521945b21/dart/lib/src/sentry_tracer.dart#L70-L76](https://github.com/getsentry/sentry-dart/blob/75bd585dbe9a7553f60f751614e6506521945b21/dart/lib/src/sentry_tracer.dart#L70-L76)
    - Java → [https://github.com/getsentry/sentry-java/blob/d1029084e146b5ff02b1359e8bdf0a7252ebf2e5/sentry/src/main/java/io/sentry/SentryTracer.java#L220-L236](https://github.com/getsentry/sentry-java/blob/d1029084e146b5ff02b1359e8bdf0a7252ebf2e5/sentry/src/main/java/io/sentry/SentryTracer.java#L220-L236)
    - Cocoa → [https://github.com/getsentry/sentry-cocoa/blob/a92d2b74174cea7de992e184f19a38d16dbb84f1/Sources/Sentry/SentryTracer.m#L273-L281](https://github.com/getsentry/sentry-cocoa/blob/a92d2b74174cea7de992e184f19a38d16dbb84f1/Sources/Sentry/SentryTracer.m#L273-L281)

We've identified the different implementations in the SDKs and are now leaning towards the latter with the reasoning that it's better to see what was going on during the transaction (could affect performance) even though it didn't complete. And the status code will help clearly indicate that this span didn't complete with the transaction. 

[...]

PR #1690, opened 2022-12-13
feat(general): Don't fail events containing a span with missing timestamp

TBS1996:
If theres a missing timestamp on a span, it will get the timestamp of the event it belongs to, and it will get the status of DeadlineExceeded. Check issue for more info

Ref #1244

### DIFF
--- a/relay-general/src/store/transactions.rs
+++ b/relay-general/src/store/transactions.rs
@@ -271,11 +271,14 @@ impl Processor for TransactionsProcessor {
             normalize_transaction_name(&mut event.transaction)?;
         }
 
+        validate_transaction(event)?;
+
         let spans = event.spans.value_mut().get_or_insert_with(|| Vec::new());
 
         for span in spans {
             if let Some(span) = span.value_mut() {
                 if span.timestamp.value().is_none() {
+                    // event timestamp guaranteed to be `Some` due to validate_transaction call
                     span.timestamp.set_value(event.timestamp.value().cloned());
                     span.status = Annotated::new(SpanStatus::DeadlineExceeded);
                 }
@@ -285,8 +288,6 @@ impl Processor for TransactionsProcessor {
                 ));
             }
         }
-
-        validate_transaction(event)?;
 
         set_default_transaction_source(event);
 
--- a/relay-general/src/store/transactions.rs
+++ b/relay-general/src/store/transactions.rs
@@ -283,6 +283,7 @@ impl Processor for TransactionsProcessor {
             if let Some(val) = span.value_mut() {
                 if val.timestamp.value().is_none() {
                     val.timestamp.set_value(event.timestamp.value().cloned());
+                    val.status = Annotated::new(SpanStatus::DeadlineExceeded);
                 }
             } else {
                 return Err(ProcessingAction::InvalidTransaction(
--- a/relay-general/src/store/transactions.rs
+++ b/relay-general/src/store/transactions.rs
@@ -814,45 +814,6 @@ mod tests {
     }
 
     #[test]
-    fn test_discards_transaction_event_with_span_with_missing_timestamp() {
-        let mut event = Annotated::new(Event {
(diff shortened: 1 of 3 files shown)


relay-general/src/store/transactions.rs
```diff
@@ -421,6 +423,63 @@ mod tests {
...
+
+        assert_eq!(
+            event.spans.value().unwrap()[0]
+                .value()
+                .unwrap()
+                .status
+                .value()
+                .unwrap(),
+            &SpanStatus::DeadlineExceeded
+        );
```
jjbayer, line 460: I think we can just move this assertion into the previous test. Both tests execute the same function on the same testing data, so there's no reason to split the two asserts into separate tests.

relay-general/src/store/transactions.rs
```diff
@@ -275,18 +271,23 @@ impl Processor for TransactionsProcessor {
...
-        validate_transaction(event)?;
-
         let spans = event.spans.value_mut().get_or_insert_with(|| Vec::new());
 
         for span in spans {
-            if span.value().is_none() {
+            if let Some(val) = span.value_mut() {
+                if val.timestamp.value().is_none() {
+                    val.timestamp.set_value(event.timestamp.value().cloned());
+                    val.status = Annotated::new(SpanStatus::DeadlineExceeded);
```
jjbayer: nit: I would rename `val` to `span` here to increase readability.

jjbayer (commented):
Looks good to me, just two minor comments.

jjbayer (approved):
Please add a changelog as discussed, apart from that, this looks good!

relay-general/src/store/transactions.rs
```diff
@@ -275,18 +271,23 @@ impl Processor for TransactionsProcessor {
...
 
-        validate_transaction(event)?;
-
         let spans = event.spans.value_mut().get_or_insert_with(|| Vec::new());
 
         for span in spans {
-            if span.value().is_none() {
+            if let Some(span) = span.value_mut() {
+                if span.timestamp.value().is_none() {
+                    span.timestamp.set_value(event.timestamp.value().cloned());
```
jan-auer, line 282: Syntactically, the event timestamp could be `None` here. Is there a validation that checks this before?

If we normalize the event timestamp somewhere else (I think we default it to the receive timestamp), we should also apply the same logic here and put the same default.

jjbayer: We did a quick survey and it seems that the event timestamp is changed in these places:

* [`process_minidump`](https://github.com/getsentry/relay/blob/1edaf631c61968066bd2f905a0b8dcfa22dc7558/relay-server/src/utils/native.rs#L204), which runs as part of `create_placehodlers` _before_ light normalization, which should be fine.
* [`normalize_timestamps`](https://github.com/getsentry/relay/blob/1edaf631c61968066bd2f905a0b8dcfa22dc7558/relay-general/src/store/normalize.rs#L532): This runs _after_ transaction processor in light normalization, which would be a problem, but if I understand correctly, the clock drift processor fixes all timestamps in the event by calling [`process_child_values`](https://github.com/getsentry/relay/blob/1edaf631c61968066bd2f905a0b8dcfa22dc7558/relay-general/src/store/clock_drift.rs#L122), so it will alter the event timestamp and the span timestamp in the same way.

olksdr: Maybe adding test to verify these findings would be a good idea.

jan-auer (approved):

relay-general/src/store/transactions.rs
TBS1996: - we have a test that asserts a valid event timestamp: https://github.com/getsentry/relay/blob/026536a9d761cc8dd85982e161e5b30a4b4b022f/relay-general/src/store/transactions.rs#L418-L420
- `process_minidump` is for error events and not transactions
- `normalize_timestamps` is tested with clock drift correction already

PR #1692, opened 2022-12-13
feat(relay): Implement decaying rule [TET-607]

iambriccardo:
A decaying function is a function that interpolates the value of the sample rate based on a set of parameters (e.g. `start`, `end`, `fromSampleRate`). The rationale behind the decaying function is to be able to systematically change the sample rate based on multiple parameters. This should be especially helpful with the behavior of latest releases, in which the number of transactions grows non immediately and we want to inversely follow that trend.

There exist currently two types of decaying functions:

- `constant` → returns the `sampleRate` irrespectively of the time of evaluation.
    
    ```python
    {
    	...
    	"decayingFn": {
    		"type": "constant"	
      }
    	...
    }
    ```
    
- `linear` → returns a decreasing sample rate from `sampleRate` to `decayedSampleRate` based on the time of evaluation.
    
    ```python
    { 
      ...
    	"decayingFn": {
    		"function": "linear",
    		"decayedSampleRate": 0.6
    	}
    	...
    }
    ```
    
_By default, if no decaying function is specified, the `constant` function will be used, which returns the `sampleRate` at any point within the specified time range._

Both decaying functions require time ranges, that are fetched from the previously existing `timeRange` field:

```python
{ 
	...
  "timeRange": {
		"start": "2022-10-21 18:50:25+00:00",
		"end": "2022-10-21 19:50:25+00:00"
	},
	"decayingFn": {
		"function": "linear",
		"decayedSampleRate": 0.6
	}
	...
}
```

Depending on which `decayingFn` is used, the `timeRange` check will be different:

- `constant` → a valid time range can be open or closed (which means that you can have both intervals closed or at most one side of the interval open) and, if both are set, `start < end`.
- `linear` → a valid time range must be closed and `start < end`.

*In case at least one of the required conditions are not met, the rule with be considered `inactive` and won’t be matched.*

# Implementation
The implementation of decaying rules introduces the idea of an `ActiveRule` which is a rule returned only in case it is active. A rule is active if it satisfies a series of conditions that are checked in relation to the `decayingFn` used.

An `ActiveRule` has the `get_sample_rate(now)` method which returns the sample rate at point in time `now`.


relay-sampling/src/lib.rs
```diff
@@ -355,6 +365,125 @@ impl TimeRange {
...
+    /// is equal to 0.06.
+    fn execute_linear_decay(&self, from_sample_rate: f64) -> f64 {
+        let now_timestamp = self.external_params.now.timestamp() as f64;
+        let start_timestamp = self.external_params.start.timestamp() as f64;
+        let end_timestamp = self.external_params.end.timestamp() as f64;
+
+        // We round to 2 digits in order to avoid high-precision sample rates that are more difficult
+        // to work with. Rounding is performed following the nearest integer.
+        let sample_rate_difference =
+            f64::round((from_sample_rate - self.external_params.base_sample_rate) * 100.0) / 100.0;
```
jjbayer: Why are rounded sample rates easier to work with?

relay-sampling/src/lib.rs
```diff
@@ -355,6 +365,125 @@ impl TimeRange {
...
+        let progress = (now_timestamp - start_timestamp) / (end_timestamp - start_timestamp);
+        let inverse_progress = (1.0 - progress).clamp(0.0, 1.0);
+        self.external_params.base_sample_rate + (sample_rate_difference * inverse_progress)
+    }
```
jjbayer: Instead of computing `inverse_progress`, why not just define `sample_rate_difference` as `base_rate - from_rate`, and return `from_rate + difference * progress`?

relay-sampling/src/lib.rs
```diff
@@ -355,6 +365,125 @@ impl TimeRange {
...
+
+impl DecayingFunction {
+    /// Validate the external parameters needed by the decaying function(s).
+    fn validate_external_params(
+        &self,
+        base_sample_rate: f64,
+        time_range: TimeRange,
+        now: DateTime<Utc>,
+    ) -> Option<DecayingFunctionExecutor> {
+        if let Some((start, end)) = time_range.extract_interval_tuple() {
```
jjbayer: This way, there's no need for the `extract_interval_tuple` helper function:

```suggestion
        if let (Some(start), Some(end)) = time_range {
```

relay-sampling/src/lib.rs
```diff
@@ -355,6 +365,125 @@ impl TimeRange {
...
+            if !self.validate_function_specific_params(base_sample_rate, time_range, now)
+                || end <= start
+                || now < start
```
jjbayer: `TimeRange` has a `contains` method that you could use here.

relay-sampling/src/lib.rs
```diff
@@ -355,6 +365,125 @@ impl TimeRange {
...
+    /// end = 200
+    /// from = 0.7 (70% sample rate)
+    /// base_sample_rate = 0.5 (50% sample rate)
+    /// (from - base_sample_rate) = 0.2 (20% sample rate)
+    ///
+    /// If we have now = 170 it means we are at the 70% progress of the timerange, which means we
+    /// want to inversely reduce the sample rate in the interval (from - base_sample_rate).
+    /// Thus, in the 70% case, we would take only 30% of the (from - base_sample_rate) which
+    /// is equal to 0.06.
+    fn execute_linear_decay(&self, from_sample_rate: f64) -> f64 {
```
jjbayer: nit: There are variables in this function that begin with `start_` and `end_`, so to clarify intent I would rename `from_sample_rate` to `start_sample_rate`, and create an alias `let end_sample_rate = self.external_params.base_sample_rate`.

relay-sampling/src/lib.rs
```diff
@@ -355,6 +365,125 @@ impl TimeRange {
...
+            f64::round((from_sample_rate - self.external_params.base_sample_rate) * 100.0) / 100.0;
+        let progress = (now_timestamp - start_timestamp) / (end_timestamp - start_timestamp);
+        let inverse_progress = (1.0 - progress).clamp(0.0, 1.0);
+        self.external_params.base_sample_rate + (sample_rate_difference * inverse_progress)
+    }
+}
+
+/// A struct containing the set of external params required by a decaying function.
+#[derive(Clone, Copy, Debug)]
+struct DecayingFunctionExternalParams {
```
jjbayer: I feel that the types involved in this logic are too tightly coupled to the serialization format, in that we differentiate between internal and external params.

In order to decouple them, can we construct something like that from the `SamplingRule`, e.g.

```rust

struct LinearDecay {
    // whatever model parameters a linear function needs. In theory only two (`(k, d)` for `y = kx + d`),
    // but it might make sense to store `(start, end)` timestamps and sample rates as well.
    start_time: DateTime<Utc>,
    end_time: DateTime<Utc>,
    start_rate: f64,
    end_rate: f64,
}

enum DecayFunc {
    // ...
    Linear(LinearDecay),
}

impl SamplingRule {
   fn get_decay_fn(&self) -> DecayFunc {
       // construct DecayFunc from rule here. We can also cache it on the sample rule
   }
}
```

jjbayer (commented):
The general concept makes sense to me, but I have some comments on the code structure.

relay-sampling/src/lib.rs
iambriccardo: It is easier to debug and in general I thought we were going to keep 2 decimal digits.

relay-sampling/src/lib.rs
iambriccardo: Yep that can also work, will rewrite it.

relay-sampling/src/lib.rs
iambriccardo: Contains has the catch of allowing open intervals in both ends, I don't allow that in decaying functions.

relay-sampling/src/lib.rs
iambriccardo: How would you deserialize this? I wanted to keep my change as small as possible on the existing rule payload.

iambriccardo: Your proposal makes total sense, indeed, it is the simplest way of doing it but this requires the removal of the timerange, which we can for sure do but we need to be careful with backward compatibility.

relay-sampling/src/lib.rs
jjbayer: By multiplying `sample_rate_difference * inverse_progress`, you potentially get a lot more decimals again, right? If we want to do 1% steps of decay (e.g. `0.7, 0.69, 0.68, ...`) I would round when the computation is done, e.g. in `get_sample_rate`.

relay-sampling/src/lib.rs
jjbayer: I would leave the existing rule payload as-is, and create these helper structs independently.

iambriccardo: This would end up creating duplication within the json. Do we want that? And how do we treat the TimeRange, because it won't be needed anymore if we keep intervals within the functions themselves.

relay-sampling/src/lib.rs
iambriccardo: What you are saying is correct but I did that on purpose.

I had at first rounded only the sample difference because it was behaving quite weird, in the sense that I was getting 0.7 - 0.2 = 0.4999999 due to floating arithmetic approximations I think. And this was going to screw up the whole calculation because we started doing that with an ill number.

I think that we could simplify everything and round at the end, with some more loss propagated through multiplications but that is fine. The only problem I see is that we will have weird numbers in tests if we choose for example the 50% progress time.

jan-auer: Do increased precision and floating point glitches really make this more difficult to work with? In the end, you can just make your "correct" calculation and use the result. If you ever have to debug print the value and receive `0.49999...`, it is still clear what the value is. Besides, interpolation will give you all sorts of values anyway. 

Also, please do not round to integer percentages. We need more precision than that, considering that a realistic uniform sample rate is 1%. You can, of course, round your rule targets, but please do not round here.

I would suggest to remove complexity and not round anywhere in Relay.

iambriccardo: Great! Thanks for the explanation Jan!

relay-sampling/src/lib.rs
```diff
@@ -355,6 +357,48 @@ impl TimeRange {
...
+struct DecayingFunctionParams {
+    sample_rate: f64,
+    progress_ratio: f64,
+}
+
+/// A decaying function definition.
+///
+/// A decaying function is responsible of decaying the sample rate from a value to another following
+/// a given curve.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
```
```suggestion
#[derive(Default, Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
```

relay-sampling/src/lib.rs
```diff
@@ -355,6 +357,48 @@ impl TimeRange {
...
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "type")]
+pub enum DecayingFunction {
+    #[serde(rename_all = "camelCase")]
+    Linear {
+        decayed_sample_rate: f64,
+    },
+    Constant,
```
```suggestion
    #[default]
    Constant,
```

relay-sampling/src/lib.rs
```diff
@@ -355,6 +357,48 @@ impl TimeRange {
...
+        }
+    }
+}
+
+impl Default for DecayingFunction {
+    fn default() -> Self {
+        Self::Constant
+    }
+}
+
```
```suggestion
```

relay-sampling/src/lib.rs
```diff
@@ -381,8 +427,58 @@ impl SamplingRule {
...
+    /// Returns the sample rate of this specific sampling rule based on the existence of the
+    /// decaying function.
+    ///
+    /// In case of any problems that render the decaying function impossible to compute, we are
+    /// going to fallback to the base sample rate.
+    pub fn get_sample_rate(&self, now: DateTime<Utc>) -> f64 {
+        if let Some(params) = self.get_decaying_function_params(now) {
+            self.decaying_fn.call(params)
+        } else {
+            self.sample_rate
```
jjbayer: Now that `sample_rate` is not the base sample rate anymore, I believe that instead of falling back to this value we should make sure that invalid rules do not match any events. We already have the `is_active` mechanism for that, should we extend that function so it returns false for invalid rules?

https://github.com/getsentry/relay/blob/0e8bd2273e37e8fc0bc71c78f05cef07c28266a1/relay-sampling/src/lib.rs#L741-L742

relay-sampling/src/lib.rs
```diff
@@ -381,8 +427,58 @@ impl SamplingRule {
...
+            if !self.validate_function_specific_params() || end <= start || now < start {
+                return None;
+            }
+
+            let now_timestamp = now.timestamp() as f64;
+            let start_timestamp = start.timestamp() as f64;
+            let end_timestamp = end.timestamp() as f64;
+            let progress_ratio = ((now_timestamp - start_timestamp)
+                / (end_timestamp - start_timestamp))
+                .clamp(0.0, 1.0);
```
jjbayer: What happens when `now > end`? Do we return the decay sample rate because of this clamp, or is there some validation somewhere that disallows `now > end`?

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -182,18 +204,37 @@ mod tests {
         state
     }
 
-    fn state_with_rule(
+    fn state_with_non_decaying_rule(
```
jjbayer: I would leave this function name as it was. The rename makes the diff harder to read.

relay-sampling/src/lib.rs
```diff
@@ -2296,6 +2443,9 @@ mod tests {
                         start: Some(Utc.ymd(1970, 10, 30).and_hms(0, 0, 0)),
                         end: Some(Utc.ymd(3000, 10, 10).and_hms(0, 0, 0)),
                     },
+                    decaying_fn: DecayingFunction::Linear {
+                        decayed_sample_rate: 0.7,
+                    },
```
jjbayer: By changing this helper function, all the tests that use it are now testing `DecayingFunction::Linear` instead of `DecayingFunction::Constant`. Would it not be better to leave this helper function as it was, and test the new linear behavior in another way?

relay-sampling/src/lib.rs
iambriccardo: In my mind, if you pass a `timeRange` without `decayingFn`, you will use directly/implicitly the `Constant` function which only uses the `sampleRate` and the rule by itself is valid.

relay-sampling/src/lib.rs
iambriccardo: `now > end` is fine mathematically speaking (thanks to our implementation) but it will be not reached due to the `is_active()` check.

relay-sampling/src/lib.rs
iambriccardo: @jan-auer this is a good point, with the new implementation, we would need to decide what to do in case of invalid params. We could either:
1. Implement a mechanism that automatically falls back either to `decayedSampleRate` or `sampleRate` based on the function type.
2. Make automatically the rule invalid by performing a validation of params before `get_sample_rate`
3. Propagate some error but this is generally not suggested and looking at the current implementation it is not done

_Just to clarify, this can happen for example if `timeRange` has `start` and not `end`, irrespectively if we specify a decaying function or not._

relay-sampling/src/lib.rs
iambriccardo: The decaying function behavior is tested in the other tests because what we care is to test the sample rate calculation. In this file we test on the other hand the matching logic which doesn't have anything to do with the sample rate.

The `decaying_fn` specified in this test is added because it is required by the struct, I could convert the `Linear` to `Constant` to signal that we don't care about it specifically.

jjbayer: 👍 thanks!

relay-sampling/src/lib.rs
jjbayer: We could write a custom deserializer for `SamplingConfig::rules`, which removes structurally invalid rules and logs an error for them. But `get_decaying_function_params` also depends on `now`, and that part of the "validation" should go into `is_active` IMO.

iambriccardo: I think that first we would need to clarify whether we want to allow open intervals or not.

If we wanted to do it better, I would have redefined the timerange with only closed intervals and then automatically use the constant decaying if not specified. Then for each function build custom extra validation and fallback mechanism(s).

I don't know tho, if there is a proper rationale behind the decision of open intervals. For DS we don't need it (as of now).

CHANGELOG.md
```diff
@@ -2,9 +2,10 @@
 
 ## Unreleased
 
-**Features:**
+**Features**:
 
 - Add support for `limits.keepalive_timeout` configuration. ([#1645](https://github.com/getsentry/relay/pull/1645))
+- Add support for decaying functions in sampling rules. ([#1692](https://github.com/getsentry/relay/pull/1692))
```
```suggestion
- Add support for decaying functions in dynamic sampling rules. ([#1692](https://github.com/getsentry/relay/pull/1692))
```

relay-sampling/src/lib.rs
```diff
@@ -427,57 +399,95 @@ impl SamplingRule {
     ///
```
jjbayer, line 453: Let's update this doc comment.

relay-sampling/src/lib.rs
```diff
@@ -427,57 +399,95 @@ impl SamplingRule {
...
+        match self.decaying_fn {
+            DecayingFunction::Linear {
+                decayed_sample_rate,
+            } => {
+                if let TimeRange {
+                    start: Some(start),
+                    end: Some(end),
+                } = self.time_range
+                {
+                    if self.sample_rate > decayed_sample_rate && start < end && now >= start {
```
jjbayer: Because `time_range.contains` is not called in this branch anymore, we need to also check that `now` is `< end` (feel free to change `<=` to `<` and vice versa). 
```suggestion
                    if self.sample_rate > decayed_sample_rate && start <= now && now < end {
```

We do not need to check `start < end` explicitly then, because it follows from `start <= now < end`.

relay-sampling/src/lib.rs
```diff
@@ -427,57 +399,95 @@ impl SamplingRule {
...
-            let now_timestamp = now.timestamp() as f64;
-            let start_timestamp = start.timestamp() as f64;
-            let end_timestamp = end.timestamp() as f64;
-            let progress_ratio = ((now_timestamp - start_timestamp)
-                / (end_timestamp - start_timestamp))
-                .clamp(0.0, 1.0);
+/// A sampling rule that has been successfully matched and that contains all the required data
+/// to return the sample rate.
+#[derive(Debug, Clone, Copy)]
+pub struct MatchingRule {
```
jjbayer: `ActiveRule` might actually be a better name now that I think of it, because it is created by `is_active`.

```suggestion
pub struct ActiveRule {
```

relay-sampling/src/lib.rs
```diff
@@ -427,57 +399,95 @@ impl SamplingRule {
...
     }
+}
 
-    /// Validates the external params specifically to each decaying function.
-    fn validate_function_specific_params(&self) -> bool {
-        match self.decaying_fn {
-            DecayingFunction::Linear {
+/// A struct representing the evaluation context of a sample rate.
+#[derive(Debug, Clone, Copy)]
+enum SampleRateEvaluator {
```
jjbayer: We usually declare types before using them, so In this case, declare `SampleRateEvaluator`, then `MatchingRule`, then `SamplingRule`. See [internal docs](https://www.notion.so/sentry/HOWTO-Code-Rust-at-Sentry-7b35f165b10b4492bb95ebe1471a9ada#2695321713704369bccc5fa83ebedfa8).

relay-sampling/src/lib.rs
jjbayer: Apparently this was not caught by any test case. Maybe we should add one?

iambriccardo: Actually this was done with a purpose but for the previous implementation, now we should explicitly verify like you said because the check is in `is_active`.

relay-sampling/src/lib.rs
iambriccardo: Thanks!

relay-sampling/src/lib.rs
jjbayer: ^ One line above, it says `Returns whether the sampling rule is active`. We should change it to "Returns an `ActiveRule` if the rule is active", or something like that.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -122,17 +125,32 @@ pub fn should_keep_event(
...
     processing_enabled: bool,
 ) -> SamplingResult {
-    let matching_trace_rule =
-        match get_trace_sampling_rule(processing_enabled, sampling_project_state, dsc, ip_addr) {
-            Ok(spec) => spec,
-            Err(sampling_result) => return sampling_result,
-        };
+    // For consistency reasons we take a snapshot in time and use that time across all code that
+    // requires it.
+    let now = Utc::now();
```
jjbayer, line 130: Good call!

relay-sampling/src/lib.rs
```diff
@@ -492,6 +477,21 @@ impl SampleRateEvaluator {
...
+    pub id: RuleId,
+    evaluator: SampleRateEvaluator,
+}
+
+impl ActiveRule {
+    /// Gets the sample rate for the specific rule.
+    pub fn get_sample_rate(&self, now: DateTime<Utc>) -> f64 {
+        self.evaluator.evaluate(now)
+    }
+}
```
jjbayer: nit: Because `SamplingRule` uses active rule, I would change the declaration order to 

```
SampleRateEvaluator
ActiveRule
SamplingRule
```

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -684,35 +684,16 @@ mod tests {
...
+            Some(now + DateDuration::days(1)),
         );
 
         assert_eq!(
             prepare_and_get_sampling_rule(1.0, EventType::Transaction, &project_state, now)
                 .unwrap()
                 .unwrap()
                 .sample_rate,
-            0.7
+            0.44999999999999996
```
jjbayer: This looks flaky, should we rather do something like `(expected - result).abs() < eps` instead? See https://rust-lang.github.io/rust-clippy/master/#float_cmp

jjbayer (approved):
Couple of nitpicks, apart from those this looks good to me!

relay-server/src/utils/dynamic_sampling.rs
iambriccardo: Very good call! I totally overlooked the precision, especially considering that it can differ from architecture to architecture.

jjbayer (approved):

PR #1695, opened 2022-12-14
feat(transactions): Adding support for transaction (re)naming rules

### LINKED ISSUE #1657: Support transaction renaming rules
Given the [protocol](https://www.notion.so/sentry/Clusterer-infra-design-1353df44638b4167af7358d25431ed9c#c5b6345c3e9245b7b460f8b3d2c76106), add support in relay for transaction renaming rules.

### Acceptance criteria
- [ ] Relay only applies the first matching rule.
- [ ] Relay ensures the path ends with a `/` when checking whether a rule matches, as detailed [here](https://www.notion.so/sentry/Rule-formats-c25e476637a2483cb33ca00ae48a7c5a#c45c3cae5b2b44b5b733b4fd15ffee61).
- [ ] Transaction renaming rules must be applied before applying any filters (such as dynamic sampling or inbound).
- [ ] Relay must preserve the original transaction name in the transaction's `meta` when applying the rules.
- [ ] The exact format to find the old name in a transaction's meta is documented [here](https://www.notion.so/sentry/Transaction-name-normalization-changes-42516662dc684a9abcc8d0aa2212396c#63d4f5d4472345aca08c5c6b95bf7481).
- [ ] When a transaction name has been sanitized, the transaction's source is set to `sanitized`.

olksdr:
These changes introduce the support for naming rules, which currently can be applied to only `url` transaction source:

* relay always applies first matching rule
* rules applied in light normalization, before extraxting any information from the event
* original transaction name is preserved in meta, and also remark added with the rule, which was applied
* `transaction_info.source` is changed to `sanitized` after rule is applied

This PR also contains changes to the current implementation of `Glob`, adding the ability to apply the matching rule to replace parts of the matching pattern - which maybe could be done separately.

Also few existing structs were re-used / re-purposed to support new rule format, which also up for the discussions.

closes: [#1657](https://github.com/getsentry/relay/issues/1657)

@getsentry/owners-ingest also requesting advice on structuring the code better. Since I currently re-using few utility structs and also existing protocol struct (like `TransactionSource`) to keep changes smaller and also re-used existing protocol for new things.


relay-common/src/utils.rs
```diff
@@ -40,9 +43,13 @@ impl Glob {
         pattern.push_str(&regex::escape(&glob[last..]));
         pattern.push('$');
 
+        // Prepare the new pattern replacing matches for `**` and `.` with non-capturing groups.
+        let replacer = pattern.replace("(.)", "(?:.)").replace("(.*?)", "(?:.*?)");
+
         Glob {
             value: glob.to_string(),
             pattern: Regex::new(&pattern).unwrap(),
+            replacer: Regex::new(&replacer).unwrap(),
```
jjbayer: Rather than compiling and storing an additional regex, should we just add an option to `Glob::new` to tell it which groups to capture?

relay-general/src/store/transactions/processor.rs
```diff
@@ -1,21 +1,77 @@
...
+    /// - glob pattern matches the transaction name
+    pub fn apply_transaction_rename_rule(
+        &self,
+        transaction: &mut Annotated<String>,
+        info: &mut std::option::Option<TransactionInfo>,
+        source: &TransactionSource,
+    ) -> ProcessingResult {
+        let now = Utc::now();
+        transaction.apply(|transaction, meta| {
+            let tran = format!("{}/", transaction.as_str());
```
jjbayer: Haven't thought through the consequences, but shouldn't we only add a `/` if it's not already there?

relay-general/src/store/transactions/rules.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+}
+
+/// Contains transaction attribute the rule must only be applied to.
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub struct RuleScope {
+    /// The source of the transaction.
+    pub source: TransactionSource,
+}
```
jjbayer, line 33: nit: Let's put this above `TransactionNameRule`, see https://www.notion.so/sentry/HOWTO-Code-Rust-at-Sentry-7b35f165b10b4492bb95ebe1471a9ada#2695321713704369bccc5fa83ebedfa8.

relay-general/src/store/transactions/rules.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// The rule describes how transaction name should be changed.
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub struct TransactionNameRule {
+    /// The pattern which will be applied to transaction name.
+    pub pattern: Glob,
+    /// Date time when the rule expires and it should not be applied anymore.
+    pub expiry: DateTime<Utc>,
+    /// Object containing transaction attributes the rules must only be applied to.
+    #[serde(default)]
```
jjbayer: Does it even make sense to `default` these two fields when they are missing? I would just make them required.

relay-general/src/store/transactions/rules.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub source: TransactionSource,
+}
+
+impl Default for RuleScope {
+    fn default() -> Self {
+        Self {
+            source: TransactionSource::Url,
+        }
+    }
+}
```
jjbayer, line 41: See comment above, I believe `RuleScope` should not have a default.

relay-common/src/utils.rs
olksdr: Thanks for pointing this!

It was a quick test on my side, to make sure the replacements and patterns are working. But you're definitely, it does not make any sense to keep 2 regexes inside the struct. I made the `Glob` configurable through the builder, with default behaviour like it was before (but it might be good idea to switch the builder around) and added customer deserialize function, see [b90a8ba](https://github.com/getsentry/relay/pull/1695/commits/b90a8bafd7f75cbc0f7ab8f3fbafd807ddd4025c)

relay-general/src/store/transactions/processor.rs
olksdr: Theoretically it should not matter, if there are 1 or 2 slashed at the end. The `**` pattern at the end of each rule will match anything anyway. If I, of course, see it correctly.

relay-general/src/store/transactions/rules.rs
olksdr: done in [b90a8ba](https://github.com/getsentry/relay/pull/1695/commits/b90a8bafd7f75cbc0f7ab8f3fbafd807ddd4025c)

relay-general/src/store/transactions/rules.rs
olksdr: So, this is the same comment as https://github.com/getsentry/relay/pull/1695#discussion_r1049440446

relay-general/src/store/transactions/rules.rs
olksdr: I'm was using our docs for rules definition format, and set all the defaults as the relay supposed to use. From my point of few it's better to get all of those set on deserialization rather than using `Option` and then check additionally if it's set and `unwrap` with some defaults. 

E.g. `RuleScope` which is optional completely, but we default to `url` - and it much more convenient to set it straight away to what we expect to be default. 

Let me know what you think.

relay-server/src/actors/project.rs
```diff
@@ -136,9 +136,12 @@ pub struct ProjectConfig {
     pub span_attributes: BTreeSet<SpanAttribute>,
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub metric_conditional_tagging: Vec<TaggingRule>,
-    /// Exposable features enabled for this project
+    /// Exposable features enabled for this project.
     #[serde(skip_serializing_if = "BTreeSet::is_empty")]
     pub features: BTreeSet<Feature>,
+    /// Transaction renaming rules.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub tx_name_rules: Vec<TransactionNameRule>,
```
jjbayer, line 145: Please also add to `LimitedProjectConfig`.

relay-common/src/utils.rs
```diff
@@ -56,6 +121,26 @@ impl Glob {
         self.pattern.is_match(value)
     }
 
+    /// Currently support replacing only all `*` in the input string with provided replacement.
+    /// If no match is found, then a copy of the string is returned unchanged.
+    pub fn apply(&self, input: &str, replacement: &str) -> String {
```
jjbayer: nit
```suggestion
    pub fn replace_captures(&self, input: &str, replacement: &str) -> String {
```

relay-general/src/protocol/transaction.rs
```diff
@@ -20,6 +23,8 @@ pub enum TransactionSource {
     View,
     /// Named after a software component, such as a function or class name.
     Component,
+    /// The transaction name was updated to remove high cardinality parts.
```
iker-barriocanal: Probably a nit -- a "sanitized" transaction name doesn't necessarily require getting high cardinality parts removed. Currently, we are only going to support the removal feature, but let's not limit ourselves to that.
```suggestion
    /// The transaction name was updated to reduce the name cardinality.
```

relay-general/src/protocol/transaction.rs
```diff
@@ -1,12 +1,15 @@
...
+use serde::{Deserialize, Serialize};
+
 use crate::processor::ProcessValue;
 use crate::protocol::Timestamp;
 use crate::types::{Annotated, Empty, ErrorKind, FromValue, IntoValue, SkipSerialization, Value};
 
 /// Describes how the name of the transaction was determined.
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
```
iker-barriocanal, line 12: lol TIL kebab-case, I didn't know there was a name for this. It's made my day 😂

relay-general/src/store/transactions/processor.rs
```diff
@@ -1,21 +1,89 @@
...
+            let slash_is_present = transaction
+                .chars()
+                .last()
+                .map(|c| c == '/')
+                .unwrap_or_default();
+
+            // Add new `/` at the end of the transaction if there isn't one.
+            if !slash_is_present {
+                transaction.push('/');
+            }
```
iker-barriocanal: What do you think about using a context manager here? The overall idea is:
```rs
with_slashed_transaction(|transaction_name| {
  // do the logic below
});
```
Then, `with_slashed_transaction` automatically handles this logic of (not) adding the `/` at the end.

I personally find this much easier to read and follow the semantics.

relay-general/src/store/transactions/processor.rs
```diff
@@ -1,21 +1,89 @@
...
+        Self {
+            normalize_names,
+            tx_name_rules,
+        }
+    }
+
+    /// Applies the rule if any found to the transaction name.
+    ///
+    /// It find the first rule matching the criteria:
+    /// - source matchining the one provided in the rule sorce, default `url`
```
iker-barriocanal: There's actually no default in the code in the method below, and that's correct. Using `url` by default is something controlled by the rule generation, and that lives in sentry and should be independent of how relay behaves.
```suggestion
    /// - source matchining the one provided in the rule sorce
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -1,21 +1,89 @@
...
+                    // Adding `/` at the end of the name, ensures that rules like /<something>/*/**
+                    // will always match the string.
+                    && rule.pattern.is_match(transaction)
+            });
+            let result = rule.map(|r| (r.pattern.pattern(), r.apply(transaction)));
+
+            if let Some((rule, Some(mut result))) = result {
+                if !slash_is_present {
+                    let _ = transaction.pop();
+                    let _ = result.pop();
```
```suggestion
                    transaction.pop();
                    result.pop();
```
Why not? Same on line 83.

relay-general/src/store/transactions/processor.rs
```diff
@@ -1,21 +1,89 @@
...
+                }
+                if &result != transaction {
+                    meta.set_original_value(Some(transaction.clone()));
+                    // add also the rule which was applied to the transaction name
+                    meta.add_remark(Remark::new(RemarkType::Substituted, rule));
```
iker-barriocanal: Ideally, we take this remark from the rule's redaction instead of hardcoding it. Feel free to apply this change in this PR or leave it for a follow-up one.

relay-general/src/store/transactions/processor.rs
```diff
@@ -1,21 +1,89 @@
...
+    /// - glob pattern matches the transaction name
+    ///
+    /// Note: we add `/` at the end of the transaction name if there isn't one, to make sure that
+    /// patterns like `/<something>/*/**` where we have `**` at the end are a match.
+    pub fn apply_transaction_rename_rule(
+        &self,
+        transaction: &mut Annotated<String>,
+        info: &mut std::option::Option<TransactionInfo>,
+        source: &TransactionSource,
+    ) -> ProcessingResult {
```
iker-barriocanal: There's `self` and another 3 parameters here, all of them coming from the transaction. Why don't we change the API of this method and pass directly the transaction instead of each element we need? Then, the method will extract and modify what it needs to. The benefit is further extensibility in the future -- no need to modify the API if we are planning to add more conditions. It also makes the code in the caller easier, e.g. `self.apply_transaction_rename_rule(&mut transaction)` instead of passing the 3 params and the code to extract these.

relay-general/src/store/transactions/processor.rs
```diff
@@ -266,8 +334,16 @@ impl Processor for TransactionsProcessor {
...
+        self.apply_transaction_rename_rule(
+            &mut event.transaction,
+            event.transaction_info.value_mut(),
+            &transaction_source,
+        )?;
+
         // Normalize transaction names for URLs transaction sources only.
-        if event.get_transaction_source() == &TransactionSource::Url && self.normalize_names {
+        if transaction_source == TransactionSource::Url && self.normalize_names {
             normalize_transaction_name(&mut event.transaction)?;
```
iker-barriocanal: Do we still want to do this, even if we applied the transaction renaming rules? Should we not do this if we apply the transaction renaming rules, at least during internal dogfooding?

relay-server/src/actors/project.rs
iker-barriocanal, line 145: Out of scope of this PR -- in the future, we can exclude transaction name rules that are already expired.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - Parse string as number to handle a release bug. ([#1637](https://github.com/getsentry/relay/pull/1637))
 - Expand Profiling's discard reasons. ([#1661](https://github.com/getsentry/relay/pull/1661), [#1685](https://github.com/getsentry/relay/pull/1685))
 - Allow to rate limit profiles on top of transactions. ([#1681](https://github.com/getsentry/relay/pull/1681))
+- Support transaction naming rules. The rules will be currently applied to transactions where the transaction source set to `url`. ([#1695](https://github.com/getsentry/relay/pull/1695))
```
iker-barriocanal:
```suggestion
- Support transaction naming rules. ([#1695](https://github.com/getsentry/relay/pull/1695))
```
I've mentioned this somewhere in a comment below: let's not focus this PR on having the source set to `url`. This is the default use case and the only one we currently support, but that's defined in `sentry` and not in relay. In fact, if the code is a bit generalized we could already be supporting more sources out of the box.

relay-common/src/utils.rs
```diff
@@ -6,45 +6,110 @@ use regex::Regex;
...
 
-/// A simple glob matcher.
-///
-/// Supported are `?` for a single char, `*` for all but a slash and
-/// `**` to match with slashes.
-#[derive(Clone)]
-pub struct Glob {
-    value: String,
-    pattern: Regex,
+/// Glob options is used to configure the behaviour underlying regex.
```
iker-barriocanal: I find this sentence difficult to understand, and without the example below I've not been able to. I suggest the following but feel free to modify it to a different one.
```suggestion
/// Glob options represent the underlying regex emulating the globs.
```

relay-common/src/utils.rs
```diff
@@ -6,45 +6,110 @@ use regex::Regex;
...
+pub struct GlobBuilder<'g> {
+    value: &'g str,
+    opts: GlobPatternOpts<'g>,
+}
+
+impl<'g> GlobBuilder<'g> {
+    /// Create a new builder with all the captures enabled by default.
+    pub fn new(value: &'g str) -> Self {
+        let opts = GlobPatternOpts {
+            star: "([^/]*?)",
```
iker-barriocanal, line 29: Why the `?` in `star` here? The `?` tries to find as less matching groups as possible, so in `/abc/` the non-null groups are `a`, `b`, and `c`. We're interested in a matching group of `abc`, and we accomplish that by removing `?`.

Am I missing something?

relay-common/src/utils.rs
```diff
@@ -6,45 +6,110 @@ use regex::Regex;
...
+            double_star: "(.*?)",
```
iker-barriocanal, line 30: With the double star, the `?` doesn't make any difference in this case. However, do we need it? I'm really not sure if I'm missing something.

relay-common/src/utils.rs
```diff
@@ -6,45 +6,110 @@ use regex::Regex;
...
+            question_mark: "(.)",
```
iker-barriocanal, line 31: Fantastic idea of surrounding the regexes with parenthesis!

relay-common/src/utils.rs
```diff
@@ -6,45 +6,110 @@ use regex::Regex;
...
 }
 
-impl Glob {
-    /// Creates a new glob from a string.
-    pub fn new(glob: &str) -> Glob {
-        let mut pattern = String::with_capacity(glob.len() + 100);
+/// `GlobBuilder` provides the posibility to fine tune the final [`Glob`], mainly what capture
+/// groups will be enabled in the underlying regex.
+#[derive(Debug)]
+pub struct GlobBuilder<'g> {
```
iker-barriocanal, line 20: I believe all this glob builder logic belongs to a different PR -- the complexity is enough to be on its own; it's easy to make mistakes when regexes, globs and custom logic is involved; and it makes reviewing the core functionality this PR is introducing more complicated.

iker-barriocanal (commented):
Still pending review of the glob builder, the rules module, and tests. Leaving some feedback for now.

relay-common/src/utils.rs
olksdr: This regex will match everything expect `/`.

relay-common/src/utils.rs
olksdr: this also supports proper globs, when you can have the glob like `/foo/bar/**/this` there `**` matches any number of slashes and stuff in between.

relay-general/src/store/transactions/processor.rs
olksdr: if I remember correctly from our conversations, yes we want to keep this in as well.

relay-common/src/utils.rs
olksdr: There is no much logic, just a simple addition to make sure that existing or better to say, required rule application cane be done, and this happens only in `replace_captures` function. The rest is just a helper code.

relay-general/src/store/transactions/processor.rs
olksdr: It might make the code a little bit prettier in this one case but one will have to jump to different function to check what's going on. We could maybe come back to this in the followup PRs.

This also can be complicated to do, since we have few cases, if the rules applied we should remove the added `/` and save the results in to proper places, or another case when we clean up the added slash afterwards. 

If you could sketch out  how you envision the code of the `with_slashed_transaction` would look like? 

Should this function be added to `Event` protocol or keep it here and get event as the input data?

relay-server/src/actors/project.rs
olksdr: Where do you think it's better to exclude those rules? 
We still have to check expiration of the rule as close to the place where we want to apply them as possible. Since they can expire any moment, even while being in transit between different relays.

relay-general/src/store/transactions/processor.rs
olksdr: Simplified args for this function. 
Transaction does not contain transaction info - we still need it to apply changes and check the source. We could get the entire `Event` in, but then function will have access to way too much info than it requires.

relay-common/src/utils.rs
```diff
@@ -6,45 +6,110 @@ use regex::Regex;
...
-///
-/// Supported are `?` for a single char, `*` for all but a slash and
-/// `**` to match with slashes.
-#[derive(Clone)]
-pub struct Glob {
-    value: String,
-    pattern: Regex,
+/// Glob options represent the underlying regex emulating the globs.
+#[derive(Debug)]
+struct GlobPatternOpts<'g> {
```
jjbayer: nit: I would like to give this struct a more descriptive name but I don't have one. Maybe something like `GlobPatternGroups` or `GlobPatternBuildingBlocks`?

relay-common/src/utils.rs
jjbayer: Note that these three patterns already exist on master branch, they were just copied to a different location: https://github.com/getsentry/relay/blob/6e2f7aeacd0c3ee3b0f4f959556e9ecd963802d7/relay-common/src/utils.rs#L33-L35

relay-common/src/utils.rs
```diff
@@ -167,6 +260,29 @@ mod tests {
...
+
+        assert_eq!(
+            g.replace_captures("/foo/some/bar/here/store", "*"),
+            "/foo/*/bar/here/store"
+        );
+        assert_eq!(g.replace_captures("/foo/testing/bar/", "*"), "/foo/*/bar/");
+        assert_eq!(
+            g.replace_captures("/foo/testing/1/", "*"),
+            "/foo/testing/1/"
+        );
```
jjbayer: Should we add some more test coverage for `replace_captures`? We could do something like

```rust
for (pattern, star, double_star, question_mark, expected_result) in [
    ("**", false, true, false, "*"),
    // ...
] {
    let g = Glob::builder(pattern)
            .capture_star(star)
            .capture_double_star(double_star)
            .capture_question_mark(question_mark)
            .build();
    // test
}
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -1,21 +1,95 @@
...
+                    .and_then(|i| i.source.value().map(|s| s == source))
+                    .unwrap_or_default()
+            };
+
+            let rule = self.tx_name_rules.iter().find(|rule| {
+                source_match(&rule.scope.source)
+                    && rule.expiry > now
+                    // Adding `/` at the end of the name, ensures that rules like /<something>/*/**
+                    // will always match the string.
+                    && rule.pattern.is_match(transaction)
```
jjbayer: How about we move this logic into a `fn matches()` in `impl TransactionNameRule`?

relay-general/src/store/transactions/rules.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Note: currently only `url` source for rules supported.
+    pub fn apply(&self, value: &str) -> Option<String> {
+        match &self.redaction {
+            RedactionRule::Replace(Replace { substitution }) => {
+                let result = self.pattern.replace_captures(value, substitution);
+                Some(result)
+            }
+            _ => {
+                relay_log::trace!("Replacement rule type is unsupported!");
+                None
```
jjbayer: The only reason this can return `None` is because of the `#[serde(other)]` option on `RedactionRule`, right? I would instead change the signature to `-> String`, and return `value.to_owned()` here.

Ideally, we would remove unsupported replacement rules on deserialization. Then this branch practically becomes dead code.

relay-general/src/store/transactions/rules.rs
```diff
@@ -0,0 +1,195 @@
...
+    #[serde(default = "default_substitution")]
+    substitution: String,
+}
+
+/// Describes what to do with the matched pattern.
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+#[serde(tag = "method", rename_all = "snake_case")]
+pub enum RedactionRule {
+    Replace(Replace),
+    #[serde(other, skip_serializing)]
```
jjbayer:
```suggestion
    #[serde(other)]
```

If we skip serializing, an internal Relay passes on this config to a downstream Relay which then assumes the default substitution, which is not what we want.

relay-general/src/store/transactions/rules.rs
jjbayer: Or, if we add a `matches` function as suggested below, that could automatically return false for any unsupported rule.

olksdr:
@jjbayer @iker-barriocanal  added more tests, reworked some functions to get your review comments in.  PTAL!

relay-general/src/store/transactions/processor.rs
iker-barriocanal: My thought was to leave it here, but adding it to the `Event` might not be a bad idea either. My idea is the following, where `closure` is a function with the logic below:

```
def with_slashed_transaction(tx, closure):
  if slash_is_present(tx):
      closure()
  else:
      add_slash(tx)
      closure()
      remove_slash(tx)
```

relay-general/src/store/transactions/processor.rs
iker-barriocanal: > function will have access to way too much info than it requires

What's the problem here? Is it that `Event` would be a mutable reference, and we don't want to make the whole `Event` mutable?

relay-server/src/actors/project.rs
iker-barriocanal: I was specific enough --  I was referring to the serialization part. Before checking if the vector is empty, discard expired rules and then check for the vector's emptiness. Especially at the beginning, this won't make much of a difference, since we're setting 90d TTLs.

relay-general/src/store/transactions/rules.rs
iker-barriocanal: I personally would not set any defaults. If relay can't deserialize a rule because it doesn't know about it, then ignore the rule.

iker-barriocanal (approved):

relay-general/src/store/transactions/processor.rs
```diff
@@ -53,28 +50,20 @@ impl<'r> TransactionsProcessor<'r> {
...
 
-            if let Some((rule, Some(mut result))) = result {
-                if !slash_is_present {
-                    transaction.pop();
-                    result.pop();
-                }
+            if let Some((rule, mut result)) = result {
                 if &result != transaction {
+                    if !slash_is_present {
+                        transaction.pop();
```
jjbayer: I think we should copy these two lines to line 59, that is, re-remove the added slash even if no rule matches at all. In other words, the original, unmodified transaction name should always survive. Maybe it's safer to use a `Cow` for the transaction name,  and evaluate the rule matching / replacement on a copy when `slash_is_present`.

jjbayer (commented):
Looking good, just one more note on the `slash_is_present` logic.

relay-general/src/store/transactions/processor.rs
olksdr: Currently the slash is always removed if it was added, on the line `77` is the check, even if the rule is not matching. 

I was thinking to copy the string and use it for the checks and manipulations - but I also want to avoid unnecessary copies if we can avoid it.

But I see that the code now a bit longer, and keeping code more compact would help here. I'll look into this.

jjbayer: Sorry, I missed that. Yeah, to make this logic more resilient against future refactors it would make sense to make it more compact. Either by using a `Cow` or by something like a context manager as @iker-barriocanal suggested. The rust equivalent to a context manager would be the [RAII](https://doc.rust-lang.org/rust-by-example/scope/raii.html) pattern.

jjbayer (approved):

relay-general/src/store/transactions/processor.rs
olksdr: @jjbayer  have a look into this commit [8199c5c](https://github.com/getsentry/relay/pull/1695/commits/8199c5caebfcf27b5d8fb11c1650bd8a4338334a), I tried to use `Cow` and hide the checks in the rule impl:
* it doesn't change the transaction
* it uses owned data only if needed
* and also combined the check and apply into one function

relay-general/src/store/transactions/rules.rs
```diff
@@ -74,6 +76,32 @@ pub struct TransactionNameRule {
...
+    pub fn match_and_apply(
+        &self,
+        mut transaction: Cow<String>,
+        transaction_info: &TransactionInfo,
+    ) -> Option<String> {
+        let slash_is_present = transaction
+            .chars()
+            .last()
+            .map(|c| c == '/')
+            .unwrap_or_default();
```
jjbayer: btw found https://doc.rust-lang.org/std/primitive.str.html#method.ends_with

jjbayer (approved):
nice!

relay-general/src/store/transactions/rules.rs
olksdr: shame on me 🤦 

fixed in [33fed75](https://github.com/getsentry/relay/pull/1695/commits/33fed75ac29fbdb45a3b72b824f223bd6aeb9172)

PR #1697, opened 2022-12-15
revert(profiling): Remove concurrent profiling

phacops:
This piece of code was duplicating profiles and trimming them to the right duration for each transaction they were associated with.

As we want to be able to only accept profiles associated with a transaction, we have to send the profile in the same envelope as transaction so dynamic sampling works properly. We can't start multiple profilers at the time (Android API we're using doesn't allow that) or we can't do the processing the server was doing to get one profile and trim on the client. We will keep trimming `sample` and `cocoa` formats.

Therefore, we decided to remove this concurrent profiling support.


relay-profiling/src/android.rs
```diff
@@ -141,83 +133,23 @@ pub fn expand_android_profile(payload: &[u8]) -> Result<Vec<Vec<u8>>, ProfileErr
...
-                        Duration::from_nanos(transaction.relative_start_ns)
-                    }
-                };
-                event.time = substract_to_timestamp(event.time, relative_start);
-                true
-            } else {
-                false
-            }
-        });
+    let transaction = &profile.transactions[0];
```
jan-auer: This can panic if `profile.transactions` is empty but `has_transaction_metadata` returns false.

relay-profiling/src/android.rs
```diff
@@ -141,83 +133,23 @@ pub fn expand_android_profile(payload: &[u8]) -> Result<Vec<Vec<u8>>, ProfileErr
...
-            cpu: cpu
-                .checked_sub(duration)
-                .or_else(|| Some(Duration::default())),
-            wall: wall
-                .checked_sub(duration)
-                .or_else(|| Some(Duration::default())),
-        },
-        _ => unimplemented!(),
-    }
+    Ok(items)
```
jan-auer: This function now always returns a vector with exactly one element. This can be simplified.

jan-auer (commented):
This changes android profile parsing, yet there's still unlimited expansion in `expand_cocoa_profile` and `expand_sample_profile`. Shouldn't we remove that, too?

relay-profiling/src/android.rs
jan-auer: Fixed in 0c109eb324e50c7e087ce9d1365a2046a76fd535. @phacops could you double-check that the logic was even correct? It seems to me as if the _intent_ of this was rather to check `if transactions.is_empty() || profile.has_tx() { return }`

jan-auer:
`test_expand_multiple_transactions` is still failing now. It asserts that `events` of the profile have been filtered, but expansion no longer does that. @phacops please confirm if that can be changed in the test.

relay-profiling/src/android.rs
phacops: I'll remove this when we'll remove the expansion for all formats.

relay-profiling/src/android.rs
phacops: The logic was correct. This was to provide compatibility with the previous format where we'd receive transaction metadata as top level fields. If we had an empty `transactions` array and top level fields (as previous Android SDKs were sending the profile), we can just encode the payload and move on.

jan-auer (approved):

Zylphrex (approved):

relay-profiling/src/android.rs
```diff
@@ -126,97 +126,46 @@ impl AndroidProfile {
...
+            return Err(ProfileError::InvalidTransactionMetadata);
         }
 
-        match serde_json::to_vec(&new_profile) {
-            Ok(payload) => items.push(payload),
-            Err(_) => {
-                return Err(ProfileError::CannotSerializePayload);
-            }
-        };
+        profile.transaction = Some(transaction.clone());
```
jan-auer: If you move this statement down below all the other assignments, you can remove this clone. You will just have to clone the transaction name below.

relay-profiling/src/android.rs
```diff
@@ -126,97 +126,46 @@ impl AndroidProfile {
...
-            wall: wall
-                .checked_sub(duration)
-                .or_else(|| Some(Duration::default())),
-        },
-        _ => unimplemented!(),
+pub fn parse_android_profile(payload: &[u8]) -> Result<Vec<u8>, ProfileError> {
+    let profile = parse_profile(payload)?;
+    match serde_json::to_vec(&profile) {
+        Ok(payload) => Ok(payload),
+        Err(_) => Err(ProfileError::CannotSerializePayload),
```
jan-auer: This `match` is essentially a manual version of this: 
```rust
serde_json::to_vec(&profile)
    .map_err(|_| ProfileError::CannotSerializePayload)
```

relay-profiling/src/android.rs
```diff
@@ -126,97 +126,46 @@ impl AndroidProfile {
...
+        // this is for compatibility
+        profile.active_thread_id = transaction.active_thread_id;
+        profile.duration_ns = transaction.duration_ns();
+        profile.trace_id = transaction.trace_id;
+        profile.transaction_id = transaction.id;
+        profile.transaction_name = transaction.name;
+    }
+
+    if !profile.sampled_profile.is_empty() {
+        profile.parse()?;
```
jan-auer, line 152: Does this `parse()` and subsequent removal of events have to happen after setting the transaction? If not, I would turn this around to avoid a possibly redundant operation.

relay-profiling/src/android.rs
```diff
@@ -126,97 +126,46 @@ impl AndroidProfile {
...
-            Err(_) => {
-                return Err(ProfileError::CannotSerializePayload);
-            }
-        };
+fn parse_profile(payload: &[u8]) -> Result<AndroidProfile, ProfileError> {
+    let mut profile: AndroidProfile =
+        serde_json::from_slice(payload).map_err(ProfileError::InvalidJson)?;
 
-        return Ok(items);
+    if profile.transactions.is_empty() && !profile.has_transaction_metadata() {
```
jan-auer: This can be joined with the next `if` to simplify control flow:

```rust
if let Some(transaction) = profile.transactions.drain(..).next() {
    // ...
} else if !profile.has_transaction_metadata() {
    return Err(ProfileError::NoTransactionAssociated);
}
```

jan-auer (commented):
Preliminary review of Android.

relay-server/src/actors/processor.rs
```diff
@@ -967,35 +967,46 @@ impl EnvelopeProcessorService {
...
         }
 
-        let context = &state.envelope_context;
-        let mut new_profiles = Vec::new();
-
-        while let Some(item) = envelope.take_item_by(|item| item.ty() == &ItemType::Profile) {
-            match relay_profiling::expand_profile(&item.payload()[..]) {
-                Ok(payloads) => new_profiles.extend(payloads),
+        envelope.retain_items(|item| match item.ty() {
+            ItemType::Profile => match relay_profiling::expand_profile(&item.payload()[..]) {
```
jan-auer:
```suggestion
            ItemType::Profile => match relay_profiling::expand_profile(&item.payload()) {
```

relay-profiling/src/android.rs
phacops: Moving the statement lower always gives me this and I'm not certain why.
```
149 |         profile.transaction = Some(transaction);
    |                                    ^^^^^^^^^^^ value used here after partial move
```

relay-server/src/actors/processor.rs
```diff
@@ -967,35 +967,46 @@ impl EnvelopeProcessorService {
...
+                Ok(payload) => {
+                    if payload.len() <= self.config.max_profile_size() {
+                        item.set_payload(ContentType::Json, &payload[..]);
+                        true
+                    } else {
+                        let context = &state.envelope_context;
+                        context.track_outcome(
```
jan-auer: Here and below:

```suggestion
                        state.envelope_context.track_outcome(
```

relay-profiling/src/cocoa.rs
```diff
@@ -139,29 +132,35 @@ impl CocoaProfile {
...
+    }
+
+    profile.remove_single_samples_per_thread();
 
-        return Ok(items);
+    if profile.sampled_profile.samples.is_empty() {
+        return Err(ProfileError::NotEnoughSamples);
+    }
+
+    if profile.transactions.is_empty() && profile.has_transaction_metadata() {
```
jan-auer: This logic here is different than in the android case. In android case, this early return doesn't exist. Is that intentional?

relay-profiling/src/android.rs
jan-auer: Turns out this is not possible since the mutable borrow lives for the entire statement including the `else` branch :/

relay-profiling/src/cocoa.rs
phacops: Yes, it's intentional because in the Android case, we still need to go through the `parse` logic to extract an `AndroidTrace` even if the transaction is set on top level fields.

relay-profiling/src/android.rs
phacops: This is just the heaviest operation you can think of in terms of parsing compared to just validating a few things first.

jan-auer (approved):

Issue #1689, opened 2022-12-12
Use tokio::time::Instant in all services


jan-auer:
Using tokio's `Instant` allows us to run tests more efficiently and mock timers. Otherwise, in release builds it's just a wrapper around the native `std::time::Instant`.

jjbayer:
After replacing the `Instant`, add tokio `pause` to any relay tests that currently use `sleep`.

PR #1698, opened 2022-12-15
ref(server): Change std::time::instant to tokio::time::instant

TBS1996:
Changing form `std::time::instant` to `tokio::time::instant` allows us to use convenient methods while making unit tests such as instantly mocking a time delay. 

references #1689 

#skip-changelog


jan-auer (approved):
Please move the imports into the group with other external imports.

iker-barriocanal (approved):

PR #1701, opened 2022-12-15
fix: Properly escape generic type in docstring

emmatyping:
This fixes two errors in the documentation builds of `relay-server`, which fail due to a generic type not being escaped.

This is currently blocking the self-hosted release today.

_#skip-changelog_


hubertdeng123 (approved):

iker-barriocanal (approved):

chadwhitacre (approved):

chadwhitacre:
Force-merging to unblock [self-hosted release](https://github.com/getsentry/publish/issues/1666).

PR #1702, opened 2022-12-15
feat(replays): Leave non-critical fields as generic "Value" types

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/239

Leaves non-critical fields as generic `Value` types which can be re-serialized later without consequence to PII scrubbing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Apply dynamic sampling to transactions from older SDKs and even in case Relay cannot load project information. This avoids accidentally storing 100% of transactions. ([#1667](https://github.com/getsentry/relay/pull/1667))
 - Replay recording parser now uses the entire body rather than a subset. ([#1682](https://github.com/getsentry/relay/pull/1682))
 - Fix a potential OOM in the Replay recording parser. ([#1691](https://github.com/getsentry/relay/pull/1691))
+- Fix type error in replay recording parser. ([#1702](https://github.com/getsentry/relay/pull/1702))
 
 **Internal**:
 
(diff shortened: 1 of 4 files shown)


iker-barriocanal (commented):
Aren't we losing the type safety by setting everything as `Value`? What's the problem we're facing here?

cmanallen:
@iker-barriocanal Type safety is not a requirement.  We have no desire to validate the payload.  We're parsing and stripping PII.  If we could guarantee no PII then we would not parse this at all.

cmanallen:
@iker-barriocanal The problem we're facing is that these values may not be well typed in their implementation.  It's better to just accept values we don't care about and let our javascript implementation sort out the quirks.

jan-auer:
@cmanallen is there a desired target type for all these fields? For example, SDKs might send an identifier as either string or number, but you always stringify it for your backend (can follow up with an example). That way, you reduce the burden on all other systems that touch this data to handle all permutations or edge-cases. We noticed with the event protocol in the past that leaving completely untyped and not enforcing constraints early on leads to much added complexity down the road.

If you need an urgent bugfix, we can still merge this. A conversation on the role of normalization and schemas will definitely take a little more time.

cmanallen:
@jan-auer

> is there a desired target type for all these fields?

No.  The contract is maintained between the recording component and the player component.  Both are open source projects outside the control of Sentry.  The rrweb maintainers bear the burden of their schema definitions.

jan-auer:
In that case it would probably be better if we define no schema at all and use a plain `Annotated<Value>` instead. You can still run a generic scrubber on the basic types (object, string, int) without further knowledge of any schema.

cmanallen:
Let's leave the schema in place for now.  I'll review the generic schema implementation independently.

olksdr (approved):

PR #1703, opened 2022-12-15
ci: Increase timeout for tests

emmatyping:
It seems that these tests are flaking due to getting canceled as they exceed the 15 minute mark. Ideally tests would be sped up, but to avoid currently failing CI, I think this change makes the most sense.

_#skip-changelog_


jan-auer (approved):
We should investigate why those tests are so slow and how we can speed them up, though. 20 minutes is unacceptable.

PR #1707, opened 2022-12-19
fix(crash): Use main SDK to send crash reports

jan-auer:
Relay captures crashes with the Sentry Native SDK and its breakpad backend. By
default, this uses `curl` as transport. In our Docker build, we dynamically link
with the installed development library. Since the build runs on CentOS 7, this
is an old library lacking many security features and patches. Even though the
production container has a more recent libcurl installed, the Native SDK
uploader fails the SSL handshake with sentry.io.

This PR removes the dependency on `curl` entirely and replaces it with a custom
transport function. We parse the envelope and send it using the main Sentry SDK.

### DIFF
--- a/relay-crash/src/lib.rs
+++ b/relay-crash/src/lib.rs
@@ -2,6 +2,7 @@
 //!
 //! Use [`CrashHandler`] to configure and install a crash handler.
 
+use std::fmt;
 use std::path::Path;
 
 #[cfg(unix)]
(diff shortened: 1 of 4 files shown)


relay-crash/src/lib.rs
```diff
@@ -13,19 +14,63 @@ mod native {
...
 
+/// A transport function used to send envelopes to Sentry.
+pub type Transport = fn(envelope: &[u8]);
+
+/// Serializes a sentry_native envelope and passes the buffer to the [`Transport`] function.
+unsafe extern "C" fn transport_proxy(
+    envelope: *const native::sentry_envelope_s,
+    tx_pointer: *mut std::ffi::c_void,
+) {
+    if envelope.is_null() || tx_pointer.is_null() {
```
jan-auer, line 26: This is overly defensive, since the Native SDK will not send a null pointer for the envelope, and the transport pointer is set below.

relay-crash/src/lib.rs
```diff
@@ -13,19 +14,63 @@ mod native {
...
     database: &'a str,
+    transport: Option<Transport>,
     release: Option<&'a str>,
     environment: Option<&'a str>,
 }
 
+impl<'a> fmt::Debug for CrashHandler<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let transport = match self.transport {
+            Some(_) => "Some(fn)",
```
jan-auer, line 61: `fn(&[u8])` does not implement `Debug`, hence we have to do it ourselves. We could also wrap the transport in a custom type and implement `Debug` just for that, but I found this to be less code.

relay-crash/src/lib.rs
```diff
@@ -13,19 +14,63 @@ mod native {
     include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 }
 
+/// A transport function used to send envelopes to Sentry.
+pub type Transport = fn(envelope: &[u8]);
+
+/// Serializes a sentry_native envelope and passes the buffer to the [`Transport`] function.
+unsafe extern "C" fn transport_proxy(
```
jan-auer, line 22: We do not have access to the Sentry SDK in this crate, so instead we provide a hookable callback and inject Sentry-related logic outside.

Swatinem (approved):
lgtm

jan-auer:
I've verified that the crash reporter works and can pass the envelope to the callback. However, the Rust SDK cannot parse the native Envelope since it has a too strict `Event` protocol implemented that it parses eagerly. There are now two options:
1. Make a change to the Rust SDK
2. Send the Envelope manually

jan-auer:
I would like to move forward with this PR even though it will not successfully send the crash to Sentry. This will need to be fixed in `sentry-rust`, and then we need to upgrade to the latest version, which can be a separate PR.

olksdr (approved):

Issue #1603, opened 2022-11-21
Update ProjectLocal Actor


HazAT:
- [ ] https://github.com/getsentry/relay/pull/1710

PR #1710, opened 2022-12-20
ref: Convert project local actor to tokio

jjbayer:
- Replace `ProjectLocalSource` by a `ProjectLocalSourceService` that accepts a `FetchOptionalProjectState` message.
- Replace the internally used `UpdateLocalStates` by an `mpsc::channel`.

https://github.com/getsentry/relay/issues/1603

#skip-changelog


relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,108 @@
...
+use relay_system::Sender;
+use relay_system::Service;
 
 use crate::actors::project::ProjectState;
 use crate::actors::project_cache::FetchOptionalProjectState;
 
+/// Service interface of the local project source.
+#[derive(Debug)]
+pub enum LocalProjectSource {
+    /// TODO: docs
```
jjbayer:
```suggestion
    /// Fetches the project state for a given project key. Returns `None` if the project state is not available locally.
```

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,108 @@
...
+    type Response = AsyncResponse<Option<Arc<ProjectState>>>;
+    fn from_message(
+        message: FetchOptionalProjectState,
+        sender: Sender<Option<Arc<ProjectState>>>,
+    ) -> Self {
+        Self::FetchOptionalProjectState(message, sender)
+    }
+}
+
+/// TODO: docs
```
jjbayer:
```suggestion
/// A service which periodically loads project states from disk.
```

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,108 @@
...
+use relay_system::Service;
 
 use crate::actors::project::ProjectState;
 use crate::actors::project_cache::FetchOptionalProjectState;
 
+/// Service interface of the local project source.
+#[derive(Debug)]
+pub enum LocalProjectSource {
+    /// Fetches the project state for a given project key. Returns `None` if the project state is not available locally.
+    FetchOptionalProjectState(FetchOptionalProjectState, Sender<Option<Arc<ProjectState>>>),
```
jan-auer: Since there's only a single message, you can make this a `struct` instead of an enum. See the second example here:
https://getsentry.github.io/relay/relay_system/trait.Interface.html#implementating-interfaces

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Remove concurrent profiling. ([#1697](https://github.com/getsentry/relay/pull/1697))
 - Use the main Sentry SDK to submit crash reports instead of a custom curl-based backend. This removes a dependency on `libcurl` and ensures compliance with latest TLS standards for crash uploads. Note that this only affects Relay if the hidden `_crash_db` option is used. ([#1707](https://github.com/getsentry/relay/pull/1707))
 - Support transaction naming rules. ([#1695](https://github.com/getsentry/relay/pull/1695))
+- Convert project local actor to tokio. ([#1710](https://github.com/getsentry/relay/pull/1710))
```
jan-auer: So far we haven't included these in the changelog.

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,108 @@
...
-impl Message for UpdateLocalStates {
-    type Result = ();
-}
+            loop {
+                tokio::select! {
+                    biased;
 
-impl Handler<UpdateLocalStates> for LocalProjectSource {
-    type Result = ();
+                    _ = ticker.tick() => self.poll_local_states().await,
```
jan-auer: This used to run in a background thread and didn't block the main loop. Are we surely fine with moving this in?

relay-server/src/actors/project_local.rs
jjbayer: I considered it, but I wasn't sure how to name these types then. Would I call the `Interface` `FetchOptionalProjectState`, and rename `LocalProjectSourceService` back to `LocalProjectSource`?

relay-server/src/actors/project_local.rs
jjbayer: Good point, we actually want to run this in the background, not `await` it. Should I just do a `tokio::spawn` here?

jan-auer: I think you could just do a completely separate `tokio::spawn` that constantly loops and sleeps. As long as there is no blocking operations (i.e. it either uses `tokio::fs` or it uses `spawn_blocking`), it's sufficiently isolated.

relay-server/src/actors/project_local.rs
jan-auer: You can keep the interface on `LocalProjectSource` since you need the sender:

```rust
pub struct LocalProjectSource(FetchOptionalProjectState, Sender<Option<Arc<ProjectState>>>);

impl Interface for LocalProjectSource {}
```

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,104 @@
...
-}
+impl Service for LocalProjectSourceService {
+    type Interface = LocalProjectSource;
+
+    /// Spawns two loops:
+    /// 1. one for periodically polling local states, and
+    /// 2. one to handle both external messages _and_ updates to the local cache.
+    fn spawn_handler(mut self, mut rx: relay_system::Receiver<Self::Interface>) {
+        let project_path = self.config.project_configs_path();
+        let (mut state_tx, mut state_rx) = mpsc::channel(1);
```
jjbayer: There is no need to keep more than one item in the channel. Ideally, we would even overwrite previous items. So maybe we should use an `Arc<Mutex>` instead of a channel?

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,104 @@
...
+            relay_log::info!("project local cache started");
+            loop {
+                tokio::select! {
+                    Some(message) = rx.recv() => self.handle_message(message),
+                    Some(states) = state_rx.recv() => self.local_states = states,
 
-    fn handle(&mut self, message: UpdateLocalStates, _context: &mut Context<Self>) -> Self::Result {
-        self.local_states = message.states;
+                    else => break,
+                }
```
jjbayer: Did not add a bias here, because it's unclear to me what branch should be prioritized.

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,109 @@
...
-            // Block entire actor on first local state read, such that we don't e.g. drop events on
-            // startup
-            .wait(context);
-    }
 
-    fn stopped(&mut self, _ctx: &mut Self::Context) {
-        relay_log::info!("project local cache stopped");
+    fn handle_fetch_optional_project_state(
+        &self,
+        message: FetchOptionalProjectState,
```
olksdr: Could we instead of the message, just provide the `project_key`? 

There is already a `handle_message` function, which kind of just forwards the message again to another function.  What the reason for this?  
I would either see all this in one function (since there is no logic and nothing really happens), or really then isolate this function and accept parameters which are only required for this function to preform  the required operation.

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,109 @@
...
-use futures01::{sync::oneshot, Future};
 
 use relay_common::{ProjectId, ProjectKey};
 use relay_config::Config;
 use relay_log::LogError;
+use relay_system::AsyncResponse;
+use relay_system::FromMessage;
+use relay_system::Interface;
+use relay_system::Sender;
+use relay_system::Service;
```
olksdr: You might want merge all these imports here.

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,109 @@
...
-struct UpdateLocalStates {
-    states: HashMap<ProjectKey, Arc<ProjectState>>,
-}
+impl Service for LocalProjectSourceService {
+    type Interface = LocalProjectSource;
+
+    /// Spawns two loops:
+    /// 1. one for periodically polling local states, and
+    /// 2. one to handle both external messages _and_ updates to the local cache.
+    fn spawn_handler(mut self, mut rx: relay_system::Receiver<Self::Interface>) {
```
olksdr: you could also import `relay_system::Receiver`, since all the other stuff is already imported.

relay-server/src/actors/project_local.rs
```diff
@@ -115,9 +144,19 @@ fn load_local_states(projects_path: &Path) -> io::Result<HashMap<ProjectKey, Arc
...
+            path: std::path::PathBuf,
+        ) -> tokio::io::Result<(std::path::PathBuf, ProjectState)> {
+            let file = std::fs::File::open(&path)?;
+            let reader = std::io::BufReader::new(file);
+            Ok((path, serde_json::from_reader(reader)?))
+        }
 
+        // serde_json is not async, so spawn a blocking task here:
+        let handle = tokio::task::spawn_blocking(move || parse_file(path));
+        let (path, state) = handle.await??;
```
olksdr: I would simply inline this 
```suggestion
        let (path, state) = tokio::task::spawn_blocking(move || parse_file(path)).await??;
       
```

relay-server/src/actors/project_local.rs
```diff
@@ -115,9 +144,19 @@ fn load_local_states(projects_path: &Path) -> io::Result<HashMap<ProjectKey, Arc
             continue;
         }
 
-        let state = serde_json::from_reader(io::BufReader::new(fs::File::open(&path)?))?;
-        let mut sanitized = ProjectState::sanitize(state);
+        fn parse_file(
```
olksdr: It could be better to refactor this to outside standalone function.

olksdr: It could be better to refactor this to outside standalone function.

relay-server/src/actors/project_local.rs
```diff
@@ -1,81 +1,109 @@
...
+            // populated.
+            poll_local_states(&project_path_copy, &mut state_tx_copy).await;
 
-impl Handler<UpdateLocalStates> for LocalProjectSource {
-    type Result = ();
+            relay_log::info!("project local cache started");
+            loop {
+                tokio::select! {
+                    Some(message) = rx.recv() => self.handle_message(message),
+                    Some(states) = state_rx.recv() => self.local_states = states,
```
olksdr: Why do you spawn 2 tasks with 2 loops here?  Maybe there is a reason for this I cannot see yet.

See `AggregatorService`: https://github.com/getsentry/relay/blob/82ed865de41159b2114c3805de177e4e27696e7b/relay-metrics/src/aggregation.rs#L1985-L1996 
where I also used ticker and everything kind of fit together in one loop. 

And if you need to spawn , do it in the `poll_local_states`.  

This will make the `spawn_handler` as the entry point for the service much smaller and easier to understand.

relay-server/src/actors/project_local.rs
olksdr: We might want to handle incoming message asap, and the rest can wait a bit.

relay-server/src/actors/project_local.rs
```diff
@@ -136,31 +175,18 @@ fn load_local_states(projects_path: &Path) -> io::Result<HashMap<ProjectKey, Arc
...
-
-            thread::sleep(config.local_cache_interval());
+async fn poll_local_states(
+    path: &Path,
+    tx: &mut mpsc::Sender<HashMap<ProjectKey, Arc<ProjectState>>>,
+) {
+    let states = load_local_states(path).await;
+    match states {
+        Ok(states) => {
+            let _ = tx.send(states).await;
```
olksdr: Should the error be handled here which can happen when the channel is closed? Or maybe just log it?  This also comes to other instances of `send` where  (if) the result is returned.

olksdr (commented):
Overall it looks good to me! 

I left few comments, which we can also discuss. Let me know what you think?

relay-server/src/actors/project_local.rs
```diff
@@ -136,31 +144,42 @@ fn load_local_states(projects_path: &Path) -> io::Result<HashMap<ProjectKey, Arc
...
+            "failed to load static project configs: {}",
+            LogError(&error)
+        ),
+    };
+}
 
-    let _ = thread::spawn(move || {
-        let path = config.project_configs_path();
-        let mut sender = Some(sender);
+async fn spawn_poll_local_states(
```
olksdr, line 135: only one thing I this, should this function be moved to the top, before it used first time? 

and same for `poll_local_states`

olksdr (approved):

olksdr (approved):

relay-server/src/actors/project_local.rs
jjbayer: It might make the diff harder to read, but yes, absolutely.

PR #1713, opened 2022-12-20
ref(utils): Add unit test for event and trace rules

TBS1996:
Removes a python integration tests that checks if trace rules are applied after event rules, and replaces it with unit tests in rust that targets the specific functions more directly.

#skip-changelog

### DIFF
--- a/relay-server/src/utils/dynamic_sampling.rs
+++ b/relay-server/src/utils/dynamic_sampling.rs
@@ -284,8 +284,7 @@ mod tests {
         })
     }
 
-    ///
-    ///
+    /// checks if
     #[test]
     fn test_trace_rules_applied_after_event_rules() {
         // a transaction rule that drops everything
--- a/relay-server/src/utils/dynamic_sampling.rs
+++ b/relay-server/src/utils/dynamic_sampling.rs
@@ -303,37 +303,37 @@ mod tests {
         // a trace rule that keeps everything
         let trace_state = state_with_rule(Some(1.0), RuleType::Trace, SamplingMode::Received);
 
-        let matching_envelope = new_envelope(true, "healthcheck");
-        let non_matching_envelope = new_envelope(true, "test1");
+        let healthcheck_envelope = new_envelope(true, "healthcheck");
+        let other_envelope = new_envelope(true, "test1");
 
-        let matching_event = Event {
+        let healthcheck_event = Event {
             id: Annotated::new(EventId::new()),
             ty: Annotated::new(EventType::Transaction),
             transaction: Annotated::new("healthcheck".to_owned()),
             ..Event::default()
         };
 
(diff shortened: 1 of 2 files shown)


relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -254,6 +260,112 @@ mod tests {
...
+                value: "healthcheck".into(),
+                options: Default::default(),
+            }),
+        );
+
+        // a trace rule that keeps everything
+        let trace_state = state_with_rule(Some(1.0), RuleType::Trace, SamplingMode::Received);
+
+        let matching_envelope = new_envelope(true, "healthcheck");
+        let non_matching_envelope = new_envelope(true, "test1");
```
jjbayer: `matching` and `non_matching` are ambiguous, because both envelopes match exactly one of the two rules. So I would rename to something like:

```suggestion
        let healthcheck_envelope = new_envelope(true, "healthcheck");
        let other_envelope = new_envelope(true, "test1");
```

Same goes for the events below.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -254,6 +260,112 @@ mod tests {
...
+        };
+
+        let non_matching_event = Event {
+            id: Annotated::new(EventId::new()),
+            ty: Annotated::new(EventType::Transaction),
+            transaction: Annotated::new("test1".to_owned()),
+            ..Event::default()
+        };
+
+        // if it matches, the transaction should be dropped
```
```suggestion
        // if it matches the transaction rule, the transaction should be dropped
```

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -254,6 +260,112 @@ mod tests {
...
+        let should_drop = should_keep_event(
+            matching_envelope.dsc(),
+            Some(&matching_event),
+            None,
+            &event_state,
+            Some(&trace_state),
+            false,
+        );
+
+        // if it doesn't match, the transaction should be kept
```
```suggestion
        // if it doesn't match the transaction rule, the transaction should be kept
```

relay-server/src/utils/dynamic_sampling.rs
TBS1996: good idea, I've changed it in my latest commit

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -254,6 +260,112 @@ mod tests {
...
+
+        state_with_config(SamplingConfig {
+            rules,
+            mode,
+            next_id: None,
+        })
+    }
+
+    ///
+    ///
```
jjbayer: Let's either remove these, or better yet, write a short description of what the test does.

jjbayer (approved):
See note about a comment, apart from that, this seems good to go!

PR #1714, opened 2022-12-21
ref(actix): Migrate RedisProjectSource from actor to utility class

### LINKED ISSUE #1604: Update ProjectRedis Actor

olksdr:
Instead of using the `actix::Actor` and instead of migrating this to our new `Service` this change introduces the simple utility class `RedisProjectSource` which just provides only one helper method to fetch project config. 

The call to this method is wrapped into `tokio::task::spawn_blocking` which should be executed on the separated thread pool.  And the rest of the functionality stays practically the same. 

fix: #1604 

#skip-changelog


jjbayer (approved):
LGTM, nice simplification!

PR #1715, opened 2022-12-21
ref(actix): Migrate UpstreamRelay to tokio

jan-auer:
Migrates the `UpstreamRelay` actor to a tokio-based `Service`. The internals of
the upstream are separated into components:
- `SharedClient`: Abstraction for sending HTTP requests and handling responses.
- `ConnectionMonitor`: Maintain network outage state and
  reestablish a connection when down.
- `AuthMonitor`: Maintain authentication state and handle
  re-authentication.
- `Broker`: Subscribes to updates using the two monitors, waits for queued
  requests from the inbound channel, and then uses `SharedClient` to send them.

#skip-changelog


jan-auer:
Update on the current state:
 - All functionality has been ported, though there needs to be one more walkthrough of the old actor to verify this.
 - Both the outage monitor and outage handling are not yet working correctly. There are multiple TODOs on this.
 - The interface through which upstream queries are constructed and handled is not ideal.
 - Serialization of queries happens eagerly on the main thread. This should be moved to a later point

I am going to spend a little time over the next days on this on the side, so feel free to keep this assigned to me.

relay-server/src/actors/upstream.rs
```diff
@@ -1,58 +1,104 @@
...
-//!   * messages that do no use Relay authentication
-//!
-//!     These are messages for requests that originate as user sent events and use whatever
-//!     authentication headers were provided by the original request.
-//!
-//!  * messages used internally by Relay
-//!
-//!    These are messages that Relay sends in order to coordinate its work and do not result
-//!    directly in a HTTP message being send to the upstream server.
-//!
```
olksdr: We used to have nice docs here. It would be great to get them back in refreshed form

relay-server/src/actors/upstream.rs
```diff
@@ -1,58 +1,104 @@
...
 use std::sync::Arc;
 
-use ::actix::fut;
-use ::actix::prelude::*;
-use actix_web::http::{header, Method};
-use futures01::{future, prelude::*, sync::oneshot};
 use itertools::Itertools;
-use serde::de::DeserializeOwned;
-use serde::ser::Serialize;
+use relay_log::LogError;
```
olksdr: Could you, please, move `relay_log`  down to all relay imports?

relay-server/src/actors/upstream.rs
```diff
@@ -66,9 +112,25 @@ pub enum UpstreamRequestError {
...
     ChannelClosed,
+
+    #[error("upstream permanently denied authentication")]
+    AuthDenied,
 }
 
 impl UpstreamRequestError {
+    /// Returns the status code of the HTTP request sent to the upstream.
+    ///
+    /// If this error is the result of sending a request to the upstream, this method returns `Sone`
```
olksdr:
```suggestion
    /// If this error is the result of sending a request to the upstream, this method returns `Some`
```

relay-server/src/actors/upstream.rs
```diff
@@ -235,1043 +252,1229 @@ impl fmt::Display for RequestPriority {
...
 
-impl<F> RequestBuilderTransformer for F
-where
-    F: FnMut(RequestBuilder) -> Result<Request, UpstreamRequestError> + Send + 'static,
-{
-    fn build_request(&mut self, builder: RequestBuilder) -> Result<Request, UpstreamRequestError> {
-        self(builder)
+    /// Whether this request should retry on network errors.
+    ///
+    /// Defaults to `true` and should be disabled if there is an external retry mechnism. Note that
```
olksdr:
```suggestion
    /// Defaults to `true` and should be disabled if there is an external retry mechanism. Note that
```

relay-server/src/actors/upstream.rs
```diff
@@ -235,1043 +252,1229 @@ impl fmt::Display for RequestPriority {
...
 ///
-///  1. `RateLimited` for a `429` status code.
-///  2. `ResponseError` in all other cases.
-fn handle_response(
-    response: Response,
-    intercept_status_errors: bool,
-    max_response_size: usize,
-) -> ResponseFuture<Response, UpstreamRequestError> {
-    let status = response.status();
+/// The implementing type directly constitutes the request payload.
```
olksdr: This sentence does not make sense to me. Could you, please, rephrase it?

relay-server/src/actors/upstream.rs
```diff
@@ -235,1043 +252,1229 @@ impl fmt::Display for RequestPriority {
...
+        false
+    }
+
+    fn route(&self) -> &'static str {
+        "response"
     }
 }
 
-impl Supervised for UpstreamRelay {}
+/// A shared, asynchonous client to build and execute requests.
```
olksdr:
```suggestion
/// A shared, asynchronous client to build and execute requests.
```

relay-server/src/actors/upstream.rs
```diff
@@ -235,1043 +252,1229 @@ impl fmt::Display for RequestPriority {
...
-            return Box::new(fut::err(()));
+    /// Handles an HTTP response returned from the upstream.
+    ///
+    /// If the response indicates success via 2XX status codes, `Ok(response)` is returned.
+    /// Otherwise, the response is consumed and an error is returned. If `intercept_status_errors`
+    /// is set to `true` on the request, depending on the status code and details provided in the
+    /// payload, one of the following errors is returned:
+    ///
+    ///  1. `RateLimited` for a `429` status code.
+    ///  2. `ResponseError`  in all other cases, containing the status and details.
```
olksdr:
```suggestion
    ///  2. `ResponseError` in all other cases, containing the status and details.
```

relay-server/src/actors/upstream.rs
```diff
@@ -235,1043 +252,1229 @@ impl fmt::Display for RequestPriority {
...
+    /// entries with low priority.
+    pub fn dequeue(&mut self) -> Option<Entry> {
+        self.high.pop_front().or_else(|| self.low.pop_front())
     }
+}
 
-    fn priority(&self) -> RequestPriority {
-        RequestPriority::Immediate
-    }
+/// The state of authentication state.
```
olksdr: I think it's already give the understanding what it is: 

```suggestion
/// The current authentication state.
```
Or the part `state of ... state` should be rewritten in other way, since it's a bit confusing.

relay-server/src/actors/upstream.rs
```diff
@@ -235,1043 +252,1229 @@ impl fmt::Display for RequestPriority {
...
+    /// Returns the interval at which this Relay should renew authentication.
+    ///
+    /// Returns `Some` if authentication should be retried. Returns `None` if authentication is
+    /// permanent.
+    fn renew_auth_interval(&self) -> Option<std::time::Duration> {
+        if self.config.processing_enabled() {
+            // processing relays do NOT re-authenticate
+            None
+        } else {
+            // only relays the have a configured auth-interval reauthenticate
```
olksdr:
```suggestion
            // only relays that have a configured auth-interval reauthenticate
```

relay-server/src/actors/upstream.rs
```diff
@@ -235,1043 +252,1229 @@ impl fmt::Display for RequestPriority {
...
-pub trait UpstreamRequest: Send {
-    /// The HTTP method of the request.
-    fn method(&self) -> Method;
+    /// Performs a single authentication pass.
+    ///
+    /// Authentication consists of an initial request, a challenge, and a signed response including
+    /// the challenge. Throughout this sequence, the auth monitor transitions the authentication
+    /// state. If authentication succeeds, the state is set to [`AuthState::Registered`] at the end.
+    ///
+    /// If any of the requests fails, this method returns an `Err` and leaves the last
```
olksdr:
```suggestion
    /// If any of the requests fail, this method returns an `Err` and leaves the last
```

relay-server/src/actors/upstream.rs
```diff
@@ -235,1043 +252,1229 @@ impl fmt::Display for RequestPriority {
...
-/// The handler decrements the counter of in-flight HTTP requests (since one was just
-/// finished) and tries to pump the http message queue by sending a `PumpHttpMessageQueue`
+/// Maintains outage state of the connection to the upstream.
 ///
-/// Every future representing an HTTP message sent by the ClientConnector is wrapped so that when
-/// it finishes or it is dropped a message is sent back to the actor to notify it that a http connection
-/// was freed.
+///  Use [`notify_error`](Self::notify_error) and [`reset_error`](Self::reset_error) to inform the
+/// monitor of successful and failed requests. If errors persist throughout a grace period, the
+/// monitor spanws a background task to re-establish connections. During this period,
```
olksdr:
```suggestion
/// monitor spawns a background task to re-establish connections. During this period,
```

olksdr (approved):
All of my comments are about typos or docs improvements, which is not blocker. 

I haven't noticed any obvious problems with the implementation, it looks good to me. 

I say, we can test this.

relay-server/src/actors/upstream.rs
jan-auer: I've added the relevant parts of this to the `UpstreamService` docs. You're right that we should have some module-level docs though. Will add them.

PR #1717, opened 2022-12-21
ref(utils): Test if events are kept when containing unsupported rule

TBS1996:
Previously two integration tests in python checked if the events are kept when they contain an unsupported rule, both with and without processing enabled. Now both of those are tested in a single rust unit test which tests the relevant functions directly instead. 

#skip-changelog


relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -284,7 +284,71 @@ mod tests {
...
+
+        rules.push(SamplingRule {
+            condition: RuleCondition::Unsupported,
+            sample_rate: 0.0,
+            ty: RuleType::Transaction,
+            id: RuleId(1),
+            time_range: Default::default(),
+        });
+
+        // now that an unsupported rule has been pushed, it should keep the event if processing is enabled
```
```suggestion
        // now that an unsupported rule has been pushed, it should keep the event if processing is disabled
```

tests/integration/test_dynamic_sampling.py
```diff
@@ -342,124 +342,6 @@ def test_it_keeps_events(mini_sentry, relay, rule_type, event_factory):
...
-        consumer = transactions_consumer()
-
-    # add a bad condition (with the opposite sample rate to make it evident it is not applied)
-    rules = _add_sampling_config(
-        config, sample_rate=1 - sample_rate, rule_type=rule_type
-    )
-    last_rule = rules[-1]
-    last_rule["condition"]["inner"].append(
-        {
-            "op": "BadOperator",
```
jjbayer, line 387: One thing we did not verify in the Rust test is whether `BadOperator` is indeed parsed into `Unsupported`. So we should add a rust unit test that tests only the parsing, something along the lines of https://github.com/getsentry/relay/blob/60d4305600372407d619a239f42ec8c295bb2f5e/relay-metrics/src/aggregation.rs#L2170-L2207.

jjbayer (approved):

PR #1718, opened 2022-12-21
fix(sessions): Deserialize abnormal_mechanism null as None

jjbayer:
The `abnormal_mechanism` field on `Session` defaults to `AbnormalMechanism::None` when missing, but when SDKs supply an explicit `null` value, we would fail serialization.

Use the existing `null_to_default` helper function to map `null` to a valid value.


shruthilayaj (approved):
thanks!

olksdr (approved):

PR #1719, opened 2022-12-22
chore(beta): Address beta lint warnings

olksdr:
This is mostly addressing [un-inlined format args](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args) lint error, which was enabled by default in beta.

With some bash and `cargo fix` all those args should be properly inlined.

#skip-changelog


relay-general/src/protocol/measurements.rs
```diff
@@ -136,7 +134,7 @@ impl DerefMut for Measurements {
 }
 
 fn is_valid_measurement_name(name: &str) -> bool {
-    name.starts_with(|c| matches!(c, 'a'..='z' | 'A'..='Z'))
+    name.starts_with(|c: char| c.is_ascii_alphabetic())
```
jjbayer, line 137: smart linter

jjbayer (approved):

sl0thentr0py (approved):

Zylphrex (approved):

PR #1720, opened 2022-12-22
ref: Convert integration tests about dropping transactions to unit tests

TBS1996:
previously there were two integration tests that checked if the transactions got dropped or not when they should based on the sample rate, now there is a single rust unit test that is more targeted to the relevant functions

Also created a testutils file as some helper functions from dynamic_sampling.rs were needed elsewhere in the crate

#skip-changelog


TBS1996:
should work after I convert the config file to yaml and hardcode it into the test but I have to buy some christmas presents first

relay-server/src/actors/processor.rs
```diff
@@ -2318,7 +2318,12 @@ impl Service for EnvelopeProcessorService {
 mod tests {
     use chrono::{DateTime, TimeZone, Utc};
     use relay_general::pii::{DataScrubbingConfig, PiiConfig};
+    use relay_general::protocol::EventId;
+    use relay_sampling::{RuleCondition, RuleType, SamplingMode};
 
+    use crate::service::ServiceState;
+    use crate::testutils::{new_envelope, state_with_rule_and_condition};
+    use crate::utils::Semaphore as MySemaphore;
```
olksdr: Can we remove this to `TestSemaphore` instead? 
```suggestion
    use crate::utils::Semaphore as TestSemaphore;
```

relay-server/src/actors/processor.rs
```diff
@@ -2353,6 +2358,57 @@ mod tests {
...
+
+        // an empty json still produces a valid config
+        let json_config = serde_json::json!({});
+
+        let config = Config::from_json_value(json_config.clone()).unwrap();
+        let arconfig = Arc::new(Config::from_json_value(json_config).unwrap());
+
+        // it really shouldn't be necessary to start an entire service for a unit test
+        // it was ported from a python integration test
+        ServiceState::start(arconfig).unwrap();
```
olksdr, line 2460: This starts all the actors/services we have in order for this unit test to work, which is kind of heavy and does not make much sense. 

But at this point I'm not sure how to solve this yet. 

The current status:
*  the function you test uses `reject` function on `envelope_context` which triggers the `TestStore` capture functionality 
https://github.com/getsentry/relay/blob/3b42784ba9359db65bd890b8d628414e3201de8c/relay-server/src/utils/envelope_context.rs#L190-L191

* this function also is triggered on `Drop` of the `envelope_context`
* also the `reject` function triggers `track_outcome` which also requires `Outcome` actor to be available in the registry.

olksdr (commented):
I still do not like that we have to start the entire `ServiceState` just for one unit test. Maybe it's worth looking into this more closely and actually test a little bit different (only the pieces we care about).

olksdr (approved):

PR #1723, opened 2022-12-27
ref(actix): Migrate server actor to the "service" arch

### LINKED ISSUE #1606: Update Server Actor

olksdr:
This change introduces changes to migrate `Server` actor to the new `Service` architecture:
* remove `actix` dependencies from the `Server` actor
* implement `Service` for the `Server` actor
* introduce the `HttpServer` helper struct, which takes care of starting the shutting down the `actix_web` http service, and removes exposed `Recipient` from the `Server` actor - now everything is hidden and in the future can be removed from the one place. 

The former `Server` actor and current `ServerService` subscribes only to `Shutdown` watch channel and makes sure to trigger the shutdown of the http server with all its workers.

These changes are still up to discussions. 

closes: #1606 

#skip-changelog

### DIFF
--- a/relay-server/src/lib.rs
+++ b/relay-server/src/lib.rs
@@ -296,5 +296,7 @@ pub fn run(config: Config) -> anyhow::Result<()> {
 
     // Properly shutdown the new tokio runtime.
     runtime.shutdown_timeout(shutdown_timeout);
+    relay_log::info!("relay shutdown complete");
+
     Ok(())
 }
--- a/relay-server/src/lib.rs
+++ b/relay-server/src/lib.rs
@@ -281,8 +281,14 @@ pub fn run(config: Config) -> anyhow::Result<()> {
     // create an actix system, start a web server and run all relevant actors inside. See the
     // `actors` module documentation for more information on all actors.
 
+    // Create the old tokio 0.x runtime. It's required for old actix System to exist by `create_runtime` utiliy.
+    //
+    // TODO(actix): this can be removed once all the actors are on the new tokio. It looks like
+    // that this mostly needed for Upstream actor right now. ANd
     let sys = actix::System::new("relay");
-    // We also need new tokio 1.x runtime.
+    // We also need new tokio 1.x runtime. This cannot be created in the [`relay_system::Controller`] since
+    // it cannot access the `create_runtime` utilily function from there. This can be changed once
+    // the [`actix::System`] is removed.
     let runtime = utils::create_runtime("http-server-handler", 1);
     let shutdown_timeout = config.shutdown_timeout();
 
--- a/relay-server/src/lib.rs
+++ b/relay-server/src/lib.rs
@@ -269,7 +269,7 @@ mod utils;
 use relay_config::Config;
 use relay_system::Controller;
 
-use crate::actors::server::Server;
+use crate::actors::server::ServerService;
 
 /// Runs a relay web server and spawns all internal worker threads.
 ///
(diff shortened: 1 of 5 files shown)


relay-system/src/controller.rs
```diff
@@ -103,20 +104,27 @@ impl Controller {
...
+    /// new one is available.
+    ///
     /// The factory may be used to start actors in the actix system before it runs. If the factory
     /// returns an error, the actix system is not started and instead an error returned. Otherwise,
     /// the system blocks the current thread until a shutdown signal is sent to the server and all
     /// actors have completed a graceful shutdown.
-    pub fn run<F, R, E>(factory: F) -> Result<(), E>
+    pub fn run<F, R, E>(
+        handle: &tokio::runtime::Handle,
+        sys: SystemRunner,
```
jjbayer, line 117: Why not create the runtime and the `actix::SystemRunner` inside `run()`? Is there a use case where `run()` recycles an existing runtime?

relay-server/src/service.rs
```diff
@@ -320,30 +322,52 @@ where
...
+        server = listen(server, &config)?;
+        server = listen_ssl(server, &config)?;
+
+        dump_listen_infos(&server);
+        let recipient = server.start().recipient();
+        Ok(Self(recipient))
+    }
+
+    /// Triggers the shutdown process by sending [`actix_web::server::StopServer`] to the running http server.
+    pub fn shutdown(self, graceful: bool) {
```
jjbayer: Why does this take `self` by value instead of by reference?

relay-system/src/controller.rs
olksdr: This is because, if I remember it correctly - I use in `relay-server/src/lib.rs` following code
```rust
let runtime = utils::create_runtime("http-server-handler", 1);
```
And it requires that old system is already available at this point. And also I wanted to clean a little bit controller, and make sure that controller knows less about the system it runs on, just accept the system as the parameter.

relay-server/src/service.rs
olksdr: I guess it mostly to signal to the caller that once you can the shutdown, the server is not longer can be used after that.

jjbayer: The intent makes sense, but wouldn't changing it to `&mut self` prevent the need to clone [here](https://github.com/getsentry/relay/pull/1723/files#diff-a37078486e7ec6da20e15d45dc8c8cec32ae64b648bdd086d0287274336098b0R28)?

relay-system/src/controller.rs
jjbayer: But isn't spinning up the actix system kind of what the `Controller` is for? Looking at its docs

https://github.com/getsentry/relay/blob/dd9f998ee324a710528df2fa76f87a64614a45ef/relay-system/src/controller.rs#L47

olksdr: I've changed this doc and added more docs to clarify why it's done this way now.

relay-server/src/service.rs
olksdr: you're right. I've changed it.

relay-server/src/lib.rs
```diff
@@ -280,5 +280,21 @@ pub fn run(config: Config) -> anyhow::Result<()> {
...
+    // We also need new tokio 1.x runtime. This cannot be created in the [`relay_system::Controller`] since
+    // it cannot access the `create_runtime` utilily function from there. This can be changed once
+    // the [`actix::System`] is removed.
+    let runtime = utils::create_runtime("http-server-handler", 1);
+    let shutdown_timeout = config.shutdown_timeout();
+
+    Controller::run(runtime.handle(), sys, || ServerService::start(config))?;
+
+    // Properly shutdown the new tokio runtime.
+    runtime.shutdown_timeout(shutdown_timeout);
```
jjbayer, line 298: `Controller::run()` logs a message `relay shutdown complete` after returning from `sys.run()`. So by shutting down the tokio runtime with `shutdown_timeout` here, do we effectively double the possible shutdown time?

olksdr: From the [`tokio docs`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_timeout):
> Shuts down the runtime, waiting for at most duration for all spawned task to shutdown.

So basically if there is something blocking, we will wait max `shutdown_timeout` and then kill everything.

jjbayer (approved):

PR #1724, opened 2022-12-27
ci: Add GoCD deployment pipeline

joshuarli:
This adds deployment configurations for GoCD, the new internal continuous deployment tool. Please refer to the [Freight Migration Guide](https://www.notion.so/sentry/Freight-Migration-Guide-2bd47bb683254077ad31e7ad625345ae) for info on how to coordinate GoCD deploys during this transitional phase.

#skip-changelog


gocd/pipelines/relay-pop.yaml
```diff
@@ -0,0 +1,90 @@
...
+                pipeline: deploy-relay
+                stage: promote-to-pops
+        stages:
+            # checks aren't needed because it's implied by pipeline-deploy-relay-completed.
+            - deploy:
+                  approval:
+                      type: manual
+                  fetch_materials: true
+                  jobs:
+                      # XXX: Can't have pop releases yet unless we give
```
joshuarli: oh oops - this still isn't done. we'll get it done by EOW

cloudbuild.yaml
```diff
@@ -52,6 +53,26 @@ steps:
         ".",
       ]
 
+  - name: "us.gcr.io/$PROJECT_ID/relay:$COMMIT_SHA"
```
olksdr: We do not use `cloudbuild` anymore, and all the images are built directly in Github actions inside of the docker. See https://github.com/getsentry/relay/blob/d1e4d1284bf07c8ab0846f5c09ab2ef632b498c6/.github/workflows/ci.yml#L228

and later we push the image to GCR as well, here https://github.com/getsentry/relay/blob/d1e4d1284bf07c8ab0846f5c09ab2ef632b498c6/.github/workflows/ci.yml#L277

If we need to upload anything additionally to any of the buckets, we can do from the CI job as well.

joshuarli: Production is still using images from `us.gcr.io`; a recent deploy is using `us.gcr.io/sentryio/relay:31b033f3f5f1dca7023bf0cf2b6ddc2f7d604851` but I wasn't able to find what's building them... the only cloudbuild trigger using relay repo in sentryio project is called relay-branch-push which last ran Feb 9.

joshuarli: Oh, I see the cloudbuilds stopped around Feb 9. Cool, thanks for letting me know.

joshuarli: Adjusted.

.github/workflows/ci.yml
```diff
@@ -305,13 +305,19 @@ jobs:
...
           set -euxo pipefail
           docker pull -q "$IMG_VERSIONED"
           docker tag "$IMG_VERSIONED" "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}"
           docker push "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}"
 
+          docker run --rm --entrypoint cat "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}" /opt/relay-debug.zip > relay-debug.zip
+          docker run --rm --entrypoint cat "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}" /opt/relay.src.zip > relay.src.zip
+          docker run --rm --entrypoint tar "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}" -cf - /lib/x86_64-linux-gnu > libs.tar
+          gsutil -m cp ./libs.tar ./relay-debug.zip ./relay.src.zip \
+          "gs://dicd-team-devinfra-cd--getsentry-relay/deployment-assets/${{ github.event.pull_request.head.sha || github.sha }}/"
```
olksdr: Looks CI is failing to upload the artefacts to the bucket. 
If it exists the OPS team will have to update the access permissions for service account which is used here to be able to access this specific bucket.

/cc @oioki

joshuarli: Oh yeah, I need to figure out how how to authorize github actions to do this. Cloudbuild was easy because it just ran with a google service account.

joshuarli: Oh, nice. Already set up.

```
        uses: google-github-actions/auth@v1
        with:
          workload_identity_provider: projects/345757944225/locations/global/workloadIdentityPools/github-actions-pool/providers/github-actions-provider
          service_account: github-actions-gcr@sentryio.iam.gserviceaccount.com```

**Comment by joshuarli** [comment]
hmm - guess we can't rely on the builtin gcloud, need to use `google-github-actions/setup-gcloud` instead
```
ResumableUploadAbortException: 401 Anonymous caller does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).
```

https://github.com/google-github-actions/auth#authenticating-via-workload-identity-federation
> You must use the Cloud SDK version 390.0.0 or later to authenticate the bq and gsutil tools.

File: .github/workflows/ci.yml [hunk 4]
```diff
@@ -301,17 +301,30 @@ jobs:
...
+      - name: Push to us.gcr.io and upload deployment assets
         run: |
           set -euxo pipefail
           docker pull -q "$IMG_VERSIONED"
           docker tag "$IMG_VERSIONED" "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}"
           docker push "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}"
 
+          docker run --rm --entrypoint cat "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}" /opt/relay-debug.zip > relay-debug.zip
+          docker run --rm --entrypoint cat "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}" /opt/relay.src.zip > relay.src.zip
+          docker run --rm --entrypoint tar "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}" -cf - /lib/x86_64-linux-gnu > libs.tar
```
olksdr: Since `"us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}"` repeats so many times right now, it would be a good opportunity to refactor this a bit and create an env variable (which makes all of this a bit shorter), see how `IMG_VERSIONED` was configured above (line 294).

relay-log/build.rs
```diff
@@ -17,9 +17,8 @@ fn emit_release_var() -> Result<(), io::Error> {
         ));
     }
 
-    let version = std::env::var("CARGO_PKG_VERSION").unwrap();
     let revision = String::from_utf8_lossy(&cmd.stdout);
-    println!("cargo:rustc-env=RELAY_RELEASE=relay@{version}+{revision}");
+    println!("cargo:rustc-env=RELAY_RELEASE=relay@{revision}");
```
olksdr: Why is this change needed?

olksdr (approved):
Overall this looks good to me. I left a comment, let me know what you think. 

We might want to wait for @jan-auer  to also have a quick look.

relay-log/build.rs
joshuarli: read the PR description - it's part of sentry release names

jan-auer: Please revert this, we explicitly want this version number in our builds. This is relevant for our self-hosted builds.

.github/workflows/ci.yml
```diff
@@ -301,17 +301,30 @@ jobs:
...
 
-      - name: Push to us.gcr.io
+      - name: Push to us.gcr.io and upload deployment assets
         run: |
           set -euxo pipefail
           docker pull -q "$IMG_VERSIONED"
           docker tag "$IMG_VERSIONED" "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}"
           docker push "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}"
 
+          docker run --rm --entrypoint cat "us.gcr.io/sentryio/relay:${{ github.event.pull_request.head.sha || github.sha }}" /opt/relay-debug.zip > relay-debug.zip
```
jan-auer: Since we no longer need the symbols in Docker for the freight deployment, we should change the build to not even put the debug files in the docker container. Instead, we can just save them as a GitHub actions asset and then re-upload them to our devinfra.

.github/workflows/lint-pipelines.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+    lint:
+        name: Lint GoCD Pipelines
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c  # v3
+            - uses: getsentry/action-setup-gocd-cli@8391790b37dcf61bdf52d13fa1bf044c9bee9ac9  # v1.0.0
+            - name: Lint Pipelines with gocd-cli
+              run: ./.github/workflows/lint-pipelines.sh
```
jan-auer: Given the script is 3 lines and a redundant echo, can we inline it here?

.github/workflows/lint-pipelines.yml
```diff
@@ -0,0 +1,20 @@
+name: Lint Deployment Pipelines
+
+on:
+    pull_request:
+    push:
+        branches: [main, test-me-*]
```
jan-auer: - We do not have a `main` branch, this should be `master`. 
- Also, we require these deployments from all PR branches in Relay, so should we be running it there, too?
- We usually use a `/` as pattern, so this would be `test/pipeline-*` for instance.

.github/workflows/lint-pipelines.yml
```diff
@@ -0,0 +1,20 @@
...
+concurrency:
+    group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+    cancel-in-progress: true
+
+jobs:
+    lint:
+        name: Lint GoCD Pipelines
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c  # v3
```
jan-auer: Since GitHub's `actions` repo is a trusted vendor, we do not pin SHAs here. See the other actions.

```suggestion
            - uses: actions/checkout@v3  # v3
```

For readability, I would also use the version string for our own actions (i.e. `getsentry/*`)

gocd/pipelines/relay.yaml
```diff
@@ -0,0 +1,118 @@
...
+                                    /devinfra/scripts/checks/githubactions/checkruns.py \
+                                    getsentry/relay \
+                                    ${GO_REVISION_RELAY_REPO} \
+                                    "Integration Tests" \
+                                    "Test (macos-latest)" \
+                                    "Test (windows-latest)" \
+                                    "Test All Features (ubuntu-latest)" \
+                                    "self-hosted-end-to-end" \
+                                    "Sentry-Relay Integration Tests" \
+                                    "Test Python 3.8"
```
jan-auer: Did we have these listed in freight before? We can remove "Test Python 3.8" from this list, and should add "Push GCR Docker Image"

scripts/create-sentry-release
```diff
@@ -26,27 +30,24 @@ if [ -z "${SENTRY_ORG:-}" ] || [ -z "${SENTRY_PROJECT:-}" ]; then
...
-echo 'Obtaining version information...'
-VERSION="$(docker run --rm "${DOCKER_IMAGE}" --version | cut -d' ' -f2)"
-RELEASE="relay@${VERSION}+${REVISION}"  # match RELEASE in relay_log::setup
-
 echo "Creating a new release and deploy for ${RELEASE} in Sentry..."
+RELEASE="relay@${REVISION}"  # match RELEASE in relay_log::setup
 sentry-cli releases new "${RELEASE}"
 sentry-cli releases set-commits "${RELEASE}" --commit "${GITHUB_PROJECT}@${REVISION}"
-sentry-cli releases deploys "${RELEASE}" new -e release
+sentry-cli releases deploys "${RELEASE}" new -e "$ENVIRONMENT"
```
jan-auer: I'm not sure this actually makes sense. Usually, canary is a "deployment" for us, but not an "environment", so I need a bit of input on how to read the above gocd configuration. 

In freight, we used to have a "canary" pod that was part of a regular production deployment. 99% of the time, it would be no different than all others, and we want all sentry data emitted from that pod to be labeled as "production" environment. Otherwise, we don't see it in Sentry with default filters. 

When we selected a "canary" deployment in Freight, this one pod would be changed to the new revision. We didn't care about the environment, because it had a new release name anyway. 

The next production deployment reset the canary, too.

How does the current pipeline compare to this?

jan-auer (changes requested):

scripts/create-sentry-release
jan-auer: To clarify on the terminology used here: Our "canary" deployment isn't a canary in the sense that devinfra has been using it. Instead, it's an experimental deployment/tier that receives an equal share of traffic.

gocd/pipelines/relay.yaml
joshuarli: It's currently this in Freight, I'll update it.

```
          "Integration Tests",
          "Test (macos-latest)",
          "Test (windows-latest)",
          "Test All Features (ubuntu-latest)",
          "Push GCR Docker Image"
```

joshuarli: Any reason not to have `"Sentry-Relay Integration Tests"` and `"self-hosted-end-to-end"`?

scripts/create-sentry-release
joshuarli: Ah... okay. So in this case, I'll create a separate deploy-relay-experimental pipeline that will just deploy the "canary".

And then the regular deployment pipeline will deploy canary too.

joshuarli: As for sentry releases - it doesn't seem to be working as you describe it? I expected a new release name for canary deploys but they're the same.
Example:

https://freight.getsentry.net/deploys/relay/canary/131
`Created release relay@23.3.0+0d52aad0aa46887b6009db52d5d037e625b51975.`

https://freight.getsentry.net/deploys/relay/production/711
`Created release relay@23.3.0+0d52aad0aa46887b6009db52d5d037e625b51975.`

https://sentry.my.sentry.io/organizations/sentry/releases/relay%4023.3.0%2B0d52aad0aa46887b6009db52d5d037e625b51975/?project=4

relay-log/build.rs
joshuarli: Hmm... i'll have to think of a way gocd can get the semver then. 

How about this? Have an additional RELAY_SHA that'll be passed to sentry-sdk for release names. RELAY_RELEASE can stay.

Like this: https://github.com/getsentry/relay/pull/1724/commits/cb7b7a5f8e3e8cf7fab4c5692b73c54ec2646ef9

.github/workflows/ci.yml
joshuarli: Can we punt this for now? Not sure how to approach that yet... it's built in the builder image and we'd need some way of copying it to the GHA runtime. Could upload the builder image but there's surely better ways.

.github/workflows/lint-pipelines.yml
joshuarli: Corrected. re #2: pull_request: will run on everything by default, no?

gocd/pipelines/relay.yaml
jan-auer: I think we skipped those because they were often slow and unstable. The situation has improved since, so we can consider to add them. Feel free to defer this to a later time.

scripts/create-sentry-release
jan-auer: Canary deploys should have the same release as their regular deploys. The release name purely depends on the build.

.github/workflows/ci.yml
jan-auer: Yeah, fine with me 👍

gocd/pipelines/relay.yaml
joshuarli: Ok - will leave that up to the ingest team.

scripts/create-sentry-release
joshuarli: Ok, same release name then. How about environment? I made a new deploy-relay-experimental pipeline and right now it's `experimental` environment: https://github.com/getsentry/relay/pull/1724/files#diff-4717184ee67f6e9554253d98f87bbbfc8c64255329f47245d42249035c106939R65

Should it be `experimental` or `production`?

.github/workflows/lint-pipelines.yml
jan-auer: Right, I had missed `pull_request:` above. That's what we're looking for. It runs on every commit that is pushed to an open pull request, as well as the last commit of a branch when a pull request is opened on it.

scripts/create-sentry-release
jan-auer: The environment can be `production` in both cases, that's what it is. We differentiate purely by the release name (version+sha).

relay-log/build.rs
jan-auer: The point is to have both the version number and the SHA in the sentry-sdk too. I understand the limitation of not running docker-in-docker, could we consider to use the devinfra for this? I'm kinda uneasy about enforcing a specific versioning schema (i.e. change the application) just because the CI pipeline can't obtain the actual version number. 

Worst case, we just don't create the release here and let sentry auto-create it. It would be great to have it created eagerly, but I'd choose names over eager creation.

joshuarli: As discussed in Slack I'll switch to writing the release name to the deployment bucket. It's still way better than running d-in-d and pulling an entire image.

jan-auer (approved):

olksdr (approved):
🚀 🚢  

sucess

Issue #1693, opened 2022-12-13
Scrub more IP addresses 


jjbayer:
If the flag `scrub_ip_addresses` is enabled in legacy data scrubbing, we only scrub a handful of fields:

https://github.com/getsentry/relay/blob/f8841223ee689ff0ce50a9bd9d75223a4fb4ddd0/relay-general/src/pii/convert.rs#L41-L43

Should we instead (or in addition) apply the rule `@ip:replace` when this flag is enabled, such that any IP is caught?

https://github.com/getsentry/relay/blob/f8841223ee689ff0ce50a9bd9d75223a4fb4ddd0/relay-general/src/pii/builtin.rs#L89-L94

See also https://github.com/getsentry/sentry-docs/issues/5916.

PR #1725, opened 2022-12-29
feat(general): Scrub all fields with IP address

TBS1996:
This change will scrub all fields containing an IP address, rather than scrubbing certain fields known to have them 

for the unit tests, testing ip scrubbing in sdk and user have been merged as it should scrub all fields.

referencing issue #1693

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Add max replay size configuration parameter. ([#1694](https://github.com/getsentry/relay/pull/1694))
 - Add nonchunked replay recording message type. ([#1653](https://github.com/getsentry/relay/pull/1653))
 - Add `abnormal_mechanism` field to SessionUpdate protocol. ([#1665](https://github.com/getsentry/relay/pull/1665))
+- Scrub all fields with IP addresses rather than only known IP address fields. (---)
 
 **Bug Fixes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 - Add max replay size configuration parameter. ([#1694](https://github.com/getsentry/relay/pull/1694))
 - Add nonchunked replay recording message type. ([#1653](https://github.com/getsentry/relay/pull/1653))
 - Add `abnormal_mechanism` field to SessionUpdate protocol. ([#1665](https://github.com/getsentry/relay/pull/1665))
-- Scrub all fields with IP addresses rather than only known IP address fields. (---)
+- Scrub all fields with IP addresses rather than only known IP address fields. ([#1693](https://github.com/getsentry/relay/pull/1725))
 
 **Bug Fixes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 - Add max replay size configuration parameter. ([#1694](https://github.com/getsentry/relay/pull/1694))
 - Add nonchunked replay recording message type. ([#1653](https://github.com/getsentry/relay/pull/1653))
 - Add `abnormal_mechanism` field to SessionUpdate protocol. ([#1665](https://github.com/getsentry/relay/pull/1665))
-- Scrub all fields with IP addresses rather than only known IP address fields. ([#1693](https://github.com/getsentry/relay/pull/1725))
+- Scrub all fields with IP addresses rather than only known IP address fields. ([#1725](https://github.com/getsentry/relay/pull/1725))
 
 **Bug Fixes**:
 
--- a/relay-general/src/pii/convert.rs
+++ b/relay-general/src/pii/convert.rs
@@ -9,14 +9,6 @@ use crate::pii::{
 };
 use crate::processor::{SelectorPathItem, SelectorSpec, ValueType};
 
-// XXX: Move to @ip rule for better IP address scrubbing. Right now we just try to keep
-// compatibility with Python.
-static KNOWN_IP_FIELDS: Lazy<SelectorSpec> = Lazy::new(|| {
-    "($request.env.REMOTE_ADDR | $user.ip_address | $sdk.client_ip)"
-        .parse()
-        .unwrap()
-});
-
 /// Fields that the legacy data scrubber cannot strip.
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 - Add max replay size configuration parameter. ([#1694](https://github.com/getsentry/relay/pull/1694))
 - Add nonchunked replay recording message type. ([#1653](https://github.com/getsentry/relay/pull/1653))
 - Add `abnormal_mechanism` field to SessionUpdate protocol. ([#1665](https://github.com/getsentry/relay/pull/1665))
+- Scrub all fields with IP addresses rather than only known IP address fields. ([#1693](https://github.com/getsentry/relay/pull/1725))
```
```suggestion
- Scrub all fields with IP addresses rather than only known IP address fields. ([#1725](https://github.com/getsentry/relay/pull/1725))
```

relay-general/src/pii/convert.rs
```diff
@@ -39,7 +31,8 @@ pub fn to_pii_config(
     }
 
     if datascrubbing_config.scrub_ip_addresses {
-        applications.insert(KNOWN_IP_FIELDS.clone(), vec!["@anything:remove".to_owned()]);
+        let wildcard = SelectorSpec::Path(vec![SelectorPathItem::DeepWildcard]);
+        applications.insert(wildcard, vec!["@ip:replace".to_owned()]);
```
olksdr: The question is do we want to keep this consistent with old behaviour and just remove them or actually switch to replacing all the IPs as you did here.

relay-general/src/pii/convert.rs
```diff
@@ -445,12 +438,15 @@ THd+9FBxiHLGXNKhG/FRSyREXEt+NyYIf/0cyByc9tNksat794ddUqnLOg0vwSkv
...
+    /// Checks that any fields containing an IP-address is scrubbed.
+    /// Even if it doesn't make sense for there to be an IP-address
+    /// such as in the username field.
     #[test]
     fn test_user_ip_stripped() {
         let mut data = Event::from_value(
             serde_json::json!({
                 "user": {
-                    "username": "secret",
+                    "username": "73.133.27.120",
```
olksdr: it would be great to test if the scrubbing works also in , e.g. `breadcrumbs` - if you add the message with ip inside , as it's done in test https://github.com/getsentry/relay/blob/3b42784ba9359db65bd890b8d628414e3201de8c/relay-general/src/pii/convert.rs#L1197 
the protocol for `breadcrumbs` is in https://github.com/getsentry/relay/blob/3b42784ba9359db65bd890b8d628414e3201de8c/relay-general/src/protocol/breadcrumb.rs#L49 where you can see which fields are PII

relay-general/src/pii/snapshots/relay_general__pii__convert__tests__user_ip_stripped.snap
```diff
@@ -15,16 +15,61 @@ expression: data
...
     }
   },
+  "breadcrumbs": {
+    "values": [
+      {
+        "message": "[ip]"
+      }
+    ],
+    "data": {
+      "test_data": "73.133.27.120"
```
olksdr, line 25: Should the data be stripped here as well?

TBS1996: hmm that was the idea, I believed deepwilcard would check fields recursively, seems not

relay-general/src/pii/convert.rs
```diff
@@ -27,6 +19,14 @@ static DATASCRUBBER_IGNORE: Lazy<SelectorSpec> = Lazy::new(|| {
         .unwrap()
 });
 
+// XXX: Move to @ip rule for better IP address scrubbing. Right now we just try to keep
+// compatibility with Python.
```
jjbayer:
```suggestion
/// Fields that are known to contain IPs. Used for legacy IP scrubbing.
```

relay-general/src/pii/convert.rs
```diff
@@ -39,7 +39,12 @@ pub fn to_pii_config(
     }
 
     if datascrubbing_config.scrub_ip_addresses {
+        // legacy(?) scrubs all fields that are known to have IPs regardless of actual content
         applications.insert(KNOWN_IP_FIELDS.clone(), vec!["@anything:remove".to_owned()]);
+
+        // checks actual contents of all fields and scrubs where there is an IP address
+        let wildcard = SelectorSpec::Path(vec![SelectorPathItem::DeepWildcard]);
```
jjbayer: Instead of pushing to `applications` with a custom selector spec for IP scrubbing, I would simply push `@ip:replace` to `applied_rules` (see code further below). It will then use the same selector as the other default rules, which I think is what we want.

relay-general/src/pii/convert.rs
```diff
@@ -445,14 +448,30 @@ THd+9FBxiHLGXNKhG/FRSyREXEt+NyYIf/0cyByc9tNksat794ddUqnLOg0vwSkv
...
+                    "username": "73.133.27.120", // should be stripped despite not being "known ip field"
+                    "ip_address": "should be stripped despite lacking ip address",
                     "data": sensitive_vars()
+                },
+                "breadcrumbs": {
+                    "values": [
+                        {
+                            "message": "73.133.27.120",
+                            "data": {
+                                "test_data": "73.133.27.120" // test deep wildcard stripping
```
olksdr, line 458: As we talked in the meeting, can we also make sure to test that the IPs are scrubbed also from inside the strings, e.g. 

```json
"test_data": "contains ip's like 73.133.27.120 but also the text"
```

which should result in something like 

```json
"test_data": "contains ip's like [ip] but also the text"

```

or make sure this actually can work?

jjbayer (approved):

relay-general/src/pii/convert.rs
jjbayer: I would create a separate issue for this, as it would also the touch IP scrubbing for users who configured it manually through advanced data scrubbing.

PR #1726, opened 2022-12-29
chore(project): Add backoff mechanism for fetching projects

### LINKED ISSUE #1633: [INC-263] Backoff mechanism for fetching projects in case of error
With https://github.com/getsentry/relay/pull/1426 deployed in case of invalid project state (when it could not be parsed or fetched) we will reuse the old state up until the grace period is reached. 

This does not stop the Relay to try and update the state in next refresh cycle again though. We must make sure that we have a back off for these retries so we won’t DDoS our own services with requests.

olksdr:
In case of getting invalid project we try to reuse the stale state and make sure to initiate the backoff to slow down the requests to the project cache, so we don't ddos out own service.

We reset the backoff only when the new state is accepted.

closes: #1633


relay-server/src/actors/project.rs
```diff
@@ -867,9 +880,18 @@ impl Project {
...
-            ExpiryState::Updated(old) | ExpiryState::Stale(old) if state.invalid() => state = old,
+            //This also sets the next fetch attempt time.
+            ExpiryState::Updated(old) | ExpiryState::Stale(old) if state.invalid() => {
+                state = old;
+                self.next_fetch_attempt = Instant::now().checked_add(self.backoff.next_backoff());
+            }
             // If the new state is valid or the old one is expired, always use the new one.
-            _ => self.state = Some(state.clone()),
+            // And also reset the backoff at this point.
+            _ => {
```
jjbayer: What happens when `self.expiry_state()` is `Expired`, but the new `state` is `invalid()`? Shouldn't we then also set a `next_fetch_attempt`?

jjbayer (commented):
Should we add a test for this behavior? Something like https://github.com/getsentry/relay/blob/e87b92e261ffe4f563c8e5519c12c7c1e756984d/tests/integration/test_query.py#L63

relay-server/src/actors/project.rs
olksdr: Currently (and the behaviour we had before) once the old state expires we save anything we get from upstream - this case the upstream is alive, just delivers the invalid state.  Since we cannot use the old state either, we will just try to request the new one soon-ish again. 
Do you think we have to re-think this behaviour in this PR?

olksdr:
> Should we add a test for this behavior? Something like
> 
> https://github.com/getsentry/relay/blob/e87b92e261ffe4f563c8e5519c12c7c1e756984d/tests/integration/test_query.py#L63

Not sure about the test. I do not really want to introduce more "waits" in already slow integration tests. 
But I can have a look into this if you think it would be better to have one.

jjbayer (approved):

PR #1727, opened 2022-12-30
ref(actix): Migrate ProjectUpstream to `relay_system::Service`

### LINKED ISSUE #1607: Update ProjectUpstream Actor

olksdr:
Remove `actix` actor implementation and replace it with `relay_system::Service`.

Also replaces old `futures01` with new `futures`, and use `tokio::spawn` to schedule the task which fetches the `ProjectState`'s from upstream. In order to do that I had to change how the scheduling is working, so the entire service won't be blocking:

* introduces the internal state channel
* use messages to pass the current state around, like reset the backoff
  period or schedule the another fetch
* the internal message have also higher priority right now

closes: #1607 

#skip-changelog


relay-server/src/actors/project_upstream.rs
```diff
@@ -94,31 +93,67 @@ impl ProjectStateChannel {
...
+    state_tx: mpsc::UnboundedSender<UpstreamProjectSourceState>,
+    state_rx: mpsc::UnboundedReceiver<UpstreamProjectSourceState>,
 }
 
-impl UpstreamProjectSource {
+impl UpstreamProjectSourceService {
+    /// Creates a new [`UpstreamProjectSourceService`] instance.
     pub fn new(config: Arc<Config>) -> Self {
-        UpstreamProjectSource {
+        let (state_tx, state_rx) = mpsc::unbounded_channel();
```
jjbayer: TODO: is unbounded ok?

relay-server/src/actors/project_upstream.rs
```diff
@@ -182,200 +212,274 @@ impl UpstreamProjectSource {
...
+        for response in responses {
+            match response {
+                Ok((channels_batch, Ok(mut response))) => {
+                    // If a single request succeeded we reset the backoff. We decided to
+                    // only backoff if we see that the project config endpoint is
+                    // completely down and did not answer a single request successfully.
+                    //
+                    // Otherwise we might refuse to fetch any project configs because of a
+                    // single, reproducible 500 we observed for a particular project.
+                    if let Err(err) = state_tx.send(UpstreamProjectSourceState::ResetBackoff) {
```
jjbayer: Change names from `state_tx` to `inner_tx`?

relay-server/src/actors/project_upstream.rs
```diff
@@ -66,25 +72,26 @@ impl UpstreamQuery for GetProjectStates {
     }
 }
 
+/// The wrapper struct for the incoming external requests which also keeps addition inforamtion.
```
jjbayer:
```suggestion
/// The wrapper struct for the incoming external requests which also keeps addition information.
```

relay-server/src/actors/project_upstream.rs
```diff
@@ -66,25 +72,26 @@ impl UpstreamQuery for GetProjectStates {
...
 #[derive(Debug)]
 struct ProjectStateChannel {
-    sender: oneshot::Sender<Arc<ProjectState>>,
-    receiver: Shared<oneshot::Receiver<Arc<ProjectState>>>,
+    sender: BroadcastChannel<Arc<ProjectState>>,
```
jjbayer: Should we rename this to `channel`, or `inner`?

relay-server/src/actors/project_upstream.rs
```diff
@@ -93,32 +100,70 @@ impl ProjectStateChannel {
...
+        self.sender.send(Arc::new(state))
     }
 
     pub fn expired(&self) -> bool {
         Instant::now() > self.deadline
     }
 }
 
+/// Internal [`UpstreamProjectSourceService`] message protocol.
+enum UpstreamProjectSourceState {
```
jjbayer: I would rename this to `UpstreamProjectSourceInternal`, just to prevent naming confusion with `ProjectState`.

relay-server/src/actors/project_upstream.rs
```diff
@@ -182,200 +222,261 @@ impl UpstreamProjectSource {
...
-                    //  -1 to schedule repeated fetch requests (at less and less frequent intervals)
-                    //  -2 as a flag to know if a fetch is already scheduled.
-                    // Resetting it in here signals that we don't have a backoff scheduled (either
-                    // because everything went fine or because all the requests have expired).
-                    // Next time a user wants a project it should schedule fetch requests.
-                    slf.backoff.reset();
+                    // Put the channels back into the queue, we will retry again shortly.
+                    state_channels.extend(channels_batch);
+                }
+                Err(err) => {
```
jjbayer: In this case, should we not also preserve `channels_batch` and put it back into `state_channels`?

relay-server/src/actors/project_upstream.rs
olksdr: done in f8c83d2

relay-server/src/actors/project_upstream.rs
olksdr: done in a2466fc

relay-server/src/actors/project_upstream.rs
olksdr: done in f8c83d2

relay-server/src/actors/project_upstream.rs
olksdr: done in f8c83d2

relay-server/src/actors/project_upstream.rs
olksdr: done f8c83d2

jjbayer (approved):
Makes sense to me, but I would be more comfortable with a second approval.

relay-server/src/actors/project_upstream.rs
```diff
@@ -182,200 +238,222 @@ impl UpstreamProjectSource {
...
-        relay_log::info!("project upstream cache started");
+        let channels: HashMap<ProjectKey, ProjectStateChannel> =
+            self.state_channels.drain().collect();
+        let config = self.config.clone();
+        let attempt = self.backoff.attempt();
+        let inner_tx = self.inner_tx.clone();
+        let wait = self.next_backoff();
+
+        tokio::spawn(async move {
+            tokio::time::sleep(wait).await;
```
jan-auer: Instead of a `time::sleep` here, please use a `SleepHandle` like in the `OutcomeAggregator`. 

With `sleep`, you immediately schedule a task, but you capture only the channels that are registered at the time of scheduling. Everything that comes in after that until the flush happens will be missed. 

If you use a sleep handle, you can take channels right before executing the flush.

relay-server/src/actors/project_upstream.rs
```diff
@@ -182,200 +238,222 @@ impl UpstreamProjectSource {
...
+    /// Spawns the [`tokio::task`] to fetch the project configs from upstream.
+    fn schedule_fetch(&mut self) {
+        if self.state_channels.is_empty() {
+            relay_log::error!("project state schedule fetch request without projects");
+            return;
+        }
 
-        relay_log::info!("project upstream cache started");
+        let channels: HashMap<ProjectKey, ProjectStateChannel> =
+            self.state_channels.drain().collect();
```
jan-auer: This can be a `std::mem::take(&mut self.state_channels)`. My expectation would be that this observes better performance and also avoids a memory "leak", but I might be off on that.

relay-server/src/actors/project_upstream.rs
```diff
@@ -1,25 +1,49 @@
...
+/// * otherwise we get the tuple of [`ProjectStateChannel`]
+///   tree which we can re-used to re-schedule the fetch or the channel error.
+type ProjectStatesResponses = Vec<
+    Result<
+        (
+            BTreeMap<ProjectKey, ProjectStateChannel>,
+            Result<GetProjectStatesResponse, UpstreamRequestError>,
+        ),
+        (BTreeMap<ProjectKey, ProjectStateChannel>, ProjectError),
+    >,
```
jan-auer: Two comments on this, if possible:

1. Use `HashMap` here instead of `BTreeMap`. You could even create a typedef for that map, since this is used in several places.
2. It should be possible to flatten this type (and further the channel's value type) into a single structure with two fields: the result, and remaining channels.

relay-server/src/actors/project_upstream.rs
```diff
@@ -130,36 +188,34 @@ impl UpstreamProjectSource {
...
         // Pop N items from state_channels. Intuitively, we would use
-        // `self.state_channels.drain().take(n)`, but that clears the entire hashmap regardless how
+        // `state_channels.drain().take(n)`, but that clears the entire hashmap regardless how
         // much of the iterator is consumed.
         //
         // Instead, we have to collect the keys we want into a separate vector and pop them
         // one-by-one.
-        let projects: Vec<_> = (self.state_channels.keys().copied())
+        let projects: Vec<_> = (state_channels.keys().copied())
             .take(batch_size * num_batches)
```
jan-auer: If you move this `take` to `schedule_fetch`, you ensure that all channels that won't be fetched remain in the map and get appended to.

I would even go further and also move:
 - The `filter_map` + `filter` below, to further reduce allocations of the map that's being moved into the background job
 - The metric and `debug!` log, since those need the full size of the map. 

In a way, that splits the old `fetch_states` more into two functions, but the semantics of these two blocks match:
- `schedule_fetch` selects which states to fetch
- `fetch_states` only runs the fetch

relay-server/src/actors/project_upstream.rs
olksdr: done in [462a093](https://github.com/getsentry/relay/pull/1727/commits/462a09371fb1cfb026641ae515672790b4b1c593)

relay-server/src/actors/project_upstream.rs
olksdr: done in [462a093](https://github.com/getsentry/relay/pull/1727/commits/462a09371fb1cfb026641ae515672790b4b1c593)

relay-server/src/actors/project_upstream.rs
olksdr: That's a very good point , done in [462a093](https://github.com/getsentry/relay/pull/1727/commits/462a09371fb1cfb026641ae515672790b4b1c593)

relay-server/src/actors/project_upstream.rs
olksdr: move this part into the separate function [462a093](https://github.com/getsentry/relay/pull/1727/commits/462a09371fb1cfb026641ae515672790b4b1c593)

relay-server/src/actors/project_upstream.rs
```diff
@@ -1,25 +1,34 @@
...
 
 use crate::actors::project::ProjectState;
 use crate::actors::project_cache::{FetchProjectState, ProjectError};
 use crate::actors::upstream::{RequestPriority, SendQuery, UpstreamQuery, UpstreamRelay};
 use crate::statsd::{RelayCounters, RelayHistograms, RelayTimers};
-use crate::utils::{self, ErrorBoundary};
+use crate::utils::{ErrorBoundary, SleepHandle};
+
+/// The map of project keys with their project state channels.
+type ProjectStateChannels = HashMap<ProjectKey, ProjectStateChannel>;
```
jan-auer: _nit_: Move typedef below definition of `ProjectStateChannel`.

relay-server/src/actors/project_upstream.rs
```diff
@@ -93,32 +103,73 @@ impl ProjectStateChannel {
...
+    pub fn send(self, state: ProjectState) {
+        self.channel.send(Arc::new(state))
     }
 
     pub fn expired(&self) -> bool {
         Instant::now() > self.deadline
     }
 }
 
+/// This is the [`UpstreamProjectSourceService`] interface.
```
jan-auer, line 121: _nit_: Since the interface describes the behavior, it would be more straight-forward for a reader to have the bulk of information here. Especially, it would be good to describe what the `UpstreamProjectSource` is and what it can do. I would not link to the service implementation here, but the other way around link to the interface from the service.

Of course, description on internals, such as buffer queues etc can and should remain on the service implementation.

relay-server/src/actors/project_upstream.rs
```diff
@@ -191,191 +230,240 @@ impl UpstreamProjectSource {
...
+                public_keys: channels_batch.keys().copied().collect(),
+                full_config: config.processing_enabled(),
+                no_cache: channels_batch.values().any(|c| c.no_cache),
+            };
+
+            // count number of http requests for project states
+            metric!(counter(RelayCounters::ProjectStateRequest) += 1);
+
+            let future_request = compat::send(UpstreamRelay::from_registry(), SendQuery(query))
+                .map(|response| match response {
```
jan-auer: _nit_: You can wrap the asynchronous parts of the loop into an `async` block to avoid using the future trait extensions (might need to be async move):

```rust
// ...

for channels_batch in cache_batches {
  // ...
  let query = GetProjectStates { .. };
  
  requests.push(async move {
    let request = compat::send(...);
    match request.await {
      Ok(response) => UpstreamResponse { .. },
      Err(_) => UpstreamResponse { .. },
    }
  });
}

let responses = future::join_all(requests).await;
// ...
```

jan-auer: Please check the other comments regarding this area first; this is more for completeness.

_nit_: The match can be pushed into the assignment to the `response` field. Note that you can even remove the type annotation on `Err`.

```rust
UpstreamResponse {
    channels_batch,
    response: match response {
        Ok(result) => result.map_err(ProjectError::UpstreamFailed),
        Err(_) => Err(ProjectError::ScheduleFailed),
    }
}
```

relay-server/src/actors/project_upstream.rs
```diff
@@ -191,191 +230,240 @@ impl UpstreamProjectSource {
...
+                        response: response.map_err(ProjectError::UpstreamFailed),
+                    },
+                    // Also propagate the channels in error case, since we will want to try to
+                    // fetch the projects again.
+                    _ => UpstreamResponse {
+                        channels_batch,
+                        response: Err::<GetProjectStatesResponse, ProjectError>(
+                            ProjectError::ScheduleFailed,
+                        ),
+                    },
```
jan-auer: This changes the behavior of the service. Unfortunately, the prior code was not entirely clear on this. Particularly `ProjectError::ScheduleFailed` was actually dead code (check it by removing `.map_err(|_| ProjectError::ScheduleFailed)` from master). 

There are two kinds of error that can happen:
 1. We fail to send a message to the upstream. This error is propagated through `join_all.and_then()`, so the handler doesn't run. In this case, we back off immediately and drop the affected channels. This, in turn, will automatically fail the project config request in the `ProjectCache`. We used to map this to `ProjectError::ScheduleFailed`, but this error was immediately dropped. So this error can be removed entirely.
 2. The upstream responds with an error. That went into a match statement, which resulted in a log and putting the channels back. 

Put together, I think we can remove `ProjectError` and simply return an `UpstreamError` here.

olksdr: > so the handler doesn't run. In this case, we back off immediately and drop the affected channels. 

So this also means we do not schedule fetches anymore, in case of the error `#1`:
* we never trigger `schedule_fetch`
* and because of https://github.com/getsentry/relay/blob/2a0976d7d9ecfd4daa6cb2390c224d09f9a79bbd/relay-server/src/actors/project_upstream.rs#L348 we will never trigger it on new message either

Do I see it correctly or am I missing something obvious?

jan-auer (approved):
:shipit:

PR #1729, opened 2023-01-09
build(py): Add sentry-pypi target

jjbayer:
As in https://github.com/getsentry/arroyo/pull/169, add a sentry-pypi target to the craft file to publish package to internal pypi.

#skip-changelog


olksdr (approved):
:shipit:

PR #1731, opened 2023-01-09
feat(replays): Use user configured PII scrubbers

cmanallen:
Removes statically defined scrubbers with data scrubbing and PII config from the project state. This includes a refactor that moves replay scrubbing on a struct.

closes: https://github.com/getsentry/replay-backend/issues/243


cmanallen:
@JoshFerge

jan-auer:
This will have to be updated to #1800 and requires a bit of cleanup in the `StringTranscoder`. I'll look into this.

relay-replays/src/recording.rs
```diff
@@ -43,104 +41,152 @@ impl From<serde_json::Error> for ParseRecordingError {
...
+    use relay_general::pii::PiiConfig;
+    use relay_general::processor::SelectorSpec;
+
+    use super::ReplayScrubber;
+
+    fn default_pii_config() -> PiiConfig {
+        let mut pii_config = PiiConfig::default();
+        pii_config.applications = [(SelectorSpec::And(vec![]), vec!["@common".to_string()])].into();
+        pii_config
+    }
```
jjbayer: To have the tests resemble the default production use case, I would use a configuration derived from `DataScrubbingConfig` here, like for example https://github.com/getsentry/relay/blob/afa6576707a77c45ed61b277f7900a0e2f835ba5/relay-general/src/protocol/replay.rs#L554-L560

jjbayer (changes requested):
The implementation looks good to me, just one remark about the test config.

relay-replays/src/recording.rs
jan-auer: Thanks, updated. This actually shows us that the filter has changed.

@cmanallen note that email addresses aren't being scrubbed anymore, apparently the default data scrubbers don't do this. No action item imho, this is in line with the other data scrubbers.

jjbayer (approved):

cmanallen:
@jjbayer @jan-auer Should this be merged?

jan-auer:
Yes. I was waiting for a response on https://github.com/getsentry/relay/pull/1731#discussion_r1094548564. 

Merging now.

jan-auer:
Yes. I was waiting for a response on https://github.com/getsentry/relay/pull/1731#discussion_r1094548564. 

Merging now.

PR #1733, opened 2023-01-09
Pick up self-hosted CI bugfix

chadwhitacre:
Picks up https://github.com/getsentry/action-self-hosted-e2e-tests/pull/3.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -370,7 +370,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v3
       - name: Run Sentry self-hosted e2e CI
-        uses: getsentry/action-self-hosted-e2e-tests@4ff2dbfdd5c986a4670efaa16d82e460598359eb
+        uses: getsentry/action-self-hosted-e2e-tests@ade3e646ffc3b6a765cfa18ec4ca720f710cb53d
         with:
           project_name: relay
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -370,7 +370,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v3
       - name: Run Sentry self-hosted e2e CI
-        uses: getsentry/action-self-hosted-e2e-tests@4ff2dbfdd5c986a4670efaa16d82e460598359eb
+        uses: getsentry/action-self-hosted-e2e-tests@ade3e646ffc3b6a765cfa18ec4ca720f710cb53d
```
hubertdeng123: bump commit here to `711694d0081a834777ca9c77c3f4c322ce7b08c4`

jjbayer: Which one is it? Should we use the latest commit on `main` like @hubertdeng123 suggests?

jjbayer (approved):

olksdr (approved):

.github/workflows/ci.yml
chadwhitacre: Either one will work, but tbh it will be good to confirm that https://github.com/getsentry/action-self-hosted-e2e-tests/commit/711694d0081a834777ca9c77c3f4c322ce7b08c4 does not regress in repos other than self-hosted. Bumped in 867666c972b96f3768ecffeb1938f7f12b0c6e5d.

PR #1734, opened 2023-01-10
feat(metrics): Count transactions toward root project [TET-627]

jjbayer:
This PR introduce new metric `count_per_root_project`, similar to `duration`,
but we associate it with root_project id and store as `sampling_metrics` vector (similar name we do in another places).

So we can use it in outcomes to know for example:

Frontend generates 8M (direct) + 120M (backend), so `count_per_root_project = 128M` for dependent projects.

or fallback to `count(duration)` == `count_per_root_project` for independent projects.

The metric is tagged with the transaction name *from the trace header*, which _should_ be low cardinality (SDKs should not set it when transaction source is `url`).

TODO:
- [x] Make it compile
- [x] Fix rust tests
- [x] Fix python integration tests

related PR https://github.com/getsentry/sentry/pull/43167


relay-server/src/actors/store.rs
```diff
@@ -950,7 +950,6 @@ struct MetricKafkaMessage {
     #[serde(flatten)]
     value: BucketValue,
     timestamp: UnixTimestamp,
-    #[serde(skip_serializing_if = "BTreeMap::is_empty")]
```
untitaker, line 958: i am reviewing this change only. i think we can revisit the schema we actually want later, but as pointed out in private conversations, the metrics indexer currently does not handle the absence of `tags` correctly.

jjbayer (commented):
The PR looks good to me. Let's make sure Search and Storage is aware that we are adding a counter metric before we deploy this.

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -499,6 +507,16 @@ fn extract_transaction_metrics_inner(
         tags_with_satisfaction.clone(),
     ));
 
+    // Count the transaction towards the root
+    sampling_metrics.push(Metric::new_mri(
+        METRIC_NAMESPACE,
```
andriisoldatenko, line 515: we need to think about internal metric?

andriisoldatenko (approved):

relay-server/src/actors/processor.rs
```diff
@@ -2156,8 +2197,12 @@ impl EnvelopeProcessorService {
                         // requires recomputation of the context.
                         state.envelope_context.update(&state.envelope);
 
+                        let has_metrics = state.extracted_metrics.project_metrics.is_empty();
```
```suggestion
                        let has_metrics = !state.extracted_metrics.project_metrics.is_empty();
```

jjbayer: I am surprised that no test case caught this.

tests/integration/test_metrics.py
```diff
@@ -919,11 +947,13 @@ def test_limit_custom_measurements(
...
 
-    # Expect exactly 3 metrics (transaction.duration, 1 builtin, 1 custom)
-    metrics = metrics_by_name(metrics_consumer, 3)
+    # Expect exactly 4 metrics:
+    # (transaction.duration, transactions.count_per_root_project, 1 builtin, 1 custom)
+    metrics = metrics_by_name(metrics_consumer, 4)
 
     assert metrics.keys() == {
         "d:transactions/duration@millisecond",
+        "c:transactions/count_per_root_project@none",
```
jjbayer, line 1059: As far as I can tell, these test cases all add this metric to current project, because there is no DSC involved. Please also add tests for the following:

* a test with a DSC where the DSC's project ID is different from the event's project ID.
* a test where we assert that the DSC's `transaction` field is used as a tag on the metric.

relay-server/src/actors/processor.rs
```diff
@@ -2156,8 +2197,12 @@ impl EnvelopeProcessorService {
...
+
+                        state.extracted_metrics.send_metrics(&state.envelope);
+
                         let envelope_response = if state.envelope.is_empty() {
-                            if state.extracted_metrics.is_empty() {
+                            if !has_metrics {
```
andriisoldatenko, line 2205: @jjbayer what about this line? 
if `has_metrics = !state.extracted_metrics.project_metrics.is_empty(); `
do we want to change it here?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Add error and sample rate fields to the replay event parser. ([#1745](https://github.com/getsentry/relay/pull/1745))
 - Add `instruction_addr_adjustment` field to `RawStacktrace`. ([#1716](https://github.com/getsentry/relay/pull/1716))
 - Add SSL support to `relay-redis` crate. It is possible to use `rediss` scheme to connnect to Redis cluster using TLS. ([#1772](https://github.com/getsentry/relay/pull/1772))
+- Add count transactions toward root project. ([#1734](https://github.com/getsentry/relay/pull/1734))
```
olksdr: Please, move this to `Unreleased` section.

relay-server/src/actors/processor.rs
```diff
@@ -763,7 +799,7 @@ impl EnvelopeProcessorService {
     /// are out of range after clock drift correction.
     fn process_sessions(&self, state: &mut ProcessEnvelopeState) {
         let received = state.envelope_context.received_at();
-        let extracted_metrics = &mut state.extracted_metrics;
+        let extracted_metrics = &mut &mut state.extracted_metrics.project_metrics;
```
olksdr: Doesn't this work just with taking one mutable reference? 

```suggestion
        let extracted_metrics = &mut state.extracted_metrics.project_metrics;
```

CHANGELOG.md
andriisoldatenko: @olksdr done.

relay-server/src/actors/processor.rs
andriisoldatenko: good catch! yeah it works without second `&mut`

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -391,7 +400,9 @@ fn extract_transaction_metrics_inner(
     aggregator_config: &AggregatorConfig,
     config: &TransactionMetricsConfig,
     event: &Event,
-    metrics: &mut Vec<Metric>, // output parameter
+    metrics: &mut Vec<Metric>,          // output parameter
+    sampling_metrics: &mut Vec<Metric>, // output parameter
+    transaction_from_dsc: Option<&str>,
```
jjbayer: nit: Because `metrics` and `sampling_metrics` are output parameters, I would put them at the end of the argument list.

relay-server/src/metrics_extraction/transactions.rs
jjbayer: @andriisoldatenko what internal metric did you mean here?

tests/integration/test_metrics.py
```diff
@@ -650,6 +659,79 @@ def test_transaction_metrics(
...
+    assert metrics["c:transactions/count_per_root_project@none"] == {
+        "timestamp": int(timestamp.timestamp()),
+        "org_id": 1,
+        "project_id": 42,
+        "retention_days": 90,
+        "tags": {"transaction": "transaction_which_starts_trace"},
+        "name": "c:transactions/count_per_root_project@none",
+        "type": "c",
+        "value": 3.0,
+    }
```
jjbayer, line 763: Isn't this use case already tested by `test_transaction_metrics`? Can we alter this test so that this metric is emitted with `project_id: 41`, where 41 is the root project. We could alter the signature of `send_transaction` to accept an explicit trace header:

`relay.send_transaction(42, transaction, trace_header={"public_key": public_key_of_project_41, "transaction": "root_transaction"})`

It would also be good to have one test case where `count_per_root_project` is emitted without a `transaction` tag (maybe I missed it).

relay-server/src/metrics_extraction/transactions.rs
andriisoldatenko: I think I've added this comment during our discussion, so are you ok with NAMESPACE? If yes we can resolve it.

tests/integration/test_metrics.py
andriisoldatenko: agree with you suggestion,

regarding:
>It would also be good to have one test case where count_per_root_project is emitted without a transaction tag (maybe I missed it).


I think we cover it separately in other tests where we can see `tags = {}`, but I can write separate test for that.

tests/integration/test_metrics.py
```diff
@@ -39,6 +40,21 @@ def metrics_by_name(metrics_consumer, count, timeout=None):
     return metrics
 
 
+def metrics_by_name_group_by_project(metrics_consumer, count, timeout=None):
```
jjbayer: This helper function does not seem to be using `count` at all. I feel like this function could call `metrics_by_name` internally, or vice versa?

tests/integration/test_metrics.py
jjbayer: Thanks, no need to add a separate test if it's already covered.

relay-server/src/metrics_extraction/transactions.rs
jjbayer: Yes, should be fine.

tests/integration/test_metrics.py
```diff
@@ -650,6 +675,87 @@ def test_transaction_metrics(
...
+            "version": 1,
+        }
+
+    transaction = generate_transaction_item()
+    transaction["timestamp"] = timestamp.isoformat()
+    transaction["measurements"] = {
+        "foo": {"value": 1.2},
+        "bar": {"value": 1.3},
+    }
+    relay.send_transaction(41, transaction, transaction_from_dsc="test")
```
jjbayer: Instead of sending this transaction to project 41, I would send it to project 42 with `project_id=41` in the DSC. And then verify that the metric has project 41, not 42. This is what I meant originally with

> a test with a DSC where the DSC's project ID is different from the event's project ID.

tests/integration/test_metrics.py
andriisoldatenko: count is huge antipattern IMO, I'll remove it from args :)

andriisoldatenko: let me explain, this helper has small problem:
```
def metrics_by_name(metrics_consumer, count, timeout=None):
    metrics = {}

    for _ in range(count):
        metric = metrics_consumer.get_metric(timeout)
        metrics[metric["name"]] = metric

    metrics_consumer.assert_empty()
    return metrics
```
if you have 2 projects and 2 metrics with the same name you will see only the last one, so that's why I decided to create a new helper function.

andriisoldatenko: And the second problem, you need manually to set the count - I prefer to get all metrics - since we have this assert_empty and then return data.

relay-server/src/metrics_extraction/transactions.rs
andriisoldatenko: fixed afa657670

jjbayer (commented):
I approve this PR. Please notify SnS when this change goes live, so they can observe the increase in stored metrics / tags.

iambriccardo (approved):
LGTM

olksdr (commented):
lgtm

olksdr (approved):

Issue #1639, opened 2022-11-25
Enumerations in public interfaces need a catch-all type


jan-auer:
- [x] https://github.com/getsentry/relay/pull/1638
- [x] Add unknown to `RuleType` and `SamplingMode` https://github.com/getsentry/relay/pull/1641
- [x] Check for more enumerations without a catch-all variant
- [x] Add `Other(u8)` to `SpanStatus` #1737 
       @jjbayer will check if we need different behavior in processing relays, and how.
- [x] Add `Unknown(String)` to `SessionStatus`. Do not extract metrics for unknown, but do forward session payloads. Remove in processing Relays. // pr: #1736 
- [x] Add `Unknown(String)` variant to `relay_profiling::MeasurementUnit`. Let Profiling team decide whether or not to drop unknowns in processing Relays. -> Not necessary right now because we don't parse profiles in non-processing relays.
- [x] `BucketValue`: Drop buckets with unknown type on deserialization. // bucketvalues are only sent from relay to relay, so no need forward compatibility in external relays
- [x] All PII config enums (see notion doc for details): skip rule on deserialization.
   Moved to separate issue: https://github.com/getsentry/relay/issues/1751
- [x] `RateLimitScope`: Skip unknown scopes (e.g. by skipping rule on deserialization). -> This already has a fallback via `QuotaScope`, see https://github.com/getsentry/relay/blob/337a8a0ed5c7f96e0c19f70757b9e54e2e99f7a5/relay-quotas/src/rate_limit.rs#L122-L123

jjbayer:
Found the following problematic enums (internal doc): https://www.notion.so/sentry/Forward-compatibility-of-enums-5eafecc16a014cbab13dbab3791ccef3

PR #1736, opened 2023-01-11
feat(general): Add unknown SessionStatus variant

TBS1996:
adds an unknown field to SessionStatus enum. This is for cases where a user has an outdated relay and receives an enum variant not in his version of relay. In those cases the name of the enum will be stored in a string and be passed on. 

#1639
#skip-changelog


relay-server/src/actors/processor.rs
```diff
@@ -2321,8 +2329,114 @@ mod tests {
...
 
+    type ProcessSessionArguments = (
+        Item,
+        DateTime<Utc>,
+        Option<&'static str>,
+        Option<net::IpAddr>,
+        SessionMetricsConfig,
+        ClockDriftProcessor,
+        Vec<Metric>,
+    );
```
olksdr: Is it used only in 1 test?

relay-server/src/actors/processor.rs
```diff
@@ -2321,8 +2329,114 @@ mod tests {
...
     use super::*;
 
+    // helper function that gives valid arguments to the process_session method
+    fn args_process_session() -> ProcessSessionArguments {
```
olksdr: This function a little bit confusing. And overall usage of tuple here as well. It might make sense create a proper struct for this and to be more explicit about the parameters used for test.

olksdr (commented):
Overall looks ok. 
The thing which bothers me it the way test is written, with a lot of indirections. 
I would personally prefer to have test be more explicit with to see what actually is tested. 

In current test I have to jump from one function to another, then to the  custom type, and back again to see what place in the huge tuple is what and what is changed.

I my opinion test should be written as explicitly as possible and be more descriptive, which also could serve as the example of how to use some pieces of code.

relay-server/src/actors/processor.rs
TBS1996: do you think it's possible to write the testing better without refactoring the code? I'm not happy with it either but the function is not easy to unit test as I'm checking like 5 different conditions where it can return false early. 

The function is used only once yes, I tried to explain it with the comment above it. Perhaps it is better to make it a closure inside of the test function similar to `should_keep`.  

I see what you mean about making a struct instead, I could make that change.

relay-server/src/actors/processor.rs
```diff
@@ -2316,13 +2324,131 @@ impl Service for EnvelopeProcessorService {
...
+                self.client_addr,
+                self.metrics_config,
+                &self.clock_drift_processor,
+                &mut self.extracted_metrics,
+            )
+        }
+    }
+
+    impl<'a> Default for TestProcessSessionArguments<'a> {
+        fn default() -> Self {
```
olksdr: You don't really have to implement the `Default` trait here, just move `default` function into `impl` block above.

relay-server/src/actors/processor.rs
```diff
@@ -2316,13 +2324,131 @@ impl Service for EnvelopeProcessorService {
...
+        received: DateTime<Utc>,
+        client: Option<&'a str>,
+        client_addr: Option<net::IpAddr>,
+        metrics_config: SessionMetricsConfig,
+        clock_drift_processor: ClockDriftProcessor,
+        extracted_metrics: Vec<Metric>,
+    }
+
+    impl<'a> TestProcessSessionArguments<'a> {
+        fn check_item_kept(&mut self) -> bool {
```
```suggestion
        fn run_session_processor(&mut self) -> bool {
```

It would be better to put in the name what exactly it does. And it process sessions.

relay-server/src/actors/processor.rs
```diff
@@ -2316,13 +2324,131 @@ impl Service for EnvelopeProcessorService {
...
+                clock_drift_processor: ClockDriftProcessor::new(None, received),
+                extracted_metrics: vec![],
+            }
+        }
+    }
+
+    /// Checks that the default test-arguments leads to the item being kept, which helps ensure the
+    /// other tests are valid
+    #[test]
+    fn test_keep_item() {
```
```suggestion
    fn test_process_session_keep_item() {
```

relay-server/src/actors/processor.rs
```diff
@@ -2316,13 +2324,131 @@ impl Service for EnvelopeProcessorService {
...
+        let mut args = TestProcessSessionArguments::default();
+        assert!(args.check_item_kept());
+    }
+
+    #[test]
+    fn test_invalid_json() {
```
```suggestion
    fn test_process_session_invalid_json() {
```

relay-server/src/actors/processor.rs
```diff
@@ -2316,13 +2324,131 @@ impl Service for EnvelopeProcessorService {
...
+        let mut args = TestProcessSessionArguments::default();
+        args.item
+            .set_payload(ContentType::Json, "this isnt valid json");
+        assert!(!args.check_item_kept());
+    }
+
+    #[test]
+    fn test_sequence_overflow() {
```
```suggestion
    fn test_process_session_sequence_overflow() {
```

relay-server/src/actors/processor.rs
```diff
@@ -2316,13 +2324,131 @@ impl Service for EnvelopeProcessorService {
...
+            },
+            "did": "user123",
+            "status": "this is not a valid status!",
+            "duration": 123.4
+        }"#,
+        );
+        assert!(!args.check_item_kept());
+    }
+    #[test]
+    fn test_invalid_timestamp() {
```
```suggestion
    fn test_process_session_invalid_timestamp() {
```

relay-server/src/actors/processor.rs
```diff
@@ -2316,13 +2324,131 @@ impl Service for EnvelopeProcessorService {
...
+        let mut args = TestProcessSessionArguments {
+            received: DateTime::from_str("2021-05-26T08:00:00+0100").unwrap(),
+            ..Default::default()
+        };
+
+        assert!(!args.check_item_kept());
+    }
+
+    #[test]
+    fn test_metrics_extracted() {
```
```suggestion
    fn test_process_session_metrics_extracted() {
```

relay-server/src/actors/processor.rs
```diff
@@ -2316,13 +2324,131 @@ impl Service for EnvelopeProcessorService {
...
+            },
+            "did": "user123",
+            "status": "this is not a valid status!",
+            "duration": 123.4
+        }"#;
+
+            item.set_payload(ContentType::Json, session);
+            let received = DateTime::from_str("2021-04-26T08:00:00+0100").unwrap();
+
+            TestProcessSessionArguments {
```
```suggestion
            Self {
```

relay-server/src/actors/processor.rs
```diff
@@ -2403,13 +2411,126 @@ impl Service for EnvelopeProcessorService {
...
+    }
+    #[test]
+    fn test_process_session_invalid_timestamp() {
+        let mut args = TestProcessSessionArguments::default();
+        args.received = DateTime::from_str("2021-05-26T08:00:00+0100").unwrap();
+        assert!(!args.run_session_producer());
+    }
+
+    #[test]
+    fn test_process_sessionprocess_session_metrics_extracted() {
```
```suggestion
    fn test_process_session_metrics_extracted() {
```

relay-general/src/protocol/session.rs
```diff
@@ -25,6 +24,8 @@ pub enum SessionStatus {
     Abnormal,
     /// The session exited cleanly but experienced some errors during its run.
     Errored,
+    /// Unknown status, for forward compatibility
```
olksdr:
```suggestion
    /// Unknown status, for forward compatibility.
```

relay-server/src/actors/processor.rs
```diff
@@ -2403,13 +2411,126 @@ impl Service for EnvelopeProcessorService {
...
+                )
+                .unwrap(),
+                clock_drift_processor: ClockDriftProcessor::new(None, received),
+                extracted_metrics: vec![],
+            }
+        }
+    }
+
+    /// Checks that the default test-arguments leads to the item being kept, which helps ensure the
+    /// other tests are valid
```
```suggestion
    /// other tests are valid.
```

olksdr (approved):
Few small things, like typo in the tests name.

Otherwise LGTM.

PR #1737, opened 2023-01-11
feat(common): Add catch-all status code variant for SpanStatus

TBS1996:
#skip-changelog

Adds a status to SpanStatus that can hold an unknown status code. Useful for forward compatibility.

#1639

### DIFF
--- a/relay-common/src/constants.rs
+++ b/relay-common/src/constants.rs
@@ -305,6 +305,9 @@ pub enum SpanStatus {
     ///
     /// Prefer PermissionDenied if a user is logged in.
     Unauthenticated = 16,
+
+    /// Other code for forward compatibility
+    Other(u8),
 }
 
 /// Error parsing a `SpanStatus`.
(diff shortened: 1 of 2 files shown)


jjbayer:
What would have to happen here is that for external relays, we pass any string on to the next relay, but for processing relays, we set an error on the event. But it's tricky to get access to the `processing_enabled` flag from deserialization.

Because we already have error handling in `FromValue`, the worst thing that can happen if we keep things as-is is that transactions are reported with the wrong span status. That is, the event is not actually dropped, and the error is even reported through meta (see linked code). I think that in such a case we can simply ask the user hosting the external relay to upgrade.

TL;DR: I suggest we close this PR and leave the behavior as-is.

https://github.com/getsentry/relay/blob/37f965b9747fd00ba4f3b01be378279bb6704687/relay-general/src/protocol/contexts/trace.rs#L127-L131

PR #1741, opened 2023-01-12
feat(config): Add unknown ruletype

TBS1996:
Adds a new enum variant to ruletype for forward-compatibility

#1639 

#skip-changelog


relay-general/src/pii/regexes.rs
```diff
@@ -88,6 +88,7 @@ pub fn get_regex_for_rule_type(
         RuleType::UsSsn => smallvec![(v, &*US_SSN_REGEX, ReplaceBehavior::replace_match())],
         RuleType::Userpath => smallvec![(v, &*PATH_REGEX, ReplaceBehavior::replace_group(1))],
 
+        RuleType::Unknown(_) => smallvec![],
```
jjbayer: I think there is one more match statement we need to change:

collect_rules: Skip unknown rules instead of inserting them:

https://github.com/getsentry/relay/blob/37f965b9747fd00ba4f3b01be378279bb6704687/relay-general/src/pii/compiledconfig.rs#L82-L84

jjbayer: nit: Let's put `RuleType::Unknown(_)` into the same match branch as `RuleType::Alias` and `RuleType::Multiple`. The comment "these ought to have been resolved in CompiledConfig" apply to all of them.

jjbayer (approved):
See single nitpick, apart from that this looks good to go!

PR #1678, opened 2022-12-06
feat(replays): Enable PII scrubbing for all organizations

cmanallen:
#skip-changelog

PR #1742, opened 2023-01-12
instr(replays): Add timer metric to recording processing

jjbayer:
We noticed an increase in total processing time which we suspect was caused by https://github.com/getsentry/relay/pull/1678.

Add a statsd metric to measure replay recording processing in isolation.

#skip-changelog

### DIFF
--- a/relay-server/src/statsd.rs
+++ b/relay-server/src/statsd.rs
@@ -298,6 +298,9 @@ pub enum RelayTimers {
 
     /// Time in milliseconds spent on converting a transaction event into a metric.
     TransactionMetricsExtraction,
+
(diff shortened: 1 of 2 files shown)


relay-server/src/statsd.rs
```diff
@@ -298,6 +298,9 @@ pub enum RelayTimers {
 
     /// Time in milliseconds spent on converting a transaction event into a metric.
     TransactionMetricsExtraction,
+
+    /// Time in milliseconds spent on parsing, normalizing and scrubbing Relay recordings.
```
```suggestion
    /// Time in milliseconds spent on parsing, normalizing and scrubbing replay recordings.
```

olksdr (approved):
lgtm!

PR #1744, opened 2023-01-13
feat: Add CrashType as a tag to allow filtering

### LINKED ISSUE #38550: UnrealEngine: Add CrashType as a tag to allow filtering
### Problem Statement

We use a self-hosted Sentry to collect crash reports from UnrealEngine using the CrashReporterClient (and it's been great so far!).
We added various tags through the CrashContext GameData field as suggested [here](https://docs.sentry.io/platforms/unreal/setup-crashreport/#configure-attributes).

However, there is several crash types (from the enum ECrashContextType) which produces CrashReports and we'd like to be able to filter on them (for example, `Stall` are less time critical than crashes).
It's not possible to add them to the CrashContext GameData easily so we'd like to look into doing this on the Sentry side.

I'm open to look into this myself but wanted to know if it's something that the sentry team is open to and get some pointer to where to look to implement this.

### Solution Brainstorm

This information is present in the `CrashContext.runtime-xml` which is sent with issues to the UE endpoint under the field `FGenericCrashContext` -> `RuntimeProperties` -> `CrashType`.
Retrieving this string and adding it as a tag would allow users to filter on it.

I'm unfamiliar with Sentry so I'm not able to propose where this could be done.

DamienDelmas:
Here's a small PR to fix [issue 38550](https://github.com/getsentry/sentry/issues/38550).

The unreal crash reporter gives some information about the type of crash that occured ( stall, crash, ...). We'd like to be able to filter crashes based on this information.

As discussed in the issue linked above, this PR adds the crash type to event tags.

I couldn't find much information about PR guidelines and it's my first commit in the Sentry codebase so, please, let me know if I can improve this PR in any way :)

### DIFF
--- a/relay-server/src/utils/unreal.rs
+++ b/relay-server/src/utils/unreal.rs
@@ -234,6 +234,14 @@ fn merge_unreal_context(event: &mut Event, context: Unreal4Context) {
         event.level = Annotated::new(Level::Warning)
     }
 
+    if let Some(crash_type) = runtime_props.crash_type.take() {
+        let tags = event.tags.value_mut().get_or_insert_with(Tags::default);
+        tags.push(Annotated::new(TagEntry::from_pair((
+            Annotated::new("crash_type".to_string()),
+            Annotated::new(crash_type),
+        ))));
+    }
+
     // Modules are not used and later replaced with Modules from the Minidump or Apple Crash Report.
     runtime_props.modules.take();
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 **Features**:
 
+- Add Unreal crash type as an event tag ([#1744](https://github.com/getsentry/relay/pull/1744))
 - Add error and sample rate fields to the replay event parser. ([#1745](https://github.com/getsentry/relay/pull/1745))
 
 ## 23.1.0


relay-server/src/utils/unreal.rs
```diff
@@ -234,6 +234,14 @@ fn merge_unreal_context(event: &mut Event, context: Unreal4Context) {
         event.level = Annotated::new(Level::Warning)
     }
 
+    if let Some(crash_type) = runtime_props.crash_type.take() {
+        let tags = event.tags.value_mut().get_or_insert_with(Tags::default);
+        tags.push(Annotated::new(TagEntry::from_pair((
+            Annotated::new("crash_type".to_string()),
+            Annotated::new(crash_type),
+        ))));
+    }
```
olksdr, line 243: I've  triggered CI to test this change. 
There might be some tests needed  just to check that everything set properly.

olksdr (commented):
LGTM.

But also I will have to bring this to the team to see if this is something we ok with. 

/cc @jan-auer

jan-auer:
We should not be adding tags for this. Instead, index the context field in storage, please. 
cc @bitsandfoxes 

**Edit:** Adding a bit more context. First of all, thank you for the contribution @DamienDelmas! In the past, we have in fact extracted tags from data to make them searchable. In modern Sentry, this is not the best option we have for a variety of reasons:
 - Tags go into a generic data store, which cannot provide optimal performance since it has to assume the common case.
 - User-defined tags share the same namespace, so this information can overwrite each other.
 - There is no typing on tags, everything is just a string (not an issue here).

For this reason, we nowadays opt to index information directly from contexts, which makes it available as fields in Issue search and Discover. This requires changes in the Sentry backend and storage. While this takes a little more effort, it's also the more sustainable solution. 

@bitsandfoxes was looking into this recently and can probably share more information.

bitsandfoxes:
We looked into the way these events get processed and since the Unreal Crash Type is already part of the context, it is searchable. We're just not yet making that accessible yet.
I'll make sure to get this done in a timely manner.

DamienDelmas:
Thank you for the explaination @jan-auer ! And thank you for your work on this feature @bitsandfoxes 😊

PR #1746, opened 2023-01-14
feat(codeowners): Find codeowner path matches with rust

NisanthanNanthakumar:
Proof of Concept PR

Benchmarked this solution agains the baseline python implementation https://github.com/getsentry/sentry/pull/43294

This code is benchmarking ~3 times faster than the baseline python implementation. So it is ready to be merged into Sentry.

### DIFF
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add utility function for matching CODEOWNER paths against a stacktrace filepath ([#1746](https://github.com/getsentry/relay/pull/1746))
+
 ## 0.8.16
 
 - The minimum required Python version is now 3.8. This release does not contain known breaking changes for Python 3.7, but we no longer guarantee compatibility.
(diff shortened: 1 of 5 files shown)


jjbayer:
@NisanthanNanthakumar I did not look at your benchmark in detail, but in my experience running a benchmark with a global LRU cache can easily skew results, because you get much better cache locality in the experiment than in production. Have you considered using pre-compiled regexes on the Python side instead?

NisanthanNanthakumar:
@jjbayer yea I realized that too. Thats why the plan is to push to production under a feature flag and run it against a sampled subset of the `sentry` org events and compare the performance between the two implementations in prod. 

WRT the pre-compiled regexes in Python suggestion, the `re` package already handles that caching optimization here: https://github.com/python/cpython/blob/5a8ed019f9b826d2ba5766688d51005624cd0699/Lib/re/__init__.py#L302

snigdhas (approved):
Thanks for walking me through it!

jjbayer (approved):
I did not check the pattern translation logic in detail, but the rest looks good to me!

We will have to [release](https://github.com/getsentry/relay#release-management) a new version of the Python library if you want to run this in production.

PR #1748, opened 2023-01-16
feat(profiling): Add profile context

Zylphrex:
This change introduces the profile context as per
https://github.com/getsentry/rfcs/blob/main/text/0047-introduce-profile-context.md.

### DIFF
--- a/relay-general/src/protocol/contexts/profile.rs
+++ b/relay-general/src/protocol/contexts/profile.rs
@@ -1,0 +1,83 @@
+use crate::types::{Annotated, Error, FromValue, Value};
+use once_cell::sync::OnceCell;
+use regex::Regex;
+
+#[derive(Clone, Debug, Default, PartialEq, Empty, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct ProfileId(pub String);
+
+impl FromValue for ProfileId {
+    fn from_value(value: Annotated<Value>) -> Annotated<Self> {
+        match value {
+            Annotated(Some(Value::String(value)), mut meta) => {
+                static PROFILE_ID: OnceCell<Regex> = OnceCell::new();
+                let regex = PROFILE_ID.get_or_init(|| Regex::new("^[a-fA-F0-9]{32}$").unwrap());
+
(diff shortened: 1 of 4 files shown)


relay-general/src/protocol/contexts/profile.rs
```diff
@@ -0,0 +1,83 @@
+use crate::types::{Annotated, Error, FromValue, Value};
+use once_cell::sync::OnceCell;
+use regex::Regex;
+
+#[derive(Clone, Debug, Default, PartialEq, Empty, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct ProfileId(pub String);
```
jjbayer: In other places, `profile_id` is declared as `EventId`, which would make the custom `FromValue` implementation unnecessary AFAIK.

jjbayer: See for example https://github.com/getsentry/relay/blob/37f965b9747fd00ba4f3b01be378279bb6704687/relay-profiling/src/android.rs#L37.

jjbayer (approved):

relay-general/src/protocol/contexts/profile.rs
```diff
@@ -0,0 +1,58 @@
+use crate::types::Annotated;
+
+use crate::protocol::EventId;
+
+/// Profile context
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+// #[metastructure(process_func = "process_trace_context")]
```
jjbayer: Remove this line?

relay-general/src/protocol/contexts/profile.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct ProfileContext {
+    /// The profile ID.
+    #[metastructure(required = "true")]
+    pub profile_id: Annotated<EventId>,
+}
+
+impl ProfileContext {
+    /// The key under which a trace context is generally stored (in `Contexts`)
```
jjbayer:
```suggestion
    /// The key under which a profile context is generally stored (in `Contexts`)
```

PR #1749, opened 2023-01-16
profiling(fix): use an unpadded base64 encoding

viglia:
Recently we switched to a new crate for `base64` data encoding/decoding (#1743).

In doing so, a regression that discards unpadded base64 profiles was introduced.

This PR restores the expected behaviour. 

#skip-changelog


Zylphrex (approved):

viglia:
#skip-changelog

Issue #1183, opened 2022-02-17
Support User-Agent Client Hints


lucas-zimerman:
- [x] Review the documentation: https://docs.sentry.io/
- [x] Search for existing issues: https://github.com/getsentry/sentry-javascript/issues
- [x] Use the latest release: https://github.com/getsentry/sentry-javascript/releases
- [x] Provide a link to the affected event from your Sentry account

## Package + Version

- [X] `@sentry/browser`
- [ ] `@sentry/node`
- [ ] `raven-js`
- [ ] `raven-node` _(raven for node)_
- [ ] other:

### Version:
```
6.17.9
```

## Description
Sentry seems to be capturing the Windows 11 platform as Windows 10.

Sample code:
```javascript
<script>
Sentry.init({
  dsn: "https://d16ae2d36f9249849c7964e9a3a8a608@o447951.ingest.sentry.io/5429213",
});
Sentry.CaptureMessage("Hello World");
</script>
```
Sample event:
https://sentry.io/organizations/sentry-sdks/issues/3024003271/events/0e4c3c7c7de840a889b18d67ca15e473/?project=5429213

AbhiPrasad:
This is being grabbed from the user-agent, and relay actually populates the browser value based on the user-agent of the client that sent the event.

`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.102 Safari/537.36 Edg/98.0.1108.55`

Interesting - it seems like Windows team made it so that user agents are frozen on Windows 10, and you need to use User-Agent Client Hints to check if it's coming from Windows 11. https://docs.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11

I've raised a ticket internally as it seems Relay does not support https://wicg.github.io/ua-client-hints/, and will transfer this ticket to the Relay repo to track there.

AbhiPrasad:
Obsolete Sentry internal JIRA ticket: https://getsentry.atlassian.net/browse/INGEST-1422

AbhiPrasad:
Sentry internal DACI: https://www.notion.so/sentry/Handling-of-new-client-hints-to-parse-user-agent-information-3590fcbbdbd542faa6fa07ba67c0ddb3

jan-auer:
Acceptance criteria that was discussed about this internally:

[...]

PR #1752, opened 2023-01-17
feat: Add clienthints support

TBS1996:
stuff i've done:

* made RawUserAgentInfo struct which holds both the "legacy" UA string as well as all the client hint headers
* made a FromUserAgentInfo trait where you pass in that struct and it tries to first get the Context first from the client hints, and if that fails it tries from the UA string. 
* made a very basic parser for browser version (SEC-CH-UA header)
* unit tests

#1183 

This PR is limited only to getting client hints from request headers and the logic of when to use that info rather than the user agent string. Check issue for next steps.


relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,233 @@ impl BrowserContext {
...
+        Some(Self {
+            name: Annotated::from(browser.family),
+            version: Annotated::from(get_version(&browser.major, &browser.minor, &browser.patch)),
+            ..BrowserContext::default()
+        })
+    }
+}
+
+#[derive(Debug, PartialEq)]
+enum Browser {
```
TBS1996: Not really useful at the moment... if there will be no use for a proper browser enum then I will remove this

relay-general/src/user_agent.rs
```diff
@@ -49,14 +81,25 @@ pub fn init_parser() {
...
 
-/// Returns the user agent string from a `Request`.
-///
-/// Returns `Some` if the event's request interface contains a `user-agent` header. Returns `None`
-/// otherwise.
-pub fn get_user_agent(request: &Annotated<Request>) -> Option<&str> {
-    let request = request.value()?;
-    let headers = request.headers.value()?;
-    get_user_agent_from_headers(headers)
+pub fn get_user_agent_from_request(request: &Annotated<Request>) -> Option<&str> {
```
TBS1996: had to change in a previous commit where it didn't return an option. Can return back if previous is considered cleaner

relay-general/src/store/normalize/user_agent.rs
```diff
@@ -6,45 +6,41 @@
...
 
-use crate::protocol::{BrowserContext, Context, Contexts, DeviceContext, Event, OsContext};
+use crate::protocol::{
+    BrowserContext, Context, ContextFromUserAgentInfo, Contexts, DeviceContext, Event, OsContext,
+};
 use crate::types::Annotated;
-use crate::user_agent::{get_user_agent, parse_device, parse_os, parse_user_agent};
+use crate::user_agent::RawUserAgentInfo;
 
 pub fn normalize_user_agent(event: &mut Event) {
```
TBS1996, line 15: simplified this function, should be same functionality

relay-general/src/protocol/contexts/device.rs
```diff
@@ -169,9 +172,83 @@ impl DeviceContext {
     }
 }
 
-#[test]
-fn test_device_context_roundtrip() {
-    let json = r#"{
+//"moto g31(w)"
+//"Mozilla/5.0 (Linux; Android 11; moto g31(w)) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Mobile Safari/537.36"
```
jjbayer:
```suggestion
```

relay-general/src/protocol/contexts/mod.rs
```diff
@@ -87,6 +88,19 @@ impl Context {
     }
 }
 
+/// Trait to get the Context both from the user agent string and also the new client hints.
+/// With an automatically derived function which tries to first get the context from client hints,
+/// if that fails it tries for the user agent string.
+pub trait ContextFromUserAgentInfo: Sized {
```
jjbayer: nit: This trait could be implemented for things other than Contexts, so no need to put `Context` in the name.

```suggestion
pub trait FromUserAgentInfo: Sized {
```

relay-general/src/protocol/contexts/mod.rs
```diff
@@ -87,6 +88,19 @@ impl Context {
...
+    fn from_client_hints(raw_contexts: &RawUserAgentInfo) -> Option<Self>;
+    fn from_user_agent(user_agent: &str) -> Option<Self>;
+
+    fn from_hints_or_ua(raw_contexts: &RawUserAgentInfo) -> Option<Self> {
+        Self::from_client_hints(raw_contexts)
+            .or_else(|| raw_contexts.user_agent.and_then(Self::from_user_agent))
```
jjbayer:
```suggestion
    fn from_client_hints(raw_info: &RawUserAgentInfo) -> Option<Self>;
    fn from_user_agent(user_agent: &str) -> Option<Self>;

    fn from_hints_or_ua(raw_info: &RawUserAgentInfo) -> Option<Self> {
        Self::from_client_hints(raw_info)
            .or_else(|| raw_info.user_agent.and_then(Self::from_user_agent))
```

relay-general/src/user_agent.rs
```diff
@@ -26,18 +26,50 @@ static UA_PARSER: Lazy<UserAgentParser> = Lazy::new(|| {
     )
 });
 
-fn get_user_agent_from_headers(headers: &Headers) -> Option<&str> {
-    for item in headers.iter() {
-        if let Some((ref o_k, ref v)) = item.value() {
-            if let Some(k) = o_k.as_str() {
-                if k.to_lowercase() == "user-agent" {
-                    return v.as_str();
+/// this has both the user agent string and the client hints, useful for the scenarios where
```
jjbayer:
```suggestion
/// This has both the user agent string and the client hints, useful for the scenarios where
```

relay-general/src/user_agent.rs
```diff
@@ -26,18 +26,50 @@ static UA_PARSER: Lazy<UserAgentParser> = Lazy::new(|| {
...
+    /// your OS, e.g. macos, android ..
+    pub sec_ch_ua_platform: Option<&'a str>,
+    /// the version number of your OS
+    pub sec_ch_ua_platform_version: Option<&'a str>,
+    /// web browser
+    pub sec_ch_ua: Option<&'a str>,
+    /// web browser version
+    pub sec_ch_ua_full_version: Option<&'a str>,
+    /// device model, e.g. samsung galaxy 3
+    pub sec_ch_ua_model: Option<&'a str>,
```
jjbayer: I would put these properties in a separate `ClientHints` struct, so you end up with a nested

```rust
struct RawUserAgentInfo {
    pub user_agent: Option<&'a str>,
    pub client_hints: Option<ClientHints<'a>>,
}
```

 and then change the signature of `from_client_hints` to accept `ClientHints`. That way, you only pass into `from_client_hints` what that function actually needs.

relay-general/src/user_agent.rs
jjbayer: Since this is a public function, please document it.

relay-general/src/user_agent.rs
```diff
@@ -26,18 +26,50 @@ static UA_PARSER: Lazy<UserAgentParser> = Lazy::new(|| {
...
-                if k.to_lowercase() == "user-agent" {
-                    return v.as_str();
+/// this has both the user agent string and the client hints, useful for the scenarios where
+/// you will use either information from client hints if it exists and if not fall back to
+/// user agent string.
+///
+/// The client hint variable names mirror the name of the headers. '<https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#user_agent_client_hints>'
+#[derive(Default, Debug)]
+pub struct RawUserAgentInfo<'a> {
+    /// the "old style" of a single UA string
```
jjbayer: We always start with a capital letter and end with punctuation.

```suggestion
    /// The "old style" of a single UA string.
```

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,233 @@ impl BrowserContext {
...
+/// the sec-ch-ua field looks something like this:
+/// "Not_A Brand";v="99", "Google Chrome";v="109", "Chromium";v="109"
+///
+/// this function attempts to detect the field with the browser, parse it as a browser enum and
+/// return that. if it fails to parse it as a browser, it uses the "other" variant of browser enum
+///
+/// returns None if no browser field detected
+fn parse_client_hint_browser<S: Into<String>>(s: S) -> Option<Browser> {
+    let s = s.into();
+    let items: Vec<&str> = s.split(',').collect();
```
jjbayer: Do you need to call `.collect()` here? It looks like an iterator would do fine.

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,233 @@ impl BrowserContext {
...
+
+    for item in items {
+        let item = item.to_lowercase();
+
+        // if it contains one of these then we can know it isn't a browser field. atm chromium
+        // browsers are the only ones supporting client hints.
+        if item.contains("brand")
```
jjbayer: Instead of working with `contains`, I would split by `;` and strip the `""` first. After that, you can do `==` checks.

jjbayer (commented):
Just a partial review, will review again after the next round of changes.

relay-general/src/protocol/contexts/browser.rs
TBS1996: the "not a brand" field is randomly changed, but it seems to always have the word "brand" so I think `contains` is the best way. heres a few examples:

`sec-ch-ua: "Google Chrome";v="87", " Not;A Brand";v="99", "Chromium";v="87"`
`"Chromium";v="86", "\"Not\\A;Brand";v="99", "Google Chrome";v="86"`

relay-general/src/protocol/contexts/browser.rs
TBS1996: neat, good catch

relay-general/src/user_agent.rs
TBS1996: Do you think ClientHints should be an option here? All its fields would be options so if there's no distinctions between it being None or it having empty fields then i think it would make it less simple this way

CHANGELOG.md
```diff
@@ -5,9 +5,11 @@
 **Features**:
 
 - Add error and sample rate fields to the replay event parser. ([#1745](https://github.com/getsentry/relay/pull/1745))
+- Add support for client hints. ([#1752](https://github.com/getsentry/relay/pull/1752))
 - Add `instruction_addr_adjustment` field to `RawStacktrace`. ([#1716](https://github.com/getsentry/relay/pull/1716))
 - Add SSL support to `relay-redis` crate. It is possible to use `rediss` scheme to connnect to Redis cluster using TLS. ([#1772](https://github.com/getsentry/relay/pull/1772))
 
+
```
jjbayer: Intentional newline?

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,223 @@ impl BrowserContext {
...
+}
+
+/// the sec-ch-ua field looks something like this:
+/// "Not_A Brand";v="99", "Google Chrome";v="109", "Chromium";v="109"
+///
+/// this function attempts to detect the field with the browser, parse it as a browser enum and
+/// return that. if it fails to parse it as a browser, it uses the "other" variant of browser enum
+///
+/// returns None if no browser field detected
+fn parse_client_hint_browser<S: Into<String>>(s: S) -> Option<Browser> {
```
olksdr: I would put this function in the `enum Browser` implementation and call it `parse` or maybe `try_from_str` in this case, which would communicated better the intention.

And why not just work with `&str` here as incoming parameter?

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -1,4 +1,7 @@
+use crate::protocol::FromUserAgentInfo;
+use crate::store::user_agent::{get_version, is_known};
 use crate::types::{Annotated, Object, Value};
+use crate::user_agent::{parse_user_agent, ClientHints};
```
olksdr: When looking on the usage these imported functions, it's sometimes not clear where they come from. It would be nicer if we always specify the module when the call the function , e.g. `user_agent::get_version()`

relay-general/src/protocol/contexts/device.rs
```diff
@@ -169,9 +172,80 @@ impl DeviceContext {
...
+        Some(Self {
+            family: Annotated::new(device),
+            ..Default::default()
+        })
+    }
+
+    fn from_user_agent(user_agent: &str) -> Option<Self> {
+        let device = parse_device(user_agent);
+
+        if !is_known(device.family.as_str()) {
```
olksdr: Will this work if you use following ?
```suggestion
        if !is_known(&device.family) {
```

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,223 @@ impl BrowserContext {
...
+            name: Annotated::new(browser.to_string()),
+            version: Annotated::new(version),
+            ..Default::default()
+        })
+    }
+
+    fn from_user_agent(user_agent: &str) -> Option<Self> {
+        let browser = parse_user_agent(user_agent);
+
+        if !is_known(browser.family.as_str()) {
```
olksdr:
```suggestion
        if !is_known(&browser.family) {
```

relay-general/src/protocol/contexts/os.rs
```diff
@@ -47,9 +49,94 @@ impl OsContext {
...
+            name: Annotated::new(platform.to_owned()),
+            version: Annotated::new(version.to_owned()),
+            ..Default::default()
+        })
+    }
+
+    fn from_user_agent(user_agent: &str) -> Option<Self> {
+        let os = parse_os(user_agent);
+
+        if !is_known(os.family.as_str()) {
```
olksdr:
```suggestion
        if !is_known(&os.family) {
```

relay-general/src/user_agent.rs
```diff
@@ -49,14 +91,26 @@ pub fn init_parser() {
...
-pub fn get_user_agent(request: &Annotated<Request>) -> Option<&str> {
-    let request = request.value()?;
-    let headers = request.headers.value()?;
-    get_user_agent_from_headers(headers)
+/// Gets the user agent string from request header type.
+pub fn get_user_agent_from_request(request: &Annotated<Request>) -> Option<&str> {
+    request
+        .value()
+        .and_then(|request| request.headers.value())
+        .and_then(|headers| get_user_agent_from_headers(headers))
```
olksdr: Since you get only one header name here, I believe `Headers` struct has convenient method `get_header` https://github.com/getsentry/relay/blob/8ffbc1b9dc1bcdffbe4c3ae998ae1bb0d4875c30/relay-general/src/protocol/request.rs#L240
which does exactly the same as the method below `get_suer_agent_from_headers`

relay-general/src/protocol/contexts/browser.rs
TBS1996: good idea as implementation, for &str I first thought i needed string for mutation but now realize that wasn't true, so I changed it now.

relay-general/src/user_agent.rs
TBS1996: a lot of tests rely on that get_user_agent_from_headers(..) is makes the header key into lowercase.  Should I add to_lowercase() on get_header? or leave it as it is?

relay-filter/src/legacy_browsers.rs
```diff
@@ -9,7 +9,7 @@ use crate::{FilterStatKey, LegacyBrowser, LegacyBrowsersFilterConfig};
 
 /// Checks if the event originates from legacy browsers.
 pub fn matches(event: &Event, browsers: &BTreeSet<LegacyBrowser>) -> bool {
-    if let Some(user_agent_string) = user_agent::get_user_agent(&event.request) {
+    if let Some(user_agent_string) = user_agent::get_user_agent_from_request(&event.request) {
```
jjbayer: Why did you rename this function? IMO it adds unnecessary diffs to the PR.

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,229 @@ impl BrowserContext {
         "browser"
     }
 }
+
+impl FromUserAgentInfo for BrowserContext {
+    fn from_client_hints(client_hints: &user_agent::ClientHints) -> Option<Self> {
+        let browser = Browser::from_ua_client_hint(client_hints.sec_ch_ua?)?;
+        let version = client_hints.sec_ch_ua_full_version?.to_owned();
```
jjbayer: `sec-ch-ua-full-version` seems to be [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Full-Version). The [significant version](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA#syntax) is probably more relevant to our users than the full version anyway, so should we just parse that from the `sec-ch-ua` hint?

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,229 @@ impl BrowserContext {
...
+            }
+
+            // "\"foo-browser\";v=\"109\"" -> "foo-browser"
+            let browser: String = item
+                .split(';')
+                .take(1)
+                .map(|s| s.trim().to_owned().replace('\"', ""))
+                .collect();
+
+            return Some(Self::from(browser));
```
jjbayer: As mentioned above, I would also return the version here.

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,229 @@ impl BrowserContext {
...
+
+impl std::string::ToString for Browser {
+    fn to_string(&self) -> String {
+        match self {
+            Browser::Chrome => "Google Chrome".to_string(),
+            Browser::Edge => "Microsoft Edge".to_string(),
+            Browser::Firefox => "Firefox".to_string(),
+            Browser::Safari => "Safari".to_string(),
+            Browser::Brave => "Brave".to_string(),
+            Browser::Opera => "Opera".to_string(),
```
jjbayer: There's a few more browsers that already support `sec-ch-ua`, let's add them to this list. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA#browser_compatibility.

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +24,229 @@ impl BrowserContext {
...
+        for item in s.split(',') {
+            // if it contains one of these then we can know it isn't a browser field. atm chromium
+            // browsers are the only ones supporting client hints.
+            if item.contains("Brand")
+            || item.contains("Chromium")
+            || item.contains("Gecko") // useless until firefox and safari support client hints
+            || item.contains("WebKit")
+            {
+                continue;
+            }
```
jjbayer: By only skipping known non-browsers, we risk accepting an unknown non-browser as browser. I would go the other way around, and only return something from this function if the browser is part of the `Browser` enum.

Then you could easily split on `;` first, to avoid the use of `contains`.

jjbayer (commented):
This looks good overall, and the test coverage seems solid! Just a couple more concerns.

relay-filter/src/legacy_browsers.rs
TBS1996: i thought it was more clear as there is a `get_user_agent_from_headers`

relay-general/src/protocol/contexts/browser.rs
TBS1996: oh good catch, yes I suppose the from_ua_client_hint should return a tuple with browser and version then

CHANGELOG.md
TBS1996: no, ill revert

relay-general/src/protocol/contexts/browser.rs
TBS1996: do you know about a list of those browsers as they appear on the client hints? For example it says `Chrome` there but in the client hints it's `Google Chrome`, and I need it to be exact for parsing

relay-general/src/protocol/contexts/browser.rs
TBS1996: this depends on whether it's worse to accidentally include a non-browser, or to wrongly exclude a browser that we didn't hardcode in. I don't really know the answer to that. Jan suggested to leave it this way for the time being

relay-general/src/user_agent.rs
```diff
@@ -26,27 +26,55 @@ static UA_PARSER: Lazy<UserAgentParser> = Lazy::new(|| {
...
-    None
-}
-
-/// Initializes the user agent parser.
-///
-/// This loads and compiles user agent patterns, which takes a few seconds to complete. The user
-/// agent parser initializes on-demand when using one of the parse methods. This function forces
-/// initialization at a convenient point without introducing unwanted delays.
-pub fn init_parser() {
-    Lazy::force(&UA_PARSER);
```
jjbayer: Was this function unused?

relay-general/src/user_agent.rs
```diff
@@ -26,27 +26,55 @@ static UA_PARSER: Lazy<UserAgentParser> = Lazy::new(|| {
...
-                if k.to_lowercase() == "user-agent" {
-                    return v.as_str();
+/// This has both the user agent string and the client hints, useful for the scenarios where
+/// you will use either information from client hints if it exists and if not fall back to
+/// user agent string.
+#[derive(Default, Debug)]
+pub struct RawUserAgentInfo<'a> {
+    /// The "old style" of a single UA string.
+    pub user_agent: Option<&'a str>,
+    pub client_hints: ClientHints<'a>,
```
jjbayer, line 91:
```suggestion
    /// User-Agent client hints.
    pub client_hints: ClientHints<'a>,
```

relay-general/src/store/normalize/user_agent.rs
```diff
@@ -6,45 +6,41 @@
...
+    let contexts = event.contexts.get_or_insert_with(|| Contexts::new());
+    normalize_user_agent_info_generic(contexts, &event.platform, &raw_ua_contexts);
 }
 
-pub fn normalize_user_agent_generic(
+pub fn normalize_user_agent_info_generic(
     contexts: &mut Contexts,
     platform: &Annotated<String>,
-    user_agent: &str,
+    raw_contexts: &RawUserAgentInfo,
```
jjbayer:
```suggestion
    user_agent_info: &RawUserAgentInfo,
```

relay-general/src/store/normalize/user_agent.rs
```diff
@@ -6,45 +6,41 @@
...
-    if let Some(contexts) = event.contexts.value_mut() {
-        // If a contexts object exists we modify in place.
-        normalize_user_agent_generic(contexts, &event.platform, user_agent);
-    } else {
-        // If a contexts object does not exist we create a new one and attempt to populate
-        // it.  If we didn't write any data to our new contexts instance we can throw it out
-        // and leave the existing contexts value as "None".
-        let mut contexts = Contexts::new();
-        normalize_user_agent_generic(&mut contexts, &event.platform, user_agent);
+    let raw_ua_contexts = RawUserAgentInfo::new(headers);
```
jjbayer:
```suggestion
    let user_agent_info = RawUserAgentInfo::new(headers);
```

relay-general/src/protocol/replay.rs
```diff
@@ -286,16 +286,23 @@ impl Replay {
...
+        let headers = match self
+            .request
+            .value()
+            .and_then(|request| request.headers.value())
+        {
+            Some(headers) => headers,
+            None => return,
         };
 
+        let mut raw_ua_contexts = RawUserAgentInfo::new(headers);
```
jjbayer:
```suggestion
        let mut user_agent_info = RawUserAgentInfo::new(headers);
```

relay-general/src/protocol/contexts/os.rs
```diff
@@ -47,9 +49,94 @@ impl OsContext {
...
+                Annotated::new("SEC-CH-UA-PLATFORM-VERSION".to_string().into()),
+                Annotated::new("13.1.0".to_string().into()),
+            )),
+        ];
+            PairList(headers)
+        });
+
+        let os = OsContext::from_hints_or_ua(&RawUserAgentInfo::new(&headers));
+
+        assert_eq!(os.clone().unwrap().name.as_str().unwrap(), "macOS");
```
jjbayer: `clone()` should not be necessary here.

relay-general/src/protocol/contexts/os.rs
```diff
@@ -47,9 +49,94 @@ impl OsContext {
...
+        assert_eq!(os.unwrap().version.as_str().unwrap(), "13.1.0");
```
jjbayer: You could use a single `assert_debug_snapshot!` here, instead of multiple `assert_eq`. See for example

https://github.com/getsentry/relay/blob/06c9d64bbaf9285274b690eaeed4992b97ce7213/relay-general/src/store/normalize.rs#L2270-L2286

relay-general/src/user_agent.rs
```diff
@@ -59,6 +87,20 @@ pub fn get_user_agent(request: &Annotated<Request>) -> Option<&str> {
...
+        if let Some((ref o_k, ref v)) = item.value() {
+            if let Some(k) = o_k.as_str() {
+                if k.to_lowercase() == "user-agent" {
+                    return v.as_str();
+                }
+            }
+        }
+    }
+    None
+}
```
jjbayer: Please move this function back to its original position above `get_user_agent`

1. That makes it easier to see if anything changed.
2. We declare functions in the order that they are used, and `get_user_agent_from_headers` is used in `get_user_agent`.

relay-general/src/protocol/contexts/device.rs
```diff
@@ -169,9 +172,80 @@ impl DeviceContext {
...
 }
 
-#[test]
-fn test_device_context_roundtrip() {
-    let json = r#"{
+impl FromUserAgentInfo for DeviceContext {
+    fn from_client_hints(client_hints: &ClientHints) -> Option<Self> {
+        let device = client_hints.sec_ch_ua_model?.to_owned();
+        Some(Self {
+            family: Annotated::new(device),
```
jjbayer: I know we talked about this, but I feel `sec-ch-ua-model` should go into the `model` field instead. Reading

https://github.com/getsentry/relay/blob/06c9d64bbaf9285274b690eaeed4992b97ce7213/relay-general/src/protocol/contexts/device.rs#L20-L23

and 

> A string containing the device version. For example "Pixel 3".

(from [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Model))

jjbayer: @jan-auer any opinion on this?

relay-general/src/user_agent.rs
```diff
@@ -26,27 +26,55 @@ static UA_PARSER: Lazy<UserAgentParser> = Lazy::new(|| {
...
+#[derive(Default, Debug)]
+pub struct ClientHints<'a> {
+    /// your OS, e.g. macos, android ..
+    pub sec_ch_ua_platform: Option<&'a str>,
+    /// the version number of your OS
+    pub sec_ch_ua_platform_version: Option<&'a str>,
+    /// name of web browser and its version
+    pub sec_ch_ua: Option<&'a str>,
+    /// device model, e.g. samsung galaxy 3
+    pub sec_ch_ua_model: Option<&'a str>,
```
jjbayer: Now that these fields are in a struct separate from the user-agent header, we can drop the prefix:

```suggestion
    /// The client's OS, e.g. macos, android...
    pub platform: Option<&'a str>,
    /// The version number of the client's OS.
    pub platform_version: Option<&'a str>,
    /// Name of the client's web browser and its version.
    pub ua: Option<&'a str>,
    /// Device model, e.g. samsung galaxy 3.
    pub ua_model: Option<&'a str>,
```

jjbayer (commented):
Almost there, just a couple of nits and the open question of whether to map `model -> family` or `model -> device`.

relay-general/src/user_agent.rs
TBS1996: I don't remember removing it on purpose so I added it back, but it also doesn't seem to be used. Which makes me wonder why we didn't get a lint warning on it

relay-general/src/user_agent.rs
TBS1996: im assuming in the future when working on this we would go a lot back and forth between the specs and the code, wouldn't it be easier to maintain when the names are exactly mirrored, even if it's a bit verbose?

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +27,171 @@ impl BrowserContext {
...
+            Annotated::new((
+                Annotated::new("SEC-CH-UA".to_string().into()),
+                Annotated::new(r#"Not_A Brand";v="99", "Google Chrome";v="109", "Chromium";v="109"#.to_string().into()),
+            )),
+        ];
+            PairList(headers)
+        });
+
+        let browser = BrowserContext::from_hints_or_ua(&RawUserAgentInfo::new(&headers));
+        assert_eq!(browser.clone().unwrap().version.as_str().unwrap(), "109");
```
jjbayer: `clone` should not be necessary here.

relay-general/src/user_agent.rs
```diff
@@ -26,6 +26,7 @@ static UA_PARSER: Lazy<UserAgentParser> = Lazy::new(|| {
     )
 });
 
+/// Should we merge it with get_user_agent_from_request?
```
jjbayer: Is this still an open question? If not, let's remove the comment.

jjbayer (approved):
Two more nitpicks and then ship it!

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Features**:
 
 - Add error and sample rate fields to the replay event parser. ([#1745](https://github.com/getsentry/relay/pull/1745))
+- Add support for client hints. ([#1752](https://github.com/getsentry/relay/pull/1752))
```
olksdr: This must be moved to `Unreleased` section

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +27,171 @@ impl BrowserContext {
...
+        if item.contains("Brand")
+            || item.contains("Chromium")
+            || item.contains("Gecko") // useless until firefox and safari support client hints
+            || item.contains("WebKit")
+        {
+            continue;
+        }
+
+        static UA_RE: OnceCell<Regex> = OnceCell::new();
+        let regex = UA_RE.get_or_init(|| Regex::new(r#""([^"]*)";v="([^"]*)""#).unwrap());
```
olksdr, line 82: It could be better to actually move this Regex outside of the function and make it global in this file, and add proper docs to it as well

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +27,171 @@ impl BrowserContext {
...
+                &browser.major,
+                &browser.minor,
+                &browser.patch,
+            )),
+            ..BrowserContext::default()
+        })
+    }
+}
+
+/// the sec-ch-ua field looks something like this:
```
olksdr:
```suggestion
/// The sec-ch-ua field looks something like this:
```

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -21,3 +27,171 @@ impl BrowserContext {
...
+/// "Not_A Brand";v="99", "Google Chrome";v="109", "Chromium";v="109"
+/// The order of the items are randomly shuffled
+///
+/// It tries to detect the "not a brand" item and the browser engine, if it's neither its assumed
+/// to be a browser and gets returend as such
+///
+/// returns None if no browser field detected
```
olksdr:
```suggestion
/// "Not_A Brand";v="99", "Google Chrome";v="109", "Chromium";v="109"
/// The order of the items are randomly shuffled.
///
/// It tries to detect the "not a brand" item and the browser engine, if it's neither its assumed
/// to be a browser and gets returned as such.
///
/// Returns None if no browser field detected.
```

olksdr (changes requested):

CHANGELOG.md
jjbayer: Good catch!

relay-general/src/protocol/contexts/browser.rs
TBS1996: Is that the recommended in relay? When I search for OnceCell< in the code I get mostly results from within functions such as here https://github.com/getsentry/relay/blob/master/relay-common/src/utils.rs#L70

it was also written like this when pairing with Jan

relay-general/src/user_agent.rs
```diff
@@ -79,3 +78,55 @@ pub fn parse_device(user_agent: &str) -> Device {
...
+#[derive(Default, Debug)]
+pub struct RawUserAgentInfo<'a> {
+    /// The "old style" of a single UA string.
+    pub user_agent: Option<&'a str>,
+    /// User-Agent client hints.
+    pub client_hints: ClientHints<'a>,
+}
+
+/// The client hint variable names mirror the name of the "SEC-CH" headers
+/// '<https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#user_agent_client_hints>'
```
olksdr:
```suggestion
/// The client hint variable names mirror the name of the "SEC-CH" headers, see
/// '<https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#user_agent_client_hints>'
```

relay-general/src/protocol/contexts/browser.rs
olksdr: Also this comment still stands, refactor this out of the function.

relay-general/src/user_agent.rs
```diff
@@ -79,3 +78,55 @@ pub fn parse_device(user_agent: &str) -> Device {
 pub fn parse_os(user_agent: &str) -> OS {
     UA_PARSER.parse_os(user_agent)
 }
+
+/// This has both the user agent string and the client hints, useful for the scenarios where
+/// you will use either information from client hints if it exists and if not fall back to
+/// user agent string.
```
olksdr:
```suggestion
/// The data container, which has both the user agent string and the client hints.
///
/// Useful for the scenarios where you will use either information from client hints if it exists 
/// and if not fall back to user agent string.
```

relay-general/src/protocol/contexts/browser.rs
TBS1996: Test comment.

olksdr (approved):

PR #1754, opened 2023-01-17
test(replays): Add timers to recording processing

cmanallen:
For canary use only.

closes: https://github.com/getsentry/replay-backend/issues/252

#skip-changelog


cmanallen:
@jjbayer @jan-auer Requires approval by someone on the ingest team.  Do you want to deploy this tomorrow morning and we can follow up async with results?

cmanallen:
Also I couldn't use the timers directly in the relay-replays module because of cross-dependency issues.  So I had to make a lot of stuff public and refactor the API a bit.  It should all work the same as before but the behavior is duplicated in the processor so we can time everything.

cmanallen:
Also @jan-auer and @jjbayer if you think this is ready to go feel free to merge it tomorrow and we can look over the results when I wake up.  Otherwise I'll try to be on early to handle peer review feedback.

relay-server/src/actors/processor.rs
```diff
@@ -1124,6 +1125,50 @@ impl EnvelopeProcessorService {
...
+        let mut events = metric!(timer(RelayTimers::ReplayRecordingProcessing), {
+            recording::deserialize(buffer)?
+        });
+
+        metric!(timer(RelayTimers::ReplayRecordingProcessing), {
+            recording::strip_pii(&mut events)
+                .map_err(recording::RecordingParseError::ProcessingAction)?
+        });
+
+        let bytes = metric!(timer(RelayTimers::ReplayRecordingProcessing), {
```
JoshFerge: do we have to create a unique timer for each thing we process to get the differentiated metrics? or does it use the module name or something?

cmanallen: Oversight.  Nice catch.

relay-server/src/actors/processor.rs
```diff
@@ -1124,6 +1125,50 @@ impl EnvelopeProcessorService {
...
+            recording::strip_pii(&mut events)
+                .map_err(recording::RecordingParseError::ProcessingAction)?
+        });
+
+        let bytes = metric!(timer(RelayTimers::ReplayRecordingSerialize), {
+            recording::serialize(events)?
+        });
+        let compressed_bytes = metric!(timer(RelayTimers::ReplayRecordingCompress), {
+            recording::compress(bytes)?
+        });
```
jjbayer, line 1183: Should we also add histogram metrics for the original compressed and decompressed size, + processed compressed and decompressed size?

Usage example: https://github.com/getsentry/relay/blob/1bf9c64d85a1a3d91c57d6f9fab77aed551e904d/relay-server/src/body/store_body.rs#L46

jjbayer (commented):
I would not merge this branch to master. Rather remove the condition on the organization ID and then deploy the branch directly to canary.

JoshFerge (commented):
based on our discussion yesterday, i'm not quite sure that this will help us solve the problem, which is finding the source of the increased memory usage. does ingest have thoughts on metrics we can add to help with the telemetry of this?

jjbayer:
> based on our discussion yesterday, i'm not quite sure that this will help us solve the problem, which is finding the source of the increased memory usage. does ingest have thoughts on metrics we can add to help with the telemetry of this?

@JoshFerge I would at least add some metrics for the size of the buffers involved (see comment above).

PR #1756, opened 2023-01-18
feat(replays): Log the event-id of a failed replay

cmanallen:
This will allow us to look up failed recordings and fix future events.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -1025,6 +1025,7 @@ impl EnvelopeProcessorService {
         let meta = state.envelope.meta().clone();
         let client_addr = meta.client_addr();
         let user_agent = meta.user_agent();
+        let event_id = state.envelope.event_id();
 
         state.envelope.retain_items(|item| match item.ty() {
             ItemType::ReplayEvent => {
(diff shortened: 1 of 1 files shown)


JoshFerge (approved):

olksdr (approved):

relay-server/src/actors/processor.rs
```diff
@@ -1112,7 +1113,8 @@ impl EnvelopeProcessorService {
                         }
                         Err(e) => {
                             relay_log::warn!(
-                                "replay-recording-event: failed to parse with message {}",
+                                "replay-recording-event: failed to parse {:?} with message {}",
+                                event_id,
```
jjbayer, line 1117: I think only `relay_log::error!` is reported to sentry: https://github.com/getsentry/relay/blob/229b7c3ddd22a2af8cbe9e14d63760ae881e76cd/relay-log/src/lib.rs#L38-L39.

Do we know how often this message would be emitted? If it's just the occasional error, then logging to sentry should be fine. If it's large bursts of data, we might have to sample down and only send every 100th message or so (e.g. by taking the replay id modulo 100).

JoshFerge: @jjbayer it should be in our logs though, no?

cmanallen: We'll figure it out later.

PR #1759, opened 2023-01-19
feat(glob): Allow escaping wildcard characters in globs

jjbayer:
For transaction name replacement rules, we might encounter the case where there's a literal `*` in a transaction name. If this `*` makes it into a rule, it should not be interpret as a "replace" command, so we will escape it. This PR adds support for escaping `*` in globs. 

#skip-changelog

Relates to https://github.com/getsentry/team-ingest/issues/35.


olksdr (approved):
I wouldn't call this `fix` but rather it's a new feature, which we did not support before.

Otherwise LGTM!

PR #1760, opened 2023-01-19
feat(protocol): Add `source` field to sdk interface

AbhiPrasad:
https://github.com/getsentry/sentry/issues/43427

We added the `source` field to the SDK event payload under the sdk interface so we can better track how users install our sdk: https://github.com/getsentry/develop/pull/815.

This PR updates the Relay schema to reflect the changed develop docs.

### DIFF
--- a/relay-general/src/protocol/clientsdk.rs
+++ b/relay-general/src/protocol/clientsdk.rs
@@ -54,6 +54,13 @@ pub struct ClientSdkInfo {
     #[metastructure(skip_serialization = "empty_deep")]
     pub packages: Annotated<Array<ClientSdkPackage>>,
 
+    /// The installation mechanism for the SDK. _Optional._
+    ///
+    /// This can be a package manager like `npm` or `cocoapods` or a custom installation mechanism
+    /// like `cdn`.
+    #[metastructure(skip_serialization = "empty")]
+    pub source: Annotated<String>,
+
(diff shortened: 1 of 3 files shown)


relay-general/src/protocol/clientsdk.rs
```diff
@@ -54,6 +54,13 @@ pub struct ClientSdkInfo {
     #[metastructure(skip_serialization = "empty_deep")]
     pub packages: Annotated<Array<ClientSdkPackage>>,
 
+    /// The installation mechanism for the SDK. _Optional._
+    ///
+    /// This can be a package manager like `npm` or `cocoapods` or a custom installation mechanism
+    /// like `cdn`.
+    #[metastructure(skip_serialization = "empty")]
+    pub source: Annotated<String>,
```
jjbayer, line 62: Is this field guaranteed to be free of sensitive data? That is, will it have an enum-like cardinality, or could this also contain paths or URLs with usernames, etc.

AbhiPrasad: Yes - it will have enum-like cardinality. The values will be fixed per SDK, and free to all possible PII.

I can update the docstring to include these details.

AbhiPrasad:
Converting this to draft as we re-examine how we can set `source`.

AbhiPrasad:
Closing as we've decided to use `packages` instead.

PR #1765, opened 2023-01-19
bug(replays): Fix error where some recordings can not be parsed due to type errors

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/258

Sentry event timestamps can be float or integer type.  Additionally id fields can be negative so we're using i32 now.

### DIFF
--- a/relay-replays/src/recording.rs
+++ b/relay-replays/src/recording.rs
@@ -313,7 +313,7 @@ struct MetaEvent {
 struct CustomEvent {
     #[serde(rename = "type")]
     ty: u8,
-    timestamp: f64,
+    timestamp: Value,
     data: CustomEventDataVariant,
 }
 
@@ -336,7 +336,7 @@ struct Breadcrumb {
 struct BreadcrumbPayload {
     #[serde(rename = "type")]
     ty: String,
-    timestamp: f64,
+    timestamp: Value,
     category: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     level: Option<String>,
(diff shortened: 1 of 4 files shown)


relay-replays/src/recording.rs
```diff
@@ -313,7 +313,7 @@ struct MetaEvent {
 struct CustomEvent {
     #[serde(rename = "type")]
     ty: u8,
-    timestamp: f64,
+    timestamp: Value,
```
olksdr: Would it be better to parse this into proper `Timestamp`? Like in the event protocol: https://github.com/getsentry/relay/blob/7834adb9d913b48461f7ed8f9712aa0e73f6f415/relay-general/src/protocol/event.rs#L331

olksdr (commented):
Overall LGTM.
Left also a small question.

relay-replays/src/recording.rs
olksdr: Or it can be just `DateTime` that `serde` supports

relay-replays/tests/fixtures/rrweb-event-5.json
```diff
@@ -1,20 +1,108 @@
...
+                        "totalJSHeapSize": 10204109,
+                        "usedJSHeapSize": 9131621
+                    }
+                }
+            }
+        }
+    },
+    {
+        "type": 5,
+        "timestamp": 1658770772,
```
jjbayer: Did this actually fail to parse as `f64`?

cmanallen: @jjbayer You're right.  I had test coverage that was failing but `1658770772.0` (instead of `1658770772`) should be reasonably parse-able by the UI.  I think its okay to leave this as f64.

jjbayer: `assert-json-diff` seems to have a more lenient compare mode, with which you might be able to keep the test cases:

https://docs.rs/assert-json-diff/latest/assert_json_diff/enum.NumericMode.html#variant.AssumeFloat

jjbayer (approved):

PR #1766, opened 2023-01-20
test(replays): try replay recording processing without PII scrubbing

JoshFerge:
a version of replay recording processing without PII scrubbing to see if memory usage still increases. meant for canary deploy only.

#skip-changelog


cmanallen (approved):

cmanallen:
@JoshFerge You'll have to remove all the dead code paths for this to deploy but this is totally functional as is.

jjbayer (approved):

Issue #1758, opened 2023-01-19
Add more tests cases for ProjectUpstream


olksdr:
We want to add more tests cases for `ProjectUpstream` behaviour e.g.:
* maybe a network outage
* project config is not parsable
* and others which can be tested for

Test with current `master` and without changes from #1727 

Follow up for: #1727 

--- 

- [ ] https://github.com/getsentry/relay/pull/1767 - added few tests to check the behaviour of relay project cache and handling the unparsable project states
- [ ] https://github.com/getsentry/relay/pull/1770 - unit tests for project upstream with mocked upstream service

PR #1767, opened 2023-01-20
test(project): Check the relay behavior for invalid states

olksdr:
Add integration test for project config:
* to check the behavior of the relay when we get the invalid project from upstream
  - the invalid state is saved in the ProjectCache and will be using till it expires
  - the incoming events are dropped because of the invalid state
* to make sure we recover when the project expires and the new one we get is valid one
  - once the old state expired the new one is saved in the ProjectCache
  - the events can be sent again and they are properly processed

And the second test makes sure to check that relay uses internal cache and re-uses stale state if the new one is invalid. 

ref: #1758 

#skip-changelog

### DIFF
--- a/tests/integration/test_projectconfigs.py
+++ b/tests/integration/test_projectconfigs.py
@@ -352,3 +352,70 @@ def test_unparsable_project_config(mini_sentry, relay):
     relay.send_event(42)
     event = mini_sentry.captured_events.get(timeout=1).get_event()
     assert event["logentry"] == {"formatted": "Hello, World!"}
+
+
+def test_cached_project_config(mini_sentry, relay):
+    project_key = 42
+    relay_config = {
+        "cache": {"project_expiry": 2, "project_grace_period": 60, "miss_expiry": 2}
+    }
+    relay = relay(mini_sentry, relay_config, wait_health_check=True)
+    mini_sentry.add_full_project_config(project_key)
+    public_key = mini_sentry.get_dsn_public_key(project_key)
+
+    # Once the event is sent the project state is requested and cached.
+    relay.send_event(42)
+    event = mini_sentry.captured_events.get(timeout=1).get_event()
+    assert event["logentry"] == {"formatted": "Hello, World!"}
+    # send a second event
+    relay.send_event(42)
+    event = mini_sentry.captured_events.get(timeout=1).get_event()
+    assert event["logentry"] == {"formatted": "Hello, World!"}
+
+    body = {"publicKeys": [public_key]}
+    packed, signature = SecretKey.parse(relay.secret_key).pack(body)
+    data = get_response(relay, packed, signature)
+    assert data["configs"][public_key]["projectId"] == project_key
+    assert not data["configs"][public_key]["disabled"]
+
+    # Introduce unparsable config.
+    config = mini_sentry.project_configs[project_key]["config"]
+    config.setdefault("dynamicSampling", {}).setdefault("rules", []).append(
+        {
+            "condition": {
+                "op": "and",
+                "inner": [
+                    {"op": "glob", "name": "releases", "value": ["1.1.1", "1.1.2"]}
+                ],
+            },
+            "sampleRate": 0.7,
+            "type": "trace",
+            "id": 1,
+            "timeRange": {
+                "start": "2022-10-10T00:00:00.000000Z",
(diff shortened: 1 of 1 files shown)


tests/integration/test_projectconfigs.py
```diff
@@ -208,3 +209,213 @@ def request_config():
...
+        }
+    } == data
+
+    try:
+        # This event will be dropped since the project state is invalid.
+        pytest.raises(HTTPError, lambda: relay.send_event(project_key))
+        time.sleep(0.5)
+        assert {str(e) for _, e in mini_sentry.test_failures} == {
+            f"Relay sent us event: error fetching project state {public_key}: missing field `type`",
+            "Relay sent us event: dropped envelope: invalid data (project_state)",
```
jjbayer, line 300: Probably out of scope for this PR, but why do we drop events here? Shouldn't relay keep events queued for a while, until it can either get a valid project config or some timeout expires?

tests/integration/test_projectconfigs.py
```diff
@@ -208,3 +209,213 @@ def request_config():
...
+    assert not data["configs"][public_key]["disabled"]
+
+    # This is still a grace period, and the state for us must be still valid, even though we get the error parsing the new state.
+    try:
+        # Give it a bit time for update to go through.
+        time.sleep(1)
+        data = get_response(relay, packed, signature)
+        assert {str(e) for _, e in mini_sentry.test_failures} == {
+            f"Relay sent us event: error fetching project state {public_key}: missing field `type`",
+        }
```
jjbayer, line 416: Should we add another sleep here that exceeds the grace period, and then assert that after that period, the event is dropped?

tests/integration/test_projectconfigs.py
olksdr: It's because in https://github.com/getsentry/relay/blob/9808ea319b5bf60de2eb9cdd399db9d54a62d989/relay-server/src/actors/processor.rs#L2142 when we call `prepare_state` function we are unable to find the project id and immediately reject the envelope https://github.com/getsentry/relay/blob/9808ea319b5bf60de2eb9cdd399db9d54a62d989/relay-server/src/actors/processor.rs#L1206

The project id wasn't set, because we had an empty cache and then we got broken project state, where we could not parse the config, so we set the state to `errored`, with basically all fields in the struct empty by default.

tests/integration/test_projectconfigs.py
olksdr: it's possible yeah, it's will just increase the wait time in integration tests. 
I will lower the grace period then and will try to test grace period for expiration as well.

tests/integration/test_projectconfigs.py
jjbayer: OK, I guess that makes sense. @jan-auer is this behavior we would like to change in the future? That is, if a project does not have a previous entry and receives a broken one, leave its envelopes in the queue until the project config is repaired or we reach a timeout?

tests/integration/test_projectconfigs.py
```diff
@@ -419,3 +419,16 @@ def test_cached_project_config(mini_sentry, relay):
 
     assert data["configs"][public_key]["projectId"] == project_key
     assert not data["configs"][public_key]["disabled"]
+
+    # Wait till grace period expires as well and we should be dropping events now.
+    time.sleep(5)
```
jjbayer, line 424: Could we shorten this `sleep` by settings `project_expiry` to 1 and `project_grace_period` to 2? Or would that make the test flaky?

jjbayer (approved):
Minor note about shaving off a few seconds, but this looks good to me!

tests/integration/test_projectconfigs.py
olksdr: Yeah, that's why I gave it longer time here, since it depends on how fast you computer is, it can make this test flaky, which I would like to avoid. 
Once we port those integration tests to Rust we could make this work and look much more elegant.

tests/integration/test_projectconfigs.py
jan-auer: Filed https://github.com/getsentry/relay/issues/1787 to keep track of this. Note that it's not the EnvelopeProcessor, but the project cache where the envelopes get dropped.

PR #1770, opened 2023-01-21
test(project_upstream): Test correct message handling

olksdr:
This change adds conditional compilation for `test` and not `test` targets, introducing the mocked `Upstream` service.

The tests check that the messages have proper flow, and project upstream manages inner queue the way it's supposed to:
* keeps internal queue
* merges same project key requests into one and broadcast the result

ref: #1758 

#skip-changelog


olksdr:
Closing this for now. We will come back to this as part of the #1818

PR #1772, opened 2023-01-23
chore(redis): enable "tls" feature for redis crate

### LINKED ISSUE #995: Allow usage of TLS for redis connection
We are trying to make setup of relay secure, as a part of sentry kubernetes deployment.
And apart from Kafka ssl problems (seen in https://github.com/getsentry/relay/issues/971), 
we are facing problems with connecting via tls to redis. 
It should be simple change - https://github.com/getsentry/relay/blob/master/relay-redis/Cargo.toml#L15 should also include TLS (exactly  `features = ["tls"]`)
Do You confirm that this is possible or will it brake another part of development as kafka did?

olksdr:
Added one more feature flag to `relay-redis` crate to be able to finetune what features are used and compiled into `relay`.
It is possible now to compile `relay-quotas` and `relay-server` with ssl support.  

This also bumps the crate version to `0.22.2` with some bug fixes in. 

fix: https://github.com/getsentry/relay/issues/995

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Features**:
 
 - Add error and sample rate fields to the replay event parser. ([#1745](https://github.com/getsentry/relay/pull/1745))
+- Add SSL support to `relay-redis` crate. ([#1772](https://github.com/getsentry/relay/pull/1772))
 
 **Internal**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 **Features**:
 
 - Add error and sample rate fields to the replay event parser. ([#1745](https://github.com/getsentry/relay/pull/1745))
-- Add SSL support to `relay-redis` crate. ([#1772](https://github.com/getsentry/relay/pull/1772))
+- Add SSL support to `relay-redis` crate. It is possible to use `rediss` scheme to connnect to Redis cluster using TLS. ([#1772](https://github.com/getsentry/relay/pull/1772))
 
 **Internal**:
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Features**:
 
 - Add error and sample rate fields to the replay event parser. ([#1745](https://github.com/getsentry/relay/pull/1745))
+- Add SSL support to `relay-redis` crate. ([#1772](https://github.com/getsentry/relay/pull/1772))
```
jjbayer: Does this mean connecting to TLS now works out of the box, by using the [`rediss://`](https://github.com/redis-rs/redis-rs#tls-support) scheme?

olksdr: Yep. If the relay compiled with `ssl`  feature it should be possible to use `rediss` scheme with TLS support.

olksdr:
@jjbayer  updated Changelog in [9c1f44a](https://github.com/getsentry/relay/pull/1772/commits/9c1f44a321bfedcef8da83a9800206f253150d0d), PTAL

jjbayer (approved):
LGTM

CHANGELOG.md
SmoshySmosh: @olksdr It doesn't look like Relay is compiled with the `ssl` feature on public docker images. Is it possible to add that capability?

olksdr: @SmoshySmosh  it looks like we do build the images with `ssl` enabled. Could you, please, open the issue in this repo with the error you get and the config you have? 
Thanks!

PR #1773, opened 2023-01-23
fix(txnames): Improve regex to properly parse the UUIDs

jjbayer:
Basically regex contained whitespaces and new line in the beginning of the regex, which make the UUID fail to match any of the UUID-like strings.

#skip-changelog


olksdr (approved):
:shipit:

PR #1775, opened 2023-01-24
chore(txnames): Replace identifiers in sanitized sources

olksdr:
This change adds `Sanitized` transaction source to transaction names normalization. 

This way, in addition to renaming rules, we can catch more high cardinality URL parts. 

#skip-changelog

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -319,8 +319,13 @@ impl Processor for TransactionsProcessor<'_> {
             event.transaction_info.value_mut(),
         )?;
 
-        // Normalize transaction names for URLs transaction sources only.
-        if event.get_transaction_source() == &TransactionSource::Url && self.normalize_names {
+        // Normalize transaction names for URLs and Sanitized transaction sources.
+        // This in addition to renaming rules can catch some high cardinality parts.
+        if matches!(
+            event.get_transaction_source(),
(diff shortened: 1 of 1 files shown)


relay-general/src/store/transactions/processor.rs
```diff
@@ -319,8 +319,13 @@ impl Processor for TransactionsProcessor<'_> {
...
 
-        // Normalize transaction names for URLs transaction sources only.
-        if event.get_transaction_source() == &TransactionSource::Url && self.normalize_names {
+        // Normalize transaction names for URLs and Sanitized transaction sources.
+        // This in addition to renaming rules can catch some high cardinality parts.
+        if matches!(
+            event.get_transaction_source(),
+            &TransactionSource::Url | &TransactionSource::Sanitized
+        ) && self.normalize_names
+        {
```
jjbayer, line 328: We could also run `normalize_transaction_name` before `apply_transaction_rename_rule`, would that result in the same name? Either way, we should think about what we put into `_meta` when both functions find something. Two separate `"rem"`s, maybe?

olksdr: If we run the normalization before the renaming rules, it can happen that the rules won't match anymore. Yeah, we might have to update the meta accordingly.

jjbayer (approved):
As discussed, to me it would feel more natural to apply the simple rules before the clustering rules, but since they are roughly equivalent, I think this change is good to go.

PR #1776, opened 2023-01-24
ref: Remove actor utils and actor language from all services

jan-auer:
Removes all mentions of "actors" from services, removes now unused
utilities relating to actix, and removes actix subscribers from the
`Controller`.

This does not fully remove actix from the service layer yet. Most
notably, the `Controller` still uses actix to catch process signals.
Once that is fixed, we can also remove actix from `relay_system` and the
remaining tests, which will just leave the web framework back.

#skip-changelog


sl0thentr0py (approved):

relay-server/src/utils/actix.rs
```diff
@@ -1,23 +1,11 @@
-use actix::System;
 use tokio::runtime::Runtime;
```
olksdr: Could we, please, also rename this file, or move `create_runtime` directly to `utils`, since there is nothing from `actix` here anymore.

olksdr (commented):
Overall looks good to me, left only one comment

olksdr (approved):
🚀

olksdr (approved):
🚀

PR #1778, opened 2023-01-24
feat(replays): Scrub recording data from generic Value type

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/259

Should test on canary first.


jjbayer (changes requested):
Hey, feel free to deploy this on Canary, but in general I think we should try to make the `Deserialize` implementation of `Event` more efficient instead of walking through an untyped `Value`. The ingest team will spend some time today to fix the deserializer.

PR #1780, opened 2023-01-25
chore(kafka): Set retention on MetricKafkaMessage

### LINKED ISSUE #1779: Add `retention_days` to Metrics Kafka message
Storage is asking us to add the data retention for metric buckets to the Kafka message Relay produces.

Relay sets data retention with a field called `retention_days` for sessions and replays already. That information is obtained from the Envelope's `retention` method, which is set in the processor here:

https://github.com/getsentry/relay/blob/bcb18a289dee2e677e24feaa1f0f93f9e4b69d62/relay-server/src/actors/processor.rs#L1187-L1191

For now, we can use the same value for metric buckets.

olksdr:
Use the event's retention and propagate it to metric kafka message. 

#skip-changelog

fix: #1779

### DIFF
--- a/relay-server/src/actors/store.rs
+++ b/relay-server/src/actors/store.rs
@@ -171,9 +171,12 @@ impl StoreService {
                         item,
                     )?;
                 }
-                ItemType::MetricBuckets => {
-                    self.produce_metrics(scoping.organization_id, scoping.project_id, item)?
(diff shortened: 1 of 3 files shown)


jan-auer (approved):
LGTM. @onewland to provide final approval

PR #1783, opened 2023-01-26
fix(filter): Ignore <anonymous> frames

### LINKED ISSUE #1750: browser extensions filter not working for chrome
We see the following form of stack trace both in our own projects as in customer projects (look for it by searching for `stack.abs_path:chrome*` in Discover):

```json
"frames": [
    {
        "filename": "chrome-extension://asdf/some.js",
        "abs_path": "chrome-extension://asdf/some.js",
        "lineno": 44,
        "colno": 65812,
        "in_app": true
    },
    {
        "function": "someFn",
        "filename": "chrome-extension://asdf/some.js",
        "abs_path": "chrome-extension://asdf/some.js",
        "lineno": 44,
        "colno": 65406,
        "in_app": true
    },
    {
        "function": "Function.defineProperty",
        "filename": "<anonymous>",
        "abs_path": "<anonymous>",
        "in_app": true
    }
]
```

Because `get_exception_source` only looks at the last stack frame, this is not caught by the browser extensions filter: https://github.com/getsentry/relay/blob/2a0976d7d9ecfd4daa6cb2390c224d09f9a79bbd/relay-filter/src/browser_extensions.rs#L107-L112.

Unsure if this is the case for all chrome extensions or just for some.

Possible solution: Look at the last frame in the stack trace that has a line number.

olksdr:
Skip all the anonymous frames and pick the first from the end of the stack which is defined and try to use it in the filter.

fix: #1750 

#skip-changelog


relay-filter/src/browser_extensions.rs
```diff
@@ -107,8 +107,15 @@ fn get_exception_value(event: &Event) -> Option<&str> {
 fn get_exception_source(event: &Event) -> Option<&str> {
     let exception = get_first_exception(event)?;
     let frames = exception.stacktrace.value()?.frames.value()?;
-    let last_frame = frames.last()?.value()?;
-    Some(last_frame.abs_path.value()?.as_str())
+    // Iterate from the tail and get the first frame which is not <anonymous>.
+    for f in frames.iter().rev() {
+        let abs_path = f.value()?.abs_path.value()?;
+        let path = abs_path.as_str();
+        if path != "<anonymous>" {
```
jan-auer: _nit_: You could introduce a constant for this and in the comment on the constant add a sentence or two on why we're skipping anonymous.

jan-auer (approved):

jan-auer (approved):

PR #1789, opened 2023-01-28
fix(pii): Scrub the objects

### LINKED ISSUE #42317: I don't understand why server side data scrubbing with default scrubbers does NOT apply for stack trace variables.
I don't understand why server side data scrubbing with default scrubbers does NOT apply for stack trace variables. I'm a long time Sentry user and this is (somewhat surprisingly) the first time I saw a clear text password in Sentry.

image

Go under `Project → Settings → Security & Privacy → Data Scrubber + Use Default Scrubbers`

> Apply default scrubbers to prevent things like passwords and credit cards from being stored

_Originally posted by @fredericgboutin-yapla in https://github.com/getsentry/sentry-php/issues/965#issuecomment-1349771830_

olksdr:
If any PII fields are found and those fields contain the objects, let's scrub the entire object by default.

This should prevent the accidental data leakage.

fix: https://github.com/getsentry/sentry/issues/42317


relay-general/src/pii/convert.rs
```diff
@@ -1410,4 +1411,32 @@ THd+9FBxiHLGXNKhG/FRSyREXEt+NyYIf/0cyByc9tNksat794ddUqnLOg0vwSkv
         process_value(&mut data, &mut pii_processor, ProcessingState::root()).unwrap();
         assert_annotated_snapshot!(data);
     }
+
+    #[test]
+    fn test_scrub_object() {
+        let mut data = Event::from_value(
+            serde_json::json!({
+                "request": {
+                    "url": "https://www.bitumequebec.ca/fr/adhesions/authentication",
```
jjbayer: Where does this URL come from? Should we replace it with an `example.com` url?

relay-general/src/pii/convert.rs
```diff
@@ -1410,4 +1411,32 @@ THd+9FBxiHLGXNKhG/FRSyREXEt+NyYIf/0cyByc9tNksat794ddUqnLOg0vwSkv
...
+                    "method": "POST",
+                    "data": {
+                      "password": "test",
+                      "profile": {
+                        "email": "testing@example.com",
+                        "password": {
+                          "$eq": "u{}{}{}]H[[[]]ww6KrA9F.x-F%%"
```
jjbayer, line 1427: Just to double check, was this copied from a real event?

relay-general/src/pii/snapshots/relay_general__pii__convert__tests__contexts.snap
```diff
@@ -76,15 +76,7 @@ expression: data
...
     },
-    "secret": {
-      "a_password_here": "[Filtered]",
-      "apiKey": "[Filtered]",
-      "api_key": "[Filtered]",
-      "foo": "bar",
-      "password": "[Filtered]",
-      "the_secret": "[Filtered]",
-      "type": "[Filtered]"
-    },
```
jjbayer, line 87: I have no personal objections against this change, but there seems to have been a product decision at some point in the past to not scrub entire objects in general. There is an entire section in our [docs](https://docs.sentry.io/product/data-management-settings/scrubbing/server-side-scrubbing/) about it. I guess we can do two things:

1. Just go ahead with this change and update the docs.
2. Only add `$object` to specific fields, e.g. `password` and `secret`, but not others like `credentials`, `auth`.

iker-barriocanal: If we add `$object` to `password`, why not do the same with `credentials`?

iker-barriocanal (commented):
Same questions as what Joris posted. Other than these, the PR looks good.

relay-general/src/pii/convert.rs
olksdr: yes, see the linked issue. The password is just a test random string here.

relay-general/src/pii/snapshots/relay_general__pii__convert__tests__contexts.snap
olksdr: The test was quite funny in first place, why the `type` field was filtered here, any ideas?

jjbayer: > If we add $object to password, why not do the same with credentials?

There's an example in the docs that lists a `credentials` object which also contains `lastLogin`, a field that does not need to be scrubbed. But I agree that there is no obvious line we can draw between "this should be scrubbed completely" and "this should be scrubbed partially".

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Internal**
+
+- Make sure to scrub all the fields with PII. If the fields contain an object, the entire object will be removed. ([#1789](https://github.com/getsentry/relay/pull/1789))
```
jjbayer, line 16: Please also add this line to `py/CHANGELOG.md`, as per https://github.com/getsentry/relay#instructions-for-changelogs.

relay-general/src/pii/snapshots/relay_general__pii__convert__tests__contexts.snap
jjbayer: > why the type field was filtered here, any ideas?

`type` was filtered because it's value was `secret` (compare to other contexts where `type` also corresponds to the top level key).

jjbayer: I only now realize that with this change, we will drop entire `Context`s. This goes against the purpose of adding contexts to events (see [docs](https://docs.sentry.io/platforms/javascript/enriching-events/context/)). But I guess users can always rename them if they do not want them scrubbed.

jjbayer (approved):
As discussed in Daily, let's go ahead with this change. @olksdr just make sure to update the docs accordingly, or create a follow-up item to do so.

CHANGELOG.md
olksdr: done in [3811a6e](https://github.com/getsentry/relay/pull/1789/commits/3811a6e55ca078b360616470883ccf3f792771d2)

iker-barriocanal (approved):

PR #1790, opened 2023-01-30
feat(relay): Implement factor based multi-matching

iambriccardo:
This PR implements a new sampling algorithm that has two new major changes, support for factors and multi-matching.

The rationale behind this new algorithm is to enable the interpolation of different matching rules due to the increasing number of biases created on the sentry side (e.g., we want to mix the latest release and team key transaction biases together). With the current system, this was not possible, because matching was performed in a FIFO fashion.

**This algorithm can result in inconsistent trace sampling, which is a drawback we decided to accept. _The algorithm does perform inconsistent sampling because it will take into consideration both transaction and trace rules while computing the final sample rate._**

The algorithm revolves around two configurations:
* **root config** -> contains the rules of the head of the trace to which the incoming event belongs.
* **non-root config** -> contains the rules of the project to which the incoming event belongs.

It also introduces the concept of `samplingValue`. A `samplingValue` is nothing more than a type of heuristic we apply to sampling, which has a value. The two types we currently have implemented are:
* `factor` -> the factor which the algorithm will apply on a matched sample rate.
* `sampleRate` -> the classic sample rate.

The new algorithm works in the following way:
1. It merges the sampling rules of the root and non-root projects, concatenating first the **transaction/error** rules of the non-root project and then the **trace** rules of the root project.
2. It performs matching from **top to bottom** with the following approach:
3. If a `factor` rule is matched, it will take the factor and multiply it with past accumulated factors (starting with default 1).
4. If a `sampleRate` rule is matched, it will take the accumulated factors and multiply that by the `sampleRate` of the matched rule.

The matching process will return the final `sampleRate` that will be used to decide whether to keep or not keep the event. _In case no matches or unsupported rules are found, the algorithm will return a no match, which will mean that we will keep the event (we prefer to oversample than drop events)._


relay-sampling/src/lib.rs
```diff
@@ -357,6 +358,25 @@ impl TimeRange {
     }
 }
 
+/// A sampling strategy definition.
+///
+/// A sampling strategy refers to the strategy that we want to use for sampling a specific rule.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "type")]
+pub enum SamplingStrategy {
```
jjbayer: This might be subjective but I would rather call this `SamplingValue` or even `SampleRate`, rather than `Strategy`.

relay-sampling/src/lib.rs
```diff
@@ -366,46 +386,51 @@ impl TimeRange {
...
-    Linear { decayed_sample_rate: f64 },
+    Linear { decayed_value: f64 },
     #[default]
     Constant,
 }
 
 /// A struct representing the evaluation context of a sample rate.
 #[derive(Debug, Clone, Copy)]
-enum SampleRateEvaluator {
+enum SamplingStrategyValueEvaluator {
```
jjbayer: See comment about naming above.

py/tests/test_processing.py
```diff
@@ -242,7 +242,10 @@ def test_validate_sampling_configuration():
         "rules": [
             {
                 "type": "trace",
-                "sampleRate": 0.7,
+                "samplingStrategy": {
+                    "type": "sampleRate",
+                    "value": 0.7
+                },
```
jjbayer: Outdated customer Relays will ignore this new field, and fail to parse this config because `sampleRate` is missing. I'm afraid the existing `unsupported` logic does not catch this.

relay-sampling/src/lib.rs
```diff
@@ -366,46 +386,51 @@ impl TimeRange {
...
                 let start_timestamp = start.timestamp() as f64;
                 let end_timestamp = end.timestamp() as f64;
-                let progress_ratio = ((now_timestamp - start_timestamp)
-                    / (end_timestamp - start_timestamp))
-                    .clamp(0.0, 1.0);
+                let mut progress_ratio =
+                    (now_timestamp - start_timestamp) / (end_timestamp - start_timestamp);
+                if !use_factor {
+                    // In case we don't use factors, we want to clamp the value between 0 and 1.
+                    progress_ratio = progress_ratio.clamp(0.0, 1.0)
```
jjbayer: The `progress_ratio` is computed entirely from timestamps, so why does it matter if the value is a factor or a sample rate < 1?

relay-sampling/src/lib.rs
```diff
@@ -357,6 +358,25 @@ impl TimeRange {
...
 
+/// A sampling strategy definition.
+///
+/// A sampling strategy refers to the strategy that we want to use for sampling a specific rule.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "type")]
+pub enum SamplingStrategy {
+    SampleRate { value: f64 },
+    Factor { value: f64 },
```
jjbayer, line 374: Please also add a doc comment above these variants explaining how they change sampling behavior.

jjbayer (changes requested):
Just a partial review for now, I think we need to think about compatibility some more.

relay-sampling/src/lib.rs
iambriccardo: I would prefer `SamplingValue` thanks for the suggestion!

iambriccardo:
Regarding the backward and forward compatibility here are my thoughts:
1. Backward compatibility (aka new Relay receives old config) -> we default to sampling at `1.0` by using the `default` value for `SamplingValue`. With this in case we receive an old config and don't find the `SamplingValue` we will just keep the event in case it matches.
2. Forward compatibility (aka old Relay receives new config) -> we first need to release a temporary fix in which we put the default `sample_rate` field to defaulting to `1.0` in case it is not found. With this in case we receive a new config and don't find `sample_rate` we will just keep the event in case it matches.

Both situations should auto-resolve once convergence is reached. The problem is that I don't know how customer relays play into this. Would need more input @jjbayer.

jjbayer:
> 2. Forward compatibility (aka old Relay receives new config) -> we first need to release a temporary fix in which we put the default `sample_rate` field to defaulting to `1.0` in case it is not found. With this in case we receive a new config and don't find `sample_rate` we will just keep the event in case it matches.

Sentry sending both the new `samplingStrategy` _and_ `sampleRate: 1.0` should actually work, also for customer relays! For documentation purposes, I would leave the `sample_rate` in the protocol in Relay though, with a doc comment explaining why it's still there.

iambriccardo:
> > 2. Forward compatibility (aka old Relay receives new config) -> we first need to release a temporary fix in which we put the default `sample_rate` field to defaulting to `1.0` in case it is not found. With this in case we receive a new config and don't find `sample_rate` we will just keep the event in case it matches.
> 
> Sentry sending both the new `samplingStrategy` _and_ `sampleRate: 1.0` should actually work, also for customer relays! For documentation purposes, I would leave the `sample_rate` in the protocol in Relay though, with a doc comment explaining why it's still there.

I am not sure I am following. You would like to keep both fields forever?

relay-sampling/src/lib.rs
iambriccardo: I have just noticed that I have clamped the wrong field 🫣, I will need to clamp the result

iambriccardo: Will fix now, I also saw I forgot to remove the `true` for `use_factor`.

py/tests/test_processing.py
```diff
@@ -242,7 +242,10 @@ def test_validate_sampling_configuration():
         "rules": [
```
jjbayer: Shouldn't this be `rulesV2` now?

relay-sampling/src/lib.rs
```diff
@@ -491,6 +507,20 @@ impl SamplingRule {
 
         None
     }
+
+    fn is_sample_rate_rule(&self) -> bool {
+        match self.sampling_value {
+            SamplingValue::SampleRate { value: _ } => true,
+            SamplingValue::Factor { value: _ } => false,
+        }
```
```suggestion
        matches!(self.sampling_value, SamplingValue::SampleRate { _ })
```

relay-sampling/src/lib.rs
```diff
@@ -491,6 +507,20 @@ impl SamplingRule {
...
+    }
+
+    fn get_sampling_base_value(&self) -> f64 {
+        match self.sampling_value {
+            SamplingValue::SampleRate { value: sample_rate } => sample_rate,
+            SamplingValue::Factor { value: factor } => factor,
+        }
+    }
```
jjbayer: nit: Can we put this helper in `impl SamplingValue`?

relay-sampling/src/lib.rs
```diff
@@ -809,79 +839,179 @@ impl Default for SamplingMode {
...
+
 /// Represents the dynamic sampling configuration available to a project.
 ///
 /// Note: This comes from the organization data
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SamplingConfig {
-    /// The ordered sampling rules for the project from highest to lowest priority.
+    /// The ordered sampling rules for the project.
+    #[serde(default, skip_deserializing)]
```
```suggestion
    /// Legacy sampling rules.
    ///
    /// We need to serialize an empty list of rules such that outdated downstream Relays
    /// accept the config, but at the same time, we do not want to actually use these rules
    /// in the current Relay.
    #[serde(default, skip_deserializing)]
```

relay-sampling/src/lib.rs
```diff
@@ -809,79 +839,184 @@ impl Default for SamplingMode {
...
+    /// The ordered sampling rules for the project.
+    ///
+    /// This field will remain here to serve only for old customer Relays to which we will
+    /// forward the sampling config. The idea is that those Relays will get the old rules as
+    /// empty array, which will result in them not sampling and forwarding sampling decisions to
+    /// upstream Relays.
+    #[serde(default, skip_deserializing)]
     pub rules: Vec<SamplingRule>,
+    /// The ordered sampling rules v2 for the project.
+    pub rules_v2: Vec<SamplingRule>,
```
jjbayer: This might make the protocol a bit confusing, but I think it makes the rest of the code / diff easier to read. Feel free to disagree.

```suggestion
    #[serde(default, skip_deserializing, rename="rules")]
    pub legacy_rules: Vec<SamplingRule>,
    /// The ordered sampling rules v2 for the project.
    #[serde(rename = "rulesV2")]
    pub rules: Vec<SamplingRule>,
```

relay-sampling/src/lib.rs
```diff
@@ -809,79 +839,184 @@ impl Default for SamplingMode {
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct MatchedRuleIds(pub Vec<RuleId>);
+
+impl MatchedRuleIds {
+    /// Creates a MatchedRuleIds struct from a string formatted with the following format:
+    /// rule_id_1;rule_id_2;...
+    pub fn from_string(value: &str) -> Result<MatchedRuleIds, ParseIntError> {
+        let mut rule_ids = vec![];
+
+        for rule_id in value.split(';') {
```
jjbayer: nit: Why a `;` and not a `,`? The outcome reason `Sampled:1,2,3` might be easier to read than `Sampled:1;2;3`

relay-sampling/src/lib.rs
```diff
@@ -809,79 +839,184 @@ impl Default for SamplingMode {
...
+                .map(|rule_id| format!("{rule_id}"))
+                .collect::<Vec<String>>()
+                .join(";")
+        )
+    }
+}
+
+/// Represents the specification for sampling an incoming event.
+#[derive(Clone, Debug, PartialEq)]
+pub struct SamplingConfigMatchResult {
```
jjbayer: nit: I would rename this to `SamplingMatch`.

relay-sampling/src/lib.rs
```diff
@@ -809,79 +839,184 @@ impl Default for SamplingMode {
...
+    pub sample_rate: f64,
+    pub seed: Uuid,
+    pub matched_rule_ids: MatchedRuleIds,
```
jjbayer, line 909: Please add doc comments to every field here.

relay-sampling/src/lib.rs
```diff
@@ -809,79 +839,184 @@ impl Default for SamplingMode {
...
+                return Some(dsc.trace_id);
             }
+        }
 
-            if !rule.condition.matches(sampling_context, ip_addr) {
-                return None;
-            }
+        if let Some(event_id) = event_id {
+            return Some(event_id.0);
+        }
```
```suggestion
        event_id.map(|id| id.0)
```

relay-sampling/src/lib.rs
```diff
@@ -809,79 +839,184 @@ impl Default for SamplingMode {
...
 
-            rule.is_active(now)
-        })
+        None
+    }
+
+    pub fn has_unsupported_rules(&self) -> bool {
```
jjbayer: Can we move this function back to the top? Just to make the diff easier to read.

relay-sampling/src/lib.rs
```diff
@@ -809,79 +839,184 @@ impl Default for SamplingMode {
...
         now: DateTime<Utc>,
-    ) -> Option<ActiveRule> {
-        let ty = if let Some(EventType::Transaction) = &event.ty.0 {
-            RuleType::Transaction
-        } else {
-            RuleType::Error
-        };
+    ) -> Option<SamplingConfigMatchResult> {
+        let mut matched_rule_ids = vec![];
+        let mut has_matched_trace_rule = false;
```
jjbayer: Instead of storing a boolean and using `get_seed`, I would just

```rust
let mut seed = event.id.value();
```

and then overwrite it with the DSC seed further down if necessary.

jjbayer (changes requested):
This looks good, mostly nitpicks at this stage. I did not look at the tests yet.

py/tests/test_processing.py
iambriccardo: Yeah, you are absolutely right, I am confused of why it didn't fail.

iambriccardo: Yep, i didn't catch it as CI was blocked and integration tests weren't run again. Will fix!

relay-sampling/src/lib.rs
iambriccardo: Lovely!

relay-sampling/src/lib.rs
iambriccardo: I would prefer to keep the `rules_v2` schema, to be more explicit about the newer version, like it is typically done in APIs. The reasoning for avoiding `legacy` is that if we upgrade again the schema, we will have to define `legacy_legacy`? I envision a simpler naming scheme with `rules_v*`. Ofc the best solution would be to implement proper versioning.

relay-sampling/src/lib.rs
iambriccardo: I made my implementation under the assumption of having trace and transaction rules mixed, as shown in the tests. For this reason I use a boolean to keep track of any seen trace.

relay-sampling/src/lib.rs
```diff
@@ -374,6 +374,15 @@ pub enum SamplingValue {
     Factor { value: f64 },
 }
 
+impl SamplingValue {
+    fn get_sampling_base_value(&self) -> f64 {
```
jjbayer: nit: Our getters usually do not have a `get_` prefix (see also [rust docs](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter)). Should we just call this `value()`?

relay-sampling/src/lib.rs
jjbayer: What I mean is that on line 1000, instead of flipping a boolean, you could overwrite the seed directly.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -1,170 +1,219 @@
...
+) -> SamplingMatchResult {
+    // We get all the required data for transaction-based dynamic sampling.
+    let event = no_match_if_none!(event);
+    let sampling_config = no_match_if_none!(&project_state.config.dynamic_sampling);
+
+    // We obtain the merged sampling configuration with a concatenation of transaction rules
+    // of the current project and trace rules of the root project.
+    let merged_config = SamplingConfigs::new(sampling_config)
+        .add_root_config(root_project_state)
+        .get_merged_config();
```
jjbayer: We should not clone configs / create new vectors for every event here. Ideally, we operate on a `get_merged_rules()` function that returns an iterator (not a vector) of rules.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -1,170 +1,219 @@
...
+
+/// The result of the dynamic sampling matching executed on the sampling config.
+#[derive(Clone, Debug, PartialEq)]
+enum SamplingMatchResult {
+    /// The incoming event/dynamic sampling context matched the sampling configuration.
+    Match {
+        sample_rate: f64,
+        matched_rule_ids: MatchedRuleIds,
+        seed: Uuid,
+    },
```
jjbayer: This looks very similar to the `SamplingMatch` defined in `relay-sampling`. Can we just use an `Option<SamplingMatch>` instead of this enum?

jjbayer (changes requested):

relay-sampling/src/lib.rs
iambriccardo: Oh yeah sorry, I get now what you mean. Will do that, sorry for the confusion.

relay-server/src/utils/dynamic_sampling.rs
iambriccardo: I wanted to discuss this indeed, I can try to implement the approach with iterators.

relay-server/src/utils/dynamic_sampling.rs
iambriccardo: They look similar but we actually change the sample rate, I would prefer to keep this `Match` and `NoMatch` for expressiveness and I can use the `SamplingMatch` inside the `Match` variant of the enum.

relay-sampling/src/lib.rs
```diff
@@ -997,14 +980,17 @@ impl SamplingConfig {
                     matched_rule_ids.push(rule.id);
 
                     if rule.ty == RuleType::Trace {
-                        has_matched_trace_rule = true
+                        seed = match dsc {
+                            Some(dsc) => Some(dsc.trace_id),
+                            None => seed,
+                        }
```
jjbayer: nit

```suggestion
                        if let Some(dsc) = dsc {
                            seed = Some(dsc.trace_id);
                        }
```

relay-sampling/src/lib.rs
```diff
@@ -888,65 +888,31 @@ pub struct SamplingMatch {
...
-                return Some(dsc.trace_id);
-            }
-        }
-
-        event_id.map(|id| id.0)
+    /// Returns an iterator of references that chains together and merges rules.
+    ///
+    /// The chaining logic will take all the non-trace rules from the project and all the trace/unsupported
+    /// rules from the root project and concatenate them.
+    pub fn merge_rules_from_configs<'a>(
```
jjbayer: nit: This function is neither a method of `SamplingMatch`, nor does it create a `SamplingMatch`, so I would make it a standalone function.

relay-sampling/src/lib.rs
```diff
@@ -888,65 +888,31 @@ pub struct SamplingMatch {
...
+    ) -> impl Iterator<Item = &'a SamplingRule> {
+        let event_rules = sampling_config
+            .rules_v2
+            .iter()
+            .filter(|&rule| rule.ty == RuleType::Transaction || rule.ty == RuleType::Error);
+
+        let parent_rules = root_sampling_config
+            .map_or(default_value, |config| &config.rules_v2)
+            .iter()
+            .filter(|&rule| rule.ty == RuleType::Trace);
```
jjbayer: Do you ever need to pass in a non-empty `default_value`? If not, I would remove that parameter and write

```suggestion
            .into_iter()
            .flat_map(|config| config.rules_v2.iter())
            .filter(|&rule| rule.ty == RuleType::Trace);
```

relay-sampling/src/lib.rs
```diff
@@ -888,65 +888,31 @@ pub struct SamplingMatch {
...
-    /// Defines which population of items a dynamic sample rate applies to.
-    #[serde(default)]
-    pub mode: SamplingMode,
-}
-
-impl SamplingConfig {
-    pub fn has_unsupported_rules(&self) -> bool {
-        !self.rules_v2.iter().all(SamplingRule::supported)
+impl SamplingMatch {
+    pub fn with_new_sample_rate(&mut self, new_sample_rate: f64) {
```
```suggestion
    /// Setter for `sample_rate`.
    pub fn set_sample_rate(&mut self, new_sample_rate: f64) {
```

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -36,78 +36,33 @@ pub enum SamplingResult {
 #[derive(Clone, Debug, PartialEq)]
 enum SamplingMatchResult {
     /// The incoming event/dynamic sampling context matched the sampling configuration.
-    Match {
-        sample_rate: f64,
-        matched_rule_ids: MatchedRuleIds,
-        seed: Uuid,
-    },
+    Match { match_result: SamplingMatch },
```
```suggestion
    Match(SamplingMatch),
```

relay-sampling/src/lib.rs
```diff
@@ -1990,37 +2264,60 @@ mod tests {
...
+                ]
+            },
+            "samplingValue": {"type": "sampleRate", "value": 0.7},
+            "type": "trace",
+            "id": 1
+        }"#;
+        let rule: Result<SamplingRule, _> = serde_json::from_str(serialized_rule);
+
+        assert!(rule.is_ok());
+        let rule = rule.unwrap();
```
jjbayer: The test will fail either way:

```suggestion
        let rule = serde_json::from_str(serialized_rule).unwrap();
```

relay-sampling/src/lib.rs
```diff
@@ -2182,408 +2572,288 @@ mod tests {
...
-            ..Default::default()
-        };
-        let result = rules.get_matching_event_rule(&old_release, None, Utc::now());
-        assert_eq!(result.unwrap().id, RuleId(5), "Did not match expected rule");
+        // early return of first rule
+        let event = mocked_event(EventType::Transaction, "healthcheck", "1.1.1", "testing");
+        let dsc =
+            mocked_dynamic_sampling_context("root_transaction", "1.1.1", "debug", "vip", "user-id");
+        let result = match_against_rules(&config, &event, &dsc, Utc::now());
+        transaction_match!(result, 0.1, event, 1);
```
jjbayer: nit: Putting `assert` in the name might help clarify the intent of the test (same goes for other macros).
```suggestion
        assert_transaction_match!(result, 0.1, event, 1);
```

relay-sampling/src/lib.rs
```diff
@@ -2813,10 +3083,75 @@ mod tests {
...
             "condition": {"op": "and", "inner": []}
         }))
         .unwrap();
         assert!(!rule.supported());
     }
+
+    #[test]
+    /// Tests the merged config of the two configs with rules.
+    fn test_get_merged_config_with_rules_in_both_project_config_and_root_project_config() {
+        match_rule_ids!(
```
jjbayer: I have hard time understanding what this test does. Any way we can make it more explicit what it actually verifies?

jjbayer (changes requested):
I think we are close to merging this. Please take a look at the remaining comments. Could you also write a PR description that includes the motivation for this change and a summary of the change itself?

relay-sampling/src/lib.rs
iambriccardo: I wanted to do the same, thanks for the confirmation!

relay-sampling/src/lib.rs
iambriccardo: This test checks whether the merged rules of two configs are equal to a specific array of rule ids. Which part of the test you don't understand?

jjbayer: Thanks, I get it now. The name change helped.

jjbayer (approved):
Thanks for contributing! Final notes:

1. Do not merge until https://github.com/getsentry/sentry/pull/44229 has landed (as discussed). Otherwise, Relay will fail to parse project configs because of the missing `rulesV2` field. 
2. Please update the PR description (see [docs](https://develop.sentry.dev/code-review/#describe-what-your-pr-does-in-a-few-sentences-in-the-description-field)).
3. There's some new deprecation warnings, but I assume they are easy to fix.

iambriccardo:
> Thanks for contributing! Final notes:
> 
> 1. Do not merge until [getsentry/sentry#44229](https://github.com/getsentry/sentry/pull/44229) has landed (as discussed). Otherwise, Relay will fail to parse project configs because of the missing `rulesV2` field.
> 2. Please update the PR description (see [docs](https://develop.sentry.dev/code-review/#describe-what-your-pr-does-in-a-few-sentences-in-the-description-field)).
> 3. There's some new deprecation warnings, but I assume they are easy to fix.

I will have also to write the changelog.

PR #1791, opened 2023-01-30
feat(replays): Remove error and session sample rate fields and replace with context fields

cmanallen:
The SDK team has determined that the `contexts` object is the more appropriate place for these fields to live.  Therefore, we can remove these unnecessary fields and allow them to be parsed as a member of the contexts object.

### DIFF
--- a/tests/integration/test_replay_events.py
+++ b/tests/integration/test_replay_events.py
@@ -6,8 +6,6 @@ def generate_replay_sdk_event():
         "type": "replay_event",
         "replay_id": "d2132d31b39445f1938d7e21b6bf0ec4",
         "replay_type": "session",
-        "error_sample_rate": 0.125,
-        "session_sample_rate": 0.5,
         "event_id": "d2132d31b39445f1938d7e21b6bf0ec4",
         "segment_id": 0,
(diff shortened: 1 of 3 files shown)


tests/integration/test_replay_events.py
```diff
@@ -38,7 +36,11 @@ def generate_replay_sdk_event():
                 "trace_id": "4C79F60C11214EB38604F4AE0781BFB2",
                 "span_id": "FA90FDEAD5F74052",
                 "type": "trace",
-            }
+            },
+            "replay": {
+                "error_sample_rate": 0.125,
+                "session_sample_rate": 0.5,
+            },
```
jjbayer, line 43: This can go into a separate PR, but we should add this new Context to the protocol. see for example 

https://github.com/getsentry/relay/blob/23c24a93dd9c26749cbef6fc49a52d70d25e91bd/relay-general/src/protocol/contexts/browser.rs#L6-L16

jjbayer (approved):

tests/integration/test_replay_events.py
cmanallen: 👍

PR #1793, opened 2023-01-30
feat(replays): Add handling for uncompressed payload types

cmanallen:
Couple of things happening here:

- Functional change allowing payloads *which could not be decompressed* to be move forward in the pipeline any way.  This allows us to accept non-compressed payloads which would have otherwise been rejected.
- The "protocol" related behavior.  Specifically compression, decompression, header parsing, and header validation have been moved to their own module.  This was done to separate the parser and the implementation details necessary to interact with the service.

### DIFF
--- a/relay-replays/src/recording/protocol.rs
+++ b/relay-replays/src/recording/protocol.rs
@@ -1,0 +1,164 @@
+/// Protocol.
+///
+/// The following section includes the behaviors responsible for deserializing a recording from
+/// its HTTP transport schema and serializing a recording to its Kafka transport schema.
+///
+/// We expect recordings to come in the format of plaintext headers (JSON encoded), then a new
+/// line character, then an optionally compressed RRWeb recording.  Failure to include any
+/// component of this schema results in an error.
+use flate2::{read::ZlibDecoder, write::ZlibEncoder, Compression};
+use serde_json::{Error as SError, Value};
+use std::fmt::Display;
+use std::io::{Error, Read, Write};
+
+/// Deserializes input bytes and returns a tuple of serialized (confirmed valid) JSON headers
+/// and the optionally decompressed body value.
+pub fn deserialize(bytes: &[u8], limit: usize) -> Result<(&[u8], Vec<u8>), ProtocolError> {
+    let (headers, body) = read(bytes)?;
+
+    // We always attempt to decompress the body value. If decompression fails we try to JSON
+    // deserialize the body bytes as is.
+    match decompress(body, limit) {
+        Ok(buf) => Ok((headers, buf)),
+        Err(_) => Ok((headers, body.into())),
+    }
+}
+
+/// Serializes the headers and body arguments into a single vec of bytes. The body value is
+/// compressed before being concatenated.  The final output is headers-bytes + "\n" (new-line
+/// character) + compressed-body-bytes.
+pub fn serialize(headers: &[u8], body: Vec<u8>) -> Result<Vec<u8>, ProtocolError> {
+    let output_bytes = compress(body)?;
+    Ok([headers.into(), vec![b'\n'], output_bytes].concat())
+}
+
+fn compress(output: Vec<u8>) -> Result<Vec<u8>, ProtocolError> {
+    let mut encoder = ZlibEncoder::new(Vec::new(), Compression::default());
+    encoder
+        .write_all(&output)
+        .map_err(ProtocolError::EncodingError)?;
+    encoder.finish().map_err(ProtocolError::EncodingError)
+}
+
+fn decompress(zipped_bytes: &[u8], limit: usize) -> Result<Vec<u8>, ProtocolError> {
+    let mut buffer = Vec::new();
+
+    let decoder = ZlibDecoder::new(zipped_bytes);
+    decoder
+        .take(limit as u64)
+        .read_to_end(&mut buffer)
+        .map_err(ProtocolError::EncodingError)?;
+
+    Ok(buffer)
+}
+
+fn read(bytes: &[u8]) -> Result<(&[u8], &[u8]), ProtocolError> {
+    match bytes.is_empty() {
+        true => Err(ProtocolError::MissingData),
+        false => {
+            let mut split = bytes.splitn(2, |b| b == &b'\n');
+            let header = split.next().ok_or(ProtocolError::MissingHeaders)?;
+
+            // Try to parse the headers to determine if they are valid JSON. This is a good sanity
+            // check to determine if our headers extraction is working properly.
+            serde_json::from_slice::<Value>(header).map_err(ProtocolError::InvalidHeaders)?;
+
+            let body = match split.next() {
+                Some(b"") | None => return Err(ProtocolError::MissingBody),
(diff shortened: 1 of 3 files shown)


relay-replays/src/recording/protocol.rs
```diff
@@ -0,0 +1,164 @@
...
+fn read(bytes: &[u8]) -> Result<(&[u8], &[u8]), ProtocolError> {
+    match bytes.is_empty() {
+        true => Err(ProtocolError::MissingData),
+        false => {
+            let mut split = bytes.splitn(2, |b| b == &b'\n');
+            let header = split.next().ok_or(ProtocolError::MissingHeaders)?;
+
+            // Try to parse the headers to determine if they are valid JSON. This is a good sanity
+            // check to determine if our headers extraction is working properly.
+            serde_json::from_slice::<Value>(header).map_err(ProtocolError::InvalidHeaders)?;
```
JoshFerge: could we define our headers to a type to not use `Value` here? may also be in general a good idea to have a source of truth for our headers schema 🤔

cmanallen: Great idea!  Added.

jan-auer:
I'd like to put this one on hold until we've made progress on the memory issue.

That said, instead of ignoring compression errors, please match on the magic bytes of the compression header. If you're reading a broken gzip compressed payload, you wanna know that there was an error decompressing it. Only if the payload is not compressed, you should pipe it through.

relay-replays/src/recording/protocol.rs
```diff
@@ -0,0 +1,171 @@
...
+/// its HTTP transport schema and serializing a recording to its Kafka transport schema.
+///
+/// We expect recordings to come in the format of plaintext headers (JSON encoded), then a new
+/// line character, then an optionally compressed RRWeb recording.  Failure to include any
+/// component of this schema results in an error.
+use flate2::{read::ZlibDecoder, write::ZlibEncoder, Compression};
+use serde::Deserialize;
+use serde_json::Error as SError;
+use std::fmt::Display;
+use std::io::{Error, Read, Write};
```
jan-auer: Please reorder imports:

```suggestion
use std::fmt::Display;
use std::io::{Error, Read, Write};

use flate2::{read::ZlibDecoder, write::ZlibEncoder, Compression};
use serde::Deserialize;
use serde_json::Error as SError;
```

relay-replays/src/recording/protocol.rs
```diff
@@ -0,0 +1,171 @@
...
+/// Deserializes input bytes and returns a tuple of serialized (confirmed valid) JSON headers
+/// and the optionally decompressed body value.
+pub fn deserialize(bytes: &[u8], limit: usize) -> Result<(&[u8], Vec<u8>), ProtocolError> {
+    let (headers, body) = read(bytes)?;
+
+    // We always attempt to decompress the body value. If decompression fails we try to JSON
+    // deserialize the body bytes as is.
+    match decompress(body, limit) {
+        Ok(buf) => Ok((headers, buf)),
+        Err(_) => Ok((headers, body.into())),
```
jan-auer, line 25: Ignoring errors here can and will lead to other errors downstream. Instead, please check if the body is compressed by inspecting magic bytes, and run decompression conditionally.

relay-replays/src/recording/protocol.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+/// Serializes the headers and body arguments into a single vec of bytes. The body value is
+/// compressed before being concatenated.  The final output is headers-bytes + "\n" (new-line
+/// character) + compressed-body-bytes.
+pub fn serialize(headers: &[u8], body: Vec<u8>) -> Result<Vec<u8>, ProtocolError> {
+    let output_bytes = compress(body)?;
+    Ok([headers.into(), vec![b'\n'], output_bytes].concat())
+}
```
jan-auer: This allocates a bunch of vectors just to create yet another one. Also, this moves `body` without actually consuming it. Either, change to `body: &[u8]` or mutate body via `body: &mut Vec<u8>`. Semantically, the first version is probably better given that you compress it:

```suggestion
pub fn serialize(headers: &[u8], body: & [u8]) -> Result<Vec<u8>, ProtocolError> {
    let output_bytes = compress(body)?;
    Ok([headers, &b"\n", &output_bytes].concat())
}
```

This also requires to change the signature of `compress`.

relay-replays/src/recording/protocol.rs
```diff
@@ -0,0 +1,171 @@
...
+/// component of this schema results in an error.
+use flate2::{read::ZlibDecoder, write::ZlibEncoder, Compression};
+use serde::Deserialize;
+use serde_json::Error as SError;
+use std::fmt::Display;
+use std::io::{Error, Read, Write};
+
+/// Deserializes input bytes and returns a tuple of serialized (confirmed valid) JSON headers
+/// and the optionally decompressed body value.
+pub fn deserialize(bytes: &[u8], limit: usize) -> Result<(&[u8], Vec<u8>), ProtocolError> {
```
jan-auer: _nit_: This function does not actually deserialize, can we give it a better name?

relay-replays/src/recording/protocol.rs
```diff
@@ -0,0 +1,171 @@
...
+        .read_to_end(&mut buffer)
+        .map_err(ProtocolError::EncodingError)?;
+
+    Ok(buffer)
+}
+
+fn read(bytes: &[u8]) -> Result<(&[u8], &[u8]), ProtocolError> {
+    match bytes.is_empty() {
+        true => Err(ProtocolError::MissingData),
+        false => {
```
jan-auer: Please do not match on a bool, use `if` instead.

```suggestion
    if bytes.is_empty() {
        return Err(ProtocolError::MissingData);
    }
    
    // ...
```

relay-replays/src/recording/protocol.rs
```diff
@@ -0,0 +1,171 @@
...
+mod tests {
+    use crate::recording::protocol::{deserialize, serialize, ProtocolError};
+    use serde_json::Value;
+
+    #[test]
+    fn test_deserialize_message() {
+        // Contains a 16 byte header followed by a new line character.
+        let payload: &[u8] = &[
+            123, 34, 115, 101, 103, 109, 101, 110, 116, 95, 105, 100, 34, 58, 51, 125, 10, 120,
+        ];
```
jan-auer:
```suggestion
        let payload = b"{\"segment_id\":3}\nx".as_slice();
```

relay-replays/src/recording/protocol.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        let result = deserialize(payload, 9999999999);
+        assert!(result.is_ok());
```
jan-auer: Instead of asserting `is_ok()` here, just unwrap. This will print the error and fail the test automatically.

jan-auer (changes requested):

jan-auer:
This has been incorporated into #1800. I'll be following up with some more module separation, for now we can close this. Thank you for the nudge!

Issue #1680, opened 2022-12-07
Error sending a metric: channel full


olksdr:
This happened few times, our statsd metrics could not be sent.

Sentry Issue: [POP-RELAY-2NV](https://sentry.my.sentry.io/organizations/sentry/issues/316057/?referrer=github_integration)

```
Error sending a metric: channel full
  caused by: channel full, maximum capacity: 100000
```

Error is coming from https://github.com/getsentry/relay/blob/576466bc15902959f55107b061a2067e33b83ae6/relay-statsd/src/lib.rs#L117

iker-barriocanal:
[Investigation doc (internal)](https://www.notion.so/sentry/Can-t-send-metrics-because-the-channel-is-full-3d464b2250fe4e2789f1884a31fd28ee).

jan-auer:
We have concluded the only viable short-term option is to reduce the amount of metrics. We are going to analyze which metrics are creating too much traffic and will reduce them. Note that this is NOT a cardinality concern, it's that our individual increments of metrics are too much.

Other considered options:
- **Increase channel size:** This option doesn’t really address the problem, and it only postpones it. The cost is the increased CPU and memory usage, which currently may be around 20% of a relay instance’s cost.
- **Increase number of channels or clients:** Same as above.
- **“Optimize” the channel**: To “optimize” the channel for higher throughput, it is recommended to move from UDP to Unix Domain Socket (UDS). UDS is another, strem-oriented protocol (supports UDP). However, there aren’t any guarantees of increased performance (we need to run benchmarks for that), comes with its own set of challenges, and requires infrastructure changes. The main challenge is based on how UDS fundamentally works, based on a special file.

PR #1797, opened 2023-01-30
ref(statsd): Add metric name as a tag for Sentry errors

iker-barriocanal:
From time to time, we still get errors from relay dropping metrics due to the channel being full, see https://github.com/getsentry/relay/issues/1680. This PR attempts to add a tag to that Sentry issue, to identify more easily what metrics may be worsening the bottleneck. Next steps TBD, based on conclusions from the errors. This code is likely to be temporary and be removed after we figure out what and how to deal with the errors.

### The approach

The original idea was to extract the error and directly include it in the error here:
https://github.com/getsentry/relay/blob/2ec2a5e47a577fb6b7660624093df95c1db05d21/relay-statsd/src/lib.rs#L115-L116

There are two problems with that approach. Firstly, the error types come from statsd's source code and it's hard to consider every case. Secondly, the code is too generic and closed that it's hard to inject something nicely without a decent amount of workarounds. Lastly, I also tried replicating the issue locally generating a lot of metrics (lots of threads, queue size to 0 and 1, sample rate 1.0), but I have not been successful.

In the end, the simplest solution was to add the name as a tag to the scope where there's access to it, wrapping the call where the actual error is logged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 **Internal**:
 
-- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1797](https://github.com/getsentry/relay/issues/1797))
+- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1797](https://github.com/getsentry/relay/pull/1797))
 
 ## 23.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Add support for client hints. ([#1752](https://github.com/getsentry/relay/pull/1752))
 
+**Internal**:
+
+- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1680](https://github.com/getsentry/relay/issues/1680))
+
 ## 23.1.1
 
 **Features**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 **Internal**:
 
-- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1680](https://github.com/getsentry/relay/issues/1680))
+- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1797](https://github.com/getsentry/relay/issues/1797))
 
 ## 23.1.1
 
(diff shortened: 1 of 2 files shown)


jan-auer:
Adding the name as tag will not help us to know, which metrics are being logged most. You simply know the one that's next in line. Compared with the increased complexity, I'm not sure this change is worth it.

relay-statsd/src/lib.rs
```diff
@@ -492,81 +492,122 @@ pub trait GaugeMetric {
...
+                    client.time_with_tags(&$crate::TimerMetric::name(&$id), $value)
+                        $(.with_tag(stringify!($k), $v))*
+                )
+            })
         })
     };
 
     // timed block
     (timer($id:expr), $($k:ident = $v:expr,)* $block:block) => {{
+        use $crate::_pred::*;
```
jjbayer: Why did this particular import move up?

iker-barriocanal:
> Adding the name as tag will not help us to know, which metrics are being logged most.

No, but metrics that are logged more often are more likely to be in tags, so I believe it can give us an approximation. Building something to efficiently (we don't want to lose even more metrics, especially these) record the metrics we log and analyzing results is too much effort, vs going through the metrics in the codebase and directly deleting some. In the end, this is about the best effort and finding a temporary solution. If this, still, doesn't provide enough value, we can call it off.

jjbayer (approved):
I'm sure there's ways to reduce code duplication here, but since these macro definitions are rarely touched, I think we can go ahead with the change as-is.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Add support for client hints. ([#1752](https://github.com/getsentry/relay/pull/1752))
 
+**Internal**:
+
+- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1680](https://github.com/getsentry/relay/issues/1680))
```
```suggestion
- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1797](https://github.com/getsentry/relay/issues/1797))
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Add support for client hints. ([#1752](https://github.com/getsentry/relay/pull/1752))
 
+**Internal**:
+
+- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1797](https://github.com/getsentry/relay/issues/1797))
```
```suggestion
- Add metric name as tag on Sentry errors from relay dropping metrics. ([#1797](https://github.com/getsentry/relay/pull/1797))
```

PR #1798, opened 2023-01-31
feat(replays): Replace use of Value type with explicit types

cmanallen:
This should be an interesting canary test.  If we remove most of the usages of `Value` (which is memory intensive) can we achieve a sustainable, stable, long-term memory growth curve?

- Not all variants have been typed.
    - I focused on the most popular payload types first.
    - Will finish the remainder if the Canary test shows promising results (or I might just type them anyway as capacity allows).
- Not all fields have had their `Value` type replaced.  Those fields will require a custom deserializer which may be overkill for a canary test (should the test go poorly).
- Verified as working against >300 different RRWeb payloads.

cc: @jan-auer @jjbayer Thoughts on this as an aide to reducing resource usage?  I think the compounding effects of `Value` could be large.

#skip-changelog

### DIFF
--- a/relay-replays/src/recording/mod.rs
+++ b/relay-replays/src/recording/mod.rs
@@ -531,9 +531,9 @@ struct MouseInteractionIncrementalSourceData {
 struct ScrollIncrementalSourceData {
     id: i32,
     #[serde(skip_serializing_if = "Option::is_none")]
-    x: Option<u64>,
+    x: Option<f64>,
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


jan-auer:
The protocol has been removed from the codebase in #1800. @jjbayer and I have agreed that we would keep it removed until there's a new need for a typed out protocol. 

Once we start processing replay payloads again, we can re-introduce a protocol. However, that protocol might better go into its own crate so it can be used better in another service.

Thank you for this contribution, we might very well come back to this.

PR #1801, opened 2023-01-31
feat(profiling): Add or remove the profile ID on the transaction's profiling context

phacops:
We added a way to set the profile ID in a profile context on the transaction. Now, we need to set the `profile_id` or remove it from the context if it's already set by the SDK in case the profile succeeds or failed to be processed in order to have accurate transaction to profile link. We don't want a transaction set with a profile ID that doesn't exist and we don't want a transaction without a profile ID that do exist.

This PR will process the profiles after the transaction is processed and then take the appropriate action and modify the profile context on the transaction.


relay-profiling/src/lib.rs
```diff
@@ -133,16 +137,20 @@ fn minimal_profile_from_json(data: &[u8]) -> Result<MinimalProfile, ProfileError
     serde_json::from_slice(data).map_err(ProfileError::InvalidJson)
 }
 
-pub fn expand_profile(payload: &[u8]) -> Result<Vec<u8>, ProfileError> {
-    let profile: MinimalProfile = minimal_profile_from_json(payload)?;
-    match profile.version {
+pub fn expand_profile(payload: &[u8]) -> (Option<EventId>, Result<Vec<u8>, ProfileError>) {
```
Zylphrex: Would a return type like `Result<(EventId, Vec<u8>), ProfileError>` be better here since this leaves the possibility of a return value like `(None, Ok(...))`. It would also clean up the `match` in the processor.

phacops: It would clean up the `match` in the processor but since I'll have to change the way to handle errors for each `parse_<platform>_profile`, I suspect it won't improve the code overall.

Now, removing the possibility of a `(None, Ok(...))` is interesting, although Rust wouldn't let us compile if we're not taking care of all the cases at which point I believe we'd see the bug. I'll try to rework it but I don't think it'd be a full win.

jan-auer:
Tests are still failing at large, probably due to the moved process function?

phacops:
> Tests are still failing at large, probably due to the moved process function?

@jan-auer I fixed the issue, but now the code is not... pretty. Maybe you can suggest a way to make it better?

relay-server/src/actors/processor.rs
```diff
@@ -1013,42 +1014,60 @@ impl EnvelopeProcessorService {
...
-            ItemType::Profile => match relay_profiling::expand_profile(&item.payload()) {
-                Ok(payload) => {
-                    if payload.len() <= self.config.max_profile_size() {
-                        item.set_payload(ContentType::Json, payload);
-                        true
-                    } else {
+            ItemType::Profile => {
+                let contexts = match state.event.value_mut() {
+                    Some(event) if event.ty.value() == Some(&EventType::Transaction) => {
+                        event.contexts.value_mut().as_mut()
```
jjbayer: Don't we also want to create `contexts` here if it does not exist? I.e. https://github.com/getsentry/relay/blob/57aacf07c7665e8d995e79648d98e8fadbabec8d/relay-general/src/protocol/replay.rs#L283

relay-server/src/actors/processor.rs
```diff
@@ -1013,42 +1014,60 @@ impl EnvelopeProcessorService {
...
+                    }
+                    _ => None,
+                };
+                match relay_profiling::expand_profile(&item.payload()) {
+                    Ok((profile_id, payload)) => {
+                        if payload.len() <= self.config.max_profile_size() {
+                            if let Some(contexts) = contexts {
+                                contexts.add(SentryContext::Profile(Box::new(ProfileContext {
+                                    profile_id: Annotated::new(profile_id),
+                                })));
```
jjbayer, line 1037: This overwrites the context if it was already set by an SDK. I assume this is intentional?

relay-server/src/actors/processor.rs
```diff
@@ -2156,6 +2174,9 @@ impl EnvelopeProcessorService {
             });
         }
 
+        // We need the event parsed in order to set the profile context on it
+        self.process_profiles(state);
```
jjbayer: This will now only process profiles for events that were not dropped by dynamic sampling. Is that OK?

relay-server/src/actors/processor.rs
phacops: This is indeed intentional.

relay-server/src/actors/processor.rs
phacops: Yes, we don't want to process profiles for transactions that were drop.

We still have one SDK that sends profiles by themselves (and so will set the `profile_id` in the context themselves until we can send that in the same envelope) and these profiles should still be processed since there's no dynamic sampling rules around profiles yet, right?

relay-server/src/actors/processor.rs
phacops: Not too sure what the right move here is, but when I had this, a lot of integration tests started failing. I'm not sure why, maybe you can help shed some lights on this?

Also, I'm not sure how realistic it is for a transaction to have 0 context, so I thought this was acceptable.

relay-profiling/src/lib.rs
```diff
@@ -133,15 +137,22 @@ fn minimal_profile_from_json(data: &[u8]) -> Result<MinimalProfile, ProfileError
     serde_json::from_slice(data).map_err(ProfileError::InvalidJson)
 }
 
-pub fn expand_profile(payload: &[u8]) -> Result<Vec<u8>, ProfileError> {
-    let profile: MinimalProfile = minimal_profile_from_json(payload)?;
-    match profile.version {
+pub fn expand_profile(payload: &[u8]) -> Result<(EventId, Vec<u8>), ProfileError> {
+    let profile: MinimalProfile = match minimal_profile_from_json(payload) {
```
jan-auer: The compiler can infer this type, no type hint needed.

```suggestion
    let profile = match minimal_profile_from_json(payload) {
```

relay-profiling/src/lib.rs
```diff
@@ -133,15 +137,22 @@ fn minimal_profile_from_json(data: &[u8]) -> Result<MinimalProfile, ProfileError
...
             Platform::Android => parse_android_profile(payload),
             Platform::Cocoa => parse_cocoa_profile(payload),
-            _ => Err(ProfileError::PlatformNotSupported),
+            _ => return Err(ProfileError::PlatformNotSupported),
         },
+    };
+    match processed_payload {
+        Ok(payload) => Ok((profile.event_id, payload)),
+        Err(err) => Err(err),
     }
```
jan-auer: _nit_: Can be written with `map`:

```suggestion
    processed_payload.map(|payload| (profile.event_id, payload))
```

relay-server/src/actors/processor.rs
jan-auer: Transactions will always have contexts in practice (the trace context is even required). Still, `get_or_insert_with` is more explicit, and also improves the quality of our tests where we had broken fixtures.

The CI failures occurred because of [this change](https://github.com/getsentry/relay/pull/1801/files/a3da2f056fc136d876b0cfe3c4101a0aa2db5140#diff-ff64e5542d3e0a29fab7ed5550d22511a6ca00cefaae3bc9a8de7b158499d792R980). `process_profiles` runs on _every_ envelope, even the ones without a profile. You unconditionally create an empty error event (without an event ID), which does not pass assertions during storage. This creates an error which caused test failures. 

You should check for the existence of an event, but within the event, you can always create the context.

relay-server/src/actors/processor.rs
```diff
@@ -1013,42 +1014,60 @@ impl EnvelopeProcessorService {
...
+                                contexts.add(SentryContext::Profile(Box::new(ProfileContext {
+                                    profile_id: Annotated::new(profile_id),
+                                })));
+                            }
+                            item.set_payload(ContentType::Json, payload);
+                            true
+                        } else {
+                            if let Some(contexts) = contexts {
+                                contexts.remove(ProfileContext::default_key());
+                            }
```
jan-auer: This is a dead branch, since otherwise execution would not enter `else` here.

relay-server/src/actors/processor.rs
jan-auer: Most of this function is constrained to processing mode. Would it make sense to:
 - Have a small function checking for the profiling feature flag above
 - Make the rest `#[cfg(feature = "processing")]` and call it after `enforce_quotas` inside `if_processing`
 
 ?

relay-server/src/actors/processor.rs
jjbayer: This branch handles the case where the payload is too large, right?

relay-server/src/actors/processor.rs
jjbayer: Yes, that should work.

jjbayer: Sounds good to me.

relay-server/src/actors/processor.rs
jan-auer: Sorry, I was getting the indentation wrong. Please disregard this comment.

relay-server/src/actors/processor.rs
```diff
@@ -999,27 +1000,38 @@ impl EnvelopeProcessorService {
...
             _ => true,
         });
+    }
 
-        if !self.config.processing_enabled() {
-            return;
-        }
-
-        envelope.retain_items(|item| match item.ty() {
+    /// Remove profiles if the feature flag is not enabled
```
Zylphrex: This comment is for the `filter_profiles` function above right?

Zylphrex (approved):

relay-server/src/actors/processor.rs
```diff
@@ -2138,9 +2156,11 @@ impl EnvelopeProcessorService {
         self.process_sessions(state);
         self.process_client_reports(state);
         self.process_user_reports(state);
-        self.process_profiles(state);
         self.process_replays(state)?;
 
+        // Remove profiles from the payload if the feature is not enabled
+        self.filter_profiles(state);
```
jan-auer, line 2174: _nit_: You can keep it with the group above. The name of the function is clear enough, and it's similar to the other functions.

relay-server/src/actors/processor.rs
```diff
@@ -999,27 +1000,38 @@ impl EnvelopeProcessorService {
...
-        envelope.retain_items(|item| match item.ty() {
+    /// Remove profiles if the feature flag is not enabled
+    #[cfg(feature = "processing")]
+    fn process_profiles(&self, state: &mut ProcessEnvelopeState) {
+        state.envelope.retain_items(|item| match item.ty() {
             ItemType::Profile => match relay_profiling::expand_profile(&item.payload()) {
-                Ok(payload) => {
+                Ok((profile_id, payload)) => {
                     if payload.len() <= self.config.max_profile_size() {
+                        if let Some(event) = state.event.value_mut() {
```
jan-auer, line 1032: Originally, we tried to keep this file free of code that actually modifies events. I do not see a quick way of refactoring this, so let's keep it. Eventually, we'll need to clean this file at large.

jan-auer (approved):

relay-server/src/actors/processor.rs
phacops: Sounds good.

PR #1802, opened 2023-01-31
feat: Add client hints for requestmeta

TBS1996:
Second PR about supporting client hints. #1183 
previous PR: #1752 

This PR adds support for client hints on the requestmeta struct, which means that the client hints can be sent as part of an event json as an optional field. This means that more information will be extracted from the client hints rather than the user agent string, as relay will first attempt to check the client hints and only falling back to the user agent string if no client hints are found.


relay-server/src/actors/processor.rs
```diff
@@ -2145,8 +2146,8 @@ impl EnvelopeProcessorService {
                 if let Some(client) = client {
                     scope.set_tag("sdk", client);
                 }
-                if let Some(user_agent) = user_agent {
-                    scope.set_extra("user_agent", user_agent.into());
+                if let Some(user_agent) = user_agent.user_agent.as_ref() {
+                    scope.set_extra("user_agent", user_agent.as_str().into());
```
TBS1996, line 2150: Should I also add the client hints field here?

relay-server/src/extractors/request_meta.rs
```diff
@@ -280,22 +288,7 @@ impl RequestMeta {
             user_agent: Some(crate::constants::SERVER.to_owned()),
             no_cache: false,
             start_time: Instant::now(),
-        }
-    }
-
-    #[cfg(test)]
-    // TODO: Remove Dsn here?
-    pub fn new(dsn: relay_common::Dsn) -> Self {
```
TBS1996, line 288: moved to mod tests{}

relay-general/src/store/mod.rs
```diff
@@ -63,6 +64,8 @@ pub struct StoreConfig {
 
     /// The SDK's sample rate as communicated via envelope headers.
     pub client_sample_rate: Option<f64>,
+
+    pub client_hints: ClientHints<String>,
```
TBS1996: Added clienthints as a separate field due to backwards compatibility wrt serialization. If this is not necessary for StoreConfig, let me know, and i'll instead use RawUserAgentInfo here.

relay-server/src/extractors/request_meta.rs
```diff
@@ -363,6 +356,12 @@ pub type PartialMeta = RequestMeta<Option<PartialDsn>>;
 impl PartialMeta {
     /// Extracts header information except for auth info.
     fn from_headers<S>(request: &HttpRequest<S>) -> Self {
+        let mut ua = RawUserAgentInfo::default();
+
+        for (key, value) in request.headers() {
+            ua.extract_header(key.as_str(), value.to_str().ok().map(str::to_string));
```
TBS1996: This solution is a bit inelegant, but since RawUserAgentInfo is defined in another crate I would have had to import Request library in General to make this a method of RawUserAgentInfo. Since the other Header type is defined by us, we could perhaps implement From\<HttpRequest\> to make it prettier.

relay-general/src/user_agent.rs
```diff
@@ -83,51 +84,119 @@ pub fn parse_os(user_agent: &str) -> OS {
...
-    pub sec_ch_ua_platform_version: Option<&'a str>,
-    /// Name of the client's web browser and its version.
-    pub sec_ch_ua: Option<&'a str>,
-    /// Device model, e.g. samsung galaxy 3.
-    pub sec_ch_ua_model: Option<&'a str>,
+impl<S: AsRef<str> + Default> RawUserAgentInfo<S> {
+    /// Checks if key matches a user agent header, in which case it sets the value accordingly.
+    /// Extracted out from from_headers so that it can be called from another crate with different
+    /// header-types. Perhaps we could make it generic over different header types instead.
+    pub fn extract_header(&mut self, key: &str, value: Option<S>) {
```
iker-barriocanal: This method sets a value but it's called `extract*` instead, can we choose a different name that reflects better its functionality?

relay-general/src/user_agent.rs
```diff
@@ -83,51 +84,119 @@ pub fn parse_os(user_agent: &str) -> OS {
...
-    /// The version number of the client's OS.
-    pub sec_ch_ua_platform_version: Option<&'a str>,
-    /// Name of the client's web browser and its version.
-    pub sec_ch_ua: Option<&'a str>,
-    /// Device model, e.g. samsung galaxy 3.
-    pub sec_ch_ua_model: Option<&'a str>,
+impl<S: AsRef<str> + Default> RawUserAgentInfo<S> {
+    /// Checks if key matches a user agent header, in which case it sets the value accordingly.
+    /// Extracted out from from_headers so that it can be called from another crate with different
+    /// header-types. Perhaps we could make it generic over different header types instead.
```
iker-barriocanal:
```suggestion
    /// Checks if key matches a user agent header, in which case it sets the value accordingly.
    ///
    /// TODO(tor): make it generic over different header types.
```
> Extracted out from from_headers so that it can be called from another crate with different header-types.

This sentence suits as a comment in the PR, but I don't think it's useful in the docstrings.

> Perhaps we could make it generic over different header types instead.

This suggestion is probably personal, but I'd leave this as a TODO.

relay-general/src/user_agent.rs
```diff
@@ -83,51 +84,119 @@ pub fn parse_os(user_agent: &str) -> OS {
...
+            }
+            _ => {}
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.user_agent.is_none() && self.client_hints.is_empty()
+    }
+
+    pub fn from_ua(s: S) -> Self {
```
iker-barriocanal: Is this method used somewhere? If it's dead code, please delete it.

relay-general/src/user_agent.rs
```diff
@@ -83,51 +84,119 @@ pub fn parse_os(user_agent: &str) -> OS {
...
+    #[test]
+    fn detect_new_field() {
+        let ua = RawUserAgentInfo::<&str>::default();
+        let size = std::mem::size_of_val(&ua);
+        //dbg!(size);
+        assert!(
+            size == 80,
+            "Make sure to update all relevant code if you add a new field to RawUserAgentInfo"
+        );
+    }
```
iker-barriocanal: What's the value of adding this test?

relay-server/src/actors/processor.rs
```diff
@@ -1095,6 +1095,11 @@ impl EnvelopeProcessorService {
                     return false;
                 }
 
+                let user_agent = RawUserAgentInfo {
+                    user_agent: meta.user_agent(),
+                    client_hints: meta.client_hints().as_deref(),
+                };
```
iker-barriocanal: Placing this piece of code here means it will run for every item in the envelope. However, the user agent is the samefor all items. Let's move this code out of `retain_items`, where the user agent was extracted before for example.

relay-general/src/store/normalize.rs
```diff
@@ -593,12 +594,25 @@ fn normalize_security_report(
...
+        insert_header(
+            "SEC-CH-UA-Platform",
+            user_agent.client_hints.sec_ch_ua_platform,
+        );
+        insert_header(
+            "SEC-CH-UA-Platform-Version",
+            user_agent.client_hints.sec_ch_ua_platform_version,
+        );
+        insert_header("SEC-CH-UA", user_agent.client_hints.sec_ch_ua);
+        insert_header("SEC-CH-UA-Model", user_agent.client_hints.sec_ch_ua_model);
```
iker-barriocanal: I think we can do better than not hardcoding the client hint names here, and having these names in a single place.

relay-server/src/extractors/request_meta.rs
```diff
@@ -199,6 +200,9 @@ pub struct RequestMeta<D = PartialDsn> {
     // NOTE: This is internal-only and not exposed to Envelope headers.
     #[serde(skip, default = "Instant::now")]
     start_time: Instant,
+
+    #[serde(default, skip_serializing_if = "ClientHints::is_empty")]
+    client_hints: ClientHints<String>,
```
iker-barriocanal: I'd move this a bit upwards and place it next to `user_agent`.

relay-server/src/extractors/request_meta.rs
iker-barriocanal: In general, I'd suggest to keep PRs focused on their scope. More small and focused PRs is better than a few big PRs.

relay-server/src/extractors/request_meta.rs
```diff
@@ -413,6 +409,21 @@ impl PartialMeta {
...
+        }
+        if self.client_hints.sec_ch_ua_platform.is_some() {
+            complete.client_hints.sec_ch_ua_platform = self.client_hints.sec_ch_ua_platform;
+        }
+        if self.client_hints.sec_ch_ua_model.is_some() {
+            complete.client_hints.sec_ch_ua_model = self.client_hints.sec_ch_ua_model;
+        }
+        if self.client_hints.sec_ch_ua.is_some() {
+            complete.client_hints.sec_ch_ua = self.client_hints.sec_ch_ua;
+        }
```
iker-barriocanal: I'm not sure what the alternative to this looks like, but I don't really like going through every client hint here. I think moving this responsibility to the client hint struct is a better choice.

relay-server/src/extractors/request_meta.rs
```diff
@@ -571,3 +583,71 @@ impl FromRequest<ServiceState> for EnvelopeMeta {
...
+  "remote_addr": "192.168.0.1",
+  "forwarded_for": "8.8.8.8",
+  "user_agent": "0x8000",
+  "no_cache": false,
+  "client_hints":  {
+    "sec_ch_ua_platform": "macOS",
+    "sec_ch_ua_platform_version": "13.1.0",
+    "sec_ch_ua": "\"Not_A Brand\";v=\"99\", \"Google Chrome\";v=\"109\", \"Chromium\";v=\"109\""
+  }
+}"#;
```
iker-barriocanal: nit: tab this to the right.

relay-server/src/extractors/request_meta.rs
```diff
@@ -571,3 +583,71 @@ impl FromRequest<ServiceState> for EnvelopeMeta {
...
+                sec_ch_ua_model: None,
+            },
+        };
+        deserialized.start_time = reqmeta.start_time;
+        assert_eq!(deserialized, reqmeta);
+    }
+
+    impl RequestMeta {
+        // TODO: Remove Dsn here?
+        pub fn new(dsn: relay_common::Dsn) -> Self {
```
iker-barriocanal: nit: place this before the test above, before tests start.

relay-server/src/actors/processor.rs
TBS1996: oh good point, implemented now

relay-general/src/store/normalize.rs
TBS1996: i added them now as associated constants.

relay-server/src/extractors/request_meta.rs
TBS1996: hmm the function itself makes use of partial moves which makes it harder to extract the logic to a different function

relay-server/src/extractors/request_meta.rs
TBS1996: right, I was worried there would be serialization issues, but i guess with json it doesnt matter

relay-general/src/user_agent.rs
TBS1996: Some parts of code need to be updated when adding a new field which the compiler doesn't notice, I was wondering if it was possible to automate that, but yeah I don't think this is good enough so i'll remove this

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 **Features**:
 
-- Add support for client hints. ([#1752](https://github.com/getsentry/relay/pull/1752))
+- Add support for client hints. ([#1752](https://github.com/getsentry/relay/pull/1752), [#1802](https://github.com/getsentry/relay/pull/1802))
```
```suggestion
- Use client hint headers instead of User-Agent when available. ([#1752](https://github.com/getsentry/relay/pull/1752), [#1802](https://github.com/getsentry/relay/pull/1802))
```

relay-general/src/protocol/replay.rs
```diff
@@ -276,8 +280,8 @@ impl Replay {
 
         let mut user_agent_info = RawUserAgentInfo::from_headers(headers);
 
-        if user_agent_info.user_agent.is_none() {
-            user_agent_info.user_agent = default_user_agent;
+        if user_agent_info.is_empty() {
+            user_agent_info = default_user_agent.clone();
```
jjbayer: This clone should be unnecessary. `normalize_user_agent_info_generic` takes a reference, so something like this should work:

```rust
let user_agent_info = if ua_from_headers.is_empty() { default_user_agent } else { &ua_from_headers};
```

relay-server/src/extractors/request_meta.rs
```diff
@@ -571,3 +571,71 @@ impl FromRequest<ServiceState> for EnvelopeMeta {
...
+                dsn: PartialDsn::from_dsn(dsn).expect("invalid DSN"),
+                client: Some("sentry/client".to_string()),
+                version: 7,
+                origin: Some("http://origin/".parse().unwrap()),
+                remote_addr: Some("192.168.0.1".parse().unwrap()),
+                forwarded_for: String::new(),
+                user_agent: Some("sentry/agent".to_string()),
+                no_cache: false,
+                start_time: Instant::now(),
+                client_hints: ClientHints::default(),
```
jjbayer, line 592: Could you please add tests which use non-default `client_hints`? To find out what tests would be affected, change to a non-default value here and see if any tests break.

jjbayer (changes requested):

relay-general/src/user_agent.rs
```diff
@@ -201,6 +226,22 @@ impl ClientHints<String> {
 mod tests {
     use super::*;
 
+    //impl<S: AsRef<str> + Default> RawUserAgentInfo<S> {
```
jjbayer:
```suggestion
```

relay-general/src/user_agent.rs
```diff
@@ -126,6 +126,31 @@ impl RawUserAgentInfo<String> {
...
+        insert_header(
+            ClientHints::SEC_CH_UA_PLATFORM_VERSION,
+            self.client_hints.sec_ch_ua_platform_version,
+        );
+        insert_header(ClientHints::SEC_CH_UA, self.client_hints.sec_ch_ua);
+        insert_header(
+            ClientHints::SEC_CH_UA_MODEL,
+            self.client_hints.sec_ch_ua_model,
+        );
+    }
```
jjbayer, line 138: Can this function only operator on `&'a str`? If not, I would put it in the generic `impl<S: ...>` block.

relay-general/src/user_agent.rs
```diff
@@ -126,6 +126,31 @@ impl RawUserAgentInfo<String> {
 }
 
 impl<'a> RawUserAgentInfo<&'a str> {
+    pub fn populate_event_headers(&self, headers: &mut Headers) {
```
jjbayer, line 115:
```suggestion
    /// Convert user-agent info to HTTP headers as stored in the `Request` interface.
    ///
    /// This function does not overwrite any pre-existing headers.
    pub fn populate_event_headers(&self, headers: &mut Headers) {
```

jjbayer (approved):
LGTM, please apply the last couple of nitpicks (if they make sense) and write PR description that includes _why_ we are doing this, and what changes it makes to data passing through sentry.

relay-general/src/user_agent.rs
TBS1996: it relies on &str being copy, do you think I should make it generic and clone it instead ?

jjbayer: What if you call `as_ref` or `as_deref` on the values you pass into `insert_header`?

TBS1996: ok neat, it works that way yeah!

PR #1803, opened 2023-01-31
chore(replays): Add test coverage for disabled ip-address pii scrubbing

cmanallen:
We're seeing an issue where ip-addresses may be improperly scrubbed.  This pull asserts that the PII scrubbing functionality (as it pertains to ip-addresses) is functioning correctly.

@jjbayer Would you say this data scrubbing configuration closely mirrors production?  I'd like to create a PIIConfig that is as close to a production use case as possible.

### DIFF
--- a/relay-general/src/protocol/replay.rs
+++ b/relay-general/src/protocol/replay.rs
@@ -551,6 +551,43 @@ mod tests {
         assert_eq!(maybe_credit_card, Some("[Filtered]"));
     }
 
+    #[test]
+    fn test_ip_address_scrubbing_disabled() {
+        // Scrubbing is explicitly disabled.
+        let mut scrub_config = simple_enabled_config();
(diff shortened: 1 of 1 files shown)


jjbayer:
> @jjbayer Would you say this data scrubbing configuration closely mirrors production? I'd like to create a PIIConfig that is as close to a production use case as possible.

Yes, and as of https://github.com/getsentry/relay/pull/1725, the `scrub_ip_addresses` setting should scrub IP addresses in any field that is `pii = true`. But whether or not users enable IP scrubbing is up to them, and I am not sure if it is on by default.

cmanallen:
@jjbayer Thanks Joris!  We ended up tracking this to the ip-address normalization code so this pull isn't needed.  We could merge it if you think the coverage is useful though.

PR #1804, opened 2023-02-01
ref(replays): Parameterize the transcoder's function

jan-auer:
An internal refactor to unblock #1731.

#skip-changelog


olksdr (approved):
lgtm

relay-replays/src/transform.rs
```diff
@@ -0,0 +1,715 @@
...
+/// struct Identity;
+///
+/// let json = "42";
+/// let json_deserializer = &mut serde_json::Deserializer::from_str(&json);
+/// let deserializer = TransformingDeserializer::new(json_deserializer, Identity);
+///
+/// let number = deserializer.deserialize_u32(deserializer).unwrap();
+/// assert_eq!(number, 42);
+/// ```
+pub struct TransformingDeserializer<'a, D, T>(D, Mut<'a, T>);
```
jjbayer: nit: change to named fields for readability?

relay-replays/src/transform.rs
```diff
@@ -0,0 +1,715 @@
...
+    fn deserialize_ignored_any<V>(mut self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: de::Visitor<'de>,
+    {
+        self.0
+            .deserialize_ignored_any(Visitor(visitor, self.1.as_mut()))
+    }
+}
+
+struct Visitor<'a, V, T>(V, &'a mut T);
```
jjbayer, line 447: nit: Rename to `TransformingVisitor` to be consistent with `TransformingSerializer`? Or drop the prefix on both.

jjbayer, line 447: nit: Name fields for readability?

jjbayer (approved):

relay-replays/src/transform.rs
jan-auer: Tried it out, but I have to admit that this leads to a bit too much line wrapping for my taste. If it's OK, I would leave it this way.

relay-replays/src/transform.rs
jan-auer: Going to rename the Deserializer instead.

PR #1805, opened 2023-02-01
bug(replays): Add default user object and fill its ip-address

### LINKED ISSUE #43509: Replay: IP address showing `0.0.0.0` without filter enabled
I was getting the server’s IP (I’m tunneling and not forwarding X-Forwarded-For) and now I see 0.0.0.0This is a result of PII stripping added recently, right?
The project doesn’t have any custom filters in Settings. On Security & Privacy tab it doesn’t have Prevent Storing of IP Addresses

image

image

image

### Expected Result

Be able to control IP address stripping

cmanallen:
If the User object is null we currently skip the ip normalization step.  Now (if we have an ip-address) we will add a default User if one does not exist and add the ip-address to that User.  This fixes a bug where a large number of replays were being ingested without an ip-address.

---

Edit: out of date.  Left for historical purposes.

I also observed this functionality:

https://github.com/getsentry/relay/blob/01514478bb1614124c360407b599c525d1371c38/relay-general/src/store/normalize.rs#L593-L650

Obviously my additions are much less than what this function encompasses.  Some of this seems related to back-end SDKs which may be outside of scope.  I did notice lines 617-620 seem to handle the case where the user's ip was submitted as `auto`.  Is this something the JS SDKs do?


cmanallen:
I decided to make the ip-address normalization function generic and implement it for event and replay.

relay-general/src/store/normalize.rs
```diff
@@ -591,6 +591,20 @@ fn normalize_security_report(
...
+        client_ip,
+    )
+}
+
+pub fn normalize_ip_addresses_generic(
+    request: &mut Annotated<Request>,
+    user: &mut Annotated<User>,
+    platform: &Annotated<String>,
+    client_ip: Option<&IpAddr>,
+) {
```
jan-auer: Thinking out loud, instead of a function with a large signature like this, we could decompose it into smaller functions. For instance, I'm noticing that we don't have to pass the `user` in. Instead, this function could return an `Option<IpAddr>` and then the caller runs `get_or_insert_with(|| infer_ip_address(/* ... */))`.

relay-general/src/store/normalize.rs
```diff
@@ -624,22 +638,21 @@ fn normalize_ip_addresses(event: &mut Event, client_ip: Option<&IpAddr>) {
     }
 
     // Copy IPs from request interface to user, and resolve platform-specific backfilling
-    let http_ip = event
-        .request
+    let http_ip = request
```
jan-auer, line 638: Similarly, here we could have two separate functions. One that resolves {{auto}}, and another one that copies IP addresses over.

relay-general/src/store/normalize.rs
cmanallen: @jan-auer I did break out some of the functionality into different units but I kept this function with the large signature.  I didn't want to duplicate implementation details across two modules.  If there's a better way to do this let me know.  I'd also like to expedite this particular pull if possible.  Its preventing ip-address ingest in prod currently.

jjbayer (approved):
@cmanallen I found it hard to judge by the diff whether your refactor had any functional changes that could negatively impact event processing, so I reverted parts of the refactor to make it easier to understand what actually changes. Let me know if I missed something.

PR #1807, opened 2023-02-02
test: Fix flaky test

andriisoldatenko:
This PR fixes flaky test by using the same approach as other similar tests in this file.

#skip-changelog

### DIFF
--- a/tests/integration/test_metrics.py
+++ b/tests/integration/test_metrics.py
@@ -739,7 +739,7 @@ def test_transaction_metrics_count_per_root_project(
 
     event, _ = transactions_consumer.get_event()
 
-    metrics_by_project = metrics_by_name_group_by_project(metrics_consumer, timeout=2)
+    metrics_by_project = metrics_by_name_group_by_project(metrics_consumer)
 
     assert metrics_by_project[41]["c:transactions/count_per_root_project@none"] == {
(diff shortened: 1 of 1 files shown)


tests/integration/test_metrics.py
```diff
@@ -739,7 +739,7 @@ def test_transaction_metrics_count_per_root_project(
 
     event, _ = transactions_consumer.get_event()
 
-    metrics_by_project = metrics_by_name_group_by_project(metrics_consumer, timeout=2)
+    metrics_by_project = metrics_by_name_group_by_project(metrics_consumer)
```
jan-auer, line 742: This halves the timeout from 2 second to 1 second. Can you explain how this improves stability? We usually had to increase the timeout rather than decrease it to avoid flakes.

jjbayer: Agreed. FWIW, I haven't heard any complaints about this test flaking in the past two days, so we can also just close this PR and only change something when the test starts flaking again.

andriisoldatenko: Agreed let's close it.

jan-auer:
> Agreed let's close it.

Thanks for the PR, @andriisoldatenko! We'll continue to monitor this test.

Issue #1712, opened 2022-12-20
Document custom measurement limit


jjbayer:
It seems that the custom measurement limit of 5 is not documented (see https://docs.sentry.io/product/performance/metrics/#custom-performance-metrics). Also, from reading the code I think we don’t create any annotation within the event payload that would pop up in our UI to tell the user why they’re seeing fewer elements than expected, which would be even better than documentation.

```[tasklist]
- [ ] https://github.com/getsentry/sentry-docs/pull/6216
- [ ] https://github.com/getsentry/relay/pull/1815
```

jjbayer:
Relevant code: https://github.com/getsentry/relay/blob/4cb923f0667da94801039ff2cbd99635221dd97b/relay-general/src/store/normalize.rs#L345

jan-auer:
Increasing priority to P0 as per request from @ale-cota.

PR #1815, opened 2023-02-03
feat(normalize): Keep meta for removed custom measurements

jan-auer:
This adds a `_meta` entry for custom measurements that have been removed during
normalization as a result of the custom measurements limit. The entry contains
a generic error message along with the payloads of all removed measurements.

See #1712


olksdr (approved):
:shipit:

PR #1819, opened 2023-02-06
ref(glob): Make the regex compilation for glob lazy

olksdr:
Currently we deserialize all the incoming glob patterns eagerly, converting the pattern to `Regex`.

This change make it so, that we only prepare the string pattern for the incoming globs, and compile the regex once we have to use it, consuming the prepared pattern.

This should reduce the deserialization time for the transaction name rules and also only the used rules will have compiled regexes in them and the rest will just wait it's turn. 

fix: https://github.com/getsentry/team-ingest/issues/33

#skip-changelog

### DIFF
--- a/relay-common/src/utils.rs
+++ b/relay-common/src/utils.rs
@@ -61,6 +61,9 @@ impl<'g> GlobBuilder<'g> {
     }
 
     /// Create a new [`Glob`] from this builder.
+    ///
+    /// This build the string pattern, which will be later lazily compiled to regex when used first
+    /// time.
     pub fn build(self) -> Glob {
         let mut pattern = String::with_capacity(&self.value.len() + 100);
         let mut last = 0;
@@ -85,7 +88,8 @@ impl<'g> GlobBuilder<'g> {
 
         Glob {
             value: self.value.to_owned(),
-            pattern: Regex::new(&pattern).unwrap(),
+            re_pattern: pattern,
+            glob: OnceCell::new(),
         }
     }
 }
@@ -97,7 +101,8 @@ impl<'g> GlobBuilder<'g> {
 #[derive(Clone)]
 pub struct Glob {
     value: String,
-    pattern: Regex,
+    re_pattern: String,
+    glob: OnceCell<Regex>,
 }
 
 impl Glob {
@@ -119,9 +124,21 @@ impl Glob {
         &self.value
     }
 
+    /// Returns the compiled regex based on the prepared pattern.
+    ///
+    /// This will return `None` if the pattern is invalid and it cannot be compiled to regex.
+    fn compiled(&self) -> Option<&Regex> {
+        self.glob
+            .get_or_try_init(|| Regex::new(&self.re_pattern))
+            .ok()
+    }
+
     /// Checks if some value matches the glob.
     pub fn is_match(&self, value: &str) -> bool {
-        self.pattern.is_match(value)
+        if let Some(glob) = self.compiled() {
+            return glob.is_match(value);
+        }
+        false
     }
 
     /// Currently support replacing only all `*` in the input string with provided replacement.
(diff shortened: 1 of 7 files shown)


relay-common/src/utils.rs
```diff
@@ -1,11 +1,49 @@
...
 
 use crate::macros::impl_str_serde;
 
+/// Wraps either the provided pattern string, which will be consumed and replaced with compiled
+/// regex, when the `compiled` function is called.
+#[derive(Debug, Clone)]
+struct GlobRegex {
+    // Note: Mutex and Arc here used to make sure this struct can be Send and Sync. Also that we
+    // could use interior mutabilty in order to be able to consume the value when needed.
+    value: Arc<Mutex<String>>,
```
jan-auer: This doesn't have to be behind a mutex. In order to construct a `Regex`, you can simply borrow the existing string.

relay-common/src/utils.rs
```diff
@@ -1,11 +1,49 @@
...
 
 use once_cell::sync::OnceCell;
 use regex::Regex;
 
 use crate::macros::impl_str_serde;
 
+/// Wraps either the provided pattern string, which will be consumed and replaced with compiled
+/// regex, when the `compiled` function is called.
+#[derive(Debug, Clone)]
+struct GlobRegex {
```
jan-auer: I'm realizing that we also have the `Pattern` type in Relay that's used in the PiiConfig that deserializes eagerly:

https://github.com/getsentry/relay/blob/35d2c744a141acc917da8904bec5c5eb4f5bf4cb/relay-general/src/pii/config.rs#L20-L59

Do you see a straight-forward way of covering both in this PR?

relay-common/src/utils.rs
```diff
@@ -1,11 +1,49 @@
...
+    /// Creates a new `GlobRegex` from the string pattern.
+    fn new(pattern: String) -> Self {
+        Self {
+            value: Arc::new(Mutex::new(pattern)),
+            re: OnceCell::new(),
+        }
+    }
+
+    /// Returns the either `Some` of the compiled regex, or `None` if there is an error while the
+    /// regex was created.
```
iker-barriocanal:
```suggestion
    /// Returns either `Some` of the compiled regex, or `None` if there is an error while the
    /// regex was created.
```

jan-auer (changes requested):
On the placement of `GlobRegex`: Given this is a deserialize helper to lazily parse globs into regular expressions, we should probably not put this in `relay-common` but rather into some utility closer to `ProjectConfig` (since this is what it's for). The common crate is also pulled into the CABI, which doesn't need this.

relay-common/src/utils.rs
```diff
@@ -1,11 +1,49 @@
...
+    ///
+    /// This will compile regex once and will reuse it later on consecutive calls.
```
iker-barriocanal:
```suggestion
    /// This will compile the regex once and will reuse it later on consecutive calls.
```

relay-common/src/utils.rs
iker-barriocanal: I don't really like this name. I don't have good suggestions that I like, maybe something like `LazyPattern`?

relay-common/src/utils.rs
```diff
@@ -61,6 +99,9 @@ impl<'g> GlobBuilder<'g> {
     }
 
     /// Create a new [`Glob`] from this builder.
+    ///
+    /// This build the string pattern, which will be later lazily compiled to regex when used first
+    /// time.
```
iker-barriocanal:
```suggestion
    /// This builds the string pattern, which will be later lazily compiled to regex when used for 
    /// the first time.
```

relay-common/src/utils.rs
```diff
@@ -1,11 +1,49 @@
...
+    /// This will compile regex once and will reuse it later on consecutive calls.
+    ///
+    /// By compiling the regex we consume value pattern, replacing it with empty string buffer,
+    /// which is very inexpensive operation.
+    fn compiled(&self) -> Option<&Regex> {
+        self.re
+            .get_or_try_init(|| {
+                let pattern = mem::take(&mut *self.value.lock().unwrap());
+                Regex::new(&pattern)
+            })
```
iker-barriocanal: How does `Regex::new` behave if the pattern is invalid? Does it panic (and the panic is propagated by `get_or_try_init`)? Do we get any information (e.g. Sentry issue) if relay fails to build the regex from a project config?

iker-barriocanal (commented):
Please add this change to the changelog.

relay-common/src/utils.rs
jan-auer: [`Regex::new`](https://docs.rs/regex/latest/regex/struct.Regex.html#method.new) returns a `Result`.

relay-common/src/utils.rs
olksdr: Yeah, it's me wanted this to be with interior mutability and actually consume the string when the regex is compiled. But I guess it's ok to keep it around as well.

relay-common/src/utils.rs
olksdr: I will look into this.

relay-common/src/utils.rs
olksdr: Not relevelavant anymore

relay-common/src/utils.rs
olksdr: Not relevant anymore

relay-common/src/utils.rs
olksdr: Not relevant anymore

relay-common/src/utils.rs
olksdr: Not relevant anymore

relay-general/src/store/transactions/rules.rs
```diff
@@ -1,28 +1,58 @@
...
-        .capture_double_star(false)
-        .capture_question_mark(false)
-        .build();
-    Ok(glob)
-}
-
-/// Default value for substitution in [`RedactionRule`].
-fn default_substitution() -> String {
-    "*".to_string()
+    Ok(LazyGlob::new(pattern))
```
jan-auer:
```suggestion
    String::deserialize(deserializer).map(LazyGlob::new)
```

jan-auer (approved):
Very elegant!

PR #1820, opened 2023-02-06
fix(time): Avoid panicking when provided timestamp is wrong

### LINKED ISSUE #1624: Relay panics on invalid or out-of-range datetime
See [Sentry issue](https://sentry.my.sentry.io/organizations/sentry/issues/315685/?project=9&query=is%3Aunresolved+is%3Afor_review+assigned_or_suggested%3A%5Bme%2C+none%5D&referrer=issue-stream&sort=inbox). The error happens specifically here:
https://github.com/getsentry/relay/blob/2a0a6b569fdc6a9f8c43cfe36dad74b3deebc0fe/relay-common/src/time.rs#L133

olksdr:
Switch to `NaiveDateTime::from_timestamp_opt` to avoid panicking when the unix timestamp is somehow wrong and cannot be parsed to get the `NaiveDateTime` out of it. See [`chrono source code`](https://github.com/chronotope/chrono/blob/6684a470abcd037aeffcde7177128c40e9740f26/src/naive/datetime/mod.rs#L150)

fix: #1624 

#skip-changelog

### DIFF
--- a/relay-server/src/actors/outcome_aggregator.rs
+++ b/relay-server/src/actors/outcome_aggregator.rs
@@ -139,19 +139,22 @@ impl OutcomeAggregator {
                 category,
             } = bucket_key;
 
-            let timestamp = UnixTimestamp::from_secs(offset * bucket_interval);
-            let outcome = TrackOutcome {
-                timestamp: timestamp.as_datetime(),
-                scoping,
-                outcome,
-                event_id: None,
-                remote_addr,
-                category,
-                quantity,
-            };
+            if let Some(timestamp) =
+                UnixTimestamp::from_secs(offset * bucket_interval).as_datetime()
+            {
+                let outcome = TrackOutcome {
+                    timestamp,
+                    scoping,
+                    outcome,
+                    event_id: None,
+                    remote_addr,
+                    category,
+                    quantity,
+                };
 
-            relay_log::trace!("Flushing outcome for timestamp {}", timestamp);
-            outcome_producer.send(outcome);
+                relay_log::trace!("Flushing outcome for timestamp {}", timestamp);
+                outcome_producer.send(outcome);
+            }
         }
     }
 
--- a/relay-server/src/actors/outcome_aggregator.rs
+++ b/relay-server/src/actors/outcome_aggregator.rs
@@ -5,6 +5,8 @@ use std::collections::HashMap;
 use std::net::IpAddr;
 use std::time::Duration;
 
+use chrono::Utc;
+
 use relay_common::{DataCategory, UnixTimestamp};
 use relay_config::{Config, EmitOutcomes};
 use relay_quotas::Scoping;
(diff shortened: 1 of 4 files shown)


relay-server/src/actors/outcome_aggregator.rs
```diff
@@ -139,19 +139,22 @@ impl OutcomeAggregator {
...
-                event_id: None,
-                remote_addr,
-                category,
-                quantity,
-            };
-
-            relay_log::trace!("Flushing outcome for timestamp {}", timestamp);
-            outcome_producer.send(outcome);
+            if let Some(timestamp) =
+                UnixTimestamp::from_secs(offset * bucket_interval).as_datetime()
```
jan-auer: _nit_: Should we fall back to some timestamp, like `::now()`? Otherwise, we're dropping an outcome which sounds worse than reporting it at the wrong time. 

Applies to all changes in this PR.

relay-server/src/actors/processor.rs
```diff
@@ -956,21 +959,31 @@ impl EnvelopeProcessorService {
         }
 
         let max_age = SignedDuration::seconds(self.config.max_secs_in_past());
-        if (received - timestamp.as_datetime()) > max_age {
+        if (received
+            - timestamp
+                .as_datetime()
+                .ok_or(ProcessingError::InvalidTimestamp)?)
```
jan-auer: _nit_: For readability, assign this to a variable an reuse.

relay-server/src/actors/processor.rs
```diff
@@ -871,7 +871,10 @@ impl EnvelopeProcessorService {
     /// At the moment client reports are primarily used to transfer outcomes from
     /// client SDKs.  The outcomes are removed here and sent directly to the outcomes
     /// system.
-    fn process_client_reports(&self, state: &mut ProcessEnvelopeState) {
+    fn process_client_reports(
+        &self,
+        state: &mut ProcessEnvelopeState,
+    ) -> Result<(), ProcessingError> {
```
jan-auer: This would fail the entire envelope if there's a single broken timestamp in one of the client reports. We should rather fail these individually, or normalize them (see my above comment).

relay-server/src/actors/outcome_aggregator.rs
olksdr: If we panic, we do not report outcome anyway. But you're right, that's what I want to also ask, if it's ok to fall back to `now` in such cases. And you answered my question 😄  , will change this.

relay-server/src/actors/processor.rs
```diff
@@ -988,7 +998,10 @@ impl EnvelopeProcessorService {
             };
 
             producer.send(TrackOutcome {
-                timestamp: timestamp.as_datetime(),
+                // If we get to this point, the unwrap should not be used anymore, since we know by
+                // now that the timestamp can be parsed, but just incase we fallback to UTC current
+                // `DateTime`.
+                timestamp: timestamp.as_datetime().unwrap_or_else(Utc::now),
```
jan-auer, line 1004: 👍 for using an unwrap with default here along with a descriptive comment. Even if the unwrap cannot fail, this line will never panic even if the rest is refactored.

jan-auer (approved):

PR #1821, opened 2023-02-06
feat(protocol): Ignore empty client hints

TBS1996:
if a client hint has an empty string, it should be treated as missing. 

#skip-changelog

### DIFF
--- a/relay-general/src/protocol/contexts/browser.rs
+++ b/relay-general/src/protocol/contexts/browser.rs
@@ -86,6 +86,10 @@ fn browser_from_client_hints(s: &str) -> Option<(String, String)> {
         let browser = captures.get(1)?.as_str().to_owned();
         let version = captures.get(2)?.as_str().to_owned();
 
+        if browser.is_empty() || version.is_empty() {
+            return None;
+        }
+
         return Some((browser, version));
     }
     None
--- a/relay-general/src/protocol/contexts/browser.rs
+++ b/relay-general/src/protocol/contexts/browser.rs
@@ -86,7 +86,7 @@ fn browser_from_client_hints(s: &str) -> Option<(String, String)> {
         let browser = captures.get(1)?.as_str().to_owned();
         let version = captures.get(2)?.as_str().to_owned();
 
-        if browser.is_empty() || version.is_empty() {
+        if browser.trim().is_empty() || version.trim().is_empty() {
             return None;
         }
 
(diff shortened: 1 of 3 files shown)


olksdr (commented):
Could you, please, also add unit tests to check this behaviour, if there are none?

TBS1996:
> Could you, please, also add unit tests to check this behaviour, if there are none?

good call, newest commit has unit tests

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -152,6 +156,27 @@ mod tests {
...
+                        .into(),
+                ),
+            ))];
+            PairList(headers)
+        });
+
+        let client_hints = RawUserAgentInfo::from_headers(&headers).client_hints;
+
+        let from_hints = BrowserContext::from_client_hints(&client_hints);
+        assert!(from_hints.is_none())
```
iker-barriocanal: nit: I'd remove these lines, and same in the files below
```suggestion
        let client_hints = RawUserAgentInfo::from_headers(&headers).client_hints;
        let from_hints = BrowserContext::from_client_hints(&client_hints);
        assert!(from_hints.is_none())
```

iker-barriocanal (approved):
Please add this PR to the changelog.

relay-general/src/protocol/contexts/browser.rs
```diff
@@ -152,6 +156,27 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_ignore_empty_browser() {
+        let headers = Headers({
+            let headers = vec![Annotated::new((
+                Annotated::new("SEC-CH-UA".to_string().into()),
+                Annotated::new(
+                    // browser field missing
+                    r#"Not_A Brand";v="99", "";v="109", "Chromium";v="109"#
```
olksdr: What if the string will be 
```suggestion
                    r#"Not_A Brand";v="99", " ";v="109", "Chromium";v="109"#
```

in this case the tests won't work even though it is kind of empty. 

What you will have to do is to trim the string, when you check  for emptiness

relay-general/src/protocol/contexts/device.rs
```diff
@@ -175,6 +175,11 @@ impl DeviceContext {
 impl FromUserAgentInfo for DeviceContext {
     fn from_client_hints(client_hints: &ClientHints) -> Option<Self> {
         let device = client_hints.sec_ch_ua_model?.to_owned();
+
+        if device.is_empty() {
```
olksdr: same here, and in the following test, you will have to `trim` the string here.

olksdr (commented):
This look great! Thanks for adding the tests. 

One more thing is, please, trim the strings when you check if those are empty and then we can merge it.

relay-general/src/protocol/contexts/browser.rs
TBS1996: ok, I have now added trim to all the checks for emptiness

olksdr (approved):

PR #1823, opened 2023-02-07
feat(replays): Add replays scrubbing diagnostic script

JoshFerge:
Adds a tool to test scrubbing implementations. not intending to merge as is, but want to show an example of how to use it. see README.md for instructions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3502,6 +3502,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "relay-scr"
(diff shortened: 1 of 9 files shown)


jjbayer:
@JoshFerge there's already a `scrub-minidump` tool, would it make sense to align the one you're adding with that one?

JoshFerge:
@jjbayer perhaps! though we're not focusing on this area of replays right now, so going to go ahead and close for now.

PR #1824, opened 2023-02-07
doc(replays): Add basic docs for Replays

jan-auer:
Contains bare minimum documentation for the `relay-replays` module. More documentation on the payload and processing functionality, similar to `relay-metrics` and `relay-profiling` should be added in a followup.

#skip-changelog

### DIFF
--- a/relay-replays/src/lib.rs
+++ b/relay-replays/src/lib.rs
@@ -1,2 +1,17 @@
+//! Session replay protocol and processing for Sentry.
+//!
+//! [Session Replay] provides a video-like reproduction of user interactions on a site or web app.
+//! All user interactions — including page visits, mouse movements, clicks, and scrolls — are
+//! captured, helping developers connect the dots between a known issue and how a user experienced
+//! it in the UI.
+//!
+//! [session replay]: https://docs.sentry.io/product/session-replay/
+
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/getsentry/relay/master/artwork/relay-icon.png",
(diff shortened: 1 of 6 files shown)


relay-replays/src/recording.rs
```diff
@@ -109,9 +145,24 @@ impl<'a> ReplayScrubber<'a> {
         }
     }
 
-    /// Parses compressed replay recording payloads and applies data scrubbers.
+    /// Parses a replay recording payloads and applies data scrubbers.
```
```suggestion
    /// Parses replay recording payloads and applies data scrubbers.
```

olksdr (approved):
:shipit:

relay-replays/src/lib.rs
```diff
@@ -1,2 +1,17 @@
+//! Session replay protocol and processing for Sentry.
+//!
+//! [Session Replay] provides a video-like reproduction of user interactions on a site or web app.
+//! All user interactions — including page visits, mouse movements, clicks, and scrolls — are
+//! captured, helping developers connect the dots between a known issue and how a user experienced
+//! it in the UI.
+//!
+//! [session replay]: https://docs.sentry.io/product/session-replay/
```
iker-barriocanal, line 8: nit, as the case is in the linked docs
```suggestion
//! [Session Replay]: https://docs.sentry.io/product/session-replay/
```

iker-barriocanal (approved):

relay-replays/src/lib.rs
jan-auer: That's intentional, my markdown utilities do that automatically. Links are case insensitive, which is a useful feature to support linking to the same document if you use the same keyword multiple times with different casing (e.g. at the beginning and in the middle of sentences).

Great catch, though :)

PR #1825, opened 2023-02-07
ref(replays): Restrict data scrubbing to sentry payloads

### LINKED ISSUE #1806: PII scrubbing breaks a replay
the Replay Recording payload has the entirety of its `_cssText` scrubbed on two of the three events below. The resulting output is also linked.

This is not a bug in PII scrubbing behavior, but session replays needs to use it correctly. we cannot remove the entirety of strings with certain words, as below you'll see one of the css selectors has the word "password".

I propose the following:
- We tweak the default PII scrubbing rules to not redact entire strings, but to do replacements (I think there are already mechanisms to do this)
- We need observability on how many of our replays are being scrubbed, and we also need to understand how these rules can break replays.
- When we re-introduce schemas, this may be less of a problem as we can be mroe discerning about what nodes we scrub

[css.txt](https://github.com/getsentry/relay/files/10563980/css.txt)
[output.txt](https://github.com/getsentry/relay/files/10564035/output.txt)

Closes https://github.com/getsentry/sentry-javascript/issues/7041

jan-auer:
Unconditional data scrubbing can can break recording payloads because it scrubs strings that should be excluded from scrubbing. Eventually, the data scrubber will have to recurse through rrweb's DOM nodes and skip safe sub-trees or fields from scrubbing.

However, the client's default is to redact all strings from the DOM, which greatly reduces the chance of sensitive data ending up in the recording payloads. This leaves just Sentry's own payloads to be scrubbed, which are always top-level events marked with `type: 5`.

This PR updates the scrubber and introduces a visitor that streams through the top-level event list. When it encounters an item with type `5`, it uses the existing transforming deserializer to scrub all strings in the event, otherwise skipping it. This solution allows us to continue scrubbing without a strict schema.

# Alternatives

During implementation, there were two main alternatives considered:

1. Implement a struct for the top-level events that contains the `type`, `timestamp`, and opaque `data`. Inspect the type, conditionally map it to another struct that wraps `data` in `ScrubbedValue`, and then serialize that struct.
2. (implemented) Deserialize the top-level events into `RawValue`, parse it once with a helper struct to obtain the type. If the type matches, send the entire value through the transformer.

The second approach ultimately ended up in less code with equal to better performance, even though it requires to parse the raw value twice.

# Safety

The `raw_value` feature in the `serde_json` deserializer is now enabled for all of Relay, since it is not possible to load the same crate another time with a separate set of features. There are no documented caveats on the `raw_value` feature in `serde_json`. The implementation uses a special token (`"$serde_json::private::RawValue"`) in a way that makes it is safe for both serialize and deserialize:

- **Serialization**: Goes through `serialize_struct` with the token as name. Triggering this behavior requires a manual implementation of `Serialize` using this token, which our code does not contain.
- **Deserialization**: Goes through `deserialize_newtype_struct` with the token as name. Again, this requires a manual implementation of `Deserialize` to trigger accidentally. 

Closes #1806

### DIFF
--- a/relay-replays/src/recording.rs
+++ b/relay-replays/src/recording.rs
@@ -6,6 +6,10 @@
 //! are likely to contain sensitive data. This module provides [`RecordingScrubber`], which applies
 //! data scrubbing on the payload of recordings while leaving their structure and required fields
 //! intact.
+//!
+//! Data scrubbing applies to only Sentry event payloads within the recording event stream,
+//! identified by `type: 5`. The scrubber skips all other node types and does not perform any
+//! validation beyond JSON parsing.
 
 use std::borrow::Cow;
 use std::cell::RefCell;
@@ -73,7 +77,7 @@ static STRING_STATE: Lazy<ProcessingState> = Lazy::new(|| {
 
 /// The [`Transform`] implementation for data scrubbing.
 ///
-/// This is used by [`EventStreamVisitor`] to scrub recording events.
+/// This is used by [`EventStreamVisitor`] and [`ScrubbedValue`] to scrub recording events.
 struct ScrubberTransform<'a> {
     processor1: Option<PiiProcessor<'a>>,
     processor2: Option<PiiProcessor<'a>>,
@@ -132,6 +136,12 @@ struct EventStreamVisitor<'a, S> {
 }
 
 impl<'a, S> EventStreamVisitor<'a, S> {
+    /// The proprietary rrweb node type that identifies Sentry payloads.
+    ///
+    /// The visitor exclusively scrubs these nodes.
+    const SENTRY_EVENT_TYPE: u8 = 5;
+
+    /// Creates a new visitor wrapping a `serializer`.
     fn new(serializer: S, scrubber: Rc<RefCell<ScrubberTransform<'a>>>) -> Self {
         Self {
             serializer,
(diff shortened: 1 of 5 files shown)


relay-replays/src/recording.rs
```diff
@@ -65,6 +71,121 @@ static STRING_STATE: Lazy<ProcessingState> = Lazy::new(|| {
...
+            ty: u8,
+        }
+
+        use serde::ser::SerializeSeq;
+        let mut seq = self.serializer.serialize_seq(v.size_hint()).map_err(s2d)?;
+
+        while let Some(raw) = v.next_element::<&'de RawValue>()? {
+            // filter for sentry-specific events
+            let helper = serde_json::from_str::<TypeHelper>(raw.get()).unwrap();
+            if helper.ty == 5 {
```
olksdr: Maybe the const with description what the type is this would be more descriptive here

relay-replays/src/recording.rs
```diff
@@ -140,12 +264,17 @@ impl<'a> RecordingScrubber<'a> {
...
+            decoder
+                .read_to_end(&mut decompressed)
+                .map_err(ParseRecordingError::Compression)?;
+
+            self.scrub_replay(&decompressed, encoder)
         }
     }
 
-    /// Parses replay recording payloads and applies data scrubbers.
+    /// Parses a replay recording payloads and applies data scrubbers.
```
```suggestion
    /// Parses a replay recording payload and applies data scrubbers.
```

olksdr (approved):
:shipit:

jan-auer:
Blocked by CI on: 
- https://github.com/getsentry/sentry/pull/44287 
- https://github.com/getsentry/sentry/pull/44289

PR #1829, opened 2023-02-09
ref: Make the pattern regex compilation lazy

olksdr:
As a followup for #1819 (based on the comment https://github.com/getsentry/relay/pull/1819#discussion_r1097515998) this PR converts the `Pattern` to `LazyPattern` and delays the `Regex` compilation to the first usage. But we do keep the incoming raw string pattern around.

If the regex cannot be compiled, we avoid creating the redact rules, which keeps the similar behaviour, since before this change we just error out on the deserialization. 

#skip-changelog

### DIFF
--- a/relay-general/src/pii/config.rs
+++ b/relay-general/src/pii/config.rs
@@ -1,10 +1,8 @@
 use std::collections::{BTreeMap, BTreeSet};
-use std::fmt;
-use std::ops::Deref;
 
 use once_cell::sync::OnceCell;
 use regex::{Regex, RegexBuilder};
-use serde::{de::Error, Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::pii::{CompiledPiiConfig, Redaction};
 use crate::processor::SelectorSpec;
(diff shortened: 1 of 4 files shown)


relay-general/src/pii/config.rs
```diff
@@ -17,51 +15,62 @@ pub enum PiiConfigError {
...
-
-impl Deref for Pattern {
-    type Target = Regex;
+/// Wrapper for the regex and the raw pattern string.
+///
+/// The regex will be compiled only when it used once, and the compiled version will be reused on
+/// consecutive calls.
+#[derive(Debug, Clone)]
+pub struct LazyPattern {
+    raw: String,
```
jan-auer: _nit_: This could be a `Cow<'static, str>` for the `From<&'static str>` case.

The `new` constructor could then have a more flexible signature:

```rust
pub fn new<S>(raw: S, ...) -> Self
where
    Cow<'static, str>: From<S>;  // or S: Into<Cow<'static, str>>
```

relay-general/src/pii/config.rs
```diff
@@ -17,51 +15,62 @@ pub enum PiiConfigError {
...
+        self.raw == other.raw
     }
 }
 
-impl From<&'static str> for Pattern {
-    fn from(pattern: &'static str) -> Pattern {
-        Pattern(Regex::new(pattern).unwrap())
+impl LazyPattern {
+    /// Create a new [`LazyPattern`] from a raw string.
+    pub fn new(raw: String, case_insensitive: bool) -> Self {
```
jan-auer: I haven't traced it down, so please disregard if this is not true: It seems like there was a default for `case_insensitive`, and that default was overwritten in just a few exceptions. The `Deserialize` function particularly needs to make an assumption on case sensitivity. 

To make this more clear, you could offer a `new()` function here that does not include `case_sensitive` and then have a builder method:

```rust
pub fn case_insensitive(mut self, case_insensitive: bool) -> Self {
    self.case_insensitive = case_insensitive;
    self
}
```

relay-general/src/pii/config.rs
```diff
@@ -17,51 +15,62 @@ pub enum PiiConfigError {
...
+    raw: String,
+    case_insensitive: bool,
+    pattern: OnceCell<Regex>,
+}
 
-    fn deref(&self) -> &Self::Target {
-        &self.0
+impl PartialEq for LazyPattern {
+    fn eq(&self, other: &Self) -> bool {
+        self.raw == other.raw
```
jan-auer: _nit_: Are two patterns identical if their casing differs?

relay-general/src/pii/config.rs
```diff
@@ -17,51 +15,62 @@ pub enum PiiConfigError {
...
+            raw,
+            case_insensitive,
+            pattern: OnceCell::new(),
+        }
+    }
+
+    /// Compiles the regex from the internal raw string.
+    pub fn compiled(&self) -> Result<&Regex, PiiConfigError> {
+        self.pattern
+            .get_or_try_init(|| {
```
jan-auer: Judging by the signature of `get_or_try_init`, this will not cache the error. This means if there's a broken regex, we will parse it over and over again. 

I would suggest to store the full `Result<Regex, regex::Error>` in the once_cell, and return references directly from this function: `-> Result<&Regex, &regex::Error>`. The caller can then use them in whichever way they need.

relay-general/src/pii/regexes.rs
```diff
@@ -52,11 +53,13 @@ pub fn get_regex_for_rule_type(
...
     match ty {
-        RuleType::RedactPair(ref redact_pair) => smallvec![(
-            kv,
-            &redact_pair.key_pattern.0,
-            ReplaceBehavior::replace_value()
-        )],
+        RuleType::RedactPair(ref redact_pair) => match redact_pair.key_pattern.compiled() {
+            Ok(pattern) => smallvec![(kv, pattern, ReplaceBehavior::replace_value())],
+            Err(err) => {
+                relay_log::error!("Unable to compile regex: {}", LogError(&err));
```
jan-auer: Note that this is called for _every_ field in _every_ event we're processing. This can easily lead to a humongous number of error logs. It would be better to log this error just once at the compilation site.

Patterns can be based on user-input, so it's not clear to me whether we should even log this as error. It's expected that invalid patterns can be part of the project config.

jan-auer (changes requested):

jan-auer (approved):

PR #1832, opened 2023-02-09
feat(protocol): Add `debug_id` to stacktrace protocol

lforst:
Ref: https://github.com/getsentry/team-webplatform-meta/issues/17

Adds a `debug_id` field to the stacktrace protocol which is used to quickly reference a particular sourcemap artifact from a stack frame. It is typed as `DebugId` but will usually just contain a regular UUID.


jan-auer:
The change itself looks good to me. Waiting for @mitsuhiko to confirm that this is the protocol we want to go with.

@lforst please also document this on [Frame Attributes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes) once this has been merged.

lforst:
> @lforst please also document this on [Frame Attributes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes) once this has been merged.

Right, thanks for the reminder: https://github.com/getsentry/develop/pull/834

lforst:
Rest peacefully

PR #1839, opened 2023-02-10
fix(txnames): Avoid replacing url encoded transaction names

olksdr:
Modified regex to avoid replacing url encoded strings, this way if there are something like `%[A-F0-9]` in the string, we will skip replacing it, since we do track only the named captures. 

/cc @jan-auer 

fix: https://github.com/getsentry/team-ingest/issues/56

#skip-changelog


relay-general/src/store/transactions/processor.rs
```diff
@@ -1882,6 +1890,16 @@ mod tests {
...
+    transaction_name_test!(
+        test_transaction_name_normalize_url_encode_1,
+        "/%2Ftest%2Fopen%20and%20help%2F1%0A",
+        "/%2Ftest%2Fopen%20and%20help%2F1%0A"
+    );
+    transaction_name_test!(
+        test_transaction_name_normalize_url_encode_2,
+        "/this/1234/%E2%9C%85/foo/bar/098123908213",
+        "/this/*/%E2%9C%85/foo/bar/*"
+    );
```
jan-auer, line 1920: Could you add a test like this?

```rust
    transaction_name_test!(
        test_transaction_name_normalize_url_encode_3,
        "/foo/hello%20world-4711/",
        "/foo/*/"
    );
    transaction_name_test!(
        test_transaction_name_normalize_url_encode_4,
        "/foo/hello%20world-0xdeadbeef/",
        "/foo/*/"
    );
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -1882,6 +1890,26 @@ mod tests {
...
+    transaction_name_test!(
+        test_transaction_name_normalize_url_encode_3,
+        "/foo/hello%20world-4711/",
+        "/foo/hello%20world-4711/"
+    );
+    transaction_name_test!(
+        test_transaction_name_normalize_url_encode_4,
+        "/foo/hello%20world-0xdeadbeef/",
+        "/foo/hello%20world-0xdeadbeef/"
+    );
```
jjbayer, line 1945: Another edge case that would be nice to test is an encoded slash (`%2F`).

jan-auer: +1

jan-auer (commented):
The change looks good, but there are still tests failing.

relay-general/src/store/transactions/processor.rs
```diff
@@ -266,26 +266,34 @@ fn normalize_transaction_name(transaction: &mut Annotated<String>) -> Processing
...
                 }
             }
         }
 
-        let changed = TRANSACTION_NAME_NORMALIZER_REGEX
-            .replace_all(trans, "*")
-            .to_string();
-        if *trans != changed && changed != "*" {
+        // Sort by the longest match end.
+        caps.sort_by_key(|(m, _)| m.end());
```
jjbayer: nit: Rename `m` to `match` everywhere?

jjbayer (approved):

relay-general/src/store/transactions/processor.rs
olksdr: `match` is actually the reserved word  😄. But I also can change it to `capture`, since it's basically what it is. What do you think?

olksdr: done in [85a517d](https://github.com/getsentry/relay/pull/1839/commits/85a517d138de733b47ad03ea6dd48d1d7aeddbbd)

PR #1840, opened 2023-02-10
fix(redis): Make `tls` feature work only if `impl` feature is enabled

olksdr:
This creates the proper dependencies between the different features. 

#skip-changelog

### DIFF
--- a/relay-redis/Cargo.toml
+++ b/relay-redis/Cargo.toml
@@ -17,8 +17,8 @@ thiserror = "1.0.20"
 
 [features]
 default = []
-tls = ["redis/tls"]
-impl = ["r2d2", "redis"]
+tls = ["redis?/tls"]
+impl = ["dep:r2d2", "dep:redis"]
 
 [dev-dependencies]
 serde_yaml = "0.8.13"


relay-redis/Cargo.toml
```diff
@@ -17,8 +17,8 @@ thiserror = "1.0.20"
 
 [features]
 default = []
-tls = ["redis/tls"]
-impl = ["r2d2", "redis"]
+tls = ["redis?/tls"]
```
jan-auer, line 20: This doesn't need a `dep:`? I wasn't sure about this from reading the docs.

jan-auer (approved):
Note that this doesn't explain why TLS isn't working in the published 23.1.1 image, which was released 2 days after the original PR introducing this.

relay-redis/Cargo.toml
olksdr: yep, it's just requires `?` for the optional feature. 
it also gives nice error 
```
feature `tls` includes `dep:redis?/tls` with both `dep:` and `/`
  To fix this, remove the `dep:` prefix.
```

PR #1842, opened 2023-02-10
ref: Update CLI dependencies

jan-auer:
This updates all dependencies related to command line parsing. Most notably,
this PR migrates from clap v2 to v4, which contains several breaking changes.
Colors and help texts have changed slightly; otherwise CLI behavior should be
largely the same.

Clap v4 comes with a much improved API and will allow us to:
 - Use derives for the entire command line interface
 - Replace manual env var parsing with clap's built-in functionality
 - Validate arguments right in the command line

All these changes can be shipped in a follow-up.

#skip-changelog


jan-auer:
I will still replace all uses of `structopt` in our tools with the same version of `clap` before merging this in.

olksdr (approved):
lgtm

jernejstrasner (approved):

PR #1843, opened 2023-02-13
fix(rate-limits): Rate limit profiles attached to transactions

### LINKED ISSUE #1612: When transactions are rate limited, profiles are sampled at 100%
When an envelope does not contain transactions, we do not attach a “sampling project state” to the envelope context: [relay/dynamic_sampling.rs at 0586fa62eb3ee75697120ea6d89efe51c740f6ad · getsentry/relay](https://github.com/getsentry/relay/blob/0586fa62eb3ee75697120ea6d89efe51c740f6ad/relay-server/src/utils/dynamic_sampling.rs#L138-L145)

But rate limit enforcement may already have removed transactions at that point, so we just sample 100% of profiles in that case: [relay/project.rs at 0586fa62eb3ee75697120ea6d89efe51c740f6ad · getsentry/relay](https://github.com/getsentry/relay/blob/0586fa62eb3ee75697120ea6d89efe51c740f6ad/relay-server/src/actors/project.rs#L886)

Example ([internal link](https://sentry.io/organizations/sentry/stats/?dataCategory=profiles&end=2022-11-16T23%3A59%3A59&project=1&start=2022-11-16T00%3A00%3A00&utc=true)):

image

Reproduce the bug in a (integration) test.

Fix: Either

* drop dependent items on rate limiting: When we remove the transaction, also remove profiles from the envelope.

* Apply dynamic sampling (i.e. attach a sampling project state to the envelope context) even if the transaction has been removed.

jjbayer:
We currently do not remove profiles from an envelope when its transaction is being rate limited. This has the unintended consequence that such an envelope, now containing only profiles, is no longer subject to dynamic sampling, because `get_sampling_key` only returns something when the envelope contains a transaction.

This PR fixes rate limiting & sampling behavior by enforcing event (i.e. transaction) rate limits also on the associated profiles, just like we do for attachments.

Fixes https://github.com/getsentry/relay/issues/1612.

### DIFF
--- a/relay-server/src/utils/rate_limits.rs
+++ b/relay-server/src/utils/rate_limits.rs
@@ -1038,6 +1038,24 @@ mod tests {
     }
 
     #[test]
+    fn test_enforce_transaction_profile_enforced() {
+        let mut envelope = envelope![Transaction, Profile];
+        let config = config_with_tx_metrics();
+
+        let mut mock = MockLimiter::default().deny(DataCategory::Transaction);
+        let limiter = EnvelopeLimiter::new(Some(&config), |s, q| mock.check(s, q));
+
+        let (enforcement, _limits) = limiter.enforce(&mut envelope, &scoping()).unwrap();
+
+        assert!(enforcement.event.is_active());
+        assert!(enforcement.profiles.is_active());
+        mock.assert_call(DataCategory::Transaction, Some(0));
+        mock.assert_call(DataCategory::Profile, None);
+
+        // TODO: Also test track_outcomes
+    }
+
+    #[test]
     fn test_enforce_transaction_attachment_enforced_metrics_extracted_indexing_quota() {
         let mut envelope = envelope![Transaction, Attachment];
         set_extracted(&mut envelope, ItemType::Transaction);
--- a/relay-server/src/utils/rate_limits.rs
+++ b/relay-server/src/utils/rate_limits.rs
@@ -266,10 +266,12 @@ impl Enforcement {
         self.event.is_active()
     }
 
-    /// Invokes [`TrackOutcome`] on all enforcements reported by the [`EnvelopeLimiter`].
-    ///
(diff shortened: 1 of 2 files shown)


relay-server/src/utils/rate_limits.rs
```diff
@@ -1037,6 +1058,22 @@ mod tests {
...
+        let mut mock = MockLimiter::default().deny(DataCategory::Transaction);
+        let limiter = EnvelopeLimiter::new(Some(&config), |s, q| mock.check(s, q));
+
+        let (enforcement, _limits) = limiter.enforce(&mut envelope, &scoping()).unwrap();
+
+        assert!(enforcement.event.is_active());
+        assert!(enforcement.profiles.is_active());
+        mock.assert_call(DataCategory::Transaction, Some(0));
+        mock.assert_call(DataCategory::Profile, None);
+    }
```
jjbayer, line 1108: This test failed before I made the functional change.

olksdr (approved):
lgtm! 

How about adding (or updating) also an integration test for this?  Just to check end-to-end behaviour.

relay-server/src/utils/rate_limits.rs
```diff
@@ -278,22 +280,33 @@ impl Enforcement {
...
+                scoping,
+                outcome: Outcome::RateLimited(limit.reason_code),
+                event_id,
+                remote_addr,
+                category: limit.category,
+                // XXX: on the limiter we have quantity of usize, but in the protocol
+                // and data store we're limited to u32.
+                quantity: limit.quantity as u32,
+            })
+    }
```
jjbayer, line 302: I factored this function out from `track_outcome` to make it unit-testable.

jjbayer:
> How about adding (or updating) also an integration test for this? Just to check end-to-end behaviour.

I would rather rely on unit tests here. We already have some integration tests that verify that rate limiting works end-to-end, and I don't want to bloat CI times. But I did add two additional asserts to verify that outcomes are generated as expected.

PR #1844, opened 2023-02-13
feat(server): Persist attachments to transactions from same envelope

### LINKED ISSUE #1761: Attachments to transactions are not persisted if sent within the same envelope
If an envelope with a transaction event and attachments is sent, the attachments are lost. If the attachment is sent via the dedicated attachments endpoint, everything is fine.

The reason for this is that we don't do any attachments processing in save_transaction_event. The attachments are written in processing store, and only in save_event (error events case) we actually copy them to persistent storage.

The fix is (in my opinion) NOT to introduce attachments processing in the transaction processing pipeline the same way it happens for errors. That is something we do because we need to mutate the event based on data found in attachments (eg stacktrace processing), but it causes lots of small operational problems such as high variance in timing of save_event.

Instead, Relay should treat attachments to transaction events as individual attachments, and write them to kafka that way. Ideally it would do so for all attachments that are not processing-relevant, even for error events.

[transaction-with-attachment.json.txt](https://github.com/getsentry/relay/files/10458746/transaction-with-attachment.json.txt)

TBS1996:
This PR will in case of transaction events, extract any attachments it may have and send them to kafka individually.

In order to make unit-testing possible I had to extract some logic from `fn handle_store_envelope` into `fn extract_kafka_messages` and `fn extract_kafka_attachments`

It should fix the following issue: #1761


relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,137 @@ impl StoreService {
...
+                        metric!(
+                            counter(RelayCounters::ProcessingMessageProduced) += 1,
+                            event_type = "attachment"
+                        );
+                    }
+                    KafkaMessage::Event(_) => metric!(
+                        counter(RelayCounters::ProcessingMessageProduced) += 1,
+                        // the unwrap is safe because in self.extract_kafka_messages() it only
+                        // pushes an event message if the event_item is_some()..
+                        event_type = &(event_item.expect("oh fuck").ty().to_string())
```
TBS1996: This is the guilty unwrap I want to remove. If there's no good way I would just use the if let Some() pattern. 

If we do that, it should always match the Some() variant. What's the best way to notify us if it ever happens that event_item is None here, when it shouldn't be?

relay-server/src/actors/store.rs
```diff
@@ -1049,3 +1139,138 @@ impl Message for KafkaMessage {
...
+    use relay_common::ProjectKey;
+
+    /// Should check these three cases:
+    /// 1. if theres no event_id, it should return the attachments.
+    /// 2. if there is an event_id and its a transaction, then it should strip the attachments from the
+    ///     event and put them in the returned vector.
+    /// 3. If there is an event_id and it's not a transaction, then the attachments should be part
+    ///     of the Event message, not as standalone attachment-messages.
+    #[test]
+    fn test_extract_kafka_messages() {
```
TBS1996: Could split into three tests

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,122 @@ impl StoreService {
...
-
-        Ok(())
     }
 
     fn produce(
         &self,
         topic: KafkaTopic,
         organization_id: u64,
-        message: KafkaMessage,
+        message: &KafkaMessage,
```
TBS1996: this created a lot of diffs. I can start a new PR that only changes the signature to make this one more clear.

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,122 @@ impl StoreService {
...
+        remote_addr: Option<String>,
+        attachments: &mut Vec<ChunkedAttachment>,
+    ) -> Vec<KafkaMessage> {
+        let mut kafkamsgs = vec![];
         if let Some(event_item) = event_item {
-            relay_log::trace!("Sending event item of envelope to kafka");
+            if matches!(event_item.ty(), ItemType::Transaction) {
+                self.extract_kafka_attachments(
+                    &mut kafkamsgs,
+                    attachments.clone(),
```
olksdr: You do need this `clone` here.

The function can own the `attachments`, by changing the signature to
```rust 
....
mut attachments: Vec<ChunkedAttachment>
```

relay-server/src/actors/store.rs
```diff
@@ -744,7 +819,7 @@ impl Service for StoreService {
 }
 
 /// Common attributes for both standalone attachments and processing-relevant attachments.
-#[derive(Debug, Serialize)]
+#[derive(Debug, Serialize, Clone)]
```
olksdr: This `Clone` is not necessary.
 ```suggestion
#[derive(Debug, Serialize)]
```

File: relay-server/src/actors/store.rs [hunk 6]
```diff
@@ -199,51 +199,122 @@ impl StoreService {
...
 
+        let remote_addr = envelope.meta().client_addr().map(|addr| addr.to_string());
+
+        let kafkamessages = self.extract_kafka_messages(
+            event_item,
+            event_id,
+            scoping,
+            start_time,
+            remote_addr,
+            &mut attachments,
```
olksdr: The signature of the function can be changed and then just move the attachments there.
```suggestion
           attachments,
```

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,122 @@ impl StoreService {
...
+                            counter(RelayCounters::ProcessingMessageProduced) += 1,
+                            event_type = "attachment"
+                        );
+                    }
+                    KafkaMessage::Event(_) => metric!(
+                        counter(RelayCounters::ProcessingMessageProduced) += 1,
+                        // The unwrap is safe because in self.extract_kafka_messages() it only
+                        // pushes an event message if the event_item is_some()..
+                        // still I would prefer to not have any unwraps.
+                        event_type = &(event_item.unwrap().ty().to_string())
```
olksdr: `unwrap` still must be removed from here

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,122 @@ impl StoreService {
...
-                event_id: event_id.ok_or(StoreError::NoEventId)?,
+                event_id: match event_id {
+                    Some(event_id) => event_id,
+                    None => return kafkamsgs,
+                },
                 project_id: scoping.project_id,
-                remote_addr: envelope.meta().client_addr().map(|addr| addr.to_string()),
-                attachments,
+                remote_addr,
+                attachments: attachments.clone(),
```
olksdr: This does not have to be cloned.

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,122 @@ impl StoreService {
...
 
-            self.produce(topic, scoping.organization_id, event_message)?;
-            metric!(
-                counter(RelayCounters::ProcessingMessageProduced) += 1,
-                event_type = &event_item.ty().to_string()
+            kafkamsgs.push(event_message);
+        } else {
+            self.extract_kafka_attachments(
+                &mut kafkamsgs,
+                attachments.clone(),
```
olksdr: This does not have to be cloned.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Features**:
+
+- Extract attachments from transaction events and send them to kafka individually. ([#1844](https://github.com/getsentry/relay/pull/1731))
```
```suggestion
- Extract attachments from transaction events and send them to kafka individually. ([#1844](https://github.com/getsentry/relay/pull/1844))
```

relay-server/src/actors/store.rs
jjbayer: Why did the signature change in the first place?

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,117 @@ impl StoreService {
             }
         }
 
+        let remote_addr = envelope.meta().client_addr().map(|addr| addr.to_string());
+
+        let kafkamessages = self.extract_kafka_messages(
```
```suggestion
        let kafka_messages = self.extract_kafka_messages(
```

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,117 @@ impl StoreService {
...
+
+    fn extract_kafka_messages(
+        &self,
+        event_item: Option<&Item>,
+        event_id: Option<EventId>,
+        scoping: Scoping,
+        start_time: Instant,
+        remote_addr: Option<String>,
+        mut attachments: Vec<ChunkedAttachment>,
+    ) -> Vec<KafkaMessage> {
```
jjbayer: There's actually no need for this helper method to store messages in a vector. It could return an iterator instead. For inspiration, see for example https://github.com/getsentry/relay/blob/4f2277c4a69c9fc6134e7cc8dbfe38c7aaf65dc1/relay-general/src/store/normalize/breakdowns.rs#L191-L194.

Use [`.chain()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain) to combine multiple iterators.

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,117 @@ impl StoreService {
...
+        let mut kafkamsgs = vec![];
         if let Some(event_item) = event_item {
-            relay_log::trace!("Sending event item of envelope to kafka");
+            if matches!(event_item.ty(), ItemType::Transaction) {
```
```suggestion
            if matches!(event_item.ty(), ItemType::Transaction) {
                // Transaction attachments are dropped in sentry.
                // In order to keep them, emit them as single attachments instead.
```

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,117 @@ impl StoreService {
...
+            }
+
             let event_message = KafkaMessage::Event(EventKafkaMessage {
                 payload: event_item.payload(),
                 start_time: UnixTimestamp::from_instant(start_time).as_secs(),
-                event_id: event_id.ok_or(StoreError::NoEventId)?,
+                event_id: match event_id {
+                    Some(event_id) => event_id,
+                    None => return kafkamsgs,
+                },
```
jjbayer: Is this correct behavior? Previously, a missing event ID would mean that `handle_store_envelope` returns a specific  `StoreError` I think we need to restore that behavior. In line 206, you could pass `event_id?` instead of `event_id`, and change the signature of `extract_kafka_messages` to `event_id: EventId`.

relay-server/src/actors/store.rs
```diff
@@ -1049,3 +1119,141 @@ impl Message for KafkaMessage {
...
+    ///     event and put them in the returned vector.
+    /// 3. If there is an event_id and it's not a transaction, then the attachments should be part
+    ///     of the Event message, not as standalone attachment-messages.
+    #[test]
+    fn test_extract_kafka_messages() {
+        let config = {
+            let json_config = serde_json::json!({});
+            Config::from_json_value(json_config).unwrap()
+        };
+        let store_service = StoreService::create(Arc::new(config)).unwrap();
```
jjbayer: It looks like `extract_kafka_messages` does not require `&self`. So you could make it an associated function or even a standalone function, then you would not have to create a service here.

relay-server/src/actors/store.rs
jjbayer: Yes 👍

jjbayer (changes requested):

relay-server/src/actors/store.rs
TBS1996: "Why did the signature change in the first place?"

after calling self.produce in the end of handle_store_envelope I want to match on message to know which log entry to make, so then I don't want the value to be consumed. 

I could have some variable keeping track of it before self.produce but that would be more messy, and also I think the signature makes more sense this way as self.produce doesn't need to consume it anyway.

relay-server/src/actors/store.rs
TBS1996: What would be the benefit of returning an iterator instead? Or do you mean that on `fn extract_kafka_attachment` return an iterator instead of mutating the Vec<KafkaMessage>?

relay-server/src/actors/store.rs
TBS1996: fixed now

relay-server/src/actors/store.rs
TBS1996: oh yes, nice observation. I am surprised there's not a lint for this.

I'm still using service_store for `self.produce_attachment_chunks` though, but maybe I should just make it directly anyway.

relay-server/src/actors/store.rs
jjbayer: Iterators are a [zero-cost abstraction](https://doc.rust-lang.org/book/ch13-04-performance.html#:~:text=Iterators%20are%20one%20of%20Rust's,of%20C%2B%2B%E2%80%9D%20(2012)%3A), which means its runtime cost is just as low as having a simple for-loop. Vectors on the other hand require memory allocations, and cause data to being copied around.

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,111 @@ impl StoreService {
...
+            event_item,
+            event_id.ok_or(StoreError::NoEventId)?,
+            scoping,
+            start_time,
+            remote_addr,
+            attachments,
+        );
+
+        for message in kafka_messages {
+            if let Ok(()) = self.produce(topic, scoping.organization_id, &message) {
```
TBS1996: Should we log it if self.produce returns an error?

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,101 @@ impl StoreService {
...
-            let event_message = KafkaMessage::Event(EventKafkaMessage {
-                payload: event_item.payload(),
-                start_time: UnixTimestamp::from_instant(start_time).as_secs(),
-                event_id: event_id.ok_or(StoreError::NoEventId)?,
-                project_id: scoping.project_id,
-                remote_addr: envelope.meta().client_addr().map(|addr| addr.to_string()),
-                attachments,
-            });
+        if event_item.is_none() && attachments.is_empty() {
+            return Ok(());
```
jjbayer, line 204: nit:
```suggestion
            // No event-related content. All done.
            return Ok(());
```

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,101 @@ impl StoreService {
...
+                        }
+                    }
+                }
             }
         }
 
         Ok(())
     }
 
+    fn extract_kafka_messages(
```
jjbayer: To clarify that this is only for envelopes with an event ID:

```suggestion
    fn extract_kafka_messages_for_event(
```

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,101 @@ impl StoreService {
...
+        scoping: Scoping,
+        start_time: Instant,
+        remote_addr: Option<String>,
+        attachments: Vec<ChunkedAttachment>,
+    ) -> impl Iterator<Item = KafkaMessage> {
+        // There might be a better way to do this:
+        let (individual_attachments, inline_attachments) = match event_item {
+            Some(event_item) => {
+                if matches!(event_item.ty(), ItemType::Transaction) {
+                    (attachments, vec![])
```
jjbayer, line 252:
```suggestion
                    // Sentry discards inline attachments for transactions, so send them as individual ones instead.
                    (attachments, vec![])
```

relay-server/src/actors/store.rs
jjbayer: The good thing about consuming it is that it ensures that the message isn't accidentally produced twice, so I strongly prefer the alternative you suggest. You can return the `AttachmentType` from `self.produce` to make it less messy.

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,101 @@ impl StoreService {
...
+                    payload: event_item.payload(),
+                    start_time: UnixTimestamp::from_instant(start_time).as_secs(),
+                    event_id,
+                    project_id,
+                    remote_addr,
+                    attachments: inline_attachments,
+                })
+            })
+            .into_iter();
+
```
jjbayer, line 274: Let's restore these log lines:

```suggestion

if ! individual_attachments.empty() {
    relay_log::trace!("Sending individual attachments of envelope to kafka");
}

```

relay-server/src/actors/store.rs
```diff
@@ -199,51 +199,101 @@ impl StoreService {
...
+                }
+            }
+            None => (attachments, vec![]),
+        };
+
+        let project_id = scoping.project_id;
+
+        let event_iterator = event_item
+            .map(|event_item| {
+                KafkaMessage::Event(EventKafkaMessage {
```
jjbayer, line 264: Let's restore these log lines:

```suggestion
                relay_log::trace!("Sending event item of envelope to kafka");
                KafkaMessage::Event(EventKafkaMessage {
```

relay-server/src/actors/store.rs
jjbayer: Good catch! If `self.produce` fails, the function should early-return with an error, just like it did before:
```suggestion
            self.produce(topic, scoping.organization_id, &message)?;
```

relay-server/src/actors/store.rs
```diff
@@ -1049,3 +1103,162 @@ impl Message for KafkaMessage {
...
     item.ty() == &ItemType::Attachment || item.ty() == &ItemType::UserReport
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use relay_common::ProjectKey;
+
+    /// Helper function to get the arguments for the `fn extract_kafka_messages(...)` method.
+    fn test_arguments_extract_kafka_msgs() -> (Instant, EventId, Scoping, Vec<ChunkedAttachment>) {
```
jjbayer: nit: To me, the `test_` prefix suggests that the function itself is a test.

```suggestion
    fn  arguments_extract_kafka_msgs() -> (Instant, EventId, Scoping, Vec<ChunkedAttachment>) {
```

relay-server/src/actors/store.rs
```diff
@@ -1049,3 +1103,162 @@ impl Message for KafkaMessage {
...
+            None,
+            attachment_vec,
+        );
+
+        assert!(kafka_messages.any(|msg| {
+            if let KafkaMessage::Attachment(_) = msg {
+                return true;
+            }
+            false
+        }));
```
jjbayer: They're all of the same type, right?

```suggestion
        assert!(kafka_messages.all(|msg| matches!(msg, KafkaMessage::Attachment(_)));
```

relay-server/src/actors/store.rs
```diff
@@ -1049,3 +1103,162 @@ impl Message for KafkaMessage {
...
+        let mut kafka_messages = get_kafka_messages();
+
+        // Checks that the attachment is a message on its own in the vector.
+        assert!(kafka_messages.any(|msg| {
+            if let KafkaMessage::Attachment(_) = msg {
+                return true;
+            }
+            false
+        }));
+    }
```
jjbayer: nit: You could use [partition](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition) here to get both types of messages at once.

jjbayer (changes requested):
Almost done, see notes about `produce()?` errors and the discussion about its signature. The rest of the comments are nitpicks.

relay-server/src/actors/store.rs
TBS1996: yeah good point, that way seems more explicit

relay-server/src/actors/store.rs
TBS1996: Right now the event and the individual attachments are being produced in the same for-loop so I can't recreate the log lines as they were, I could put some logs where we record the metrics.

Although the log messages were dealing with the logic in match statement at the start of the `fn extract_kafka_messages_for_event`, then perhaps we should write a separate log for each of the three branches there instead

relay-server/src/actors/store.rs
TBS1996: ok, I changed it now, and I realized we only care if it's attachment or not so it doesn't even look messy now.

relay-server/src/actors/store.rs
TBS1996: i changed it now but I find it a bit weird that we might send half the messages and then abruptly stop if there's an error. Meaning if there is an error with one of the attachments, the other attachments would be produced depending on whether they happen to come before or after the one with the error.

relay-server/src/actors/store.rs
```diff
@@ -199,46 +199,102 @@ impl StoreService {
...
+        start_time: Instant,
+        remote_addr: Option<String>,
+        attachments: Vec<ChunkedAttachment>,
+    ) -> impl Iterator<Item = KafkaMessage> {
+        // There might be a better way to do this:
+        let (individual_attachments, inline_attachments) = match event_item {
+            Some(event_item) => {
+                if matches!(event_item.ty(), ItemType::Transaction) {
+                    // Sentry discards inline attachments for transactions, so send them as individual ones instead.
+                    relay_log::trace!("Sending transaction event item of envelope to kafka");
```
TBS1996: Thoughts on putting log messages here?

jjbayer: How about we make it even simpler and just log a message straight before `self.produce`?

TBS1996: Would we wanna log a message for every single attachment sent then? previously we just logged that we were about to send attachments as a single log message

jjbayer (approved):
Nice! Let's not merge and deploy until Monday.

PR #1845, opened 2023-02-14
build: Use cargo 1.60 optional dependencies

jan-auer:
Uses the syntax introduced in cargo 1.60.0 to enable [optional dependencies](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies) in all crates, for instance `dep:rdkafka`.

#skip-changelog


olksdr (approved):
🚢

PR #1847, opened 2023-02-14
build(log): Upgrade log dependencies

jan-auer:
Updates `env_logger`. For human readable output we have used
`pretty_env_logger`, which is no longer compatible with the latest `env_logger`
version. This PR vendors the relevant bits of the implementation.

Previously, the `relay-log` crate hard-coded a list of all workspace members to
set log levels based on the configuration. This has been refactored:

 - The list of workspace members is now created in a build script.
 - The default level filter sets every workspace member to TRACE.
 - There is a global `max_level` filter based on the overall log level config.

NOTE: Eventually, we will want to move to `tracing` like Symbolicator did. This
PR is merely an effort to keep dependencies up-to-date. The current refactors
should help with moving to `tracing`.

#skip-changelog

### DIFF
--- a/relay-log/build.rs
+++ b/relay-log/build.rs
@@ -1,4 +1,7 @@
-use std::io;
+use std::env;
+use std::fs::{self, File};
+use std::io::{self, Write};
+use std::path::Path;
 use std::process::{Command, Stdio};
 
 fn emit_release_var() -> Result<(), io::Error> {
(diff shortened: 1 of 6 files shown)


relay-log/build.rs
```diff
@@ -21,6 +24,43 @@ fn emit_release_var() -> Result<(), io::Error> {
     Ok(())
 }
 
+fn list_crates() -> Vec<String> {
```
olksdr, line 27: Do we want this `fn` rather return `io::Result` instead of unwrapping everything? Since it's called in `emit_cratt_list` which actually properly returns result.  I think it can be used then to return nicer error messages.

relay-log/build.rs
```diff
@@ -21,6 +24,43 @@ fn emit_release_var() -> Result<(), io::Error> {
...
+            if s.starts_with("relay") {
+                crates.push(s.to_owned());
+            }
+        }
+    }
+
+    crates
+}
+
+fn emit_crate_list() -> Result<(), io::Error> {
```
olksdr, line 47: Also can return `io::Result` here.

olksdr (approved):
LGTM!

Few comments, nothing blocking from me.

relay-log/build.rs
jan-auer: Good point, I was thinking about this. 

The current old build script swallows errors silently, so I chose to expose these errors with a panic. Probably, an error is even better, but we would need additional context for why it panics. I'll follow up about this.

PR #1849, opened 2023-02-14
build: Upgrade sentry and related dependencies

jan-auer:
Upgrades the Sentry SDK, which also required an upgrade of `uuid` to stable 1.0
and `sentry-types` to the latest version across all crates. The switch to stable
UUID caused some refactor due to changes in the public API.

Most notably, the `ProjectId` type is now vendored into Relay instead of being
imported from the SDK. The SDK represents it as a string internally, while Relay
continues to treat it as a numeric value.

#skip-changelog


phacops (approved):

olksdr (approved):
lgtm!

PR #1856, opened 2023-02-16
ref: Buffer envelopes for broken project states

### LINKED ISSUE #1787: Buffer envelopes after resolving broken project states
If the upstream responds with broken states, Relay attempts to reuse old caches as long as possible. However, if there is no cache, Relay immediately drops envelopes. Our stability SLO suggests that we should buffer envelopes instead until they timeout.

To achieve this behavior, we will have to:
- Leave envelopes in the queue when states are invalid
- Expire queued envelopes when a timeout has been reached
- Check whether there should be a timeout on the envelopes or rather on the project state

Created from https://github.com/getsentry/relay/pull/1767#discussion_r1085126760

olksdr:
Check request only for the valid project configs, otherwise we reject incoming events immediately with `invalid data (project_state)`, since there is nothing to check against.

With this change we will still keep buffering even if the resolved project state is invalid. After the state is updated, or pending envelopes will be processed again.

#skip-changelog

fix: #1787


relay-server/src/actors/project.rs
```diff
@@ -942,7 +942,10 @@ impl Project {
         let state = self.valid_state();
         let mut scoping = envelope_context.scoping();
 
-        if let Some(ref state) = state {
+        // Check request only for the valid project configs, otherwise we reject incoming event
+        // immediately with "invalid data (project_state)". Since there is nothing to check
+        // against.
+        if let Some((state, false)) = state.as_ref().map(|state| (state, state.invalid())) {
```
jjbayer: Should we set `state = None` when it's invalid, such that it is treated the same way as a missing state in the code below?

jjbayer (approved):
@jan-auer what is the reason we treat invalid states different than missing states? To me it seems that in both cases, we want to buffer events and retry to fetch the config. So maybe we can get rid of the concept of an `invalid` state altogether (except maybe for logging purposes)?

relay-server/src/actors/project.rs
olksdr: Yeah, I think it makes sense in this case

relay-server/src/actors/project.rs
```diff
@@ -940,9 +940,16 @@ impl Project {
         mut envelope_context: EnvelopeContext,
     ) -> Result<CheckedEnvelope, DiscardReason> {
         let state = self.valid_state();
+        // Treat invalid state as no state at all.
+        // We cannot check anything against invalid state, since it does not contain the required
+        // information, like project id, or quotas, etc.
```
jan-auer: ~If we do this, I think we'll break the v2 endpoint, which relies on an eventual state. Shouldn't we change the places outlined in https://github.com/getsentry/relay/issues/1787#issuecomment-1406473224 instead?~

jan-auer (commented):
Reviewed again together with @jjbayer:
 1. In `update_state`, we never re-trigger fetching of states if there's an invalid state. This is now a problem, because we want to keep envelopes queued. This will require a change to `update_state`. 
 2. In `enqueue_validation`, we should skip states that are invalid
 3. You change in `check_envelope` covers the fast-path. Potentially we might want to catch this in `handle_check_envelope` though?

We explicitly take the risk of overflowing buffers now, which will be addressed through persistent queueing.

relay-server/src/actors/project.rs
```diff
@@ -859,8 +863,8 @@ impl Project {
     /// outdated.
     pub fn enqueue_sampling(&mut self, message: ProcessEnvelope) {
         match self.get_cached_state(message.envelope.meta().no_cache()) {
-            Some(_) => self.flush_sampling(message),
-            None => self.pending_sampling.push_back(message),
+            Some(state) if !state.invalid() => self.flush_sampling(message),
```
jan-auer: With our reliance on dynamic sampling this makes sense, but then please also change the signature:
- Pass `state` as parameter to `flush_sampling`
- Remove the "Ignore all errors" comment and state access within.

relay-server/src/actors/project.rs
```diff
@@ -901,19 +905,23 @@ impl Project {
             _ => self.state = Some(state.clone()),
         }
 
-        // Flush all queued `ValidateEnvelope` messages
-        while let Some((envelope, context)) = self.pending_validations.pop_front() {
-            self.flush_validation(envelope, context, state.clone());
-        }
+        if state.invalid() {
+            ProjectCache::from_registry().send(RequestUpdate::new(self.project_key, no_cache));
```
jan-auer, line 774: Either place the channel back into `self.state_channel`, or pop it in the `else` branch.

relay-server/src/actors/project.rs
```diff
@@ -901,19 +905,23 @@ impl Project {
...
+        } else {
```
jan-auer: _nit_: Early return rather than else?

relay-server/src/actors/project.rs
```diff
@@ -939,7 +947,10 @@ impl Project {
         mut envelope: Box<Envelope>,
         mut envelope_context: EnvelopeContext,
     ) -> Result<CheckedEnvelope, DiscardReason> {
-        let state = self.valid_state();
+        // Treat invalid state as no state at all.
```
jan-auer: _nit_: Create a doc-comment for `check_envelope` clearly mentioning: If you call this on an invalid project state, this will skip certain checks. This is super important, because it means you **must not** call this in the main processing path.

jan-auer (approved):

relay-server/src/actors/project.rs
```diff
@@ -684,18 +693,18 @@ impl Project {
...
+            _ => self.pending_validations.push_back((envelope, context)),
         }
     }
 
     /// Adds the project state for dynamic sampling and submits the Envelope for processing.
-    fn flush_sampling(&self, mut message: ProcessEnvelope) {
+    fn flush_sampling(&self, mut message: ProcessEnvelope, project_state: Arc<ProjectState>) {
         // Intentionally ignore all errors. Fallback sampling behavior applies in this case.
-        if let Some(state) = self.valid_state() {
+        if self.valid_state().is_some() {
```
jan-auer: ~This check is redundant now, since `get_cached_state` already checks for expiry. We could document on `flush_sampling` that this should not be called with an expired state.~

Thinking through this again, I suggested a wrong change earlier. As the comment above states, we want to skip invalid and missing project states for this assignment. This is important because the public key from the DSC can point to a project that does not exist or is unavailable in the same Sentry instance. In this case, fallback sampling behavior must apply.

jan-auer (approved):

relay-server/src/actors/project.rs
```diff
@@ -520,6 +525,13 @@ impl Project {
         }
     }
 
+    /// Retunrs `true` if backoff expired and new attempt can be triggered.
```
olksdr:
```suggestion
    /// Returns `true` if backoff expired and new attempt can be triggered.
```

relay-server/src/actors/project.rs
```diff
@@ -787,12 +808,23 @@ impl Project {
         })
     }
 
+    /// Run the checks on incoming envelopes.
```
olksdr:
```suggestion
    /// Runs the checks on incoming envelopes.
```

jan-auer (approved):
:shipit:

PR #1858, opened 2023-02-16
ref(server): Move ProjectConfig to separate crate

jjbayer:
In order to expose a function in `relay-cabi` that validates project configs, we have to move `ProjectConfig` out of `relay-server` to prevent that crate becoming a dependency. We briefly considered moving it to `relay-general`, but decided that that crate is already too large and unspecific.

Instead, move `ProjectConfig` to a new crate `relay-dynamic-config`. The name foreshadows future additions such as org-level config or even dynamic global config passed down from sentry.

This PR does not make any functional changes.

Relates to https://github.com/getsentry/team-ingest/issues/60.

#skip-changelog


relay-dynamic-config/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+processing = []
+
+[dependencies]
+# TODO: Move types used by ProjectConfig from their respective crates to this one. Eventually,
+# this crate should not depend on any sibling crates, except maybe relay-common.
+relay-auth = { path = "../relay-auth" }
+relay-common = { path = "../relay-common" }
+relay-filter = { path = "../relay-filter" }
+relay-general = { path = "../relay-general" }
+relay-sampling = { path = "../relay-sampling" }
```
jjbayer: These are the crates that `relay-cabi` also depends on, so having them here should be acceptable for now.

relay-quotas/src/scoping.rs
```diff
@@ -0,0 +1,343 @@
...
+            Ok(parsed) => parsed,
+            Err(_) => return false,
+        };
+
+        // At this stage, require that the scope is known since we have to fulfill the constraint.
+        self.scope_id(quota.scope) == Some(parsed)
+    }
+
+    /// Checks whether the quota's constraints match the current item.
+    pub fn matches(&self, quota: &Quota) -> bool {
```
jjbayer: This used to be a method on `Quota`, now it's a method on `ItemScoping`. So instead of 

```rust
quota.matches(&item_scoping)
````

we now write

```rust
item_scoping.matches(&quota)
```

This way, `ItemScoping` can remain part of `relay-quotas`.

relay-quotas/Cargo.toml
```diff
@@ -21,6 +21,7 @@ redis = [
 [dependencies]
 relay-common = { path = "../relay-common" }
 relay-log = { path = "../relay-log", optional = true }
+relay-dynamic-config = { path = "../relay-dynamic-config" }
```
jjbayer: This pulls some heavy dependencies into the previously lightweight `relay-quotas`. But the number of indirect dependencies will be reduced once we remove them from `relay-dynamic-config`.

As an alternative intermediate solution, we could add `relay-quotas` as a dependency to `relay-dynamic-config`, and move the `Quota` type back to where it was.

jan-auer: I highly favor the alternative solution you propose: Pull `relay-quotas` as a dependency into `relay-dynamic-config`. That's the true direction: The dynamic config includes quota declarations. For this reason, `relay-quota` comes with feature flags that disable all of the quota enforcement.

We should strive for smaller semantic crates that each ship their own data and configuration types, and then higher-level crates that combine these. 

If you feel like `relay-dynamic-config` is getting too large, introduce a trait: 
- Make the logic around dealing with configs dynamic
- Declare the large, composite config type at a place where you have access to everything (e.g. back in the server).

jan-auer (commented):
Please list the new crate in `relay/src/main.rs` docs. If possible, it would be great if we could also create more documentation as we're moving things.

jjbayer:
@jan-auer I inverted the `quotas` dependencies as suggested. Regarding docs: I added a minimal [crate-level documentation](https://github.com/getsentry/relay/blob/b5734d174a281dd34e5585c75b33b2eb8b54f45e/relay-dynamic-config/src/lib.rs#L1-L38), let me know if you have ideas on what else to add there.

relay-dynamic-config/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    /// Forward compatibility.
+    #[serde(other)]
+    Unknown,
+}
+
+/// Dynamic, per-DSN configuration passed down from Sentry.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(default, rename_all = "camelCase")]
+pub struct ProjectConfig {
```
olksdr: Could we also move the `Feature` and `ProjectConfig` and basically all the code from here to its own module, like `config.rs` and just leave the `lib.rs` for docs and re-imports. 

In my opinion it would make the structure a little bit more cleaner.

jjbayer: Done!

relay-server/src/actors/outcome.rs
```diff
@@ -16,6 +16,8 @@ use std::time::Duration;
 #[cfg(feature = "processing")]
 use anyhow::Context;
 use chrono::{DateTime, SecondsFormat, Utc};
+use relay_quotas::ReasonCode;
+use relay_quotas::Scoping;
```
jan-auer:
```suggestion
use relay_quotas::{ReasonCode, Scoping};
```

relay-server/src/actors/project.rs
```diff
@@ -30,14 +27,11 @@ use crate::actors::project_cache::{
 };
 use crate::envelope::Envelope;
 use crate::extractors::RequestMeta;
-use crate::metrics_extraction::sessions::SessionMetricsConfig;
-use crate::metrics_extraction::transactions::TransactionMetricsConfig;
-use crate::metrics_extraction::TaggingRule;
+use relay_dynamic_config::{Feature, LimitedProjectConfig, ProjectConfig};
```
jan-auer: _nit_: Move to external dependency block.

relay-server/src/endpoints/project_configs.rs
```diff
@@ -13,7 +13,7 @@ use crate::actors::project::{LimitedProjectState, ProjectState};
 use crate::actors::project_cache::{GetCachedProjectState, GetProjectState, ProjectCache};
 use crate::extractors::SignedJson;
 use crate::service::ServiceApp;
-use crate::utils::ErrorBoundary;
+use relay_dynamic_config::ErrorBoundary;
```
jan-auer: _nit_: Move to external dependencies block.

relay/src/main.rs
```diff
@@ -70,6 +70,7 @@
 //! [`relay-common`]: ../relay_common/index.html
 //! [`relay-config`]: ../relay_config/index.html
 //! [`relay-crash`]: ../relay_crash/index.html
+//! [`relay-dynamic-config`]: ../relay_dynamic_config/index.html
```
jan-auer, line 74: Also, add an entry in the list above

relay-dynamic-config/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+homepage = "https://getsentry.github.io/relay/"
+repository = "https://github.com/getsentry/relay"
+version = "23.1.1"
+edition = "2021"
+license-file = "../LICENSE"
+publish = false
+
+[features]
+default = []
+processing = []
```
jan-auer: Please double-check, it looks like this feature flag is unused.

jan-auer (approved):

relay-dynamic-config/Cargo.toml
jjbayer: Checked, the string `processing` does not appear anywhere else in the crate.

Issue #1751, opened 2023-01-17
Make pii config forward-compatible


jjbayer:
This is a follow-up to #1639: When we extend the PII part of project config, we want to make sure that old external Relays do not fail parsing the config. For this, we need to add fallback variants to all enums:

- [x] RuleType: https://github.com/getsentry/relay/pull/1741
- [x] Redaction: https://github.com/getsentry/relay/pull/1859
- [ ] ReplaceBehavior
- [ ] SelectorPathItem: https://github.com/getsentry/relay/pull/2209
- [ ] SelectorSpec: https://github.com/getsentry/relay/pull/2202

Ideally, we would already catch unknown variants during deserialization, so we do not have to add an enum variant in code, and we don't have to handle that variant in match statements etc.

TBS1996:
closed on recommdation from jjbayer

PR #1859, opened 2023-02-17
ref(pii): Make Redaction enum forward compatible

olksdr:
This is part of https://github.com/getsentry/relay/issues/1751, which adds forward compatibility to the `Redaction` enum.

~In order to do that I had to split the `Redaction` into 2 separate variants:~
~* `Method` - which catches all the defined and proper redaction methods~
~* `Other` - for everything else, which also currently just being ignored and won't be propagated downstream~

Just added `Other` variant to `Redaction` enum as a catch all. 

To make sure it works, I've added 2 tests in `relay-general/src/pii/redactions.rs`. 

~There is still a question if if want to report an error or warning if we get something we currently do not expect or just ignoring the variant it good enough for now.~ 
Added warning to places when and if the added variant is used.

#skip-changelog

### DIFF
--- a/relay-general/src/pii/redactions.rs
+++ b/relay-general/src/pii/redactions.rs
@@ -48,7 +48,7 @@ pub enum Redaction {
     /// Replaces the value with a hash
     Hash,
     /// Added for forward compatibility as catch-all variant.
-    #[serde(other)]
+    #[serde(other, skip_serializing)]
     Other,
 }
 
--- a/relay-general/src/pii/redactions.rs
+++ b/relay-general/src/pii/redactions.rs
@@ -47,4 +47,32 @@ pub enum Redaction {
     Mask,
     /// Replaces the value with a hash
     Hash,
+    /// Added for forward compatibility as catch-all variant.
+    #[serde(other)]
+    Other,
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    fn test_redaction_deser_method() {
+        let json = r#"{"method": "replace", "text": "[filter]"}"#;
+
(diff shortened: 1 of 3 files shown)


relay-general/src/pii/redactions.rs
```diff
@@ -29,9 +29,25 @@ impl Default for ReplaceRedaction {
...
 
 /// Defines how replacements happen.
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "snake_case", untagged)]
+pub enum Redaction {
+    /// All the reduction methods currently supported.
+    Method(RedactionMethod),
+    /// All the rest. Added for forward compatibility.
+    #[serde(skip_serializing)]
+    Other(serde_json::Value),
```
jjbayer: Could this not be a unit variant on the old `Redaction` enum?

```rust
pub enum Redaction {
    // ...
    Hash,
    
    #serde(other)
    Other,
}
```

IMO this would classify the variant as `Other` and ignore any other fields in the object.

olksdr: Yes. I still think we could actually log it or send the error in case that something strange comes in.  But all this definitely can be ignored if we do not care about this.

jjbayer: We could log a warning when the variant is `Other`, but I think it's OK regardless to drop the actual value.

olksdr: 👍  reverted all to additional `Other` variant, and added `warn!` if this variant ever used.

jjbayer (approved):

PR #1860, opened 2023-02-17
feat(py): Expose validation function for project configs

jjbayer:
Expose a function that validates arbitrary JSON against the `ProjectConfig` schema. This can be used in tests on sentry side, or possible even before writing configs to redis.

Because librelay releases are lagging behind relay master most of the time, adding new fields might pass validation even though the newest relay would reject them. To prevent this, the function has a `strict` mode which checks for stray additional fields.

Part of https://github.com/getsentry/team-ingest/issues/60.

#skip-changelog


relay-dynamic-config/Cargo.toml
```diff
@@ -14,6 +14,8 @@ default = []
 processing = []
 
 [dependencies]
+anyhow = "1.0.66"
+assert-json-diff = "2.0.2"
```
jjbayer, line 17: I added this dependency to get more meaningful error messages, i.e. to enumerate any additional fields in the JSON. I used the library we also use for testing in `relay-replays`. If anyone has better ideas on how to compare JSON, let me know.

py/tests/test_processing.py
```diff
@@ -272,3 +273,13 @@ def test_validate_sampling_configuration():
...
+
+
+def test_validate_project_config():
+    config = {"allowedDomains": ["*"], "trustedRelays": [], "piiConfig": None}
+    # Does not raise:
+    sentry_relay.validate_project_config(json.dumps(config), strict=True)
+    config["foobar"] = True
+    with pytest.raises(ValueError) as e:
+        sentry_relay.validate_project_config(json.dumps(config), strict=True)
+    assert str(e.value) == 'json atom at path ".foobar" is missing from rhs'
```
jjbayer, line 285: The error message is a bit cryptic, but at least it contains the name of the problematic field.

relay-cabi/include/relay.h
```diff
@@ -579,4 +579,9 @@ struct RelayStr relay_validate_sampling_condition(const struct RelayStr *value);
  */
 struct RelayStr relay_validate_sampling_configuration(const struct RelayStr *value);
 
+/**
+ * Validate entire project config. If `strict` is true, checks for unknown fields in the input.
```
iker-barriocanal: "checks" in the explanation of `strict` is confusing. I find more clear the sentence in the relay function itself:

> If `strict` is true, verify that reserializing the instance results in the same fields.

If you want to have a not-so-defined sentence here to leave room for modifications on the other side, I'd suggest linking to the other function in `relay-dynamic-config/src/utils.rs` and removing the ambiguity.

py/tests/test_processing.py
iker-barriocanal, line 285: question/nit: isn't more idiomatic to place this assertion inside the context manager (even if it's correct this way)?

iker-barriocanal (approved):

py/tests/test_processing.py
jjbayer: When `validate_project_config` raises an exception, the assert would never be reached. That is why it has to be outside.

relay-server/src/actors/outcome.rs
```diff
@@ -16,6 +16,8 @@ use std::time::Duration;
 #[cfg(feature = "processing")]
 use anyhow::Context;
 use chrono::{DateTime, SecondsFormat, Utc};
+use relay_quotas::ReasonCode;
+use relay_quotas::Scoping;
 use relay_system::{Interface, NoResponse};
```
olksdr: Please, sort the imports out. Move these to relay stuff

relay-cabi/include/relay.h
jjbayer: Good suggestion, changed in newest commit.

relay-server/src/actors/outcome.rs
jjbayer: Thanks, I messed something up during merge. Reverted to master version now.

PR #1862, opened 2023-02-20
ref(server): Use async/await in all endpoints

jan-auer:
Moves all endpoints and most HTTP utilities to async/await. This is to prepare
upgrading the web framework.

#skip-changelog


jan-auer:
The failing integration test shows that error reporting has changed with my refactored store endpoints. Since the endpoints now return an error, `actix_web::pipeline` logs a duplicate error message (but without the causes). This is undesirable since it would create noise in Sentry.

See the following excerpt. The first line is our manual log, the other two are now added by actix_web:

```
[relay_server::endpoints::common] ERROR: error handling request: failed to queue envelope
  caused by: envelope buffer capacity exceeded
[actix_web::pipeline] ERROR: Error occurred during request handling, status: 503 Service Unavailable failed to queue envelope
[actix_web::pipeline] DEBUG: QueueFailed(BufferError)
```

I'm looking into a quick fix. It's useful to return an error, but on the other hand we want clean error logs and reports.

**Update:** Like in the past, the only workaround is to convert `BadStoreRequest` before it hits the actix-web pipeline. Since we already needed specific boilerplate to create actix-web handlers, I decided to create a simple wrapper functions that converts the error and creates a compat future.

relay-server/src/endpoints/common.rs
```diff
@@ -323,113 +317,56 @@ fn queue_envelope(
     Ok(())
 }
 
-/// Handles Sentry events.
+/// Handles an envelope store request.
 ///
-/// Sentry events may come either directly from an HTTP request (the store endpoint calls this
-/// method directly) or are generated inside Relay from requests to other endpoints (e.g. the
+/// Sentry envelopes may come either directly from an HTTP request (the enveloipe endpoint calls
```
olksdr:
```suggestion
/// Sentry envelopes may come either directly from an HTTP request (the envelope endpoint calls
```

relay-server/src/endpoints/minidump.rs
```diff
@@ -50,16 +52,28 @@ fn infer_attachment_type(field_name: Option<&str>) -> AttachmentType {
...
-fn get_embedded_minidump(
+/// Extract a minidump from a nested multipart form.
+///
+/// This field is not a minidump (i.e. it doesn't start with the minidump magic header).
+/// It could be a multipart field containing a minidump; this happens in old versions of
+/// the Linux Electron SDK.
+///
+/// Unfortunately, the embedded multipart field is not recognized by the multipart parser
+/// as a multipart field containing a multipart body. For this case we will look if field
+/// the field starts with a '--' and manually extract the boundary (which is what follows
```
olksdr:
```suggestion
/// as a multipart field containing a multipart body. For this case we will look if 
/// the field starts with a '--' and manually extract the boundary (which is what follows
```

relay-server/src/service.rs
```diff
@@ -210,17 +210,17 @@ impl ServiceState {
         })
     }
 
-    /// Returns an atomically counted reference to the config.
-    pub fn config(&self) -> Arc<Config> {
-        self.config.clone()
+    /// Returns an reference to the Relay configuration.
```
olksdr:
```suggestion
    /// Returns a reference to the Relay configuration.
```

olksdr (approved):
lgtm 🚀

relay-server/src/extractors/shared_payload.rs
```diff
@@ -46,6 +48,29 @@ impl SharedPayload {
...
+        if let Ok(None) = result {
+            self.done.store(true, Ordering::Relaxed);
+        }
+
+        result
+    }
+
+    /// Consume the remaining response body.
+    pub async fn consume(mut self) {
+        while let Ok(Some(_)) = self.chunk().await {}
```
jjbayer, line 72: If `self.chunk()` returns an error, this function will silently fail. Is that the same behavior as we had before?

jjbayer (approved):
Did not look at the details but in general this looks good to me! Great improvement in code readability.

relay-server/src/extractors/shared_payload.rs
jan-auer: Yes, we want to consume until there's an error and then bail.

iker-barriocanal (commented):
So much simpler! 🙌

PR #1863, opened 2023-02-21
feat(server): Centralized queueing 

jjbayer:
- [x] Replace `VeqDeque` with wrapper queue referencing a central queueing service.
- [x] Minimal tests
- [ ] Write docs
- [ ] Decide whether `Arc<Mutex>` is the right synchronization mechanism here.

Fixes https://github.com/getsentry/team-ingest/issues/76.


relay-server/src/queues/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+        let queue_id = self.queue_id;
+        self.backend().len(&queue_id)
+    }
+
+    fn backend(&self) -> MutexGuard<'_, Q> {
+        (*self.backend).lock().expect("Lock was poisoned")
+    }
+
+    fn backend_mut(&mut self) -> MutexGuard<'_, Q> {
+        (*self.backend).lock().expect("Lock was poisoned")
```
jjbayer, line 49: If `Mutex::lock` returns an error, it means that another thread using the lock has panicked. In that case, we might as well panic in this thread. Not 100% comfortable with this though.

relay-server/src/queues/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+pub trait MultiQueue<K, T> {
+    fn push_back(&mut self, queue_id: &K, item: T);
+    fn pop_front(&mut self, queue_id: &K) -> Option<T>;
+    fn len(&self, queue_id: &K) -> usize;
+}
+
+/// Behaves like a queue, but defers actual work to a MultiQueue backend.
+#[derive(Debug)]
+pub struct QueueView<K, T, Q: MultiQueue<K, T>> {
+    backend: Arc<Mutex<Q>>,
```
jjbayer, line 18: Is `Arc<Mutex>` the way we want to synchronize, or should we use some kind of two-way channel instead?

relay-server/src/queues/mod.rs
```diff
@@ -0,0 +1,51 @@
+use std::marker::PhantomData;
+use std::sync::{Arc, Mutex, MutexGuard};
+
+pub mod mem;
+
+// TODO: docs on all public types and functions.
+
+/// Combine multiple logical queues into a single structure so we can efficiently handle 10.000s of queues.
+pub trait MultiQueue<K, T> {
```
jjbayer, line 9: @olksdr feel free to rename all the types in this module.

PR #1865, opened 2023-02-21
ref: Remove unsupported macos-10.15 GHA runs

asottile-sentry:
Github has removed workers for macOS 10.15. This builds the binary and wheel on newer macOS.

#skip-changelog


.github/workflows/build_library.yml
```diff
@@ -47,10 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - macos-version: "macos-10.15"
-            target: x86_64-apple-darwin
-            py-platform: macosx-10_15_x86_64
-          - macos-version: "macos-11.0"
+          - macos-version: "macos-11"
```
iker-barriocanal: I checked GitHub's recommendation to see if there was any difference between `macos-11` and `macos-11.0`, and they recommend the former (the latter may not even exist). See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.

iker-barriocanal (approved):

.github/workflows/build_library.yml
```diff
@@ -47,10 +47,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - macos-version: "macos-10.15"
```
olksdr, line 50: Do we also want to add `macos-12` to the matrix? /cc @jan-auer

jan-auer: No, we don't care about more recent versions. Initially, we only stuck with the old workers because they used to be available faster. 

Rust MacOS builds are forwards compatible and even backwards compatible usually, so we can run everything with the default IIRC

.github/workflows/build_library.yml
```diff
@@ -47,10 +47,7 @@ jobs:
...
-            target: x86_64-apple-darwin
```
jan-auer, line 51: We still need to build an x86_64 wheel, though. Don't remove the build, just upgrade the machine please.

jan-auer (changes requested):

jan-auer:
We need to check if we can continue distributing a wheel that's compatible with macOS 10.x somehow. Merely 3 years ago, Catalina was still the latest macOS release, and we're already dropping support for it in our development environment.

If we end up deprecating 10.15, we need to add a changelog entry in `py`.

jan-auer:
I'll check the wheel on a 10.15 machine later.

jan-auer (approved):

PR #1870, opened 2023-02-22
feat(metrics): Tag the sample decision on count_per_root_project

jan-auer:
Adds a binary `decision` tag to `c:transactions/count_per_root_project`. This allows us to query the effective sample rate on projects.

## Background

Dynamic sampling is based on a target sample rate, called the organization's _"fidelity"_. On top of that, Sentry computes a set of _priorities_ (formerly called "biases") to increase visibility in under-represented areas. For example, Sentry boosts releases during adoption or small projects.

The priorities currently apply on top of the fidelity rate and increase the sample rate for a subset of transactions. Overall, this increases the effective sample rate applied for a project. The extent of this is not deterministic, as it depends on the match rate of sample rules.

The **final goal** is to adjust the biases and base sample rate in such a way that the effective sample rate matches the target fidelity. 

## Follow Up

In Sentry, compute the effective sample rate ($e$) and compare it with the target sample rate ($t$). Use this to apply an adjustment to the base sample rate as: $b = t * t/e$. 

Requires https://github.com/getsentry/sentry/pull/45031
Fixes https://github.com/getsentry/team-ingest/issues/16


relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -426,17 +430,24 @@ fn extract_transaction_metrics_inner(
...
 
+    let mut root_counter_tags = BTreeMap::new();
+    if let Some(name) = transaction_from_dsc {
+        root_counter_tags.insert("transaction".to_owned(), name.to_owned());
+    }
+    let decision = match sampling_result {
+        SamplingResult::Keep => "keep".to_owned(),
+        SamplingResult::Drop(_) => "drop".to_owned(),
+    };
+    root_counter_tags.insert("decision".to_owned(), decision);
```
iker-barriocanal, line 442: Does it make sense to implement `ToString` in `SamplingResult`, and directly `sampling_result.to_string()`?

iker-barriocanal (approved):

relay-server/src/actors/processor.rs
```diff
@@ -2715,8 +2732,9 @@ mod tests {
                     TestSemaphore::new(42).try_acquire().unwrap(),
                 ),
             };
-            let result = service.sample_envelope(&mut state);
-            assert_eq!(result.is_ok(), shouldkeep);
+
+            service.compute_sampling_decision(&mut state);
```
jjbayer, line 2739: nit: `compute_sampling_decision` is now a very thin wrapper around `should_keep_event()`, so we might as well only test `should_keep_event()` here. Alternatively, if we want to test that `sample_envelope()` correctly converts a sampling result to a `Result<...>`, revert to calling `sample_envelope()` here.

jjbayer (approved):
One comment about a test, apart from that this looks good to me!

relay-server/src/metrics_extraction/transactions.rs
jan-auer: In this case I would likely not use it, because if we look at `SamplingResult` in isolation we would also render the rule IDs from `Drop` (see the `Display` implementation of `MatchedRuleIds`).

andriisoldatenko (approved):

relay-server/src/actors/processor.rs
jan-auer: Good observation. Instead of testing `should_keep_event` here, we can even remove this test, since there are unit tests for this function already. 

I'd probably test an entire `process` run then, that asserts we wire everything up correctly. Will look into this.

jan-auer:
`"decision"` and `"keep"` already exist in the indexer. See https://github.com/getsentry/sentry/pull/45031

Issue #1867, opened 2023-02-22
Strip quotes from client hints headers


TBS1996:
Currently some client hints headers contain quotes around their values, which we should strip. Example: `["Sec-Ch-Ua-Platform","\"Windows\""]`

We should strip them before we put them in the different contexts. Currently on sentry discover you find both `windows` and `"windows"`.

PR #1874, opened 2023-02-23
feat(general): Strip quotes from client hints values

TBS1996:
Currently some values from client hints have additional quotes which makes filtering in sentry discover harder, as shown here: https://github.com/getsentry/relay/issues/1867

This PR simply strips any such quotes from OS and device hints. Browser remains unchanged as the regex should take care of that.


relay-general/src/protocol/contexts/device.rs
```diff
@@ -174,9 +174,14 @@ impl DeviceContext {
...
     fn parse_client_hints(client_hints: &ClientHints<&str>) -> Option<Self> {
-        let device = client_hints.sec_ch_ua_model?.to_owned();
-
-        if device.trim().is_empty() {
+        let device = client_hints
+            .sec_ch_ua_model?
+            .trim()
+            .trim_start_matches('\"')
+            .trim_end_matches('\"')
+            .to_owned();
```
olksdr: Why just do not use [`replace`](https://doc.rust-lang.org/std/primitive.str.html#method.replace) instead all of this?

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - Add or remove the profile ID on the transaction's profiling context. ([#1801](https://github.com/getsentry/relay/pull/1801))
 - Implement a new sampling algorithm with factors and multi-matching. ([#1790](https://github.com/getsentry/relay/pull/1790)
 - Add Cloud Resource context. ([#1854](https://github.com/getsentry/relay/pull/1854))
+- Strip quotes from client hint values. ([#1874](https://github.com/getsentry/relay/pull/1874))
```
olksdr: This must be moved to `Unreleased` section.

olksdr (changes requested):
Basically the question why did you opt out for `trim_start_...` and `trim_end_...` instead of just using `replace` to remove the quotes (and other symbols if needed)? 

I assume we already know that there are only quotes around and won't be anything like this in the middle of the string.

relay-general/src/protocol/contexts/device.rs
TBS1996: I didn't think of that, I changed it now

CHANGELOG.md
TBS1996: Oops, thank you!

relay-general/src/protocol/contexts/os.rs
```diff
@@ -51,15 +51,26 @@ impl OsContext {
...
 impl FromUserAgentInfo for OsContext {
     fn parse_client_hints(client_hints: &ClientHints<&str>) -> Option<Self> {
-        let platform = client_hints.sec_ch_ua_platform?;
-        let version = client_hints.sec_ch_ua_platform_version.map(str::to_owned);
+        let platform = client_hints
+            .sec_ch_ua_platform?
+            .trim()
+            .trim_start_matches('\"')
+            .trim_end_matches('\"')
+            .to_owned();
```
olksdr: Also here you can use `replace`

olksdr (approved):
One spot left where you have to use `replace`. Other than that it's ok.

PR #1877, opened 2023-02-23
fix(replays): Enforce rate limits

jjbayer:
Rate limits for replays were checked but not enforced. See test case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Protocol validation for source map image type. ([#1869](https://github.com/getsentry/relay/pull/1869))
 - Add PHP support for profiling. ([#1871](https://github.com/getsentry/relay/pull/1871))
 
+**Bug Fixes**:
+
+- Enforce rate limits for session replays. ([#1877]https://github.com/getsentry/relay/pull/1877)
+
 **Internal**:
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - Protocol validation for source map image type. ([#1869](https://github.com/getsentry/relay/pull/1869))
 - Add PHP support for profiling. ([#1871](https://github.com/getsentry/relay/pull/1871))
 
+**Bug Fixes**:
+
+- Enforce rate limits for session replays. ([#1877]https://github.com/getsentry/relay/pull/1877)
```
```suggestion
- Enforce rate limits for session replays. ([#1877](https://github.com/getsentry/relay/pull/1877))
```

relay-server/src/utils/rate_limits.rs
```diff
@@ -562,6 +562,13 @@ where
             return false;
         }
 
+        // Remove replays independently of events
```
```suggestion
        // Remove replays independently of events.
```

CHANGELOG.md
jjbayer: Thanks. I should not write markup manually.

jan-auer (approved):

jan-auer:
Moving ahead with this as the monthly quota is not _counted_ through these rate limits. We need to follow up with correct counting, however.

PR #1878, opened 2023-02-23
feat(profiling): Deprecate fields on the profiling sample format

phacops:
The fields `relative_start_ns` and `relative_end_ns` were intended to specify the start and end of a transaction relative to a profile in the case where we'd capture one profile for many transactions. We changed our approach and are simply having 1 profile for 1 transaction.

The SDKs are starting the profile after transaction start is called and stopping it when transaction end is called. Therefore, the profile is contained within the bounds of the transaction and those 2 fields are not useful anymore (we don't need 2 fields indicating the start and stop of the transaction with regards to the profile).

For compatibility reasons with older SDKs, we're going to keep them on the profile but not require them for newer SDKs.

At the same time, since we have 1 profile for 1 transaction, we're making the field `transaction` be a single transaction and not passing it as an array anymore in `transactions`. Once again, for compatibility reasons, we'll accept both inputs and copy the transaction from `transactions[0]` to `transaction` if needed or return an error if it's in neither field.


relay-profiling/src/transaction_metadata.rs
```diff
@@ -35,11 +43,7 @@ pub struct TransactionMetadata {
 
 impl TransactionMetadata {
     pub fn valid(&self) -> bool {
-        !self.id.is_nil()
-            && !self.name.is_empty()
-            && !self.trace_id.is_nil()
-            && self.relative_cpu_start_ms <= self.relative_cpu_end_ms
-            && self.relative_start_ns < self.relative_end_ns
+        !self.id.is_nil() && !self.name.is_empty() && !self.trace_id.is_nil()
```
jjbayer, line 46: Could you add a test case that shows that a JSON payload without `relative_start_ns` / `end_ns` deserializes correctly and is now `valid`?

relay-profiling/src/sample.rs
```diff
@@ -232,20 +232,27 @@ fn parse_profile(payload: &[u8]) -> Result<SampleProfile, ProfileError> {
...
-        return Err(ProfileError::NoTransactionAssociated);
+    // Clean samples before running the checks.
+    profile.remove_single_samples_per_thread();
+
+    let transaction = match &profile.transaction {
+        Some(transaction) => transaction,
+        None => profile
+            .transactions
+            .get(0)
+            .ok_or(ProfileError::NoTransactionAssociated)?,
```
jjbayer: You could `.take()` both `profile.transaction` and `profile.transactions`, then you would not have to clone further down.

jjbayer (approved):
Two small requests, apart from those this looks good to me!

jan-auer:
> For compatibility reasons with older SDKs, we're going to keep them on the profile but not require them for newer SDKs.

Keep sending the values from ~older~ newer SDKs, please. Otherwise, all developers on single tenant or self-hosted running older versions cannot upgrade your SDK until they also upgrade Sentry. 

Making a value optional means you require a breaking SDK change.

phacops:
I'm not sure if you meant to say "keep sending values in newer SDKs" or not. Those values will keep being sent from older SDKs.

We were using these values when we received multiple transactions for a profile. We only taking the first one into account now, we're not slicing the profiles so we don't need those values.

Older SDKs will keep sending the values and we'll take them into account if we have them (even if we don't need to, but not changing the behavior there).

Newer SDKs won't have to send them because they are already sending one transaction only and we don't need the value for further processing.

relay-profiling/src/sample.rs
```diff
@@ -232,20 +232,28 @@ fn parse_profile(payload: &[u8]) -> Result<SampleProfile, ProfileError> {
         return Err(ProfileError::MissingProfileMetadata);
     }
 
-    let Some(transaction) = profile.transactions.get(0).cloned() else {
-        return Err(ProfileError::NoTransactionAssociated);
-    };
+    if profile.transaction.is_none() {
+        profile.transaction = profile.transactions.first().cloned();
```
jjbayer: I overlooked that `Vec::take` is nightly API, but this (though admittedly ugly) should work:
```suggestion
        profile.transaction = std::mem::take(&mut profile.transactions).into_iter().next();
```

relay-profiling/src/transaction_metadata.rs
```diff
@@ -13,9 +13,17 @@ pub struct TransactionMetadata {
     #[serde(default, deserialize_with = "deserialize_number_from_string")]
     pub active_thread_id: u64,
 
-    #[serde(deserialize_with = "deserialize_number_from_string")]
+    #[serde(
+        default,
```
jan-auer, line 17: To clarify, my above comment regards this change. You cannot make this field optional in SDKs if they have been sending this before. Otherwise, your SDK upgrade will break self-hosted (and temporarily single tenant) Sentry installations. Parsing of those payloads will fail, and (old) Relay will stop ingesting payloads from these newer SDK versions.

cleptric:
@phacops I Just talked to Jan, and we only need to consider single-tenant installations.
As nobody can use profiling on self-hosted right now, we do not run into the issue that a new Sentry SDK version would send data to an old Sentry self-hosted installation.

relay-profiling/src/transaction_metadata.rs
jan-auer: Please disregard this as per https://github.com/getsentry/relay/pull/1878#issuecomment-1446521577.

relay-profiling/src/sample.rs
jan-auer: That's the same as `profile.transactions.drain(..)...`

relay-profiling/src/sample.rs
```diff
@@ -232,20 +232,26 @@ fn parse_profile(payload: &[u8]) -> Result<SampleProfile, ProfileError> {
...
-    // Clean samples before running the checks.
-    profile.remove_single_samples_per_thread();
-    profile.profile.samples.retain_mut(|sample| {
-        (transaction.relative_start_ns..transaction.relative_end_ns)
-            .contains(&sample.elapsed_since_start_ns)
-    });
+    // This is to be compatible with older SDKs
+    if transaction.relative_end_ns > 0 {
+        profile.profile.samples.retain_mut(|sample| {
+            (transaction.relative_start_ns..transaction.relative_end_ns)
```
Zylphrex: Can we make `relative_end_ns` inclusive?

Zylphrex (approved):

PR #1879, opened 2023-02-24
ref(cache): Move buffering of pending envelope to ProjectCache

olksdr:
These changes moving buffering of the incoming envelopes in the `ProjectCache`.  

Current implementation still keeps, so called queue in memory and using `HashMap` with a composite key `QueueKey {key, sampling_key}`, where `sampling_key` can be the same as a key if there is no sampling project identified. The values to these keys are `Vec` of boxed `Envelope` with their `EnvelopeContext`. 

Once we get an update for project state, we check all variants of `QueueKey` which contains the current `ProjectKey` and if all the project states are cached we try to flush buffered envelopes indexed by these `QeueuKey`.

The envelops will be buffered if:
* the project state is not fetched yet
* root project is here but the sampling project state is not fetched yet
* the sampling project state is here but the root project is not fetched yet

This change also removes all the buffering from the `Project` and reduces its responsibility. Now it just keeps its own state and configuration and the envelope handling is done outside of it.


relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +373,71 @@ struct UpdateProjectState {
...
+        let keys: Vec<_> = self
+            .spool
+            .iter()
+            .filter(|(k, _)| k.key == key || k.sampling_key == key)
+            .map(|(k, _)| k)
+            .copied()
+            .collect();
+
+        // Remove all the spooled data for this project key and return it to the caller.
+        let mut values = VecDeque::new();
```
jjbayer: This might as well be a vector, right?

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +373,71 @@ struct UpdateProjectState {
...
+        for k in keys {
+            if let Some(value) = self.spool.remove(&k) {
+                values.extend(value)
+            }
+        }
+        values
```
jjbayer: I assume this is where we would need batching once we introduce disk spooling?

relay-server/src/actors/project_cache.rs
```diff
@@ -467,8 +506,18 @@ impl ProjectCacheBroker {
...
+
         self.get_or_create_project(project_key)
-            .update_state(state, no_cache)
+            .update_state(state, no_cache);
+
+        // There is no much sense to dequeue and check anything if the incoming state is invalid.
+        if !invalid {
+            let mut envelopes = self.spool.dequeue(project_key);
+            while let Some((envelope, envelope_context)) = envelopes.pop_front() {
+                self.handle_processing(envelope, envelope_context)
```
jjbayer: This puts an envelope back on the queue when one of the two project states is still missing. I think this works, but it would be neat to have an optimization to make sure that an envelope gets spooled only once, and retrieved only when both project states are present.

jjbayer: This would also prevent infinite looping, although that might be only a theoretical concern.

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +374,109 @@ struct UpdateProjectState {
...
+    sampling_key: ProjectKey,
+}
+
+/// The queue (buffer) of the incoming envelopes.
+#[derive(Debug, Default)]
+struct Queue {
+    /// Contains the cache of the incoming envelopes.
+    buffer: HashMap<QueueKey, Vec<(Box<Envelope>, EnvelopeContext)>>,
+    /// Index of the buffered project keys.
+    index: HashMap<ProjectKey, HashSet<QueueKey>>,
```
jjbayer: nit 1: We tend to use `BTreeSet` for small sets, because the `log(N)` lookup might be faster than hashing and comparing keys when the tree is small.
```suggestion
    index: HashMap<ProjectKey, BTreeSet<QueueKey>>,
```

See also https://github.com/Rufflewind/bench-maps.

relay-server/src/actors/project_cache.rs
```diff
@@ -519,30 +646,121 @@ impl ProjectCacheBroker {
...
+            (Some(state), Some((_, Some(sampling_state)))) => {
+                self.handle_processing(key, state, Some(sampling_state), envelope, envelope_context)
+            }
+
+            // Process our envelope, since there is no sampling key on this envelope.
+            (Some(state), None) => {
+                self.handle_processing(key, state, None, envelope, envelope_context)
+            }
+
+            // We have to buffe because one of the followup conditiones met:
```
```suggestion
            // We have to buffer because one of the followup conditions met:
```

relay-server/src/actors/project_cache.rs
```diff
@@ -519,30 +646,121 @@ impl ProjectCacheBroker {
...
+        project_key: ProjectKey,
+        state: Arc<ProjectState>,
+        sampling_state: Option<Arc<ProjectState>>,
+        envelope: Box<Envelope>,
+        envelope_context: EnvelopeContext,
+    ) {
+        if let Ok(CheckedEnvelope {
+            envelope: Some((envelope, envelope_context)),
+            ..
+        }) = self
```
jjbayer: Is it OK to swallow the error case here?

relay-server/src/actors/project_cache.rs
```diff
@@ -460,15 +541,61 @@ impl ProjectCacheBroker {
...
-            .update_state(state, no_cache)
+            .update_state(state.clone(), no_cache);
+
+        // There is no much sense to dequeue and check anything if the incoming state is invalid.
+        if !invalid {
+            let envelopes = self.pending_envelopes.dequeue(&project_key, |queue_key| {
+                for key in &[queue_key.key, queue_key.sampling_key] {
+                    // Do not check just updated state again.
+                    if *key == project_key {
+                        continue;
```
```suggestion
                    if *key == project_key {
                        // We know that this state is valid, so only check the other one.
                        continue;
```

relay-server/src/actors/project_cache.rs
```diff
@@ -431,7 +509,10 @@ impl ProjectCacheBroker {
 
         // Defer dropping the projects to a dedicated thread:
         let mut count = 0;
-        for (_, project) in expired {
+        for (project_key, project) in expired {
+            // Dequeue all the envelopes linked to the disposable project, which will be dropped
+            // once this for loop exits.
+            self.pending_envelopes.dequeue(&project_key, |_| true);
```
jjbayer: Good catch!

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +374,109 @@ struct UpdateProjectState {
...
+    {
+        let mut keys = self
+            .index
+            .get_mut(partial_key)
+            .map(mem::take)
+            .unwrap_or_default();
+
+        let mut result = vec![];
+        // Find those keys which match predicates.
+        for qkey in keys.drain() {
```
jjbayer: nit: This variable is called `queue_key` in other places.
```suggestion
        for queue_key in keys.drain() {
```

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +374,109 @@ struct UpdateProjectState {
...
+                if let Some(envelopes) = self.buffer.remove(&qkey) {
+                    result.extend(envelopes);
+                }
+            }
+            // Return keys into the index, where predicate is failing.
+            else {
+                self.index
+                    .entry(*partial_key)
+                    .or_insert(HashSet::from([qkey]))
+                    .insert(qkey);
```
jjbayer: This inserts the key possibly twice, which might not matter logically because it is a set, but it still seems duplicate.
```suggestion
                    .or_default()
                    .insert(qkey);
```
Alternative would be

```rust
match entry {
    Occupied(e) => e.insert(qkey),
    Vacant(e) => e.insert(HashSet::from([qkey])
}
```

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +374,109 @@ struct UpdateProjectState {
...
+        partial_key: &ProjectKey,
+        f: P,
+    ) -> Vec<(Box<Envelope>, EnvelopeContext)>
+    where
+        P: Fn(&QueueKey) -> bool,
+    {
+        let mut keys = self
+            .index
+            .get_mut(partial_key)
+            .map(mem::take)
```
jjbayer: This leaves an empty set on the index, so you might as well remove it. Or does the empty set represent some special state?
```suggestion
            .remove(partial_key)
```

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +374,109 @@ struct UpdateProjectState {
...
+
+        // Remove the index all together if it's empty.
+        let empty_index = self
+            .index
+            .get(partial_key)
+            .map(|v| v.is_empty())
+            .unwrap_or_default();
+        if empty_index {
+            self.index.remove(partial_key);
+        }
```
jjbayer: I think this would not be necessary with the suggestion I made above.

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +374,109 @@ struct UpdateProjectState {
...
+            .get_mut(partial_key)
+            .map(mem::take)
+            .unwrap_or_default();
+
+        let mut result = vec![];
+        // Find those keys which match predicates.
+        for qkey in keys.drain() {
+            if f(&qkey) {
+                if let Some(envelopes) = self.buffer.remove(&qkey) {
+                    result.extend(envelopes);
```
jjbayer, line 431: It would be nice if this function could return an iterator of envelopes, something like 

```rust
keys.drain().map(|qkey| {
    // ...
    envelopes.into_iter()
}).flatten()
```

Just need to be careful to consume the iterator when calling `dequeue` before garbage disposal.

jjbayer (changes requested):
The logic seems sound to me, just some general questions / code suggestions. Let me know if I made any wrong assumptions.

relay-server/src/actors/project_cache.rs
olksdr: makes sense , done in [435ac7b](https://github.com/getsentry/relay/pull/1879/commits/435ac7b6203eb7fc8ff74f15f05c68f043767551)

relay-server/src/actors/project_cache.rs
olksdr: done in [435ac7b](https://github.com/getsentry/relay/pull/1879/commits/435ac7b6203eb7fc8ff74f15f05c68f043767551)

relay-server/src/actors/project_cache.rs
olksdr: That's how this was done before, but  we also can rethink this behaviour now.

relay-server/src/actors/project_cache.rs
olksdr: done [435ac7b](https://github.com/getsentry/relay/pull/1879/commits/435ac7b6203eb7fc8ff74f15f05c68f043767551)

relay-server/src/actors/project_cache.rs
olksdr: done in [435ac7b](https://github.com/getsentry/relay/pull/1879/commits/435ac7b6203eb7fc8ff74f15f05c68f043767551)

relay-server/src/actors/project_cache.rs
olksdr: done in [435ac7b](https://github.com/getsentry/relay/pull/1879/commits/435ac7b6203eb7fc8ff74f15f05c68f043767551)

relay-server/src/actors/project_cache.rs
olksdr: done in [435ac7b](https://github.com/getsentry/relay/pull/1879/commits/435ac7b6203eb7fc8ff74f15f05c68f043767551)

relay-server/src/actors/project_cache.rs
olksdr: Still have to think about this a bit.

relay-server/src/actors/project_cache.rs
```diff
@@ -131,11 +133,12 @@ impl CheckEnvelope {
 /// [`CheckEnvelope`]. Once the envelope has been validated, remaining items are forwarded to the
 /// next stage:
 ///
-///  - If the envelope needs dynamic sampling, this sends [`AddSamplingState`] to the
-///    [`ProjectCache`] to add the required project state.
+///  - If the envelope needs dynamic sampling, and the project state is not cached or out of the
```
jan-auer, line 135: _nit_: We could rewrite parts of the paragraph before. It talks about stages, but now we just wait for both project states _concurrently_ (or simultaneously) and then move on to `CheckEnvelope`.

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +373,92 @@ struct UpdateProjectState {
     no_cache: bool,
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
```
jan-auer, line 375: cc @jjbayer, do we want to define the order of derives in the coding guideline?

relay-server/src/actors/project.rs
```diff
@@ -20,16 +19,15 @@ use relay_system::BroadcastChannel;
 
 use crate::actors::envelopes::{EnvelopeManager, SendMetrics};
 use crate::actors::outcome::{DiscardReason, Outcome};
-use crate::actors::processor::{EnvelopeProcessor, ProcessEnvelope};
-use crate::actors::project_cache::{
-    AddSamplingState, CheckedEnvelope, ProjectCache, RequestUpdate,
-};
+#[cfg(feature = "processing")]
+use crate::actors::processor::EnvelopeProcessor;
```
iker-barriocanal, line 23: Why is this only targeting processing relays?

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +372,83 @@ struct UpdateProjectState {
...
+impl Queue {
+    /// Creates an empty queue.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Adds the value to the queue for the provided key.
+    pub fn enqueue(&mut self, key: QueueKey, value: (Box<Envelope>, EnvelopeContext)) {
+        self.index.entry(key.key).or_default().insert(key);
+        self.index.entry(key.sampling_key).or_default().insert(key);
```
iker-barriocanal: I don't understand the purpose of another entry with `key.sampling_key`, why are we doing that?

relay-server/src/actors/project_cache.rs
iker-barriocanal: Maybe we can use [`drain_filter`](https://doc.rust-lang.org/stable/std/collections/struct.BTreeSet.html#method.drain_filter)?

relay-server/src/actors/project_cache.rs
```diff
@@ -431,15 +481,23 @@ impl ProjectCacheBroker {
...
         // Defer dropping the projects to a dedicated thread:
         let mut count = 0;
-        for (_, project) in expired {
+        for (project_key, project) in expired {
+            // Dequeue all the envelopes linked to the disposable project, which will be dropped
+            // once this for loop exits with an `Invalid(Internal)` outcome.
+            let envelopes = self.pending_envelopes.dequeue(&project_key, |_| true);
+            relay_log::with_scope(
+                |scope| scope.set_tag("project_key", project_key),
+                || relay_log::error!("eviced project with {} envelopes", envelopes.len()),
```
```suggestion
                || relay_log::error!("evicted project with {} envelopes", envelopes.len()),
```

relay-server/src/actors/project_cache.rs
```diff
@@ -468,7 +530,41 @@ impl ProjectCacheBroker {
...
+
+        let envelopes = self.pending_envelopes.dequeue(&project_key, |queue_key| {
+            // Pick envelopes which belong to the incoming project.
+            if project_key != queue_key.key {
+                return false;
+            }
+
+            // We return false if project is not cached or its state is invalid.
+            self.projects
+                .get(&queue_key.sampling_key)
```
iker-barriocanal: Why `queue_key.sampling_key`, instead of `queue_key.key`?

relay-server/src/actors/project_cache.rs
```diff
@@ -519,30 +615,90 @@ impl ProjectCacheBroker {
...
+
+            EnvelopeProcessor::from_registry().send(process);
+        }
     }
 
+    /// Checks an incoming envelope and decides either process it immediately or buffer it.
+    ///
+    /// Few conditions are checked here:
+    /// - If there is no dynamic sampling key and the project is already cached, we do strait to
+    /// processing otherwise buffer the envelopes.
```
```suggestion
    /// - If there is no dynamic sampling key and the project is already cached, we do straight to
    /// processing; otherwise, buffer the envelopes.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -519,30 +615,90 @@ impl ProjectCacheBroker {
...
+    /// - If there is a dynamic sampling key is provided, if the root and sampling project
+    /// are cached - process the envelope, buffer otherwise.
```
```suggestion
    /// - If there a dynamic sampling key is provided, and the root and sampling project
    /// are cached - process the envelope, buffer otherwise.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -519,30 +615,90 @@ impl ProjectCacheBroker {
...
+    ///
+    /// This means if the caches are hot we always process all the incoming envelopes without any
+    /// delay. But in case if the project state cannot be fetched, we keep buffering till the state
+    /// is everyntually updated.
```
```suggestion
    /// This means if the caches are hot we always process all the incoming envelopes without any
    /// delay. But in case the project state cannot be fetched, we keep buffering till the state
    /// is eventually updated.
```

iker-barriocanal (commented):
I'm a bit confused with this PR, so mostly questions.

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +372,83 @@ struct UpdateProjectState {
     no_cache: bool,
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
+struct QueueKey {
+    key: ProjectKey,
```
```suggestion
    /// Key of the project that an envelope belongs to.
    own_key: ProjectKey,
    /// Key of the project which started the trace. If the envelope has no `DynamicSamplingContext`, set to `own_key` instead.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +372,83 @@ struct UpdateProjectState {
...
+        &mut self,
+        partial_key: &ProjectKey,
+        f: P,
+    ) -> Vec<(Box<Envelope>, EnvelopeContext)>
+    where
+        P: Fn(&QueueKey) -> bool,
+    {
+        let mut result = Vec::new();
+
+        let mut keys = self.index.remove(partial_key).unwrap_or_default();
```
```suggestion
        let mut queue_keys = self.index.remove(partial_key).unwrap_or_default();
```

relay-server/src/actors/project_cache.rs
```diff
@@ -400,6 +372,83 @@ struct UpdateProjectState {
...
+        self.index.entry(key.key).or_default().insert(key);
+        self.index.entry(key.sampling_key).or_default().insert(key);
+        self.buffer.entry(key).or_default().push(value);
+    }
+
+    /// Returns the list of buffered envelopes if they satisfy a predicate.
+    pub fn dequeue<P>(
+        &mut self,
+        partial_key: &ProjectKey,
+        f: P,
```
```suggestion
        predicate: P,
```

relay-server/src/actors/project_cache.rs
jjbayer: @iker-barriocanal that's a nightly feature, unfortunately.

relay-server/src/actors/project_cache.rs
```diff
@@ -468,7 +530,41 @@ impl ProjectCacheBroker {
...
+            if project_key != queue_key.key {
+                return false;
+            }
+
+            // We return false if project is not cached or its state is invalid.
+            self.projects
+                .get(&queue_key.sampling_key)
+                // Make sure we have only cached and valid state.
+                .and_then(|p| p.valid_state())
+                .map_or(false, |s| !s.invalid())
```
jjbayer: If I'm not mistaken, this logic has a flaw:

1. Imagine `self.projects` is empty.
2. A user sends a single envelope with a project key and a distinct sampling key.
3. The project state for the event's "own" project comes in first.
4. We leave the envelope on the queue, because the sampling state is not in `self.projects` yet.
5. The project state for the sampling key comes in, but we leave the envelope on the queue because `project_key != queue_key.key`.
6. No new envelopes come in, and the envelope stays on the queue forever.

I think this problem can be overcome by treating the two partial keys symmetrically here. That is:

```suggestion
            let other_key = if queue_key.key == project_key { queue_key.sampling_key } else { queue_key.key };

            // We know that `project_key`s state has been loaded and is valid, so now check the other one:
            self.projects
                .get(&other_key)
                // Make sure we have only cached and valid state.
                .and_then(|p| p.valid_state())
                .map_or(false, |s| !s.invalid())
```

This way, we will always deque if and only if both states are loaded.

relay-server/src/actors/project_cache.rs
```diff
@@ -468,7 +530,41 @@ impl ProjectCacheBroker {
...
+        });
+
+        // Flush envelopes where both states have resolved.
+        for (envelope, envelope_context) in envelopes {
+            let sampling_state = utils::get_sampling_key(&envelope)
+                .and_then(|key| self.projects.get(&key))
+                .and_then(|p| p.valid_state());
+
+            self.handle_processing(
+                project_key,
```
jjbayer: With the suggestion I made above, `handle_processing` should take the envelope's project key, not this one.

relay-server/src/actors/project_cache.rs
```diff
@@ -519,30 +615,90 @@ impl ProjectCacheBroker {
...
+            .get_cached_state(envelope.meta().no_cache())
+            .filter(|st| !st.invalid());
+
+        // Also, fetch the project state for sampling key and make sure it's not invalid.
+        let sampling_key = utils::get_sampling_key(&envelope);
+        let sampling_state = sampling_key.and_then(|key| {
+            self.get_or_create_project(key)
+                .get_cached_state(envelope.meta().no_cache())
+                .filter(|st| !st.invalid())
+        });
```
jjbayer, line 690: This can be moved inside `if let Some(state) = `.

relay-server/src/actors/project_cache.rs
jjbayer: This is not a blocker for the current PR, but generally speaking, if all you want to do is iterate over a sequence once, there is no reason to copy its data over to a new vector.

For our case specifically, once we change to disk spooling, we will need some form of batching anyway, because the dequeued data will be too large to copy into one big vector.

relay-server/src/actors/project_cache.rs
```diff
@@ -374,13 +374,16 @@ struct UpdateProjectState {
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
 struct QueueKey {
-    key: ProjectKey,
+    root_key: ProjectKey,
```
jjbayer: I'm afraid `root` is too ambiguous here, because it's also used to refer to the "root" of the trace, which is the sampling project. I have no better name to be honest. Maybe `own_key` or `project_key`?

relay-server/src/actors/project_cache.rs
```diff
@@ -538,14 +541,15 @@ impl ProjectCacheBroker {
...
-                return false;
-            }
+            let partial_key = if queue_key.root_key == project_key {
+                queue_key.sampling_key
+            } else {
+                queue_key.root_key
+            };
 
-            // We return false if project is not cached or its state is invalid.
+            // We return false if project is not cached or its state is invalid, true otherwise.
```
jjbayer, line 550:
```suggestion
            // We return false if project is not cached or its state is invalid, true otherwise.
            // We only have to check `partial_key`, because we already know that the `project_key`s `state` is valid and loaded.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -468,7 +533,36 @@ impl ProjectCacheBroker {
...
+                .map_or(false, |s| !s.invalid())
+        });
+
+        // Flush envelopes where both states have resolved.
+        for (envelope, envelope_context) in envelopes {
+            let sampling_state = utils::get_sampling_key(&envelope)
+                .and_then(|key| self.projects.get(&key))
+                .and_then(|p| p.valid_state());
+
+            self.handle_processing(state.clone(), sampling_state, envelope, envelope_context)
```
```suggestion
            self.handle_processing(state.clone(), sampling_state, envelope, envelope_context);
```

relay-server/src/actors/project_cache.rs
jjbayer: Bumping this comment. There's no need to `get_or_create_project` the sampling state when the `project_state` is `None`.

relay-server/src/actors/project_cache.rs
```diff
@@ -519,30 +613,90 @@ impl ProjectCacheBroker {
...
     fn handle_validate_envelope(&mut self, message: ValidateEnvelope) {
-        // Preload the project cache for dynamic sampling in parallel to the main one.
-        if let Some(sampling_key) = utils::get_sampling_key(&message.envelope) {
-            self.get_or_create_project(sampling_key)
-                .prefetch(message.envelope.meta().no_cache());
-        }
+        let ValidateEnvelope { envelope, context } = message;
+
+        // Fetch the project state for our key and make sure it's not invalid.
+        let root_key = envelope.meta().public_key();
```
jjbayer: See comment on ambiguity of "root project".

https://github.com/getsentry/relay/blob/e7c862e84d3d3cf76127a2ef80dbb7f4b6235082/relay-server/src/actors/processor.rs#L224

jjbayer (approved):
Please read the comments (especially the one about `root_key`). Nothing blocking though.

relay-server/src/actors/project_cache.rs
olksdr: You're right. 
And I also think it is not important for now but we definitely will have to change this behaviour once we start handling the disk access and might have many more envelopes in persistent buffer.

relay-server/src/actors/project_cache.rs
olksdr: In other places we use projects map directly, and from what I see this is the only one place now we call `self.get_or_create_project(...).get_cached_state(...)` which initiate the upstream request to update the sampling state if it's not in the cache. 

So if the `project_state` is `None` we are most probably request the update for it, and now just queue the envelope, and at the same time we request already sampling state to be able to process those queued envelopes when state are updated. 

But maybe I'm missing something.

relay-server/src/actors/project.rs
olksdr: We have the following in the code: https://github.com/getsentry/relay/blob/381c8936e168895fd241990719a452133daaa993/relay-server/src/actors/project.rs#L798-L807
which already guards some of the code behind the feature flag, so the import must be there as well, otherwise the compiler will complain.

relay-server/src/actors/project_cache.rs
jjbayer: Sorry, you are right. I overlooked that `get_cached_state` triggers a fetch from the upstream.

jan-auer (approved):

relay-server/src/actors/project_cache.rs
jjbayer: @jan-auer what rule would you suggest? Derives are order-sensitive, so there are some limits (see https://github.com/rust-lang/rustfmt/issues/1867#issuecomment-321412488).

PR #1880, opened 2023-02-24
ref(server): Simplify HTTP server and service startup

jan-auer:
The HTTP server and `ServiceState` comprising all services are tightly coupled. This PR separates them in such a way that the service state can be started by itself and is then passed to the HttpServer during startup. Additionally, bootstrapping is streamlined in the top-level `run` method.

This change will allow to refactor the `Controller`, which is required to remove actix from our dependencies.

Follows https://github.com/getsentry/relay/pull/1723
#skip-changelog

### DIFF
--- a/relay-system/src/controller.rs
+++ b/relay-system/src/controller.rs
@@ -122,8 +122,7 @@ impl Controller {
     /// actors have completed a graceful shutdown.
     pub fn run<F, R, E>(factory: F) -> Result<(), E>
     where
-        F: FnOnce() -> Result<R, E> + 'static,
-        F: Sync + Send,
+        F: FnOnce() -> Result<R, E>,
     {
         // Spawn a legacy actix system for the controller's signals.
         let sys = actix::System::new("relay");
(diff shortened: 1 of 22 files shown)


olksdr (approved):
lgtm

relay-system/src/controller.rs
```diff
@@ -120,26 +115,26 @@ impl Controller {
...
     where
-        F: FnOnce() -> Result<R, E> + 'static,
-        F: Sync + Send,
+        F: FnOnce() -> Result<R, E>,
     {
         // Spawn a legacy actix system for the controller's signals.
         let sys = actix::System::new("relay");
 
-        // Ensure that the controller starts if no service has started it yet. It will register with
+        // Ensure that the controller starts if no service has starts it. It will register with
```
olksdr, line 125:
```suggestion
        // Ensure that the controller starts if no service has started it yet. It will register with
```

I think we should revert the comment to previous state.

olksdr (approved):
One small nit about the doc comment, otherwise LGTM!

PR #1885, opened 2023-02-24
feat(profiling): Add donet support for profiling

phacops:


jan-auer:
Would it make sense to treat everything else as a generic platform instead of listing all of them explicitly?

jan-auer (approved):

phacops:
Absolutely. I think we'd have done that already if we anticipated 3 new platforms, so our bad. We'll make the change before the next wave.

Zylphrex (approved):

PR #1886, opened 2023-02-27
feat(monitors): Ingestion for basic check-in payloads

jan-auer:
This adds ingestion for cron monitor checkin payloads through Relay. Initially, this just covers creating and updating checkins, but not monitors. Rate limits can be set and enforced, but there will be no outcomes tracked.

 - [x] New data category
 - [x] New item type
 - [x] Crate for cron-specific logic
 - [x] Rate limiting support
 - [x] Limit for item sizes
 - [x] Validator function in `processing` mode
 - [x] Kafka topic (defaulting to `ingest-monitors`)
 - [x] Kafka producer (for now as _msgpack_)
 - [x] Schema implementation
 - [x] Normalization from the cron PUT payloads

Follow-up:

 - Track outcomes
 - Define rate limiting semantics
 - Add a cron endpoint

See getsentry/sentry#45079 for the consumer.
Requires https://github.com/getsentry/ops/pull/6226


relay-common/src/constants.rs
```diff
@@ -120,6 +120,8 @@ pub enum DataCategory {
     /// This is the category for transaction payloads that were accepted and stored in full. In
     /// contrast, `transaction` only guarantees that metrics have been accepted for the transaction.
     TransactionIndexed = 9,
+    /// Cron monitor checkins.
+    Cron = 10,
```
evanpurkhiser: I wonder if we should call this monitor, we are somewhat consistently calling crons `monitors` within the codebase, and `crons` as the product name at the moment.

relay-crons/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+#[derive(Clone, Copy, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum CheckinStatus {
+    /// Checkin had no issues during execution.
+    Ok,
+    /// Checkin failed or otherwise had some issues.
+    Error,
+    /// Checkin is expectred to complete.
+    InProgress,
```
evanpurkhiser: In the future we may also have a `Late`, but for now this is good.

relay-crons/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+pub enum ProcessCheckinError {
+    /// Failed to deserialize the payload.
+    #[error("failed to deserialize checkin")]
+    Json(#[from] serde_json::Error),
+}
+
+///
+#[derive(Clone, Copy, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum CheckinStatus {
```
olksdr: Could we name it 
```suggestion
enum CheckInStatus {
```

in basically all the instances of it, in docs as well? 

Or rename it to just `Check`

relay-crons/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+fn uuid_simple<S>(uuid: &Uuid, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    uuid.as_simple().serialize(serializer)
+}
+
+/// The cron monitor checkin payload.
```
olksdr:
```suggestion
/// The cron monitor check-in payload.
```

relay-common/src/constants.rs
olksdr: I would also rename it to the `CronMonitor`
```suggestion
    CronMonitor = 10,
```

relay-crons/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Debug, Deserialize, Serialize)]
+struct Checkin {
```
olksdr:
```suggestion
struct CheckIn {
```

relay-kafka/src/config.rs
```diff
@@ -39,14 +39,16 @@ pub enum KafkaTopic {
     ReplayEvents,
     /// ReplayRecordings, large blobs sent by the replay sdk
     ReplayRecordings,
+    /// Cron monitor checkins.
+    Crons,
```
olksdr:
```suggestion
    CronMonitors,
```

olksdr (commented):
overall LGTM, but I would rename all the `Checkin` to `CheckIn` , `checkin` to `check-in`, and `Cron` to `CronMonitor`

relay-crons/src/lib.rs
evanpurkhiser: `CheckIn` is what we have been using the api

relay-crons/src/lib.rs
evanpurkhiser: 👍

relay-crons/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// Normalizes a cron monitor checkin payload.
+///
+/// Returns `None` if the payload was valid and does not have to be changed. Returns `Some` for
+/// valid payloads that were normalized.
+pub fn process_checkin(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckinError> {
+    let mut checkin = serde_json::from_slice::<Checkin>(payload)?;
+
+    // Missed status cannot be ingested, this is computed on the server.
+    if checkin.status == CheckinStatus::Missed {
```
evanpurkhiser: I wonder if there will be cases we want to allow the user to send missed checkins.

This is fine for now

relay-common/src/constants.rs
evanpurkhiser: It's worth noting in the future these may. also support heart beat checkins, at which point the word "cron" may be a bit confusing. but idk. my preference is on keeping this consistent with what we have in code. I don't think we say `CronMonitor` anywhere in the api yet.

relay-server/src/actors/processor.rs
```diff
@@ -1029,6 +1029,28 @@ impl EnvelopeProcessorService {
         });
     }
 
+    /// Normalize cron monitor checkins and remove invalid ones.
+    #[cfg(feature = "processing")]
+    fn process_checkins(&self, state: &mut ProcessEnvelopeState) {
```
evanpurkhiser:
```suggestion
    fn process_monitor_checkins(&self, state: &mut ProcessEnvelopeState) {
```

?

relay-server/src/envelope.rs
```diff
@@ -100,12 +100,14 @@ pub enum ItemType {
...
+    /// Profile event payload encoded as JSON.
     Profile,
-    /// Replay metadata and breadcrumb payload
+    /// Replay metadata and breadcrumb payload.
     ReplayEvent,
-    /// Replay Recording data
+    /// Replay Recording data.
     ReplayRecording,
+    /// Cron monitor checkin encoded as JSON.
+    Checkin,
```
evanpurkhiser: `Mon
```suggestion
    MonitorCheckin,
```

would maybe be consistent with what we have above, we don't call a replay recording `Recording` we call it a `ReplayRecording`.

relay-crons/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+/// The cron monitor checkin payload.
+#[derive(Debug, Deserialize, Serialize)]
+struct Checkin {
+    /// Unique identifier of this checkin.
+    #[serde(serialize_with = "uuid_simple")]
+    checkin_id: Uuid, // TODO(ja): roundtrip without slashes
+
+    /// Identifier of the monitor for this checkin.
+    #[serde(serialize_with = "uuid_simple")]
+    monitor_id: Uuid,
```
evanpurkhiser: I believe this is going to become `monitor_slug` as I mentioned to you.

Let me get confirmation for https://github.com/getsentry/sentry/issues/45166 first though.

relay-common/src/constants.rs
jan-auer: I'll just go with "Monitor" everywhere to keep it concise.

relay-crons/src/lib.rs
jan-auer: In the protocol it's `checkin_id` and also the URL is `checkin`, so I think either we go with the clumsier `check_in_id` for these cases or we embrace `Checkin`.

relay-server/src/envelope.rs
jan-auer: I'm usually preferring the shorter version, `ReplayRecording` was sort of an oversight.

relay-crons/src/lib.rs
jan-auer: Sounds good. I'd propose we merge as-is and follow up with a set of PRs for Relay + Sentry once this has cleared up.

relay-crons/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Status of this checkin. Defaults to `"unknown"`.
+    status: CheckinStatus,
```
jjbayer: Don't we need a `#[serde(default)]` here to derive the default in case the field is missing from the payload?

relay-crons/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Duration of this check since it has started in seconds.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    duration: Option<f64>,
+}
+
+/// Normalizes a cron monitor checkin payload.
+///
+/// Returns `None` if the payload was valid and does not have to be changed. Returns `Some` for
+/// valid payloads that were normalized.
```
jjbayer: Where do we actually return `None`? I think it's a good idea to apply copy-on-write logic here if the payload is valid and unchanged.

jan-auer:
I've updated naming to be consistently:

- `monitor` / `monitors` for everything that regards the product, the general topic (Kafka) and the data category
- `CheckIn` / `check_in` for the check-ins. This is chosen over `MonitorCheckIn` for brevity and also because it is more consistent with other models we have at Sentry (e.g. we have spans, not `TransactionSpan`). Replays are more of an exception here. Also, it's shorter.

relay-crons/src/lib.rs
jan-auer: I've had this discussion with @cleptric . The current endpoint requires this field. We're thinking about making it optional as long as we can still break this API. @evanpurkhiser would you like it to be optional right away?

relay-crons/src/lib.rs
jan-auer: Good catch, that is a leftover from an earlier version. I ended up simplifying and removing the Cow, but in that last version it makes sense again. I'll re-add it.

relay-monitors/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+/// Normalizes a monitor check-in payload.
+pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError> {
+    let mut check_in = serde_json::from_slice::<CheckIn>(payload)?;
+
+    // Missed status cannot be ingested, this is computed on the server.
+    if check_in.status == CheckInStatus::Missed {
+        check_in.status = CheckInStatus::Unknown;
+    }
+
+    Ok(serde_json::to_vec(&check_in)?)
```
jjbayer, line 82: @jan-auer did you still want to make this copy-on-write?

relay-monitors/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+enum CheckInStatus {
+    /// Check-in had no issues during execution.
+    Ok,
+    /// Check-in failed or otherwise had some issues.
+    Error,
+    /// Check-in is expectred to complete.
+    InProgress,
+    /// Monitor did not check in on time.
+    Missed,
+    /// No status was passed.
```
jjbayer, line 42:
```suggestion
    /// No status was passed.
    ///
    /// When we enable check-in deserialization in external relays, this should become `Unknown(String)` for forward compatibility.
```

relay-server/src/actors/processor.rs
```diff
@@ -1029,6 +1029,28 @@ impl EnvelopeProcessorService {
...
+                return true;
+            }
+
+            match relay_monitors::process_check_in(&item.payload()) {
+                Ok(processed) => {
+                    item.set_payload(ContentType::Json, processed);
+                    true
+                }
+                Err(error) => {
+                    // TODO: Track an outcome.
```
jjbayer, line 1046: Are we OK with not tracking outcomes here?

jjbayer (approved):
Looks good, but please do see comments, especially about outcomes.

relay-monitors/src/lib.rs
jan-auer: No, I intentionally removed that because we always need to normalize the status enum.

relay-monitors/src/lib.rs
jan-auer: Not necessarily, that depends on the kind of normalization we're doing in external Relays. It may be safer to keep it as "Unknown" rather than applying the wrong normalization (same to the metrics extraction version).

relay-server/src/actors/processor.rs
jan-auer: Yes, the backend also doesn't track outcomes yet, and it's not yet clear whether we want to track per check-in update or only per check-in.

relay-monitors/src/lib.rs
jjbayer: Right, because the deserializer maps to `Unknown`, gotcha. You could even remove the `Missed` variant then, though it also makes sense to keep that explicit.

PR #1887, opened 2023-02-27
ref(project): Skip serializing default fields

jjbayer:
Some fields that are not being sent by Sentry when they are empty are serialized by Relay. Align with Sentry and do not serialize these fields if they have default values.

**Why though?**

* The main reason is that https://github.com/getsentry/relay/pull/1860 introduced a validation function that can be used Sentry-side to check if Sentry accidentally produces any fields that Relay cannot use. This has already surfaced some inconsistencies (see https://github.com/getsentry/sentry/pull/45124), but it requires that Sentry and Relay produce the same fields.
* Reduced payload size when sending configs to downstream Relays.

### DIFF
--- a/relay-general/src/pii/config.rs
+++ b/relay-general/src/pii/config.rs
@@ -12,6 +12,12 @@ use crate::processor::SelectorSpec;
 
 const COMPILED_PATTERN_MAX_SIZE: usize = 262_144;
 
+/// Helper method to check whether a flag is false.
+#[allow(clippy::trivially_copy_pass_by_ref)]
+pub(crate) fn is_flag_default(flag: &bool) -> bool {
+    !*flag
+}
+
 #[derive(Clone, Debug, thiserror::Error)]
 pub enum PiiConfigError {
     #[error("could not parse pattern")]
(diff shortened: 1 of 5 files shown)


relay-general/src/pii/config.rs
```diff
@@ -121,7 +127,7 @@ pub struct MultipleRule {
     /// A reference to other rules to apply
     pub rules: Vec<String>,
     /// When set to true, the outer rule is reported.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "is_flag_default")]
```
jjbayer, line 130: `#[serde(default)]` has been here for 4 years, so every Relay should be able to handle the field's absence. I checked the age of the `default` functionality for the other fields as well.

olksdr (approved):
lgtm

relay-general/src/pii/config.rs
```diff
@@ -199,23 +205,29 @@ pub struct RuleSpec {
...
 pub struct Vars {
     /// The default secret key for hashing operations.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub hash_key: Option<String>,
 }
 
+impl Vars {
+    fn is_empty(&self) -> bool {
+        self.hash_key.is_none()
```
iker-barriocanal, line 214: Should we also consider post-trimmed empty strings as empty vars? In the tests below (from the diff), I only see `hash_key` being `null`.

iker-barriocanal (approved):

relay-general/src/pii/config.rs
jjbayer: You might have more context than me here, when would `hash_key` be an empty string?

Issue #1855, opened 2023-02-15
Apply default scurbbing to span and breadcrumb data


cleptric:
We started to ship updates in the PHP SDKs and will ship Python/Ruby soon, where we removed sensitive data from `http.client` span descriptions and moved them into the span data instead. 

This was decided in this [RFC](https://github.com/getsentry/rfcs/blob/main/text/0038-scrubbing-sensitive-data.md).
https://github.com/getsentry/relay/pull/1593 reverted this behavior, as the scrubbing of the span description introduced a regression.

We would now need to scrub the values of `span.data.http.query` and `span.data.http.fragment` as well as `breadcrumb.data.http.query` and `breadcrumb.data.http.fragment` by default in Relay.

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/relay/pull/1953
- [ ] https://github.com/getsentry/relay/pull/1955
```

cleptric:
Related to https://github.com/getsentry/team-webplatform-meta/issues/2

iker-barriocanal:
```[tasklist]
# Fields to scrub
- [ ] `span.data.http.query`
- [ ] `span.data.http.fragment`
- [ ] `breadcrumb.data.http.query`
- [ ] `breadcrumb.data.http.fragment`
```

PR #1889, opened 2023-02-28
feat(scrubbing): Scrub `span.data.http.query` with default scrubbers

iker-barriocanal:
Ref: https://github.com/getsentry/relay/issues/1855

`span.data.http.query` may contain sensitive data, which is currently not scrubbed. This PR makes relay to scrub that field with default data scrubbers.

### DIFF
--- a/relay-general/src/protocol/span.rs
+++ b/relay-general/src/protocol/span.rs
@@ -1,4 +1,6 @@
-use crate::protocol::{JsonLenientString, OperationType, SpanId, SpanStatus, Timestamp, TraceId};
+use crate::protocol::{
+    DataElement, JsonLenientString, OperationType, SpanId, SpanStatus, Timestamp, TraceId,
+};
 use crate::types::{Annotated, Object, Value};
 
 #[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
@@ -45,7 +47,7 @@ pub struct Span {
 
     /// Arbitrary additional data on a span, like `extra` on the top-level event.
     #[metastructure(pii = "maybe")]
-    pub data: Annotated<Object<Value>>,
+    pub data: Annotated<DataElement>,
 
     // TODO remove retain when the api stabilizes
     /// Additional arbitrary fields for forwards compatibility.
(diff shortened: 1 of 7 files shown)


cleptric (approved):
LGTM, we may add an additional test case that checks a more complex query, like `api_token=123&process_id=123` -> `api_token=[Filtered]&process_id=123`

relay-general/src/protocol/types.rs
```diff
@@ -1117,6 +1117,26 @@ impl schemars::JsonSchema for Timestamp {
...
+
+    #[metastructure(additional_properties, retain = "true", pii = "maybe")]
+    pub other: Object<Value>,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct HttpElement {
+    #[metastructure(pii = "true")]
+    pub query: Annotated<String>,
```
jjbayer: `span.data` is [documented as free-form](https://develop.sentry.dev/sdk/event-payloads/span/), so when a user sends `{"http": {"query": {"not": "a string"}}}`, we should still accept it.

@iker-barriocanal could you check if the scrubbing still works when you change `String` to `Value` here?

jjbayer (changes requested):

relay-general/src/protocol/types.rs
iker-barriocanal: Great catch! This is fixed in 03af9eb46153a99ecada9db886da4c6dab5ffbaf, and continues to work as expected.

relay-general/src/protocol/span.rs
```diff
@@ -70,7 +74,12 @@ mod tests {
   "op": "operation",
   "span_id": "fa90fdead5f74052",
   "trace_id": "4c79f60c11214eb38604f4ae0781bfb2",
-  "status": "ok"
+  "status": "ok",
+  "data": {
+    "http": {
+      "query": "is_sample_query=true"
+    }
+  }
```
jjbayer, line 89: Should we add some random extra fields in `data` and `data.http` here to validate that arbitrary data is still accepted?

jjbayer (approved):

PR #1890, opened 2023-02-28
chore(sentry): Add environment config option

### LINKED ISSUE #996: Add sentry environment to sentry config section
We are using sentry relay in staging and production environments and would like to have one sentry project for both of them.

This could be done by adding an `environment` field to the sentry config section:

```
    relay:
      mode: proxy
      port: 8080
    sentry: 
      enabled: true
      dsn: https://...@....ingest.sentry.io/...
      environment: staging
```

Then we could switch between environments with sentry website UI:

Screen Shot 2021-05-17 at 17 25 06

olksdr:
This adds the `environment` configuration option, which also earlier could be set with env variable `SENTRY_ENVIRONMENT` to make this a bit more explicit, which environment must be reported to Sentry. 

fixes https://github.com/getsentry/relay/issues/996

#skip-changelog

### DIFF
--- a/relay-log/src/setup.rs
+++ b/relay-log/src/setup.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::env;
 use std::io::{self, Write};
 use std::path::PathBuf;
@@ -84,6 +85,9 @@ pub struct SentryConfig {
     /// Enables reporting to Sentry.
     pub enabled: bool,
 
+    /// Sets the environment for this service.
+    pub environment: Option<Cow<'static, str>>,
+
     /// Internal. Enables crash handling and sets the absolute path to where minidumps should be
     /// cached on disk. The path is created if it doesn't exist. Path must be UTF-8.
     pub _crash_db: Option<PathBuf>,
@@ -103,6 +107,7 @@ impl Default for SentryConfig {
                 .parse()
                 .ok(),
             enabled: false,
+            environment: None,
             _crash_db: None,
         }
     }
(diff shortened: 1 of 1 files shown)


relay-log/src/setup.rs
```diff
@@ -84,6 +85,9 @@ pub struct SentryConfig {
     /// Enables reporting to Sentry.
     pub enabled: bool,
 
+    /// Sets the environment for this service.
+    pub environment: Option<Cow<'static, str>>,
```
jjbayer, line 89: Why is this a `Cow`? Is there a use case where we generate `SentryConfig` with a static string?

relay-log/src/setup.rs
```diff
@@ -176,6 +181,7 @@ pub fn init(config: &LogConfig, sentry: &SentryConfig) {
             in_app_include: vec!["relay"],
             release: Some(RELEASE.into()),
             attach_stacktrace: config.enable_backtraces,
+            environment: sentry.environment.clone(),
```
jjbayer, line 184: Probably more a question for the SDK, but what happens when both `SENTRY_ENVIRONMENT` and `environment` are set?

jjbayer (approved):
Two questions, but they are not blockers.

relay-log/src/setup.rs
olksdr: This will behave like `sentry-rust` SDK behaves,  and [according to this part](https://github.com/getsentry/sentry-rust/blob/5be18e5b3a5379b9ca45577bef1c814591ef3c87/sentry/src/defaults.rs#L92) the env will be used when there is nothing set in the config options.

relay-log/src/setup.rs
olksdr: That's basically because `sentry-rust` has this type for environment in `ClientOptions` and I did not want to mess with it. 

Do you think we should change to something else? What would you propose?

jjbayer: OK, that makes sense, all good then. I would have expected a `String` here and wondered where the `Cow` came from. I guess with `Cow` you could even deserialize without owning, although we don't do that for configs AFAIK.

PR #1892, opened 2023-02-28
feat(general): Fix leaked ip address

### LINKED ISSUE #1740: If the user ip address contains an invalid IP address, it will get leaked in the redaction error message
If the user IP address contains an invalid IP address as seen here:

https://github.com/getsentry/relay/blob/156c2ae42bb2fd21430636d613b83d597c45710f/relay-general/src/pii/convert.rs#L450

then in the redaction it will contain the entire value:

https://github.com/getsentry/relay/blob/156c2ae42bb2fd21430636d613b83d597c45710f/relay-general/src/pii/snapshots/relay_general__pii__convert__tests__ip_stripped.snap#L92

If it contained an incorrectly formatted IP address it would get leaked in this redaction.

TBS1996:
#1740 

before this, if an annotated object failed to be deserialized, the "original value" would be saved in the meta, where it wouldn't get scrubbed. This means that sensitive data could potentially be leaked in the meta. 

This PR will run pii scrubbing also in the `original_value` meta-field, in order to stop sensitive data from getting leaked as in the example shown in the linked issue. 

to do this we made `meta`  optional in `fn apply_all_rules` as `original_value` doesnt have a meta (since its part of a meta itself).


jjbayer:
> always scrub the val in meta for the scrubbed field

Let's try to implement the first option, as discussed with @jan-auer .

relay-general/src/pii/snapshots/relay_general__pii__processor__tests__scrub_original_value.snap
```diff
@@ -0,0 +1,180 @@
...
+                                13,
+                            ),
+                        ),
+                    },
+                ],
+                errors: [],
+                original_length: Some(
+                    22,
+                ),
+                original_value: None,
```
TBS1996, line 68: previously the original value would be leaked here. now since IpAddr is set to scrub anything it gets removed completely. if it's a value that is set to only be scrubbed partially you might see something like `" [ip]"`

relay-general/src/processor/chunks.rs
```diff
@@ -157,19 +157,28 @@ where
 
 /// Splits the string into chunks, maps each chunk and then joins chunks again, emitting
 /// remarks along the process.
-pub fn process_chunked_value<F>(value: &mut String, meta: &mut Meta, f: F)
+pub fn process_chunked_value<F>(value: &mut String, meta: &mut Option<&mut Meta>, f: F)
```
TBS1996: I would particularly like some input on this function

TBS1996:
I'm also looking for more ways to unit tests. Is there an object which can fail to be deserialized (meaning it's not a string..) and it also has pii scrubbing?

relay-general/src/pii/processor.rs
```diff
@@ -59,6 +59,15 @@ impl<'a> Processor for PiiProcessor<'a> {
         meta: &mut Meta,
         state: &ProcessingState<'_>,
     ) -> ProcessingResult {
+        if let Some(crate::types::Value::String(original_value)) = meta.original_value_as_mut() {
+            if self
+                .apply_all_rules(None, state, Some(original_value))
+                .is_err()
```
TBS1996: this implies the value should be hard-deleted. in that case it's safest to simply delete `original_value`.

relay-general/src/pii/processor.rs
```diff
@@ -59,6 +59,15 @@ impl<'a> Processor for PiiProcessor<'a> {
         meta: &mut Meta,
         state: &ProcessingState<'_>,
     ) -> ProcessingResult {
+        if let Some(crate::types::Value::String(original_value)) = meta.original_value_as_mut() {
```
jjbayer: Please import `Value`.

relay-general/src/processor/chunks.rs
jjbayer: For simplicity's sake, I would leave this function as it was, and call it with an empty `Meta::default()` instead. We could even do this on the highest level, i.e. call `apply_all_rules` with a `Meta::default()` instead of `None`.

relay-general/src/pii/processor.rs
```diff
@@ -43,7 +43,7 @@ impl<'a> PiiProcessor<'a> {
                 #[allow(clippy::needless_option_as_deref)]
                 for rule in rules {
                     let reborrowed_value = value.as_deref_mut();
-                    apply_rule_to_value(meta, rule, state.path().key(), reborrowed_value)?;
+                    apply_rule_to_value(&mut meta, rule, state.path().key(), reborrowed_value)?;
```
jjbayer: In case we keep the optional: It should be possible to pass an `Option<&mut Meta>` here by doing the same as for `value`, that is `let reborrowed_meta = meta.as_deref_mut();`

relay-general/src/pii/processor.rs
```diff
@@ -59,6 +59,15 @@ impl<'a> Processor for PiiProcessor<'a> {
         meta: &mut Meta,
         state: &ProcessingState<'_>,
     ) -> ProcessingResult {
+        if let Some(crate::types::Value::String(original_value)) = meta.original_value_as_mut() {
+            if self
```
jjbayer:
```suggestion
            // Also apply pii scrubbing to the original value (set by normalization or other processors),
            // such that we do not leak sensitive data through meta.
            if self
```

relay-general/src/pii/processor.rs
```diff
@@ -373,9 +388,58 @@ mod tests {
...
+            ..Default::default()
+        };
+
+        let pii_config = to_pii_config(&scrubbing_config).unwrap();
+        let mut pii_processor = PiiProcessor::new(pii_config.compiled());
+
+        process_value(&mut data, &mut pii_processor, ProcessingState::root()).unwrap();
+
+        assert_debug_snapshot!(&data);
+    }
```
jjbayer, line 439: Please also add a test where the the original value is a credit card number, so we can see if it is properly scrubbed.

jjbayer (changes requested):
Looks good, just one suggestion for simplification.

> Is there an object which can fail to be deserialized (meaning it's not a string..) and it also has pii scrubbing?

Have you tried putting a number in the `ip_addr` field?

relay-general/src/pii/processor.rs
```diff
@@ -12,6 +12,7 @@ use crate::processor::{
     process_chunked_value, Chunk, Pii, ProcessValue, ProcessingState, Processor, ValueType,
 };
 use crate::protocol::{AsPair, IpAddr, NativeImagePath, PairList, Replay, User};
+use crate::types::Value;
```
jjbayer: nit: You can merge this import with the line below.

relay-general/src/pii/processor.rs
```diff
@@ -59,6 +60,32 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+
+        if let Some(Value::String(original_value)) = meta.original_value_as_mut() {
+            if let Some(parent) = state.iter().next() {
+                let attrs = FieldAttrs {
+                    pii: Pii::True,
+                    ..Default::default()
+                };
+
+                let new_state = parent.enter_static(
+                    "",
```
TBS1996: should i put something in the key variable?

relay-general/src/pii/processor.rs
```diff
@@ -59,6 +60,32 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                    Some(Cow::Owned(attrs)),
+                    enumset::enum_set!(ValueType::String),
+                );
```
jjbayer:
```suggestion
                let path = state.path();
                let new_state = parent.enter_borrowed(
                    path.key().unwrap_or(""),
                    Some(Cow::Borrowed(state.attrs())),
                    enumset::enum_set!(ValueType::String),
                );
```

relay-general/src/pii/processor.rs
jjbayer: Yes, the name of the current field (see suggestion below). Without that, name-based rules will not match.

relay-general/src/pii/processor.rs
```diff
@@ -59,6 +59,28 @@ impl<'a> Processor for PiiProcessor<'a> {
         meta: &mut Meta,
         state: &ProcessingState<'_>,
     ) -> ProcessingResult {
+        // Also apply pii scrubbing to the original value (set by normalization or other processors),
+        // such that we do not leak sensitive data through meta. Deletes `original_value` if an Error
+        // value is returned.
+
+        if let Some(Value::String(original_value)) = meta.original_value_as_mut() {
```
jjbayer: nit

```suggestion
        if let Some(Value::String(original_value)) = meta.original_value_as_mut() {
            // Also apply pii scrubbing to the original value (set by normalization or other processors),
            // such that we do not leak sensitive data through meta. Deletes `original_value` if an Error
            // value is returned.
```

jjbayer (commented):
Nice! Please add a test that proves that the repro case in https://github.com/getsentry/relay/issues/1740#issuecomment-1416003596 is now fixed.

relay-general/src/pii/processor.rs
```diff
@@ -59,6 +59,27 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                    path.key().unwrap_or(""),
+                    Some(Cow::Borrowed(state.attrs())),
+                    enumset::enum_set!(ValueType::String),
+                );
+
+                if self
+                    .apply_all_rules(&mut Meta::default(), &new_state, Some(original_value))
+                    .is_err()
+                {
+                    meta.set_original_value(Option::<String>::None);
```
jjbayer, line 79:
```suggestion
                    // `apply_all_rules` returned `DeleteValueHard` or `DeleteValueSoft`, so delete the original as well.
                    meta.set_original_value(Option::<String>::None);
```

jjbayer (approved):

PR #1894, opened 2023-02-28
feat(profiling): Remove idle samples at the start and end of a profile and useless metadata

phacops:
SDKs might send threads with only idle samples and some threads might contains a lot of idle samples at the beginning or end because they're background threads and only execute a little bit of work.

We want to remove all consecutive idle samples from the beginning of a profile to the first real sample and from the last real sample to the end. At the same time, threads with only idle samples will be cleaned up.

Additionally, we're removing all thread and queue metadata without a sample.


relay-profiling/src/sample.rs
```diff
@@ -226,6 +226,52 @@ impl SampleProfile {
...
+                }
+                None => true,
+            });
+    }
+
+    fn cleanup_thread_metadata(&mut self) {
+        if let Some(thread_metadata) = &mut self.profile.thread_metadata {
+            let mut thread_ids: HashSet<String> = HashSet::new();
+            for sample in &self.profile.samples {
+                thread_ids.insert(sample.thread_id.to_string());
```
jjbayer, line 254: It's unfortunate that we have to convert to string here... could we change `thread_metadata` to contain `u64` thread IDs instead? Or would that fail parsing certain payloads?

relay-profiling/src/sample.rs
```diff
@@ -226,6 +226,52 @@ impl SampleProfile {
         self.profile
             .strip_pointer_authentication_code(&self.platform, &self.device.architecture);
     }
+
+    fn remove_idle_samples_at_the_edge(&mut self) {
+        self.remove_idle_samples_at_the_start();
+        self.profile.samples.reverse();
+        self.remove_idle_samples_at_the_start();
+        self.profile.samples.reverse();
```
jjbayer: How many samples are usually in a profile? Instead of reversing the vector in place, could you make this work with a reverse iterator? Unfortunately there is no `retain_reverse` method.

jjbayer (changes requested):

relay-profiling/src/sample.rs
phacops: We did this because JSON allows only keys to be strings, not integer. Plus, `u64` doesn't exist in the JSON spec so some platforms will send us `u64` like thread IDs as strings anyway.

Now, we may be able to actually parse the key as `u64` the same way we parse a `string` or `u64` into a `u64` elsewhere but it'll be a problem downstream so I'll have to adjust that first.

I would like to keep that as is in the PR thought and see if I can refactor in `vroom` to support both (which is doable) and then change it here.

relay-profiling/src/sample.rs
phacops: There's a sample every 10ms and I'd say the average duration of a profile is less than 1s. Max duration is 30s. So we could have an average of 100 and max 3000 samples per thread. Usually we see 10 threads in a `cocoa` profile for example. `python`, `rust` and `node` don't have as many though.

I'm trying another approach by finding the position and then using `drain`. Maybe you'll like that one better?

relay-profiling/src/sample.rs
```diff
@@ -228,27 +228,27 @@ impl SampleProfile {
...
-                    if stack.is_empty() && !thread_ids.contains(&sample.thread_id) {
-                        false
-                    } else {
-                        thread_ids.insert(sample.thread_id);
-                        true
-                    }
-                }
-                None => true,
-            });
+        if let Some(end) = self.profile.samples.iter().rev().position(|sample| {
```
jjbayer: As far as I know, `.rev().position()` gives you the offset from the end, so I think we need `.rposition()` here.

```suggestion
        if let Some(end) = self.profile.samples.iter().rposition(|sample| {
```

Example:

```rust
let xs = vec!['a', 'b', 'c', 'd'];
println!("{:?}", xs.iter().rev().position(|item| *item == 'c'));
println!("{:?}", xs.iter().rposition(|item| *item == 'c'));

// prints
Some(1)
Some(2)
```

You can verify this by removing the last sample from your test case -- then `end` will be `2`.

relay-profiling/src/sample.rs
```diff
@@ -226,6 +226,52 @@ impl SampleProfile {
...
+
+        if let Some(end) = self.profile.samples.iter().rev().position(|sample| {
+            match self.profile.stacks.get(sample.stack_id) {
+                Some(stack) => !stack.is_empty(),
+                None => false,
+            }
+        }) {
+            if end > 0 {
+                self.profile.samples.drain(end..);
+            }
```
jjbayer:
```suggestion
            self.profile.samples.truncate(end + 1); // With the suggestion above, `end` is the index of the last element we want to keep
```

jjbayer (changes requested):
Thanks for iterating on this! Apart from the indexing bug this looks good to go.

jjbayer (approved):

Zylphrex (approved):

PR #1895, opened 2023-02-28
feat(mobile): Synthesize device.class based on specs from device context

edwardgou-sentry:
Adds and uses `normalize_device_context` which creates a new device.class context value by checking a combination of the device family, processor_frequency, processor_count, and memory_size.

The `device.class` context value will be used in the sentry issues, performance, and discover products for the device classification project (https://github.com/getsentry/sentry/issues/44449).


relay-general/src/store/normalize/contexts.rs
```diff
@@ -199,11 +204,45 @@ fn normalize_response(response: &mut ResponseContext) {
...
+            {
+                device.class = Annotated::new(DEVICE_CLASS_LOW);
+            } else if device.processor_frequency.value().unwrap() < &2500
+                || device.memory_size.value().unwrap() < &(6 * GIB)
+            {
+                device.class = Annotated::new(DEVICE_CLASS_MEDIUM);
+            } else {
+                device.class = Annotated::new(DEVICE_CLASS_HIGH);
+            }
+        }
```
jan-auer: You can avoid these unwraps with an `if let` statement similar to the one above:

```suggestion
        } else if let (Some(&freq), Some(&proc), Some(&mem)) = (
            device.processor_frequency.value(),
            device.processor_count.value(),
            device.memory_size.value(),
        ) {
            if freq < 2000 || mem < 4 * GIB || proc < 8 {
                device.class = Annotated::new(DEVICE_CLASS_LOW);
            } else if freq < 2500 || mem < 6 * GIB {
                device.class = Annotated::new(DEVICE_CLASS_MEDIUM);
            } else {
                device.class = Annotated::new(DEVICE_CLASS_HIGH);
            }
        }
```

relay-general/src/protocol/contexts/device.rs
```diff
@@ -160,6 +160,10 @@ pub struct DeviceContext {
     /// Whether location support is available on the device.
     pub supports_location_service: Annotated<bool>,
 
+    // The performance class of the device, stored as a number.
+    // This value is synthesized from the device's specs in normalize_device_context.
+    pub class: Annotated<u64>,
```
jan-auer: Since the device class is an enumeration with semantic values, can we make it an `enum` instead? For instance, what should happen when an SDK sends the value 42? 

If you want to allow any number, you can introduce a so-called "newtype":

```rust
#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq Serialize)]
struct DeviceClass(pub u64);

impl DeviceClass {
    const LOW: Self = Self(1);
    const MEDIUM: Self = Self(2);
    const HIGH: Self = Self(3);
}

// use as `DeviceClass::LOW`
```

jan-auer (commented):
Thanks for the contribution!

relay-general/src/store/normalize/contexts.rs
edwardgou-sentry: 👍 Thanks Jan, updated to if let!

relay-general/src/protocol/contexts/device.rs
edwardgou-sentry: Hey Jan, I went with the newtype that you suggested!
We have plans to support more device class values in the future and theres use in being able to compare the number values so I think this works best for us.

jan-auer (approved):
This will allow clients to send unknown device class values, which we will store without modification. If this is OK for the rest of the product, this PR can be merged.

PR #1896, opened 2023-02-28
feat(threads): Add state as an optional key in the threads interface

### LINKED ISSUE #45196: Extract thread state in the threads interface in relay
Thread state is currently sent by the Android SDK but isn't extracted. We want to extract and store it on the event json. Currently, we capture the JVM [states](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.State.html), but in the ANRv2 implementation, we expect to capture to states the Android VM [provides](https://cs.android.com/android/platform/superproject/+/master:art/runtime/thread_state.h). Note, these values are for Android only, other SDKs could send other values.

We only need to save thread state in nodestore, we don't anticipate running any queries against this field.

shruthilayaj:
Thread state is an optional string currently sent by the Android SDK on the threads interface but isn't extracted. We want to extract and store it on the event json. Currently, we capture the JVM [states](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.State.html), but in the ANRv2 implementation, we expect to capture to states the Android VM [provides](https://cs.android.com/android/platform/superproject/+/master:art/runtime/thread_state.h). Note, these values are for Android only, other SDKs could send other values.

We only need to save thread state in nodestore, we don't anticipate running any queries against this field.

[Here](https://gist.github.com/shruthilayaj/2151eac05009466f4549c5f2ad48a11f) is an example event payload with thread states.

closes https://github.com/getsentry/sentry/issues/45196

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Protocol validation for source map image type. ([#1869](https://github.com/getsentry/relay/pull/1869))
 - Strip quotes from client hint values. ([#1874](https://github.com/getsentry/relay/pull/1874))
 - Add Dotnet, Javascript and PHP support for profiling. ([#1871](https://github.com/getsentry/relay/pull/1871), [#1876](https://github.com/getsentry/relay/pull/1876), [#1885](https://github.com/getsentry/relay/pull/1885))
+- Add `thread.state` field to protocol. ([#1896](https://github.com/getsentry/relay/pull/1896))
 
 **Bug Fixes**:
 
--- a/relay-general/src/protocol/thread.rs
+++ b/relay-general/src/protocol/thread.rs
@@ -125,6 +125,9 @@ pub struct Thread {
     /// A flag indicating whether the thread was responsible for rendering the user interface.
     pub main: Annotated<bool>,
 
+    #[metastructure(skip_serialization = "empty")]
+    pub state: Annotated<String>,
+
     /// Additional arbitrary fields for forwards compatibility.
     #[metastructure(additional_properties)]
     pub other: Object<Value>,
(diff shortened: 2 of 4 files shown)


iker-barriocanal (commented):
LGTM.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Strip quotes from client hint values. ([#1874](https://github.com/getsentry/relay/pull/1874))
 - Add Dotnet, Javascript and PHP support for profiling. ([#1871](https://github.com/getsentry/relay/pull/1871), [#1876](https://github.com/getsentry/relay/pull/1876), [#1885](https://github.com/getsentry/relay/pull/1885))
 - Scrub `span.data.http.query` with default scrubbers. ([#1889](https://github.com/getsentry/relay/pull/1889))
+- Add `thread.state` field to protocol. ([#1896](https://github.com/getsentry/relay/pull/1896))
```
iker-barriocanal, line 13: Please also add an entry to `py/CHANGELOG.md`.

iker-barriocanal (approved):

relay-general/src/protocol/thread.rs
```diff
@@ -125,6 +125,10 @@ pub struct Thread {
     /// A flag indicating whether the thread was responsible for rendering the user interface.
     pub main: Annotated<bool>,
 
+    /// Thread state at the time of the crash.
+    #[metastructure(skip_serialization = "empty")]
+    pub state: Annotated<String>,
```
iker-barriocanal, line 130: Now that I'm thinking about it, it's probably best to make this an `enum` instead. From your PR description, I assume you're focusing mostly on Android for now, and you link to two places where there's a list of possible values. We can start from these.

iker-barriocanal (changes requested):

iker-barriocanal (approved):
Approving as `String` because only Android has 38 different options, and I don't think it's worth the effort of typing every case at the moment.

PR #1593, opened 2022-11-18
Revert "fix(pii): scrub PII in span description [INGEST-1709]"

olksdr:
Reverts getsentry/relay#1590

#skip-changelog

PR #1901, opened 2023-03-01
feat(scrubbing): Scrub `breadcrumb.data.http.query` & `data.other` with default scrubbers

iker-barriocanal:
Ref: https://github.com/getsentry/relay/issues/1855

`breadcrumb.data.http.query` may contain sensitive data, which is currently not scrubbed. This PR makes relay to scrub that field with default data scrubbers.

`data.other` is also scrubbed to be consistent with the current IP address scrubbing in breadcrumb's data. This should not cause any problems with https://github.com/getsentry/relay/pull/1593, where the biggest issue was scrubbing span descriptions (which they are still not scrubbed).


relay-general/src/protocol/types.rs
```diff
@@ -1123,7 +1123,7 @@ pub struct DataElement {
     #[metastructure(pii = "maybe")]
     pub http: Annotated<HttpElement>,
 
-    #[metastructure(additional_properties, retain = "true", pii = "maybe")]
+    #[metastructure(additional_properties, retain = "true", pii = "true")]
```
iker-barriocanal, line 1129: This change doesn't seem to affect the changes made to `Span`, and is motivated by this test (which fails otherwise):

https://github.com/getsentry/relay/blob/00b83a3e9a77a5e407cdfdf8a4822151b953d750/relay-general/src/pii/convert.rs#L438

iker-barriocanal: The `data` property in `Breadcrumb` is also marked as `pii == true`.

jjbayer: If `breadcrumb.data` was already `pii = true`, why do we even need to change its type? Does `pii = true` not descend into the `Value`?

iker-barriocanal: I haven't dove into the code, but apparently, it does not descend -- the test linked above would have passed on the first attempt. Maybe it acts the other way around, where `pii = false` means no `pii` checks on child elements? I'd have to check the code; but I don't think that's an issue.

jjbayer: I modified the "span" test, I think we now scrub too much in `span.data`: https://github.com/getsentry/relay/blob/112580138f9ceb4875e2137956962dc8978e9d5c/relay-general/src/pii/processor.rs#L1037

An easy way around this would be to have separate `SpanData` and `BreadcrumbData` structs, instead of a common `DataElement`.

jjbayer: With this change

1. `breadcrumb.data.http.query` is scrubbed by default
2. `breadcrumb.data.whatever` is scrubbed by default
3. `breadcrumb.data.http.whatever` is _not_ scrubbed by default

Is this really the behavior we want? Because if we also want the 3rd case to be scrubbed by default (ie. scrub everything in `data`), we might as well change the type of `breadcrumb.data` back to `Value`. `test_breadcrumb_data_object_is_scrubbed` and `test_breadcrumb_data_string_is_scrubbed` seem to pass just fine with that simplified schema.

jjbayer (changes requested):

iker-barriocanal:
Closing, see https://github.com/getsentry/relay/issues/1915#issuecomment-1465796416.

PR #1905, opened 2023-03-03
ref(monitors): Use monitor_slug over monitor_id

evanpurkhiser:
We will be preferring slug over GUIDs, since it is likely we'll remove (or at least hide) them in the future

#skip-changelog

### DIFF
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -59,8 +59,7 @@ struct CheckIn {
     check_in_id: Uuid,
 
     /// Identifier of the monitor for this check-in.
-    #[serde(serialize_with = "uuid_simple")]
-    monitor_id: Uuid,
+    monitor_slug: String,
 
     /// Status of this check-in. Defaults to `"unknown"`.
     status: CheckInStatus,
@@ -91,7 +90,7 @@ mod tests {
     fn test_json_roundtrip() {
         let json = r#"{
   "check_in_id": "a460c25ff2554577b920fcfacae4e5eb",
-  "monitor_id": "4dc8556e039245c7bd569f8cf513ea42",
+  "monitor_slug": "my-monitor",
   "status": "in_progress",
   "duration": 21.0
 }"#;
--- a/tests/integration/test_crons.py
+++ b/tests/integration/test_crons.py
@@ -1,7 +1,7 @@
 def generate_check_in():
     return {
         "check_in_id": "a460c25ff2554577b920fcfacae4e5eb",
-        "monitor_id": "4dc8556e039245c7bd569f8cf513ea42",
+        "monitor_slug": "my-monitor",
         "status": "in_progress",
         "duration": 21.0,
     }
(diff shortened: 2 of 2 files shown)


relay-monitors/src/lib.rs
```diff
@@ -59,8 +59,7 @@ struct CheckIn {
     check_in_id: Uuid,
 
     /// Identifier of the monitor for this check-in.
-    #[serde(serialize_with = "uuid_simple")]
-    monitor_id: Uuid,
+    monitor_slug: String,
```
iker-barriocanal, line 62: Do we need to consider backward compatibility?

iker-barriocanal, line 62: Is the docstring still up-to-date?

tests/integration/test_crons.py
```diff
@@ -1,7 +1,7 @@
 def generate_check_in():
     return {
         "check_in_id": "a460c25ff2554577b920fcfacae4e5eb",
-        "monitor_id": "4dc8556e039245c7bd569f8cf513ea42",
+        "monitor_id": "monitor-id",
```
iker-barriocanal:
```suggestion
        "monitor_slug": "monitor-id",
```
Same in the assertion below.

relay-monitors/src/lib.rs
jan-auer: No, the SDK has not been released and we can break the API for crons currently.

jan-auer (commented):
@evanpurkhiser, since you're using the monitor slug in URLs too, please consider: 
- whether you want to restrict the allowed character set for the slug, especially if it's used in URLs. We do this for releases and environments, e.g. in sessions [here](https://github.com/getsentry/relay/blob/f7ff9f0303a4e48720b63ecf7a9a4413be1f973d/relay-server/src/actors/processor.rs#L572-L575). 
- limit the length, especially if there's a limit on the database column. We can either truncate or reject. This would fit in `process_check_in`.
- reject payloads where the slug is an empty string in `process_check_in`.

relay-monitors/src/lib.rs
evanpurkhiser: Yes it is

evanpurkhiser:
> * whether you want to restrict the allowed character set for the slug, especially if it's used in URLs. We do this for releases and environments, e.g. in sessions [here](https://github.com/getsentry/relay/blob/f7ff9f0303a4e48720b63ecf7a9a4413be1f973d/relay-server/src/actors/processor.rs#L572-L575).
> * limit the length, especially if there's a limit on the database column. We can either truncate or reject. This would fit in process_check_in.

We're restricting this [on the sentry side](https://github.com/getsentry/sentry/blob/11e54749a85d07127b3bc593d9a41575ecdc8dab/src/sentry/api/validators/monitor.py#L92) right now, but I assume we'll want to do the same validation within relay as well correct?

jan-auer:
> I assume we'll want to do the same validation within relay

Yes, I'd propose to do that. The linked validator doesn't run in the ingest path, and if we move this kind of validation to Relay we can react in more ways, such as normalizing them. That can be important if we also ingest other data (e.g. attachment) that require the same normalization to be applied.

Same for the maximum length. Maybe it's better to truncate rather than reject?

jan-auer (approved):
G2G, validation can come in a follow up.

PR #1906, opened 2023-03-06
Revert "ref(cache): Move buffering of pending envelope to ProjectCache"

olksdr:
Reverts getsentry/relay#1879

Introduced regression: [RELAY-2NFR
](https://sentry.my.sentry.io/organizations/sentry/issues/393031/events/7dae7f8e304e456ea5355a2169f3a075/?project=4)

The project key on the ProjectState in only few cases was missing, when the majority of messages went through ok.  

#skip-changelog


jjbayer (approved):
Could you update the description with the reason why we're reverting (e.g. link to sentry issue).

Note there's also the `Trigger: revert` label to automatically revert and skip CI.

PR #1907, opened 2023-03-07
ref(cache): Move buffering of pending envelope to ProjectCache

olksdr:
Take two as an attempt to fix the problem we had while deploying https://github.com/getsentry/relay/pull/1879 which was reverted in #1906 

----

These changes moving buffering of the incoming envelopes in the ProjectCache.

Current implementation still keeps, so called queue in memory and using HashMap with a composite key QueueKey {key, sampling_key}, where sampling_key can be the same as a key if there is no sampling project identified. The values to these keys are Vec of boxed Envelope with their EnvelopeContext.

Once we get an update for project state, we check all variants of QueueKey which contains the current ProjectKey and if all the project states are cached we try to flush buffered envelopes indexed by these QeueuKey.

The envelops will be buffered if:

the project state is not fetched yet
root project is here but the sampling project state is not fetched yet
the sampling project state is here but the root project is not fetched yet
This change also removes all the buffering from the Project and reduces its responsibility. Now it just keeps its own state and configuration and the envelope handling is done outside of it.

fix https://github.com/getsentry/team-ingest/issues/76

### DIFF
--- a/relay-server/src/actors/project_cache.rs
+++ b/relay-server/src/actors/project_cache.rs
@@ -633,45 +633,47 @@ impl ProjectCacheBroker {
     fn handle_processing(&mut self, envelope: Box<Envelope>, envelope_context: EnvelopeContext) {
         let project_key = envelope.meta().public_key();
 
-        // The `Envelope` and `EnvelopeContext` will be dropped if the `Project::check_envelope()`
-        // function returns any error, which will also be ignored here.
-        if let Some(project) = self.projects.get_mut(&project_key) {
-            if let Some(own_project_state) = project.valid_state().filter(|s| !s.invalid()) {
-                if let Ok(CheckedEnvelope {
-                    envelope: Some((envelope, envelope_context)),
-                    ..
-                }) = project.check_envelope(envelope, envelope_context)
-                {
-                    let sampling_state = utils::get_sampling_key(&envelope)
-                        .and_then(|key| self.projects.get(&key))
-                        .and_then(|p| p.valid_state());
-
-                    let mut process = ProcessEnvelope {
-                        envelope,
-                        envelope_context,
-                        project_state: own_project_state.clone(),
-                        sampling_project_state: None,
-                    };
-
-                    if let Some(sampling_state) = sampling_state {
-                        if own_project_state.organization_id == sampling_state.organization_id {
-                            process.sampling_project_state = Some(sampling_state)
-                        }
-                    }
-
-                    EnvelopeProcessor::from_registry().send(process);
-                }
-            } else {
-                relay_log::with_scope(
-                    |scope| scope.set_tag("project_key", project_key),
-                    || relay_log::error!("project has no valid cached state"),
-                );
-            }
-        } else {
+        let Some(project) = self.projects.get_mut(&project_key) else {
             relay_log::with_scope(
                 |scope| scope.set_tag("project_key", project_key),
                 || relay_log::error!("project could not be found in the cache"),
             );
+            return;
+        };
+
+        let Some(own_project_state) = project.valid_state().filter(|s| !s.invalid()) else {
+            relay_log::with_scope(
+                |scope| scope.set_tag("project_key", project_key),
+                || relay_log::error!("project has no valid cached state"),
+            );
+            return;
+        };
+
+        // The `Envelope` and `EnvelopeContext` will be dropped if the `Project::check_envelope()`
+        // function returns any error, which will also be ignored here.
+        if let Ok(CheckedEnvelope {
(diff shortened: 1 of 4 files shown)


relay-server/src/actors/project_cache.rs
```diff
@@ -468,7 +533,60 @@ impl ProjectCacheBroker {
...
+                    utils::get_sampling_key(&envelope)
+                        .and_then(|key| self.projects.get(&key))
+                        .and_then(|p| p.valid_state()),
+                )
+            } else {
+                (
+                    self.projects.get(&own_key).and_then(|p| p.valid_state()),
+                    Some(state.clone()),
+                )
+            };
```
olksdr: A new addition is here, made in [105b54c](https://github.com/getsentry/relay/pull/1907/commits/105b54cd46dfb225619f506a6ab2a3579a349d8d) - basically if my thinking is correct, we always have to use the own (root) project state for processing

This was not the case, and if the sampling project state update comes in first, after getting all the envelopes for it, we would use this dynamic project state in 2 places.

olksdr: \cc @jjbayer if you could, please, verify my thinking here.

jjbayer: Yes, good catch! I remember we fixed this for the project key but apparently we had a similar bug for the project state. We need more tests with `sampling_key != own_key`.

I think what would simplify this code and make it safer against confusion would be to change the signature of `handle_processing` to `fn handle_processing(envelope, envelope_context)`:

* `handle_processing` already does `self.projects.get_mut(project_key)` internally, So we don't gain much from passing `own_state` into it.
* and then we move `utils::get_sampling_key(&envelope).and_then(|key| self.projects.get(&key))` into `handle_processing` as well.

olksdr: Great suggestions!
Addressed this in [0074eb3](https://github.com/getsentry/relay/pull/1907/commits/0074eb372507a488b03907756eb48e003f1c47d4)

olksdr:
Refactor `handle_processing` a bit and also add more logging and docs in [`2389abe`](https://github.com/getsentry/relay/pull/1907/commits/2389abe3d4f278cccd453e16e6c6bdec56ec3fa9)

@jjbayer @jan-auer, please, take another look.

relay-server/src/actors/project_cache.rs
```diff
@@ -625,36 +623,55 @@ impl ProjectCacheBroker {
...
-        &mut self,
-        state: Arc<ProjectState>,
-        sampling_state: Option<Arc<ProjectState>>,
-        envelope: Box<Envelope>,
-        envelope_context: EnvelopeContext,
-    ) {
+    ///
+    /// The following pre-conditions must be met before calling this function:
+    /// - Envelope's project state must be cached and valid.
+    /// - Optional: if dynamic sampling keys exists, the dyanmic project state must be cached and valid.
```
```suggestion
    /// - If dynamic sampling keys exists, the sampling project state must be cached and valid.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -625,36 +623,55 @@ impl ProjectCacheBroker {
...
-                envelope,
-                envelope_context,
-                project_state: state.clone(),
-                sampling_project_state: None,
-            };
-
-            if let Some(sampling_state) = sampling_state {
-                if state.organization_id == sampling_state.organization_id {
-                    process.sampling_project_state = Some(sampling_state)
+        if let Some(project) = self.projects.get_mut(&project_key) {
```
jjbayer: Personally I would write this as a [guard](https://en.wikipedia.org/wiki/Guard_(computer_science)) to keep the error handling close to the check, i.e.

```rust
let Some(project) = self.projects.get_mut(&project_key) else {
  // log error
  return
};

let Some(own_project_state) = ... else {
   // log error
   return
}

if let Ok(CheckedEnvelope {...}) = ... {
    ...
}

jjbayer (approved):

relay-server/src/actors/project_cache.rs
olksdr: Yeah, this can look better. I will change.

olksdr: done [cdf587a](https://github.com/getsentry/relay/pull/1907/commits/cdf587abffa419bc995b3c86bb45e962c2237fc8)

relay-server/src/actors/project_cache.rs
olksdr: also done in [56f32b8](https://github.com/getsentry/relay/pull/1907/commits/56f32b8f089d0932fb8aa463814474fb79edac27)

PR #1909, opened 2023-03-08
ref(buffering): Move the buffering to the dedicated service

olksdr:
Refactor the queue, which was added in #1907, into the service, which only responsible for buffering (queueing, dequeueing, removing) the envelopes (and still envelope context) and communicated through the message passing from and to this service. 

related to https://github.com/getsentry/team-ingest/issues/78 
follow up for #1907 

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_buffer.rs
+++ b/relay-server/src/actors/project_buffer.rs
@@ -1,0 +1,181 @@
+use std::collections::{BTreeMap, BTreeSet};
+
+use tokio::sync::mpsc;
+
+use relay_common::ProjectKey;
+use relay_system::{FromMessage, Interface, Service};
+
+use crate::envelope::Envelope;
+use crate::utils::EnvelopeContext;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct QueueKey {
+    pub own_key: ProjectKey,
+    pub sampling_key: ProjectKey,
+}
+
+impl QueueKey {
+    pub fn new(own_key: ProjectKey, sampling_key: ProjectKey) -> Self {
+        Self {
+            own_key,
+            sampling_key,
+        }
+    }
+}
+
+/// Adds the envelope and the envelope context to the internal buffer.
+#[derive(Debug)]
+pub struct Enqueue {
+    key: QueueKey,
+    value: (Box<Envelope>, EnvelopeContext),
+}
+
+impl Enqueue {
+    pub fn new(key: QueueKey, value: (Box<Envelope>, EnvelopeContext)) -> Self {
+        Self { key, value }
+    }
+}
+
+/// Removes messages from the internal buffer and streams them to the sender.
+#[derive(Debug)]
+pub struct DequeueMany {
+    keys: Vec<QueueKey>,
+    sender: mpsc::UnboundedSender<(Box<Envelope>, EnvelopeContext)>,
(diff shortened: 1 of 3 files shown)


relay-server/src/actors/project_buffer.rs
```diff
@@ -0,0 +1,181 @@
...
+use tokio::sync::mpsc;
+
+use relay_common::ProjectKey;
+use relay_system::{FromMessage, Interface, Service};
+
+use crate::envelope::Envelope;
+use crate::utils::EnvelopeContext;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct QueueKey {
```
jan-auer, line 17: _nit_: Add a doc comment for this struct.

relay-server/src/actors/project_buffer.rs
```diff
@@ -0,0 +1,181 @@
...
+impl DequeueMany {
+    pub fn new(
+        keys: Vec<QueueKey>,
+        sender: mpsc::UnboundedSender<(Box<Envelope>, EnvelopeContext)>,
+    ) -> Self {
+        Self { keys, sender }
+    }
+}
+
+/// Removes the provided keys from the internal buffer.
```
jan-auer, line 60: _nit_: Mention that this drops the envelopes and logs an error.

relay-server/src/actors/project_buffer.rs
```diff
@@ -0,0 +1,181 @@
...
+impl FromMessage<RemoveMany> for Buffer {
+    type Response = relay_system::NoResponse;
+
+    fn from_message(message: RemoveMany, _: ()) -> Self {
+        Self::RemoveMany(message)
+    }
+}
+
+#[derive(Debug)]
+pub struct BufferService {
```
jan-auer, line 123:
```suggestion
/// In-memory implementation of the [`Buffer`] interface.
#[derive(Debug)]
pub struct BufferService {
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -0,0 +1,181 @@
...
+    keys: BTreeSet<QueueKey>,
+}
+
+impl RemoveMany {
+    pub fn new(project_key: ProjectKey, keys: BTreeSet<QueueKey>) -> Self {
+        Self { project_key, keys }
+    }
+}
+
+/// The Interface for [`BufferService`] service.
```
jan-auer: _nit_: We could describe a bit better here what this interface is for (and how to use it, if necessary).

relay-server/src/actors/project_cache.rs
```diff
@@ -372,102 +375,70 @@ struct UpdateProjectState {
...
-    /// Returns the list of buffered envelopes if they satisfy a predicate.
-    pub fn dequeue<P>(
-        &mut self,
-        partial_key: &ProjectKey,
-        predicate: P,
-    ) -> Vec<(Box<Envelope>, EnvelopeContext)>
-    where
-        P: Fn(&QueueKey) -> bool,
-    {
+    /// Sends the message to [`BufferService`] to dequeue the envelopes.
```
jan-auer, line 404: _nit_: Describe that the envelopes come back via the `buffer_rx` and are sent directly to `handle_processing`. This will make it easier for us to trace the control flow.

jan-auer (approved):

PR #1910, opened 2023-03-08
Assert array fields are capped to 100 items

cmanallen:
Caps the size of the `error_ids`, `trace_ids`, and `urls` arrays to a maximum of 100 array items.  This will prevent very large messages from being produced to our Snuba consumer.  It will also reduce the number of "failed to produce message to Kafka ..." log messages in Relay.

#skip-changelog

closes: https://github.com/getsentry/replay-backend/issues/289

### DIFF
--- a/relay-general/src/protocol/replay.rs
+++ b/relay-general/src/protocol/replay.rs
@@ -271,7 +271,6 @@ impl Replay {
         if let Some(items) = self.urls.value_mut() {
             items.truncate(100);
         }
-
     }
 
     fn normalize_ip_address(&mut self, ip_address: Option<RealIPAddr>) {
--- a/relay-general/src/protocol/replay.rs
+++ b/relay-general/src/protocol/replay.rs
@@ -260,6 +260,8 @@ impl Replay {
     }
 
     fn normalize_array_fields(&mut self) {
+        // TODO: This should be replaced by the TrimmingProcessor.
+        // https://github.com/getsentry/relay/pull/1910#pullrequestreview-1337188206
         if let Some(items) = self.error_ids.value_mut() {
             items.truncate(100);
         }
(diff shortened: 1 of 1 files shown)


relay-general/src/protocol/replay.rs
```diff
@@ -256,6 +256,27 @@ impl Replay {
...
+            items.truncate(100);
+            items.to_owned()
+        });
+        self.trace_ids.set_value(new_trace_ids);
+
+        let new_urls = self.urls.value_mut().as_mut().map(|items| {
+            items.truncate(100);
+            items.to_owned()
+        });
+        self.urls.set_value(new_urls);
```
olksdr: So if you would like to truncate the items, you could do it directly without copying the data, since you have here a `&mut` to that list.  
I think something like following should work:

```suggestion
        if let Some(items) = self.error_ids.value_mut() {
            items.truncate(100);
        }

        if let Some(items) = self.trace_ids.value_mut() {
            items.truncate(100);
        }

        if let Some(items) = self.urls.value_mut() {
            items.truncate(100);
        }

```

cmanallen:
@olksdr Ah, thanks.  That makes much more sense.

relay-general/src/protocol/replay.rs
```diff
@@ -539,6 +554,35 @@ mod tests {
...
+            urls: Annotated::new(urls),
+            error_ids: Annotated::new(error_ids),
+            trace_ids: Annotated::new(trace_ids),
+            ..Default::default()
+        });
+
+        let replay_value = replay.value_mut().as_mut().unwrap();
+        replay_value.normalize_array_fields();
+
+        assert!(replay_value.error_ids.value().unwrap().len() == 100);
```
bruno-garcia, line 583: Worth also checking for `99` and making sure ur func isn't messing up with that (testing the "other direction")

jjbayer (commented):
@cmanallen This looks OK to me, but could you take a look if the existing `bag_size` mechanism would suite your needs as well?

For example:

https://github.com/getsentry/relay/blob/0e35d8d807541a670c2f2dc453ad66f59d351c28/relay-general/src/protocol/event.rs#L466-L468

cmanallen:
@jjbayer How should I write tests for this?  It seems like constructing the Annotated replay directly does not truncate any values.

jjbayer:
> @jjbayer How should I write tests for this? It seems like constructing the Annotated replay directly does not truncate any values.

I have not interfaced with this code TBH, but I think you need to run the `TrimmingProcessor` explicitly: https://github.com/getsentry/relay/blob/ee8d51d4426ccc2b6e931f41c88add75e53c1ca5/relay-general/src/store/trimming.rs#L20

olksdr:
Hey @cmanallen, 
did you have an opportunity to look into this again? Especially in the comment from @jjbayer https://github.com/getsentry/relay/pull/1910#issuecomment-1466558003 ?

cmanallen:
@olksdr I did but I unfortunately don't have the time to follow up on the feedback right now.  If possible I'd like to merge this to solve the immediate problem but if the pulls needs to stay open until I've added it I understand.

cc @jjbayer

jjbayer (approved):
@cmanallen assuming the product can deal with truncated replays, this looks OK to me. Please add a note / todo to the code referencing the alternative solution with the TrimmingProcessor.

PR #1911, opened 2023-03-08
feat(mobile): Move device.class from contexts to tags

edwardgou-sentry:
`device.class` is currently being calculated and stored in the `device` context. This change moves the calculation and storage into tags, under the `device.class` key. This is because we need `device.class` to be surfaced by various tag endpoints in the Sentry product, and we don't need `device.class` in contexts.


jan-auer:
We're usually not adding synthetic tags to the product anymore as this has been causing issues. Tags should be reserved for information the user sets. If you need the device class served, we should ideally serve it from the contexts:
- They are automatically indexed and with minimal configuration can be queried from Discover
- They are visible on issue details by default.

Are there other cases where you need to expose the device class?

edwardgou-sentry:
So I've spoken to @jan-auer. For context:
**Why**: We want `device.class` to be stored as a tag because Sentry has product areas and endpoints (ie `/tags`) that fetch all tag values. We want `device.class` to be included in these areas.
**Problem**: We want to avoid adding these synthetic tags, they've caused problems for Sentry in the past. 

I've taken a look at the `/tags` endpoint, and I wasn't able to find any way to cleanly mesh non-tags data (ie contexts) in the query/response. I don't think we have an alternative solution that satisfies the product requirement other than storing `device.class` as a tag.

relay-general/src/store/normalize.rs
```diff
@@ -661,6 +663,41 @@ fn normalize_logentry(logentry: &mut Annotated<LogEntry>, _meta: &mut Meta) -> P
...
+                    } else if freq < 2500 || mem < 6 * GIB {
+                        tags.insert(tag_name, Annotated::new(DeviceClass::MEDIUM.to_string()));
+                    } else {
+                        tags.insert(tag_name, Annotated::new(DeviceClass::HIGH.to_string()));
+                    }
+                }
+            }
+        }
+    }
+}
```
jjbayer, line 674: nit: IMO readability would benefit from splitting out a helper function, i.e.

```rust
fn normalize_device_class(event: &mut Event) {
     if let Some(device_class) = get_device_class(event) {
         tags.insert("device.class".to_owned(), Annotated::new(device_class.to_string());
    }
}

File: relay-general/src/protocol/device_class.rs [hunk 2]
```diff
@@ -0,0 +1,17 @@
...
+use std::fmt;
+
+#[derive(Debug, FromValue, IntoValue, Empty, Clone, PartialEq)]
+pub struct DeviceClass(pub u64);
+
+impl DeviceClass {
+    pub const LOW: Self = Self(1);
+    pub const MEDIUM: Self = Self(2);
+    pub const HIGH: Self = Self(3);
+}
```
jjbayer, line 46: Now that the device class is a tag, what is the benefit of representing it as an integer (except a few characters saved)? Why not just `"low", "medium", "high"`?

relay-general/src/store/normalize.rs
```diff
@@ -661,6 +663,41 @@ fn normalize_logentry(logentry: &mut Annotated<LogEntry>, _meta: &mut Meta) -> P
     logentry.apply(|le, meta| logentry::normalize_logentry(le, meta))
 }
 
+// Reads device specs (family, memory, cpu, etc) from context and sets the device.class tag to high, medium, or low.
+fn normalize_device_class(event: &mut Event) {
+    let tags = &mut event.tags.value_mut().get_or_insert_with(Tags::default).0;
+    let tag_name = "device.class".to_string();
+    if let Some(ref mut contexts) = event.contexts.value_mut() {
+        if let Some(Context::Device(ref mut device)) = contexts.get_context_mut("device") {
```
jjbayer: These references do not have to be mutable. But I realize there is no `Contexts::get_context` helper. Feel free to copy paste `Contexts::get_context_mut` to create an immutable version.

relay-general/src/store/normalize.rs
```diff
@@ -661,6 +663,41 @@ fn normalize_logentry(logentry: &mut Annotated<LogEntry>, _meta: &mut Meta) -> P
...
+                    if freq < 2000 || proc < 8 || mem < 4 * GIB {
+                        tags.insert(tag_name, Annotated::new(DeviceClass::LOW.to_string()));
+                    } else if freq < 2500 || mem < 6 * GIB {
+                        tags.insert(tag_name, Annotated::new(DeviceClass::MEDIUM.to_string()));
+                    } else {
+                        tags.insert(tag_name, Annotated::new(DeviceClass::HIGH.to_string()));
+                    }
+                }
+            }
+        }
```
jjbayer, line 672: 1. This potentially overwrites a user tag with the same name. Are we fine with that?
2. Will the product be able to handle cases where this function is unable to derive a device class, and a user tag with the same name exists? If not, I would delete the tag here if no device class could be derived.

jjbayer (changes requested):

relay-general/src/store/normalize.rs
edwardgou-sentry: 👍  I've moved most of the logic into `from_contexts` under `device_class.rs`

relay-general/src/protocol/device_class.rs
edwardgou-sentry: Hey Joris, we plan to add additional device classes as higher performing devices get released in the market. Rather than store explicitly as `high medium and low`, we want to store them as number values (even though they'll be strings). This is so we have some ascending sequence of numbers and map those numbers to user facing values of `high medium and low` as we scale up. (ie one day we may have 4 -> low, 5 -> medium, 6 -> high)

relay-general/src/store/normalize.rs
edwardgou-sentry: Good point! I've updated them to be immutable and added the `get_context` helper!

relay-general/src/store/normalize.rs
edwardgou-sentry: 1. Yes, we are ok with overwriting user tags with the same name. This should only be a sentry provided tag.
2. 👍 Added a line to remove any user provided values for the tag in `normalize_device_class`

edwardgou-sentry:
Hey @jjbayer , thanks for the review! I've made some updates to the branch to resolve the changes requested. Thanks!

relay-general/src/protocol/device_class.rs
jjbayer: 👍 Makes sense.

jjbayer (approved):
Thanks for the changes!

PR #1912, opened 2023-03-08
ref(actix): Refactor shutdown without actix

jan-auer:
Removes a dependency on `actix` from Relay's shutdown functionality. Instead, this uses tokio directly to get notified about process signals. The `Controller` no longer creates an actix system as a result of this, which also means that it is now reduced to a service that simply handles shutdowns.

The Controller's API is purely static. This makes it easier to use the controller from various places across the code base. This is subject to be revised in a near-term followup.

Since the web server still requires an actix system, this is now handled by the `HttpServer` service internals. It creates a new thread and blocks on the actix system inside of this thread. The shutdown message is then forwarded into that thread. This required to make startup of the `HttpServer` asynchronous, as we need to wait for the service to start in order to obtain a handle to the inner actix server.

#skip-changelog


relay-system/src/controller.rs
```diff
@@ -40,190 +71,205 @@ impl ShutdownHandle {
...
-        tx.send(Some(Shutdown { timeout })).ok();
-
-        // Delay the shutdown for 100ms to allow recipients of the shutdown signal to execute their
-        // error handlers. Once `System::stop` is called, futures won't be polled anymore and we
-        // will not be able to print error messages.
-        let when = timeout.unwrap_or_else(|| Duration::from_secs(0)) + Duration::from_millis(100);
+    /// This waits for the first shutdown signal and then conditionally waits for the shutdown
+    /// timeout. If the shutdown timeout is interrupted by another signal, this function resolves
+    /// immediately.
+    pub async fn shutdown() {
```
jan-auer: The duality of `shutdown` and `shutdown_handle` is not ideal. I'm now thinking to refactor one more time so that:

- `Controller` has `shutdown_handle()` and `trigger_shutdown`. 
- `ShutdownHandle` has `async notified()` and `async finished()`.

jan-auer (commented):
I'd still like to add some unit tests before merging this.

relay-system/src/lib.rs
```diff
@@ -1,11 +1,19 @@
...
-//! actors must implement a handler for the [`Shutdown`] message and register with the controller to
-//! receive this signal. See the struct level documentation for more information.
+//! Relay's system is based on [`tokio`]. To use any of these components, ensure a tokio runtime is
+//! available.
 //!
-//! See the [`Controller`] struct for more information.
+//! # Services
+//!
+//! The basic building block in Relay are asynchronous [`Service`]s. Each service implements an
+//! [`Interface`], which consists of one or messages that can be sent to the service using its
```
olksdr:
```suggestion
//! [`Interface`], which consists of one or many messages that can be sent to the service using its
```

olksdr (approved):
:shipit:

cleptric (approved):

PR #1913, opened 2023-03-08
ref(monitors): Normalize monitor slug parameter

evanpurkhiser:
This attempts to convert any string passed in the `monitor_slug`
parameter into the style slug sentry will expect on consumption


relay-monitors/src/lib.rs
```diff
@@ -78,6 +82,23 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
...
+        .replace_all(check_in.monitor_slug.as_str(), "")
+        .to_lowercase();
+
+    static SLUGIFIER: OnceCell<Regex> = OnceCell::new();
+    let slugifier_regex = SLUGIFIER.get_or_init(|| Regex::new(r"[\s_\-]+").unwrap());
+
+    check_in.monitor_slug = slugifier_regex
+        .replace_all(clean_slug.as_str(), "-")
+        .get(..SLUG_LENGTH)
+        .unwrap_or("")
```
evanpurkhiser: We probably actually want to reject here. But I'm not sure how to reject from the processor?

iker-barriocanal: This function returns a `Result`, and its error is a `ProcessCheckInError`. I assume you're looking for that?

relay-monitors/src/lib.rs
```diff
@@ -78,6 +82,23 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
         check_in.status = CheckInStatus::Unknown;
     }
 
+    static SLUG_CLEANER: OnceCell<Regex> = OnceCell::new();
+    let slug_cleaner_regex = SLUG_CLEANER.get_or_init(|| Regex::new(r"[^a-zA-Z0-9\s_-]").unwrap());
```
iker-barriocanal: What do you need the `SLUG_CLEANER` for? Why is it `static`? I think what you want to do `static` is `slug_cleaner_regex`?

Same with the `SLUGIFIER` below.

iker-barriocanal (commented):
Can we also add a unit test in rust to cover the test? Besides the integration test.

relay-monitors/src/lib.rs
```diff
@@ -78,6 +82,23 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
...
+        .to_lowercase();
+
+    static SLUGIFIER: OnceCell<Regex> = OnceCell::new();
+    let slugifier_regex = SLUGIFIER.get_or_init(|| Regex::new(r"[\s_\-]+").unwrap());
+
+    check_in.monitor_slug = slugifier_regex
+        .replace_all(clean_slug.as_str(), "-")
+        .get(..SLUG_LENGTH)
+        .unwrap_or("")
+        .trim_matches('-')
```
dashed, line 105: @evanpurkhiser should you also trim underscores as well?

jan-auer:
@evanpurkhiser updated with:
 - Using `Lazy` instead of `OnceCell` (same crate), so the regexes are a one-liner now
 - Changed how slugs get trimmed and truncated. Truncation now happens after trimming.
 - Reject check-ins with an empty slug (after trimming)
 - Added unit tests

jan-auer (approved):

PR #1916, opened 2023-03-10
fix(normalize): No original_value for untouched transactions

jjbayer:
Pattern-based identifier scrubbing would set an original value even when the transaction was unchanged.

#skip-changelog

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -1524,6 +1524,69 @@ mod tests {
         "###);
     }
 
+    /// When no identifiers are scrubbed, we should not set an original value.
+    #[test]
+    fn test_transaction_name_skip_original_value() {
+        let json = r#"
+        {
+            "type": "transaction",
+            "transaction": "/foo/static/page",
+            "transaction_info": {
+              "source": "url"
+            },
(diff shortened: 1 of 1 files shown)


relay-general/src/store/transactions/processor.rs
```diff
@@ -1524,6 +1529,43 @@ mod tests {
...
+        let mut event = Annotated::<Event>::from_json(json).unwrap();
+
+        process_value(
+            &mut event,
+            &mut TransactionsProcessor::new(true, &[]),
+            ProcessingState::root(),
+        )
+        .unwrap();
+
+        assert!(event.meta().is_empty());
```
jjbayer, line 1566: This assertion failed before adding the early return to `normalize_transaction_name`.

relay-general/src/store/transactions/processor.rs
```diff
@@ -282,6 +282,11 @@ fn normalize_transaction_name(transaction: &mut Annotated<String>) -> Processing
             }
         }
 
+        if caps.is_empty() {
+            // Nothing to do for this transaction.
+            return Ok(());
+        }
```
jjbayer, line 288: This also prevents copying the string from `trans` to `changed` and back if they are the same.

olksdr (approved):

PR #1917, opened 2023-03-10
feat(normalization): Optionally mark scrubbed transactions as sanitized

jjbayer:
If the corresponding feature flag is set, mark transactions that were scrubbed by regex patterns as `sanitized` instead of `url`.

### Motivation

If we discovered identifiers in a URL transaction by pattern matching, we previously did not risk marking it as `sanitized`, because it could theoretically still contain user names or other high-cardinality identifiers. But in practice this seems to conservative, and we'd rather assume that a URL scrubbed by pattern matching is low-cardinality enough.

### Risks

If there is a second identifier in the URL that neither the pattern-matching nor the clusterer discovered, we mark something high-cardinality as low-cardinality. This might add noise to the product and an increase in transaction metrics being rate-limited by the cardinality limiter.

https://github.com/getsentry/team-ingest/issues/69

### DIFF
--- a/relay-cabi/src/processing.rs
+++ b/relay-cabi/src/processing.rs
@@ -124,8 +124,7 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
         measurements_config: None, // only supported in relay
         breakdowns_config: None,   // only supported in relay
         normalize_user_agent: config.normalize_user_agent,
-        normalize_transaction_name: false, // only supported in relay
-        tx_name_rules: &[],                // only supported in relay
+        transaction_name_config: Default::default(), // only supported in relay
         is_renormalize: config.is_renormalize.unwrap_or(false),
     };
     light_normalize_event(&mut event, &light_normalization_config)?;
--- a/relay-cabi/src/processing.rs
+++ b/relay-cabi/src/processing.rs
@@ -127,7 +127,7 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
         transaction_name_config: Default::default(), // only supported in relay
         is_renormalize: config.is_renormalize.unwrap_or(false),
     };
-    light_normalize_event(&mut event, &light_normalization_config)?;
+    light_normalize_event(&mut event, light_normalization_config)?;
     process_value(&mut event, &mut *processor, ProcessingState::root())?;
     RelayStr::from_string(event.to_json()?)
 }
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -455,6 +455,7 @@ mod tests {
 
     use chrono::offset::TimeZone;
     use chrono::{Duration, Utc};
+    use insta::assert_debug_snapshot;
     use similar_asserts::assert_eq;
 
     use crate::processor::process_value;
@@ -2030,6 +2031,22 @@ mod tests {
            }
          }
          "###);
+    }
+
+    #[test]
+    fn test_normalize_transaction_names() {
+        let should_be_replaced = [
+            "/aaa11111-aa11-11a1-a11a-1aaa1111a111",
+            "/1aa111aa-11a1-11aa-a111-a1a11111aa11",
+            "/00a00000-0000-0000-0000-000000000001",
+            "/test/b25feeaa-ed2d-4132-bcbd-6232b7922add/url",
+        ];
+        let replaced = should_be_replaced.map(|s| {
+            let mut s = Annotated::new(s.to_owned());
+            scrub_identifiers(&mut s).unwrap();
+            s.0.unwrap()
+        });
+        assert_debug_snapshot!(replaced);
     }
 
     macro_rules! transaction_name_test {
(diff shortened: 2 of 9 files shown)


relay-cabi/src/processing.rs
```diff
@@ -124,8 +124,7 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
         measurements_config: None, // only supported in relay
         breakdowns_config: None,   // only supported in relay
         normalize_user_agent: config.normalize_user_agent,
-        normalize_transaction_name: false, // only supported in relay
-        tx_name_rules: &[],                // only supported in relay
+        transaction_name_config: Default::default(), // only supported in relay
```
jjbayer, line 126: Instead of adding a third transaction name related parameter to `LightNormalizationConfig`, I decided to put it in its own struct.

relay-general/src/store/transactions/processor.rs
```diff
@@ -11,19 +11,29 @@ use crate::types::{Annotated, Meta, ProcessingAction, ProcessingResult, Remark,
 
 use super::TransactionNameRule;
 
+/// Configuration around removing high-cardinality parts of URL transactions.
+#[derive(Debug, Default)]
+pub struct TransactionNameConfig<'r> {
+    /// True if identifiers should be erased from the transaction name.
+    ///
+    /// Applies both pattern-based rules and rules discovered by the clusterer.
```
```suggestion
    /// True if regex patterns should be applied to erase identifiers from the transaction name.
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -259,12 +282,13 @@ fn set_default_transaction_source(event: &mut Event) {
 /// Normalize the transaction name.
 ///
 /// Replaces UUIDs, SHAs and numerical IDs in transaction names by placeholders.
-fn normalize_transaction_name(transaction: &mut Annotated<String>) -> ProcessingResult {
+fn scrub_identifiers(transaction: &mut Annotated<String>) -> Result<bool, ProcessingAction> {
```
jjbayer, line 271:
```suggestion
/// Returns `Ok(true)` if the name was changed.
fn scrub_identifiers(transaction: &mut Annotated<String>) -> Result<bool, ProcessingAction> {
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -340,9 +365,16 @@ impl Processor for TransactionsProcessor<'_> {
...
+        ) && self.name_config.scrub_identifiers
         {
-            normalize_transaction_name(&mut event.transaction)?;
+            let changed = scrub_identifiers(&mut event.transaction)?;
+            if changed && self.name_config.mark_scrubbed_as_sanitized {
+                let source = &mut event
+                    .transaction_info
+                    .get_or_insert_with(Default::default)
+                    .source;
+                source.set_value(Some(TransactionSource::Sanitized));
```
jjbayer, line 363: This is the actual relevant change.

relay-general/src/store/transactions/processor.rs
```diff
@@ -11,19 +11,27 @@ use crate::types::{Annotated, Meta, ProcessingAction, ProcessingResult, Remark,
...
 }
 
 impl<'r> TransactionsProcessor<'r> {
-    pub fn new(normalize_names: bool, tx_name_rules: &'r [TransactionNameRule]) -> Self {
-        Self {
-            normalize_names,
-            tx_name_rules,
-        }
+    pub fn new(name_config: &'r TransactionNameConfig<'r>) -> Self {
+        Self { name_config }
```
olksdr: Just suggestion, so if instead of providing reference here and implementing default with static name config, we could accept here a reference to `LightNormalizationConfig` and then build the config inside of the new? Like following:
```suggestion
    pub fn new(config: &'r LightNormalizationConfig) -> Self {
        let name_config = TransactionNameConfig {
            scrub_identifiers: config.transaction_name_config.scrub_identifiers,
            mark_scrubbed_as_sanitized: config.transaction_name_config.mark_scrubbed_as_sanitized,
            rules: config.transaction_name_config.rules,
        };
        Self { name_config }
  }

```

This way you could still `#[derive(Default)]` on `TransactionsProcessor` and in `normalize.rs` you could just do 
```rust
let mut transactions_processor =
            transactions::TransactionsProcessor::new(config);
```

And of course update tests to use new `new` function 😄 

Right now I'm not sure what would be better,  maybe having this the way it is now it's a bit clearer and have better separation of concerns.

olksdr (approved):
overall LGTM

relay-general/src/store/transactions/processor.rs
olksdr: I have a better idea now, let's make `light_normalize_event` accept not a reference but a a moved variable of `LightNormalizationConfig` and then we could easily move `transaction_name_config` out of it and move it into `new` function. 
This way I think the code will code much cleaner and also there won't be additional jumps through the reference when that not really needed.

@jjbayer  Let me know what you think and we could pair on this.

jjbayer: Good idea, done in https://github.com/getsentry/relay/pull/1917/commits/e2633be36ffdf09aead4955f507b778e5b41488c.

olksdr (approved):

PR #1920, opened 2023-03-13
ref(buffering): Add BufferService with SQLite backend

olksdr:
In PR:
* add config option to enable persistent buffer
* add migrations to create DB and required table for storing the envelopes
  - migrations are built-in
  - migrations run on start up and if the failure happens relay won't start at all
* when persist the envelopes to the disk, the outcome won't be sent
* keep a memory buffer and spool it when it hits defined memory limit
* update `BufferService` to handle the sqlite pooling and insert/select/delete operations
* when Relay stops we try to spool in-memory buffer to disk as well

fix: https://github.com/getsentry/team-ingest/issues/78


olksdr:
Closing this for now, since it's actually needed still.

migrations/20230311081917_envelopes_table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS envelopes (
+  id              INTEGER PRIMARY KEY AUTOINCREMENT,
+  own_key         TEXT,
+  sampling_key    TEXT,
```
jjbayer, line 4: Can we make these fixed width 32 byte arrays? That should make the lookup faster (at least in theory).

relay-server/src/actors/project_buffer.rs
```diff
@@ -189,11 +213,130 @@ impl Service for BufferService {
...
+
+    /// Handles the dequeueing messages from the internal buffer.
+    ///
+    /// This method removes the envelopes from the buffer and stream them to the sender.
+    async fn handle_dequeue(&self, message: DequeueMany) -> Result<(), BufferError> {
+        let DequeueMany { keys, sender } = message;
+
+        for key in keys {
+            let mut envelopes = sqlx::query(
+                "DELETE FROM envelopes WHERE own_key = ? AND sampling_key = ? RETURNING envelope",
```
jjbayer: I think we will need a db index on `(own_key, sampling_key)`.

.gitignore
```diff
@@ -17,3 +17,6 @@ target
 # NPM utilities for building docs
 /node_modules/
 package-lock.json
+
+# Default sqlite db files
+envelope-buffer.*
```
jjbayer: If we make the default path `.relay/envelope-buffer.sqlite3` or `.relay/db/envelope-buffer.sqlite3`, we don't need this entry. The db is also a runtime file so I think it would be a good place:

https://github.com/getsentry/relay/blob/7c4f618d44fe5a2bdedc6c779407871621344e94/.gitignore#L10-L11

relay-config/src/config.rs
```diff
@@ -736,6 +736,21 @@ impl Default for Http {
...
+fn default_persistent_buffer_path() -> PathBuf {
+    PathBuf::from("envelope-buffer.db")
+}
+
+/// Controls internal caching behavior.
+#[derive(Serialize, Deserialize, Debug)]
+struct PersistentBuffer {
+    /// True if the Relay should persist the incoming envelopes to the disk in case they cannot be
+    /// processed right away.
+    enabled: bool,
```
jjbayer: IMO this flag is redundant, because to disable persistence we can just comment out `persistent_envelope_buffer`.

relay-config/src/config.rs
```diff
@@ -1676,6 +1697,24 @@ impl Config {
...
+    }
+
+    /// Returns the path to the persistent envelope buffer file.
+    pub fn cache_persistent_buffer_path(&self) -> PathBuf {
+        self.values
+            .cache
+            .persistent_envelope_buffer
+            .as_ref()
+            .map_or(default_persistent_buffer_path(), |b| b.path.clone())
+    }
```
jjbayer: nit: Can we just have a `cache_persistent_buffer()` method that returns a reference to the config struct?

relay-server/src/actors/project_buffer.rs
```diff
@@ -189,11 +213,130 @@ impl Service for BufferService {
...
+
+    /// Handles the enqueueing messages into the internal buffer.
+    async fn handle_enqueue(&self, message: Enqueue) -> Result<(), BufferError> {
+        let Enqueue {
+            key,
+            value: mut managed_envelope,
+        } = message;
+
+        managed_envelope.spool();
+        let envelope_bytes = managed_envelope.take_envelope().to_vec()?;
```
jjbayer: `take_envelope()` is an ugly hack (that's on me). How about we introduce a `ManagedEnvelope::into_parts()` method for this use case? That could internally give back the permit.

relay-server/src/actors/project_buffer.rs
```diff
@@ -189,11 +213,130 @@ impl Service for BufferService {
...
+            .bind(key.sampling_key.to_string())
+            .fetch(&self.db);
+
+            while let Some(row) = envelopes.try_next().await? {
+                let envelope_bytes_slice: &[u8] = row.try_get("envelope")?;
+                let envelope_bytes = bytes::Bytes::from(envelope_bytes_slice);
+                let envelope = Envelope::parse_bytes(envelope_bytes)?;
+
+                // TODO: do not use the "standalone"  here , rather make sure to hande over the
+                // permit. We have to plug into the system.
```
jjbayer: +1

relay-server/src/actors/project_buffer.rs
```diff
@@ -189,11 +213,130 @@ impl Service for BufferService {
...
+        let RemoveMany { project_key, keys } = message;
+
+        let mut count = 0;
+        for key in keys {
+            let result =
+                sqlx::query("DELETE FROM envelopes where own_key = ? AND sampling_key = ?")
+                    .bind(key.own_key.to_string())
+                    .bind(key.sampling_key.to_string())
+                    .execute(&self.db)
+                    .await?;
```
jjbayer: What is the use case for dropping envelopes from disk?

relay-server/src/actors/project_cache.rs
```diff
@@ -793,6 +794,23 @@ impl Service for ProjectCacheService {
 
             // Channel for envelope buffering.
             let (buffer_tx, mut buffer_rx) = mpsc::unbounded_channel();
+            let buffer = if config.cache_persistent_buffer_enabled() {
+                match SqliteBufferService::from_path(config.cache_persistent_buffer_path()).await {
```
jjbayer: Why is it being initialized within `merge_state`?

relay-server/src/utils/buffer.rs
```diff
@@ -67,3 +75,33 @@ impl BufferGuard {
...
+    if !config.cache_persistent_buffer_enabled() {
+        return Ok(());
+    }
+
+    relay_log::info!("Configuring the persistent envelope buffer");
+
+    let options = SqliteConnectOptions::new()
+        .filename(PathBuf::from("sqlite://").join(config.cache_persistent_buffer_path()))
+        .journal_mode(SqliteJournalMode::Wal)
+        .create_if_missing(true);
```
jjbayer: Why does this code exist 2x, once in `from_path` and once here?

jjbayer (commented):
Preliminary review, mostly questions and bikeshedding.

relay-config/src/config.rs
iker-barriocanal: Or set it to `None`. I agree.

relay-config/src/config.rs
```diff
@@ -769,6 +784,11 @@ struct Cache {
     file_interval: u32,
     /// Interval for evicting outdated project configs from memory.
     eviction_interval: u32,
+    /// The settings to configure persistent buffering.
+    ///
+    /// When enabled all the incoming envelope will be persisted to the disk instead of keeping
+    /// them in memory.
```
```suggestion
    /// When enabled all incoming envelopes will be persisted to disk instead of keeping
    /// them in memory.
```

I'd even go one step further and not comment about implementation details. What this will do is persist to disk, but it'll be possible to have some stuff in memory.

```suggestion
    /// When enabled all incoming envelopes will be persisted to disk.
```

relay-server/src/lib.rs
```diff
@@ -278,6 +278,9 @@ use relay_system::Controller;
 use crate::actors::server::HttpServer;
 use crate::service::ServiceState;
 
+// Re-export utility function to for setting up the persistent envelope buffer if configured.
```
iker-barriocanal:
```suggestion
// Re-export utility function for setting up the persistent envelope buffer if configured.
```

relay-server/src/utils/managed_envelope.rs
```diff
@@ -255,6 +253,15 @@ impl ManagedEnvelope {
         relay_common::instant_to_date_time(self.envelope().meta().start_time())
     }
 
+    /// Set the `done` flag to true.
+    ///
+    /// This allows us to take the envelope out and drop the context without reporting the
+    /// rejection.
```
iker-barriocanal: I think this keeps it to the point? Your call.
```suggestion
    /// Remove the envelope and drop the context, without reporting the rejection.
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -1,11 +1,35 @@
...
 
+use crate::envelope::{Envelope, EnvelopeError};
 use crate::utils::ManagedEnvelope;
 
+/// The set of errors which can happend while working the the buffer.
+#[derive(Debug, thiserror::Error)]
+pub enum BufferError {
+    /// Describes the errors linked with the `Sqlite` backed buffer.
+    #[error("failed to fetch data from the database")]
+    DatabaseError(#[from] sqlx::Error),
```
iker-barriocanal, line 43: A database is a type of backend of a buffer. Should we abstract it? Feel free to address this in a follow-up PR, or not do it at all until we actually use another backend.

relay-server/src/actors/project_buffer.rs
```diff
@@ -1,11 +1,35 @@
...
+use sqlx::sqlite::{SqliteConnectOptions, SqliteJournalMode, SqlitePoolOptions};
+use sqlx::{Pool, Row, Sqlite};
 use tokio::sync::mpsc;
 
+use crate::envelope::{Envelope, EnvelopeError};
 use crate::utils::ManagedEnvelope;
 
+/// The set of errors which can happend while working the the buffer.
+#[derive(Debug, thiserror::Error)]
+pub enum BufferError {
```
iker-barriocanal, line 31: How are we going to handle the errors of accessing the buffer? For example, the buffer no longer exists, or relay no longer has permission to write to it, or max memory is reached (do we cover this in `DatabaseError`), etc.

relay-server/src/actors/project_buffer.rs
```diff
@@ -189,11 +213,130 @@ impl Service for BufferService {
...
 }
+
+/// [`Buffer`] interface implementation backed by SQLite.
+#[derive(Debug)]
+pub struct SqliteBufferService {
+    db: Pool<Sqlite>,
+}
+
+impl SqliteBufferService {
+    /// Creates a new [`SqliteBufferService`] from the provide path to the sqlite database file.
```
iker-barriocanal: nit + to be consistent with the naming, as per the docs above too.
```suggestion
    /// Creates a new [`SqliteBufferService`] from the provided path to the SQLite database file.
```

.gitignore
iker-barriocanal: Should we create a directory to place these, and ignore the filename instead? I think that would make understanding and managing this easier. Or, as Joris suggests above, place the file somewhere in `.relay/`, although I prefer a separate directory.

relay-server/src/actors/project_buffer.rs
```diff
@@ -189,11 +213,130 @@ impl Service for BufferService {
...
+            value: mut managed_envelope,
+        } = message;
+
+        managed_envelope.spool();
+        let envelope_bytes = managed_envelope.take_envelope().to_vec()?;
+        sqlx::query("INSERT INTO envelopes (own_key, sampling_key, envelope) VALUES (?, ?, ?)")
+            .bind(key.own_key.to_string())
+            .bind(key.sampling_key.to_string())
+            .bind(envelope_bytes)
+            .execute(&self.db)
```
iker-barriocanal: `execute` should insert one row, should we check for that and log an error if not?

relay-server/src/actors/project_buffer.rs
iker-barriocanal: TIL `RETURNING`.

relay-server/src/actors/project_buffer.rs
iker-barriocanal: Why not?

relay-server/src/actors/project_buffer.rs
```diff
@@ -189,11 +213,130 @@ impl Service for BufferService {
...
+        }
+
+        Ok(())
+    }
+
+    /// Handles the remove request.
+    ///
+    /// This remove all the envelopes from the internal buffer for the provided keys.
+    /// If any of the provided keys are still have the envelopes, the error will be logged with the
+    /// number of envelopes dropped for the specific project key.
```
iker-barriocanal:
```suggestion
    /// This removes all the envelopes from the internal buffer for the provided keys.
    /// If any of the provided keys still have the envelopes, the error will be logged with the
    /// number of envelopes dropped for the specific project key.
```

relay-server/src/utils/buffer.rs
```diff
@@ -67,3 +75,33 @@ impl BufferGuard {
...
+
+    // All the DB using async and must be run in the context of the tokio runtime.
+    // Also, the DB must be created and migrations run before all the servises start, and if there
+    // are any errors we must bail out ASAP.
```
iker-barriocanal:
```suggestion
    // All DB operations are async and must be run in the context of the tokio runtime.
    // Also, the DB must be created and migrations run before all the services start, and if there
    // are any errors we must bail out ASAP.
```

relay-server/src/utils/buffer.rs
```diff
@@ -67,3 +75,33 @@ impl BufferGuard {
...
+    let setup_rt = create_runtime("setup", 1);
```
iker-barriocanal: nit
```suggestion
    let setup_rt = create_runtime("buffer-setup", 1);
```

relay-server/src/utils/buffer.rs
```diff
@@ -67,3 +75,33 @@ impl BufferGuard {
...
+    setup_rt.block_on(async move {
+        let pool = SqlitePool::connect_with(options).await?;
```
iker-barriocanal: Q: where is the number of connections of the pool configured? (Could we make it part of the `PersistentBuffer` config?)

migrations/20230311081917_envelopes_table.sql
iker-barriocanal, line 1: Q: Will we ever need migrations? If/when we need to make a change, we have to deploy it to make it effective, and that means creating the database from scratch again. Is there anything preventing us from making changes to this file and calling it a day?

I don't intend to bring any actionable items with this question, I think it's good to have this structure. I'm asking for curiosity and to make sure that I understand how this works.

relay-server/src/actors/project_cache.rs
```diff
@@ -793,6 +794,23 @@ impl Service for ProjectCacheService {
...
+                    Ok(buf) => buf.start(),
+
+                    // TODO: how to propagate this error to the entire rely and stop it?
+                    // Should this panic? But this will only kill the service and not the relay.
```
iker-barriocanal: I don't know about propagating the error to the entire relay, but I do think in case of an error it should panic and kill the service. We can postpone this decision for now since it's not critical yet.

iker-barriocanal (commented):
This is good stuff, looks good!

Some comments on top of what Joris suggests.

migrations/20230311081917_envelopes_table.sql
olksdr: I would keep the creation of the DB and tables separately, so the `migrations` folder in my opinion is a good choice for this. 
We might also want to re-evaluate the table structure or add new tables to the DB for more types of the data we want to store.

migrations/20230311081917_envelopes_table.sql
olksdr: I will have to look into this, how the indicies on the binary fields work.

relay-server/src/actors/project_buffer.rs
olksdr: yeah, that would definitely make my life easier 😄

relay-server/src/actors/project_buffer.rs
olksdr: If this statement fails we should get an error

relay-server/src/actors/project_buffer.rs
olksdr: yes, we have an index , see `migrations`

relay-server/src/actors/project_buffer.rs
olksdr: Right now it mimics the current behaviour, if the project expired (maybe it was deleted etc.) we will drop everything we have for it. This normally should not happen. For example, the memory implementation does not drop anything, otherwise we would get many errors.

relay-server/src/actors/project_cache.rs
olksdr: It is not, it's in the `spawn_handler`

relay-server/src/utils/buffer.rs
olksdr: Right now I have no idea how to stop the Relay from an actor, I can produce the error, kill one actor but the rest of the actors are up and running. I want to make sure it fails on start if the Relay cannot read file or run the migrations.

migrations/20230311081917_envelopes_table.sql
jjbayer: > That should make the lookup faster

I'm basing this on an assumption that `TEXT` fields store a reference to a string that is saved somewhere else, while fixed-width types are stored "inline". But this assumption might be wrong as I have no idea how sqlite does it.

migrations/20230311081917_envelopes_table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS envelopes (
+  id              INTEGER PRIMARY KEY AUTOINCREMENT,
```
jjbayer, line 2: I'm curious, do we even need an `id` column? It might help with getting envelopes in a specific order, but we have no requirement of referencing a specific row by ID.

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
+            count_mem_envelopes: 0,
+            db: None,
+        };
+
+        // Only iof persistent buffer enabled, we create the pool and set the config.
+        if let Some(buffer_config) = config.cache_persistent_buffer() {
+            let path = buffer_config.buffer_path().to_owned();
+            let options = SqliteConnectOptions::new()
+                .filename(PathBuf::from("sqlite://").join(&path))
+                .journal_mode(SqliteJournalMode::Wal)
```
jjbayer, line 233: A comment about why we chose `Wal` would be nice here.

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
+            let db = SqlitePoolOptions::new()
+                .max_connections(buffer_config.max_connections())
+                .min_connections(buffer_config.min_connections())
+                .connect_with(options)
+                .await?;
+
+            // Set the buffer memory limit, which must not be bigger then `envelope_buffer_size`.
+            let limit = buffer_config
+                .memory_limit()
+                .unwrap_or(config.envelope_buffer_size() / 2);
```
jjbayer: nit: I would change the signature of `memory_limit()` to `-> usize` and do the `unwrap_or` _and_ the capping to `envelope_buffer_size` in that function's body.

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
+        let Self {
+            db,
+            config,
+            memory_limit,
+            ref mut buffer,
+            ..
+        } = self;
+
+        // Buffer to disk only if the DB and config are provided.
+        if let (Some(db), Some(config), Some(memory_limit)) = (db, config, memory_limit) {
```
jjbayer: nit: If `db`, `config` and `memory_limit` are either all `None` or all `Some`, I would group them together in a struct.

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
+            // And if the count of in memory envelopes is over the defined max buffer size.
+            if self.count_mem_envelopes > *memory_limit as i64 {
+                // Reject all the enqueue requests if we exceed the max size of the buffer.
+                let current_size = std::fs::metadata(config.buffer_path())
+                    .ok()
+                    .map(|meta| meta.len());
+                if current_size.map_or(false, |size| size >= config.max_buffer_size()) {
+                    return Err(BufferError::Full(current_size.unwrap_or_default()));
+                }
```
```suggestion
                let current_size = std::fs::metadata(config.buffer_path())
                    .ok()
                    .map_or(0, |meta| meta.len());
                if current_size > config.max_buffer_size() {
                    return Err(BufferError::Full(current_size));
                }
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
+            service.config = Some(buffer_config.to_owned());
         }
+
+        Ok(service)
+    }
+
+    /// Tries to save in-memory buffer to disk.
+    ///
+    /// It will spool to disk only if the persistent storage enabled in the configuration.
+    fn try_enqueue(&mut self) -> Result<(), BufferError> {
```
jjbayer: `try_spool`

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
+        } = self;
+
+        // Buffer to disk only if the DB and config are provided.
+        if let (Some(db), Some(config), Some(memory_limit)) = (db, config, memory_limit) {
+            // And if the count of in memory envelopes is over the defined max buffer size.
+            if self.count_mem_envelopes > *memory_limit as i64 {
+                // Reject all the enqueue requests if we exceed the max size of the buffer.
+                let current_size = std::fs::metadata(config.buffer_path())
+                    .ok()
+                    .map(|meta| meta.len());
```
jjbayer: With the write-ahead log, a large number of db pages might not have been written to disk yet, so the file size might actually be smaller than the true db size.If this check is just about having _some_ limit on disk growth, it might be good enough though.

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
+                    );
+
+                    // Flatten all the envelopes
+                    let envelopes = buf
+                        .into_iter()
+                        .flat_map(|(k, vals)| {
+                            vals.into_iter()
+                                .map(move |v| (k, v.into_envelope().to_vec()))
+                        })
+                        .collect::<Vec<_>>();
```
jjbayer: nit: Instead of `collect` + `chunks`, we could maybe use the iterator and repeatedly call `take(N)` on it?

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
             }
         }
+
+        // Persistent buffer is configured, lets try to get data from the disk.
+        if let Some(db) = &self.db {
+            // TODO: we also want to limit the number of fetched envelopes from the DB.
+            // DO NOT fetch all the envelopes at once.
+            for key in keys {
+                let mut envelopes = sqlx::query(
+                "DELETE FROM envelopes WHERE own_key = ? AND sampling_key = ? RETURNING envelope",
```
jjbayer: Do we want to `ORDER BY id` here, to maintain the order of events?

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
     ///
     /// This method removes the envelopes from the buffer and stream them to the sender.
-    fn handle_dequeue(&mut self, message: DequeueMany) {
+    async fn handle_dequeue(&mut self, message: DequeueMany) -> Result<(), BufferError> {
         let DequeueMany { keys, sender } = message;
-        for key in keys {
-            for value in self.buffer.remove(&key).unwrap_or_default() {
+        for key in &keys {
+            for value in self.buffer.remove(key).unwrap_or_default() {
+                self.count_mem_envelopes -= 1;
```
jjbayer, line 434: To maintain order of envelopes, shouldn't we yield from the disk first, and then from memory? Or is the order completely irrelevant?

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
                 sender.send(value).ok();
             }
         }
+
+        // Persistent buffer is configured, lets try to get data from the disk.
+        if let Some(db) = &self.db {
+            // TODO: we also want to limit the number of fetched envelopes from the DB.
+            // DO NOT fetch all the envelopes at once.
```
jjbayer: Why not, actually? It seems like `fetch` returns a stream, so it's not like everything is loaded into memory at once. Does the awaiting `try_next` loop block writes to the db while it is streaming?

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,237 @@ impl FromMessage<RemoveMany> for Buffer {
...
+            for key in keys {
+                let mut envelopes = sqlx::query(
+                "DELETE FROM envelopes WHERE own_key = ? AND sampling_key = ? RETURNING envelope",
+            )
+            .bind(key.own_key.to_string())
+            .bind(key.sampling_key.to_string())
+            .fetch(db);
+
+                while let Some(row) = envelopes.try_next().await? {
```
jjbayer: Is there a chance that one row returns an error, but the next one will succeed again? In that case, I would catch the error here and continue the loop, so we don't lose all the envelopes at once.

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +159,240 @@ impl FromMessage<RemoveMany> for Buffer {
...
+            for key in keys {
+                // TODO: remove hardcoded number for the limit.
+                // If the requested permits are available, let use them and fetch the envelopes.
+                if let Ok(mut permits) = self.buffer_guard.try_reserve(100) {
+                    let mut envelopes = sqlx::query(
+                "DELETE FROM envelopes WHERE id IN (SELECT id FROM envelopes WHERE own_key = ? AND sampling_key = ? LIMIT ?) RETURNING envelope",
+            )
+            .bind(key.own_key.to_string())
+            .bind(key.sampling_key.to_string())
+            .bind(100)
```
jjbayer: I left a comment on one of the previous commits. If `fetch` returns a stream, why bother limiting it to 100 entries? Can we instead just exit the loop when we are out of permits, and the remaining envelopes remain on disk?

relay-server/src/utils/buffer.rs
```diff
@@ -67,3 +93,34 @@ impl BufferGuard {
...
+                let migrator = Migrator::new(Path::new("./migrations")).await?;
+                migrator
+                    .run(&pool)
+                    .await
+                    .map_err(project_buffer::BufferError::from)
+            })
+            .ok();
+    }
+    Ok(())
+}
```
jjbayer: How about instead of running all this init code twice, we only run it in `BufferService::create`, and do one of two things if it fails:

1. log an error and fall back to a pure memory buffer,
2. log an error and call `std::process::exit()`?

migrations/20230311081917_envelopes_table.sql
olksdr: This is just to make sure we have a primary key in the table. Since project key and own key can can many records assigned to them

relay-server/src/actors/project_buffer.rs
olksdr: I will have to move the `memory_limit()` function out of the `PersistentBuffer` config, since I do not have access to `envelope_buffer_size` in there. That's why I did it here. 
But I might have to move all this getters to top level instead to keep it consistent.

relay-server/src/actors/project_buffer.rs
olksdr: yeah,  that's true, but the file still should be close to the theoretical size. Since data from WAL will eventually migrate to db. 

Another way would be to issue a query to DB and try to request the size that way. I just wanted to avoid creating another query. But maybe it will be  better and more precise way.

relay-server/src/actors/project_buffer.rs
olksdr: I will have to try it out. Yeah, it sounds like better idea.

relay-server/src/actors/project_buffer.rs
olksdr: This does not really matter I think. But we could do it, yes.

relay-server/src/actors/project_buffer.rs
olksdr: Once delete run, the data will be deleted (if I'm not mistaken), the stream part of it just streaming the collected data. if there are 100k records, all of them will get deleted and will be streamed to the caller, which with this implementation will block the service and will fail when all the permits are exhausted.

I will still have to collect all the envelopes and save them to disk again. I will try that instead actually. 

But I also noticed a bug, that if the `keys` we get aren't used (either error happens, or permits are exhausted) those have to be send back to push into index for the next iteration.

migrations/20230311081917_envelopes_table.sql
jjbayer: Why do you need a primary key though?

relay-server/src/actors/project_buffer.rs
```diff
@@ -321,32 +372,68 @@ impl BufferService {
...
+                    .fetch(db);
+
+                    loop {
+                        match envelopes.try_next().await {
+                            Ok(Some(row)) => match Self::extract_envelope(row, permits.pop()) {
+                                Ok(managed_envelope) => {
+                                    sender.send(managed_envelope).ok();
+                                }
+                                Err(err) => relay_log::error!(
+                                    "failed to extractt envelope from the buffer: {}",
```
```suggestion
                                    "failed to extract envelope from the buffer: {}",
```

relay-server/src/actors/project_cache.rs
```diff
@@ -801,15 +830,18 @@ impl Service for ProjectCacheService {
...
-                Err(err) => {
-                    relay_log::error!("failed to start buffer service: {}", LogError(&err));
-                    return;
-                }
-            };
+            let buffer =
+                match BufferService::create(buffer_guard, project_cache, config.clone()).await {
+                    Ok(buffer) => buffer.start(),
+                    // This should never happens, since on the startup Relay tries to create a DB and
+                    // run migrations, and it that fails the Relay will not start.
```
```suggestion
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -321,32 +372,68 @@ impl BufferService {
...
+                    )
+                    .bind(key.own_key.to_string())
+                    .bind(key.sampling_key.to_string())
+                    .fetch_one(db)
+                    .await;
+
+                    // Make sure to save the key which will be sent back if there are some more
+                    // records left in the db.
+                    if result.map_or(true, |row| !row.is_empty()) {
+                        back_keys.insert(key);
```
jjbayer: When would this happen? Only when `try_next().await` returns an error? In that case, I would skip the extra db query and just push to `back_keys` on line 407.

relay-server/src/actors/project_cache.rs
```diff
@@ -165,6 +165,17 @@ impl UpdateRateLimits {
...
+pub struct BufferIndex {
+    project_key: ProjectKey,
+    keys: BTreeSet<QueueKey>,
+}
+
+impl BufferIndex {
+    pub fn new(project_key: ProjectKey, keys: BTreeSet<QueueKey>) -> Self {
+        Self { project_key, keys }
+    }
+}
```
jjbayer: I dislike the complexity of adding this new message type, especially because it now requires passing the buffering service now suddenly needs to have a notion of a `project_key`. Can we make `DequeueMany` have an `AsyncResponse` instead, and we send the `back_keys` via that channel instead?

jjbayer (commented):
Looks good, I think this is ready for load testing.

relay-server/src/actors/project_buffer.rs
olksdr: I've rewritten this condition a little bit, to make cleat when exactly we have to hit DB again to check if there is more data for a key: 
1. when error occurs while fetching the data 
2. since we read in batches, which in my opinion we must do, otherwise we could lose a lot of data if something goes wrong, or we just push way to match data into the memory (and also we have to give fair share to all the keys which must be fetched from disk) - in this case we still have to check DB if there is data left and if there is, we should keep the key  around for the next round

relay-server/src/actors/project_cache.rs
olksdr: If the `AsyncResponse` is introduced we will have to `await` on it and block, which we want to avoid in the processing pipeline. This will also will require introducing `async`  context where is currently none, project cache and project broker currently sync and do not await on anything. 

The point here was to have everything as async as possible and do not block on any stages.

relay-server/src/utils/buffer.rs
```diff
@@ -45,9 +45,27 @@ impl BufferGuard {
         self.capacity.saturating_sub(self.available())
     }
 
+    /// Reserver the number of requested permits.
```
iker-barriocanal:
```suggestion
    /// Reserves the number of requested permits.
```

relay-server/src/utils/semaphore.rs
```diff
@@ -46,6 +46,30 @@ impl Semaphore {
         self.capacity.load(Ordering::Relaxed)
     }
 
+    /// Tries to reserve the number of requested permits.
+    ///
+    /// Returns Some of the list of permits if the requested number of permits is available.
+    ///
```
iker-barriocanal:
```suggestion
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +167,345 @@ impl FromMessage<RemoveMany> for Buffer {
...
+        // Buffer to disk only if the DB and config are provided.
+        if let Some(BufferSpoolConfig {
+            db,
+            max_disk_size,
+            max_memory_size,
+        }) = spool_config
+        {
+            // And if the count of in memory envelopes is over the defined max buffer size.
+            if self.count_mem_envelopes > *max_memory_size as i64 {
+                // Reject all the enqueue requests if we exceed the max size of the buffer.
```
iker-barriocanal: Question not necessarily related to the PR. If we have too many envelopes in memory and in the buffer, this actor will reject the incoming message. What happens with the envelope in that case? In case it's dropped, how can we prevent that?

relay-server/src/actors/project_buffer.rs
iker-barriocanal: I think the order matters (best effort), but prioritizing one or another depends on what we want to achieve. But I don't know what we should prioritize: yielding from disk first vs memory will be slower, but it may also generate a situation where we are dealing with new envelopes while older ones are enqueued. I'd leave this as a follow-up to deal with when we have more information.

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +167,345 @@ impl FromMessage<RemoveMany> for Buffer {
...
     }
 
     /// Handles the dequeueing messages from the internal buffer.
     ///
     /// This method removes the envelopes from the buffer and stream them to the sender.
-    fn handle_dequeue(&mut self, message: DequeueMany) {
-        let DequeueMany { keys, sender } = message;
-        for key in keys {
-            for value in self.buffer.remove(&key).unwrap_or_default() {
+    async fn handle_dequeue(&mut self, message: DequeueMany) -> Result<(), BufferError> {
```
iker-barriocanal, line 425: This function does a lot of things, could we break it down into smaller functions?

relay-server/src/actors/project_cache.rs
jjbayer: > we will have to await on it and block, which we want to avoid in the processing pipeline.

OK good point. I wonder if the index that maps from project key to queue keys should just move into the buffer service, that would also make this additional message type unnecessary (maybe something for a follow-up refactor).

Either way, please add a doc comment to this type to explain why it exists.

relay-server/src/actors/project_buffer.rs
jjbayer: Thanks! I think I misread the nested loop on my last review. Instead of delegating `back_keys` to the project cache  in the `Ok(None)` case, why can't we keep requesting batches until there's nothing in the db? Something like

```
FOR EVERY key
    LOOP
        envelopes := fetch_and_delete_batch(batch_size=100)
        IF envelopes.is_empty()
            BREAK
        handle_envelopes(envelopes)
```

olksdr: yeah, if that feels better to you, I can definitely do that. Will make a another commit with a change.

relay-server/src/actors/project_buffer.rs
olksdr: done in [6bc057b](https://github.com/getsentry/relay/pull/1920/commits/6bc057b6f66070b601529ae11b0cf6d615944972)

relay-server/src/actors/project_buffer.rs
olksdr: if we hit the limit, it will behave the way it is now, just start dropping the incoming envelopes.

relay-server/src/actors/project_buffer.rs
olksdr: @jjbayer  done in [6bc057b](https://github.com/getsentry/relay/pull/1920/commits/6bc057b6f66070b601529ae11b0cf6d615944972), please, take another look if that looks better for you.

relay-config/src/config.rs
```diff
@@ -1676,6 +1709,65 @@ impl Config {
...
+    /// Set to smaller between `cache.persistent_envelope_buffer.max_memory_size` and `cache.envelope_buffer_size`.
+    /// Default set to half of `cache.envelope_buffer_size`
+    pub fn cache_persistent_buffer_max_memory_size(&self) -> usize {
+        self.values
+            .cache
+            .persistent_envelope_buffer
+            .as_ref()
+            .and_then(|b| b.max_memory_size)
+            .unwrap_or(self.envelope_buffer_size() / 2)
+    }
```
jjbayer, line 1759: nit: This seems like a lot of code for getting values, where the only benefit seems to be setting a default for `max_memory_size` that depends on `envelope_buffer_size`. I would personally prefer a single `fn persistent_envelope_buffer()` set appropriate defaults in the `Default` implementation, and maybe log an error if `max_memory_size > envelope_buffer_size`.

relay-server/src/actors/project_cache.rs
```diff
@@ -164,6 +165,22 @@ impl UpdateRateLimits {
     }
 }
 
+/// Updates buffer index for [`ProjectKey`] with keys list of the [`QueueKey`]s.
+///
+/// This message is sent from the project buffer in case of the error while fetching the data from
+/// the persistent buffer, ensuring that we stil have the index pointing to the keys, which could be found in the
```
```suggestion
/// the persistent buffer, ensuring that we still have the index pointing to the keys, which could be found in the
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +166,335 @@ impl FromMessage<RemoveMany> for Buffer {
...
+                // Reject all the enqueue requests if we exceed the max size of the buffer.
+                let db_file_path =
+                    config
+                        .cache_persistent_buffer_path()
+                        .ok_or(BufferError::DatabaseSizeError(Error::from(
+                            ErrorKind::NotFound,
+                        )))?;
+                let current_size = std::fs::metadata(db_file_path)
+                    .ok()
+                    .map_or(0, |meta| meta.len());
```
jjbayer: nit: This could go into a helper method `self.estimate_db_size() -> usize`.

relay-server/src/actors/project_buffer.rs
jjbayer: Any update on this?

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +166,335 @@ impl FromMessage<RemoveMany> for Buffer {
...
+                        break None;
+                    }
+
+                    for envelope in envelopes {
+                        match self.extract_envelope(envelope) {
+                            Ok(managed_envelope) => {
+                                sender.send(managed_envelope).ok();
+                            }
+                            Err(err) => relay_log::error!(
+                                "failed to extractt envelope from the buffer: {}",
```
```suggestion
                                "failed to extract envelope from the buffer: {}",
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -113,66 +166,335 @@ impl FromMessage<RemoveMany> for Buffer {
...
+    /// managed envelopes and send to back into processing pipeline.
+    ///
+    /// If the error happens in delete/fetch phase, the key is returned to the index and will be
+    /// re-tried later again.
+    async fn fetch_and_delete(
+        &self,
+        db: &Pool<Sqlite>,
+        key: QueueKey,
+        sender: &mpsc::UnboundedSender<ManagedEnvelope>,
+    ) -> Option<QueueKey> {
```
jjbayer: nit: We could change this signature to `Result<(), QueueKey>` to make the intent clearer. I.e. `break None` would become `break Ok(()`, and `break Some(key)` would become `break Err(key)`.

relay-server/src/actors/project_buffer.rs
jjbayer: I would double check any ordering questions with @jan-auer, just in case there are any caveats.

jjbayer (approved):
Looks good! There are still some open questions & suggestions, but none of them are blocking from my point of view.

migrations/20230311081917_envelopes_table.sql
olksdr: For cases if we really decided to give a priority  and order the un-spooled envelopes. 
Do you think it's better to remove it?

relay-config/src/config.rs
olksdr: We currently have same pattern for everything in the config, expose the top level functions for different parts for the config.  I decided to go with it as well. But we might revisit this in the follow-up again

relay-server/src/actors/project_buffer.rs
olksdr: It was a helper but it was also inlined later to satisfy  the borrow checker. And also it is just used in one place anyway.  I can also can have a look into how to refactor this quickly as well.

relay-server/src/actors/project_buffer.rs
olksdr: Forgot to give an update before. 
I looked into this and could not find the best way to accomplish this without `Ittertools`, which I did not want to use. 

I will try today again, if I can come up with something.

migrations/20230311081917_envelopes_table.sql
jjbayer: You're right, let's keep it in case we need ordering.

relay-server/src/actors/project_buffer.rs
olksdr: @jjbayer done something different in [05050ce](https://github.com/getsentry/relay/pull/1920/commits/05050cea15bae42306b352b13d7889c0beb9453e). what do you think?

jjbayer: Looks great!

relay-server/src/actors/project_cache.rs
```diff
@@ -165,12 +165,17 @@ impl UpdateRateLimits {
     }
 }
 
-pub struct BufferIndex {
+/// Updates buffer index for [`ProjectKey`] with keys list of the [`QueueKey`]s.
```
```suggestion
/// Updates the buffer index for [`ProjectKey`] with the [`QueueKey`] keys.
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -371,6 +366,57 @@ impl BufferService {
         Ok(())
     }
 
+    /// Tries to delete the envelops from persistent buffer in batches, extract and convert them to
+    /// managed envelopes and send to back into processing pipeline.
+    ///
+    /// If the error happens in delete/fetch phase, the key is returned to the index and will be
+    /// re-tried later again.
```
iker-barriocanal: > will be re-tried later again.

I'd reword this, as it's an expectation of the caller.

> If the error happens in the deletion/fetching phase, a key is returned to allow retrying later.

iker-barriocanal (commented):
The PR looks good to me, I'd address the order topic before deploying, in [this conversation](https://github.com/getsentry/relay/pull/1920#discussion_r1146222370).

relay-server/src/actors/project_buffer.rs
olksdr: Jan responded offline: this should not be matter for our current setup.

PR #1921, opened 2023-03-13
chore: Apply import grouping from coding style

jan-auer:
Applies the Sentry [import order guidelines](https://develop.sentry.dev/rust/#import-order) on the entire codebase. One notable difference is that public re-exports (`pub use`) are grouped with other imports.

On top of default settings, this is equivalent to the following `rustfmt` configuration:

```toml
imports_granularity = "Module"
group_imports = "StdExternalCrate"  # nightly only
```

Since the latter option is nightly-only at the time of this PR, we do not add a rustfmt configuration to the codebase yet. For more information on the used options, refer to the [official documentation](https://rust-lang.github.io/rustfmt/?version=v1.5.2&search=import).

See also https://github.com/getsentry/develop/pull/869.

#skip-changelog


olksdr (approved):
🚀

sl0thentr0py (approved):

jernejstrasner (approved):

PR #1923, opened 2023-03-13
feat(config): Add links to docs in YAML config file

### LINKED ISSUE #949: Add documentation links to config file
In his great [post on Sentry 10](https://medium.com/@twunde/the-unofficial-guide-to-upgrading-sentry-on-premise-9-to-sentry-10-5e8b4e476349), Thomas Wunderlich mentions:

> There are two sets of documentation that are useful to read, the first is the operating guidelines: https://docs.sentry.io/product/relay/operating-guidelines/,the second is the full list of config options https://docs.sentry.io/product/relay/options/. These pages are so useful that I normally add the following comment to the top of the config file:
> 
> ```
> # Please see the relevant documentation:
> # Performance Tuning: https://docs.sentry.io/product/relay/operating-guidelines/
> # All config options: https://docs.sentry.io/product/relay/options/
> ```

This is something our config generator in `relay config init` can and should do.

iker-barriocanal:
Closes https://github.com/getsentry/relay/issues/949.

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -199,7 +199,7 @@ trait ConfigObject: DeserializeOwned + Serialize {
 
         match Self::format() {
             ConfigFormat::Yaml => {
-                f.write_all(CONFIG_HEADER.as_bytes())?;
+                f.write_all(CONFIG_YAML_HEADER.as_bytes())?;
                 serde_yaml::to_writer(&mut f, self)
                     .with_context(|| ConfigError::file(ConfigErrorKind::CouldNotWriteFile, &path))?
             }
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -193,8 +193,20 @@ trait ConfigObject: DeserializeOwned + Serialize {
             .with_context(|| ConfigError::file(ConfigErrorKind::CouldNotWriteFile, &path))?;
 
         match Self::format() {
-            ConfigFormat::Yaml => serde_yaml::to_writer(&mut f, self)
-                .with_context(|| ConfigError::file(ConfigErrorKind::CouldNotWriteFile, &path))?,
+            ConfigFormat::Yaml => {
+                f.write_all(b"# Please see the relevant documentation:\n")
+                    .unwrap();
+                f.write_all(
+                    b"# Performance tuning: https://docs.sentry.io/product/relay/operating-guidelines/\n"
+                )
+                .unwrap();
+                f.write_all(
+                    b"# All config options: https://docs.sentry.io/product/relay/options/\n",
+                )
+                .unwrap();
+                serde_yaml::to_writer(&mut f, self)
+                    .with_context(|| ConfigError::file(ConfigErrorKind::CouldNotWriteFile, &path))?
+            }
             ConfigFormat::Json => serde_json::to_writer_pretty(&mut f, self)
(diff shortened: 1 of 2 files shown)


relay-config/src/config.rs
```diff
@@ -193,8 +193,20 @@ trait ConfigObject: DeserializeOwned + Serialize {
...
+                f.write_all(b"# Please see the relevant documentation:\n")
+                    .unwrap();
+                f.write_all(
+                    b"# Performance tuning: https://docs.sentry.io/product/relay/operating-guidelines/\n"
+                )
+                .unwrap();
+                f.write_all(
+                    b"# All config options: https://docs.sentry.io/product/relay/options/\n",
+                )
+                .unwrap();
```
iker-barriocanal: I tried writing these with a single buffer, but I was not able to without adding a bunch of `\n`s and making it unreadable.

iker-barriocanal: I've also experimented writing these as part of `MinimalConfig`, but I think doing that is an overkill at this point considering we only want to do this for YAML files (since [JSON doesn't support comments](https://www.json.org/json-en.html)).

relay-config/src/config.rs
```diff
@@ -193,8 +193,20 @@ trait ConfigObject: DeserializeOwned + Serialize {
             .with_context(|| ConfigError::file(ConfigErrorKind::CouldNotWriteFile, &path))?;
 
         match Self::format() {
-            ConfigFormat::Yaml => serde_yaml::to_writer(&mut f, self)
-                .with_context(|| ConfigError::file(ConfigErrorKind::CouldNotWriteFile, &path))?,
+            ConfigFormat::Yaml => {
+                f.write_all(b"# Please see the relevant documentation:\n")
```
jan-auer: Could you pull these out into a single top-level constant (can be a `&str` if that's easier)?

relay-config/src/config.rs
```diff
@@ -193,8 +193,20 @@ trait ConfigObject: DeserializeOwned + Serialize {
...
+                    .unwrap();
```
jan-auer: Instead of `unwrap`, please bubble the I/O error here.

jan-auer (changes requested):

relay-config/src/config.rs
jan-auer: Example:

```rust
// put this somewhere top-level
static CONFIG_HEADER: &str = r###"# Please see the relevant documentation.
# Performance tuning: https://docs.sentry.io/product/relay/operating-guidelines/
# All config options: https://docs.sentry.io/product/relay/options/
"###;

f.write_all(CONFIG_HEADER.as_bytes())?
```

iker-barriocanal: Good suggestion, done in 348dece234459d5fbf914398677487aab0af91cb.

I spent too much time trying to do something like that, but I was using `const` and too many workarounds instead. So TIL the difference: https://github.com/rust-lang/rfcs/blob/master/text/0246-const-vs-static.md.

jan-auer: It works the same with `const`, actually. See this playground:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=77507675c75105ce7f3166c7382a245b

iker-barriocanal: Hmmm, then I don't know what I was doing wrong 😕.

jan-auer (approved):

PR #1924, opened 2023-03-13
chore(ci): Skip generating credentials file with google auth

olksdr:
We should not need this, since we do not pass the creds to other steps. And having only `gcloud` utility authenticated should be just enough. 

#skip-changelog


jan-auer (approved):

olksdr:
Closing this, the cred file is needed for now.

PR #1922, opened 2023-03-13
ci(gha): Remove unmaintained actions-rs actions

jan-auer:
Removes the unmaintained Rust actions from `actions-rs`, since they contain a
deprecated command that will be removed soon. While GitHub frequently updates
its preinstalled rust toolchain, we keep explicit `rustup toolchain install` in
our jobs to ensure we're always on latest.

The `cargo` invocations remain in CI for readability, instead of calling the
equivalent in the Makefile for now.

For more information on the actions, see the original issue.

Fixes #1919
#skip-changelog

PR #1925, opened 2023-03-13
fix(ci): Use correct Rust toolchain in Beta CI

jan-auer:
Fixes a regression in #1922 and uses the stated toolchain instead of stable in
the Beta CI jobs.

#skip-changelog


olksdr (approved):

iker-barriocanal (approved):
🚀

Issue #1818, opened 2023-02-06
Switch from global service registry to dependency "injection"


olksdr:
Currently we use a global registry https://github.com/getsentry/relay/blob/861966fdc58b7f7815c99ca51928df63a57ff90f/relay-server/src/service.rs#L35 with all the services having a record there.

This introduces few challenges:
* it's kind of difficult to test separate parts of the services when they are depend on each other 
* it's not possible to predict the oder in which services are shutting down, and we might be losing some data

```[tasklist]
### Subtasks
- [ ] https://github.com/getsentry/relay/issues/1610
- [ ] https://github.com/getsentry/relay/issues/1711
- [ ] https://github.com/getsentry/relay/issues/1629
- [ ] https://github.com/getsentry/relay/issues/1608
```

As an example how this could look like you can refer to #1770, which introduces also a small isolated test for only 1 service.  This will have to be changed for all services:

```[tasklist]
### PRs
- [ ] https://github.com/getsentry/relay/pull/1770
- [ ] https://github.com/getsentry/relay/pull/1926
- [ ] https://github.com/getsentry/relay/pull/1927
- [ ] https://github.com/getsentry/relay/pull/1928
- [ ] https://github.com/getsentry/relay/pull/1932
- [ ] https://github.com/getsentry/relay/pull/1946
- [ ] https://github.com/getsentry/relay/pull/1947
- [ ] https://github.com/getsentry/relay/pull/1956
```

olksdr:
At this point we still have few things left where the global registry is used.

Here is the quick grep:

```bash 
relay-server/src/utils/managed_envelope.rs
191:        let outcome_aggregator = TrackOutcome::from_registry();

relay-server/src/endpoints/public_keys.rs
12:    let relay_cache = RelayCache::from_registry();

relay-server/src/endpoints/outcomes.rs
13:    let producer = OutcomeProducer::from_registry();

```
and

```bash
relay-server/src/actors/envelopes.rs
110:                        ProjectCache::from_registry().send(UpdateRateLimits::new(

relay-server/src/endpoints/common.rs
289:        EnvelopeProcessor::from_registry().send(ProcessMetrics {
309:        ProjectCache::from_registry().send(ValidateEnvelope::new(event_context));
318:        ProjectCache::from_registry().send(ValidateEnvelope::new(managed_envelope));

relay-server/src/endpoints/health_check.rs
16:    Ok(match HealthCheck::from_registry().send(message).await {

[...]

PR #1926, opened 2023-03-13
ref: Inject services in project cache and project upstream

olksdr:
Getting rid of the global registry and switching just to using injected addresses of the service to send the messages to.

ref: #1818 

/cc @jjbayer 

#skip-changelog


iker-barriocanal (approved):
Nicee!

jjbayer (approved):

PR #1928, opened 2023-03-13
ref(project): Inject services into  the `Project`

olksdr:
Inject services into the `Project`.

ref #1818 

#skip-changelog


relay-system/src/service.rs
```diff
@@ -856,6 +862,7 @@ pub fn channel<I: Interface>(name: &'static str) -> (Addr<I>, Receiver<I>) {
     interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
 
     let receiver = Receiver {
+        addr: addr.clone(),
```
olksdr, line 865: @jan-auer I had to add one more field to the `Receiver` to be able to extract the address of the service this receiver belongs to. 
If you have another idea how that could be accomplished better, please, let me know.

relay-system/src/service.rs
```diff
@@ -828,6 +829,11 @@ impl<I: Interface> Receiver<I> {
             }
         }
     }
+
+    /// Address of the service this [`Receiver`] belongs to.
+    pub fn service_address(&self) -> Addr<I> {
```
jan-auer: What if we call this just `addr`? It's the `Addr` that belongs to this `Receiver`.

relay-server/src/actors/project.rs
```diff
@@ -399,12 +403,22 @@ pub struct Project {
 
 impl Project {
     /// Creates a new `Project`.
-    pub fn new(key: ProjectKey, config: Arc<Config>) -> Self {
+    pub fn new(
+        key: ProjectKey,
+        config: Arc<Config>,
+        project_cache: Addr<ProjectCache>,
+        envelope_manager: Addr<EnvelopeManager>,
+        #[cfg(feature = "processing")] envelope_processor: Addr<EnvelopeProcessor>,
```
jjbayer: nit: could we pass them in the same order as the fields are declared?

relay-server/src/actors/project.rs
```diff
@@ -823,12 +839,44 @@ mod tests {
...
+                            tokio::select! {
+                                Some(_) = rx.recv() => (),
+                                else => break,
+                            }
+                        }
+                    });
+                }
+            }
+        };
+    }
```
jjbayer: Nice! We can probably put this in `testutils.rs`. Should we call this `dummy_service`?

relay-server/src/actors/project.rs
```diff
@@ -399,12 +403,22 @@ pub struct Project {
...
+    ) -> Self {
         Project {
             backoff: RetryBackoff::new(config.http_max_retry_interval()),
             next_fetch_attempt: None,
             last_updated_at: Instant::now(),
             project_key: key,
+            project_cache,
+            #[cfg(feature = "processing")]
+            envelope_processor,
+            envelope_manager,
```
jjbayer: Out of scope for this PR, but the necessity of passing 3 `Addr`s into every single `Project` leads me to believe that we have a flaw in our design. Maybe we need to move even more functionality from `Project` to `ProjectCache`?

When we are done with https://github.com/getsentry/relay/issues/1818, we should probably draw a diagram of the current data flow, and see how we can improve it.

jjbayer (approved):

relay-server/src/actors/project.rs
jan-auer: Instead of `testutils.rs`, put it in `relay_test/`. That way, we can also use it in other crates.

Also, I think you can write this easily without a macro. See the [original Gist](https://gist.github.com/jan-auer/68305125c13aaab7355a1c99b3b65012#file-main-rs-L214-L232).

relay-system/src/service.rs
olksdr: This was removed, and not needed anymore with new changes.

relay-server/src/actors/project.rs
olksdr: This also is not needed right now. 
As the followup I will look into this and add tests and also more test utils including suggestions.

relay-server/src/actors/project.rs
```diff
@@ -386,6 +386,7 @@ enum GetOrFetch<'a> {
 /// Projects can define multiple keys, in which case this structure is duplicated for each instance.
 #[derive(Debug)]
 pub struct Project {
+    aggregator: mpsc::UnboundedSender<Aggregator>,
```
jjbayer: Why can this not be an `Addr<Aggregator>`?

relay-server/src/actors/project_cache.rs
```diff
@@ -386,7 +378,9 @@ struct UpdateProjectState {
 #[derive(Debug)]
 struct ProjectCacheBroker {
     config: Arc<Config>,
+    aggregator: mpsc::UnboundedSender<Aggregator>,
```
jjbayer: Why is this not an `Addr`?

relay-server/src/actors/project_cache.rs
```diff
@@ -507,9 +502,14 @@ impl ProjectCacheBroker {
             no_cache,
         } = message;
 
-        self.get_or_create_project(project_key)
+        let request_update = self
+            .get_or_create_project(project_key)
             .update_state(state.clone(), no_cache);
 
+        if let Some(reqeust_update) = request_update {
+            self.handle_request_update(reqeust_update)
```
```suggestion
        if let Some(request_update) = request_update {
            self.handle_request_update(request_update)
```

relay-server/src/actors/project_cache.rs
```diff
@@ -550,26 +550,45 @@ impl ProjectCacheBroker {
...
 
     fn handle_get(&mut self, message: GetProjectState, sender: ProjectSender) {
-        self.get_or_create_project(message.project_key)
+        let request_update = self
+            .get_or_create_project(message.project_key)
             .get_state(sender, message.no_cache);
+
+        if let Some(request_update) = request_update {
+            self.handle_request_update(request_update)
+        }
```
jjbayer: To be honest, I am not sure whether this is actually better than passing the `Addr`s to `Project`. I feel that it will be easy to forget to send the update request in future refactors. I am in favor of reverting back to https://github.com/getsentry/relay/pull/1928/commits/3940b2ce83cf1fed2d54d8c8ce542944ba0ed029 and merging that version.

Alternatively, we could pass a callback or an `Addr` into `Project::get_state` such that the project takes care of sending the `RequestUpdate` again (similar as you do with `flush_buckets`).

relay-server/src/actors/project.rs
olksdr: We have a circular dependency here, since aggregator already have an address of the cache project. Cache project was started first, and we could not provide the address of the aggregator here. Using channels here is one of the solution which we can use instead.

relay-server/src/actors/project_cache.rs
olksdr: See previous comment.

relay-server/src/actors/project_cache.rs
olksdr: Right now the service does not know address of themselves, which can be fixed of course.
Passing address as a parameter might be better here, that they using this function will force the caller to provide a correct type.

olksdr: changed this in [81baec3](https://github.com/getsentry/relay/pull/1928/commits/81baec31435fd9712919fb15ec480cb8c4ad1530)

relay-server/src/service.rs
```diff
@@ -155,6 +159,7 @@ impl ServiceState {
         let aggregator = AggregatorService::new(
             config.aggregator_config().clone(),
             Some(project_cache.clone().recipient()),
+            aggregator_receiver,
```
jjbayer: Not for this PR, but as discussed offline, I think we could generalize this approach and always create the `Addr`s and corresponding `rx` _before_  we create the services.

Right now the `AggregatorService` has two channels for the same data type, which should not be necessary.

relay-metrics/src/aggregation.rs
```diff
@@ -1502,6 +1503,7 @@ pub struct AggregatorService {
     receiver: Option<Recipient<FlushBuckets, NoResponse>>,
     state: AggregatorState,
     cost_tracker: CostTracker,
+    projects: mpsc::UnboundedReceiver<Aggregator>,
```
jjbayer: Is there a better name we can give to this field? Maybe `project_rx` or `rx_from_projects`?

tests/integration/test_projectconfigs.py
```diff
@@ -389,29 +389,21 @@ def test_cached_project_config(mini_sentry, relay):
...
-
-    # This is still a grace period, and the state for us must be still valid, even though we get the error parsing the new state.
     try:
         # Give it a bit time for update to go through.
-        time.sleep(1)
+        time.sleep(3)
         data = get_response(relay, packed, signature)
+        assert data["configs"][public_key]["disabled"]
+
+        from pprint import pprint
```
jjbayer: Why this import? What happened to the asserts in this test?

relay-server/src/actors/project_cache.rs
jjbayer: Thanks, it looks much safer now!

jjbayer (approved):

relay-metrics/src/aggregation.rs
olksdr: yep, renamed.

tests/integration/test_projectconfigs.py
olksdr: Removed, used for some debugging. And also reverted test

relay-server/src/service.rs
olksdr: @jjbayer  if you can give a final look at this?  😄 
I've refactored this a bit to use `Addr<...>` now

jjbayer: Even better!

jjbayer (approved):

Issue #1888, opened 2023-02-28
Make it impossible to track outcomes for items that remain in the envelope


jjbayer:
Before https://github.com/getsentry/relay/pull/1877, we reported rate limiting outcomes for replay items that remained in the envelope. This was possible because `track_outcome` uses the output of `EnvelopeLimiter::enforce`, but that function did not actually remove replays from the envelope:

https://github.com/getsentry/relay/blob/82af3929b38b21fa177a0564dc3e85ee42ed5a01/relay-server/src/utils/rate_limits.rs#L408-L410

Not sure what's the exact solution here, but generally speaking, any function that creates an outcome should _consume_ the item that has been dropped.

```[tasklist]
### Tasks
- [x] Refactor EnvelopeContext: https://github.com/getsentry/relay/pull/1936
- [ ] https://github.com/getsentry/relay/pull/1944
- [x] Wontfix: https://github.com/getsentry/relay/pull/1940
- [x] Make test work without running System
- [x] Wontfix: Make inner envelope private -> Won't work, because we should not update the envelope summary after the event has been extracted.
```

jjbayer:
Outcomes can now be produced right where the envelope items are dropped, but the initial goal of simplifying outcome reporting in rate limits has not been achieved.

Let's take a more top-level look at how to restructure Relay's data flow and the envelope data type (hierarchical containers instead of flat list of items?) instead.

PR #1936, opened 2023-03-15
ref(server): Move Envelope into EnvelopeContext

jjbayer:
Make `EnvelopeContext` the owner of `Envelope`. and rename it to `ManagedEnvelope`.

In follow-up PRs, this will allow us to:

- Move outcome tracking into `retain_item`.
- Let `ManagedEnvelope` manage all modifications to `Envelope`, such that the `EnvelopeSummary` is always up-to-date.

#skip-changelog

ref: https://github.com/getsentry/relay/issues/1888

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -62,7 +62,7 @@ use crate::metrics_extraction::transactions::{extract_transaction_metrics, Extra
 use crate::service::REGISTRY;
 use crate::statsd::{RelayCounters, RelayHistograms, RelayTimers};
 use crate::utils::{
-    self, get_sampling_key, ChunkedFormDataAggregator, EnvelopeContext, FormDataIter,
+    self, get_sampling_key, ChunkedFormDataAggregator, FormDataIter, ManagedEnvelope,
     SamplingResult,
 };
 
@@ -302,19 +302,19 @@ struct ProcessEnvelopeState {
     /// ID.
     project_id: ProjectId,
 
-    /// The envelope context before processing.
-    envelope_context: EnvelopeContext,
+    /// The managed envelope before processing.
+    managed_envelope: ManagedEnvelope,
 }
 
 impl ProcessEnvelopeState {
     /// Returns a reference to the contained [`Envelope`].
     fn envelope(&self) -> &Envelope {
-        self.envelope_context.envelope()
+        self.managed_envelope.envelope()
     }
 
     /// Returns a mutable reference to the contained [`Envelope`].
     fn envelope_mut(&mut self) -> &mut Envelope {
-        self.envelope_context.envelope_mut()
+        self.managed_envelope.envelope_mut()
     }
 
     /// Returns whether any item in the envelope creates an event in any relay.
@@ -405,7 +405,7 @@ pub struct ProcessEnvelopeResponse {
     /// This is `Some` if the envelope passed inbound filtering and rate limiting. Invalid items are
     /// removed from the envelope. Otherwise, if the envelope is empty or the entire envelope needs
     /// to be dropped, this is `None`.
-    pub envelope: Option<EnvelopeContext>,
+    pub envelope: Option<ManagedEnvelope>,
 }
 
 /// Applies processing to all contents of the given envelope.
@@ -419,7 +419,7 @@ pub struct ProcessEnvelopeResponse {
 ///  - Rate limiters and inbound filters on events in processing mode.
 #[derive(Debug)]
 pub struct ProcessEnvelope {
-    pub envelope_context: EnvelopeContext,
+    pub envelope: ManagedEnvelope,
     pub project_state: Arc<ProjectState>,
     pub sampling_project_state: Option<Arc<ProjectState>>,
 }
@@ -808,10 +808,10 @@ impl EnvelopeProcessorService {
     /// Both are removed from the envelope if they contain invalid JSON or if their timestamps
     /// are out of range after clock drift correction.
     fn process_sessions(&self, state: &mut ProcessEnvelopeState) {
-        let received = state.envelope_context.received_at();
+        let received = state.managed_envelope.received_at();
         let extracted_metrics = &mut state.extracted_metrics.project_metrics;
         let metrics_config = state.project_state.config().session_metrics;
-        let envelope = state.envelope_context.envelope_mut();
+        let envelope = state.managed_envelope.envelope_mut();
         let client = envelope.meta().client().map(|x| x.to_owned());
         let client_addr = envelope.meta().client_addr();
 
@@ -895,7 +895,7 @@ impl EnvelopeProcessorService {
 
         let mut timestamp = None;
         let mut output_events = BTreeMap::new();
-        let received = state.envelope_context.received_at();
+        let received = state.managed_envelope.received_at();
 
         let clock_drift_processor = ClockDriftProcessor::new(state.envelope().sent_at(), received)
             .at_least(MINIMUM_CLOCK_DRIFT);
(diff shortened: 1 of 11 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -808,14 +811,19 @@ impl EnvelopeProcessorService {
...
+            .map(|x| x.to_owned());
+        let client_addr = state.envelope_context.envelope().meta().client_addr();
 
         let clock_drift_processor =
-            ClockDriftProcessor::new(envelope.sent_at(), received).at_least(MINIMUM_CLOCK_DRIFT);
+            ClockDriftProcessor::new(state.envelope_context.envelope().sent_at(), received)
+                .at_least(MINIMUM_CLOCK_DRIFT);
 
-        envelope.retain_items(|item| {
+        state.envelope_context.envelope_mut().retain_items(|item| {
```
jjbayer: The borrow checker did not like my use of `state.envelope_mut()` here.

relay-server/src/actors/processor.rs
```diff
@@ -1052,77 +1060,87 @@ impl EnvelopeProcessorService {
...
-            _ => true,
-        });
+                },
+                _ => true,
+            });
+        for (outcome, category, quantity) in outcomes {
+            state
+                .envelope_context
+                .track_outcome(outcome, category, quantity);
+        }
```
jjbayer: This is just a temporary workaround which will be removed once we move outcome tracking into `retain_items`.

relay-server/src/actors/processor.rs
```diff
@@ -1139,102 +1157,107 @@ impl EnvelopeProcessorService {
...
-            _ => true,
-        });
+                _ => true,
+            });
+
+        for (outcome, category, quantity) in outcomes {
+            state
+                .envelope_context
+                .track_outcome(outcome, category, quantity);
+        }
```
jjbayer: This is just a temporary workaround which will be removed once we move outcome tracking into retain_items.

relay-server/src/utils/envelope_context.rs
```diff
@@ -91,23 +99,41 @@ impl EnvelopeContext {
...
+        self.envelope.as_mut()
+    }
+
+    /// Take the envelope out of the context and replace it with a dummy.
+    ///
+    /// Note that after taking out the envelope, the envelope summary is incorrect.
+    pub(crate) fn take_envelope(&mut self) -> Box<Envelope> {
+        let dummy_envelope = Envelope::parse_bytes(Bytes::from("{}")).unwrap();
+        std::mem::replace(&mut self.envelope, dummy_envelope)
+    }
```
jjbayer, line 129: This method should not exist because it leaves the `EnvelopeContext` in an inconsistent state. It is only used in `handle_submit` at the end of the envelope's life time.

relay-server/src/utils/envelope_context.rs
```diff
@@ -60,6 +61,7 @@ impl Handling {
 /// management. It is automatically reclaimed when the context is dropped along with the envelope.
 #[derive(Debug)]
 pub struct EnvelopeContext {
+    envelope: Box<Envelope>,
```
jjbayer, line 70: Should we rename `EnvelopeContext` to `ScopedEnvelope`?

relay-server/src/utils/envelope_context.rs
```diff
@@ -91,23 +98,40 @@ impl EnvelopeContext {
...
+    pub fn envelope_mut(&mut self) -> &mut Envelope {
+        self.envelope.as_mut()
+    }
+
+    /// Take the envelope out of the context and replace it with a dummy.
+    ///
+    /// Note that after taking out the envelope, the envelope summary is incorrect.
+    pub(crate) fn take_envelope(&mut self) -> Box<Envelope> {
+        Box::new(self.envelope.take_items())
+    }
```
jjbayer, line 122: This method should not exist because it leaves the EnvelopeContext in an inconsistent state. It is only used in `handle_submit` at the end of the envelope's life time.

relay-server/src/actors/processor.rs
iker-barriocanal: In line 814, you have a non-mutable (?) reference. Could you make that mutable, and reuse that same reference here?

relay-server/src/utils/envelope_context.rs
iker-barriocanal: SGTM, but no strong opinions.

relay-server/src/utils/envelope_context.rs
```diff
@@ -72,15 +73,21 @@ pub struct EnvelopeContext {
...
-            event_id: envelope.event_id(),
-            remote_addr: meta.client_addr(),
-            scoping: meta.get_partial_scoping(),
+            envelope,
+            summary,
+            start_time,
+            received_at: relay_common::instant_to_date_time(start_time),
+            event_id,
+            remote_addr,
+            scoping,
```
iker-barriocanal: This is personal preference, but in this case at least, I prefer the previous approach instead of having variables. However, if you decide to leave the variables, I'd also extract `received_at`.

relay-server/src/utils/envelope_context.rs
```diff
@@ -91,23 +98,40 @@ impl EnvelopeContext {
...
+    /// Returns a mutable reference to the contained [`Envelope`].
+    pub fn envelope_mut(&mut self) -> &mut Envelope {
+        self.envelope.as_mut()
+    }
+
+    /// Take the envelope out of the context and replace it with a dummy.
+    ///
+    /// Note that after taking out the envelope, the envelope summary is incorrect.
+    pub(crate) fn take_envelope(&mut self) -> Box<Envelope> {
+        Box::new(self.envelope.take_items())
```
iker-barriocanal, line 121: Out of curiosity, how does the rust handle the memory allocated to `envelope` here? Since the object is not destroyed, I assume the same amount of memory is saved?

relay-server/src/utils/envelope_context.rs
```diff
@@ -91,23 +98,40 @@ impl EnvelopeContext {
...
+    }
+
     /// Update the context with new envelope information.
```
iker-barriocanal:
```suggestion
    /// Update the context with envelope information.
```

iker-barriocanal (approved):

relay-server/src/utils/envelope_context.rs
jjbayer: The `Box::new` will cause a new allocation. `take_items` will copy all the headers, but the envelope items themselves remain where they are in memory. Only the pointer to those items is moved.

relay-server/src/actors/processor.rs
jjbayer: Seems to work, thanks.

relay-server/src/utils/envelope_context.rs
jjbayer: I need the variables because the envelope is being moved now, so I cannot reference `meta` anymore. But good catch, will also make a variable for `received_at`.

relay-server/src/utils/envelope_context.rs
olksdr: With all this refactoring `EnvelopeContext` does not make much sense to me now. I would rename it to proposed `ScopedEnvelope`

olksdr (commented):
I think these changes fine for the initial iteration.  I would just rename the `EnvelopeContext`  for now and that's it.

relay-server/src/utils/managed_envelope.rs
```diff
@@ -91,23 +99,40 @@ impl EnvelopeContext {
...
     #[cfg(test)]
-    pub fn standalone(envelope: &Envelope) -> Self {
+    pub fn standalone(envelope: Box<Envelope>) -> Self {
         Self::new_internal(envelope, None)
     }
 
-    /// Computes an envelope context from the given envelope and binds it to the processing queue.
+    /// Computes an managed envelope from the given envelope and binds it to the processing queue.
     ///
     /// To provide additional scoping, use [`EnvelopeContext::scope`].
```
```suggestion
    /// To provide additional scoping, use [`ManagedEnvelope::scope`].
```

PR #1938, opened 2023-03-15
ref(server): Move from actix-web to axum

jan-auer:
Replaces `actix-web` with [`axum`](https://docs.rs/axum/) in the web endpoint layer and upgrades related dependencies.

## Motivation

Relay has used `actix-web` as web framework. Due to a failed upgrade attempts in the past, Relay has remained on a version from 2018, which relies on outdated dependencies and predates modern language features such as `Futures` in the standard library. As such, the development experience has been suboptimal. Most importantly, however, some of the dependencies required by actix-web contain vulnerabilities or are unmaintained and advised against at this point.

The [latest version](https://docs.rs/actix-web/4.3.1/actix_web) of `actix-web` supports standard futures and has added support for the stable version of `tokio`, the async runtime. However, the differences between these versions are large enough to consider other web frameworks.

The community currently prefers `axum`, which is developed closer to the `tokio` project and provides quicker upgrade paths on more recent tokio versions. Additionally, `axum` is a relatively thin routing layer built on top of `hyper`, which is already used inside `reqwest` as our HTTP client for outbound requests. Additionally, `axum` is built on top of `tower`, a composable service library with wide-spread community support offering building blocks for common HTTP functionality.

## Details

Actix-web required a proprietary system executor built on top of an older version of the `tokio` runtime. It runs a central accept loop and an additional thread pool to handle web requests. This system has been removed. Web requests are now handled within the main executor.

Similar to actix-web, axum offers handler functions for endpoints and extractors to resolve structured information from web requests. Changes to the prior implementation include:
- The maximum body size is no longer checked directly in the extractor or endpoint implementation. Instead, this is now handled via a body size limit middleware.
- The server no longer consumes request bodies if the request is terminated early. This can be expanded to more cases in follow-ups to speed up request handling.

Hyper offers low-level SSL utilities. Achieving SSL support via `native-tls` is slightly more involving, and we are not yet ready to move to `rustls`. Since the best practices in operating guidelines require to use a reverse proxy in front of Relay, we're dropping native SSL support. Outbound HTTPS support and TLS for redis and kafka are now enabled by default.

As part of this migration, we're embracing extraction more than before. Instead of functional utilities, many inputs to endpoint handlers composed from and written as extractors, which reduces endpoints to a minimal set of processing logic, mostly delegating to services.

Axum's router is less capable than actix-web. Most notably, endpoints that used to fall back to forwarding now need to handle forwarding explicitly. For this reason, the forward endpoint is written as a function that can be called directly on a request for convenience. Other differences are minor, for instance that non-numeric project IDs are now reported as error instead of 404.

## Dependencies

Since the actix-web framework was the last dependency to rely on `failure`, `futures 0.1` and `bytes 0.4`, these dependencies have now been removed from the dependency tree. This also resolves outstanding security issues and ensures Relay continues to build with upcoming compiler releases.

There is still a number of duplicate dependencies in the tree, which can be reviewed separately.

## Performance Impact

Performance was tested using the load testing infrastructure. CPU usage is similar to the actix-web baseline, usually even faster. Same for request roundtrip times. Memory consumption is 20% less.

Note that most of Relay's resource consumption occurs outside of HTTP request handling, either in endpoint logic or internal services.

## Breaking Changes

This section is also part of the changelog.

Configuration:
- SSL support has been dropped. As per [official guidelines](https://docs.sentry.io/product/relay/operating-guidelines/), Relay should be operated behind a reverse proxy, which can perform SSL termination.
- Connection config options `max_connections`, `max_pending_connections`, and `max_connection_rate` no longer have an effect. Instead, configure the reverse proxy to handle connection concurrency as needed.

Endpoints:
- The security endpoint no longer forwards to upstream if the mime type doesn't match supported mime types. Instead, the request is rejected with a corresponding error.
- Passing store payloads as `?sentry_data=<base64>` query parameter is restricted to `GET` requests on the store endpoint. Other endpoints require the payload to be passed in the request body.
- Requests with an invalid `content-encoding` header will now be rejected. Exceptions to this are an empty string and `UTF-8`, which have been sent historically by some SDKs and are now treated as identity (no encoding). Previously, all unknown encodings were treated as identity.
- Temporarily, response bodies for some errors are rendered as plain text instead of JSON. This will be addressed in an upcoming release.

Metrics:
- The `route` tag of request metrics uses the route pattern instead of schematic names. There is an exact replacement for every previous route. For example, `"store-default"` is now tagged as `"/api/:project_id/store/"`.
- Statsd metrics `event.size_bytes.raw` and `event.size_bytes.uncompressed` have been removed.

## Tasks

- [x] Replace actix-web and actix with axum, hyper, and tower in dependencies
- [x] Upgrade related dependencies, such as `bytes`
- [x] Remove the actix system and subthread from the `HttpServer` service
- [x] Configure the axum/hyper server appropriately
  - [x] Remove SSL support and leave a validation error
  - [x] Add notes about SSL removal everywhere and update docs
  - [x] Support graceful shutdown
  - [x] Configure HTTP settings
  - [x] Configure connection concurrency
- [x] Migrate middlewares
  - [x] Metrics
  - [x] Start time
  - [x] Sentry error logging
  - [x] Logging
  - [x] Panic handling (via [tower-http](https://docs.rs/tower-http/latest/tower_http/catch_panic/index.html))
- [x] Migrate extractors
  - [x] Start time
  - [x] Forwarded for header
  - [x] Request meta
  - [x] Envelope meta
  - [x] Signed JSON
- [x] Migrate body utilties
  - [x] Store body decompression
  - [x] Decompression (may require [tower-http](https://docs.rs/tower-http/latest/tower_http/decompression/index.html))
  - [x] Body peeking
- [x] Migrate endpoints
  - [x] Health check
  - [x] Events (testing)
  - [x] Project configs (note forward predicate)
  - [x] Public keys
  - [x] Outcomes
  - [x] Store (legacy)
  - [x] Store
  - [x] Envelope
  - [x] Security & CSP
  - [x] Minidump (note missing trailing slash)
  - [x] Attachments
  - [x] Unreal
  - [x] Forward
- [x] Migrate endpoint registration
- [x] Ensure request body decompression is operational
  - [x] `gzip`
  - [x] `deflate`
  - [x] `br`
- [x] Ensure equivalence of tracked metrics
  - [x] Route names on generic request metrics
  - [x] Track `RequestSizeBytesRaw` and `RequestSizeBytesUncompressed`
- [x] Port details of invalid HTTP spec handling
  - [x] Ensure robust routing
  - [x] Ensure body limits are applied on all endpoints
  - [x] Ensure connection keep-alive works: body is consumed and connection is not stalled on error
  - [x] Ensure multipart is resilient against missing newlines
  - [x] Check occurrences of 415 status codes
- [x] Load test using `test-factory`

## Follow-up

These tasks are not part of this PR. They can be addressed independently in smaller follow-ups:

Web layer behavior:
- Render all response errors as JSON

Web layer ergonomics:
- Refactor `ApiErrorResponse` to include the status code
- Switch to `multer` directly instead of axum's Multipart extractor
- Refactor auth extraction (`RequestMeta` and `PartialMeta`)
- Generate errors for all extractors instead of unwrapping internally: `ForwardedFor`, `StartTime`
- Bundle all request decompression middleware into a single Layer and Server

Performance / Correctness:
- Fix "dropped envelope: internal error" error ([RELAY-CRJ](https://sentry.my.sentry.io/organizations/sentry/issues/315716))
- Benchmark with HTTP/1 half-close enabled
- Stop streaming request bodies earlier upon errors (defer extraction)

Service layer:
- Use Addrs from ServiceState instead of Registry in endpoints
- Improve `ServiceState` runtime handling

General cleanup and follow-up:
- Move from `log` to `tracing`
- Check usage of the `url` crate in favor of reqwest/axum's URL
- Remove HTTP utilities that wrap upstream responses (directly expose `http` types)


relay-server/src/endpoints/attachments.rs
```diff
@@ -1,54 +1,49 @@
...
-async fn store_attachment(
+async fn handle(
+    state: ServiceState,
     meta: RequestMeta,
-    request: HttpRequest<ServiceState>,
-) -> Result<HttpResponse, BadStoreRequest> {
-    let envelope = extract_envelope(&request, meta).await?;
-    common::handle_envelope(request.state(), envelope).await?;
-    Ok(HttpResponse::Created().finish())
+    Path(path): Path<AttachmentPath>,
```
jjbayer, line 40: nit: Why not just
```suggestion
    Path(event_id): Path<EventId>,
```
?

jjbayer: The route for this endpoint also has a `:projectId` path parameter, do we ignore that one?

relay-server/src/endpoints/security_report.rs
```diff
@@ -1,103 +1,92 @@
...
+struct SecurityReportQuery {
     sentry_release: Option<String>,
     sentry_environment: Option<String>,
 }
 
-async fn extract_envelope(
-    request: &HttpRequest<ServiceState>,
+#[derive(Debug, FromRequest)]
+#[from_request(state(ServiceState))]
+struct SecurityReportParams {
```
jjbayer, line 25: As discussed offline, I would rename this to `SecurityReportRequest`, because that is what it represents IMO (same goes for `*Params` structs in other endpoints).

relay-server/src/endpoints/security_report.rs
```diff
@@ -1,103 +1,92 @@
...
-                | ("application", "expect-ct-report", Some("json"))
-                | ("application", "expect-staple-report", None)
-        )
+async fn handle(
+    state: ServiceState,
+    mime: Mime,
+    params: SecurityReportParams,
+) -> Result<impl IntoResponse, BadStoreRequest> {
+    if !is_security_mime(mime) {
+        return Ok(StatusCode::UNSUPPORTED_MEDIA_TYPE.into_response());
```
jjbayer, line 82: I assume the actix route with the `.filter` used to respond with 404 here? If so, are we OK with changing that behavior?

jjbayer (commented):
Overall this looks good to me! Left a few questions. I wonder if we should split this PR into batches of endpoints, so we can start merging some of it.

relay-server/src/endpoints/attachments.rs
jan-auer: The project_id is used by `RequestMeta` to check authentication. This is also why we need a struct that matches the `event_id` path param, so that it doesn't fail deserializing when there are two parameters (it wouldn't know which one to choose).

relay-server/src/endpoints/security_report.rs
jan-auer: See the PR description, this is actually a breaking change. The former `filter` skipped the route, which means the request was forwarded to the upstream. We could consider bringing this back, but it looks like this is not needed in practice.

jan-auer:
As it turns out, hyper does not exhaust request bodies if they are dropped by the "user" (i.e. endpoint implementation). There's an optimistic implementation that tries to read exactly one chunk, but for larger requests, it closes the connection:

- https://github.com/hyperium/hyper/blob/47f614fd99bf9e937efc874880e3276caf4c34b1/src/proto/h1/dispatch.rs#L196-L199
- https://github.com/hyperium/hyper/blob/47f614fd99bf9e937efc874880e3276caf4c34b1/src/proto/h1/dispatch.rs#L209-L212
- https://github.com/hyperium/hyper/blob/47f614fd99bf9e937efc874880e3276caf4c34b1/src/proto/h1/conn.rs#L741-L747

This means that we'll have to provide a middleware similar to what we used to do in actix. This time, this would be possible as a tower layer.

relay-server/src/actors/processor.rs
```diff
@@ -2514,9 +2514,10 @@ impl EnvelopeProcessorService {
                 encoder.finish()?
             }
             HttpEncoding::Br => {
-                let mut encoder = BrotliEncoder::new(Vec::new(), 5);
+                // Use default buffer size (via 0), medium quality (5), and the default lgwin (22).
+                let mut encoder = BrotliEncoder::new(Vec::new(), 0, 5, 22);
```
jan-auer, line 2518: We had to switch from `brotli2` to `brotli`. The decompression middleware from `tower-http` is built on the latter, and having both dependencies in your binary does not compile.

relay-server/src/actors/server.rs
```diff
@@ -18,202 +22,123 @@ use crate::statsd::RelayCounters;
...
+        // TODO(ja): These are unsupported for now:
+        //     .maxconn(config.max_connections())
+        //     .maxconnrate(config.max_connection_rate())
+        //     .backlog(config.max_pending_connections())
+
+        let handle = Handle::new();
+        let server = axum_server::bind(config.listen_addr())
+            .http_config(http_config)
+            .addr_incoming_config(addr_config)
+            .handle(handle.clone());
```
jan-auer: To simplify startup even further, I'm thinking of moving the server into the spawn function, so that everything is contained within one function. In the constructor, we can keep binding to the socket to catch occupied ports before starting the web server.

relay-server/src/constants.rs
```diff
@@ -25,3 +25,6 @@ pub const UNREAL_USER_HEADER: &str = "unreal_user_id";
 /// The default retention for events if the server does not specify a value in project
 /// configurations.
 pub const DEFAULT_EVENT_RETENTION: u16 = 90;
+
+/// Maximum size of JSON request bodies.
+pub const MAX_JSON_SIZE: usize = 262_144;
```
jan-auer, line 30: Moved from `SignedJson` so that it can be used for `DefaultBodyLimit` in middlewares.

relay-server/src/body/store_body.rs
```diff
@@ -1,71 +0,0 @@
...
-        return None;
-    }
-
-    let (_, value) = form_urlencoded::parse(req.query_string().as_bytes())
-        .find(|(key, _)| key == "sentry_data")?;
-
-    Some(value)
-}
-
-fn decode_bytes<B>(body: B, limit: usize) -> Result<Bytes, PayloadError>
```
jan-auer, line 49: Querystring and decoding have moved to the `store` endpoint, which is the only endpoint that supports this encoding.

relay-server/src/body/peek_line.rs
```diff
@@ -1,144 +0,0 @@
...
-///
-/// # Cancel Safety
-///
-/// This function is _not_ cancellation safe. If canceled, partially read data may not be put back
-/// into the request. Additionally, it is not safe to read the body after an error has been returned
-/// since data may have been partially consumed.
-///
-/// [`Payload`]: actix_web::dev::Payload
-/// [`SharedPayload`]: crate::extractors::SharedPayload
-pub async fn peek_line<S>(
```
jan-auer, line 29: This is no longer needed as `EnvelopeMeta` is now directly extracted in the endpoint along with the body.

relay-server/src/endpoints/common.rs
```diff
@@ -225,39 +242,6 @@ pub fn event_id_from_items(items: &Items) -> Result<Option<EventId>, BadStoreReq
     Ok(None)
 }
 
-/// Creates a preconfigured CORS middleware builder for store requests.
-///
-/// To configure CORS, register endpoints using `resource()` and finalize by calling `register()`, which
-/// returns an App. This configures POST as allowed method, allows default sentry headers, and
-/// exposes the return headers.
-pub fn cors(app: App<ServiceState>) -> CorsBuilder<ServiceState> {
```
jan-auer, line 233: The code has moved to middlewares as it has become a layer, but is still registered as route layer on the store endpoints.

relay-server/src/endpoints/common.rs
```diff
@@ -373,68 +357,26 @@ pub async fn handle_envelope(
...
-        .body(id.0.as_hyphenated().to_string())
-}
+#[derive(Debug)]
+pub struct TextResponse(pub Option<EventId>);
 
-/// A helper for creating Actix routes that are resilient against double-slashes
-///
-/// Write `normpath("api/store")` to create a route pattern that matches "/api/store/",
-/// "api//store", "api//store////", etc.
-pub fn normpath(route: &str) -> String {
```
jan-auer, line 393: This has moved to middelwares. Normalizing the path requires a middleware that _wraps_ the router, see `server.rs` which explains this in a comment.

relay-server/src/endpoints/envelope.rs
```diff
@@ -1,27 +1,114 @@
...
+
+        if !matches!(result, Err(BadEventMeta::MissingAuth)) {
+            return Ok(Self {
+                meta: result?,
+                body: request.extract().await?,
+            });
+        }
+
+        // TODO(ja): Improve PartialMeta / RequestMeta to parse Auth and meta separately, then merge
+        // into RequestMeta.
```
jan-auer: This will likely be left for a follow up and the TODO removed. The way `PartialMeta` and `RequestMeta` are being parsed is highly unergonomic. I would appreciate some input on how this could be structured well.

relay-server/src/endpoints/envelope.rs
```diff
@@ -31,21 +118,20 @@ struct StoreResponse {
...
+    let id = common::handle_envelope(&state, envelope).await?;
+    Ok(Json(StoreResponse { id }))
+}
+
+pub fn route<B>(config: &Config) -> MethodRouter<ServiceState, B>
+where
+    B: axum::body::HttpBody + Send + 'static,
+    B::Data: Send,
+    B::Error: Into<axum::BoxError>,
+{
```
jan-auer, line 133: The route function has an unwieldy signature due to the generic body (see `endpoints/mod.rs` for a comment on that). On the upside, it hides route layers and private extractors from the public signature. 

I think there's no right or wrong, but the benefit of having a `route` function here is that it makes the overall router configuration very readable and concise.

relay-server/src/endpoints/forward.rs
```diff
@@ -218,87 +188,97 @@ impl UpstreamRequest for ForwardRequest {
...
+/// the request body back to the origin. Regardless of the incoming connection, the connection to
+/// the upstream uses its own HTTP version and transfer encoding.
+///
+/// # Usage
+///
+/// This endpoint is both a handler and a request function:
+///
+/// - Use it as [`Handler`] directly in router methods when registering this as a route.
+/// - Call this manually from other request handlers to conditionally forward from other endpoints.
+pub fn forward<B>(state: ServiceState, req: Request<B>) -> impl Future<Output = Response>
```
jan-auer, line 276: This is a special case, see the doc comment. 

This function is necessary because it dynamically adds a layer on the handler based on request information (the path). I have not found a more concise way to implement this.

relay-server/src/endpoints/minidump.rs
```diff
@@ -60,56 +65,38 @@ fn infer_attachment_type(field_name: Option<&str>) -> AttachmentType {
...
-        .into_future()
-        .compat();
-
-    let (data, _) = future
-        .await
-        .map_err(|(err, _)| BadStoreRequest::InvalidMultipart(err))?;
-
-    Ok(data.map(Bytes::from))
+    let stream = future::ok::<_, Infallible>(payload.clone()).into_stream();
+    let mut multipart = multer::Multipart::new(stream, boundary);
```
jan-auer, line 74: Here, we're extracting a custom multipart body from an already obtained payload (this is *not* nested multipart!). Since the `axum` extractor has no constructor that allows us to pass it a payload directly, we have to use `multer` directly.

Eventually, we can disable the multipart feature in axum and always use our own multipart extractor built directly with `multer`. This would also allow us to configure some of the internal limits. The `multipart` utilities file has a few hacks for this, right now.

relay-server/src/endpoints/mod.rs
```diff
@@ -18,27 +18,56 @@ mod statics;
...
-        .configure(attachments::configure_app)
-        .configure(unreal::configure_app)
-        // `forward` must be last as it creates a wildcard proxy
-        .configure(forward::configure_app)
+#[rustfmt::skip]
+pub fn routes<B>(config: &Config) -> Router<ServiceState, B>
+where
+    B: axum::body::HttpBody + Send + 'static,
+    B::Data: Send + Into<Bytes>,
+    B::Error: Into<axum::BoxError>,
```
jan-auer, line 34: Unfortunately, with the decompression middleware, we're no longer using the default axum Body type. This means we either have to:
- make all functions generic
- declare the concrete body type that changes with middlewares
- move routes directly into server startup

I've opted with the first approach to keep the endpoints somewhat independent from middlewares. It seems this is not really intended by axum, so we can also try to follow one of the other approaches.

relay-server/src/endpoints/mod.rs
```diff
@@ -18,27 +18,56 @@ mod statics;
...
+    let store_routes = Router::new()
+        // Legacy store path that is missing the project parameter.
+        .route("/api/store/", store::route(config))
+        .route("/api/:project_id/store/", store::route(config))
+        .route("/api/:project_id/envelope/", envelope::route(config))
+        .route("/api/:project_id/security/", security_report::route(config))
+        .route("/api/:project_id/csp-report/", security_report::route(config))
+        // No mandatory trailing slash here because people already use it like this.
+        .route("/api/:project_id/minidump", minidump::route(config))
+        .route("/api/:project_id/minidump/", minidump::route(config))
```
jan-auer, line 62: Note how these are now two routes. Before, the trailing slash was made optional by `normpath`.

relay-server/src/extractors/mime.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use axum::extract::FromRequestParts;
+use axum::http::request::Parts;
+use axum::http::{header, StatusCode};
+use axum::response::IntoResponse;
+
+use crate::utils::ApiErrorResponse;
+
+#[derive(Clone, Debug)]
+pub struct Mime(mime::Mime);
```
jan-auer, line 12: axum provides typed header extractors via the `headers` crate, which has an opaque `ContentType` header (wrapping mime). For security reports, we need to inspect the parsed mime however, so we need to create our own here.

tests/integration/test_security_report.py
```diff
@@ -176,42 +180,40 @@ def test_security_report_cors(mini_sentry, relay):
...
+    resp = relay.send_security_report(
+        project_id=42,
+        content_type="application/json; charset=utf-8",
+        payload="",  # Invalid, but irrelevant for this test
+        release="01d5c3165d9fbc5c8bdcf9550a1d6793a80fc02b",
+        environment="production",
+        origin="http://valid.com",
+    )
+
+    assert set(split_header(resp.headers["access-control-expose-headers"])) == set(
```
jan-auer, line 214: The old test was actually wrong. `access-control-expose-headers` is supposed to go on the response of the actual request, but not on the preflight.

tests/integration/test_store.py
```diff
@@ -72,8 +72,11 @@ def test_options_response(mini_sentry, relay, method_to_test):
     }
 
     result = relay.send_options(project_id, headers)
-
-    assert result.ok == should_succeed
+    assert result.ok, result
```
jan-auer, line 74: The previous test asserted invalid behavior. The preflight should always succeed, its response headers declare whether the stated request would be valid.

jan-auer (commented):
There are a few comments marked with "TODO(ja)" that I will still resolve before merging this.

relay-server/src/actors/server.rs
jjbayer: That sounds good to me. It looks like the fields on `Self` are not being used outside of `spawn_handler`, so they might as well be local.

relay-server/src/endpoints/mod.rs
jjbayer: As discussed offline, a meta trait would be nice here, something like

```rust
trait RelayBody: axum::body::HttpBody + Send + 'static {}
```

relay-server/src/endpoints/minidump.rs
```diff
@@ -1,17 +1,22 @@
...
 use crate::constants::{ITEM_NAME_BREADCRUMBS1, ITEM_NAME_BREADCRUMBS2, ITEM_NAME_EVENT};
-use crate::endpoints::common::{self, BadStoreRequest};
+use crate::endpoints::common::{self, BadStoreRequest, TextResponse};
 use crate::envelope::{AttachmentType, ContentType, Envelope, Item, ItemType};
-use crate::extractors::RequestMeta;
+use crate::extractors::{RawContentType, RequestMeta};
 use crate::service::ServiceState;
-use crate::utils::{consume_field, get_multipart_boundary, MultipartError, MultipartItems};
+use crate::utils;
+// use crate::utils::{consume_field, get_multipart_boundary, MultipartError, MultipartItems};
```
olksdr:
```suggestion
```

can we remove this commented out code?

relay-server/src/endpoints/common.rs
```diff
@@ -86,30 +107,32 @@ impl BadStoreRequest {
...
                 // now executed asynchronously in `EnvelopeProcessor`.
-                HttpResponse::Forbidden().json(&body)
-            }
-            BadStoreRequest::PayloadError(e) if matches!(e.get_ref(), PayloadError::Overflow) => {
-                HttpResponse::PayloadTooLarge().json(&body)
+                (StatusCode::FORBIDDEN, body).into_response()
             }
+            // BadStoreRequest::PayloadError(e) if matches!(e.get_ref(), PayloadError::Overflow) => {
+            //     HttpResponse::PayloadTooLarge().json(&body)
+            // }
```
jjbayer: What happened to this variant?

relay-server/src/endpoints/common.rs
```diff
@@ -86,30 +107,32 @@ impl BadStoreRequest {
...
             _ => {
                 // In all other cases, we indicate a generic bad request to the client and render
                 // the cause. This was likely the client's fault.
-                HttpResponse::BadRequest().json(&body)
+                (StatusCode::BAD_REQUEST, body).into_response()
```
jjbayer, line 132: I realize this did not change in this PR, but I would like if we matched on all other `BadStoreRequest` variants explicitly here, to make it visible exactly what cases we consider a client error.

relay-server/src/extractors/signed_json.rs
```diff
@@ -70,42 +70,34 @@ fn get_header<'a, S>(
...
+            .parse::<RelayId>()
+            .map_err(|_| SignatureError::MalformedHeader("x-sentry-relay-id"))?;
 
-    let body = body::request_body(&request, MAX_JSON_SIZE).await?;
-
-    let inner = relay
-        .public_key
-        .unpack(&body, &signature, None)
-        .map_err(SignatureError::from)?;
+        // Dump out header value even if not string
```
olksdr:
```suggestion
        // Dump out header value even if it's not a string
```

relay-server/src/endpoints/common.rs
```diff
@@ -373,68 +357,26 @@ pub async fn handle_envelope(
...
-    debug_assert!(!id.is_nil());
-
-    // the minidump client expects the response to contain an event id as a hyphenated UUID
-    // i.e. xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
-    HttpResponse::Ok()
-        .content_type("text/plain")
-        .body(id.0.as_hyphenated().to_string())
-}
+#[derive(Debug)]
+pub struct TextResponse(pub Option<EventId>);
```
jjbayer, line 358: If we always `unwrap_or_default` with a debug assert in `into_response`, why even allow a `None` value here?

relay-server/src/middlewares/mod.rs
```diff
@@ -0,0 +1,13 @@
+mod cors;
```
olksdr, line 10: Maybe some small top level doc here would be good to have.

olksdr (approved):
🚢  it!

relay-server/src/endpoints/project_configs.rs
```diff
@@ -172,14 +151,20 @@ async fn get_project_configs(
     Ok(Json(GetProjectStatesResponseWrapper { configs, pending }))
 }
 
-pub fn configure_app(app: App<ServiceState>) -> App<ServiceState> {
-    app.resource("/api/0/relays/projectconfigs/", |r| {
-        r.name("relay-projectconfigs");
-        r.post()
-            .filter(VersionPredicate)
-            .with_async(|b, v| Box::pin(get_project_configs(b, v)).compat());
+fn predicate(Query(query): Query<VersionQuery>) -> bool {
```
jjbayer:
```suggestion
fn has_compatible_version(Query(query): Query<VersionQuery>) -> bool {
```

relay-server/src/endpoints/project_configs.rs
```diff
@@ -172,14 +151,20 @@ async fn get_project_configs(
...
-        r.post().f(crate::endpoints::forward::forward_compat);
+pub async fn handle<B>(state: ServiceState, mut req: Request<B>) -> Result<impl IntoResponse>
+where
+    B: axum::body::HttpBody + Send + 'static,
+    B::Data: Send,
+    B::Error: Into<axum::BoxError>,
+{
+    let data = req.extract_parts().await?;
+    Ok(if predicate(data) {
+        inner.call(req, state).await
```
jjbayer, line 178: nit: Why is the inner function called `inner` now, and why do we call it with `call`?

```suggestion
        get_project_configs(req, state).await
```

relay-server/src/utils/unreal.rs
```diff
@@ -83,7 +83,7 @@ pub fn expand_unreal_envelope(
 
         let mut item = Item::new(ItemType::Attachment);
         item.set_filename(file.name());
-        item.set_payload(content_type, file.data());
+        item.set_payload(content_type, file.data().to_owned());
```
jjbayer, line 87: Does this make a copy of the data? Can we move it in somehow instead?

jjbayer (approved):
Left a couple of comments, none of them are blocking. This PR is too large to catch subtle bugs during review, but overall this looks good to me!

relay-server/src/endpoints/common.rs
jan-auer: Good point. The original error does no longer exist. Instead, this is caught earlier in extractors, particularly by `BytesRejection`.

relay-server/src/endpoints/common.rs
jan-auer: Good point. Doing it now will just list a ton of errors without additional values, I'm afraid. But instead we can start to remove error variants from `BadStoreRequest` and instead make more specific errors for the individual endpoints. Most of these errors should over time end up in extractor rejections.

relay-server/src/endpoints/common.rs
jan-auer: Unfortunately, we need to allow `None` here because of static typing. Our endpoints need to be defensive, which is why we're not going to error in production in case the event ID ends up missing. In tests we'd of course still know about this bug. 

I haven't found a good way to address this yet in a type-safe manner. Maybe we could start to type envelopes into event-carrying and event-less Envelopes, one of which have a guaranteed event_id. That's a much bigger change, though.

relay-server/src/utils/unreal.rs
jan-auer: Yes, this requires an upstream change in Symbolic. Adding a TODO for this and similar cases.

PR #1939, opened 2023-03-15
ref(protocol): make enum out of metric names

### LINKED ISSUE #1937: Create enums for builtin metric names and tags
From both [sessions](https://github.com/getsentry/relay/blob/bc974833c9cc45ce4267f70c6a4434b7c143aaf4/relay-server/src/metrics_extraction/sessions.rs) and [transactions](https://github.com/getsentry/relay/blob/bc974833c9cc45ce4267f70c6a4434b7c143aaf4/relay-server/src/metrics_extraction/transactions.rs), we currently extract a fixed set of metrics with a fixed set of tags.

These metric names and tag names are not well documented at the moment. In addition to the missing documentation, we have the problem that we want to set the metric name as a statsd tag for some statsd emissions:

https://github.com/getsentry/relay/blob/bc974833c9cc45ce4267f70c6a4434b7c143aaf4/relay-metrics/src/aggregation.rs#L1703-L1706

But with custom measurements, this statsd tag is becoming too high-cardinality to be usable.

We can solve both problems by creating `enums` for session / transaction metrics and their tags.

Easiest would be to just create 4 enums. Bonus task: Unite these enums with `MetricNamespace` somehow, to make it impossible to emit a session metric in the transaction namespace or vice versa.

TBS1996:
References #1937 

Create an enum for Session metrics and for Transaction metrics

#skip-changelog


relay-metrics/src/protocol.rs
```diff
@@ -163,6 +163,90 @@ fn is_valid_name(name: &str) -> bool {
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TransactionsKind<'a> {
+    Duration,
+    User,
+    CountPerRootProject,
+    MeasurementsFramesFrozen,
```
TBS1996: There are so many `measurements.`  variants, and there's also a place in the code which dynamically creates names that begin with `measurements.`. 

Perhaps we could create a `Measurements(String)` variant.

relay-metrics/src/aggregation.rs
```diff
@@ -1578,19 +1578,17 @@ impl AggregatorService {
     fn normalize_metric_name(key: &mut BucketKey) -> Result<(), AggregateMetricsError> {
         key.metric_name = match MetricResourceIdentifier::parse(&key.metric_name) {
             Ok(mri) => {
-                if matches!(mri.namespace, MetricNamespace::Unsupported) {
```
TBS1996: potential issue: when we had the `unsupported` variant and we failed to parse the namespace we propagated this error in specific, and for other error-cases we propagated `invalidcharacters` error. 

should we let `MetricResoureIdentifier::parse` return `aggregatemetricserror` directly instead? or perhaps create a custom error variant?

relay-metrics/src/protocol.rs
```diff
@@ -163,6 +163,121 @@ fn is_valid_name(name: &str) -> bool {
...
+            },
+            "user" => Ok(Self::User),
+            "count_per_root_project" => Ok(Self::CountPerRootProject),
+            _ => Ok(Self::Custom { name, ty, unit }),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Measurementkind<'a> {
```
TBS1996: Is there a place to get a more complete overview of the most common variants?

relay-metrics/src/protocol.rs
```diff
@@ -173,89 +288,111 @@ fn is_valid_name(name: &str) -> bool {
...
+    Session(SessionsKind),
+}
 
-    fn from_str(ns: &str) -> Result<Self, Self::Err> {
-        match ns {
-            "sessions" => Ok(MetricNamespace::Sessions),
-            "transactions" => Ok(MetricNamespace::Transactions),
-            _ => Ok(MetricNamespace::Unsupported),
+impl<'a> MetricResourceIdentifier<'a> {
+    pub fn unit(&self) -> MetricUnit {
```
TBS1996: we have to be careful that it returns the right unit here

relay-metrics/src/protocol.rs
```diff
@@ -173,89 +288,111 @@ fn is_valid_name(name: &str) -> bool {
...
+    pub fn name(&self) -> String {
+        match self {
+            Self::Transaction(t) => format!("{t}"),
+            Self::Session(s) => format!("{s}"),
+        }
+    }
 
-impl fmt::Display for MetricNamespace {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    pub fn ty(&self) -> MetricType {
```
TBS1996: likewise, have to be careful that it returns the right metrictype here. If a variant always returns hte same type but in reality it can be different types then we have to encode the type in the variant itself

relay-metrics/src/protocol.rs
```diff
@@ -173,89 +288,111 @@ fn is_valid_name(name: &str) -> bool {
...
+                SessionsKind::Session => MetricType::Counter,
+                SessionsKind::Error => MetricType::Set,
+                SessionsKind::User => MetricType::Set,
+            },
         }
     }
+
+    /// Parses and validates an MRI of the form `<ty>:<ns>/<name>@<unit>`
+    pub fn parse(s: &'a str) -> Result<Self, ParseMetricError> {
+        //<ty>    |    <ns>/<name>@<unit>
```
TBS1996: are these comments annoying or helpful?

relay-metrics/src/protocol.rs
```diff
@@ -173,89 +288,111 @@ fn is_valid_name(name: &str) -> bool {
...
+                    "error" => SessionsKind::Error,
+                    _ => return Err(ParseMetricError(())),
+                };
+                Ok(Self::Session(kind))
+            }
+            _ => Err(ParseMetricError(())),
+        }
+    }
+
+    fn format_from_params(ns: &str, name: &str, unit: MetricUnit, ty: MetricType) -> String {
```
TBS1996: normalization from the statsd format

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -333,12 +333,9 @@ fn extract_transaction_metrics_inner(
                         None => continue,
                     };
 
-                    let unit = measurement.unit.value();
```
TBS1996, line 336: not needed as `breakdowns` should always be MetricUnit::Duration::Milliseconds

relay-metrics/src/aggregation.rs
jjbayer: See comment below.

relay-metrics/src/aggregation.rs
```diff
@@ -1578,19 +1579,17 @@ impl AggregatorService {
...
                 let mut metric_name = mri.to_string();
                 // do this so cost tracking still works accurately.
                 metric_name.shrink_to_fit();
                 metric_name
             }
             Err(_) => {
-                relay_log::debug!("invalid metric name {:?}", key.metric_name);
-                return Err(AggregateMetricsErrorKind::InvalidCharacters.into());
+                relay_log::debug!("invalid metric namespace {:?}", key.metric_name);
+                return Err(AggregateMetricsErrorKind::UnsupportedNamespace.into());
```
jjbayer: Let's unite all parse errors into a new `InvalidIdentifier` kind.

```suggestion
                relay_log::debug!("invalid metric identifier {:?}", key.metric_name);
                return Err(AggregateMetricsErrorKind::InvalidIdentifier.into());
```

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
+}
 
-    fn from_str(ns: &str) -> Result<Self, Self::Err> {
-        match ns {
-            "sessions" => Ok(MetricNamespace::Sessions),
-            "transactions" => Ok(MetricNamespace::Transactions),
-            _ => Ok(MetricNamespace::Unsupported),
+/// Enumerates the most common transaction-names, with a catch-all for any other names.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TransactionsKind<'a> {
```
jjbayer: naming: 
```suggestion
pub enum TransactionMetric<'a> {
```

relay-metrics/src/protocol.rs
jjbayer: Yes, https://github.com/getsentry/sentry/blob/9d4330eb11c7a3b216deb5dd2050fcf12b7ab408/src/sentry/relay/config/measurements.py#L26-L48

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
+/// The namespace of a metric.
+///
+/// Namespaces allow to identify the product entity that the metric got extracted from, and/or
+/// identify the use case that the metric belongs to. These namespaces cannot be defined freely,
+/// instead they are defined by Sentry. Over time, there will be more namespaces as we introduce
+/// new metrics-based products.
+///
+/// Right now this successfully deserializes any kind of string, but in reality only `"sessions"`
+/// (for release health) and `"transactions"` (for metrics-enhanced performance) is supported.
+/// Everything else is dropped both in the metrics aggregator and in the store service.
```
jjbayer: This paragraph should be moved to wherever we parse the namespace.

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
         }
     }
 }
 
+/// The namespace of a metric.
+///
+/// Namespaces allow to identify the product entity that the metric got extracted from, and/or
+/// identify the use case that the metric belongs to. These namespaces cannot be defined freely,
+/// instead they are defined by Sentry. Over time, there will be more namespaces as we introduce
+/// new metrics-based products.
```
jjbayer: Please move this to `fn namespace()`

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
+                TransactionsKind::User => MetricType::Set,
+            },
+            Self::Session(s) => match s {
+                SessionsKind::Session => MetricType::Counter,
+                SessionsKind::Error => MetricType::Set,
+                SessionsKind::User => MetricType::Set,
+            },
+        }
+    }
+    pub fn namespace(&self) -> &str {
```
jjbayer: What if we keep the `Namespace` enum and restore its documentation, and return that here?

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
+    pub fn ty(&self) -> MetricType {
+        match self {
+            Self::Transaction(t) => match t {
+                TransactionsKind::Custom { ty, .. } => *ty,
+                TransactionsKind::Breakdowns(_) => MetricType::Distribution,
+                TransactionsKind::CountPerRootProject => MetricType::Counter,
+                TransactionsKind::Duration(_) => MetricType::Distribution,
+                TransactionsKind::Measurements { .. } => MetricType::Distribution,
+                TransactionsKind::User => MetricType::Set,
+            },
```
jjbayer: Let's move this to `TransactionKind::ty()` (same for `SessionsKind::ty()`).

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
+    pub fn unit(&self) -> MetricUnit {
+        match self {
+            Self::Transaction(t) => match t {
+                TransactionsKind::Custom { unit, .. } => *unit,
+                TransactionsKind::Measurements { unit, .. } => *unit,
+                TransactionsKind::Duration(unit) => MetricUnit::Duration(*unit),
+                TransactionsKind::Breakdowns(_) => MetricUnit::Duration(DurationUnit::MilliSecond),
+                TransactionsKind::User => MetricUnit::None,
+                TransactionsKind::CountPerRootProject => MetricUnit::None,
+            },
```
jjbayer: Please put a `unit()` method on `TransactionsKind` and return it here. The transaction-specific logic should be as close to the transaction-specific type as possible.

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
-    /// aggregating. For instance, an MRI of `c:something_new/foo@none` has the namespace
-    /// `something_new`, but as Relay doesn't support that namespace, it gets deserialized into
-    /// this variant.
-    ///
-    /// Relay currently drops all metrics whose namespace ends up being deserialized as
-    /// `unsupported`. We may revise that in the future.
-    Unsupported,
+/// Enumerates the most common session-names.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum SessionsKind {
```
jjbayer:
```suggestion
pub enum SessionMetric {
```

relay-metrics/src/protocol.rs
jjbayer: I would remove them, but use the names from the comments in the actual variable names, e.g. `let (ty, ns_name_unit) = ...`

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
+            Self::User => format!("user"),
+            Self::CountPerRootProject => format!("count_per_root_project"),
+            Self::Measurements { kind, .. } => format!("{kind}"),
+            Self::Custom { name, .. } => format!("{name}"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Measurementkind<'a> {
```
jjbayer:
```suggestion
pub enum Measurement<'a> {
```

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
-            "transactions" => Ok(MetricNamespace::Transactions),
-            _ => Ok(MetricNamespace::Unsupported),
+/// Enumerates the most common transaction-names, with a catch-all for any other names.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TransactionsKind<'a> {
+    User,
+    Duration(DurationUnit),
+    CountPerRootProject,
+    Breakdowns(&'a str),
+    Measurements {
```
jjbayer:
```suggestion
    Measurement {
```

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
-            "sessions" => Ok(MetricNamespace::Sessions),
-            "transactions" => Ok(MetricNamespace::Transactions),
-            _ => Ok(MetricNamespace::Unsupported),
+/// Enumerates the most common transaction-names, with a catch-all for any other names.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TransactionsKind<'a> {
+    User,
+    Duration(DurationUnit),
+    CountPerRootProject,
+    Breakdowns(&'a str),
```
jjbayer:
```suggestion
    Breakdown(&'a str),
```

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
+    Measurements {
+        kind: Measurementkind<'a>,
+        unit: MetricUnit,
+    },
+    Custom {
```
jjbayer: Is this variant instantiated anywhere? I think this should move into `MeasurementKind`. That is, there are custom _measurements_, but not fully custom transaction metrics.

relay-metrics/src/protocol.rs
```diff
@@ -163,88 +164,225 @@ fn is_valid_name(name: &str) -> bool {
...
+/// Enumerates the most common transaction-names, with a catch-all for any other names.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TransactionsKind<'a> {
+    User,
+    Duration(DurationUnit),
+    CountPerRootProject,
+    Breakdowns(&'a str),
+    Measurements {
+        kind: Measurementkind<'a>,
+        unit: MetricUnit,
```
jjbayer: The unit should be defined by the `MeasurementKind`. Custom measurements (see comment below) should have a configurable unit though.

```suggestion
```

jjbayer (changes requested):
This is starting to look good!

relay-metrics/src/aggregation.rs
```diff
@@ -2615,56 +2606,85 @@ mod tests {
...
+    #[test]
+    fn test_capped_iter_empty() {
+        let buckets = vec![];
+
+        let mut iter = CappedBucketIter::new(buckets.into_iter(), 200);
+        assert!(iter.next().is_none());
+    }
+
     #[test]
     fn test_aggregator_cost_tracking() {
```
TBS1996: im using `breakdowns` variant here as it allows custom names, which is required here to make the bucketkeys different from each other

relay-metrics/src/aggregation.rs
```diff
@@ -2615,56 +2606,85 @@ mod tests {
...
         };
+
         let fixed_cost = bucket_key.cost() + mem::size_of::<BucketValue>();
         for (metric_name, metric_value, expected_added_cost) in [
             (
-                "c:transactions/foo@none",
+                "c:transactions/breakdowns.foo_counter@none",
                 MetricValue::Counter(42.),
-                fixed_cost,
+                fixed_cost + 7,
```
TBS1996: I had to change these numbers to make the cost add up.. I'll look into why I had to do this. It should be clear before merging.

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,137 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TransactionsKind {
+    User,
+    Duration(DurationUnit),
+    CountPerRootProject,
+    Breakdowns(String),
+    Measurements {
+        kind: Measurementkind,
+        unit: MetricUnit,
+    },
```
jjbayer: I was wrong about measurements before. They are dynamically configured and validated by event normalization: https://github.com/getsentry/relay/blob/73ea5adc59ec80f9eb300d7717441f6262b54bc2/relay-general/src/store/normalize.rs#L352.

What this means is we do not need / should not have a `MeasurementKind`, and just use `Measurements(String, MetricUnit)` to get the name and the unit.

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,137 @@
...
+}
+
+impl FromStr for TransactionsKind {
```
jjbayer: Where is `TransactionsKind::from_str` used?

relay-metrics/src/protocol.rs
```diff
@@ -217,22 +217,31 @@ impl fmt::Display for MetricNamespace {
...
+pub struct MetricResourceIdentifier {
     /// The metric type.
     pub ty: MetricType,
     /// The namespace/usecase for this metric. For example `sessions` or `transactions`. In the
     /// case of the statsd protocol, a missing namespace is converted into the valueconverted into
     /// the value `"custom"`.
     pub namespace: MetricNamespace,
     /// The actual name, such as `duration` as part of `d:transactions/duration@ms`
-    pub name: &'a str,
+    pub name: String,
```
TBS1996: changed from &str to String because I ran in to many lifetime issues in the From<TransactionsKind> for MetricResourceIdentifier function

relay-metrics/src/protocol.rs
```diff
@@ -141,7 +141,7 @@ relay_common::impl_str_serde!(MetricType, "a metric type string");
 
 /// An error returned by [`Metric::parse`] and [`Metric::parse_all`].
 #[derive(Clone, Copy, Debug)]
-pub struct ParseMetricError(());
+pub struct ParseMetricError(pub ());
```
TBS1996: why not just make it a unit-like struct? `pub struct ParseMetricError;`

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,218 @@
...
+
+        if let Some(sdk) = value.sdk {
+            map.insert("sdk".to_string(), sdk);
+        }
+
+        map
+    }
+}
+
+impl From<SessionsErrorTags> for BTreeMap<String, String> {
```
TBS1996: I hate how wet this code is

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,218 @@
...
+use relay_common::{MetricUnit, UnixTimestamp, Uuid};
 use relay_general::protocol::{
     AbnormalMechanism, SessionAttributes, SessionErrored, SessionLike, SessionStatus,
 };
-use relay_metrics::{Metric, MetricNamespace, MetricUnit, MetricValue};
+use relay_metrics::{
+    CounterType, Metric, MetricNamespace, MetricResourceIdentifier, MetricType, MetricValue,
+};
+
+/// Enumerates the most common session-names.
```
jjbayer:
```suggestion
/// Enumerates the metrics extracted from session payloads.
```

relay-metrics/src/protocol.rs
jjbayer: I'm confident we could make it work, but let's call it low priority for now. You can also make it a [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html).

relay-metrics/src/protocol.rs
```diff
@@ -284,7 +284,7 @@ pub(crate) fn validate_tag_value(tag_value: &mut String) {
 ///
 /// Returns [`MetricUnit::None`] if no unit is specified. Returns `None` if the name or value are
 /// invalid.
-fn parse_name_unit(string: &str) -> Option<(&str, MetricUnit)> {
+pub fn parse_name_unit(string: &str) -> Option<(&str, MetricUnit)> {
```
jjbayer: Does this need to be public?

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,218 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub enum SessionsKind {
```
jjbayer: This is what it represents.

```suggestion
pub enum SessionMetric {
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,218 @@
...
+        match self {
+            SessionsKind::Session { .. } => MetricType::Counter,
+            SessionsKind::Error { .. } => MetricType::Set,
+            SessionsKind::User { .. } => MetricType::Set,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SessionsErrorTags {
```
jjbayer: The metric name is `session.error`.

```suggestion
pub struct SessionErrorTags {
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,218 @@
...
+    fn with_abnormal(
+        status: Option<SessionStatus>,
+        abnormal_mechanism: String,
+        common: (String, Option<String>, Option<String>),
+    ) -> Self {
+        let mut session = Self::new(status, common);
+        session.abnormal_mechanism = Some(abnormal_mechanism);
+        session
+    }
+}
```
jjbayer: I would personally remove these helper methods. Without them, metrics construction becomes even more explicit. And a parameter like `(String, Option<String>, Option<String>)` always harbors the risk of accidentally switching parameter values.

relay-server/src/metrics_extraction/sessions.rs
jjbayer: You could play around with `BTreeMap::from_iter`, `Option::into_iter`, `flat_map`, etc but I don't think it will help readability.

jjbayer (commented):
The `sessions.rs` part looks good to me!

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,173 @@
...
     AbnormalMechanism, SessionAttributes, SessionErrored, SessionLike, SessionStatus,
 };
-use relay_metrics::{Metric, MetricNamespace, MetricUnit, MetricValue};
+use relay_metrics::{
+    CounterType, Metric, MetricNamespace, MetricResourceIdentifier, MetricType, MetricValue,
+};
+
+/// Enumerates the metrics extracted from session payloads.
+#[derive(Clone, Debug, PartialEq)]
+pub enum SessionsMetric {
```
jjbayer:
```suggestion
pub enum SessionMetric {
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,173 @@
...
+    Session {
```
jjbayer:
```suggestion
    /// The number of sessions collected in the given time frame.
    Session {
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,173 @@
...
+        counter: CounterType,
+        timestamp: UnixTimestamp,
+        tags: SessionSessionTags,
+    },
+    User {
```
jjbayer:
```suggestion
    /// The number of unique session users for the given time frame.
    User {
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,173 @@
...
+        distinct_id: String,
+        timestamp: UnixTimestamp,
+        tags: SessionUserTags,
+    },
+    Error {
```
jjbayer:
```suggestion
    /// The number of sessions that errored in the given time frame.
    ///
    /// Because multiple session updates can be received for the same session ID,
    /// this is collected as a [`MetricType::Set`] metric rather than a simple counter.
    Error {
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,173 @@
...
+        id: Uuid,
```
jjbayer:
```suggestion
        session_id: Uuid,
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -38,125 +196,163 @@ pub fn extract_session_metrics<T: SessionLike>(
...
-            MetricValue::Counter(session.total_count() as f64),
-            timestamp,
-            with_tag(&tags, "session.status", "init"),
-        ));
+        target.push(
+            SessionsMetric::Session {
+                counter: session.total_count() as f64,
+                timestamp,
+                tags: SessionSessionTags {
+                    status: SessionStatus::Unknown("init".to_string()),
```
jjbayer: I realize now that because `"init"` is not a valid session status, we should just make `SessionSessionTags::status` etc. a `String`.

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -38,125 +196,163 @@ pub fn extract_session_metrics<T: SessionLike>(
...
-    tags.insert("release".to_owned(), attributes.release.clone());
-    if let Some(ref environment) = attributes.environment {
-        tags.insert("environment".to_owned(), environment.clone());
-    }
-    if let Some(client) = client {
-        tags.insert("sdk".to_owned(), client.to_owned());
-    }
+    let release = attributes.release.clone();
+    let environment = attributes.environment.clone();
+    let sdk = client.map(|s| s.to_owned());
```
jjbayer: nit: these instances are cloned here to be later be either cloned again, or moved (`session[crashed]`). I would make them references here, and clone them everywhere below, i.e. `let release = attributes.release;`

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
-    AggregatorConfig, DurationUnit, Metric, MetricNamespace, MetricUnit, MetricValue,
+    AggregatorConfig, CounterType, DistributionType, DurationUnit, Metric, MetricNamespace,
+    MetricResourceIdentifier, MetricType, MetricValue,
 };
 
 use crate::metrics_extraction::conditional_tagging::run_conditional_tagging;
 use crate::statsd::RelayCounters;
 use crate::utils::SamplingResult;
 
+/// Enumerates the most common transaction-names, with a catch-all for any other names.
```
jjbayer:
```suggestion
/// Enumerates the metrics extracted from transaction payloads.
```

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
+#[derive(Clone, Debug, PartialEq)]
+enum TransactionsMetric {
```
jjbayer:
```suggestion
enum TransactionMetric {
```

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
+    User {
+        value: String,
+        timestamp: UnixTimestamp,
+        tags: TransactionsUserTags,
```
jjbayer:
```suggestion
        tags: TransactionUserTags,
```

jjbayer: same for other tag enums.

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
+        value: DistributionType,
+        timestamp: UnixTimestamp,
+        tags: TransactionsDurationTags,
+    },
+    CountPerRootProject {
+        value: CounterType,
+        timestamp: UnixTimestamp,
+        tags: TransactionsCPRTags,
+    },
+    Breakdowns {
```
jjbayer:
```suggestion
    Breakdown {
```

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
+        name: String,
+        timestamp: UnixTimestamp,
+        value: DistributionType,
+        tags: TransactionsBreakdownTags,
+    },
+    Measurements {
```
jjbayer: This represents one measurement, not multiple.

```suggestion
    Measurement {
```

relay-server/src/metrics_extraction/sessions.rs
```diff
@@ -1,15 +1,173 @@
...
+        tags: SessionSessionTags,
+    },
+    User {
+        distinct_id: String,
+        timestamp: UnixTimestamp,
+        tags: SessionUserTags,
+    },
+    Error {
+        id: Uuid,
+        timestamp: UnixTimestamp,
```
jjbayer: Random idea: we could get rid of the `timestamp` redundancy by having an `.into_metric(timestamp: UnixTimestamp)` method instead of `.into()`.

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
+        kind: String,
```
jjbayer:
```suggestion
        name: String,
```

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
+    measurement_rating: Option<String>,
+    release: Option<String>,
+    dist: Option<String>,
+    environment: Option<String>,
+    transaction: Option<String>,
+    platform: Option<String>,
+    transaction_status: Option<String>,
+    transaction_op: Option<String>,
+    http_method: Option<String>,
+    other: BTreeMap<String, String>,
```
jjbayer: nit: Because these are all optional, we could make it an enum instead, and keep them in a vector:

```rust
tags: vec![
    TransactionMeasurementTag::Release("foo".to_string()),
    // ...
]
```
    
But I'm also fine with keeping it as-is.

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
+struct TransactionsBreakdownTags {
+    release: Option<String>,
+    dist: Option<String>,
+    environment: Option<String>,
+    transaction: Option<String>,
+    platform: Option<String>,
+    transaction_status: Option<String>,
+    transaction_op: Option<String>,
+    http_method: Option<String>,
+    other: BTreeMap<String, String>,
```
jjbayer: I did not check line-by-line, but is this type identical to `TransactionMeasurementTags`? If so, they can be one type.

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -160,23 +611,33 @@ fn get_transaction_name(
...
+}
+
```
jjbayer: nit: Let's put all these types into a separate module:

```
metrics_extraction/
    transactions/
        mod.rs
        types.rs
```

(same for sessions).

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -1,20 +1,473 @@
...
+impl TransactionsMeasurementTags {
+    fn new(measurement_rating: Option<String>, universal_tags: UniversalTags) -> Self {
```
jjbayer: If we're keeping `UniversalTags` (which I think makes sense because of `extract_universal_tags`, let's store it in a nested fashion too, e.g.

```rust
struct TransactionMeasurementTags {
    universal_tags: UniversalTags,
    measurement_rating: Option<String>,
}
```

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -160,23 +611,33 @@ fn get_transaction_name(
...
+struct UniversalTags {
+    release: Option<String>,
+    dist: Option<String>,
+    environment: Option<String>,
+    transaction: Option<String>,
+    platform: Option<String>,
+    transaction_status: Option<String>,
+    transaction_op: Option<String>,
+    http_method: Option<String>,
+    other: BTreeMap<String, String>,
```
jjbayer:
```suggestion
    custom_tags: BTreeMap<String, String>,
```

The only source for this `other` category are custom tags, right?

jjbayer (changes requested):

relay-server/src/metrics_extraction/sessions.rs
TBS1996: only for SessionSessionTags or for the other ones 'tags'-structs too?

relay-server/src/metrics_extraction/sessions.rs
TBS1996: i like it, I'll make a trait

relay-server/src/metrics_extraction/transactions.rs
TBS1996: or maybe a hashset of enums? if its not bad for performance when iterating

relay-server/src/metrics_extraction/transactions.rs
TBS1996: yeah, at least thats how i interpret this code from `extract_universal_tags`

```
let custom_tags = &config.extract_custom_tags;
    if !custom_tags.is_empty() {
        // XXX(slow): event tags are a flat array
        if let Some(event_tags) = event.tags.value() {
            for tag_entry in &**event_tags {
                if let Some(entry) = tag_entry.value() {
                    let (key, value) = entry.as_pair();
                    if let (Some(key), Some(value)) = (key.as_str(), value.as_str()) {
                        if custom_tags.contains(key) {
                            tags.other.insert(key.to_owned(), value.to_owned());
                        }
                    }
                }
            }
        }
    }
    ```

File: relay-server/src/metrics_extraction/transactions.rs (hunk 48 above)
Comment (jjbayer) [reply relay-server/src/metrics_extraction/transactions.rs]: Also fine, I'd use a `BTreeSet` for small sets. If the necessary traits for that cannot be easily derived, I would stick with a vector though.

File: relay-server/src/metrics_extraction/sessions.rs (hunk 39 above)
Comment (jjbayer) [reply relay-server/src/metrics_extraction/sessions.rs]: If the others only accept known `SessionStatus` tags, I think we can leave that type there.

File: relay-config/src/config.rs [hunk 53]
```diff
@@ -1720,7 +1720,7 @@ impl Config {
             .persistent_envelope_buffer
             .as_ref()
             .map(|b| b.max_connections)
-            .unwrap_or(buffer_max_connections())
```
TBS1996, line 1723: this ones gives a clippy warning, unsure how it ended up in main branch.

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -38,125 +179,138 @@ pub fn extract_session_metrics<T: SessionLike>(
...
 
-    let mut tags = BTreeMap::new();
-    tags.insert("release".to_owned(), attributes.release.clone());
-    if let Some(ref environment) = attributes.environment {
-        tags.insert("environment".to_owned(), environment.clone());
-    }
-    if let Some(client) = client {
-        tags.insert("sdk".to_owned(), client.to_owned());
-    }
+    let common_tags = CommonTags {
```
TBS1996: adding this type instead of the references reintroduces that early clone you mentioned, although having owned types means you don't need to clone it the last time it's used here.

relay-metrics/src/protocol.rs
```diff
@@ -521,10 +513,13 @@ impl Metric {
...
-            raw_namespace.parse().ok()?,
-            name,
-            unit,
+        let mut metric = Self::new(
+            MetricResourceIdentifier {
+                ty,
+                namespace: raw_namespace.parse().ok()?,
+                name,
+                unit,
+            },
```
jjbayer: Are the changes in this file actually necessary, or can they be reverted?

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -1,15 +1,156 @@
...
+use relay_metrics::{
+    CounterType, Metric, MetricNamespace, MetricResourceIdentifier, MetricType, MetricValue,
+};
+
+use crate::metrics_extraction::IntoMetric;
+
+/// Enumerates the metrics extracted from session payloads.
+#[derive(Clone, Debug, PartialEq)]
+pub enum SessionMetric {
+    /// The number of sessions collected in the given time frame.
```
jjbayer: There's no timestamp anymore.

```suggestion
    /// The number of sessions collected in a given time frame.
```

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -1,15 +1,156 @@
...
+    Session {
+        counter: CounterType,
+        tags: SessionSessionTags,
+    },
+    /// The number of unique session users for the given time frame.
```
jjbayer:
```suggestion
    /// The number of unique session users for a given time frame.
```

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -1,15 +1,156 @@
...
+    User {
+        distinct_id: String,
+        tags: SessionUserTags,
+    },
+    /// The number of sessions that errored in the given time frame.
```
jjbayer:
```suggestion
    /// The number of sessions that errored in a given time frame.
```

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1,20 +1,219 @@
...
 
 use crate::metrics_extraction::conditional_tagging::run_conditional_tagging;
+use crate::metrics_extraction::IntoMetric;
 use crate::statsd::RelayCounters;
 use crate::utils::SamplingResult;
 
+/// Enumerates the metrics extracted from transaction payloads.
+#[derive(Clone, Debug, PartialEq)]
+enum TransactionMetric {
+    User {
```
jjbayer:
```suggestion
    /// A set metric counting unique users.
    User {
```

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1,20 +1,219 @@
...
+        value: String,
+        tags: TransactionUserTags,
+    },
+    Duration {
```
jjbayer, line 19:
```suggestion
    /// A distribution metric for the transaction duration.
    ///
    /// Also used to count transactions, as any distribution metric features a counter.
    Duration {
```

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1,20 +1,219 @@
...
+        unit: DurationUnit,
+        value: DistributionType,
+        tags: TransactionDurationTags,
+    },
+    CountPerRootProject {
```
jjbayer, line 27:
```suggestion
    /// An internal counter metric used to compute dynamic sampling biases.
    ///
    /// See https://github.com/getsentry/sentry/blob/d3d9ed6cfa6e06aa402ab1d496dedbb22b3eabd7/src/sentry/dynamic_sampling/prioritise_projects.py#L40.
    CountPerRootProject {
```

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1,20 +1,219 @@
...
+        value: CounterType,
+        tags: TransactionCPRTags,
+    },
+    Breakdown {
```
jjbayer, line 32:
```suggestion
    A metric created from [`relay_general::Breakdowns`].
    Breakdown {
```

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1,20 +1,219 @@
...
+        name: String,
+        value: DistributionType,
+        tags: TransactionBreakdownTags,
+    },
+    Measurement {
```
jjbayer, line 38:
```suggestion
    /// A metric created from a [`relay_general::Measurement`].
    Measurement {
```

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1,20 +1,219 @@
...
+}
+
+impl From<TransactionUserTags> for BTreeMap<String, String> {
+    fn from(value: TransactionUserTags) -> Self {
+        UniversalTags::into_str_keys(value.universal_tags)
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+enum UniversalTags {
```
jjbayer: nit: In the `sessions` module, this is called `CommonTag`. Let's use the same name here (and use singular, because this represents a single tag key).

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -1,15 +1,156 @@
...
+            SessionMetric::Error { .. } => MetricType::Set,
+            SessionMetric::User { .. } => MetricType::Set,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SessionErrorTags {
+    common_tags: CommonTags,
+}
```
jjbayer: In this case I would take a shortcut and declare `tags: CommonTags` in the `Session` variant.

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -1,15 +1,156 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
```
jjbayer, line 31:
```suggestion
/// Tags that are set on the `session` counter metric.
#[derive(Clone, Debug, PartialEq, Eq)]
```

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -1,15 +1,156 @@
...
+pub struct SessionSessionTags {
+    status: String,
+    common_tags: CommonTags,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
```
jjbayer, line 38:
```suggestion
/// Tags that are set on the `user` set metric.
#[derive(Clone, Debug, PartialEq, Eq)]
```

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -1,15 +1,156 @@
...
+pub struct SessionUserTags {
+    status: Option<SessionStatus>,
+    abnormal_mechanism: Option<String>,
+    common_tags: CommonTags,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
```
jjbayer, line 46:
```suggestion
/// Tags that are set on all session metrics.
#[derive(Clone, Debug, PartialEq, Eq)]
```

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -332,63 +532,67 @@ fn extract_transaction_metrics_inner(
...
-            ));
+            metrics.push(
+                TransactionMetric::User {
+                    value,
+                    tags: TransactionUserTags {
+                        universal_tags: tags,
+                    },
+                }
+                .into_metric(timestamp),
+            );
```
jjbayer: Please put the actual extraction logic in `mod.rs`.

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1170,22 +1375,26 @@ mod tests {
...
-                    tags
-                }
-            )]
+            &[TransactionMetric::Measurement {
+                name: "lcp".to_string(),
+                value: 41.0,
+                unit: MetricUnit::Duration(DurationUnit::MilliSecond),
+                tags: measurement_tags
+            }
+            .into_metric(UnixTimestamp::from_secs(1619420402))]
```
jjbayer: Why construct the metric we test against via `TransactionMetric`, and not leave it as it was?

Alternatively, you can also use `assert_debug_snapshot!` instead of `assert_eq!` here.

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1,20 +1,219 @@
...
+    universal_tags: BTreeMap<UniversalTags, String>,
+}
+
+impl From<TransactionUserTags> for BTreeMap<String, String> {
+    fn from(value: TransactionUserTags) -> Self {
+        UniversalTags::into_str_keys(value.universal_tags)
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
jjbayer, line 143: doc comment please.

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -1,20 +1,219 @@
...
+        let mut map = UniversalTags::into_str_keys(value.universal_tags);
+        map.insert("decision".to_string(), value.decision);
+        map
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct TransactionDurationTags {
+    universal_tags: BTreeMap<UniversalTags, String>,
+}
```
jjbayer: nit: Define a single `UniversalTags` newtype used in both `user` and `duration` metrics. On that, you can implement `From<...> for BTreeMap` again.

`struct UniversalTags(BTreeMap<UniversalTag, String>)`

jjbayer (changes requested):

relay-server/src/metrics_extraction/transactions/types.rs
TBS1996: the `new_mri` function is changed so i couldnt leave it as it is, it seemed natural to change it to the new way, maybe theres something i dont understand here

TBS1996: update: reverted the new_mri function so this is changed back too

relay-metrics/src/protocol.rs
TBS1996: seems not needed anymore, was only needed when we had it as an enum, so nice catch

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -1170,10 +1153,17 @@ mod tests {
         )
         .unwrap();
         metrics.retain(|m| m.name.contains("lcp"));
+        let mut tags = extract_universal_tags(event.value().unwrap(), &config);
+        tags.0.insert(
+            CommonTag::Custom("satisfaction".to_owned()),
+            "frustrated".to_owned(),
+        );
+        tags.0.insert(CommonTag::Platform, "other".to_string());
```
jjbayer: It does not look like this code does anything:

```suggestion
```

jjbayer (approved):
Looks like some of my suggestions were not valid rust code.

See comment about dead code in test, apart from that, I think we can merge this!

After merging this, let's do a follow-up to improve doc comments, based on what gets rendered to https://getsentry.github.io/relay/relay/.

PR #1940, opened 2023-03-15
ref(quotas): Refactor outcomes for rate limiting

jjbayer:
Using the new `retain_items` function introduced in https://github.com/getsentry/relay/pull/1944, refactor rate limit enforcement to be more tightly coupled to outcome reporting.

This should make it easier to add rate limiting for new data types.

For events and attachments, I was not able to untangle the relationship between `CategoryLimit` and `retain_item`, so I'll leave that to a follow-up PR.

### DIFF
--- a/relay-server/src/utils/rate_limits.rs
+++ b/relay-server/src/utils/rate_limits.rs
@@ -275,12 +275,11 @@ impl Enforcement {
         self.event.is_active()
     }
 
-    #[cfg(test)] // TODO: remove function
-    fn get_outcomes(
-        self,
-        envelope: &Envelope,
-        scoping: &Scoping,
-    ) -> impl Iterator<Item = TrackOutcome> {
+    /// Invokes [`TrackOutcome`] on event-related enforcements reported by the [`EnvelopeLimiter`].
+    ///
+    /// Datatypes with 1:1 relationships between outcomes and enforcement are handled automatically
+    /// by retain_items.
+    fn track_event_outcomes(&self, envelope: &Envelope, scoping: &Scoping) {
         let timestamp = relay_common::instant_to_date_time(envelope.meta().start_time());
         let scoping = *scoping;
         let event_id = envelope.event_id();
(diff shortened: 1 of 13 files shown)


relay-server/src/utils/rate_limits.rs
```diff
@@ -1145,38 +1134,38 @@ mod tests {
...
+        //     .collect::<Vec<_>>();
+
+        // assert_eq!(
+        //     outcomes,
+        //     vec![
+        //         (DataCategory::TransactionIndexed, 1),
+        //         (DataCategory::Profile, 1),
+        //         (DataCategory::Transaction, 1)
+        //     ]
+        // );
```
jjbayer, line 1156: TODO: Restore these tests.

relay-server/src/utils/rate_limits.rs
```diff
@@ -273,12 +275,11 @@ impl Enforcement {
...
-    fn get_outcomes(
-        self,
-        envelope: &Envelope,
-        scoping: &Scoping,
-    ) -> impl Iterator<Item = TrackOutcome> {
+    /// Invokes [`TrackOutcome`] on event-related enforcements reported by the [`EnvelopeLimiter`].
+    ///
+    /// Datatypes with 1:1 relationships between outcomes and enforcement are handled automatically
+    /// by retain_items.
+    fn track_event_outcomes(&self, envelope: &Envelope, scoping: &Scoping) {
```
jjbayer, line 282: Eventually I would like to get rid of this function and emit all outcomes from `retain_item`.

relay-server/src/utils/rate_limits.rs
```diff
@@ -565,44 +553,54 @@ where
...
 
         // Remove profiles even if the transaction is not rate limited
-        if enforcement.profiles.is_active() && item.ty() == &ItemType::Profile {
-            return false;
+        let limit = &enforcement.profiles;
+        if limit.is_active() && item.ty() == &ItemType::Profile {
+            return RetainItem::Drop(
+                Outcome::RateLimited(limit.reason_code.clone()),
+                limit.category,
+                1,
```
jjbayer, line 583: Theoretically, the quantity set by the `limit` could defer from the number passed here. I found no good solution for this.

jjbayer:
Reopen when we have a solution for all rate limiting outcomes.

PR #1941, opened 2023-03-15
feat(normalization): Trim logger smartly

### LINKED ISSUE #1827: Increase character limit for the logger
A customer gets it's value from the logger but it's been capped at 64 which truncates the information.
Is there a reason for this or can we up the limit to a fair amount like for instance the TagValue => 200

`"_meta": {
    "logger": {
      "": {
        "rem": [
          [
            "!limit",
            "s",
            61,
            64
          ]
        ],
        "len": 86
      }
    }
  },`

can be found in this [file ](https://github.com/getsentry/relay/blob/6ebaf355ef20a2c10c5d1ced99f5a296b31c396c/relay-general/src/processor/attrs.rs#L117-L138)

iker-barriocanal:
This PR changes how we trim loggers for events coming from Java platforms. Chars from the end are now prioritized (so the string is trimmed from the beginning), and it's also attempted to trim full words in `.`s.

We are only interested in doing this for Java at this moment; because loggers in Java are fully qualified names (including the package, so potentially very long strings) where the class name at the end is more important, and because it's the only platform we've received this type of feedback on.

Resolves https://github.com/getsentry/relay/issues/1827.

### Examples

These examples are from the tests, see the diff for more examples and/or edge cases.

```txt
# Before
this_is-way_too_long-and_we_only_have_one_word-so_we_cant_smart_trim

# After
*is-way_too_long-and_we_only_have_one_word-so_we_cant_smart_trim
```

```txt
# Before
super_out.this_is_already_out_too.this_part-is-kept.this_right_here-is_a-very_long_word

# After
*.this_part-is-kept.this_right_here-is_a-very_long_word
```

### DIFF
--- a/relay-general/src/store/normalize.rs
+++ b/relay-general/src/store/normalize.rs
@@ -1042,22 +1042,35 @@ fn shorten_logger(logger: String) -> String {
     tokens.reverse(); // Prioritize chars from the end of the string.
     tokens.pop();
 
-    remove_logger_extra_chars(&mut tokens);
-    remove_logger_word(&mut tokens);
+    let word_cut = remove_logger_extra_chars(&mut tokens);
+    if word_cut {
+        remove_logger_word(&mut tokens);
+    }
 
     tokens.reverse();
     "*".to_owned() + &tokens.iter().join("")
 }
 
 /// Remove as many tokens as needed to match the maximum char limit defined in
-/// [`MaxChars::Logger`], and an extra token for the logger prefix.
-fn remove_logger_extra_chars(tokens: &mut Vec<&str>) {
+/// [`MaxChars::Logger`], and an extra token for the logger prefix. Returns
+/// whether a word has been cut.
+///
+/// A word is considered any non-empty substring that doesn't contain a `.`.
+fn remove_logger_extra_chars(tokens: &mut Vec<&str>) -> bool {
     // Leave one slot of space for the prefix
     let mut remove_chars = tokens.len() - MaxChars::Logger.limit() + 1;
+    let mut word_cut = false;
     while remove_chars > 0 {
-        tokens.pop();
+        if let Some(c) = tokens.pop() {
+            if !word_cut && c != "." {
+                word_cut = true;
+            } else if word_cut && c == "." {
+                word_cut = false;
+            }
+        }
         remove_chars -= 1;
     }
+    word_cut
 }
 
 /// If the `.` token is present, removes all tokens from the end of the vector
@@ -2283,6 +2296,22 @@ mod tests {
                 "event_id": "7637af36578e4e4592692e28a1d6e2ca",
                 "platform": "java",
                 "logger": "super_out.this_is_already_out_too.this_part-is-kept.this_right_here-is_a-very_long_word",
+            })
+            .into(),
+        );
+
+        let mut processor = NormalizeProcessor::default();
+        process_value(&mut event, &mut processor, ProcessingState::root()).unwrap();
+        assert_annotated_snapshot!(event);
+    }
+
+    #[test]
+    fn test_normalize_logger_word_leading_dots() {
+        let mut event = Event::from_value(
+            serde_json::json!({
+                "event_id": "7637af36578e4e4592692e28a1d6e2ca",
+                "platform": "java",
+                "logger": "io.this-tests-the-smart-trimming-and-word-removal-around-dot.words",
             })
             .into(),
         );
--- a/relay-general/src/store/normalize.rs
+++ b/relay-general/src/store/normalize.rs
@@ -804,6 +804,19 @@ impl<'a> Processor for NormalizeProcessor<'a> {
             event.client_sdk.set_value(self.get_sdk_info());
         }
 
+        if event.platform.as_str() == Some("java") {
+            if let Some(event_logger) = event.logger.value_mut() {
+                let trimmed_logger = event_logger.as_str().trim();
+                let logger_len = bytecount::num_chars(trimmed_logger.as_bytes());
+
+                if logger_len > MaxChars::Logger.limit() {
+                    *event_logger = shorten_logger(trimmed_logger);
+                } else if trimmed_logger != event_logger.as_str() {
+                    *event_logger = trimmed_logger.to_owned();
+                }
+            }
+        }
+
         // Normalize connected attributes and interfaces
         self.normalize_spans(event);
         self.normalize_trace_context(event);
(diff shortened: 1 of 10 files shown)


relay-general/src/store/normalize.rs
```diff
@@ -1001,6 +1002,78 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
         Ok(())
     }
+
+    fn process_event_logger(
+        &mut self,
+        event_logger: &mut EventLogger,
+        meta: &mut Meta,
+        state: &ProcessingState,
+    ) -> ProcessingResult {
+        event_logger.process_child_values(self, state)?;
```
jjbayer: Why even call `process_child_values`? `EventLogger` will always be a leaf node AFAIK.

relay-general/src/store/normalize.rs
```diff
@@ -1001,6 +1002,78 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
+            event_logger.set_logger(trimmed_logger.to_owned());
+        }
+
+        Ok(())
+    }
+}
+
+static LOGGER_DELIMITER_PATTERN: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"\.").expect("Invalid logger separator Regex"));
+static LOGGER_TRIMMED_PREFIX: &str = "*";
```
```suggestion
const LOGGER_TRIMMED_PREFIX: &str = "*";
```

relay-general/src/store/normalize.rs
```diff
@@ -1001,6 +1002,78 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
+
+fn shorten_logger(logger: &str) -> String {
+    let mut tokens = LOGGER_DELIMITER_PATTERN.split(logger).collect_vec();
```
jjbayer: Using regex here seems like overkill to me.
```suggestion
    let mut tokens = logger.split(LOGGER_DELIMITER_CHAR).collect_vec();
```

relay-general/src/store/normalize.rs
```diff
@@ -1001,6 +1002,78 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
+    tokens.reverse(); // Prioritize keeping chars from the end
+
+    let logger_len = bytecount::num_chars(logger.as_bytes());
+    let target_size = logger_target_size(tokens);
+    let target_idx = logger_len - target_size;
+
+    LOGGER_TRIMMED_PREFIX.to_owned() + &logger[target_idx..]
```
jjbayer: `logger[target_idx..]` could panic if `target_idx` is not a character boundary.

E.g. `&"Löwe"[2..]` will result in 

```
thread 'main' panicked at 'byte index 2 is not a char boundary; it is inside 'ö' (bytes 1..3) of `Löwe`'
```

relay-general/src/store/normalize.rs
```diff
@@ -1001,6 +1002,78 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
+    if MaxChars::Logger.limit() < logger_prefix_len {
+        return 0;
+    }
+
+    let max = MaxChars::Logger.limit() - logger_prefix_len;
+    let mut cummulative = 0;
+
+    for token in tokens {
+        // Consider one slot for the delimiter between tokens.
+        let current = bytecount::num_chars(token.as_bytes()) + 1;
```
jjbayer: It's a good idea to use `num_chars` for limiting the string, but the computed `target_size` that is used for slicing the string should be in bytes. Best add some unit tests with strings that contain a lot of special characters to see the effects.

jjbayer (commented):
I like the idea of a dedicated `EventLogger` type, _but_ if we want to make this normalization dependent on other fields like `platform`, I think it's easier to add a function to `light_normalize_event` instead, similar to `normalize_logentry`.

iker-barriocanal:
> but if we want to make this normalization dependent on other fields like `platform`, I think it's easier to add a function to `light_normalize_event` instead, similar to `normalize_logentry`

@jjbayer if we need to do this platform-dependent, I was thinking of doing it in the normalize processor instead. I don't think we need it as part of the light normalization.

relay-general/src/store/normalize.rs
```diff
@@ -806,6 +806,19 @@ impl<'a> Processor for NormalizeProcessor<'a> {
             event.client_sdk.set_value(self.get_sdk_info());
         }
 
+        if event.platform.as_str() == Some("java") {
+            if let Some(event_logger) = event.logger.value_mut() {
+                let trimmed_logger = event_logger.as_str().trim();
+                let logger_len = bytecount::num_chars(trimmed_logger.as_bytes());
+
+                if logger_len > MaxChars::Logger.limit() {
```
jjbayer: You do the same check in `shorten_logger`, so I would remove it here.

relay-general/src/store/normalize.rs
```diff
@@ -806,6 +806,19 @@ impl<'a> Processor for NormalizeProcessor<'a> {
             event.client_sdk.set_value(self.get_sdk_info());
         }
 
+        if event.platform.as_str() == Some("java") {
+            if let Some(event_logger) = event.logger.value_mut() {
+                let trimmed_logger = event_logger.as_str().trim();
```
jjbayer: Can we move this basic trimming inside `shorten_logger`?

relay-general/src/store/normalize.rs
```diff
@@ -1005,6 +1018,68 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
+/// String.
+///
+/// To shorten the logger, all extra chars that don't fit into the maximum limit
+/// are removed, from the beginning of the logger.  Then, if the remaining
+/// substring contains a `.` somewhere but in the end, all chars until `.`
+/// (exclusive) are removed.
+///
+/// Additionally, the new logger is prefixed with `*`, to indicate it was
+/// shortened.
+fn shorten_logger(logger: &str) -> String {
```
jjbayer: nit: I would rename this to `trim_logger`, since that is the word we use in other places (e.g. `TrimmingProcessor`).

relay-general/src/store/normalize.rs
```diff
@@ -1005,6 +1018,68 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
+    let logger_len = bytecount::num_chars(logger.as_bytes());
+    if logger_len <= MaxChars::Logger.limit() {
+        return logger.to_owned();
```
jjbayer: You could change the interface of this function to consume a `String`, then you could return `logger` here without extra allocation:

`fn shorten_logger(logger: String) -> String {`

relay-general/src/store/normalize.rs
```diff
@@ -1005,6 +1018,68 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
+    }
+
+    let mut tokens = logger.split("").collect_vec();
+    // Remove empty str tokens from the beginning and end.
+    tokens.pop();
+    tokens.reverse(); // Prioritize chars from the end of the string.
+    tokens.pop();
+
+    remove_logger_extra_chars(&mut tokens);
+    remove_logger_word(&mut tokens);
```
jjbayer: Shouldn't this be called only if `remove_logger_extra_chars` cut a word in half? If my logger is `foo.bar`, wouldn't `remove_logger_word` crop it to `.bar`?

relay-general/src/store/normalize.rs
```diff
@@ -1005,6 +1018,68 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
+
+    tokens.reverse();
+    "*".to_owned() + &tokens.iter().join("")
```
```suggestion
    tokens.join("*")
```

iker-barriocanal: `join` is the contrary operation to `split`. We only want the `*` at the beginning to indicate the logger was trimmed, not between each token (where a token is a character).

relay-general/src/store/normalize.rs
iker-barriocanal: It's a good suggestion. However, I've moved the trimming to this function too and `trim` returns an `str`, so we still have an extra allocation in this case.

relay-general/src/store/normalize.rs
iker-barriocanal: That's an interesting case I didn't think about. Should be addressed in db4d016.

relay-general/src/store/normalize.rs
iker-barriocanal: Unless there're strong opinions about changing that name, I'd rather not. The reason is that what this method does isn't really trimming:

> In computer programming, trimming or stripping is a string manipulation in which leading and trailing whitespace is removed from a string.

We don't trim here. We remove a non-whitespace substring from the string and insert a `*`. I'd like to avoid any future confusion coming from the wording.

relay-general/src/store/normalize.rs
jjbayer: Whoops, I guess I meant something like
```suggestion
    format!("*{}", tokens.join(""))
```

relay-general/src/store/normalize.rs
```diff
@@ -808,13 +808,9 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
         if event.platform.as_str() == Some("java") {
             if let Some(event_logger) = event.logger.value_mut() {
-                let trimmed_logger = event_logger.as_str().trim();
-                let logger_len = bytecount::num_chars(trimmed_logger.as_bytes());
-
-                if logger_len > MaxChars::Logger.limit() {
-                    *event_logger = shorten_logger(trimmed_logger);
-                } else if trimmed_logger != event_logger.as_str() {
-                    *event_logger = trimmed_logger.to_owned();
+                let shortened = shorten_logger(event_logger.clone());
```
jjbayer: nit: It must be possible to prevent this clone. Maybe change the signature to `shorten_logger(&mut String) -> ()`? Then the check below becomes unnecessary.

jjbayer (approved):

relay-general/src/store/normalize.rs
iker-barriocanal: If we change the signature to accept a `&mut String`, we remove a cloning here. However, in cases where the logger is smaller than the max size, we will always perform an allocation (if we clone the String, not always) which is not the case for cloning; plus, making the code less readable (why do we need to pass a mutable string?). So optimizing to remove extra allocations doesn't always happen.

I don't have data to back up the following, but my guess is that the performance improvement is not noticeable. This makes me optimize how the code is used instead.

Of course, I may be missing something that makes this possible _and_ readable. If that's the case, I'm happy to update the code. If not, I'll go ahead and merge it like this.

jjbayer: > I may be missing something that makes this possible and readable.

Opened a little draft PR, let me know what you think: https://github.com/getsentry/relay/pull/1950

jjbayer:
@iker-barriocanal One thing I missed during review: Should we add a `Meta` remark when we change the string? Something like

https://github.com/getsentry/relay/blob/1c47b22e2ddee545a9b76e09f8eca450e83b8965/relay-general/src/store/trimming.rs#L424-L429

jjbayer (approved):

PR #1944, opened 2023-03-16
ref(server): Track outcomes in retain_items

jjbayer:
Add a function to `ManagedEnvelope` which tightly couples removal of items with outcome reporting.

This iteration only covers the easy cases (mostly in processing), where there is a 1-to-1 relationship between dropped item and produced outcome. The rate limiting case will have to be handled later.

ref: https://github.com/getsentry/relay/issues/1888

#skip-changelog


relay-server/src/utils/managed_envelope.rs
```diff
@@ -43,23 +43,34 @@ impl Handling {
     }
 }
 
+/// Represents the decision on whether or not to keep an envelope item.
+pub enum RetainItem {
```
iker-barriocanal: Nice!

relay-server/src/utils/managed_envelope.rs
```diff
@@ -70,17 +81,24 @@ pub struct EnvelopeContext {
     done: bool,
 }
 
-impl EnvelopeContext {
-    /// Computes an envelope context from the given envelope.
-    fn new_internal(envelope: &Envelope, slot: Option<SemaphorePermit>) -> Self {
+impl ManagedEnvelope {
+    /// Computes an managed envelope from the given envelope.
```
iker-barriocanal:
```suggestion
    /// Computes a managed envelope from the given envelope.
```

relay-server/src/utils/managed_envelope.rs
```diff
@@ -91,26 +109,66 @@ impl EnvelopeContext {
...
+    pub fn update(&mut self) -> &mut Self {
+        self.event_id = self.envelope().event_id();
+        self.summary = EnvelopeSummary::compute(self.envelope());
         self
     }
 
+    /// Retains only the items specified by the predicate.
+    ///
+    /// In other words, remove all elements where `f(&item)` returns `false`. This method operates
+    /// in place and preserves the order of the retained items.
```
iker-barriocanal: Nice wording of docs, I see what you did here 😄

relay-server/src/utils/managed_envelope.rs
```diff
@@ -43,23 +43,34 @@ impl Handling {
...
 }
 
+/// Represents the decision on whether or not to keep an envelope item.
+pub enum RetainItem {
+    /// Keep the item.
+    Keep,
+    /// Drop the item and log an outcome for it.
+    Drop(Outcome, DataCategory, usize), // TODO: derive DataCategory and usize from Item
+    /// Drop the item without logging an outcome.
+    DropSilently,
```
iker-barriocanal: As a follow-up, would it make sense to document under what circumstances we should use each of these?

iker-barriocanal (commented):
Overall the PR LGTM.

I understand that another branch with some refactoring was merged to this branch. It may be my preference, but I find this approach harder to review. I prefer merging PRs to master even if they are not fully ready yet and merge follow-ups to master too. It keeps diffs smaller, removing noise from reviews and making it easier to provide better feedback or spot unwanted actions.

jjbayer:
> I prefer merging PRs to master even if they are not fully ready yet and merge follow-ups to master too.

@iker-barriocanal Sorry about that! The intention was as you said, but github still showed the large diff presumably because I had merge conflicts with the current master branch. Updated now so the diff should look more managable.

Note that I put this PR back in draft mode today, because I still want to change one thing.

relay-server/src/utils/managed_envelope.rs
iker-barriocanal: Not sure if it changed since my last review or if I missed it, but the predicate `retain_items` expects returns a `RetainItem` enum, and docstrings mention `false`. Could we update them?

relay-server/src/utils/managed_envelope.rs
```diff
@@ -129,6 +140,33 @@ impl ManagedEnvelope {
...
+                    outcomes.push((outcome, category, item.quantity()));
+                }
+
+                false
+            }
+            RetainItem::DropSilently => false,
+        });
+        for (outcome, category, quantity) in outcomes {
+            self.track_outcome(outcome, category, quantity);
+        }
```
iker-barriocanal, line 166: There's an `outcomes` vector that we populate during `retain_items`. Then, here, we track the outcomes of that vector. Is there any reason to not call `track_outcome` directly from inside `retain_items`?

relay-server/src/utils/managed_envelope.rs
```diff
@@ -175,6 +213,16 @@ impl ManagedEnvelope {
         }
     }
 
+    /// Returns `true` if the indexed data category should be used for reporting.
+    ///
+    /// If metrics have been extracted from the event item, we use the indexed category
+    /// (for example, [TransactionIndexed](`DataCategory::TransactionIndexed`)) for reporting
+    /// rate limits and outcomes, because reporting of the main category will be handled by
+    /// the metrics aggregator.
```
iker-barriocanal: nit

> because reporting of the main category will be handled by the metrics aggregator.

I think "_main_ data category" may be a bit confusing, I'd add an example as with `TransactionIndexed`:

```suggestion
    /// If metrics have been extracted from the event item, we use the indexed category
    /// (for example, [TransactionIndexed](`DataCategory::TransactionIndexed`)) for reporting
    /// rate limits and outcomes, because reporting of the main category (for example, [Transaction]
    /// (`DataCategory::Transaction`) for processed transactions) will be handled by the metrics
    /// aggregator.
```

iker-barriocanal (approved):
🚀

relay-server/src/utils/managed_envelope.rs
jjbayer: `track_outcome` borrows `self`, which is not allowed in this scope because `retain_items` borrows `self.envelope` as mutable. I also dislike the vector here though, open to any ideas how to get around the borrowing issue.

relay-server/src/utils/managed_envelope.rs
jjbayer: Good catch, rewrote the entire paragraph now.

relay-server/src/utils/managed_envelope.rs
iker-barriocanal: I see. I think it's fine, it was more like a nit. Let's trust the borrow checker over me 😄.

PR #1945, opened 2023-03-16
feat(general): Indicate if OS-version may be frozen

TBS1996:
Implements the following Sentry-PR: https://github.com/getsentry/sentry/pull/45369

Figured relay would be a better fit, a benefit to implementing it in relay is that the OS-version is only potentially frozen if it's from the user-agent, not from the client hints, which is easier to figure out here.

### DIFF
--- a/relay-general/src/protocol/contexts/os.rs
+++ b/relay-general/src/protocol/contexts/os.rs
@@ -253,6 +253,7 @@ OsContext {
 
         let os = OsContext::from_hints_or_ua(&RawUserAgentInfo::from_headers(&headers)).unwrap();
 
+        // Checks that the '>=' prefix is added.
         assert_eq!(os.version.value().unwrap(), ">=10.15.7");
     }
 
--- a/relay-general/src/protocol/contexts/os.rs
+++ b/relay-general/src/protocol/contexts/os.rs
@@ -73,14 +73,33 @@ impl FromUserAgentInfo for OsContext {
 
     fn parse_user_agent(user_agent: &str) -> Option<Self> {
         let os = parse_os(user_agent);
+        let mut version = get_version(&os.major, &os.minor, &os.patch);
 
         if !is_known(&os.family) {
             return None;
         }
 
+        let name = os.family.into_owned();
+
+        // Since user-agent strings freeze the OS-version at windows 10 and mac os 10.15.7,
+        // we will indicate that the version may in reality be higher.
+        if name == "Windows" {
+            if let Some(v) = version.as_mut() {
+                if v == "10" {
+                    v.insert_str(0, ">=");
+                }
+            }
+        } else if name == "Mac OS X" {
+            if let Some(v) = version.as_mut() {
+                if v == "10.15.7" {
+                    v.insert_str(0, ">=");
+                }
+            }
+        }
+
         Some(Self {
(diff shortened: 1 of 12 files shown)


relay-general/src/protocol/contexts/os.rs
```diff
@@ -73,14 +73,33 @@ impl FromUserAgentInfo for OsContext {
...
         }
 
+        let name = os.family.into_owned();
+
+        // Since user-agent strings freeze the OS-version at windows 10 and mac os 10.15.7,
+        // we will indicate that the version may in reality be higher.
+        if name == "Windows" {
+            if let Some(v) = version.as_mut() {
+                if v == "10" {
+                    v.insert_str(0, ">=");
```
TBS1996, line 89: I figured doing this instead of `*v = ">=10".to_owned()' would prevent a heap allocation (provided the vec doesn't reach capacity). Let me know if this is wrong.

relay-general/src/protocol/contexts/os.rs
```diff
@@ -191,7 +210,7 @@ OsContext {
             let headers = vec![
             Annotated::new((
                 Annotated::new("user-agent".to_string().into()),
-                Annotated::new(r#"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"#.to_string().into()),
+                Annotated::new(r#"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"#.to_string().into()),
```
TBS1996, line 213: Since I wanted to test this functionality in a separate test, I changed the version here so that it wouldn't trigger the '>=' prefix

AbhiPrasad:
Before we decide to take this option, what about the option of adding extra metadata that we can render in the UI?

So if we detect that an os-version may be frozen, we send additional info down with the event that the UI/product can use to indicate this (maybe even tell people about client hints).

TBS1996:
> Before we decide to take this option, what about the option of adding extra metadata that we can render in the UI?
> 
> So if we detect that an os-version may be frozen, we send additional info down with the event that the UI/product can use to indicate this (maybe even tell people about client hints).


like a tooltip info giving some more info sounds good to me, although I don't think our users can do much about it. If the end-user doesn't have client hints they would have to switch to a browser that supports it. 

would it be too primitive to simply do an equality check in the UI? if version == ">=10" then show tooltip?

AbhiPrasad:
> would it be too primitive to simply do an equality check in the UI? if version == ">=10" then show tooltip?

I think this makes a lot of sense to me. We should probably then create a GH issue to track this after this gets merged in!

jjbayer (commented):
Looks good to me! Please also add a unit test for the Windows case.

relay-general/src/protocol/contexts/os.rs
jjbayer: Sounds right to me.

jjbayer (approved):

AbhiPrasad (approved):

iker-barriocanal (approved):
🚀

PR #1946, opened 2023-03-17
ref(processor) Inject dependencies into envelope processor service

olksdr:
Remove from registry from the `EnvelopeProcessorService`.

ref: #1818 

#skip-changelog


relay-server/src/actors/processor.rs
```diff
@@ -2907,20 +2930,25 @@ mod tests {
     }
 
     fn create_test_processor(config: Config) -> EnvelopeProcessorService {
+        let (envelope_manager, _) = mock_service("envelope_manager", (), |&mut (), _| {});
+        let (outcome_aggregator, _) = mock_service("outcome_aggregator", (), |&mut (), _| {});
+        let (project_cache, _) = mock_service("project_cache", (), |&mut (), _| {});
+        let (upstream_relay, _) = mock_service("upstream_relay", (), |&mut (), _| {});
```
jjbayer, line 2936: nit: Should we make the third argument to `mock_service` optional, or create a wrapper `noop_service` or similar that takes only one argument?

relay-server/src/actors/processor.rs
```diff
@@ -2907,20 +2930,25 @@ mod tests {
...
             rate_limiter: None,
             #[cfg(feature = "processing")]
             geoip_lookup: None,
         }
     }
 
-    #[test]
-    #[ignore = "The REGISTRY panics if the Addr of an service (that is not yet started) is queried,
-                hence this test fails. The old SystemRegistry returned dummy Addr's hence this did
-                not fail."]
```
jjbayer, line 2922: Nice!

relay-server/src/actors/processor.rs
```diff
@@ -538,7 +540,14 @@ pub struct EnvelopeProcessorService {
...
 impl EnvelopeProcessorService {
     /// Creates a multi-threaded envelope processor.
-    pub fn new(config: Arc<Config>, _redis: Option<RedisPool>) -> anyhow::Result<Self> {
+    pub fn new(
+        config: Arc<Config>,
+        _redis: Option<RedisPool>,
+        envelope_manager: Addr<EnvelopeManager>,
+        outcome_aggregator: Addr<TrackOutcome>,
+        project_cache: Addr<ProjectCache>,
+        upstream_relay: Addr<UpstreamRelay>,
```
jjbayer, line 549: nit: We could think about putting these in a separate `Receivers` struct to reduce the number of arguments.

jjbayer (approved):

relay-server/src/actors/processor.rs
olksdr: yep, this is in the following PR: https://github.com/getsentry/relay/pull/1947

relay-server/src/actors/processor.rs
olksdr: yeah, I will look into this in the follow-ups

PR #1947, opened 2023-03-18
ref(utils): Remove from_registry from utils

olksdr:
Remove some usage of the global registry from the utils. 
This also includes a small refactoring of the `ProjectCache`, where the context was introduced to hold the services that `ProjectCache` and `Project`s need.  

#skip-changelog

ref: #1818


relay-server/src/actors/project_cache.rs
```diff
@@ -378,17 +378,46 @@ struct UpdateProjectState {
     no_cache: bool,
 }
 
+/// Holds the addresses of all services required for [`ProjectCache`].
+#[derive(Debug, Clone)]
+pub struct ProjectCacheContext {
```
jjbayer: nit(naming): I would rename this to `Services` because it is a better description of what's inside. Other mods can then call `project_cache::Services::new()`.

Or even better, call it `AddressBook` :)

jjbayer (approved):
Can we also get rid of `Registry::aggregator()` or is it too early for that?

relay-server/src/actors/project_cache.rs
olksdr: Even though I like the `AddressBook`, but still renamed it to `Services` for now 😄

olksdr:
> Can we also get rid of `Registry::aggregator()` or is it too early for that?

There is still one place, where this is used and it will be cleaned up in the next PR.

PR #1949, opened 2023-03-20
fix(build): Skip push to registry for dependabot

olksdr:
Dependabot does not have the write permissions to push images to container registry, so we want to skip the push for it.

related to https://github.com/getsentry/relay/pull/1948

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,7 +241,7 @@ jobs:
 
       - name: Push to ghcr.io
         # Do not run this on forks as they do not have access to secrets
-        if: "!github.event.pull_request.head.repo.fork"
+        if: "!github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'"
         run: |
           set -euxo pipefail
           docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -269,7 +269,7 @@ jobs:
 
     # Skip redundant checks for library releases
     # Skip if run on a fork
```
iker-barriocanal: I'd mention here that access to secrets is required, as in the line diff above.
```suggestion
    # Skip if run without access to secrets (e.g. a fork)
```

.github/workflows/ci.yml
```diff
@@ -448,7 +448,7 @@ jobs:
       - name: Run Sentry self-hosted e2e CI
         # Skip if a fork as they cannot upload to ghcr and this test attempts to pull
         # the image from ghcr
```
iker-barriocanal: Same suggestion as above.
```suggestion
        # Skip if they cannot upload to ghcr (e.g. a fork) and this test attempts to pull
        # the image from ghcr
```

jjbayer (approved):
I see a minor risk in skipping self-hosted e2e CI for dependabot PRs, but it seems acceptable.

iker-barriocanal (approved):
I was wondering if we should use `github.triggering_actor` instead of `github.actor`? Docs [here](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), and mini blog-post [here](https://github.blog/changelog/2022-07-19-differentiating-triggering-actor-from-executing-actor/). I'm not really sure of the differences between the two, but IIUC using `github.triggering_actor` will run these workflows if a Sentry member triggers the workflow? Approving since this is already existing behavior, your call to evaluate if we make this change.

iker-barriocanal:
@jjbayer I was also thinking about the risk of skipping self-hosted tests, but these are manually merged by a sentry member so the workflow runs on master. Ideally, all tests run on a PR, but at least we run them for the commit.

PR #1950, opened 2023-03-20
ref: no clone

jjbayer:


relay-general/src/store/normalize.rs
```diff
@@ -807,11 +807,9 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
 
         if event.platform.as_str() == Some("java") {
-            if let Some(event_logger) = event.logger.value_mut() {
-                let shortened = shorten_logger(event_logger.clone());
-                if shortened != event_logger.as_str() {
-                    *event_logger = shortened;
-                }
+            if let Some(event_logger) = event.logger.value_mut().take() {
+                let shortened = shorten_logger(event_logger);
+                event.logger.set_value(Some(shortened));
```
jjbayer, line 812: In case `shorten_logger` does not change the original string, it is never copied.

jjbayer:
One thing I missed during review: Should we add a `Meta` remark when we change the string? Something like

https://github.com/getsentry/relay/blob/1c47b22e2ddee545a9b76e09f8eca450e83b8965/relay-general/src/store/trimming.rs#L424-L429

jjbayer:
Merged into feature branch manually.

PR #1951, opened 2023-03-20
fix(pii): Scrub sensitive cookies

### LINKED ISSUE #1706: pii: Scrub `session` cookie by default
It seems we do not replace the value of the `session` cookie by default. To reproduce, add a `"session"` key to `sensitive_vars()` in the following test case:

https://github.com/getsentry/relay/blob/e0d15baeea247b2433f5def923301db86a2cd866/relay-general/src/pii/convert.rs#L381

The key `session` should probably not be treated as sensitive _outside_ of cookies, so we cannot just add it to the `@password:filter` rule. It could be a special case added to `@common` / `@common:filter` though.

jjbayer:
Always scrub cookies that are known to be session cookies.

Fixes https://github.com/getsentry/relay/issues/1706

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Normalize monitor slug parameters into slugs. ([#1913](https://github.com/getsentry/relay/pull/1913))
 - Smart trim loggers for Java platforms. ([#1941](https://github.com/getsentry/relay/pull/1941))
 - PII scrub `span.data` by default. ([#1953](https://github.com/getsentry/relay/pull/1953))
-- Scrub sensitive cookies. ([#1951](https://github.com/getsentry/relay/pull/1951))
+- Scrub sensitive cookies. ([#1951](https://github.com/getsentry/relay/pull/1951)))
 
 ## 23.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Normalize monitor slug parameters into slugs. ([#1913](https://github.com/getsentry/relay/pull/1913))
 - Smart trim loggers for Java platforms. ([#1941](https://github.com/getsentry/relay/pull/1941))
 - PII scrub `span.data` by default. ([#1953](https://github.com/getsentry/relay/pull/1953))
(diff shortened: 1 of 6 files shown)


iker-barriocanal (approved):

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
 - Indicate if OS-version may be frozen with '>=' prefix. ([#1945](https://github.com/getsentry/relay/pull/1945))
 - Normalize monitor slug parameters into slugs. ([#1913](https://github.com/getsentry/relay/pull/1913))
 - Smart trim loggers for Java platforms. ([#1941](https://github.com/getsentry/relay/pull/1941))
-
+- Scrub sensitive cookies. ([#1951](https://github.com/getsentry/relay/pull/1951)
```
jjbayer: Please also add this entry to `py/CHANGELOG.md`. See https://github.com/getsentry/relay#instructions-for-changelogs.

relay-general/src/pii/convert.rs
```diff
@@ -186,6 +221,34 @@ THd+9FBxiHLGXNKhG/FRSyREXEt+NyYIf/0cyByc9tNksat794ddUqnLOg0vwSkv
...
+        "csrf": "my csrf",
+        "xsrf": "my xsrf",
+        "_xsrf": "my _xsrf",
+        "_csrf": "my _csrf",
+        "csrf-token": "my csrf-token",
+        "csrf_token": "my csrf_token",
+        "xsrf-token": "my  xsrf-token",
+        "xsrf_token": "my  xsrf_token",
+        "fastcsrf": "my fastcsrf",
+        "_iris_csrf": "my _iris_csrf",
```
jjbayer, line 248: Let's add one non-sensitive entry here, to test that it is _not_ scrubbed, e.g.

```suggestion
        "_iris_csrf": "my _iris_csrf",
        "not_sensitive": "keep this",
```

jjbayer (commented):
I cannot formally approve, but this looks good to go to me!

CHANGELOG.md
mdtro: done! Thank you!

PR #1952, opened 2023-03-20
fix(codeowners): Correct behavior of ** in regex file matching

jangjodi:
[WOR-2572](https://getsentry.atlassian.net/browse/WOR-2572)

_#skip-changelog_

[WOR-2572]: https://getsentry.atlassian.net/browse/WOR-2572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


relay-common/src/glob.rs
```diff
@@ -77,7 +77,9 @@ fn translate_codeowners_pattern(pattern: &str) -> Option<Regex> {
     for (i, ch) in iterator {
         if let Some(skip_amount) = num_to_skip {
             num_to_skip = Some(skip_amount - 1);
-            continue;
+            if num_to_skip > Some(0) {
```
jangjodi, line 81: This prevents everything after the `*` in the pattern from being skipped

relay-common/src/glob.rs
```diff
@@ -87,10 +89,16 @@ fn translate_codeowners_pattern(pattern: &str) -> Option<Regex> {
                 let right_anchored = i + 2 == pattern.len();
                 let trailing_slash =
                     i + 2 < pattern.len() && pattern.chars().nth(i + 2) == Some('/');
+                let star_star_slash = pattern.chars().nth(i + 1) == Some('*')
+                    && pattern.chars().nth(i + 2) == Some('/');
```
jangjodi: This allows the `/` after `**/` to be optional. This is needed for these cases:

- **/something/file.py
- something/**/file.py

relay-common/src/glob.rs
lobsterkatie: Can you add this as a comment directly in the code?

relay-common/src/glob.rs
lobsterkatie: Same here.

relay-common/src/glob.rs
```diff
@@ -211,35 +221,91 @@ mod tests {
...
 
+        // Matches "apps/" anywhere it appears
         let pattern = "apps/";
         let regex = translate_codeowners_pattern(pattern).unwrap();
         assert!(regex.is_match(b"apps/file.txt"));
         assert!(regex.is_match(b"/apps/file.txt"));
         assert!(regex.is_match(b"/dir/apps/file.txt"));
         assert!(regex.is_match(b"/dir/subdir/apps/file.txt"));
 
+        // Matches leading "docs/<something>", where <x> is a single dir or filename
```
lobsterkatie:
```suggestion
        // Matches leading "docs/<x>", where <x> is a single dir or filename
```

lobsterkatie (approved):
LGTM

relay-common/src/glob.rs
```diff
@@ -89,12 +89,13 @@ fn translate_codeowners_pattern(pattern: &str) -> Option<Regex> {
                 let right_anchored = i + 2 == pattern.len();
                 let trailing_slash =
                     i + 2 < pattern.len() && pattern.chars().nth(i + 2) == Some('/');
-                let leading_star_slash = pattern.chars().nth(i + 2) == Some('/');
+                let star_star_slash = pattern.chars().nth(i + 1) == Some('*')
```
jjbayer: One thing I missed in my previous review: `pattern.chars().nth()` has a complexity of `O(n)`, because it has to go through the string and parse UTF-8 every time. So it would be better to `.collect()` the chars into a `Vec` on line 68 and iterate over that, so you can access neighboring chars in `O(1)`.

jangjodi: Thank you for your suggestion. I have changed this

relay-common/src/glob.rs
```diff
@@ -65,36 +65,38 @@ fn translate_codeowners_pattern(pattern: &str) -> Option<Regex> {
...
+            // Prevents everything after the * in the pattern from being skipped
             if num_to_skip > Some(0) {
                 continue;
             }
         }
+
         if ch == '*' {
             // Handle double star (**) case properly
-            if i + 1 < pattern.len() && pattern.chars().nth(i + 1) == Some('*') {
+            if i + 1 < pattern.len() && pattern_vec[i + 1] == '*' {
```
jjbayer:
```suggestion
            if pattern_vec.get(i + 1) == Some('*') {
```

See https://develop.sentry.dev/rust/#use-get-instead-of-slice-syntax.

relay-common/src/glob.rs
```diff
@@ -65,36 +65,38 @@ fn translate_codeowners_pattern(pattern: &str) -> Option<Regex> {
...
                 let left_anchored = i == 0;
-                let leading_slash = i > 0 && pattern.chars().nth(i - 1) == Some('/');
+                let leading_slash = i > 0 && pattern_vec[i - 1] == '/';
```
jjbayer: Here we need the `i > 0` check because `0 - 1` will panic in debug builds and overflow in release builds.

relay-common/src/glob.rs
```diff
@@ -65,36 +65,38 @@ fn translate_codeowners_pattern(pattern: &str) -> Option<Regex> {
...
                 let right_anchored = i + 2 == pattern.len();
-                let trailing_slash =
-                    i + 2 < pattern.len() && pattern.chars().nth(i + 2) == Some('/');
-                let star_star_slash = pattern.chars().nth(i + 1) == Some('*')
-                    && pattern.chars().nth(i + 2) == Some('/');
+                let trailing_slash = i + 2 < pattern.len() && pattern_vec[i + 2] == '/';
```
jjbayer:
```suggestion
                let trailing_slash = pattern_vec.get(i + 2) == Some('/');
```

relay-common/src/glob.rs
```diff
@@ -65,36 +65,38 @@ fn translate_codeowners_pattern(pattern: &str) -> Option<Regex> {
...
+                let star_star_slash =
+                    i + 2 < pattern.len() && pattern_vec[i + 1] == '*' && pattern_vec[i + 2] == '/';
```
jjbayer: `pattern_vec[i + 1] == '*'` is already given by the check in line 88, so `star_star_slash` looks equivalent to `trailing_slash` to me.

jjbayer (changes requested):

relay-common/src/glob.rs
jangjodi: Good catch, thank you. I have changed this and changed the slicing to `.get()`

jjbayer (approved):
Sorry for the delay.

Issue #1915, opened 2023-03-09
Align PII data scrubbing in `span.data` and `breadcrumb.data`


iker-barriocanal:
https://github.com/getsentry/relay/issues/1855 requests PII scrubbing specific fields in `span.data` and `breadcrumb.data`. Looking at `data` in the protocol, in [breadcrumbs](https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/) is undefined what the parameters are, but [spans](https://develop.sentry.dev/sdk/performance/span-data-conventions/) have it more defined and it's expected to follow [OpenTelemetry's semantics](https://github.com/open-telemetry/opentelemetry-specification/blob/24de67b3827a4e3ab2515cd8ab62d5bcf837c586/specification/trace/semantic_conventions/README.md).

In some case, such as in [here](https://github.com/getsentry/relay/pull/1725/files#diff-56c69297080cffba95fe1c78d629f7c1bc74ddaadeaa249b88580db771374b93R458), we're scrubbing breadcrumb `data` fields where it doesn't really make sense for an IP address to be, with the goal to remove all PII sensitive data that may have accidentally been leaked. However, this is not the case in spans.

This brings us to some questions about what the behavior should be, and what should relay do. Is there any reason not to PII scrub the whole `data` objects in spans and breadcrumbs? Is there any reason not to align the PII data scrubbing behavior of spans and breadcrumbs?

iker-barriocanal:
We decided to follow what this issue suggests -- to not type all the structured data in `span.data` and `breadcrumb.data`, and perform PII scrubbing in the whole objects. We'll monitor the effects and take action accordingly.

PR #1953, opened 2023-03-20
feat(pii-scrubbing): PII scrub span.data by default

iker-barriocanal:
This PR reverts https://github.com/getsentry/relay/pull/1889 and enables PII data scrubbing of `span.data` by default. For the context of the reason for the revert, see https://github.com/getsentry/relay/issues/1915#issuecomment-1465796416.

Related to: https://github.com/getsentry/relay/pull/1955.
Ref: https://github.com/getsentry/relay/issues/1855, https://github.com/getsentry/relay/issues/1915.


jjbayer (approved):
LGTM, @iker-barriocanal could you post in the client infra channel when this goes live? Just so everybody's aware that they will have to start explicitly typing [fields in `span.data`](https://develop.sentry.dev/sdk/performance/span-data-conventions/)  in Relay if the new behavior breaks something for them.

PR #1956, opened 2023-03-21
ref(envelope_manager): Remove from_registry calls from the service

olksdr:
This PR removes the `from_registry` calls from the `EnvelopeMangerService`.
Removed un-used code.

#skip-changelog

ref: #1818

### DIFF
--- a/relay-server/src/actors/envelopes.rs
+++ b/relay-server/src/actors/envelopes.rs
@@ -11,9 +11,7 @@ use relay_log::LogError;
 use relay_metrics::{Aggregator, Bucket, MergeBuckets};
 use relay_quotas::Scoping;
 use relay_statsd::metric;
-#[cfg(feature = "processing")]
-use relay_system::Addr;
-use relay_system::{FromMessage, NoResponse};
+use relay_system::{Addr, FromMessage, NoResponse};
 use tokio::sync::oneshot;
 
 use crate::actors::outcome::{DiscardReason, Outcome};
--- a/relay-server/src/actors/envelopes.rs
+++ b/relay-server/src/actors/envelopes.rs
@@ -8,7 +8,7 @@ use relay_common::ProjectKey;
 use relay_config::{Config, HttpEncoding};
 use relay_general::protocol::ClientReport;
 use relay_log::LogError;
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/envelopes.rs
```diff
@@ -201,15 +198,29 @@ impl FromMessage<SendMetrics> for EnvelopeManager {
 #[derive(Debug)]
 pub struct EnvelopeManagerService {
     config: Arc<Config>,
+    aggregator: Addr<Aggregator>,
+    enveloper_processor: Addr<EnvelopeProcessor>,
+    test_store: Addr<TestStore>,
+    upstream_relay: Addr<UpstreamRelay>,
```
jjbayer, line 204: nit: Should we make this an "address book" again?

relay-server/src/service.rs
```diff
@@ -120,20 +111,7 @@ impl ServiceState {
...
-            let rt = create_runtime("store-rt", 1);
-            let store = StoreService::create(config.clone())?.start_in(&rt);
-            envelope_manager.set_store_forwarder(store);
-            _store_runtime = Some(rt);
-        }
-
-        let envelope_manager = envelope_manager.start();
-        let test_store = TestStoreService::new(config.clone()).start();
-
+        let (envelope_manager, envelope_manager_rx) = channel(EnvelopeProcessorService::name());
```
jjbayer: I assume this is to resolve the circular dependency between manager and processor? Could you add a comment explaining this?

jjbayer (approved):

relay-server/src/actors/envelopes.rs
olksdr: This for me still fine and I would keep it separated for now and I would like to keep this still as it is.

relay-server/src/service.rs
olksdr: Yeah, we have to get an address of the manager before we can use it in the processor, still we also have to have a processor address in the manager here.

olksdr: done in [f0de37a](https://github.com/getsentry/relay/pull/1956/commits/f0de37a6ec95588d53b41fa020fd952b7b9562be)

PR #1960, opened 2023-03-22
feat(normalization): Mark scrubbed transactions as sanitized

iker-barriocanal:
All transactions that got their names scrubbed will be marked as sanitized, as long as the org feature flag is enabled.

6961c5e includes a small refactor around the usage of feature flags. Although it may seem like it, this doesn't change the behavior of relay around the application of transaction rules, since the initial flag belonging to `scrub_identifiers` controls the generation of rules in `sentry`, [see](https://github.com/getsentry/sentry/blob/242fd1dd0654e85bf027df59131db516a0b55395/src/sentry/relay/config/__init__.py#L203).

Ref: https://github.com/getsentry/team-ingest/issues/93

#skip-changelog

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -354,13 +354,13 @@ impl Processor for TransactionsProcessor<'_> {
             &TransactionSource::Url | &TransactionSource::Sanitized
         ) && self.name_config.scrub_identifiers
         {
-            let changed = scrub_identifiers(&mut event.transaction)?;
-            if changed && self.name_config.mark_scrubbed_as_sanitized {
-                let source = &mut event
+            scrub_identifiers(&mut event.transaction)?;
+            if self.name_config.mark_scrubbed_as_sanitized {
+                event
                     .transaction_info
                     .get_or_insert_with(Default::default)
-                    .source;
-                source.set_value(Some(TransactionSource::Sanitized));
+                    .source
+                    .set_value(Some(TransactionSource::Sanitized));
             }
         }
 
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -341,26 +341,27 @@ impl Processor for TransactionsProcessor<'_> {
                 .set_value(Some("<unlabeled transaction>".to_owned()))
         }
 
-        // Apply the rule if any found
-        self.apply_transaction_rename_rule(
-            &mut event.transaction,
-            event.transaction_info.value_mut(),
-        )?;
+        if self.name_config.scrub_identifiers {
+            // Apply the rule if any found
+            self.apply_transaction_rename_rule(
+                &mut event.transaction,
+                event.transaction_info.value_mut(),
+            )?;
 
-        // Normalize transaction names for URLs and Sanitized transaction sources.
-        // This in addition to renaming rules can catch some high cardinality parts.
-        if matches!(
-            event.get_transaction_source(),
-            &TransactionSource::Url | &TransactionSource::Sanitized
-        ) && self.name_config.scrub_identifiers
-        {
-            scrub_identifiers(&mut event.transaction)?;
-            if self.name_config.mark_scrubbed_as_sanitized {
-                event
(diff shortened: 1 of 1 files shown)


relay-general/src/store/transactions/processor.rs
```diff
@@ -354,13 +354,13 @@ impl Processor for TransactionsProcessor<'_> {
...
-                let source = &mut event
+            scrub_identifiers(&mut event.transaction)?;
+            if self.name_config.mark_scrubbed_as_sanitized {
+                event
                     .transaction_info
                     .get_or_insert_with(Default::default)
-                    .source;
-                source.set_value(Some(TransactionSource::Sanitized));
+                    .source
+                    .set_value(Some(TransactionSource::Sanitized));
```
jjbayer: The code structure is starting to look misleading (not your fault). I think it should look something like this:

```rust
// This comes from a feature flag that already toggles both rules and patterns,
// so we might as well put it around both function calls.
if self.name_config.scrub_identifiers {
    self.apply_transaction_rename_rule(
        &mut event.transaction,
        event.transaction_info.value_mut(),
    )?;
    if matches!(
            event.get_transaction_source(),
            &TransactionSource::Url | &TransactionSource::Sanitized) {
            
        scrub_identifiers(&mut event.transaction)?;
            
        if self.name_config.mark_scrubbed_as_sanitized {
            // set as sanitized
        }
    }
}
```

jjbayer (commented):
@iker-barriocanal did you check if affected orgs currently have any high-cardinality transactions that would destroy the UX with this change?

iker-barriocanal:
@jjbayer the list of orgs affected by the feature flag is [here](https://github.com/getsentry/getsentry/blob/7940f4560bb770b02dcfa7e66739c82fc3da7fa0/getsentry/features.py#L287). I think it's fine, but we can also consider limiting the flag to Sentry orgs for now.

relay-general/src/store/transactions/processor.rs
iker-barriocanal: That snippet changes the behavior. You're including `apply_transaction_rename_rule` inside the `scrub_identifiers ` feature flag now, and I don't think we want to do that.

In my opinion, considering the two feature flags the code is okayish as it is. Once we remove the feature flags, we can consider moving things around.

jjbayer: > That snippet changes the behavior.

They are equivalent in behavior because `name_config.scrub_identifiers` is tied to the same feature flag that controls the emission of clusterer rules in Sentry:

https://github.com/getsentry/relay/blob/5560fb555a6e4167023083a247fffff4b429891a/relay-server/src/actors/processor.rs#L2216-L2218

So if the flag is "on" you get both, if the flag is "off" you get none.

But with your version of the code, the intend seems to be to _only_ mark txns as sanitized when they passed through pattern-based normalization.

Either way, IMO this code should be restructured, but it does not have to be in this PR. https://github.com/getsentry/team-ingest/issues/14 will also require a change here (see [comment](https://github.com/getsentry/team-ingest/issues/14#issuecomment-1415776831)), so we can defer until then.

jjbayer (approved):
> I think it's fine, but we can also consider limiting the flag to Sentry orgs for now.

@iker-barriocanal My question was specifically whether or not the affected org(s) still had a lot of URL transactions with identifiers that we failed to parameterize. In the mean time I found the answer myself: the affected customer project only has a handful of high-volume but low-cardinality transactions labeled as `url`, so we can go ahead with the change.

iker-barriocanal:
@jjbayer I checked the projects for that org and `sentry` org and I believe it's fine. I haven't checked other sentry orgs, but for the purpose of this dogfooding I expect no issues. I should have written a more elaborated answer from the beginning.

relay-general/src/store/transactions/processor.rs
iker-barriocanal: Ohh, I see. Good suggestion, thanks! Addressed in 6961c5e. I also had to add the flag in tests, so that the scrubbing would run: 5d847ab.

PR #1962, opened 2023-03-23
feat(monitors): Validate monitor_config for upserts

evanpurkhiser:
Adds an optional `monitor_config` key for checkins

### DIFF
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -60,6 +60,46 @@ where
     uuid.as_simple().serialize(serializer)
 }
 
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+#[serde(tag = "type")]
+enum Schedule {
+    Crontab { value: String },
+    Interval { value: u64, unit: IntervalName },
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum IntervalName {
+    Year,
+    Month,
+    Week,
+    Day,
+    Hour,
+    Minute,
+}
+
+/// The monitor configuration playload for upserting monitors during check-in
+#[derive(Debug, Deserialize, Serialize)]
+struct MonitorConfig {
+    /// The monitor schedule configuration
+    schedule: Schedule,
+
+    /// How long (in minutes) after the expected checkin time will we wait until we consider the
+    /// checkin to have been missed.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    checkin_margin: Option<u64>,
+
+    /// How long (in minutes) is the checkin allowed to run for in in_rogress before it is
+    /// considered failed.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    max_runtime: Option<u64>,
+
+    /// tz database style timezone string
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    timezone: Option<String>,
+}
+
 /// The monitor check-in payload.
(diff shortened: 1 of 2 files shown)


relay-monitors/src/lib.rs
```diff
@@ -60,6 +60,48 @@ where
     uuid.as_simple().serialize(serializer)
 }
 
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+#[serde(tag = "schedule_type", content = "schedule")]
```
evanpurkhiser: Let me know if this is the way to go here, combined with the `flatten` in the field this seems to work pretty well.

relay-monitors/src/lib.rs
```diff
@@ -60,6 +60,48 @@ where
...
+struct MonitorConfig {
+    /// The monitor schedule configuration
+    #[serde(flatten)]
+    // Note that the `schedule_type` key is flattened from the Schedule enum marked by serde
+    schedule: Schedule,
+
+    /// How long (in minutes) after the expected checkin time will we wait until we consider the
+    /// checkin to have been missed.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    checkin_margin: Option<i8>,
```
jan-auer: This means it will fail to deserialize if you set a value that is larger than 127, and it allows you to set negative values. Is that what you want? Otherwise, I'd propose to use `u64` or `u32`.

relay-monitors/src/lib.rs
```diff
@@ -76,6 +118,10 @@ struct CheckIn {
     /// Duration of this check since it has started in seconds.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     duration: Option<f64>,
+
+    /// monitor configuration to support upserts.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    monitor_config: Option<MonitorConfig>,
```
jan-auer, line 122: _nit_: The `config` suffix seems redundant in protocol, but no formal objection on this.

relay-monitors/src/lib.rs
```diff
@@ -60,6 +60,48 @@ where
     uuid.as_simple().serialize(serializer)
 }
 
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+#[serde(tag = "schedule_type", content = "schedule")]
+enum Schedule {
+    Crontab(String),
+    Interval(i8, IntervalName),
```
jan-auer: This variant is not extensible. If you're relatively certain that you will only ever need this tuple for interval schedules, then feel free to keep it. Otherwise, I'd propose to make this a `struct`, so you can add new fields.

jan-auer (approved):
The PR is good to go. Please see comments below, those are up to your discretion to address.

relay-monitors/src/lib.rs
evanpurkhiser: This is a little tricky, since the monitor object actually has a `config` as it's own property. 

https://github.com/getsentry/sentry/blob/ad9fa09acb39820a12908d57e0b04a29c71b3a45/src/sentry/monitors/models.py#L175

There are other fields on a monitor such as it's name and project that are NOT part of the config. I wanted to keep it this way for consistency.

That said, from an API ingest pov we could simply call this `monitor` and it would be OK, let me bring it up and see what everyone think.s

relay-monitors/src/lib.rs
evanpurkhiser: Yeah I want `u64` you're right.

relay-monitors/src/lib.rs
evanpurkhiser: I've updated the configuration schema to not use a tuple pair like this.

PR #1964, opened 2023-03-23
ref(tx-processor): Apply transaction rules after UUID scrubbing

iker-barriocanal:
Applying transaction rules before UUID scrubbing makes it impossible to track what rules were applied. It's interesting to know what rules are useful to reduce the high cardinality, so that we can use them again over other rules that are less interesting or provide no value. Applying these rules after UUID scrubbing leaves no room for further modification of the transaction name, so the resulting transaction name is guaranteed to be the direct output of a rule.

This is the first step in the process of discarding the rules that provide no value and prioritizing those that do.

Ref: https://github.com/getsentry/team-ingest/issues/14.

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -362,10 +362,8 @@ impl Processor for TransactionsProcessor<'_> {
                 event.transaction_info.value_mut(),
             )?;
 
-            if !matches!(
-                event.get_transaction_source(),
-                &TransactionSource::Sanitized
-            ) && self.name_config.mark_scrubbed_as_sanitized
+            if matches!(event.get_transaction_source(), &TransactionSource::Url)
+                && self.name_config.mark_scrubbed_as_sanitized
             {
                 event
                     .transaction_info
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -448,7 +448,7 @@ mod tests {
     use super::*;
     use crate::processor::process_value;
     use crate::protocol::{Contexts, SpanId, TraceContext, TraceId, TransactionSource};
-    use crate::store::LazyGlob;
+    use crate::store::{LazyGlob, RedactionRule, RuleScope};
     use crate::testutils::assert_annotated_snapshot;
     use crate::types::Object;
 
@@ -2129,4 +2129,50 @@ mod tests {
         "open-12345-close",
         "open-12345-close"
     );
+
+    #[test]
+    fn test_scrub_identifiers_before_rules() {
+        // There's a rule matching the transaction name. However, the UUID
+        // should be scrubbed first. Scrubbing the UUID makes the rule to not
+        // match the transformed transaction name anymore.
+
+        let mut event = Annotated::<Event>::from_json(
+            r#"{
+                "type": "transaction",
+                "transaction": "/remains/rule-target/1234567890",
+                "transaction_info": {
+                    "source": "url"
+                },
+                "timestamp": "2021-04-26T08:00:00+0100",
+                "start_timestamp": "2021-04-26T07:59:01+0100",
+                "contexts": {
+                    "trace": {
+                        "trace_id": "4c79f60c11214eb38604f4ae0781bfb2",
+                        "span_id": "fa90fdead5f74053"
+                    }
+                }
+            }"#,
+        )
+        .unwrap();
+
+        process_value(
+            &mut event,
+            &mut TransactionsProcessor::new(TransactionNameConfig {
+                scrub_identifiers: true,
+                rules: &[TransactionNameRule {
+                    pattern: LazyGlob::new("/remains/*/1234567890/".to_owned()),
+                    expiry: Utc.with_ymd_and_hms(3000, 1, 1, 1, 1, 1).unwrap(),
+                    scope: RuleScope::default(),
+                    redaction: RedactionRule::default(),
+                }],
+                ..Default::default()
+            }),
+            ProcessingState::root(),
+        )
+        .unwrap();
+
+        // Annotate the snapshot instead of comparing transaction names, to also
+        // make sure the event's _meta is correct.
+        assert_annotated_snapshot!(event);
+    }
 }
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -1925,7 +1925,7 @@ mod tests {
         let json = r#"
         {
             "type": "transaction",
-            "transaction": "/foo/2fd4e1c67a2d28fced849ee1bb76e7391b93eb12/user",
+            "transaction": "/foo/rule-target/user",
             "transaction_info": {
               "source": "url"
             },
@@ -1999,7 +1999,7 @@ mod tests {
                      "s"
                    ]
                  ],
-                 "val": "/foo/2fd4e1c67a2d28fced849ee1bb76e7391b93eb12/user"
+                 "val": "/foo/rule-target/user"
                }
              }
            }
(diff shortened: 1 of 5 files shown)


relay-general/src/store/transactions/processor.rs
```diff
@@ -341,28 +341,32 @@ impl Processor for TransactionsProcessor<'_> {
...
-        ) && self.name_config.scrub_identifiers
-        {
-            let changed = scrub_identifiers(&mut event.transaction)?;
-            if changed && self.name_config.mark_scrubbed_as_sanitized {
-                let source = &mut event
-                    .transaction_info
-                    .get_or_insert_with(Default::default)
-                    .source;
-                source.set_value(Some(TransactionSource::Sanitized));
+        if self.name_config.scrub_identifiers {
```
jjbayer, line 350: nit: Should we rename the `scrub_identifiers` flag in config to `normalize`?

I've been thinking about naming for a while, I think we need unambiguous names to refer to the different parts of transaction name handling but haven't been able to come up with good ones.

relay-general/src/store/transactions/processor.rs
```diff
@@ -341,28 +341,32 @@ impl Processor for TransactionsProcessor<'_> {
...
+                    // TODO(iker): we also mark transaction as sanitized after
+                    // applying renaming rules. Once we remove this feature
+                    // flag, relay should only mark transactions as sanitized
+                    // once.
+                    event
+                        .transaction_info
+                        .get_or_insert_with(Default::default)
+                        .source
+                        .set_value(Some(TransactionSource::Sanitized));
+                }
```
jjbayer: It might not matter in practice, but I would put this block _after_ `apply_transaction_rename_rule`. Basically

1. Apply sanitation step 1
2. Apply sanitation step 2
3. Mark as sanitized

Right now the order is

1. Apply sanitation step 1
1. Mark as sanitized
1. Apply sanitation step 2

which feels unnatural to me.

relay-general/src/store/transactions/processor.rs
iker-barriocanal: I agree the names could be better, but I don't have any suggestions (besides being very explicit and setting extremely long names, which doesn't really improve things I guess). Since I expect to remove these feature flags soonish, I didn't rename them. That said, happy to refactor the names if you come up with a good suggestion.

relay-general/src/store/transactions/processor.rs
iker-barriocanal: Docs of `mark_scrubbed_as_sanitized` are as follows:

> True if transaction names scrubbed by regex patterns should be marked as `TransactionSource::Sanitized`.
>
> Transaction names modified by clusterer rules are always marked as such.

So the code follows these steps:

1. Apply sanitation step 1.
2. Mark as sanitized, resulting from step 1.
3. Apply sanitation step 2 (which also marks as sanitized, so no `4` needed).

I think 1 and 2 belong together. I agree that this could be improved, and that's why I added the `TODO`, but I wanted to minimize refactoring as much as possible until we have more information on how this looks and what we do with the feature flags..

jjbayer: > True if transaction names scrubbed by regex patterns should be marked as TransactionSource::Sanitized.

That doc was written before you removed the `if changed` condition though (by which it actually belonged to pattern-based matching). By removing that condition, the real consequence of `mark_scrubbed_as_sanitized` is that _every_ `url` transaction will be marked as sanitized as long as the `normalize-transaction-names` feature flag is on.

relay-general/src/store/transactions/processor.rs
```diff
@@ -57,7 +57,13 @@ impl<'r> TransactionsProcessor<'r> {
...
                 if let Some((rule, result)) = result {
                     if *transaction != result {
-                        meta.set_original_value(Some(transaction.clone()));
+                        // If another rule was applied before, we don't want to
+                        // rename the transaction name to keep the original one.
+                        // We do want to continue adding remarks though, in
+                        // order to keep track of all rules applied.
+                        if meta.original_value().is_none() {
+                            meta.set_original_value(Some(transaction.clone()));
+                        }
```
iker-barriocanal, line 66: I tried to do this with `get_or_insert_with`, but I couldn't figure out how. Happy to change it.

relay-general/src/store/transactions/processor.rs
```diff
@@ -1675,7 +1684,7 @@ mod tests {
         let json = r#"
         {
             "type": "transaction",
-            "transaction": "/foo/2fd4e1c67a2d28fced849ee1bb76e7391b93eb12/user/123/0/",
+            "transaction": "/foo/rule-target/user/123/0/",
```
iker-barriocanal, line 1686: The reason for these changes is to avoid UUID scrubbing rules running first in all fields. They are already running for `123` so that is good enough.

relay-general/src/store/transactions/processor.rs
```diff
@@ -356,13 +356,22 @@ impl Processor for TransactionsProcessor<'_> {
...
+                    // once.
                     event
                         .transaction_info
                         .get_or_insert_with(Default::default)
                         .source
                         .set_value(Some(TransactionSource::Sanitized));
                 }
             }
+
+            self.apply_transaction_rename_rule(
```
jjbayer, line 360: When `mark_scrubbed_as_sanitized` is enabled (which it will be soon), the rules matching on `source: url` will fail to match, so `apply_transaction_rename_rule` will essentially become dead code. Maybe the previously failing tests already reflected this, I did not check in detail.

This brings me back to the point that we should "mark as sanitized" after applying the rules.

jjbayer (changes requested):
See https://github.com/getsentry/relay/pull/1964#discussion_r1148924220.

relay-general/src/store/transactions/processor.rs
iker-barriocanal: This is a good catch! It should now be addressed, including a test to avoid this from happening in the future.

relay-general/src/store/transactions/processor.rs
jjbayer: I think this is nice and explicit.

relay-general/src/store/transactions/processor.rs
```diff
@@ -342,26 +348,30 @@ impl Processor for TransactionsProcessor<'_> {
...
+            }
+
+            self.apply_transaction_rename_rule(
+                &mut event.transaction,
+                event.transaction_info.value_mut(),
+            )?;
+
+            if !matches!(
+                event.get_transaction_source(),
+                &TransactionSource::Sanitized
```
jjbayer: Wouldn't this set `source=sanitized` for every source (`route`, `custom`, etc.). I think we need to match on `TransactionSource::Url` here explicitly.

jjbayer (changes requested):

relay-general/src/store/transactions/processor.rs
iker-barriocanal: Heh, I think I got confused _again_ from the code structure. Let's see when we can remove the feature flags here and make the code a bit easier to follow.

Thanks for putting in the effort!

relay-general/src/store/transactions/processor.rs
```diff
@@ -362,10 +362,8 @@ impl Processor for TransactionsProcessor<'_> {
                 event.transaction_info.value_mut(),
             )?;
 
-            if !matches!(
-                event.get_transaction_source(),
-                &TransactionSource::Sanitized
-            ) && self.name_config.mark_scrubbed_as_sanitized
+            if matches!(event.get_transaction_source(), &TransactionSource::Url)
+                && self.name_config.mark_scrubbed_as_sanitized
```
jjbayer, line 366: As discussed offline, we might want to add an an additional condition which is `scrub_identifiers_changed_something OR rename_rules.len() > 0`, to prevent projects that do not have any rules yet from marking everything as sanitized. But that is actually out of scope for this PR.

jjbayer (approved):

PR #1965, opened 2023-03-23
ref: Use sparse registry in dockerfile

untitaker:
from-scratch builds become significantly faster that way

#skip-changelog

### DIFF
--- a/Dockerfile.builder
+++ b/Dockerfile.builder
@@ -24,7 +24,7 @@ ARG GID=10000
 
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
     | sh -s -- -y --profile minimal --default-toolchain=${RUST_TOOLCHAIN_VERSION} \
-    && echo -e '[registries.crates-io]\nprotocol = "sparse"\n[net]\ngit-fetch-with-cli = true' > $CARGO_HOME/config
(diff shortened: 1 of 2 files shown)


jjbayer (approved):
SGTM

olksdr (commented):
for all our image builds we use `Dockerfile.builder` and the `Dockerfile` is deprecated and will be removed soon. 

Could you, please, also add the change to the `Dockerfile.builder` file?

Dockerfile.builder
```diff
@@ -24,7 +24,7 @@ ARG GID=10000
 
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
     | sh -s -- -y --profile minimal --default-toolchain=${RUST_TOOLCHAIN_VERSION} \
-    && echo -e "[net]\ngit-fetch-with-cli = true" > $CARGO_HOME/config \
+    && echo -e '[registries.crates-io]\nprotocol = "sparse"\n[net]\ngit-fetch-with-cli = true' > $CARGO_HOME/config
```
olksdr:
```suggestion
    && echo -e '[registries.crates-io]\nprotocol = "sparse"\n[net]\ngit-fetch-with-cli = true' > $CARGO_HOME/config \
```

olksdr (approved):

PR #1967, opened 2023-03-27
fix(quota): New reason code for fallback rate limit

### LINKED ISSUE #1929: External Relay reports rate limit without reason code
A user with rate limiting problems sent us an extract of their Relay log, and it seems there is no `reason` being tracked:

```
2023-03-01T08:41:52Z [relay_server::actors::outcome_aggregator] TRACE: Outcome aggregation requested:
TrackOutcome { timestamp: 2023-03-01T08:41:52.837099513Z, scoping: Scoping { organization_id: ...,
project_id: ProjectId(...), project_key: ProjectKey("..."), key_id:
Some(...) }, outcome: RateLimited(None), event_id: Some(EventId(...),
remote_addr: Some(127.0.0.1), category: Error, quantity: 1 }
```

This also shows in the stats_v2 API endpoint:

```json
{
  "by": {
    "outcome": "rate_limited",
    "reason": "none"
  },
  "totals": {
    "sum(quantity)": 552830
  }
}
```

Is reason code "none" a valid one? If not, figure out how it was created.

* In Sentry, we never seem to generate a quota object without a reason code (double check).
* The only place in the Relay code base where we construct a `RateLimited` variant with reason `None` is when the envelope is empty (see below). Is this an expected code path? Why would we emit outcomes for an empty envelope for which the contents were already rate limited?

https://github.com/getsentry/relay/blob/ee8d51d4426ccc2b6e931f41c88add75e53c1ca5/relay-server/src/actors/project.rs#L748-L750

Config of the external Relay:

```yaml
relay:
  mode: managed
  upstream: "https://sentry.io/"
  host: 0.0.0.0
  port: 8080
logging:
  level: trace
  format: auto
sentry:
  enabled: true
  dsn: "..."
http:
  timeout: 65
  connection_timeout: 65
limits:
  max_concurrent_requests: 10000
  max_connection_rate: 10000
  max_session_count: 10000
```

Internal issue: https://getsentry.atlassian.net/browse/FEEDBACK-1775.

jjbayer:
When a non-processing Relay receives a 429 response when sending envelopes to the upstream, it logs a `RateLimited` outcome with a reason code that is parsed from the `X-Sentry-Rate-Limits` header. However, if that header is not available, it logs an outcome with `RateLimited(None)`. This happens for example when Sentry SaaS rejects the envelope on the nginx layer.

Because reason code `None` is hard to debug, replace it with a dedicated reason code that we can search for.

Follow-up work: Modify the nginx anti-abuse layer to send an `X-Sentry-Rate-Limits` header.

Fixes https://github.com/getsentry/relay/issues/1929.

#skip-changelog

### DIFF
--- a/tests/integration/test_store.py
+++ b/tests/integration/test_store.py
@@ -148,11 +148,11 @@ def test_store_rate_limit(mini_sentry, relay):
     with pytest.raises(HTTPError):
         relay.send_event(project_id, {"message": "invalid"})
 
-    # Generated outcome has no reason code:
+    # Generated outcome has reason code 'generic':
     outcome_envelope = mini_sentry.captured_events.get(timeout=1)
     outcome = json.loads(outcome_envelope.items[0].payload.bytes)
     assert outcome["rate_limited_events"] == [
-        {"reason": "", "category": "error", "quantity": 1}
+        {"reason": "generic", "category": "error", "quantity": 1}
     ]
 
     # This event should arrive
--- a/tests/integration/test_store.py
+++ b/tests/integration/test_store.py
@@ -135,7 +135,7 @@ def test_store_rate_limit(mini_sentry, relay):
             return "", 429, {"retry-after": "2"}
 
     # Disable outcomes so client report envelopes do not interfere with the events we are looking for
-    config = {"outcomes": {"emit_outcomes": False}}
+    config = {"outcomes": {"emit_outcomes": "as_client_reports"}}
     relay = relay(mini_sentry, config)
     project_id = 42
     mini_sentry.add_basic_project_config(project_id)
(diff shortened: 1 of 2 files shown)


tests/integration/test_store.py
```diff
@@ -148,6 +148,13 @@ def store_event():
     with pytest.raises(HTTPError):
         relay.send_event(project_id, {"message": "invalid"})
 
+    # Generated outcome has reason code 'generic':
+    outcome_envelope = mini_sentry.captured_events.get(timeout=1)
+    outcome = json.loads(outcome_envelope.items[0].payload.bytes)
+    assert outcome["rate_limited_events"] == [
+        {"reason": "generic", "category": "error", "quantity": 1}
+    ]
```
jjbayer, line 156: In https://github.com/getsentry/relay/pull/1967/commits/e62d5695ed5dcfc0e07c2ea6bf2c77d8a9e4f90d, this outcome had `"reason": ""`.

olksdr (approved):

PR #1970, opened 2023-03-27
feat(normalization): Changes device class logic for iphone devices to use a mode to class mapping instead of checking processor_frequency

edwardgou-sentry:
Sentry SDKs on iOS devices are unable to retrieve `processor_frequency` which we needed to determine the `device.class` for iPhones. This is a workaround by adding a hardcoded iPhone model to `device.class` mapping instead.


relay-general/src/protocol/device_class.rs
```diff
@@ -50,3 +45,74 @@ impl fmt::Display for DeviceClass {
...
+    "iPhone14,2",
+    "iPhone14,3",
+    "iPhone14,6",
+    "iPhone14,7",
+    "iPhone14,8",
+    "iPhone15,2",
+    "iPhone15,3",
+];
+
+fn device_model_to_class(device_model: &str) -> Option<DeviceClass> {
```
iker-barriocanal: This function is mapping from `str`s to `DeviceClass`. I think the best approach is to have a `Map`?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Features**
 
 - Allow monitor checkins to paass `monitor_config` for monitor upserts. ([#1962](https://github.com/getsentry/relay/pull/1962))
+- Adds a model to device class mapping for iPhones. ([#1970](https://github.com/getsentry/relay/pull/1970))
```
iker-barriocanal: nit: this PR changes the way we set the device class (instead of using the processor frequency, we use the device's model.

relay-general/src/protocol/device_class.rs
```diff
@@ -17,15 +17,10 @@ impl DeviceClass {
...
                 if family == "iPhone" || family == "iOS" || family == "iOS-Device" {
-                    if let Some(processor_frequency) = device.processor_frequency.value() {
-                        if processor_frequency < &2000 {
-                            return Some(DeviceClass::LOW);
-                        } else if processor_frequency < &3000 {
-                            return Some(DeviceClass::MEDIUM);
-                        } else {
-                            return Some(DeviceClass::HIGH);
-                        }
+                    if let Some(model) = device.model.value() {
```
iker-barriocanal, line 22: Related to my comment above -- what if there's no model, but the processor frequency is available? We can prioritize using the model over the frequency, but changing this behavior is a breaking change. Shouldn't the tests below still pass, _and_ new tests for this new functionality also pass?

iker-barriocanal (commented):
nit: the context of the PR should be "normalization" or something similar.

CHANGELOG.md
iker-barriocanal: This PR changes normalization, and as per the [changelog instructions](https://github.com/getsentry/relay#instructions-for-changelogs), please also update the `py/CHANGELOG`.

edwardgou-sentry: Updated!

relay-general/src/protocol/device_class.rs
edwardgou-sentry: Hey @iker-barriocanal , so this feature isn't being used by/released to any customers yet, we are in the process of rolling it out which is why I think we can go back on this logic. I confirmed with the sdk team that iOS devices are unable to send `processor_frequency` so there should never be a case where we would enter the code block to check for `processor_frequency` which is why I removed it to simplify. Let me know if this is fine!

relay-general/src/protocol/device_class.rs
edwardgou-sentry: Hey @iker-barriocanal, I wasn't able to find a way to easily create a static HashMap. Some recommendations online were to use `lazy_static` or `phf`, but I noticed we don't import/use these anywhere in our code. Is the concern here that `match` might not be as performant as something like a `Map`?

iker-barriocanal: I'd use `Lazy`, example in the codebase with a `BTreeMap` [here](https://github.com/getsentry/relay/blob/e8b2acc2cedd38db4ab5bc5d5d13f9996ceca528/relay-general/src/pii/builtin.rs#L12), and docs [here](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html#). I believe you'll be able to use [`get`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get) in `from_contexts` and get rid of this method.

I believe `match` is pretty fast, but in each arm of the `match` there are plenty of string comparisons. However, I'm pointing this out for code readability more than performance itself.

iker-barriocanal (commented):
The PR looks good, but let's update the mapping first.

edwardgou-sentry:
Thanks @iker-barriocanal ! I've replaced the match function with a map as suggested!

relay-general/src/protocol/device_class.rs
```diff
@@ -50,3 +47,57 @@ impl fmt::Display for DeviceClass {
         self.0.to_string().fmt(f)
     }
 }
+
+static APPLE_DEVICE_MODEL_TO_CLASS_MAP: Lazy<BTreeMap<&str, DeviceClass>> = Lazy::new(|| {
+    let mut map = BTreeMap::new();
+    map.insert("iPhone1,1", DeviceClass::LOW);
+    map.insert("iPhone1,2", DeviceClass::LOW);
```
iker-barriocanal: You could simplify this and make it more readable with `BTreeMap::from`, see example in [rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cfdd8ef7be54e9d20b3ba71863174b85).

iker-barriocanal (approved):
Nice!

PR #1972, opened 2023-03-28
doc: document pii fields

### LINKED ISSUE #6087: List fields scrubbed by default data scrubbers
### Core or SDK?

Core Sentry product

### Which part? Which one?

https://docs.sentry.io/product/data-management-settings/scrubbing/server-side-scrubbing/

### Description

The current documentation described nicely what we try to scrub from events, but not where we scrub the information from.

For example, credit card numbers are scrubbed from additional context, but not from tags or exception value. See the example event that uses randomly generated credit card number and IP address. The same value was used on all fields with same name (password, IP and card).

Screenshot 2023-01-17 at 11 59 21

### Suggested Solution

Have a list that informs where the scrubbed is applied (fields that match `pii = "true"` in the relay repository where the scrubbing is done).

TBS1996:
A tool which takes the path of a rust crate or workspace and outputs all the fields of structs or enums with the metaattribute 'pii = "true"' (or custom values), with the full path. E.g. `Event.csp.effective_directive`. 

The current use case will be to get all the pii=true fields under the Event struct, in order to display it in sentry-docs.

if needed, in the future it can be trivially generalized to look for any custom metaattribute key, not just "pii". 

I made the interface similar to the document-metrics tool in order to make it easier to maintain. 

I have a complementary PR on `sentry-docs`: https://github.com/getsentry/sentry-docs/pull/6730
 
Example output: https://github.com/getsentry/relay/blob/10283cd7b15b6ce80599f17b478bb763110eb5ec/tools/document-pii/src/snapshots/document_pii__tests__pii_true.snap
How it might look on the docs: https://i.imgur.com/R0K2xba.png

#skip-changelog


tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,575 @@
...
+
+    #[derive(Debug)]
+    static ref PII_VALUES: Mutex<Vec<String>> = Mutex::new(vec![]);
+}
+
+#[derive(PartialEq, Eq, Hash, Clone, Debug, Default)]
+struct TypeAndField {
+    type_name: String, //shoud be full path
+    field_name: String,
+}
```
iker-barriocanal: For better readability, what do you think of the following?
```suggestion
struct TypeAndField {
    rust_path: String,
    name: String,
}
```
Suggesting `rust_path` to note the difference between a filesystem path.

iker-barriocanal: Also, IIUC, should we rename `TypeAndField` to `Token` or something similar? It seems this is the word used to refer to this concept from what I see googling a bit (like [here](https://ruairidh.dev/build-your-own-ast/)), although I may be wrong.

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,575 @@
...
+
+fn find_rs_files(dir: &str) -> Vec<std::path::PathBuf> {
+    let walker = WalkDir::new(dir).into_iter();
+    let mut rs_files = Vec::new();
+
+    for entry in walker.filter_map(walkdir::Result::ok) {
+        if entry.clone().path().to_string_lossy().contains("target/") {
+            continue;
+        }
+        if entry.file_type().is_file() && entry.path().extension().unwrap_or_default() == "rs" {
```
iker-barriocanal: Does this panic if a file doesn't have an extension?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,575 @@
...
+}
+
+#[derive(Debug, Clone)]
+enum EnumOrStruct {
+    ItemStruct(ItemStruct),
+    ItemEnum(ItemEnum),
+}
+
+// Need this to wrap lazy_statics in a mutex. everything is single-threaded so it shouldn't matter.
+unsafe impl Send for EnumOrStruct {}
```
iker-barriocanal: There are some mutexes at the top. Why do we need them?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,575 @@
...
+        }
+    }
+    rs_files
+}
+
+#[derive(Debug, Clone)]
+enum EnumOrStruct {
+    ItemStruct(ItemStruct),
+    ItemEnum(ItemEnum),
+}
```
iker-barriocanal: Related to my comment above, I'd change this to `TokenType`. We can also simplify the enum options by removing `Item`:
```suggestion
enum EnumOrStruct {
    Struct(ItemStruct),
    Enum(ItemEnum),
}
```

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,575 @@
...
+            rs_files.push(entry.into_path());
+        }
+    }
+    rs_files
+}
+
+#[derive(Debug, Clone)]
+enum EnumOrStruct {
+    ItemStruct(ItemStruct),
+    ItemEnum(ItemEnum),
```
iker-barriocanal: Have we considered [type aliases](https://doc.rust-lang.org/reference/items/type-aliases.html)? It's okay if we don't want to include these at the moment and approach it in a later iteration; but in that case, could we add a comment to note that?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,575 @@
...
+    fn visit_item_use(&mut self, node: &'ast syn::ItemUse) {
+        if should_keep_itemuse(node) {
+            self.use_statements.push(node.clone());
+        }
+        syn::visit::visit_item_use(self, node);
+    }
+}
+
+// We are only interested in the use-statements that are defined in relay, which is the ones
+// start start with "relay" or "crate".
```
iker-barriocanal: § `super`?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    if !Path::new(&format!("{}/Cargo.toml", &chosen_dir)).exists() {
+        panic!("Cargo.toml not found. Either run this script from a rust crate/workspace, or pass in a valid path with the -p flag");
+    }
+
+    let paths = find_rs_files(chosen_dir);
+
+    if paths.is_empty() {
+        panic!("No rust files found in {current_dir}");
```
iker-barriocanal: I don't think we should panic here. If we choose a directory and there's nothing to deal with, that's ok.

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,574 @@
...
+        output = output.replace("{{Unnamed}}.", &unnamed_replace);
+        output_vec.push(output);
+    }
+    output_vec.sort();
+    for x in &output_vec {
+        println!("{x}");
+    }
+}
+
+fn rust_file_to_use_path(file_path: &Path) -> String {
```
iker-barriocanal: IIUC this method attempts to get the `use` path from a rust file, and we are trying to do that by dealing with the file system. Is there a way to get this by information in the AST?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,574 @@
...
+        .map(|part| part.as_os_str().to_string_lossy().to_string())
+        .filter(|part| part != "src")
+        .collect::<Vec<_>>();
+
+    if is_module {
+        module_path.push(file_stem);
+    }
+
+    let mut use_path = module_path.join("::");
+    use_path = use_path.split_once("rust::").unwrap().1.into();
```
TBS1996: im realizing now this is referencing the folder structure on my local PC, looking into fixing this now

iker-barriocanal (changes requested):
Since you can run it from the CLI and you're including the command in the PR description (good!), including the output would also be helpful to understand what the code is trying to do before diving into it. If you find the output is too long, you could consider using the [`<details>`](https://gist.github.com/scmx/eca72d44afee0113ceb0349dd54a84a2) feature.

I see that, in general, we're building a lot of custom logic by moving through the file system. Is there more information the AST provides us that we can use, to avoid rebuilding our own logic?

tools/document-pii/src/main.rs
TBS1996: no, the .extension() returns an Option with a possible extension, so this is just normalizing it. I'll try to rewrite it for readability

tools/document-pii/src/main.rs
TBS1996: I needed them to mutate them, as it has to implement DerefMut, although I could change to refcell instead of mutex

tools/document-pii/src/main.rs
TBS1996: good call, I was unnecessarily focused on namespaces here

tools/document-pii/src/main.rs
TBS1996: hmm im sorry i dont get what you mean, what do you mean that we dont want to include these at the moment?

tools/document-pii/src/main.rs
TBS1996: oh damn, I didn't think of that. thanks! need to look into that

tools/document-pii/src/main.rs
TBS1996: yeah agreed, I wrote it before i put the logic about checking for a Cargo.toml file.

tools/document-pii/src/main.rs
TBS1996: unfortunately not, that would have made this task a lot easier

tools/document-pii/src/main.rs
TBS1996: im not sure if i agree, 'Token' may be more technical but wouldn't it be more readable with a name that decribes the content more literally?

tests/test_pii_docs/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2021"
+
+
+[workspace]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+synstructure = "0.12.3"
+proc-macro2 = "1.0.8"
```
iker-barriocanal: Since this is for the test project, why do we need these?

tests/test_pii_docs/src/main.rs
```diff
@@ -0,0 +1,39 @@
+fn main() {
+    println!("Hello, world!");
+}
+
+pub mod anotherfile;
```
iker-barriocanal: nit (?)
```suggestion
pub mod anothermod;
```

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+
+// Due to global variables, have to both run the tests sequentially as well as clearing the globals
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
+
+    use serial_test::serial;
+
+    use super::*;
+    const RUST_CRATE: &str = "../../tests/test_pii_docs";
```
iker-barriocanal:
```suggestion
    const TEST_CRATE: &str = "../../tests/test_pii_docs";
```

iker-barriocanal: We only use this variable to create the `PathBuf` in unit tests. Could we create it directly here?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+    }
+
+    #[test]
+    #[serial]
+    fn test_use_statements() {
+        clear_globals();
+        let rust_crate = PathBuf::from_str(&RUST_CRATE).unwrap();
+        run(&rust_crate, &vec![], None);
+        let use_statements = USE_STATEMENTS.lock().unwrap().clone();
+        insta::assert_debug_snapshot!(use_statements);
```
iker-barriocanal: With tests, we want to ensure that the behavior is what we expect; i.e. the program produces the correct output. We're looking for an output format similar to what's in the PR description, but it seems to me that tests are checking intermediate variables instead, which doesn't provide much value. Could we add a test to ensure the output is what we expect? And we can get rid of the other tests.

tools/document-pii/src/main.rs
iker-barriocanal: Can we do that? I don't think we need mutexes.

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+
+        UseTree::Rename(use_rename) => {
+            leading_path.segments.push(use_rename.rename.clone().into());
+            quote::quote!(#leading_path).to_string()
+        }
+        _ => quote::quote!(#leading_path).to_string(),
+    }
+}
+
+fn print_error(error: &anyhow::Error) {
```
iker-barriocanal, line 136: Is this a helper method for development purposes?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+    // When iterating recursively, this keeps track of the current path
+    static ref CURRENT_PATH: Mutex<Vec<TypeAndField>> = Mutex::new(vec![]);
+    // a set of CURRENT_PATH whenever it hits a pii=true field
+    static ref PII_TYPES: Mutex<BTreeSet<Vec<TypeAndField>>> = Mutex::new(BTreeSet::new());
+    // All the structs and enums, where the key is the full path. The string in the tuple represents
+    // the module path to the item. Now that I think of it, should be able to just get it from the key.
+    static ref ALL_TYPES: Mutex<HashMap<String, EnumOrStruct>> = Mutex::new(HashMap::new());
+    /// The values that we wanna match on, (true, false, maybe)
+    static ref PII_VALUES: Mutex<Vec<String>> = Mutex::new(vec![]);
+    /// All the use statements in a given module.
```
iker-barriocanal: Could we add to the docstrings here what the `BTreeMap` maps from and to? It'd help reading the code.

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+            };
+            let file_content = fs::read_to_string(path.as_path()).unwrap();
+            let syntax_tree: syn::File = syn::parse_file(&file_content).unwrap();
+            hashmap_filler.visit_file(&syntax_tree);
+        }
+    }
+}
+
+impl<'ast> Visit<'ast> for FileSyntaxVisitor {
+    fn visit_item_struct(&mut self, node: &'ast ItemStruct) {
```
iker-barriocanal:
```suggestion
    fn visit_struct(&mut self, node: &'ast ItemStruct) {
```

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+
+        let struct_name = format!("{}::{}", self.module_path, node.ident);
+        self.insert_use_statements(vec![struct_name.clone()]);
+        ALL_TYPES
+            .lock()
+            .unwrap()
+            .insert(struct_name, EnumOrStruct::Struct(node.clone()));
+    }
+
+    fn visit_item_enum(&mut self, node: &'ast ItemEnum) {
```
iker-barriocanal:
```suggestion
    fn visit_enum(&mut self, node: &'ast ItemEnum) {
```

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+        let enum_name = format!("{}::{}", self.module_path, node.ident);
+        self.insert_use_statements(vec![enum_name.clone()]);
+        ALL_TYPES
+            .lock()
+            .unwrap()
+            .insert(enum_name, EnumOrStruct::Enum(node.clone()));
+    }
+
+    fn visit_item_use(&mut self, i: &'ast syn::ItemUse) {
```
iker-barriocanal:
```suggestion
    fn visit_use_statement(&mut self, i: &'ast syn::ItemUse) {
```
Not very sure about this.

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+
+        self.insert_use_statements(use_statements);
+    }
+}
+
+#[derive(Ord, PartialOrd, PartialEq, Eq, Hash, Clone, Debug, Default)]
+struct TypeAndField {
+    // Full path of a type. E.g. relay_common::protocol::Event, rather than just 'Event'.
+    qualified_type_name: String,
+    field_name: String,
```
iker-barriocanal: Was this the same as `ident`? Can we use the same name for the same concept?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,650 @@
...
+        }
+    }
+
+    fn visit_variant(&mut self, node: &'ast Variant) {
+        for field in node.fields.iter() {
+            self.visit_field(field);
+        }
+    }
+
+    fn visit_field(&mut self, node: &'ast Field) {
```
iker-barriocanal: This method is doing several things, could we split it up to make it easier to read and understand?

iker-barriocanal (changes requested):

tools/document-pii/src/main.rs
TBS1996: you cant make a PathBuf as a normal const unfortunately, so would have to do some something like lazy_static

tools/document-pii/src/main.rs
TBS1996: while the use_statements is an intermediate variable, its not "instead" because we also test the final output, I guess you could make a case that its redundant but at least like this if something breaks then it makes it easier to debug, whats your thoughts on that?

tools/document-pii/src/main.rs
TBS1996: would be useful for debugging mainly I think, if you give some wrong values. I copied it from tools/document-metrics as I decided to make my code structure the same as that one, I think that will make it easier to maintain in the future

tools/document-pii/src/main.rs
TBS1996: good idea! done.

tools/document-pii/src/main.rs
TBS1996: I can't change this, as it's a part of a pre-defined trait

tools/document-pii/src/main.rs
TBS1996: in the rust source code, I believe ident is always just the name of the item, not including the full path which is what we have here.

tools/document-pii/src/main.rs
TBS1996: Ok, so the function does 3 things:

1. push and pop to/from the CURRENT_PATH
2. if it's PII then add it to PII_VALUES
3. recursion logic (with the double for-loops)

since both `1` and `2` are just simple statements without any branching, I moved the recursion logic to a separate function, is this what you had in mind?

.github/workflows/deploy.yml
TBS1996, line 1: looking into ways of unifying the job of exporting metrics and exporting PII fields atm

iker-barriocanal, line 1: I think the PR diff is big enough with the script itself. Increasing the scope and adding more complexity is going to make the review harder, and we should consider splitting the work into several PRs.

tools/document-pii/src/main.rs
iker-barriocanal: I think that the variable doesn't represent what the output could look like. If there's a bug in the piece of code that translates that variable to the actual output, this test wouldn't catch that and pass.

tools/document-pii/src/main.rs
iker-barriocanal: Is the full path the same as `qualified_type_name`? What is the difference between `field_name` and `ident`?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+        self.recursion(node);
+
+        CURRENT_PATH.lock().unwrap().pop();
+    }
+}
+
+/// if you have a field such as "foo: Foo<Bar<Baz>>" this function can take the type of the field
+/// and return a vector of the types like: ["Foo", "Bar", "Baz"].
+fn type_to_string(ty: &Type) -> Vec<String> {
+    fn recursion(ty: &Type, segments: &mut Vec<String>) {
```
iker-barriocanal: It's pretty hard to know what `recursion` is supposed to accomplish without reading the code. Could we choose a name that better describes its goal?

Also, is there any reason to have `recursion` inside `type_to_string`?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+}
+
+fn main() {
+    let cli = Cli::parse();
+
+    match cli.run() {
+        Ok(()) => (),
+        Err(error) => {
+            print_error(&error);
+            std::process::exit(1);
```
iker-barriocanal: We could return a `Result` from main, and if there's an error, it will automatically be printed out and exist with code 1.

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+    qualified_type_name: String,
+    field_name: String,
+}
+
+fn find_rs_files(dir: &PathBuf) -> Vec<std::path::PathBuf> {
+    let walker = WalkDir::new(dir).into_iter();
+    let mut rs_files = Vec::new();
+
+    for entry in walker.filter_map(walkdir::Result::ok) {
+        if !entry.clone().path().to_string_lossy().contains("/src/") {
```
iker-barriocanal: nit: why do we need to clone the entry here?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+    static ref PII_TYPES: Mutex<BTreeSet<Vec<TypeAndField>>> = Mutex::new(BTreeSet::new());
+    // All the structs and enums, where the key is the full path. The string in the tuple represents
+    // the module path to the item. Now that I think of it, should be able to just get it from the key.
+    static ref ALL_TYPES: Mutex<HashMap<String, EnumOrStruct>> = Mutex::new(HashMap::new());
+    /// The values that we wanna match on, (true, false, maybe)
+    static ref PII_VALUES: Mutex<Vec<String>> = Mutex::new(vec![]);
+    /// Maps the name of the module to all of the use statements therein, plus the path of the
+    /// types that are defined there.
+    static ref USE_STATEMENTS: Mutex<BTreeMap<String, BTreeSet<String>>> = Mutex::new(BTreeMap::new());
+}
```
olksdr: Why do you use `lazy_static`, with `Mutex` and with a lot of globs? Why do not just keep the data in the structs which manipulate the data, initialize everything in `main` and pass the data around when needed?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+    fn insert_use_statements(&self, use_statements: Vec<String>) {
+        USE_STATEMENTS
+            .lock()
+            .unwrap()
+            .entry(self.module_path.clone())
+            .or_default()
+            .extend(use_statements);
+    }
+
+    fn fill_pii_hashmaps(paths: &Vec<PathBuf>) {
```
iker-barriocanal: This method is only called once and doesn't use `self`, is there any reason to not extract it from `FileSyntaxVisitor`?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+    current_type: String,
+}
+
+impl TypeVisitor {
+    /// Takes a Field (could be struct or enum) and visit the types that it consist of if we have
+    /// the full path to it in USE_STATEMENTS.
+    fn recursion(&mut self, node: &Field) {
+        let local_paths = USE_STATEMENTS
+            .lock()
+            .unwrap()
```
olksdr: Moving all the data into the places when the actual data is needed will eliminate usage of `Mutex`, which is not needed at all, and also will get rid of all those unwraps everywhere.

tools/document-pii/src/main.rs
TBS1996: Whenever I try to change it to something else it doesn't seem to be an improvement. Would you prefer btw if we make a single Globals struct which has all those types in the lazy_static as fields?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+}
+
+#[derive(Ord, PartialOrd, PartialEq, Eq, Hash, Clone, Debug, Default)]
+struct TypeAndField {
+    // Full path of a type. E.g. relay_common::protocol::Event, rather than just 'Event'.
+    qualified_type_name: String,
+    field_name: String,
+}
+
+fn find_rs_files(dir: &PathBuf) -> Vec<std::path::PathBuf> {
```
iker-barriocanal, line 30: In `fill_pii_hashmaps` we use `Path`s instead of `PathBuf`s. Could we change the return type and make the following methods easier?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+            .unwrap()
+            .entry(self.module_path.clone())
+            .or_default()
+            .extend(use_statements);
+    }
+
+    fn fill_pii_hashmaps(paths: &Vec<PathBuf>) {
+        for path in paths {
+            // first iterate through all rust files to create the hashmap
+            let mut hashmap_filler = FileSyntaxVisitor {
```
iker-barriocanal: Could we set a name for `hashmap_filler` that better represents its goal?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+        .split_at(use_path.find(crate_name).unwrap())
+        .1
+        .replace('-', "_");
+
+    use_path
+}
+
+/// Checks if a file is a module, which is needed to convert a file path to a module path
+fn is_file_module(file_path: &Path) -> bool {
+    let parent_dir = file_path.parent().unwrap();
```
olksdr: Instead of using `unwrap`s , which does not really matter for tooling , but we still wanna enforce best practices, return the `Result` and propagate the error to the `main` (maybe), where the error could be handled and properly printed / showed to the caller, with proper error message. 
You could also use `anyhow` crate for this and add the `context` to each error to provide even more information on why the error happens. 


Same goes about all the `unwrap`s in this and other functions.

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,696 @@
...
+
+        // Before we can iterate over the pii fields properly, we make a mapping between all
+        // paths to types and their AST node.
+        FileSyntaxVisitor::fill_pii_hashmaps(&rust_file_paths);
+
+        {
+            // Sets the pii_values from Cli to a global variable so it can be accessed in the
+            // 'Visit' trait.
+            let mut guard = PII_VALUES.lock().unwrap();
+            guard.clear();
```
iker-barriocanal: At the start of the program the global variables are empty. Is there any reason why we are clearing an already-clear vector?

iker-barriocanal (changes requested):
Blocking the PR for the usage of the mutexes, let's get rid of them. On the other hand, we can do some work to improve the readability of the code: better names, splitting big functions into smaller ones (see the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)), etc.

olksdr (dismissed):
I have lots of questions about the chosen approach to structure the code and passing data around. 

Even though this is just a cli tool, but we still have to enforce some code quality and best practices: 
* like returning `Result` to propagate the error, instead of unwrapping things everywhere
* stop using `Mutex` where it's not needed
* remove globs, which do not really needed here either
* add more documentation (some functions and structs do not have anything)


Also, please, resolve the conflicts in the Cargo.lock to make sure that CI runs.

tools/document-pii/src/main.rs
TBS1996: It's very related to it so thats why i put it there, I added the Self keyword now since it creates many FileSyntaxVisitors in a loop, do you think i should extract it anyway?

tools/document-pii/src/main.rs
TBS1996: hmm is there a best practice wrt to this pattern in general? I want to have a function that returns a Vec<String> as thats cleaner to deal with on the outside, but I need to build it recursively with a function that mutates a vector. 

I thought putting it inside is best to make it clear it's only needed for this function

.github/workflows/deploy.yml
TBS1996: ok, I removed it now

tools/document-pii/src/main.rs
TBS1996: I think itd add some complexity as it's the equivalent of &str so I can't as easily return it from a function, ill look into it

tools/document-pii/src/main.rs
TBS1996: I thought i should make the interface identical to document-metrics for simplicity, which uses the print_error way, would your suggestion change any behaviour? I can do it of course, just wondering why it wasnt done like this on the metrics tool

tools/document-pii/src/main.rs
TBS1996: yes thats the same, and I didn't use the word 'ident' in my code but I guess you are referring to the way its used in the rust source. Ident is used as identifier in general, not just for fields, so It's not as explicit to just call it ident. I could rename field_name to field_ident though!

tools/document-metrics/Cargo.toml
```diff
@@ -14,7 +14,7 @@ clap = { version = "4.1.4", features = ["derive"] }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.55"
 serde_yaml = "0.9.17"
-syn = "1.0.39"
+syn = { version = "1.0.39", features = ["full"] }
```
TBS1996: I'm pretty confused why I had to do this, after merging with master this crate suddenly complained it needed the "full" feature flag, even though it worked without it before

this is the document-metrics crate which i didn't touch at all.

TBS1996:
`fn test_convert_sensitive_fields_too_large()` is now failing, although I didn't touch any code in relay beside the tool i made. I think it must have something to do with the updated Cargo.lock file, when I merged i had merge conflicts on it, and then I simply deleted it and then 'cargo build' to generate a new one.

according to the test, the "problem" seems to be that im able to compile the regex even though it shouldn't be? I'm looking into it but any input is appreciated

tools/document-pii/src/main.rs
TBS1996: true but there already is code that directly tests the output, this use_statements test is still helpful if i want to modify the code dealing with use_statements

Cargo.lock
```diff
@@ -2859,13 +2912,22 @@ dependencies = [
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "567664f262709473930a4bf9e51bf2ebf3348f2e748ccc50dea20646858f8f29"
+dependencies = [
+ "bitflags",
+]
+
 [[package]]
 name = "regex"
-version = "1.7.1"
+version = "1.8.1"
```
olksdr: There was quite big release for the regex crate https://github.com/rust-lang/regex/blob/master/CHANGELOG.md#180-2023-04-20 which changes the behaviour a little bit and adds quite some improvements. 

That's can be the reason why one of the tests is failing now.

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,692 @@
...
+                module_path: rust_file_to_use_path(path.as_path()),
+                use_statements: HashMap::new(),
+                all_types: HashMap::new(),
+            };
+            let file_content = fs::read_to_string(path.as_path())?;
+            let syntax_tree: syn::File = syn::parse_file(&file_content)?;
+            visitor.visit_file(&syntax_tree);
+            node_map.extend(visitor.all_types);
+            use_statements.extend(visitor.use_statements);
+        }
```
olksdr: Why do you create `Self` per path. The visitor already have access to self and can insert in the fields everything what's needed, I do not understand the reason this is done this way. Could you, please, explain your thinking behind it?

tools/document-pii/src/main.rs
olksdr: @TBS1996  why did you resole this without addressing the comment? 
Can we, please, remove unwraps and return `Result` here?

olksdr (commented):
The `main.rs` is quite big and difficult to navigate. Please, split the main and move the parsing part into the separate module. 
This way the code organization will be better and easier to see what's going one.

tools/document-pii/src/main.rs
TBS1996: I believe it was an oversight when doing refactoring, and it's no longer needed, thanks for pointing it out, it now only creates one

TBS1996:
@olksdr 

I've now refactored into 3 files, and removed some unwraps I overlooked. It now only unwraps within the trait methods, since I cannot change the return signature, should I let it be like that or maybe do some other thing?

Cargo.lock
TBS1996: Ok, so im wondering if we want to continue having this test, why should we test that a regex is failing at a certain point?

tools/document-pii/src/main.rs
TBS1996: my bad, I did it on a different function and didn't see that this one also would require a change

Cargo.lock
iker-barriocanal: The test has the following comment:

> The data won't be scrubbed here, since the regex cannot be compiled.

Testing a sensitive field is not being scrubbed when a regex cannot be compiled seems a legit use case to me. Why do you think it's not?

Cargo.lock
iker-barriocanal, line 1: There are a bunch of package upgrades, do we need them all? Have you tested there are no breaking changes on each of them? Or known security vulnerabilities?

tools/document-pii/Cargo.toml
iker-barriocanal, line 1: nit: let's remove the empty lines in this file.

tools/document-pii/src/main.rs
iker-barriocanal: I may have missed it, but what test is testing the output? The output should be something like it used to be in the PR description.

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,266 @@
...
+    pii_values: &Vec<String>,
+) -> anyhow::Result<BTreeSet<Vec<TypeAndField>>> {
+    let mut pii_types = BTreeSet::new();
+    let mut current_path = vec![];
+    match item {
+        Some(path) => {
+            if let Some(structorenum) = { all_types.get(path).cloned() } {
+                let module_path = path
+                    .rsplit_once("::")
+                    .ok_or_else(|| anyhow!("invalid module path: {}", path))?
```
iker-barriocanal: I'm not familiar with the `anyhow` usage, but is this how we're supposed to use it? My understanding is that you let the error pop up and add more context with `.context()`.

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,266 @@
...
+                    if ident == name {
+                        if let Some(proc_macro2::TokenTree::Punct(punct)) = iter.next() {
+                            if punct.as_char() == '=' {
+                                if let Some(proc_macro2::TokenTree::Literal(lit_val)) = iter.next()
+                                {
+                                    let mut lit_val = lit_val.to_string();
+                                    lit_val.pop(); // remove superfluous quotes. "\"true\"" -> "true"
+                                    lit_val.remove(0);
+                                    if lit_val == value {
+                                        return true;
```
iker-barriocanal: Here we have 9 indentation levels, could we split the function into multiple so that it's more readable?

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,266 @@
...
+        for field_type in &field_types {
+            for use_path in &local_paths {
+                if use_path.split("::").last().unwrap() == field_type.trim() {
+                    let enum_or_struct = { self.all_types.get(use_path).cloned() };
+
+                    if let Some(enum_or_struct) = enum_or_struct {
+                        match enum_or_struct {
+                            EnumOrStruct::Enum(itemenum) => {
+                                let current_type = self.current_type.clone();
+                                let module_path = self.module_path.clone();
```
iker-barriocanal: Here we have 6 indentation levels too. In general, don't be afraid to split your code into functions, even if it's only called once at this point. It makes the code easier to read, understand, extend, and test.

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,251 @@
...
+use crate::EnumOrStruct;
+
+use anyhow::anyhow;
+use syn::punctuated::Punctuated;
+use syn::visit::Visit;
+use syn::{ItemEnum, ItemStruct, UseTree};
+
+/// Iterates over the rust files to collect all the types and use_statements, which is later
+/// used for recursively looking for pii_fields afterwards.
+pub struct ItemCollector {
```
iker-barriocanal: Since we're dealing with the AST:
```suggestion
pub struct AstItemCollector {
```

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// Gets both a mapping of the full type to a type and its actual AST node, and also the
+    /// use_statements in its module, which is needed to fetch the types that it referes to in its
+    /// fields.
+    #[allow(clippy::type_complexity)]
+    pub fn get_types_and_use_statements(
+        paths: &Vec<PathBuf>,
+    ) -> anyhow::Result<(
+        HashMap<String, EnumOrStruct>,
+        BTreeMap<String, BTreeSet<String>>,
+    )> {
```
iker-barriocanal: I agree with clippy here -- I think we could make this method simpler than this return type. If we make a struct with meaningful names, that's already a great start!

iker-barriocanal (changes requested):
I haven't done an in-depth review since tests are not passing, let's make sure all tests are passing to ensure we're not introducing any breaking changes.

In general, I find the code a bit hard to read and understand. I think we could start by splitting methods into smaller ones, grouping the same logic together, and using more meaningful names.

I remember reviewing the PR and seeing a description of the expected output of this tool. There's now a screenshot on the result of this PR + other code, but it's not clear what this tool is expected to produce. We could make this very clear with a test and adding the output to the PR description. What do you think?

tools/document-pii/src/main.rs
TBS1996: `test_pii_true`, `test_pii_false` and `test_pii_all` are testing the output, if you look at the snapshot files its how the files that get sent to the sentry docs will look like

Cargo.lock
TBS1996: well we don't want the regex to fail, and now the new update to regex makes it not fail, so id say thats a good thing, we can't keep the test as it is now unless we try to stop regex from being upgraded

Cargo.lock
TBS1996: I never intended to upgrade anything, I'm trying to revert it

TBS1996: ok fixed it

tools/document-pii/src/item_collector.rs
TBS1996: Ok I did it now, I thought it'd be ugly but then i remembered destructuring structs is possible, so now the original reason for the tuple is gone :P

tools/document-pii/src/pii_finder.rs
TBS1996: ok fixed, it's now flatter than denmark

tools/document-pii/src/pii_finder.rs
TBS1996: revised it now

tools/document-pii/src/pii_finder.rs
TBS1996: do you mean like appending multiple error messages instead of exiting at the first error ?

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,281 @@
...
+            Type::Array(arr) => self.visit_field_types(&arr.elem),
+            Type::BareFn(bare_fn) => bare_fn
+                .inputs
+                .iter()
+                .for_each(|ty| self.visit_field_types(&ty.ty)),
+            Type::Group(group) => self.visit_field_types(&group.elem),
+            Type::ImplTrait(_) => {}
+            Type::Infer(_) => {}
+            Type::Macro(_) => {}
+            Type::Never(_) => {}
```
TBS1996: left them like this so it's easier to see if i missed some inner Type

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,293 @@
...
+
+pub struct TypesAndUseStatements {
+    // Maps the path name of an item to its actual AST node.
+    pub all_types: HashMap<String, EnumOrStruct>,
+    // Maps the use_statements to different modules. For use in constructing the full path
+    // of an item from its type name.
+    pub use_statements: BTreeMap<String, BTreeSet<String>>,
+}
+
+impl AstItemCollector {
```
olksdr, line 73: Can we, please, put `impl` right after the struct definition (according our [`style guidelines`](https://develop.sentry.dev/rust/#declaration-order)) ?

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,295 @@
...
+
+use anyhow::anyhow;
+use syn::punctuated::Punctuated;
+use syn::visit::Visit;
+use syn::{ItemEnum, ItemStruct, UseTree};
+
+use crate::EnumOrStruct;
+
+/// Iterates over the rust files to collect all the types and use_statements, which is later
+/// used for recursively looking for pii_fields afterwards.
```
olksdr: This is the container, and does not iterate over anything.  
```suggestion
/// The types and use statements items collected from the rust files.
```

tools/document-pii/src/item_collector.rs
olksdr, line 1: Some top-level docs would be great here.

tools/document-pii/src/pii_finder.rs
olksdr, line 1: Some top-level docs would be great here as well.

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,281 @@
...
+        .filter(|use_path| {
+            let last_use_path = use_path.split("::").last().unwrap().trim();
+            field_types
+                .iter()
+                .any(|field_type| field_type.trim() == last_use_path)
+        })
+        .collect()
+}
+
+impl<'a> PiiFinder<'a> {
```
olksdr: Please, move `impl` to right after the struct definition.

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,281 @@
...
+            Type::Array(arr) => self.visit_field_types(&arr.elem),
+            Type::BareFn(bare_fn) => bare_fn
+                .inputs
+                .iter()
+                .for_each(|ty| self.visit_field_types(&ty.ty)),
+            Type::Group(group) => self.visit_field_types(&group.elem),
+            Type::ImplTrait(_) => {}
+            Type::Infer(_) => {}
+            Type::Macro(_) => {}
+            Type::Never(_) => {}
```
olksdr: can we combine all of the empty Types together at the end? 

```suggestion
            Type::Infer(_) | Type::Macro(_) | Type::Never(_) => {}
```

tools/document-pii/src/item_collector.rs
TBS1996: but it implements `visit` with which it uses to visit all the AST trees, isn't that consdered iterating over?

tools/document-pii/src/pii_finder.rs
TBS1996: ok, I added some explanations there, not 100% sure if it's how its done.

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,325 @@
+//! Collecting full paths to items and AST-nodes of types
+//!
+//! This module will iterate over all the rust files given and collect all of the full path names
+//! and the actual AST-node for the types defined in those paths. This is later needed for finding
+//! PII fields recursively.
```
olksdr:
```suggestion
//! Contains the helper types and functions which help to iterate over all the given rust files, 
//! collect all of the full path names and the actual AST-node for the types defined in those paths. 
//! This is later needed for finding PII fields recursively.
```

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,325 @@
...
+        self.scoped_paths
+            .entry(self.module_path.clone())
+            .or_default()
+            .extend(use_statements);
+    }
+
+    /// Gets both a mapping of the full type to a type and its actual AST node, and also the
+    /// use_statements in its module, which is needed to fetch the types that it referes to in its
+    /// fields.
+    pub fn get_types_and_scoped_paths(paths: &[PathBuf]) -> anyhow::Result<TypesAndScopedPaths> {
```
olksdr:
```suggestion
    pub fn collect(paths: &[PathBuf]) -> anyhow::Result<TypesAndScopedPaths> {
```

Since the struct name and doc already describes what it collects, we do not have to put this in to the name of the function. Simpler names will work much better here.

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,325 @@
...
+
+impl AstItemCollector {
+    fn insert_scoped_paths(&mut self, use_statements: Vec<String>) {
+        self.scoped_paths
+            .entry(self.module_path.clone())
+            .or_default()
+            .extend(use_statements);
+    }
+
+    /// Gets both a mapping of the full type to a type and its actual AST node, and also the
```
olksdr:
```suggestion
    /// Gets both a mapping of the full path to a type and its actual AST node, and also the
```

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,325 @@
...
+        // Currently this script can't handle glob imports, which, we shouldn't use anyway.
+        UseTree::Glob(_) => vec![quote::quote!(#leading_path).to_string()],
+    }
+}
+
+fn crate_name_from_file(file_path: &Path) -> anyhow::Result<String> {
+    let file_str = file_path
+        .as_os_str()
+        .to_str()
+        .ok_or_else(|| anyhow!("Invalid file path: {}", file_path.display()))?;
```
olksdr: what if we use `to_string_lossy()` instead? Does it change much? 

```suggestion
        let file_str = file_path.as_os_str().to_string_lossy();
```

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,325 @@
...
+    let mut module_path = file_path
+        .parent()
+        .ok_or_else(|| {
+            anyhow!(
+                "Invalid file path (unable to find parent directory): {}",
+                file_path.display()
+            )
+        })?
+        .components()
+        .map(|part| part.as_os_str().to_string_lossy().replace('-', "_"))
```
olksdr: The `replace` called in multiple places on, what it looks like, same things. Maybe some the things, like the file path can be first resolved, then normalized once and you can avoid normalazing same things over and over again?

tools/document-pii/src/item_collector.rs
```diff
@@ -0,0 +1,325 @@
...
+// Checks if a file is a Rust module.
+fn is_file_module(file_path: &Path) -> anyhow::Result<bool> {
+    let parent_dir = file_path
+        .parent()
+        .ok_or_else(|| anyhow!("Invalid file path: {}", file_path.display()))?;
+    let file_stem = file_path
+        .file_stem()
+        .ok_or_else(|| anyhow!("Invalid file path: {}", file_path.display()))?
+        .to_string_lossy()
+        .into_owned();
```
olksdr:
```suggestion
        .to_string_lossy();
```

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    pub fn run(self) -> anyhow::Result<()> {
+        // User must either provide the path to a rust crate/workspace or be in one when calling this script.
+        let path = match self.path.clone() {
+            Some(path) => path,
+            None => std::env::current_dir()?,
+        };
+
+        if !path.join("Cargo.toml").exists() {
+            panic!("Please provide the path to a rust crate/workspace");
```
olksdr:
```suggestion
            anyhow::bail!("Please provide the path to a rust crate/workspace");
```

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+        match self.output {
+            Some(ref path) => self.write_pii(File::create(path)?, &output_vec)?,
+            None => self.write_pii(std::io::stdout(), &output_vec)?,
+        }
+
+        Ok(())
+    }
+}
+
+/// Represent the pii fields in a format that will be used in the final output.
```
olksdr: In some cases you use PII in some pii, let's use PII everywhere in the docs, please.  There are many places like this. 

```suggestion
/// Represents the PII fields in a format that will be used in the final output.
```

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,301 @@
...
+            all_types,
+            scoped_paths,
+            pii_values,
+            current_path: vec![],
+            pii_types: BTreeSet::new(),
+        })
+    }
+
+    fn visit_type_path(&mut self, path: &TypePath) {
+        let scoped_paths = self.scoped_paths.get(&self.module_path).unwrap().clone();
```
olksdr, line 93: can we remove `unwrap` here ? and return the error instead?

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,301 @@
...
+        get_field_types(&path.path, &mut field_types);
+
+        let use_paths = get_matching_scoped_paths(&field_types, &scoped_paths);
+        for use_path in use_paths {
+            if let Some(enum_or_struct) = self.all_types.get(use_path).cloned() {
+                // Theses values will be changed when recursing, so we save them here so when we
+                // return to this function after the match statement, we can set them back.
+                let current_type = self.current_type.clone();
+                let module_path = self.module_path.clone();
+                self.module_path = use_path.rsplit_once("::").unwrap().0.to_owned();
```
olksdr, line 105: same here

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,301 @@
...
+fn has_pii_value(pii_values: &[String], field: &Field) -> bool {
+    field.attrs.iter().any(|attribute| {
+        pii_values
+            .iter()
+            .any(|pii_value| has_attr_value(attribute, "metastructure", "pii", pii_value))
+    })
+}
+
+/// Finds all the pii fields recursively of a given type.
+pub fn find_pii_fields_of_type(
```
olksdr: it fees like this function could be on `TypesAndScopedPaths` struct from `item_collector` module.

tools/document-pii/src/pii_finder.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    match value {
+        EnumOrStruct::Struct(itemstruct) => visitor.visit_item_struct(itemstruct),
+        EnumOrStruct::Enum(itemenum) => visitor.visit_item_enum(itemenum),
+    };
+    Ok(visitor.pii_types)
+}
+
+/// Finds all the pii fields recursively of all the types in the rust crate/workspace.
+pub fn find_pii_fields_of_all_types(
```
olksdr: also this could be on `TypesAndScopedPaths`

olksdr (commented):
Left some comments to follow up on.

olksdr (commented):
Left some comments to follow up on.

tools/document-pii/src/item_collector.rs
TBS1996: that works well, thanks!

tools/document-pii/src/pii_finder.rs
TBS1996: I dont think so, because it's called from the visit trait which doesn't return anything 

`visit_field -> visit_field_types -> visit_type_path`

tools/document-pii/src/item_collector.rs
TBS1996: Rewrote it now so that it does this replace at the very end of the function instead

tests/test_pii_docs/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "test_pii_docs"
+version = "0.1.0"
+edition = "2021"
+
+
+[workspace]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
iker-barriocanal:
```suggestion
```

Do we also need the `workspace` attribute?

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,233 @@
...
+
+/// Prints documentation for metrics.
+#[derive(Debug, Parser, Default)]
+#[command(verbatim_doc_comment)]
+pub struct Cli {
+    /// Optional output path. By default, documentation is printed on stdout.
+    #[arg(short, long)]
+    pub output: Option<PathBuf>,
+
+    /// Path to the rust crate/workspace
```
iker-barriocanal:
```suggestion
    /// Path to the rust crate/workspace.
```

tools/document-pii/src/main.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    use crate::item_collector::{AstItemCollector, TypesAndScopedPaths};
+
+    use super::*;
+
+    const RUST_TEST_CRATE: &str = "../../tests/test_pii_docs";
+
+    fn get_types_and_use_statements() -> TypesAndScopedPaths {
+        let rust_crate = PathBuf::from_slash(RUST_TEST_CRATE);
+        let rust_file_paths = find_rs_files(&rust_crate);
```
iker-barriocanal, line 170: Since the test crate is only for testing purposes, the number and name of files won't change often. What do you think about hardcoding the list of files, to not rely on `find_rs_files`?

iker-barriocanal (commented):
I don't have strong opinions against not moving forward with his state, but would love to see what @olksdr thinks.

tests/test_pii_docs/Cargo.toml
TBS1996: It's pretty much a dummy file, the crate is never actually run, just for testing purposes. There just need to be a `Cargo.toml` file existing, and i created with normal `cargo new` command.

tools/document-pii/src/main.rs
TBS1996: I could do that, but isn't it better to verify that the `find_rs_files` works as expected, as is done now?

TBS1996: I made now a specific test to see that it got all the files

iker-barriocanal (approved):
Nice! I think this is good to be merged.

PR #1975, opened 2023-03-29
build(auth): Update cryptography dependencies

jan-auer:
This updates `ed25519-dalek`, `hmac`, `rand`, and `sha2` to their respective
latest versions. The Ed25519 crate is currently in final development for version
2. We're using this version as this allows us to run the latest versions of all
other crates.

This allows us to remove the `failure` dependency completely from our dependency
tree.

#skip-changelog

### DIFF
--- a/relay-auth/src/lib.rs
+++ b/relay-auth/src/lib.rs
@@ -208,7 +208,6 @@ impl SecretKey {
         let header_encoded = BASE64URL_NOPAD.encode(&header);
         header.push(b'\x00');
         header.extend_from_slice(data);
-        // TODO: Digest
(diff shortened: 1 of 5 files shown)


relay-auth/src/lib.rs
```diff
@@ -284,6 +288,16 @@ impl fmt::Debug for SecretKey {
 
 relay_common::impl_str_serde!(SecretKey, "a secret key");
 
+/// Represents the public key of an relay.
```
olksdr:
```suggestion
/// Represents the public key of a relay.
```

olksdr (approved):
Except `TODOs` about digest, this looks good to me.

PR #1976, opened 2023-03-29
feat(normalization): Don't mark as sanitized if there aren't any rules

iker-barriocanal:
If there aren't any rules, a transaction should not be promoted to `source:sanitized` because the clusterer may not have generated enough rules yet. The exception is if we've scrubbed UUIDs, in which case we still consider the transaction as sanitized.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,22 +7,26 @@
 This release contains major changes to the web layer, including TCP and HTTP handling as well as all web endpoint handlers. Due to these changes, some functionality was retired and Relay responds differently in specific cases.
 
 Configuration:
+
 - SSL support has been dropped. As per [official guidelines](https://docs.sentry.io/product/relay/operating-guidelines/), Relay should be operated behind a reverse proxy, which can perform SSL termination.
 - Connection config options `max_connections`, `max_pending_connections`, and `max_connection_rate` no longer have an effect. Instead, configure the reverse proxy to handle connection concurrency as needed.
 
 Endpoints:
+
 - The security endpoint no longer forwards to upstream if the mime type doesn't match supported mime types. Instead, the request is rejected with a corresponding error.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,22 +7,26 @@
 This release contains major changes to the web layer, including TCP and HTTP handling as well as all web endpoint handlers. Due to these changes, some functionality was retired and Relay responds differently in specific cases.
 
 Configuration:
+
```
iker-barriocanal, line 10: These blank lines have been added by the formatter, so I thought about leaving them in place to avoid saving without formatting every time there's a change in the changelog.

relay-general/src/store/transactions/processor.rs
```diff
@@ -350,19 +350,29 @@ impl Processor for TransactionsProcessor<'_> {
         if self.name_config.scrub_identifiers {
             // Normalize transaction names for URLs and Sanitized transaction sources.
             // This in addition to renaming rules can catch some high cardinality parts.
+            let mut sanitize_transaction = false;
```
```suggestion
            let mut sanitized = false;
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -350,19 +350,29 @@ impl Processor for TransactionsProcessor<'_> {
...
+
             if matches!(
                 event.get_transaction_source(),
                 &TransactionSource::Url | &TransactionSource::Sanitized
             ) {
-                scrub_identifiers(&mut event.transaction)?;
+                let scrubbed_identifiers = scrub_identifiers(&mut event.transaction);
+                if scrubbed_identifiers.map_or(false, |scrubbed| scrubbed) {
```
jjbayer: AFAIK we still want to early return the error here if `scrub_identifiers` returns one.

```suggestion
                let scrubbed_identifiers = scrub_identifiers(&mut event.transaction)?;
                if scrubbed_identifiers {
```

jjbayer (approved):

PR #1978, opened 2023-03-29
instr(txnames): Log statsd metrics for transaction name changes

jjbayer:
Log changes made to the transaction name by the `TransactionProcessor` to get an idea of what percentage of transactions is affected, and by which mechanism.

We can consider this a temporary metric and delete it once we are done with this initiative.

#skip-changelog


relay-server/src/statsd.rs
```diff
@@ -479,6 +479,16 @@ pub enum RelayCounters {
...
+    /// The number of transaction events processed grouped by transaction name modifications.
+    /// This metric is tagged with:
+    ///  - `source_in`: The source of the transaction name before normalization.
+    ///    See the [transaction source
+    ///    documentation](https://develop.sentry.dev/sdk/event-payloads/properties/transaction_info/)
+    ///    for all valid values.
+    ///  - `change`: The mechanism that changed the transaction name.
+    ///    Either `"none"`, `"pattern"`, `"rule"`, or `"both"`.
+    ///  - `source_out`: The source of the transaction name after normalization.
+    TransactionNameChanges,
```
jjbayer, line 470: I considered adding a `set` metric as well to get the % of _unique_ names scrubbed, but I did not want to risk overwhelming statsd with that.

relay-server/src/utils/statsd.rs
```diff
@@ -0,0 +1,63 @@
...
+use relay_general::protocol::Event;
+use relay_general::types::{Annotated, RemarkType};
+
+use crate::statsd::RelayCounters;
+
+/// Log statsd metrics about transaction name modifications.
+///
+/// We have to look at event & meta before and after the modification is made,
+/// so we delegate to `f` in the middle of the function.
+pub fn log_transaction_name_metrics<F, R>(event: &mut Annotated<Event>, mut f: F) -> R
```
jjbayer, line 11: This function turned out rather convoluted. We cannot log metrics from `relay-general`, so we have to analyze the event and its meta from `relay-server` to determine what changes were made.

relay-server/src/utils/statsd.rs
```diff
@@ -0,0 +1,63 @@
...
+where
+    F: FnMut(&mut Annotated<Event>) -> R,
+{
+    let Some(inner) = event.value() else {
+            return f(event);
+        };
```
olksdr: Formatting a little bit off.

olksdr (approved):

relay-server/src/utils/statsd.rs
jjbayer: Thanks, interesting that the auto formatter did not catch it.

jjbayer:
```suggestion
    let Some(inner) = event.value() else {
        return f(event);
    };
```

PR #1979, opened 2023-03-29
feat(thread-protocol): Extract lock mechanism

### LINKED ISSUE #46510: Extract lock reason in the threads interface in relay

shruthilayaj:
The android sdk is going to start sending lock mechanism
as a part of the thread interface which we are going
to want to extract and store on events json.

image (2)

closes https://github.com/getsentry/sentry/issues/46510


relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +69,21 @@ impl Empty for ThreadId {
...
+    #[metastructure(field = "type", required = "true")]
+    pub ty: Annotated<u64>,
+    #[metastructure(skip_serialization = "empty")]
+    pub address: Annotated<String>,
+    #[metastructure(skip_serialization = "empty")]
+    pub package_name: Annotated<String>,
+    #[metastructure(skip_serialization = "empty")]
+    pub class_name: Annotated<String>,
+    #[metastructure(skip_serialization = "empty")]
+    pub thread_id: Annotated<ThreadId>,
```
jjbayer, line 170: This allows outdated external relays to forward new fields if we ever add them:

```suggestion
    pub thread_id: Annotated<ThreadId>,
    
    /// Additional arbitrary fields for forwards compatibility.
    #[metastructure(additional_properties)]
    pub other: Object<Value>,
```

romtsn (approved):
lgtm, thanks!

relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +69,25 @@ impl Empty for ThreadId {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct LockReason {
+    #[metastructure(field = "type", required = "true")]
+    pub ty: Annotated<u64>,
```
jjbayer: Do we accept any integer for `type`? If not, we could make it an enum.

jjbayer (approved):

relay-general/src/protocol/thread.rs
shruthilayaj: good question, @romtsn do you happen to know if it's a small set of integers here?

romtsn: this looks like the entire set of values, so definitely an enum would suit here
image

shruthilayaj: oh great, will update 👍

shruthilayaj: @jjbayer I made lock reason an enum most relevant changes in this commit https://github.com/getsentry/relay/pull/1979/commits/cbc72bde891cb4330200407bcc6445101763c9e0

relay-general/src/protocol/thread.rs
```diff
@@ -69,11 +71,101 @@ impl Empty for ThreadId {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "jsonschema", schemars(rename_all = "lowercase"))]
+
```
jjbayer:
```suggestion
```

relay-general/src/protocol/thread.rs
```diff
@@ -69,11 +71,101 @@ impl Empty for ThreadId {
...
+    fn serialize_payload<S>(&self, s: S, _behavior: SkipSerialization) -> Result<S::Ok, S::Error>
+    where
+        Self: Sized,
+        S: Serializer,
+    {
+        Serialize::serialize(&self.to_string(), s)
+    }
+}
+
+impl ProcessValue for LockReasonType {}
```
jjbayer: You could get around this boilerplate by using a fallback variant, like in the code linked below. This would also make it possible for external Relays to forward new variants again.

I realize that this means that effectively, any integer is accepted again, but at least it documents the expected variants.

https://github.com/getsentry/relay/blob/92f9ac91d306310d8b3f912cdad97d686d147583/relay-general/src/protocol/debugmeta.rs#L489-L491

relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +71,115 @@ impl Empty for ThreadId {
...
+            Annotated(Some(Value::I64(val)), mut meta) => {
+                match LockReasonType::from_android_lock_reason_type(val as u64) {
+                    Some(value) => Annotated(Some(value), meta),
+                    None => {
+                        meta.add_error(ErrorKind::InvalidData);
+                        meta.set_original_value(Some(val));
+                        Annotated(None, meta)
+                    }
+                }
+            }
```
iker-barriocanal, line 123: Are these branches the same? If yes, could we merge them?

relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +71,115 @@ impl Empty for ThreadId {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "jsonschema", schemars(rename_all = "lowercase"))]
+
```
iker-barriocanal:
```suggestion
```

relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +71,114 @@ impl Empty for ThreadId {
...
+        Self: Sized,
+        S: Serializer,
+    {
+        Serialize::serialize(&self.to_string(), s)
+    }
+}
+
+impl ProcessValue for LockReasonType {}
+
+#[derive(Clone, Debug, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
```
jjbayer, line 149: Please add a `///` doc comment above all public types and variants.

relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +71,114 @@ impl Empty for ThreadId {
...
 }
 
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "jsonschema", schemars(rename_all = "lowercase"))]
+pub enum LockReasonType {
+    Locked = 1,
+    Waiting = 2,
+    Sleeping = 4,
+    Blocked = 8,
```
jjbayer, line 84:
```suggestion
    Blocked = 8,
    // This enum does not have a `fallback_variant` because we consider it unlikely to be extended. If it is,
    // The error added to `Meta` will tell us to update this enum.
```

relay-general/src/protocol/thread.rs
```diff
@@ -128,6 +238,10 @@ pub struct Thread {
     #[metastructure(skip_serialization = "empty")]
     pub state: Annotated<String>,
 
+    /// Lock object that is blocking the thread
```
jjbayer:
```suggestion
    /// Lock object that is blocking the thread.
```

jjbayer (approved):
Please take a look at the comments below before merging.

relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +71,116 @@ impl Empty for ThreadId {
...
+impl fmt::Display for LockReasonType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
+            LockReasonType::Locked => write!(f, "locked"),
+            LockReasonType::Waiting => write!(f, "waiting"),
+            LockReasonType::Sleeping => write!(f, "sleeping"),
+            LockReasonType::Blocked => write!(f, "blocked"),
+        }
+    }
+}
```
jjbayer:
```suggestion
```

relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +71,116 @@ impl Empty for ThreadId {
...
+                meta.set_original_value(Some(value));
+                Annotated(None, meta)
+            }
+        }
+    }
+}
+
+impl IntoValue for LockReasonType {
+    fn into_value(self) -> Value {
+        Value::String(self.to_string())
```
jjbayer: I missed this on my previous review: An upstream Relay has to be able to parse this again after re-serialization, so we have to serialize as an integer (see related suggestions below).

```suggestion
        Value::U64(self as u64)
```

relay-general/src/protocol/thread.rs
```diff
@@ -69,6 +71,116 @@ impl Empty for ThreadId {
...
+    }
+
+    fn serialize_payload<S>(&self, s: S, _behavior: SkipSerialization) -> Result<S::Ok, S::Error>
+    where
+        Self: Sized,
+        S: Serializer,
+    {
+        Serialize::serialize(&self.to_string(), s)
```
jjbayer:
```suggestion
        Serialize::serialize(&(*self as u64), s)
```

relay-general/src/protocol/thread.rs
```diff
@@ -205,4 +322,69 @@ mod tests {
...
+  "state": "BLOCKED",
+  "lock_reason": {
+    "type": "locked",
+    "package_name": "android.database.sqlite",
+    "class_name": "SQLiteConnection",
+    "thread_id": 2
+  },
+  "other": "value"
+}"#;
+        assert_eq!(output, thread.to_json_pretty().unwrap());
```
jjbayer:
```suggestion
        assert_eq!(input, thread.to_json_pretty().unwrap());
```

relay-general/src/protocol/thread.rs
jjbayer: Please also add a doc comment to `LockReason` and `LockReasonType` type itself.

jjbayer (changes requested):

relay-general/src/protocol/thread.rs
shruthilayaj: Ah I see 👍

relay-general/src/protocol/thread.rs
shruthilayaj: ah yup, will do

shruthilayaj: i'm going get roman to help me with docstrings for each of the LockReasonType enums (i'm not super familiar) but will get those in before merging! you can still review the other changes

jjbayer (approved):
Thanks!

PR #1981, opened 2023-03-29
feat: Use sentry-kafka-schemas types for metrics

untitaker:
Depends on https://github.com/getsentry/sentry-kafka-schemas/pull/74 landing.

- [x] remove git dep on PR branch
- [x] resolve open TODOs (use of unwrap)


relay-server/src/actors/store.rs
```diff
@@ -563,16 +563,50 @@ impl StoreService {
...
                     name: bucket.name,
-                    value: bucket.value,
-                    timestamp: bucket.timestamp,
-                    tags: bucket.tags,
-                    retention_days: retention,
+                    type_: flat_type,
+                    value: flat_value,
+                    timestamp: bucket.timestamp.as_secs().try_into().unwrap(),
+                    // TODO: slow (conversion from btreemap to hashmap)
+                    tags: bucket.tags.into_iter().collect(),
```
jjbayer: Any way around this? Can you configure the code generator to choose `BTreeMap` instead?

relay-server/src/actors/store.rs
```diff
@@ -563,16 +563,50 @@ impl StoreService {
         let payload = item.payload();
 
         for bucket in Bucket::parse_all(&payload).unwrap_or_default() {
+            let (flat_type, flat_value) = match bucket.value {
+                BucketValue::Distribution(value) => (
+                    ingest_metrics_v1::IngestMetricType::D,
```
jjbayer: Is there a way to rename this variant to `Distribution` (while still naming it `d` in json)?

relay-server/src/actors/store.rs
```diff
@@ -563,16 +563,50 @@ impl StoreService {
...
+                        |scope| {
+                            scope.set_extra("metric_message.name", bucket.name.into());
+                        },
+                        || {
+                            relay_log::error!("attempted to produce Gauge");
+                        },
+                    );
+                    continue;
+                }
+            };
```
jjbayer: nit: How about we put all this conversion logic in a `TryFrom<Bucket> for IngestMetric` impl?

relay-server/src/actors/store.rs
untitaker: i believe it would require us to write custom rust types for the bucketvalue and then patch them into the generated code using https://docs.rs/typify/latest/typify/struct.TypeSpaceSettings.html#method.with_replacement -- i don't think that's worth it

relay-server/src/actors/store.rs
untitaker: I ended up string-replacing HashMap with BTreeMap in the generated code: https://github.com/getsentry/sentry-kafka-schemas/pull/78

relay-server/src/actors/store.rs
untitaker: i put it into a helper function. the conversion function should be infallible except for the gauge case

untitaker:
blocked on https://github.com/getsentry/sentry-kafka-schemas/pull/78

untitaker:
this is actually ready for review. i will update the types in 24h. the tests pass locally with the pr branch patched in

relay-general/tests/snapshots/test_fixtures__event_schema.snap
```diff
@@ -1427,15 +1427,56 @@ expression: "relay_general::protocol::event_json_schema()"
...
-      "type": "string",
-      "enum": [
-        "error",
-        "csp",
-        "hpkp",
-        "expectct",
-        "expectstaple",
-        "transaction",
-        "default"
+      "oneOf": [
```
untitaker, line 1430: cargo bumped schemars, and as a result schemars apparently decided to make this change to the event schema. i think the change is harmless, but we can probably also just pin the direct schemars dep to a different version

jjbayer: Looks like it's deliberate, to maintain the doc comments on enum variants 👍

relay-server/src/actors/store.rs
```diff
@@ -1008,6 +1001,53 @@ struct UserReportKafkaMessage {
...
+            ingest_metrics_v1::IngestMetricValue::NumbersMetricValue(value.iter_values().collect()),
+        ),
+        BucketValue::Counter(value) => (
+            ingest_metrics_v1::IngestMetricType::C,
+            ingest_metrics_v1::IngestMetricValue::NumberMetricValue(value),
+        ),
+        BucketValue::Set(values) => (
+            ingest_metrics_v1::IngestMetricType::S,
+            ingest_metrics_v1::IngestMetricValue::NumbersMetricValue(
+                values.into_iter().map(f64::from).collect(),
```
jjbayer: It makes me a bit uneasy to convert a hash to a floating point number here, but it works because `f64` can represent any `u32`.

jjbayer (approved):

relay-server/src/actors/store.rs
untitaker: this is definitely an oversight, to be fixed in https://github.com/getsentry/sentry-kafka-schemas/pull/83

PR #1982, opened 2023-03-29
feat(profiling): Copy transaction tags to the profile

phacops:
As we slim down profiles to just profiles data, we need to extract tags from transactions and pass them over to use in our aggregations.

This PR will copy values we need from the event and store them as `tags` on the profile or in the proper fields if they are present on the profile already.


relay-profiling/src/lib.rs
```diff
@@ -126,15 +127,18 @@ fn minimal_profile_from_json(data: &[u8]) -> Result<MinimalProfile, ProfileError
...
         Ok(profile) => profile,
         Err(err) => return Err(err),
     };
     let processed_payload = match profile.version {
-        Version::V1 => parse_sample_profile(payload),
+        Version::V1 => parse_sample_profile(payload, tags),
         Version::Unknown => match profile.platform.as_str() {
-            "android" => parse_android_profile(payload),
+            "android" => parse_android_profile(payload, tags),
             "cocoa" => parse_cocoa_profile(payload),
```
iker-barriocanal: Not for cocoa?

iker-barriocanal (approved):

relay-profiling/src/lib.rs
phacops: This is an old format from an old SDK and it's not sending the profile and transaction in the same envelope and extracting tags requires the transaction and profile in the same envelope.

So, not useful to add it since we'll never have them together.

relay-profiling/src/android.rs
```diff
@@ -147,7 +151,18 @@ fn parse_profile(payload: &[u8]) -> Result<AndroidProfile, ProfileError> {
...
-    } else if !profile.has_transaction_metadata() {
+    } else if profile.has_transaction_metadata() {
+        profile.transaction = Some(TransactionMetadata {
+            active_thread_id: profile.active_thread_id,
+            id: profile.transaction_id,
+            name: profile.transaction_name.clone(),
+            relative_cpu_end_ms: 0,
+            relative_cpu_start_ms: 0,
+            relative_end_ns: 0,
+            relative_start_ns: 0,
```
iker-barriocanal: If you derive `Default` in `TransactionMetadata`, you can omit explicitly setting these by `..Default::default()`.

relay-profiling/src/android.rs
```diff
@@ -160,8 +178,30 @@ fn parse_profile(payload: &[u8]) -> Result<AndroidProfile, ProfileError> {
     Ok(profile)
 }
 
-pub fn parse_android_profile(payload: &[u8]) -> Result<Vec<u8>, ProfileError> {
-    let profile = parse_profile(payload)?;
+pub fn parse_android_profile(
```
iker-barriocanal, line 181: This code looks identical to `parse_sample_profile`, is this intentional?

phacops: It's not identical. `profile` is a different type and we're not setting the `transaction_name` on the `SampleProfile` in `parse_sample_profile`.

I'd be willing to refactor it but I don't see how we can beat the simplicity of this vs making a common function to copy tags on both structs.

iker-barriocanal: Oh sorry, I missed that. I agree this is simpler.

relay-server/src/actors/processor.rs
```diff
@@ -1080,52 +1083,78 @@ impl EnvelopeProcessorService {
...
+                                version: 1,
+                                extract_custom_tags: BTreeSet::from(["device.class".to_string()]),
+                                ..Default::default()
+                            },
+                        );
+
+                        // Overwrite the transaction name to keep the original one
+                        tags.insert(
+                            "transaction".to_owned(),
+                            event.get_transaction_source().to_string(),
```
jjbayer: Don't you want `event.transaction.value()` here?

iker-barriocanal (approved):

relay-server/src/metrics_extraction/transactions.rs
```diff
@@ -161,7 +161,7 @@ fn get_transaction_name(
 }
 
 /// These are the tags that are added to all extracted metrics.
-fn extract_universal_tags(
+pub fn extract_universal_tags(
```
jjbayer: Exposing this function means that changes to what tags are extracted for metrics by default will also affect profiles. This might be exactly what you want (if the `group by` for profiles should always be the same as for metrics), but it might also be an unwanted side effect, depending on the use case.

relay-server/src/actors/processor.rs
```diff
@@ -1080,52 +1083,78 @@ impl EnvelopeProcessorService {
...
+                        );
```
Zylphrex: Nice 👍

Zylphrex (approved):

relay-server/src/metrics_extraction/transactions.rs
phacops: Right. We were OK with that since getting the normalized transaction name wasn't trivial. Now that we realized we need the original transaction name, reaching to this function to extract tags is unnecessary, we can write our own and not have the dependency.

relay-server/src/actors/processor.rs
phacops: Indeed.

JonasBa (approved):

relay-profiling/src/android.rs
```diff
@@ -160,8 +175,30 @@ fn parse_profile(payload: &[u8]) -> Result<AndroidProfile, ProfileError> {
...
+
+    if let Some(release) = tags.get("release") {
+        profile.release = release.to_owned();
+    }
+
+    if let Some(environment) = tags.get("environment") {
+        profile.environment = environment.to_owned();
+    }
+
+    profile.tags = tags;
```
jjbayer: Out of curiosity: Why do we store release, transaction name and environment in _both_ tags _and_ dedicated fields? I.e., why not remove them from `tags`?

relay-profiling/src/android.rs
```diff
@@ -250,4 +287,31 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_copy_tags() {
+        let mut tags: BTreeMap<String, String> = BTreeMap::new();
+        tags.insert("release".to_string(), "some-random-release".to_string());
+        tags.insert(
+            "transaction".to_string(),
+            "some-random-transaction".to_string(),
+        );
```
jjbayer: nit

```suggestion
        let tags = BTreeMap::from([
            ("release".to_string(), "some-random-release".to_string()),
            ("transaction".to_string(), "some-random-transaction".to_string()),
        ]);
```

relay-profiling/src/sample.rs
```diff
@@ -315,8 +318,28 @@ fn parse_profile(payload: &[u8]) -> Result<SampleProfile, ProfileError> {
...
-pub fn parse_sample_profile(payload: &[u8]) -> Result<Vec<u8>, ProfileError> {
-    let profile = parse_profile(payload)?;
+pub fn parse_sample_profile(
+    payload: &[u8],
+    tags: BTreeMap<String, String>,
+) -> Result<Vec<u8>, ProfileError> {
+    let mut profile = parse_profile(payload)?;
+
+    if let Some(transaction_name) = tags.get("transaction") {
+        if let Some(ref mut transaction) = profile.transaction {
```
jjbayer, line 332: `parse_android_profile` also sets `profile.transaction_name` here, is that difference intentional?

Either way it seems like these two functions could share some code, maybe unite them and pass a boolean argument for setting `profile.transaction_name`?

jjbayer (approved):
Left a few questions, but none of them are blocking.

relay-profiling/src/android.rs
phacops: No real reason, it's mostly because it started as collecting tags and evolved into something else. I'll remove them.

relay-profiling/src/sample.rs
phacops: `parse_profile` called in `parse_sample_profile` and `parse_android_profile` are not returning the same type (`SampleProfile` or `AndroidProfile`).

I can't make a function which would accept both types and let me set the fields directly, can I?

jjbayer: You are right, you would need a trait for that. Or refactor the `*Profile` structs to contain a common `BaseProfile` with [`#[serde(flatten)]`](https://serde.rs/attr-flatten.html#factor-out-frequently-grouped-keys). But that would be out of scope for this PR.

Zylphrex (approved):
LGTM

PR #1983, opened 2023-03-30
feat(replays): add replay_id onto event from dynamic sampling context

### LINKED ISSUE #1958: Pluck `replay_id` out of dynamic sampling context and add it to context if it exists
As part of Connecting Backend Errors With Replays, https://github.com/getsentry/replay-backend/issues/251, we decided that we will add replay_id to the dynamic sampling context (https://github.com/getsentry/sentry-javascript/issues/7555).

In Relay, we should pluck the replay_id out of the dynamic sampling context and add it to the context of the event. This will then be written as a column in the various clickhouse tables. I will plan on making a PR in one of our next sprints.

Original RFC https://github.com/getsentry/rfcs/pull/60

JoshFerge:
- [x] Defines a new context `ReplayContext` that will contain the replay_id on events (errors and transactions, etc.). Note that the SDKs never add this value to the event, it is always added in relay via the DSC.
- [x] Adds replay_id to dynamic sampling context (DSC) schema 
- [x] If replay_id exists on DSC, add it onto the event being processed in the replay context

Related: https://github.com/getsentry/rfcs/pull/60 https://github.com/getsentry/team-webplatform-meta/issues/41
https://github.com/getsentry/sentry/issues/45529

Closes: https://github.com/getsentry/relay/issues/1958


relay-server/src/actors/processor.rs
```diff
@@ -1756,9 +1757,19 @@ impl EnvelopeProcessorService {
         // event id. To be defensive, we always overwrite to ensure consistency.
         event.id = Annotated::new(event_id);
 
-        // In processing mode, also write metrics into the event. Most metrics have already been
-        // collected at this state, except for the combined size of all attachments.
         if self.config.processing_enabled() {
+            // add the replay_id to the event's context if it exists on the DSC
+            if let Some(dsc) = envelope.dsc() {
```
JoshFerge, line 1762: i wasn't sure where to stick this bit of logic but added here for now in `finalize_event`

relay-server/src/actors/processor.rs
```diff
@@ -2262,6 +2279,7 @@ impl EnvelopeProcessorService {
             if_processing!({
                 self.process_unreal(state)?;
                 self.create_placeholders(state);
+                self.enrich_event_with_dsc(state);
```
JoshFerge: wasn't sure where to put this logic, lmk if approach makes sense or not.

JoshFerge:
need to look into the unit test failures, and address the replay context object usage in the replay event payload.

relay-general/src/protocol/contexts/replay.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::types::Annotated;
+
+/// Replay context
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct ReplayContext {
+    /// The replay ID.
+    #[metastructure(required = "true")]
+    pub replay_id: Annotated<EventId>,
+}
```
olksdr, line 19: Please, also add the `other` for forward compatibility.  Like it's done in all other contexts. 


```rust    
    #[metastructure(additional_properties, retain = "true", pii = "maybe")]
    pub other: Object<Value>,
```

relay-sampling/src/lib.rs
```diff
@@ -1164,6 +1164,8 @@ pub struct DynamicSamplingContext {
     /// user object).
     #[serde(flatten, default)]
     pub user: TraceUserContext,
+    /// if the event occured during a session replay, the associated replay_id is added to the DSC
```
```suggestion
    /// If the event occurred during a session replay, the associated replay_id is added to the DSC.
```

olksdr (changes requested):

relay-general/src/protocol/contexts/replay.rs
JoshFerge: sounds good. FWIW i was copying the implementation of profiles context which does not have other https://github.com/getsentry/relay/blob/b62b0f2947afaccbfb79fb70015fe2ab634bbc22/relay-general/src/protocol/contexts/profile.rs#L7

JoshFerge: but this is actually helpful in our case -- thanks for the suggestion.

relay-server/src/actors/processor.rs
```diff
@@ -2167,6 +2167,23 @@ impl EnvelopeProcessorService {
         Ok(())
     }
 
+    #[cfg(feature = "processing")]
+    /// Adds data from the DSC onto the event, only replay_id for now.
+    fn enrich_event_with_dsc(&self, state: &mut ProcessEnvelopeState) {
+        let envelope = state.managed_envelope.envelope_mut();
```
```suggestion
        let envelope = state.managed_envelope.envelope();
```

relay-server/src/actors/processor.rs
```diff
@@ -2167,6 +2167,23 @@ impl EnvelopeProcessorService {
...
+
+        if let Some(event) = state.event.value_mut() {
+            if let Some(dsc) = envelope.dsc() {
+                if let Some(replay_id) = dsc.replay_id {
+                    let contexts = event.contexts.get_or_insert_with(Contexts::new);
+                    contexts.add(SentryContext::Replay(Box::new(ReplayContext {
```
jjbayer: This overwrites a replay context if it was already on the event. Is that fine?

relay-general/src/protocol/contexts/replay.rs
```diff
@@ -0,0 +1,55 @@
+use crate::protocol::EventId;
+use crate::types::{Annotated, Object, Value};
+
+/// Replay context
```
jjbayer: Some more information would be nice here, see for example 

https://github.com/getsentry/relay/blob/8afa67264a69d5d9105264bf94254aa0bfb10dd8/relay-general/src/protocol/contexts/os.rs#L6-L9

relay-general/src/protocol/contexts/replay.rs
```diff
@@ -0,0 +1,55 @@
...
+pub struct ReplayContext {
+    /// The replay ID.
+    pub replay_id: Annotated<EventId>,
+    /// Additional arbitrary fields for forwards compatibility.
+    #[metastructure(additional_properties, retain = "true")]
+    pub other: Object<Value>,
+}
+
+impl ReplayContext {
+    /// The key under which a replay context is generally stored (in `Contexts`)
```
jjbayer:
```suggestion
    /// The key under which a replay context is generally stored (in `Contexts`).
```

relay-server/src/actors/processor.rs
jjbayer: I would rather put it in store normalization, like we do for the trace context:

https://github.com/getsentry/relay/blob/8afa67264a69d5d9105264bf94254aa0bfb10dd8/relay-general/src/store/normalize.rs#L195-L201

which pulls information from the DSC here:

https://github.com/getsentry/relay/blob/8afa67264a69d5d9105264bf94254aa0bfb10dd8/relay-server/src/actors/processor.rs#L1892

jjbayer (changes requested):
In general this looks good to me. I would just move the logic to store normalization.

relay-server/src/actors/processor.rs
JoshFerge: yup, the client should not be setting the replay context on events.

relay-general/src/protocol/contexts/replay.rs
bruno-garcia: Can we get clippy to suggest ending comments with ` .`? Or just add the `.` ?

relay-server/src/actors/processor.rs
JoshFerge: makes sense!

relay-general/src/store/.normalize.rs.pending-snap
```diff
@@ -0,0 +1,16 @@
...
+{"run_id":"1680636747-15499000","line":2862,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2415,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2485,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2148,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2038,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2734,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2085,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2611,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2630,"new":null,"old":null}
+{"run_id":"1680636747-15499000","line":2207,"new":null,"old":null}
```
jjbayer: Please remove this file.

jjbayer (changes requested):
The move to normalization looks good, but that means that `enrich_event_with_dsc` can now be removed, right? After that, this should be good to merge.

Please also remove the pending snapshot file, and run `cargo insta review` to fix the [test failures](https://github.com/getsentry/relay/actions/runs/4612548879/jobs/8153575530?pr=1983#step:5:873).

JoshFerge:
> The move to normalization looks good, but that means that `enrich_event_with_dsc` can now be removed, right? After that, this should be good to merge.
> 
> Please also remove the pending snapshot file, and run `cargo insta review` to fix the [test failures](https://github.com/getsentry/relay/actions/runs/4612548879/jobs/8153575530?pr=1983#step:5:873).

okay should be good! thanks for the review.

jjbayer (approved):

relay-general/src/protocol/contexts/replay.rs
```diff
@@ -0,0 +1,61 @@
+use crate::protocol::EventId;
+use crate::types::{Annotated, Object, Value};
+
+/// Replay context
```
olksdr:
```suggestion
/// Replay context.
```

olksdr (approved):
:shipit:

PR #1984, opened 2023-03-30
feat(protocol): Add origin to trace context and span

philipphofmann:
Adds origin to both trace context and the span. The origin of the span indicates what created the span.

Related RFC https://github.com/getsentry/rfcs/pull/73 and develop docs PR https://github.com/getsentry/develop/pull/887


relay-general/src/protocol/contexts/trace.rs
```diff
@@ -102,6 +102,9 @@ pub struct TraceContext {
     /// should not ever send this value.
     pub client_sample_rate: Annotated<f64>,
 
+    /// The origin of the trace indicates what created the trace.
+    pub origin: Annotated<String>,
```
olksdr: Do we have a list of predefined origins, which we could convert to `enum` to get more type safety? Or this is just some always random string, which will come from the user?

iker-barriocanal: This. If the latter, what character set should we allow for this field? How long should this field be?

philipphofmann: We don't have a predefined list, but a pattern to follow https://develop.sentry.dev/sdk/performance/trace-origin/. We could enforce alphanumeric characters: a-z , A-Z , and 0-9 and underscores: _ as pointed out in the develop docs.

philipphofmann: I added a `OriginType` similar to `OperationType` https://github.com/getsentry/relay/blob/1d57db414a11af96348747c307d40a7852018142/relay-general/src/protocol/contexts/mod.rs#L31-L33

relay-general/src/protocol/contexts/trace.rs
```diff
@@ -102,6 +102,10 @@ pub struct TraceContext {
     /// should not ever send this value.
     pub client_sample_rate: Annotated<f64>,
 
+    /// The origin of the trace indicates what created the trace (see `OriginType` docs).
```
iker-barriocanal: nit:
```suggestion
    /// The origin of the trace indicates what created the trace (see [`OriginType`] docs).
```

relay-general/src/protocol/span.rs
```diff
@@ -47,6 +49,10 @@ pub struct Span {
     #[metastructure(pii = "true")]
     pub data: Annotated<Object<Value>>,
 
+    /// The origin of the span indicates what created the span (see `OriginType` docs).
```
iker-barriocanal: nit:
```suggestion
    /// The origin of the span indicates what created the span (see [`OriginType`] docs).
```

relay-general/src/protocol/span.rs
```diff
@@ -47,6 +49,10 @@ pub struct Span {
...
+    #[metastructure(max_chars = "enumlike")]
+    pub origin: Annotated<OriginType>,
```
iker-barriocanal: Another nit: if we could add this before `data` prop, we would live the "grouping" properties at the end of the structure.

iker-barriocanal (changes requested):
> We could enforce alphanumeric characters: a-z , A-Z , and 0-9 and underscores: _ as pointed out in the develop docs.

Let's do this. It can easily be done by adding `allow_chars` to the metastructure of the property, see [example](https://github.com/getsentry/relay/blob/3673a76f8efcb9827ce4e7d1fec60a36aab6c5ba/relay-general/src/protocol/tags.rs#L7). Blocking the PR for this.

relay-general/src/protocol/contexts/trace.rs
```diff
@@ -102,6 +102,10 @@ pub struct TraceContext {
     /// should not ever send this value.
     pub client_sample_rate: Annotated<f64>,
 
+    /// The origin of the trace indicates what created the trace (see [OriginType] docs).
+    #[metastructure(max_chars = "enumlike", allow_chars = "a-zA-Z0-9_.")]
+    pub origin: Annotated<OriginType>,
```
philipphofmann, line 107: @iker-barriocanal, can I somehow add the metastructure to the definition of `OriginType` so we reuse it? I'm new to Rust and Relay and have no clue how to achieve this.

relay-general/src/protocol/mod.rs
```diff
@@ -66,3 +66,7 @@ pub use self::transaction::*;
 pub use self::types::*;
 pub use self::user::*;
 pub use self::user_report::*;
+
+/// Origin type such as `auto.http`.
+/// Follows the pattern described in the [develop docs](https://develop.sentry.dev/sdk/performance/trace-origin/).
+pub type OriginType = String;
```
iker-barriocanal: Sorry I missed this earlier, but let's move this to `relay-general/protocol/contexts/mod.rs`, like `OperationType`.

relay-general/src/protocol/contexts/trace.rs
iker-barriocanal: No worries, it's a good question!

I don't think you want to do that. The reason is that what you're doing with the `metastructure` is not adding it to the type `OriginType`, but the fields `TraceContext.origin` and `Span.origin`. `OriginType` per se is only the type of these two fields, which happen to be the same and have the same restrictions in `TraceContext` and `Span`. To have the metastructure in `OriginType`, I guess you'd have to create a `struct` (not `Type`), and I'm not sure if a specific processing function too (see `process_func` in `TraceContext`).

I'd leave it like it is -- we don't need a `struct` for now and it's simpler to follow this way (I can look at `TraceContext` and see its requirements, instead of navigating to the definition of `OriginType` and also look there).

iker-barriocanal (approved):

PR #1991, opened 2023-03-31
chore(gocd): Auto approve and run deploy if the check stage succeeds

olksdr:
Currently every stage must be approved manually, but it makes more sense to automate this.

This change makes sure that if the `check` stage succeeds the `deploy` stage will start automatically, and if it fails it should not be possible to trigger it even manually.

 
#skip-changelog

### DIFF
--- a/gocd/pipelines/relay.yaml
+++ b/gocd/pipelines/relay.yaml
@@ -47,7 +47,7 @@ pipelines:
 
       - deploy-production:
           approval:
-            type: manual
+            type: success
             allow_only_on_success: true
           fetch_materials: true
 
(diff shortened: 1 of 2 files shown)


gocd/pipelines/relay.yaml
```diff
@@ -82,7 +82,7 @@ pipelines:
           # Manually completes this pipeline run.
           # This will trigger relay-pop deployments.
           approval:
-            type: manual
+            type: success
```
jjbayer, line 85: Does this mean PoPs will automatically deploy if the Relay deploy was successful?

jan-auer (approved):

gocd/pipelines/relay.yaml
olksdr: Nope, since deploy for PoPs will have to be approved manually again, see 
https://github.com/getsentry/relay/blob/a4290b4d53bce5781d80b1b3ff1be88b234b8240/gocd/pipelines/relay-pop.yaml#L27-L29

PR #1992, opened 2023-03-31
chore(buffering): Handle buffering messages sequentially and add more docs 

olksdr:
This is just small fixes and mostly adding more docs about how sqlite behaves. 

This also should makes sure we handle all the incoming messages to the buffer service sequentially, which makes sure that we always have `write` before `read`.  

related to https://github.com/getsentry/team-ingest/issues/53

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_buffer.rs
+++ b/relay-server/src/actors/project_buffer.rs
@@ -263,6 +263,58 @@ impl BufferService {
             )))
     }
 
+    /// Saves the provided buffer to the disk.
+    ///
+    /// It returns an error if the spooling failed.
+    async fn do_spool(
+        db: Pool<Sqlite>,
+        buffer: BTreeMap<QueueKey, Vec<ManagedEnvelope>>,
+    ) -> Result<(), BufferError> {
+        // A RESERVED lock will be acquired when the first INSERT, UPDATE, or DELETE statement is executed.
+        // Once we started spooling our in-memory buffer, the dequeue from the disk
+        // will be blocked till this operations either fails, or commits.
+        sqlx::query("BEGIN").execute(&db).await?;
+
+        // A builder type for constructing queries at runtime.
+        // This by default creates a prepared sql statement, which is cached and
+        // re-used for sequential queries.
+        let mut query_builder: QueryBuilder<Sqlite> =
+            QueryBuilder::new("INSERT INTO envelopes (own_key, sampling_key, envelope) ");
+
+        // Flatten all the envelopes
+        let envelopes = buffer.into_iter().flat_map(|(k, vals)| {
+            vals.into_iter()
+                .map(move |v| (k, v.into_envelope().to_vec()))
+        });
+
+        // Since we have 3 variables we have to bind, we devide the SQLite limit by 3
+        // here to prepare the chnunks which will be preparing the batch inserts.
+        let mut envelopes = stream::iter(envelopes).chunks(SQLITE_LIMIT_VARIABLE_NUMBER / 3);
+        while let Some(chunk) = envelopes.next().await {
+            query_builder.push_values(chunk.into_iter(), |mut b, (key, value)| match value {
+                Ok(value) => {
+                    b.push_bind(key.own_key.to_string())
+                        .push_bind(key.sampling_key.to_string())
+                        .push_bind(value);
+                }
+                Err(err) => {
+                    relay_log::error!("failed to serialize the envelope: {}", LogError(&err))
+                }
+            });
+            query_builder.build().execute(&db).await?;
+            // Reset the builder to initial state set by `QueryBuilder::new` function,
+            // so it can be reused for another chunk.
+            query_builder.reset();
+        }
+
+        // The SQL command "COMMIT" does not actually commit the changes to disk. It just turns autocommit back on.
+        // Then, at the conclusion of the command, the regular autocommit logic takes over and causes the actual commit to disk to occur.
+        // TODO: re-try failed commit?
+        sqlx::query("COMMIT").execute(&db).await?;
+
+        Ok(())
+    }
+
     /// Tries to save in-memory buffer to disk.
     ///
     /// It will spool to disk only if the persistent storage enabled in the configuration.
@@ -296,62 +348,11 @@ impl BufferService {
 
                 // spawn the task to enqueue the entire buffer
                 tokio::spawn(async move {
-                    // A RESERVED lock will be acquired when the first INSERT, UPDATE, or DELETE statement is executed.
-                    // Once we started spooling our in-memory buffer, the dequeue from the disk
-                    // will be blocked till this operations either fails, or commits.
-                    if let Err(err) = sqlx::query("BEGIN").execute(&db).await {
-                        relay_log::error!("failed to initiate a transaction: {}", LogError(&err));
-                    }
-
-                    // A builder type for constructing queries at runtime.
-                    // This by default creates a prepared sql statement, which is cached and
-                    // re-used for sequential queries.
-                    let mut query_builder: QueryBuilder<Sqlite> = QueryBuilder::new(
-                        "INSERT INTO envelopes (own_key, sampling_key, envelope) ",
-                    );
-
-                    // Flatten all the envelopes
-                    let envelopes = buf.into_iter().flat_map(|(k, vals)| {
-                        vals.into_iter()
-                            .map(move |v| (k, v.into_envelope().to_vec()))
-                    });
-
-                    // Since we have 3 variables we have to bind, we devide the SQLite limit by 3
-                    // here to prepare the chnunks which will be preparing the batch inserts.
-                    let mut envelopes =
-                        stream::iter(envelopes).chunks(SQLITE_LIMIT_VARIABLE_NUMBER / 3);
-                    while let Some(chunk) = envelopes.next().await {
-                        query_builder.push_values(chunk.into_iter(), |mut b, v| match v.1 {
-                            Ok(envelope_bytes) => {
-                                b.push_bind(v.0.own_key.to_string())
-                                    .push_bind(v.0.sampling_key.to_string())
-                                    .push_bind(envelope_bytes);
-                            }
-                            Err(err) => {
-                                relay_log::error!(
-                                    "failed to serialize the envelope: {}",
-                                    LogError(&err)
-                                )
-                            }
-                        });
-
-                        let query = query_builder.build();
-                        if let Err(err) = query.execute(&db).await {
-                            relay_log::error!(
-                                "failed to buffer envelopes to disk: {}",
-                                LogError(&err)
-                            )
-                        }
-                        // Reset the builder to initial state set by `QueryBuilder::new` function,
-                        // so it can be reused for another chunk.
-                        query_builder.reset();
-                    }
-
-                    // The SQL command "COMMIT" does not actually commit the changes to disk. It just turns autocommit back on.
-                    // Then, at the conclusion of the command, the regular autocommit logic takes over and causes the actual commit to disk to occur.
-                    // TODO: re-try failed commit?
-                    if let Err(err) = sqlx::query("COMMIT").execute(&db).await {
-                        relay_log::error!("failed to commit a transaction: {}", LogError(&err));
+                    if let Err(err) = Self::do_spool(db, buf).await {
+                        relay_log::error!(
+                            "failed to spool memory buffer to the disk: {}",
+                            LogError(&err)
+                        );
                     }
                 });
             }
--- a/relay-server/src/actors/project_buffer.rs
+++ b/relay-server/src/actors/project_buffer.rs
@@ -267,14 +267,9 @@ impl BufferService {
     ///
     /// It returns an error if the spooling failed.
     async fn do_spool(
-        db: Pool<Sqlite>,
+        db: &Pool<Sqlite>,
         buffer: BTreeMap<QueueKey, Vec<ManagedEnvelope>>,
     ) -> Result<(), BufferError> {
-        // A RESERVED lock will be acquired when the first INSERT, UPDATE, or DELETE statement is executed.
-        // Once we started spooling our in-memory buffer, the dequeue from the disk
-        // will be blocked till this operations either fails, or commits.
-        sqlx::query("BEGIN").execute(&db).await?;
-
         // A builder type for constructing queries at runtime.
         // This by default creates a prepared sql statement, which is cached and
         // re-used for sequential queries.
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/project_buffer.rs
```diff
@@ -296,6 +296,16 @@ impl BufferService {
 
                 // spawn the task to enqueue the entire buffer
                 tokio::spawn(async move {
+                    // A RESERVED lock will be acquired when the first INSERT, UPDATE, or DELETE statement is executed.
+                    // Once we started spooling our in-memory buffer, the dequeue from the disk
+                    // will be blocked till this operations either fails, or commits.
+                    if let Err(err) = sqlx::query("BEGIN").execute(&db).await {
+                        relay_log::error!("failed to initiate a transaction: {}", LogError(&err));
```
jan-auer: Shouldn't you exit early if begin fails? I'd refactor this to an async function that returns an error and then catch it in the spawn to log errors.

relay-server/src/actors/project_buffer.rs
```diff
@@ -296,6 +296,16 @@ impl BufferService {
 
                 // spawn the task to enqueue the entire buffer
                 tokio::spawn(async move {
+                    // A RESERVED lock will be acquired when the first INSERT, UPDATE, or DELETE statement is executed.
+                    // Once we started spooling our in-memory buffer, the dequeue from the disk
+                    // will be blocked till this operations either fails, or commits.
```
jan-auer: Where does it do that blocking?

olksdr: Another query, which will try to access this table will be blocked till the lock is released. This is also partially address the `fence` suggestion from yesterday. There will be definitely only 1 query at a time.

relay-server/src/actors/project_buffer.rs
```diff
@@ -290,55 +332,39 @@ impl BufferService {
...
+                            );
                         }
-                        // Reset the builder to initial state set by `QueryBuilder::new` function,
-                        // so it can be reused for another chunk.
-                        query_builder.reset();
-                    }
-                });
+
+                        // The SQL command "COMMIT" does not actually commit the changes to disk. It just turns autocommit back on.
+                        // Then, at the conclusion of the command, the regular autocommit logic takes over and causes the actual commit to disk to occur.
```
iker-barriocanal: I'm not sure if I understand this correctly; but if I do, this comment is not correct. Since we're using the [WAL journal mode](https://www.sqlite.org/wal.html#how_wal_works), a `COMMIT` will [add a checkpoint](https://www.sqlite.org/wal.html#automatic_checkpoint) (so committing the changes to disk?). Then, indeed, the default committing policy is again in-place.

iker-barriocanal (commented):
Other than the comment below, the rest looks good to me.

relay-server/src/actors/project_buffer.rs
olksdr: I do not really understand your comment here. 
The comment for this is taken from the docs, just to make sure to clarify what's happening while we call `commit` on the open transaction.

relay-server/src/actors/project_buffer.rs
```diff
@@ -263,10 +270,51 @@ impl BufferService {
...
+            QueryBuilder::new("INSERT INTO envelopes (own_key, sampling_key, envelope) ");
+
+        // Flatten all the envelopes
+        let envelopes = buffer.into_iter().flat_map(|(k, vals)| {
+            vals.into_iter()
+                .map(move |v| (k, v.into_envelope().to_vec()))
+        });
+
+        // Since we have 3 variables we have to bind, we devide the SQLite limit by 3
+        // here to prepare the chnunks which will be preparing the batch inserts.
```
```suggestion
        // here to prepare the chunks which will be preparing the batch inserts.
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -263,10 +270,51 @@ impl BufferService {
...
+            // Reset the builder to initial state set by `QueryBuilder::new` function,
+            // so it can be reused for another chunk.
+            query_builder.reset();
+        }
+        Ok(())
+    }
+
     /// Tries to save in-memory buffer to disk.
     ///
     /// It will spool to disk only if the persistent storage enabled in the configuration.
```
jjbayer, line 315:
```suggestion
    /// It will spool to disk only if the persistent storage enabled in the configuration
    /// and a db connection can be acquired.
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -290,55 +338,26 @@ impl BufferService {
...
+                    // Once we started spooling our in-memory buffer, the dequeue from the disk
+                    // will be blocked till this operations either fails, or commits.
+                    sqlx::query("BEGIN").execute(&mut connection).await?;
+
+                    Self::do_spool(&mut connection, buf).await?;
+
+                    // The SQL command "COMMIT" does not actually commit the changes to disk. It just turns autocommit back on.
+                    // Then, at the conclusion of the command, the regular autocommit logic takes over and causes the actual commit to disk to occur.
+                    // TODO: re-try failed commit?
+                    sqlx::query("COMMIT").execute(&mut connection).await?;
```
jjbayer: It looks like `sqlx` has a RAII wrapper for this: https://docs.rs/sqlx/latest/sqlx/struct.Transaction.html

Not sure if you want the automatic rollback in failure cases though.

relay-server/src/actors/project_buffer.rs
```diff
@@ -380,6 +399,13 @@ impl BufferService {
         sender: &mpsc::UnboundedSender<ManagedEnvelope>,
     ) -> Result<(), QueueKey> {
         loop {
+            // By default this creates a prepared statement which is cached and re-used.
+            //
+            // Removing envelopes from the on-disk buffer in batches has following implicatations:
```
```suggestion
            // Removing envelopes from the on-disk buffer in batches has following implications:
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -290,55 +338,26 @@ impl BufferService {
...
+                    // will be blocked till this operations either fails, or commits.
+                    sqlx::query("BEGIN").execute(&mut connection).await?;
+
+                    Self::do_spool(&mut connection, buf).await?;
+
+                    // The SQL command "COMMIT" does not actually commit the changes to disk. It just turns autocommit back on.
+                    // Then, at the conclusion of the command, the regular autocommit logic takes over and causes the actual commit to disk to occur.
+                    // TODO: re-try failed commit?
+                    sqlx::query("COMMIT").execute(&mut connection).await?;
+                }
```
jjbayer: Is it ok to do nothing if the connection cannot be acquired?

relay-server/src/actors/project_buffer.rs
```diff
@@ -290,55 +338,26 @@ impl BufferService {
                     return Err(BufferError::Full(estimated_db_size));
```
jan-auer: As a follow-up, we should: 
- try to spool as much as possible to fill up the disk (spool as many envelopes as possible)
- potentially: keep the rest in memory, and let the upstream service handle the case that we can no longer buffer. this must not "block" the caller on pending spooling, though.
- propagate an error up, via healthcheck and as 503 on queueing in the common endpoint handler.

relay-server/src/actors/project_buffer.rs
```diff
@@ -263,10 +270,51 @@ impl BufferService {
...
+        // Flatten all the envelopes
+        let envelopes = buffer.into_iter().flat_map(|(k, vals)| {
+            vals.into_iter()
+                .map(move |v| (k, v.into_envelope().to_vec()))
+        });
+
+        // Since we have 3 variables we have to bind, we devide the SQLite limit by 3
+        // here to prepare the chnunks which will be preparing the batch inserts.
+        let mut envelopes = stream::iter(envelopes).chunks(SQLITE_LIMIT_VARIABLE_NUMBER / 3);
+        while let Some(chunk) = envelopes.next().await {
```
jan-auer: From offline conversation: Intention here was to not use itertools. However, using some async function on a synchronous iterator from the `futures` crate doesn't buy us anything. Let's use itertools here.

relay-server/src/actors/project_buffer.rs
```diff
@@ -263,10 +270,51 @@ impl BufferService {
...
+        // A builder type for constructing queries at runtime.
+        // This by default creates a prepared sql statement, which is cached and
+        // re-used for sequential queries.
+        let mut query_builder: QueryBuilder<Sqlite> =
+            QueryBuilder::new("INSERT INTO envelopes (own_key, sampling_key, envelope) ");
+
+        // Flatten all the envelopes
+        let envelopes = buffer.into_iter().flat_map(|(k, vals)| {
+            vals.into_iter()
+                .map(move |v| (k, v.into_envelope().to_vec()))
```
jan-auer: This loop can take quite a while, and we're not yielding. This means that the thread on the main runtime that runs this future is going to be blocked for a long time. Should we either yield here, or defer this into a `spawn_blocking`? The latter is semantically wrong, and the CPU time has to be spent, but it should be spent in a context where we do not block other futures for too long.

Example:

```rust
let mut envelopes = Vec::new();
for k, vals in buffer {
  for val in vals {
    envelopes.insert((k, val.into_envelope().to_vec()));
    tokio::task::yield_now().await;  // TODO: consider doing this every Nth iteration
  }
}
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -380,6 +399,13 @@ impl BufferService {
...
     ) -> Result<(), QueueKey> {
         loop {
+            // By default this creates a prepared statement which is cached and re-used.
+            //
+            // Removing envelopes from the on-disk buffer in batches has following implicatations:
+            // 1. It is faster to delete from the DB in batches.
+            // 2. Make sure that if we panic and deleted envelopes cannot be read out fully, we do not lose all of them,
+            // but only one batch, and the rest of them will stay on disk for the next iteration
+            // to pick up.
             let mut envelopes = sqlx::query(
```
jan-auer, line 394: From offline conversation: Provided this could be a DELTE without subquery if a specific feature is enabled, let's add that dependency ourselves, enable the feature, and leave a comment in cargo toml, why we have it.

relay-server/src/actors/project_buffer.rs
```diff
@@ -519,14 +546,22 @@ impl Drop for BufferService {
...
-                relay_log::error!("dropped {} envelopes", count);
+            let buffer = std::mem::take(&mut self.buffer);
+            if let Some(BufferSpoolConfig { db, .. }) = &self.spool_config {
+                let db = db.clone();
+                tokio::spawn(async move {
+                    let Some(mut connection) = db.try_acquire() else {
+                        relay_log::error!("failed to get connection while dropping the project buffer");
+                        return;
+                    };
+                    if let Err(err) = Self::do_spool(&mut connection, buffer).await {
```
jan-auer: We drop this service mostly when the runtime stops, which means we should no longer start async tasks. Also, our expectation would be that when we drop the service, we close the DB and it's connections, which now isn't true because the DB lives on in a spawned task. 

This kind of procedure should probably be implemented in response to the `shutdown` signal.

For the MVP, it's enough if we drop the buffer and log outcomes + errors.

jan-auer (approved):
Approved provided that open comments are addressed. Thanks!

relay-server/src/actors/project_buffer.rs
iker-barriocanal: Do you mind linking to these docs then? Because I'm misunderstanding something.

relay-server/src/actors/project_buffer.rs
```diff
@@ -232,6 +234,10 @@ impl BufferService {
                 // 3. Disk I/O operations tends to be more sequential using WAL.
                 // 4. WAL uses many fewer fsync() operations and is thus less vulnerable to problems on systems where the fsync() system call is broken.
                 .journal_mode(SqliteJournalMode::Wal)
+                // WAL mode is safe from corruption with synchronous=NORMAL.
+                // When synchronous is NORMAL, the SQLite database engine will still sync at the most critical moments, but less often than in FULL mode.
+                // Which guarantees good balance between safety and speed.
+                .synchronous(SqliteSynchronous::Normal)
```
iker-barriocanal, line 239: I also find the following relevant, from the [docs](https://www.sqlite.org/pragma.html#:~:text=A%20transaction%20committed%20in%20WAL%20mode%20with%20synchronous%3DNORMAL%20might%20roll%20back%20following%20a%20power%20loss%20or%20system%20crash.):

> A transaction committed in WAL mode with synchronous=NORMAL might roll back following a power loss or system crash.

relay-server/src/actors/project_buffer.rs
iker-barriocanal: IMO `try_spool` should return an error in that case. If I call `try_spool` and it doesn't return an error, I'd expect I spooled. If there's any inner detail (i.e. connection could not be acquired), I'd expect I get the error so that I can choose to retry again if I want to.

relay-server/src/actors/project_buffer.rs
iker-barriocanal: (also adding "the":)
```suggestion
            // Removing envelopes from the on-disk buffer in batches has the following implications:
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -407,6 +400,12 @@ impl BufferService {
     ) -> Result<(), QueueKey> {
         loop {
             // By default this creates a prepared statement which is cached and re-used.
+            //
+            // Removing envelopes from the on-disk buffer in batches has following implicatations:
```
iker-barriocanal: nit: I'm not sure if I like the "on-disk" here since this in the end is an abstraction, but let's move forward with this for now.

iker-barriocanal (approved):

relay-server/src/actors/project_buffer.rs
olksdr: https://www.sqlite.org/lockingv3.html, see transaction control.

relay-server/src/actors/project_buffer.rs
olksdr: This explicit open/close transaction has to go away now. And we will rely on the pool to do the right thing, since we handle all the messages sequentially now

relay-server/src/actors/project_buffer.rs
olksdr: As agreed, adding this to followups and will jump one some of the items after merging this PR.

relay-server/src/actors/project_buffer.rs
olksdr: This is the related issue https://github.com/rusqlite/rusqlite/issues/1111, which explains the problem with this. 
If we want to enable this feature, we will have to build `sqlite` on our own. There is no easy way to just enable the feature.

relay-server/src/actors/project_buffer.rs
iker-barriocanal: Thanks! I misunderstood what checkpointing was, thus my confusion. It makes sense.

PR #1993, opened 2023-03-31
chore(build): bump Rust version to 1.68.2

olksdr:
Bumps toolchain to the latest current one: `1.68.2`

#skip-changelog


jan-auer (approved):
G2G. 

Let's automate this, however. We can inject the latest version from the build job, which means we can still use caches while always being on latest.

PR #1996, opened 2023-03-31
instr(txnames): Measure unique tx names

jjbayer:
Same as https://github.com/getsentry/relay/pull/1978, but counting unique transaction names.

> Log changes made to the transaction name by the TransactionProcessor to get an idea of what percentage of transactions is affected, and by which mechanism.
>
> We can consider this a temporary metric and delete it once we are done with this initiative.

#skip-changelog

### DIFF
--- a/relay-server/src/utils/statsd.rs
+++ b/relay-server/src/utils/statsd.rs
@@ -1,8 +1,11 @@
+use std::hash::Hasher;
+
+use fnv::FnvHasher;
 use relay_common::EventType;
 use relay_general::protocol::Event;
 use relay_general::types::{Annotated, RemarkType};
 
-use crate::statsd::RelayCounters;
+use crate::statsd::{RelayCounters, RelaySets};
 
 /// Log statsd metrics about transaction name modifications.
 ///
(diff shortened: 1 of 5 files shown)


relay-server/src/utils/statsd.rs
```diff
@@ -20,6 +23,14 @@ where
         return f(event);
     }
 
+    // Create a hash of the old name so we can track it in a set metric:
+
+    let old_name = {
```
olksdr:
```suggestion
    // Create a hash of the old name so we can track it in a set metric.
    let old_name = {
```

relay-server/src/utils/statsd.rs
```diff
@@ -58,6 +69,12 @@ where
         changes = changes,
         source_out = new_source.as_str(),
     );
+    relay_statsd::metric!(
+        set(RelaySets::TransactionNameChanges) = old_name,
```
olksdr, line 72: Will this create different tags for different hashes?

olksdr (approved):

relay-server/src/utils/statsd.rs
jjbayer: No, it will buffer a bit more data, but the amount of tags is the same as for the counter metric. In the end, it should only transmit one unique count per flush: https://github.com/statsd/statsd/blob/master/docs/metric_types.md#sets.

PR #1998, opened 2023-04-03
ref(dynamic-sampling): Refactor dynamic sampling

iambriccardo:
This PR refactors the dynamic sampling file and moves some functions to the `relay-sampling` crate for better separation of concerns.


relay-sampling/src/lib.rs
```diff
@@ -1039,6 +1018,96 @@ impl SamplingConfig {
     pub fn has_unsupported_rules(&self) -> bool {
         !self.rules_v2.iter().all(SamplingRule::supported)
     }
+
+    /// Checks whether unsupported rules result in a direct keep of the event or depending on the
+    /// type of Relay an ignore of unsupported rules.
+    fn check_unsupported_rules(
+        processing_enabled: bool,
+        sampling_config: &SamplingConfig,
```
iker-barriocanal: `check_unsupported_rules` was a top-level function, and it's now part of `SamplingConfig`, even if it doesn't need `self`. Why are we doing this change? Did you mean to change the `sampling_config` for `self`?

relay-sampling/src/lib.rs
```diff
@@ -1039,6 +1018,96 @@ impl SamplingConfig {
...
+        }
+
+        Some(())
+    }
+
+    /// Returns an iterator of references that chains together and merges rules.
+    ///
+    /// The chaining logic will take all the non-trace rules from the project and all the trace/unsupported
+    /// rules from the root project and concatenate them.
+    fn merge_rules_from_configs<'a>(
```
iker-barriocanal: Similar to above: why are we making this function part of `SamplingConfig`?

iker-barriocanal (commented):
I have quickly skimmed through the diff only, I haven't done an in-depth review.

Let's try to make PRs smaller and more focused on one goal. Ideally, if you move code there's nothing else that changes, that way is much easier to spot bugs and suggest improvements. I'd suggest you split this PR in two (or the number you consider necessary), especially if you want to add (even minor) improvements.

Thanks for your contribution!

relay-sampling/src/lib.rs
iambriccardo: I did change this since I wanted to group them as associated functions of the `SamplingConfig`. Does the ingest team have a preference for this? I generally prefer to be explicit about the association.

iker-barriocanal: I don't think we have any written document on how to contribute to relay yet.

It's good to have the associated function as part of `SamplingConfig`, but doing that change raises the question of why it accepts a `&SamplingConfig` instead of `&self`. And the follow-up: if you do that change, the scope of the PR grows from moving stuff around to changing functionality. I feel this is better suited for a follow-up PR -- we first move stuff, then we improve. Let me know if I'm missing something or if you think otherwise.

iambriccardo: That is sound, I did ask myself the same question regarding the `&self`. I can lift those functions out and in a followup PR improve.

PR #2004, opened 2023-04-04
feat(general): Add android product names

### LINKED ISSUE #1297: Map Android model to product name
### Problem Statement

For reference: https://github.com/jaredrummler/AndroidDeviceNames

Came up internally:

> They would like to see the Android device name (i.e., Samsung Galaxy), rather than device model number for their Discover results (not sure what field they’re referring to exactly). They have to do an additional step right now to google these model names. Is it possible to translate the device model number to a name inside the product?

### Solution Brainstorm

Instead of increasing the size of the SDK with this mapping that would ultimately get outdated all the time, we could do this on the backend. We need to get ingest folks involved on this one.

TBS1996:
#1297

This PR introduces a hashmap that maps from android models to the more human-friendly product-names. e.g. `NE2211` -> `OnePlus 10 Pro 5G`
 
The result is put in the `device.name` field, leaving the original `device.model` untouched.


android_models.csv
TBS1996, line 1: where should we put this file?

relay-general/src/store/mod.rs
```diff
@@ -66,6 +67,9 @@ pub struct StoreConfig {
 
     /// The SDK's sample rate as communicated via envelope headers.
     pub client_sample_rate: Option<f64>,
+
+    /// A csv file that can map from android model number
+    pub android_csv: PathBuf,
```
TBS1996: will remove if we will have a constant location for the csv

relay-general/src/store/normalize/contexts.rs
```diff
@@ -199,12 +204,44 @@ fn normalize_response(response: &mut ResponseContext) {
     }
 }
 
-pub fn normalize_context(context: &mut Context) {
+static ANDROID_MAP: OnceCell<Result<HashMap<String, String>, std::io::Error>> = OnceCell::new();
+
+fn get_android_product_name(model: &str) -> Option<String> {
+    let mymap: &Result<HashMap<String, String>, std::io::Error> = ANDROID_MAP.get_or_init(|| {
```
TBS1996: current behaviour: if it fails to initiate the map, it won't try again, hence returning a result. 

presumably, if it fails once it will fail every other time too.

we should be notified somehow if it fails.

relay-server/src/actors/processor.rs
```diff
@@ -1871,6 +1873,9 @@ impl EnvelopeProcessorService {
             );
         }
 
+        let mut android_csv = PathBuf::from(self.config.path());
```
olksdr: This won't compile the file into Relay, which we decided to do if I recall it right.  This will always try to open and read the file from the path. 

You have to look into something like [`include_str!`](https://doc.rust-lang.org/std/macro.include_str.html) or `include_bytes!` macros.

relay-general/src/store/normalize/contexts.rs
olksdr: It's not necessary if we compile Relay with file together, see my previous comment.  And we will fail if the file is not there while we build the binary.

olksdr (dismissed):

relay-general/src/store/normalize/contexts.rs
```diff
@@ -199,12 +204,44 @@ fn normalize_response(response: &mut ResponseContext) {
...
     match context {
         Context::Runtime(runtime) => normalize_runtime_context(runtime),
         Context::Os(os) => normalize_os_context(os),
         Context::Response(response) => normalize_response(response),
-        _ => (),
+        Context::Device(device) => {
+            if let Some(model) = device.as_ref().model.value() {
+                if let Some(product_name) = get_android_product_name(model) {
+                    device.model.set_value(Some(product_name));
+                }
```
iker-barriocanal: nit: if possible, I'd use [`and_then`](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then) here, I think it's more readable. Something like:
```suggestion
                get_android_product_name(model).and_then(|product_name| device.model.set_value(Some(product_name)));
                }
```

relay-general/src/store/normalize/contexts.rs
```diff
@@ -199,12 +204,44 @@ fn normalize_response(response: &mut ResponseContext) {
...
+        for line in reader.lines() {
+            let line = line?;
+            let columns: Vec<&str> = line.split(',').collect();
+
+            if columns.len() >= 4 {
+                let key = columns[3].trim().to_string();
+                let value = columns[1].trim().to_string();
+                android_map.insert(key, value);
+            }
+        }
```
iker-barriocanal: Adding another dependency to deal with a single CSV file may not be worth it, but I don't like parsing it manually for the number of things that can go wrong. See, for example, [this StackOverflow](https://stackoverflow.com/a/10141147) answer about the delimiter (`,` or `;`).

I guess it's fine to hardcode it like this since we're inserting the file at compile time.

relay-general/src/store/normalize/contexts.rs
```diff
@@ -215,6 +252,20 @@ mod tests {
     use super::*;
     use crate::protocol::LenientString;
 
+    #[test]
+    fn test_get_product_name() {
```
iker-barriocanal, line 229: Let's also try a model that isn't on the list.

android_models.csv
iker-barriocanal: I guess somewhere in `relay-general` I guess. I'd create a `files/` directory in there, but I don't know about the best practices or if this is too much. @olksdr or @jjbayer any suggestions here?

relay-general/src/store/normalize/android_map.rs
TBS1996, line 1: script to generate this:

```python
input_file = 'android_models.csv'
output_file = 'android_map.rs'


def filter(key, val) -> bool:
    normalized_key = key.replace('_', " ").replace('-', ' ').lower()
    normalized_val = val.replace('_', " ").replace('-', ' ').lower()
    if normalized_val in normalized_key:
        return True
    if key == "" or val == "":
        return True
    return False

with open(input_file, 'r') as csv_file:
    csv_reader = csv.reader(csv_file)

    with open(output_file, 'w') as output:
        output.write("""use once_cell::sync::Lazy;
use std::collections::HashMap;

static ANDROID_MAP: Lazy<HashMap<&'static str, &'static str>> = Lazy::new(|| {\n\tlet data = vec![\n""")
        for row in csv_reader:
            key = row[3].replace('"', "").replace('\\', "")
            val = row[1].replace('"', "").replace('\\', "")
            if filter(key, val):
                continue

            formatted_line = f'\t\t("{key}", "{val}"),\n'
            output.write(formatted_line)
        output.write("""];

    data.iter().cloned().collect()
});
""")

print(f"Output written to {output_file}")
```

TBS1996: thoughts on this approach? 

pros: compiled time check, no run-time parsing, we won't store a huge static &str in memory that we only use for creating a hashmap

cons: gotta manually run this python script (I could make a reminder to do this at fixed intervals if i cant automate it). Also increases the code-base by a lot, not sure if that matters

TBS1996:
currently failing a lot of integration tests because it expects the model values to be equal to the value in the User Agent. 

looking into this..

im wondering if we want to put this in device.model at all, i notice that there's no events in sentry discover that has a non-emtpy device.name. Would that be a better fit?

relay-general/src/store/normalize/contexts.rs
```diff
@@ -215,6 +226,21 @@ mod tests {
...
```
TBS1996, line 229: this test feels a bit silly now since it's only checking that the values in the hashmap exist and nothing more

iker-barriocanal (changes requested):
I see you hardcoded the values into the file, by running the script. Doing it with a script is better than doing it manually, but when you do so please also include the script to update the models if the file is updated. 

Have you looked at [this comment](https://github.com/getsentry/relay/pull/2004#discussion_r1157184878) from Olek? It includes a couple of suggestions of macros that could be useful to add the file to relay, without running any scripts. Having a script is a worse alternative long-term, since you need to maintain the script and deal with the edge-cases of it creating a file (should it overwrite the file? what if we want to extend the file?).

android_models.csv
TBS1996: i went ahead with your files/ idea, so we can discuss now if this is a good solution

relay-general/src/store/normalize/contexts.rs
TBS1996: are you thinking of this crate?: https://crates.io/crates/csv

now we are back to run-time parsing so this is more relevant again.

TBS1996:
there is a circular dependency in getsentry/sentry. i'll make a PR there to temporarily disable a test

files/android_models.csv
iker-barriocanal, line 1: Since this is compiled in `relay-general`, it should exist in that directory. `/files/` look good to me, I'm not sure if there are any best practices here.

relay-general/src/protocol/contexts/device.rs
```diff
@@ -1,8 +1,27 @@
...
+use once_cell::sync::Lazy;
+
 use crate::protocol::FromUserAgentInfo;
 use crate::store::user_agent::is_known;
 use crate::types::{Annotated, Object, Value};
 use crate::user_agent::{parse_device, ClientHints};
 
+/// A hashmap that translates from the android model to the more human-friendly product-names.
+/// E.g. NE2211 -> OnePlus 10 Pro 5G
+pub static ANDROID_MAP: Lazy<HashMap<&'static str, &'static str>> = Lazy::new(|| {
```
iker-barriocanal: `ANDROID_MAP` doesn't really tell me anything. I don't have any suggestion that I really like, but I think this one is better:
```suggestion
pub static ANDROID_MODEL_NAMES: Lazy<HashMap<&'static str, &'static str>> = Lazy::new(|| {
```

relay-general/src/protocol/contexts/device.rs
```diff
@@ -1,8 +1,27 @@
...
+    let mut map = HashMap::new();
+    let android_str = include_str!("../../../../files/android_models.csv");
```
iker-barriocanal: From the [docs](https://doc.rust-lang.org/std/macro.include_str.html):

> So, for instance, an invocation with a Windows path containing backslashes \ would not compile correctly on Unix.

I think we should add a note about that.

TBS1996:
getting a CI error because it `github-actions-gcr` dont have proper access
https://github.com/getsentry/relay/actions/runs/4627412932/jobs/8187451550?pr=2004

not sure why it doesn't have..

relay-general/src/protocol/contexts/device.rs
```diff
@@ -1,8 +1,27 @@
...
+
+    for line in android_str.lines() {
+        let fields: Vec<&str> = line.split(',').collect();
+        if fields.len() >= 4 {
+            map.insert(fields[3].trim(), fields[1].trim());
```
iker-barriocanal: Currently, the first line (the header) is also included in the map I believe, and reordering of columns can break this. I think using the first line and extracting the column indices from there is more solid -- I'd look at the indices of `Device` and `Model`, so if Google inserts another column at the beginning, this will still work. I believe changing the column name is less likely than introducing new column names or reordering them.

relay-general/src/store/normalize/contexts.rs
```diff
@@ -215,6 +225,21 @@ mod tests {
...
+        let model = "NE2211";
+        let product_name = ANDROID_MAP.get(model).unwrap();
+        assert_eq!(product_name, &"OnePlus 10 Pro 5G");
```
iker-barriocanal: nit: I think we can simplify this a bit like you did with the example of `foobar` below:
```suggestion
        assert_eq!(ANDROID_MAP.get("NE2211").unwrap(), &"OnePlus 10 Pro 5G");
```

iker-barriocanal (changes requested):
The PR looks good! Let's address the headers comment and we're good to go.

relay-general/src/protocol/contexts/device.rs
TBS1996: hmm yeah good point, I can do this. although if something change it doesn't really matter since it will fail the test so we would be alerted

relay-general/src/store/normalize/contexts.rs
TBS1996: good point! thats definitively better

files/android_models.csv
TBS1996: so youre thinking a /files/ directory inside of relay-general crate?

relay-general/src/protocol/contexts/device.rs
TBS1996: ok, but what circumstances would this be useful? I guess something about self-hosted on windows?

relay-general/src/protocol/contexts/device.rs
```diff
@@ -1,8 +1,46 @@
...
+
+    let header = lines.next().expect("CSV file should have a header");
+
+    let header_fields: Vec<&str> = header.split(',').collect();
+    let model_index = header_fields.iter().position(|&s| s.trim() == "Model");
+    let product_name_index = header_fields
+        .iter()
+        .position(|&s| s.trim() == "Marketing Name");
+
+    let (model_index, product_name_index) = match (model_index, product_name_index) {
```
TBS1996, line 27: i wish rust had an unwrap_or_return one-liner..

relay-general/src/store/normalize/contexts.rs
```diff
@@ -204,7 +204,17 @@ pub fn normalize_context(context: &mut Context) {
...
         Context::Response(response) => normalize_response(response),
-        _ => (),
+        Context::Device(device) => {
+            if let Some(product_name) = device
+                .as_ref()
+                .model
+                .value()
+                .and_then(|model| ANDROID_MODEL_NAMES.get(model.as_str()))
+            {
+                device.name.set_value(Some(product_name.to_string()))
```
TBS1996, line 214: I want to highlight that from the earliest commits I changed it from setting the value to "model" to setting it to "name".  

on struct `devicecontext` the namefield just has the comment `Name of the device` which seems a bit unclear but the product name of it sounds like a good fit. It's also not used at all. If you go to sentry discover and filter events by those that have a non-empty device.name field, you get zero results: 

https://sentry.sentry.io/discover/homepage/?field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&name=All+Events&query=%21device.name%3A%22%22&sort=-timestamp&statsPeriod=14d&yAxis=count%28%29

this way we also get  to keep the original model name which could be useful. Let me know if there's any objections to this.

relay-general/src/protocol/contexts/device.rs
```diff
@@ -1,8 +1,46 @@
...
+/// A hashmap that translates from the android model to the more human-friendly product-names.
+/// E.g. NE2211 -> OnePlus 10 Pro 5G
+pub static ANDROID_MODEL_NAMES: Lazy<HashMap<&'static str, &'static str>> = Lazy::new(|| {
+    let mut map = HashMap::new();
+    // Note that windows paths with backslashes '\' won't work on unix systems.
+    let android_str = include_str!("../../../../files/android_models.csv");
+
+    let mut lines = android_str.lines();
+
+    let header = lines.next().expect("CSV file should have a header");
```
TBS1996, line 19: i think unwraps are fine here because if it panics it will fail the tests, and the string is static so it will be the same in tests as in production

relay-general/src/protocol/contexts/device.rs
iker-barriocanal: Yes, and a contributor too for example. I'm not sure how many people are in that situation so my ask is not to do a fancy solution for them.

relay-general/src/store/normalize/contexts.rs
iker-barriocanal: Well, now it's also checking the headers are passed correctly 😛.

relay-general/src/protocol/contexts/device.rs
```diff
@@ -1,8 +1,46 @@
...
+
+    let header_fields: Vec<&str> = header.split(',').collect();
+    let model_index = header_fields.iter().position(|&s| s.trim() == "Model");
+    let product_name_index = header_fields
+        .iter()
+        .position(|&s| s.trim() == "Marketing Name");
+
+    let (model_index, product_name_index) = match (model_index, product_name_index) {
+        (Some(model_index), Some(product_name_index)) => (model_index, product_name_index),
+        (_, _) => return HashMap::new(),
```
iker-barriocanal: It's good that we have the test, but I think the compilation should error in this case.

files/android_models.csv
iker-barriocanal: I'm not convinced of `files` as a name, but I can't think of better names right now. However, the CSV file is compiled with `relay-general`, so it should be somewhere inside `relay-general/`.

iker-barriocanal (changes requested):
Let's move the CSV file first, see my comment above about it.

relay-general/src/protocol/contexts/device.rs
TBS1996: i added a small comment, is it what you had in mind?

relay-general/src/protocol/contexts/device.rs
TBS1996: so regular unwraps? is it right to say compilation would fail though? it's static but it's lazy, so I thought it doesn't get executed until the first access, rather than during compilation

files/android_models.csv
TBS1996: ok, but unless someone suggest a better name i'll leave it as files/ for now

relay-general/src/protocol/contexts/device.rs
iker-barriocanal: yep yep

relay-general/src/protocol/contexts/device.rs
iker-barriocanal: Yeah you're right, sorry for my comment and the confusion it generated. I see you're now using `unwrap`, let's put it back as it was before so that it doesn't panic in runtime, and log a sentry error so that we're aware something changed, see [this example](https://github.com/getsentry/relay/blob/c3b339e151c1e548ede489a01c65db82472c8751/relay-server/src/actors/relays.rs#L323).

iker-barriocanal (approved):
🚀

PR #2010, opened 2023-04-05
ref(metrics): Don't report unknown transaction sources to Sentry

iker-barriocanal:
All new events of this issue are coming from old SDK versions or users setting the source manually. The issue is no longer useful or actionable.

#skip-changelog


jjbayer (approved):
LGTM, if a user notices that their transactions disappear in `<< unparametrized >>`, they can always contact us.

olksdr (approved):
🚀

PR #2011, opened 2023-04-05
instr(buffers): Log buffer sizes as statsd metrics

jjbayer:
Introduce statsd metrics around persistent buffering.

- [x] num envelopes in memory
- [x] db size (bytes)
- [x] read / write operations

#skip-changelog


relay-server/src/actors/project_buffer.rs
```diff
@@ -198,7 +204,12 @@ impl BufferService {
         let db = SqlitePoolOptions::new().connect_with(options).await?;
         sqlx::migrate!("../migrations").run(&db).await?;
 
-        Ok(())
+        let count: i64 = sqlx::query("SELECT count(*) FROM envelopes")
```
jjbayer: We could also get the number of bytes here, and create a metric for that, but it's a heavy operation when the db is big:

```suggestion
        let count: i64 = sqlx::query("SELECT count(*), sum(length(envelope)) FROM envelopes")
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -186,10 +187,15 @@ pub struct BufferService {
     project_cache: Addr<ProjectCache>,
     spool_config: Option<BufferSpoolConfig>,
     count_mem_envelopes: i64,
+    count_disk_envelopes: i64,
+
+    #[cfg(test)]
+    untracked: bool,
```
jjbayer, line 191:
```suggestion
    /// Create untracked envelopes when loading from disk. Only use in tests.
    untracked: bool,
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -535,3 +590,104 @@ impl Drop for BufferService {
...
+        let config = Config::from_json_value(serde_json::json!({
+            "cache": {
+                "persistent_envelope_buffer": {
+                    "path": dbg!(std::env::temp_dir().join(Uuid::new_v4().to_string())),
+                    "max_memory_size": 2,
+                }
+            }
+        }))
+        .unwrap();
+
```
```suggestion
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -535,3 +590,104 @@ impl Drop for BufferService {
...
+        let captures = relay_statsd::with_capturing_test_client(|| {
+            rt.block_on(async {
+                let (project_cache, _) = mock_service("project_cache", (), |&mut (), _| {});
+                let mut service = BufferService::create(buffer_guard, project_cache, config.into())
+                    .await
+                    .unwrap();
+                service.untracked = true; // so we do not panic when dropping envelopes
+
+                for _ in 0..5 {
```
jjbayer, line 680:
```suggestion
                // Send 5 envelopes
                for _ in 0..5 {
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -535,3 +590,104 @@ impl Drop for BufferService {
...
+                    service
+                        .handle_enqueue(Enqueue {
+                            key,
+                            value: empty_envelope(),
+                        })
+                        .await
+                        .unwrap();
+                }
+                let (tx, mut rx) = mpsc::unbounded_channel();
```
jjbayer, line 691:
```suggestion

                // Dequeue everything
                let (tx, mut rx) = mpsc::unbounded_channel();
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -198,7 +205,12 @@ impl BufferService {
         let db = SqlitePoolOptions::new().connect_with(options).await?;
         sqlx::migrate!("../migrations").run(&db).await?;
 
-        Ok(())
+        let count: i64 = sqlx::query("SELECT count(*) FROM envelopes")
```
olksdr: Select `count` on the big tables can be very slow, when it starts doing full table scans. But also since we always delete from the table, there is no easy way around this.

relay-server/src/statsd.rs
```diff
@@ -42,6 +42,20 @@ pub enum RelayHistograms {
     ///
     /// The queue size can be configured with `cache.event_buffer_size`.
     EnvelopeQueueSize,
+    /// The number of envelopes waiting for project states in memory.
+    ///
+    /// This number is always <= `EnvelopeQueueSize`.
+    ///
+    /// The memory buffer size can be configured with `cache.persistent_envelope_buffer.max_memory_size`.
```
olksdr:
```suggestion
    /// The memory buffer size can be configured with `spool.envelopes.max_memory_size`.
```

relay-server/src/statsd.rs
```diff
@@ -42,6 +42,20 @@ pub enum RelayHistograms {
...
+    BufferEnvelopesMemory,
+    /// The number of envelopes waiting for project states on disk.
+    ///
+    /// During normal operation, this number should be 0. Only when
+    /// the in-memory buffer overflows (for example, we cannot fetch project states from the upstream)
+    /// will the disk buffer be used.
+    ///
+    /// The disk buffer size can be configured with `cache.persistent_envelope_buffer.max_disk_size`.
```
olksdr:
```suggestion
    /// The disk buffer size can be configured with `spool.envelopes.max_disk_size`.
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -341,8 +358,17 @@ impl BufferService {
 
             let buf = std::mem::take(buffer);
             self.count_mem_envelopes = 0;
-            Self::do_spool(db, buf).await?;
```
olksdr, line 347: On the line `351` we also get estimated database file size. I think it also would be great to have a metric on that, to see how the file grow with time. What do you think, does it make sense?

relay-server/src/actors/project_buffer.rs
```diff
@@ -353,6 +379,10 @@ impl BufferService {
         let envelope_row: Vec<u8> = row.try_get("envelope")?;
         let envelope_bytes = bytes::Bytes::from(envelope_row);
```
olksdr, line 378: What do you think we also can get the histogram of envelopes size distribution?

olksdr (approved):
Overall LGTM! 
I left only few questions I would also like to hear your opinion on.

relay-server/src/actors/project_buffer.rs
jjbayer: Hmm that's surprising, I would have expected the db to have knowledge about its table sizes. Anyway, I think we can live with it if it only happens once during initialization?

relay-server/src/actors/project_buffer.rs
jjbayer: I considered it, but I'm not sure if its misleading if it only logs the size _before_ it spools. But I'm happy to add it anyway.

relay-server/src/actors/project_buffer.rs
jjbayer: Yes, I'm surprised we don't have it for incoming envelopes. We do have it for outgoing envelopes, maybe that's enough for estimations?

https://github.com/getsentry/relay/blob/004fe550b7b6ce02357b74dcd6b0bbb213dc508b/relay-server/src/statsd.rs#L133-L134

relay-server/src/actors/project_buffer.rs
olksdr: We could actually test it, just create bug table (~50 - 100 GB) with some inserts and deletes and run the `count` on it.

relay-server/src/actors/project_buffer.rs
olksdr: On one side yes, but it'll also give us at least some overview what's going on with the system I guess.

relay-server/src/actors/project_buffer.rs
olksdr: Hm, you might be right. Is one request can have multiple envelopes in the body here? Which wouldn't give us actual size we would like to see.

relay-server/src/actors/project_buffer.rs
```diff
@@ -186,10 +187,16 @@ pub struct BufferService {
...
+    count_disk_envelopes: i64,
+
+    #[cfg(test)]
+    /// Create untracked envelopes when loading from disk. Only use in tests.
+    untracked: bool,
 }
 
 impl BufferService {
-    async fn setup(path: &PathBuf) -> Result<(), BufferError> {
+    /// Set up database and return the current number of envelopes.
```
iker-barriocanal:
```suggestion
    /// Set up the database and return the current number of envelopes.
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -302,12 +318,16 @@ impl BufferService {
                     relay_log::error!("failed to serialize the envelope: {}", LogError(&err));
                 }
             });
-            query_builder.build().execute(db).await?;
+            let result = query_builder.build().execute(db).await?;
+            relay_statsd::metric!(counter(RelayCounters::BufferWrites) += 1);
```
iker-barriocanal, line 325: This will generate a ton of metrics. Can we use a counter here for each loop iteration, and log the metric outside of the loop? The result should be pretty similar, with less overhead.

relay-server/src/actors/project_buffer.rs
```diff
@@ -398,12 +440,15 @@ impl BufferService {
             .bind(key.sampling_key.to_string())
             .fetch(db).peekable();
 
+            relay_statsd::metric!(counter(RelayCounters::BufferReads) += 1);
```
iker-barriocanal, line 435: Same suggestion here, should we have a counter and log a single metric instead of doing so on each iteration?

iker-barriocanal (approved):
I haven't looked at the tests, but the PR overall looks good.

relay-server/src/actors/project_buffer.rs
jjbayer: It took roughly 10 seconds for a billion rows (97 GB) on my machine. Still too much if you ask me. I refactored the metric to log the estimated db size in bytes instead.

relay-server/src/actors/project_buffer.rs
jjbayer: I considered it, but I wanted the metric to be as real time as possible. This logs one metric for each batch of 333 envelopes, so I hope it's ok.

relay-server/src/actors/project_buffer.rs
jjbayer: Same argument as before: the batch size is 100, so I hope the volume is acceptable.

relay-server/src/actors/project_buffer.rs
```diff
@@ -265,18 +270,27 @@ impl BufferService {
...
+        };
+        let page_size: i64 = match rows.next().await {
+            Some(row) => row?.try_get(0)?,
+            None => return Err(BufferError::DatabaseError(sqlx::Error::RowNotFound)),
+        };
+        let size = page_count * page_size;
+
+        relay_statsd::metric!(histogram(RelayHistograms::BufferDiskSize) = size as u64);
+
+        Ok(size)
```
jjbayer, line 288: @olksdr I completely changed the way we compute the db size so we can log it as a metric more often. Let me know if you approve.

olksdr: Yeah, that would definitely work, and we do not have to all the time hit the filesystem.

olksdr (approved):
:shipit:

relay-server/src/actors/project_buffer.rs
iker-barriocanal: Seems reasonable, yes 👍

PR #2012, opened 2023-04-05
ref(buffering): Refactor relay config

olksdr:
Refactor the configuration setup for persistent buffering:
* add the top level `spool` option
* currently `spool` contains the `envelopes` configuration - this is done this way because we are planning to add more spooling to more places, which might require different setup
* add very rough estimate of how many envelopes can we fit into configured memory limit - should we have some pre-defined minimum which we must enforce?  

#skip-changelog

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -764,7 +764,11 @@ pub struct EnvelopeSpool {
     /// If not set the befault is 524288000 bytes (500MB).
     #[serde(default = "spool_envelopes_max_disk_size")]
     max_disk_size: usize,
-    /// The maximum number of envelopes to keep in the memory buffer before spooling them to disk.
+    /// The maximum bytes to keep in the memory buffer before spooling envelopes to disk, in bytes.
+    ///
+    /// This is a hard upper bound. Internally, this is converted to an envelope count by dividing
+    /// this number by the maximum envelope size, so in practice, because the avg. envelope
+    /// size is well below the maximum, we start spooling to disk long before this hard limit is reached.
     #[serde(default = "spool_envelopes_max_memory_size")]
     max_memory_size: usize,
 }
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -726,31 +726,66 @@ impl Default for Http {
     }
 }
 
-fn buffer_max_connections() -> u32 {
-    30
+/// Default for max memory size, 500 MB.
+fn spool_envelopes_max_memory_size() -> usize {
+    524288000
 }
 
-fn buffer_min_connections() -> u32 {
+/// Default for max disk size, 500 MB.
+fn spool_envelopes_max_disk_size() -> usize {
+    524288000
+}
+
+/// Default for min connections to keep open in the pool.
+fn spool_envelopes_min_connections() -> u32 {
     10
 }
 
-/// Controls internal caching behavior.
+/// Default for max connections to keep open in the pool.
+fn spool_envelopes_max_connections() -> u32 {
+    20
+}
+
+/// Persistent buffering configuration for incoming envelopes.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct PersistentBuffer {
-    /// The path to the persistent buffer file.
-    path: PathBuf,
+pub struct EnvelopeSpool {
+    /// The path to the persistent spool file.
+    ///
+    /// If set, this will enable the buffering for incoming envelopes.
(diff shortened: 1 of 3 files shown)


relay-server/src/actors/project_buffer.rs
```diff
@@ -243,15 +243,19 @@ impl BufferService {
...
             let spool_config = BufferSpoolConfig {
                 db,
-                max_disk_size: config.cache_persistent_buffer_max_disk_size(),
-                max_memory_size: config.cache_persistent_buffer_max_memory_size(),
+                max_disk_size: config.spool_envelopes_max_disk_size(),
+                // It is a rough extimation for how many envelopes we can fit in the
+                // configured memory limit, taking that 1 enveloper is 1 MB.
+                //
+                // TODO: Can we calculate the size of the envelope?
+                max_envelopes_count: config.spool_envelopes_max_memory_size() / (1024 * 1024),
```
jjbayer: Something like this?

```suggestion
                max_envelopes_count: config.spool_envelopes_max_memory_size() / config.max_envelope_size(),
```

relay-config/src/config.rs
```diff
@@ -726,31 +726,66 @@ impl Default for Http {
...
+    #[serde(default = "spool_envelopes_min_connections")]
     min_connections: u32,
     /// The maximum size of the buffer to keep, in bytes.
     ///
-    /// If not set the befault is 10737418240 bytes or 10 GB.
-    max_disk_size: Option<usize>,
+    /// If not set the befault is 524288000 bytes (500MB).
+    #[serde(default = "spool_envelopes_max_disk_size")]
+    max_disk_size: usize,
     /// The maximum number of envelopes to keep in the memory buffer before spooling them to disk.
```
```suggestion
    /// The maximum bytes to keep in the memory buffer before spooling envelopes to disk, in bytes.
    ///
    /// This is a hard upper bound. Internally, this is converted to an envelope count by dividing this number by the maximum envelope size, so in practice, because the avg. envelope size is well below the maximum, we start spooling to disk long before this hard limit is reached.
```

jjbayer (changes requested):

relay-server/src/actors/project_buffer.rs
olksdr: done in [15f57c7](https://github.com/getsentry/relay/pull/2012/commits/15f57c7a6b6d26639f69f7c74c864e2ff441c495)

relay-config/src/config.rs
olksdr: also done in [15f57c7](https://github.com/getsentry/relay/pull/2012/commits/15f57c7a6b6d26639f69f7c74c864e2ff441c495)

jjbayer (approved):

relay-config/src/config.rs
```diff
@@ -726,31 +726,70 @@ impl Default for Http {
...
-    #[serde(default = "buffer_min_connections")]
+    #[serde(default = "spool_envelopes_min_connections")]
     min_connections: u32,
     /// The maximum size of the buffer to keep, in bytes.
     ///
-    /// If not set the befault is 10737418240 bytes or 10 GB.
-    max_disk_size: Option<usize>,
-    /// The maximum number of envelopes to keep in the memory buffer before spooling them to disk.
-    max_memory_size: Option<usize>,
+    /// If not set the befault is 524288000 bytes (500MB).
```
iker-barriocanal, line 764: I wonder if, when generating the docs for this value, it could be possible to include the docs of the default value. Could be an interesting feature for the compiler or whatever is in charge of doing it.

relay-config/src/config.rs
```diff
@@ -1699,63 +1734,34 @@ impl Config {
...
     }
 
-    /// Returns `true` if the persistent envelope buffer is enabled, `false` otherwise.
-    pub fn cache_persistent_buffer_enabled(&self) -> bool {
-        self.values.cache.persistent_envelope_buffer.is_some()
-    }
-
     /// Returns the path of the buffer file if the `cache.persistent_envelope_buffer.path` is configured.
-    pub fn cache_persistent_buffer_path(&self) -> Option<PathBuf> {
+    pub fn spool_envelopes_path(&self) -> Option<PathBuf> {
```
iker-barriocanal, line 1738: I don't know what best practices are, but should we panic or something if the config has `spool.envelopes` without a path? It's basically wrong config.

iker-barriocanal (commented):
Post-close review with a couple of questions, the PR looks good.

PR #2013, opened 2023-04-05
feat: Apply schema validation to all topics in local development

untitaker:
Alternative to https://github.com/getsentry/relay/pull/1981 -- apply
schema validation to all outgoing messages when Relay runs in
development mode.

This means that messages are validated as part of integration tests.

The additional dependencies are compiled unconditionally, but unused
in release builds. The only way to conditionally compile them is to
introduce yet another featureflag, but that seems burdensome to use
since we use `--all-features` everywhere right now. We have had similar
issues introducing optional SSL support in Kafka a long time ago, IIRC

This validation does _not_ run in sentry's local development because (I think)
it uses release builds there. In order to make that happen we could add
a "validate schemas" flag to config.yml, off by default.


relay-kafka/src/producer/mod.rs
```diff
@@ -36,6 +39,10 @@ pub enum ClientError {
     #[error("failed to serialize json message")]
     InvalidJson(#[source] serde_json::Error),
 
+    /// Failed to run schema validation on message.
```
jjbayer, line 46: Can this be conditional?

```suggestion
    #[cfg(debug_assertions)]
    /// Failed to run schema validation on message.
```

relay-kafka/src/producer/schemas.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+pub fn validate_message_schema(topic: KafkaTopic, message: &[u8]) -> Result<(), SchemaError> {
+    let default_assignments = TopicAssignments::default();
+    let logical_topic_name = match default_assignments.get(topic) {
+        TopicAssignment::Primary(logical_topic_name) => logical_topic_name,
+        _ => return Err(SchemaError::InvalidLogicalTopic),
+    };
+
+    let schema = match sentry_kafka_schemas::get_schema(logical_topic_name, None) {
```
jjbayer: I know this is only for debug builds, but should we keep a cache of schemas by topic name so it isn't loaded from disk for every message? Maybe even put that cache in the `sentry-kafka-schemas` library?

jjbayer: nit: Should the library have a `get_latest_schema` helper to make it more explicit what the `None` does? Or an explicit enum with variant `SchemaVersion::Latest` instead of an `Option`?

relay-kafka/Cargo.toml
```diff
@@ -18,6 +18,8 @@ rmp-serde = { version = "1.1.1", optional = true }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = { version = "1.0.55", optional = true }
 thiserror = "1.0.38"
+sentry-kafka-schemas = "0.0.25"
+jsonschema = "0.17.0"
```
jjbayer, line 22: I wonder if it would make sense to hide the `jsonschema` dependency in `sentry-kafka-schemas`. That is, if we removed this dependency here, and `sentry-kafka-schemas` exposed a `validate(topic, message)` function. 

That way, the library client (Relay) would not have to deal with compiling schemas, and the way that `sentry-kafka-schemas` validates would be completely encapsulated (i.e. you could change the library or even the format it uses for schema definition and validation).

jjbayer (changes requested):
I like this approach! It has zero runtime overhead in production but is more automatic than relying on manually crafted tests.

Pressing "Request changes" because of some open questions.

relay-kafka/src/producer/schemas.rs
untitaker: Maybe sometime but for now we don't want to be opinionated in how people use this library, so we settled for the simplest interface that comes with the least amount of dependencies (aka, return schema as string). This goes for the Python library as well: It requires you to bring your own schema validation library right now, and figure out yourself where it makes sense to store the schema. As such it doesn't have many dependencies. This is a good thing if you're only using the generated types in your service.

I also don't want to build any cache right now because I foresee a (potential, not certain) future where we want/need people to iterate on schemas and relay in parallel (i.e. temporarily pointing relay to a local checkout of the schemas repo), and it would be good if changes to the schema files were picked up immediately in local development without recompiling.

I'm also not convinced that this schema validation library is the best one to use. I detest the fact that the schema and the payload has to be deserialized into intermediate `serde_json::Value`, but it seems that that is commonpractice for all schema validation libraries.

untitaker: I'm happy to add a `get_latest_schema` helper (or rename to have `get_schema` and `get_schema_version`) -- how strongly do you feel about this happening?

relay-kafka/Cargo.toml
untitaker: See my reply to the local cache comment. For now I would like to avoid that. There's still many unsolved questions around how the workflow in Relay will look like, what performance requirements we have on the entire system, etc.

relay-kafka/src/producer/mod.rs
untitaker: yes, it has to be, fixed

relay-kafka/src/producer/schemas.rs
jjbayer: > it would be good if changes to the schema files were picked up immediately in local development without recompiling.

I agree, I was thinking more of an in-memory cache that picks up the changes when you restart Relay (or re-run a test). I see your argument against putting such a cache in the library, but I still think it makes sense to put it _somewhere_ (e.g. the `KafkaClient`). I'm happy to contribute to this PR for this. Or do you see a use case for an actual hot reload, where you keep a local Relay instance running and edit the schema while sending in events?

jjbayer: > I'm happy to add a get_latest_schema helper

Only if you think it is a good idea.

> how strongly do you feel about this happening?

Not strongly at all, I just noticed I had to jump into the library code to see what the `None` was for.

relay-kafka/Cargo.toml
jjbayer: OK! Definitely not a blocker.

> As such it doesn't have many dependencies. This is a good thing if you're only using the generated types in your service.

Makes sense, though this could be solved with cargo features. I.e. there could be a `codegen` feature and a `validation` feature, and clients use either-or. My gut feeling is that 90% of clients that want to perform validation will not care about what schema definition language or validation library is used under the hood, hence the idea to hide it inside the library.

untitaker: i'm reconsidering my response now. one of the reasons i was hesitant to change this was because python currently also works the same (it just exposes the raw schema, doesn't prescribe a schema validation library)

but at the same time it would be good to have a vetted recommended library that everybody ought to use (for both python and rust)

untitaker:
@jjbayer after deliberating back and forth I believe we should defer some of the API decisions until we have more feedback from other teams. For that purpose I have summarized the discussion about API here: https://github.com/getsentry/sentry-kafka-schemas/issues/105

Let me know if I missed anything, for now I would like to merge this PR as-is

jjbayer (approved):

untitaker:
I believe a recent change in sentry-kafka-schemas has made it so that InvalidSchema is returned for every call to get_schema. Need to investigate, but this PR is broken right now (it won't validate anything at all)

untitaker:
yikes! https://github.com/getsentry/sentry-kafka-schemas/pull/107

untitaker:
I added `LogError` somewhere so that a schema validation error for metrics is actually printed to stdout/err

PR #2016, opened 2023-04-06
feat(buffering): Enable auto-vacuum and fallback to cache.envelope_buffer_size if spool is full

olksdr:
This change makes sure we fall back to `cache.envelope_buffer_size` if the disk stays full, but also enabling auto-vacuum let us reclaim un-used space be able to spool more. 

Also refactored some of the function and switched to early returns. 

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_buffer.rs
+++ b/relay-server/src/actors/project_buffer.rs
@@ -14,7 +14,7 @@ use sqlx::sqlite::{
     SqliteConnectOptions, SqliteJournalMode, SqlitePoolOptions, SqliteRow, SqliteSynchronous,
 };
 use sqlx::{Pool, QueryBuilder, Row, Sqlite};
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, Semaphore};
 
 use crate::actors::project_cache::{ProjectCache, UpdateBufferIndex};
 use crate::envelope::{Envelope, EnvelopeError};
@@ -31,9 +31,6 @@ const SQLITE_LIMIT_VARIABLE_NUMBER: usize = 999;
 /// The set of errors which can happend while working the the buffer.
 #[derive(Debug, thiserror::Error)]
 pub enum BufferError {
-    #[error("failed to store the envelope in the buffer, max size {0} reached")]
-    Full(u64),
-
     #[error("failed to move envelope from disk to memory")]
     CapacityExceeded(#[from] crate::utils::BufferError),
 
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/project_buffer.rs
```diff
@@ -223,7 +225,12 @@ impl BufferService {
 
         // Only if persistent envelopes buffer file path provided, we create the pool and set the config.
         if let Some(path) = config.spool_envelopes_path() {
-            relay_log::info!("Using the buffer file: {}", path.to_string_lossy());
+            relay_log::info!(
+                "Using the buffer file {}, max memory size {}, max disk size {}",
```
jan-auer: Let's make these separate info logs.

relay-server/src/actors/project_buffer.rs
```diff
@@ -242,6 +249,12 @@ impl BufferService {
                 // When synchronous is NORMAL, the SQLite database engine will still sync at the most critical moments, but less often than in FULL mode.
                 // Which guarantees good balance between safety and speed.
                 .synchronous(SqliteSynchronous::Normal)
+                // The freelist pages are moved to the end of the database file and the database file is truncated to remove the freelist pages at every
+                // transaction commit. Note, however, that auto-vacuum only truncates the freelist pages from the file.
+                // Auto-vacuum does not defragment the database nor repack individual database pages the way that the VACUUM command does.
+                //
+                // This will helps us to keep the file size under some controll.
```
jan-auer:
```suggestion
                // This will helps us to keep the file size under some control.
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -402,6 +416,18 @@ impl BufferService {
...
+        // The auto-vacuum will kick in on each delete from the database, and the spool status will
+        // be checked and we more space is available the `is_disk_full` flag set to `false`.
+        //
+        // It falls back to the `cache.envelope_buffer_size`, and will try to keep the rest
+        // of the incoming envelopes in the memory till the `BufferGuard` is exhausted and it
+        // starts dropping the envelopes.
+        let is_disk_full = self.spool_config.as_ref().map_or(true, |s| s.is_disk_full);
+        if !is_disk_full {
+            self.try_spool().await?;
+        }
```
jan-auer:
```suggestion
        if let Some(ref config) = self.spool_config {
            if !config.is_disk_full {
                self.try_spool().await?;
            }
        }
```

jan-auer (approved):

PR #2017, opened 2023-04-07
ref(server): Log sentry errors for web handler panics

jan-auer:
Panics in web endpoint handlers are caught by a middleware and transformed into
a 500 Internal Server Error. With this PR, we additionally log a sentry error
with stack trace.

#skip-changelog


olksdr (approved):
🚀

PR #2018, opened 2023-04-07
feat(protocol): Change lock_reason to a held_locks dictionary

romtsn:
A thread can hold multiple locks, so this PR changes the recently introduced `lock_reason` attribute to a dictionary/map, where the key is the lock object address and the value is the existing `LockReason` struct. I was also thinking to make it just an Array, but I think it'd be easier and faster to look up by a lock address when we analyze.

Additionally, I haven't found any docs about how to update snapshot files, so I've added a section to readme about that, lemme know if that makes sense.

https://cs.android.com/android/platform/superproject/+/master:development/tools/bugreport/src/com/android/bugreport/stacks/ThreadSnapshot.java;l=53


romtsn:
> Additionally, I haven't found any docs about how to update snapshot files, so I've added a section to readme about that, lemme know if that makes sense.

Ah, ok, it's here https://develop.sentry.dev/rust/#development-environment, should I remove it from the readme then?

olksdr (approved):
lgtm!

shruthilayaj (approved):

relay-general/src/protocol/thread.rs
```diff
@@ -233,9 +233,10 @@ pub struct Thread {
     #[metastructure(skip_serialization = "empty")]
     pub state: Annotated<String>,
 
-    /// Represents an instance of a held lock (java monitor object) in a thread.
-    #[metastructure(skip_serialization = "empty")]
-    pub lock_reason: Annotated<LockReason>,
+    /// Represents a collection of locks (java monitor objects) held by a thread.
+    ///
+    /// A map of lock object addresses and their respective lock reason/details.
+    pub held_locks: Annotated<Object<LockReason>>,
```
iker-barriocanal, line 239: Consider any backward-compatibility if there already are SDKs released with `lock_reason` in the protocol.

iker-barriocanal (approved):

relay-general/src/protocol/thread.rs
romtsn: no, not yet, so we're good here. but thanks for the reminder!

PR #2019, opened 2023-04-07
ref(buffering): Use estimated envelope size to calculate the buffer mem usage

olksdr:
This change adds the estimated envelope size to the managed envelope and use it to calculate and limit the buffer memory usage.

#skip-changelog


relay-server/src/utils/rate_limits.rs
```diff
@@ -142,6 +142,9 @@ pub struct EnvelopeSummary {
 
     /// Whether the envelope contains an event which already had the metrics extracted.
     pub event_metrics_extracted: bool,
+
+    /// The payload size of this envelope.
```
iker-barriocanal, line 146: nit:
```suggestion
    /// The estimated payload size of this envelope.
```
?

relay-server/src/utils/managed_envelope.rs
```diff
@@ -305,6 +306,18 @@ impl ManagedEnvelope {
...
     }
 
+    /// Returns estimated size of this envelope.
+    ///
+    /// This is just an estimated size, which in reality can be somewhat bigger, depending on the
+    /// list of additional attributes allocated on all of the inner types.
+    pub fn estimated_size(&self) -> usize {
+        // Always round it up to next 100 bytes.
+        (f64::ceil(
+            (self.context.summary.payload_size + size_of::<Self>() + size_of::<Envelope>()) as f64
```
iker-barriocanal, line 316: Isn't `EnvelopeContext` what you want to add here? The size of the envelope is already computed as part of the `context.summary.payload_size`, so IIUC we miss the rest of the type?
```suggestion
            (self.context.summary.payload_size + size_of::<Self>() + size_of::<EnvelopeContext>()) as f64
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -175,7 +175,7 @@ impl FromMessage<RemoveMany> for Buffer {
 struct BufferSpoolConfig {
     db: Pool<Sqlite>,
     max_disk_size: usize,
-    max_envelopes_count: usize,
+    max_memory_size: usize,
```
iker-barriocanal, line 178: Should we add any docs about the units used here?

iker-barriocanal (changes requested):
The PR looks good overall, see the comment on the estimation of the memory used.

relay-server/src/utils/managed_envelope.rs
olksdr: Nope. 
The `EnvelopeContext` should already in included in `Self` at leat all the stack allocated portion, and since the `Envelope` is in the `Box`, we count only the pointer size but not the stack fields. That's why additionally to this I add the envelope here as well. 
Same goes about payload size from summary , which calculated for every time in the envelope. That's also in practice often escapes to heap (I assume 😄 from what I can see). That why added here payload size as external field summary.

relay-server/src/actors/project_buffer.rs
olksdr: I can take the doc part in my next PR.

relay-server/src/utils/managed_envelope.rs
iker-barriocanal: Ohh I see, that makes sense!

iker-barriocanal (approved):
Nice!

relay-server/src/utils/rate_limits.rs
jjbayer: I would even add a note here that this number excludes the cost of item headers and the envelope headers.

jjbayer (commented):
How far off is the estimated number of bytes? Should we add a unit test that compares the actual serialized size of a realistic payload to the estimated size?

PR #2020, opened 2023-04-07
feat(protocol) Add new `mechanism` fields to protocol to support exception groups

mattjohnsonpint:
Adds 4 new fields to the `mechanism` protocol to support exception groups, per [Sentry RFC 79](https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md)

### DIFF
--- a/relay-general/src/protocol/mechanism.rs
+++ b/relay-general/src/protocol/mechanism.rs
@@ -155,7 +155,12 @@ pub struct Mechanism {
     /// - .NET Examples:  `"InnerException"`, `"InnerExceptions[0]"`, `"InnerExceptions[1]"`
     ///
     /// - JavaScript Examples: `"cause"`, `"errors[0]"`, `"errors[1]"`
-    #[metastructure(required = "false", nonempty = "true", max_chars = "enumlike")]
+    #[metastructure(
+        required = "false",
+        nonempty = "true",
+        max_chars = "enumlike",
+        deny_chars = " \t\r\n"
+    )]
     pub source: Annotated<String>,
 
     /// An optional boolean value, set `true` when the exception is the platform-specific exception
--- a/relay-general/src/protocol/mechanism.rs
+++ b/relay-general/src/protocol/mechanism.rs
@@ -144,6 +144,43 @@ pub struct Mechanism {
     ///   as the user explicitly captured the exception (and therefore kind of handled it)
     pub handled: Annotated<bool>,
 
+    /// An optional string value describing the source of the exception.
+    ///
+    /// For chained exceptions, this should contain the platform-specific name of the property or
+    /// attribute (on the parent exception) that this exception was acquired from. In the case of
+    /// an array, it should include the zero-based array index as well.
+    ///
+    /// - Python Examples: `"__context__"`, `"__cause__"`, `"exceptions[0]"`, `"exceptions[1]"`
+    ///
+    /// - .NET Examples:  `"InnerException"`, `"InnerExceptions[0]"`, `"InnerExceptions[1]"`
+    ///
+    /// - JavaScript Examples: `"cause"`, `"errors[0]"`, `"errors[1]"`
+    #[metastructure(required = "false", nonempty = "true", max_chars = "enumlike")]
+    pub source: Annotated<String>,
+
+    /// An optional boolean value, set `true` when the exception is the platform-specific exception
+    /// group type.  Defaults to `false`.
+    ///
+    /// For example, exceptions of type `ExceptionGroup` (Python), `AggregateException` (.NET), and
+    /// `AggregateError` (JavaScript) should have `"is_exception_group": true`.  Other exceptions
+    /// can omit this field.
+    pub is_exception_group: Annotated<bool>,
+
+    /// An optional numeric value providing an ID for the exception relative to this specific event.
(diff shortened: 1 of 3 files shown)


relay-general/src/protocol/mechanism.rs
```diff
@@ -144,6 +144,43 @@ pub struct Mechanism {
...
+    /// can omit this field.
+    pub is_exception_group: Annotated<bool>,
+
+    /// An optional numeric value providing an ID for the exception relative to this specific event.
+    /// It is referenced by the `parent_id` to reconstruct the logical tree of exceptions in an
+    /// exception group.
+    ///
+    /// This should contain an unsigned integer value starting with `0` for the last exception in
+    /// the exception values list, then `1` for the previous exception, etc.
+    pub exception_id: Annotated<u64>,
```
mattjohnsonpint, line 180: Question: I chose `Annotated<u64>` because I didn't see any other places where we use smaller sized integers.  Is that for compatibility reasons? Or just by convention?

For all practical purposes, this could be a `u16` (or even a `u8`).   Does it matter?

relay-general/tests/snapshots/test_fixtures__event_schema.snap
```diff
@@ -2222,6 +2232,14 @@ expression: "relay_general::protocol::event_json_schema()"
                 "null"
               ]
             },
+            "is_exception_group": {
+              "description": " An optional boolean value, set `true` when the exception is the platform-specific exception\n group type.  Defaults to `false`.\n\n For example, exceptions of type `ExceptionGroup` (Python), `AggregateException` (.NET), and\n `AggregateError` (JavaScript) should have `\"is_exception_group\": true`.  Other exceptions\n can omit this field.",
+              "default": null,
```
mattjohnsonpint, line 2240: Question: Is there a way to make the default `false`?  This really isn't a three-state value.  It's either `true` or `false`.  If omitted, should be treated as `false`.

relay-general/src/protocol/mechanism.rs
```diff
@@ -144,6 +144,43 @@ pub struct Mechanism {
...
+    /// For chained exceptions, this should contain the platform-specific name of the property or
+    /// attribute (on the parent exception) that this exception was acquired from. In the case of
+    /// an array, it should include the zero-based array index as well.
+    ///
+    /// - Python Examples: `"__context__"`, `"__cause__"`, `"exceptions[0]"`, `"exceptions[1]"`
+    ///
+    /// - .NET Examples:  `"InnerException"`, `"InnerExceptions[0]"`, `"InnerExceptions[1]"`
+    ///
+    /// - JavaScript Examples: `"cause"`, `"errors[0]"`, `"errors[1]"`
+    #[metastructure(required = "false", nonempty = "true", max_chars = "enumlike")]
```
olksdr: This will give you `128` chars. Is this enough for your use-case?

olksdr (approved):
lgtm

relay-general/tests/snapshots/test_fixtures__event_schema.snap
olksdr: I do not think it's possible to set a `default` on the field at the moment. But in code you can treat empty (or `null`) field as `false`.

iker-barriocanal: I believe the event schema has some flaws, so I wouldn't worry about it too much anyway.

relay-general/src/protocol/mechanism.rs
iker-barriocanal: Let's also limit the allowed characters for strings, see [example here](https://github.com/getsentry/relay/blob/3673a76f8efcb9827ce4e7d1fec60a36aab6c5ba/relay-general/src/protocol/tags.rs#L7).

relay-general/src/protocol/mechanism.rs
iker-barriocanal: I think `u64` is fine. I remember having a conversation similar to this around 32 bits, but I don't remember exactly what the conclusions were, it's possible that they were around compatibility with other librarires.

iker-barriocanal (approved):
Thanks for contributing and putting the effort to write these amazing docs! Approving to not keeping you blocked, but please address the comment about limiting the character set in the string.

relay-general/src/protocol/mechanism.rs
mattjohnsonpint: Yes, that's plenty.  Thanks.

mattjohnsonpint: I added `deny_chars = " \t\r\n"` to prevent whitespace.  I don't want to specifically list allowable chars, because different languages have different conventions and it could be problematic to try to include all possible values.

iker-barriocanal: That makes sense. To prevent whitespace, it may be more readable to add `trim_whitespace` (and `nonempty`?), see [example](https://github.com/getsentry/relay/blob/ab2a3dca7a3daaecfcdf097d898c766fd6b69476/relay-general/src/protocol/event.rs#L383-L385).

mattjohnsonpint: Trimming wouldn't catch whitespace in the middle of a string.

PR #2022, opened 2023-04-08
ref(registry): Remove global service registry

### LINKED ISSUE #1610: Fix tests depending on broken `SystemRegistry` behavior
```[tasklist]
### List of failing tests
- [ ] https://github.com/getsentry/relay/blob/1d57db414a11af96348747c307d40a7852018142/relay-server/src/actors/processor.rs#L2941
```

olksdr:
This PR finalizes migration to service dependencies injection instead of using the global registry.

In this part the changes mostly located in the endpoints, and their dependencies.

This also allowed to uncomment few `ignored` tests as they are no longer depend on the global registry or all the services. 

related to: https://github.com/getsentry/relay/issues/1818
fixes: https://github.com/getsentry/relay/issues/1610

### DIFF
--- a/relay-server/src/actors/project_cache.rs
+++ b/relay-server/src/actors/project_cache.rs
@@ -23,9 +23,9 @@ use crate::actors::project_local::{LocalProjectSource, LocalProjectSourceService
 #[cfg(feature = "processing")]
 use crate::actors::project_redis::RedisProjectSource;
 use crate::actors::project_upstream::{UpstreamProjectSource, UpstreamProjectSourceService};
+use crate::actors::test_store::TestStore;
 use crate::actors::upstream::UpstreamRelay;
 
-use crate::service::REGISTRY;
 use crate::statsd::{RelayCounters, RelayGauges, RelayHistograms, RelayTimers};
 use crate::utils::{self, BufferGuard, GarbageDisposal, ManagedEnvelope};
 
@@ -210,12 +210,6 @@ pub enum ProjectCache {
     MergeBuckets(MergeBuckets),
     FlushBuckets(FlushBuckets),
     UpdateBufferIndex(UpdateBufferIndex),
-}
-
-impl ProjectCache {
-    pub fn from_registry() -> Addr<Self> {
-        REGISTRY.get().unwrap().project_cache.clone()
-    }
 }
 
 impl Interface for ProjectCache {}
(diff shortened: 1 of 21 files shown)


relay-server/src/extractors/signed_json.rs
```diff
@@ -77,7 +78,7 @@ where
     B: axum::body::HttpBody + Send + 'static,
     B::Data: Send,
     B::Error: Into<axum::BoxError>,
-    S: Send + Sync,
+    S: Send + Sync + ServiceRegistry,
```
jan-auer: Instead of this trait, you can make this non-generic and require `ServiceState`. This is also what we do in a few other extractors.

relay-server/src/service.rs
```diff
@@ -87,6 +90,7 @@ pub fn create_runtime(name: &str, threads: usize) -> Runtime {
 pub struct ServiceState {
     config: Arc<Config>,
```
jan-auer, line 84: `ServiceState` is cloned very often by axum, and also from us. Cloning it is somewhat expensive now, since we clone a lot of fields: All the runtimes, and then all fields in the registry. It would now make sense to change the internals of `ServiceState` to something like:


```rust
#[derive(Clone, Debug)]
struct ServiceState {
    inner: Arc<StateInner>,
}

impl ServiceState {
    // accessor methods for every service here, accessing through `self.inner.*`
}

#[derive(Debug)] // no Clone!
struct StateInner {
    // all Addrs here
    // all runtimes here
    // buffer guard
}
```

relay-server/src/actors/project_cache.rs
```diff
@@ -819,12 +820,14 @@ impl Service for ProjectCacheService {
...
         let Self {
+            buffer_guard,
             config,
             services,
             redis,
         } = self;
-        let buffer_guard = services.buffer.clone();
         let project_cache = services.project_cache.clone();
+        let outcome_aggregator = services.outcome_aggregator.clone();
+        let test_store = services.test_store.clone();
```
iker-barriocanal, line 830: I see that we clone all the services. Is it to avoid issues with a mutable reference? Or, why are we doing that?

relay-server/src/actors/project_cache.rs
```diff
@@ -835,16 +838,23 @@ impl Service for ProjectCacheService {
...
-                    Ok(buffer) => buffer.start(),
-                    Err(err) => {
-                        relay_log::error!("failed to start buffer service: {}", LogError(&err));
-                        // NOTE: The process will exit with error if the buffer file could not be
-                        // opened or the migrations could not be run.
-                        std::process::exit(1);
-                    }
-                };
+            let buffer = match BufferService::create(
+                buffer_guard,
```
iker-barriocanal, line 842: Not related to this PR, but I wonder if we could somehow abstract the buffer guard and make it part of the buffer service?

iker-barriocanal (changes requested):
Out of scope for this PR, but `test_store` seems to me that it's related to tests. For this approach the PR looks good, a couple of questions to discuss wrt to the approach.

relay-server/src/actors/project_cache.rs
olksdr: This is still in followups: https://github.com/getsentry/team-ingest/issues/53

relay-server/src/actors/project_cache.rs
olksdr: The clone is very cheap here and in fact it's all the same objects, which just contains the inner rx channel and the counter for number of messages in the queue. This way each service can own the addresses of the services its communicate with.

relay-server/src/actors/project_cache.rs
iker-barriocanal: Oh yeah my bad, I completely forgot we had that.

iker-barriocanal (approved):

PR #2023, opened 2023-04-11
fix(processing): Set start time on envelope meta when it's deserialized from spool entry.

olksdr:
This adds support for processing time for spooled envelopes:
* preserve the current processing time, using `start_time` field on envelope meta
* use the record to restore it once we get the envelope from the persistent spool

#skip-changelog


relay-server/src/actors/project_buffer.rs
```diff
@@ -302,6 +306,24 @@ impl BufferService {
...
+    fn now() -> Duration {
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+    }
+
+    /// Returns the timestamp in form of fixed byte array.
+    ///
+    /// Timestamp represents the point in time before the provided duration.
+    fn timestamp(elapsed: Duration) -> Option<[u8; 16]> {
```
iker-barriocanal: `timestamp` is confusing as the function name. I can't think of names I really like, but I think `then_timestamp` is a better choice.

relay-server/src/actors/project_buffer.rs
```diff
@@ -302,6 +306,24 @@ impl BufferService {
...
+    /// Returns duration since UNIX_EPOCH.
+    fn now() -> Duration {
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+    }
+
+    /// Returns the timestamp in form of fixed byte array.
+    ///
+    /// Timestamp represents the point in time before the provided duration.
```
iker-barriocanal: I think this is easier to understand:
```suggestion
    /// Returns the timestamp before the provided duration, in form of fixed byte array.
```

migrations/20230311081917_envelopes_table.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE IF NOT EXISTS envelopes (
   id              INTEGER PRIMARY KEY AUTOINCREMENT,
+  timestamp       BLOB,
```
iker-barriocanal: The `timestamp` name is confusing since it doesn't tell us what this timestamp is referring to, or what this time means/represents.

iker-barriocanal (changes requested):
The dance we're doing parsing multiple times the timestamp is quite complex, I think more than necessary. Why not store the `start_time` directly? See [date and time data types](https://www.sqlite.org/datatype3.html#date_and_time_datatype), and [date and time functions](https://www.sqlite.org/lang_datefunc.html) in SQLite. If we're changing the approach, you may want to ignore my comments below.

olksdr:
@iker-barriocanal addressed you comments in the last commit, PTAL!

relay-server/src/actors/project_buffer.rs
```diff
@@ -392,7 +401,25 @@ impl BufferService {
...
+            .unwrap_or_default()
+            .checked_sub(received_at)
+            .unwrap_or_default();
+
+        let start_time = Instant::now()
+            // Subtract the elapsed time (the time the envelope was kept in the spool) from the current
+            // Instant to get the timestamp when the request started.
+            .checked_sub(elapsed)
+            // If we fail to get the instant from the timestamp, we fallback to `now()`.
+            .unwrap_or_else(Instant::now);
```
iker-barriocanal: Why are we shifting the `start_time`? I see you added it to the PR description but I don't know the context on why we need it:

> preserve the current processing time, using start_time field on envelope meta

From the docs of `start_time`:

> The time at which the request started.

If we shift this field, it'll no longer hold and may be quite confusing in the future when we forget a bit about this.

I assume that this hack is the fastest way to keep the total processing time of an envelope correct and ship this fast. If this is the case and we want to move forward with it, let's add a comment to fix this in `start_time`'s docs and create a new follow-up item in the [buffering meta issue](https://github.com/getsentry/team-ingest/issues/53). I'm not blocking the PR because of this, but let's make sure we address this sooner than later.

migrations/20230311081917_envelopes_table.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE IF NOT EXISTS envelopes (
   id              INTEGER PRIMARY KEY AUTOINCREMENT,
+  received_at     INTEGER,
```
iker-barriocanal: SQLite's docs on [date and time datatypes](https://www.sqlite.org/datatype3.html#date_and_time_datatype) mention that the integer is expected to represent the number of seconds. This now represents milliseconds. We either change the format in the database to represent milliseconds or if we go with a bit more dirty solution I'd say to rename the column to `received_at_ms`. What do you think?

relay-server/src/actors/project_buffer.rs
```diff
@@ -392,7 +401,25 @@ impl BufferService {
...
+
+        envelope.set_start_time(start_time);
```
iker-barriocanal: If I understand this math correctly, `received_at` and `start_time` is the same, and the calculations on `elapsed` don't perform any changes. The only difference may be the time it takes from `SystemTime::now` to `Instant::now`, so a few nanoseconds? This results on the same thing:

```suggestion
        let received_at = Duration::from_millis(received_at as u64);
        envelope.set_start_time(received_at);
```

iker-barriocanal (changes requested):
Could we add some tests to verify this is working as intended?

The PR title (and the changelog, when applicable) already describe _what_ the PR does, but we don't know _why_ we are doing it (both the goal the PR is trying to accomplish and the approach we went with). Reading these reasons make understanding, reviewing, and reminding (in the future) easier.

migrations/20230311081917_envelopes_table.sql
olksdr: As the docs state the SQLite does not have notion of datetime types in it. 
> SQLite does not have a storage class set aside for storing dates and/or times.
Meaning users basically free to use whatever they want. 

In our case the `INTEGER` is an `INTEGER` nothing else, and nothing more, which we happened to use it for storing the timestamp in the form of `i64`, which we then get back. Database in this case does not care what the number in there  and what it represents. 

Using milliseconds gives us better precision and I do not see the point to put that in the name of the field, which feels very wrong.

relay-server/src/actors/project_buffer.rs
olksdr: I do not really understand this comment. Could you, please, rephrase it? 

What's happening here is that we preserve the time the envelopes came in when we spool it and then restore it when we deserialize the envelope back for processing. 

What do you mean by hack? At this point this is the working solution, which might be addressed later in the followup refactors, if the underlying structures (e.g. `Envelope` or `ManagedEnvelope`, etc.) are changed.

relay-server/src/actors/project_buffer.rs
olksdr: Calculate the `received_at` to set the `start_time` is the entire point of it - in order to restore the time when the request started. Exactly what the docs say.

olksdr (commented):
There are also more docs on [`Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html) in order to understand why the conversion between `SystemTime` and `Instant` is needed and what the `Instant` is - it can help to understand how the `start_time` field is used in our code. 

> Could we add some tests to verify this is working as intended?

There is a integration test , which uses this code, but I also will have a look into adding a unit test. 

> The PR title (and the changelog, when applicable) already describe what the PR does, but we don't know why we are doing it (both the goal the PR is trying to accomplish and the approach we went with).

The PR description also kind of implies the `why`:

to support the processing time for the envelope, which is broken with persistent spooling , and has to be fixed by restoring the correct `start_time` of the request

migrations/20230311081917_envelopes_table.sql
iker-barriocanal: Yeah that makes sense 👍

iker-barriocanal (approved):
LGTM! I was wrong in my previous review so ignore the previous comments. Good stuff!

migrations/20230311081917_envelopes_table.sql
jjbayer:
```suggestion
  received_at     INTEGER,  -- milliseconds since epoch
```

relay-server/src/actors/project_buffer.rs
```diff
@@ -395,7 +404,13 @@ impl BufferService {
     fn extract_envelope(&self, row: SqliteRow) -> Result<ManagedEnvelope, BufferError> {
         let envelope_row: Vec<u8> = row.try_get("envelope")?;
         let envelope_bytes = bytes::Bytes::from(envelope_row);
-        let envelope = Envelope::parse_bytes(envelope_bytes)?;
+        let mut envelope = Envelope::parse_bytes(envelope_bytes)?;
+
+        let received_at: i64 = row.try_get("received_at")?;
+        let start_time = StartTime::from_timestmap(received_at as u64);
```
```suggestion
        let start_time = StartTime::from_timestamp(received_at as u64);
```

relay-server/src/extractors/start_time.rs
```diff
@@ -19,6 +19,26 @@ impl StartTime {
     pub fn into_inner(self) -> Instant {
         self.0
     }
+
+    /// Returns the [`StartTime`] corresponding to provided timestamp.
+    pub fn from_timestmap(timestamp: u64) -> Self {
```
jjbayer:
```suggestion
    pub fn from_timestamp(timestamp: u64) -> Self {
```

relay-server/src/extractors/start_time.rs
```diff
@@ -19,6 +19,26 @@ impl StartTime {
...
+
+        let start_time = Instant::now()
+            // Subtract the elapsed time from the current `Instant` to get the timestamp
+            // of the start time.
+            .checked_sub(elapsed)
+            // If we fail to get the `Instant` from the timestamp, we fallback to `now()`.
+            .unwrap_or_else(Instant::now);
+
+        Self(start_time)
+    }
```
jjbayer, line 41: Instead of this function, can we use `UnixTimestamp::to_instant`?

https://github.com/getsentry/relay/blob/e5d7d5b64d9a46897729b2bac500713310f19bc0/relay-common/src/time.rs#L155

jjbayer (changes requested):
Looks good in general! One open question about code reuse.

relay-server/src/extractors/start_time.rs
olksdr: This will introduce another layer of indirection:
* this wrapper assumes the internal timestamp in seconds, which will have to be fixed
* this also will require more additions to the `impl` of `UnixTimestamp` than this PR already makes

I can work on followup to improve `UnixTimestamp` to handle better precision and then use this wrapper inside of this function (or instead).  What do you think?

jjbayer: Ah right, `UnixTimestamp` counts seconds. Never mind.

jjbayer (approved):

PR #2025, opened 2023-04-11
feat(profiling): Add a data category for indexed profiles

phacops:
We're adding a new data category for indexed profiles in preparation of count processed profiles.


relay-common/src/constants.rs
```diff
@@ -107,6 +107,9 @@ pub enum DataCategory {
     /// Session updates. Quantity is the number of updates in the batch.
     Session = 5,
     /// A profile
+    ///
+    /// This is the category for profiles sent alongside transactions but the transaction was
+    /// dropped.
     Profile = 6,
```
maheskett: @phacops This category should be for all profiles, regardless of whether the transaction was sampled. Also, if a profile is rate limited there should be a rate limited outcome for `Profile` even if the transaction is accepted.

jjbayer (changes requested):
> if the transaction is processed only, emit an outcome for a processed profile
> if the transaction is indexed, an indexed profile outcome will be emitted later in the pipeline

@phacops the reason why we have two different data categories for transactions is that they can make their way to storage in two different shapes: either as raw transactions (when the sampling decision is "keep") or as metrics (when the sampling decision is "drop").

But for profiles, I assume we are not going to extract metrics when the sampling decision is "drop", or are we?

If not, I wonder if instead of introducing a new data category, it would make more sense to keep a single data category, but make sure that [an outcome with reason `Sampled`](https://github.com/getsentry/relay/blob/751bc4b2f4e3c6fc7373d8ca7cddf0812f68dbbc/relay-server/src/actors/outcome.rs#L160-L161) is generated for every profile that is dropped by dynamic sampling.

So instead of 

```
processed profiles = profiles[outcome:accepted] + profiles_indexed[outcome:accepted]
```

you would have

```
processed profiles = profiles[outcome:accepted] + profiles[outcome:sampled]
```

With that we should be able to calculate all stats and billing correctly.

phacops:
Going into a different direction.

PR #2026, opened 2023-04-12
feat(errors): Add flag to error when trace is full

### LINKED ISSUE #48773: Start flagging errors with traces in Relay
We would like to flag incoming errors when they have full traces connected to them.

iambriccardo:
This PR implements the flagging of the error "having a full trace" when the `trace` rules return a `keep` sampling decision given the error's DSC.

To achieve this, the PR adds an `Option<Event>` to the sampling algorithm.

Closes https://github.com/getsentry/sentry/issues/48773


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 **Internal**:
 
 - Add `txNameReady` flag to project config. ([#2128](https://github.com/getsentry/relay/pull/2128))
+- Tag incoming errors with the new `sampled` context in case their DSC is sampled. ([#2026](https://github.com/getsentry/relay/pull/2026))
```
```suggestion
- Tag incoming errors with the new `sampled` field in case their DSC is sampled. ([#2026](https://github.com/getsentry/relay/pull/2026))
```

relay-general/src/protocol/contexts/trace.rs
```diff
@@ -106,6 +106,11 @@ pub struct TraceContext {
     #[metastructure(max_chars = "enumlike", allow_chars = "a-zA-Z0-9_.")]
     pub origin: Annotated<OriginType>,
 
+    /// Track whether the trace connected to this event has been sampled entirely.
+    ///
+    /// This flag only applies to events with [`Error`] type.
+    pub sampled: Annotated<bool>,
```
jjbayer: I never know whether `sampled` means "drop" or "keep". Can we make this `sampling_decision: "drop" | "keep"` instead? Maybe even put the serialized `SamplingResult` here?

relay-server/src/actors/processor.rs
```diff
@@ -1994,8 +1997,8 @@ impl EnvelopeProcessorService {
 
         // The DSC can only be computed if there's a transaction event. Note that `from_transaction`
         // below already checks for the event type.
-        let Some(event) = state.event.value() else { return };
-        let Some(key_config) = state.project_state.get_public_key_config() else { return };
+        let Some(event) = state.event.value() else { return; };
+        let Some(key_config) = state.project_state.get_public_key_config() else { return; };
```
jjbayer: Why?

jjbayer (changes requested):
I might not have the full picture, but this PR refactors a lot of code where I'm not sure it's necessary. Since we already have a two step approach of 1. computing the sampling decision and 2. acting upon the sampling decision (`sample_envelope`), could we structure the code something like this?

```mermaid
flowchart TD
    compute_sampling_decision -->|SampleResult| t

    subgraph sample_envelope
      t{{event_type}}
      t -->|transaction| x[drop envelope]
      t -->|error| m[annotate event]
    end
```

relay-server/src/actors/processor.rs
iambriccardo: I am confused, I didn't add this, maybe I have use the reformat code from PyCharm.

relay-sampling/src/lib.rs
```diff
@@ -993,8 +995,15 @@ pub fn merge_configs_and_match(
...
-    // We also decide to use the sampling mode of the project to which the event belongs.
-    match_result.set_sample_rate(match sampling_config.mode {
+    // The determination of the sampling mode occurs with the following priority:
+    // 1. Non-root project sampling mode
+    // 2. Root project sampling mode
+    // 3. SamplingMode::Unsupported
+    let sampling_mode = sampling_config.map_or(
+        root_sampling_config.map_or(SamplingMode::Unsupported, |config| config.mode),
+        |config| config.mode,
+    );
```
jjbayer: I would make this failure more explicit:

```suggestion
    let Some(primary_config) = sampling_config.or(root_sampling_config) else {
        relay_log::error!("Cannot sample without config");
        return None;
    }
    let sampling_mode = primary_config.sampling_mode;
```

relay-sampling/src/lib.rs
```diff
@@ -1072,19 +1082,23 @@ impl SamplingMatch {
         // 3. /transaction is matched with a transaction rule with a factor of 4 and uses as seed abc -> 0.2 * 4 = 0.8 sample rate
         //
         // We can see that we have 3 different samples rates but given the same seed, the random number generated will be the same.
-        let mut seed = event.id.value().map(|id| id.0);
+        let mut seed = event
+            .map_or(Annotated::empty(), |event| event.id.clone())
+            .value()
+            .map(|id| id.0);
```
```suggestion
        let mut seed = event.and_then(|e| e.id.value()).map(|id| id.0);
```

relay-sampling/src/lib.rs
```diff
@@ -1072,19 +1082,23 @@ impl SamplingMatch {
...
 
+        let event_ty = event.and_then(|event| event.ty.value());
         for rule in rules {
             let matches = match rule.ty {
                 RuleType::Trace => match dsc {
                     Some(dsc) => rule.condition.matches(dsc, ip_addr),
                     _ => false,
                 },
-                RuleType::Transaction => match event.ty.0 {
+                RuleType::Transaction => event.map_or(false, |event| match event_ty {
```
jjbayer: Do you need `event_ty` somewhere else? If not, this should work:

```suggestion
                RuleType::Transaction => event.map_or(false, |event| match event.py {
```

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -53,6 +79,31 @@ fn get_sampling_match(
...
+        root_project_state,
+        dsc,
+        None,
+        ip_addr,
+        // For consistency reasons we take a snapshot in time and use that time across all code that
+        // requires it.
+        Utc::now(),
+    );
+    SamplingResult::determine_from_sampling_match(sampling_result)
+}
```
jjbayer: Regardless of what we discussed today, I would remove this function and just adapt the interface of `should_keep_event` (and inline `determine_from_sampling_match` again) to reduce code duplication. We should maybe rename `should_keep_event` to `get_sampling_decision` or similar now that it has become so generic.

relay-general/src/protocol/contexts/trace.rs
jjbayer: Feel free to ignore, a boolean might be easier to work with. It just needs to be properly documented.

jjbayer (changes requested):
Looks good in general! Some minor comments, and I did not yet review the tests.

relay-sampling/src/lib.rs
iambriccardo: Damn, 10x more expressive, I love it!

relay-sampling/src/lib.rs
iambriccardo: Overlooked this, thanks for the heads up!

py/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.8.24
+
+- Add optional event to respect signature of the `get_sampling_result` method. ([#2026](https://github.com/getsentry/relay/pull/2026))
```
jjbayer: I would remove this changelog entry, because the observable behavior of the exposed python function does not change (or does it?). If we keep it, please move it to `## Unreleased`.

relay-general/src/protocol/contexts/trace.rs
```diff
@@ -106,6 +106,11 @@ pub struct TraceContext {
     #[metastructure(max_chars = "enumlike", allow_chars = "a-zA-Z0-9_.")]
     pub origin: Annotated<OriginType>,
 
+    /// Track whether the trace connected to this event has been sampled entirely.
+    ///
+    /// This flag only applies to events with [`Error`] type.
```
jjbayer:
```suggestion
    /// This flag only applies to events with [`Error`] type that have an associated dynamic sampling context.
```

relay-sampling/src/lib.rs
```diff
@@ -54,13 +54,13 @@
 //! ## [`SamplingConfig`]
 //!
 //! ```json
-#![doc = include_str!("../tests/fixtures/sampling_config.json")]
+#![doc = include_str ! ("../tests/fixtures/sampling_config.json")]
```
```suggestion
#![doc = include_str!("../tests/fixtures/sampling_config.json")]
```

relay-sampling/src/lib.rs
```diff
@@ -54,13 +54,13 @@
...
 //! ```
 //!
 //! ## [`DynamicSamplingContext`]
 //!
 //! ```json
-#![doc = include_str!("../tests/fixtures/dynamic_sampling_context.json")]
+#![doc = include_str ! ("../tests/fixtures/dynamic_sampling_context.json")]
```
jjbayer: Did the auto formatter make this change?

```suggestion
#![doc = include_str!("../tests/fixtures/dynamic_sampling_context.json")]
```

relay-sampling/src/lib.rs
```diff
@@ -993,8 +994,15 @@ pub fn merge_configs_and_match(
...
-    match_result.set_sample_rate(match sampling_config.mode {
+    // The determination of the sampling mode occurs with the following priority:
+    // 1. Non-root project sampling mode
+    // 2. Root project sampling mode
+    let Some(primary_config) = sampling_config.or(root_sampling_config) else {
+        relay_log::error!("cannot sample without at least one sampling config");
+        return None;
+    };
+    let sampling_mode = primary_config.mode;
+    let sample_rate = match sampling_mode {
```
```suggestion
    let sample_rate = match primary_config.mode {
```

relay-sampling/src/lib.rs
```diff
@@ -1295,7 +1304,7 @@ impl DynamicSamplingContext {
 
         let contexts = event.contexts.value()?;
         let context = contexts.get(TraceContext::default_key())?.value()?;
-        let Context::Trace(ref trace) = context.0 else { return None };
+        let Context::Trace(ref trace) = context.0 else { return None; };
```
```suggestion
        let Context::Trace(ref trace) = context.0 else { return None };
```

relay-server/src/actors/processor.rs
```diff
@@ -2237,18 +2240,95 @@ impl EnvelopeProcessorService {
...
+                if let (Some(Trace(boxed_context)), Some(sampling_result)) =
+                    (context, sampling_result)
+                {
+                    boxed_context.sampled = Annotated::new(match sampling_result {
+                        SamplingResult::Keep => true,
+                        SamplingResult::Drop(_) => false,
+                    });
+                }
+            }
+            None => {}
```
jjbayer: If there is no event, there's no need to compute the dynamic sampling decision. So I would make this a guard at the beginning of the function, that is

```rust
let event = Some(state.event.value_mut()) else { return };

File: relay-server/src/actors/processor.rs [hunk 15]
```diff
@@ -2237,18 +2240,95 @@ impl EnvelopeProcessorService {
...
+                // In case we have no contexts object, we have to create it.
+                if event.contexts.value().is_empty() {
+                    event.contexts = Annotated::new(Contexts::new());
+                }
+
+                // We want to get the specific trace context, or we want to create it in case
+                // it is not there.
+                let context = event.contexts.value_mut().as_mut().map(|context| {
+                    context
+                        .get_or_insert_with(TraceContext::default_key(), || Trace(Box::default()))
```
jjbayer: If `sampling_result` is `None`, we create a trace context here, but then we do not set `sampled`.

Do we want to overwrite the `sampled` information written by a prior Relay? Depending on that question, I would either: 

1. If not: Add another guard, something like `let Some(sampling_result) = ... { else return }`
2. If so: change the `if` on line 2319 to always write the outcome of `get_sampling_result` (even if it's none).

jjbayer (changes requested):

relay-server/src/actors/processor.rs
iambriccardo: Good observation, I forgot about those two lines when adding the tri-state. What we want is to not run the mutation in case the sampling result is None.

py/CHANGELOG.md
iambriccardo: Good point!

jjbayer (approved):

PR #2027, opened 2023-04-12
feat(monitors): Pass environment through

evanpurkhiser:
Supports specifying the checkins environment

### DIFF
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -113,6 +113,10 @@ struct CheckIn {
     /// Status of this check-in. Defaults to `"unknown"`.
     status: CheckInStatus,
 
+    /// The environment to associate the check-in with
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    environment: Option<String>,
+
     /// Duration of this check since it has started in seconds.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     duration: Option<f64>,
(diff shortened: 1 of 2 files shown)


rjo100 (approved):
thank you for this!!

olksdr (approved):

relay-monitors/src/lib.rs
```diff
@@ -113,6 +113,10 @@ struct CheckIn {
     /// Status of this check-in. Defaults to `"unknown"`.
     status: CheckInStatus,
 
+    /// The environment to associate the check-in with
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    environment: Option<String>,
```
iker-barriocanal, line 118: Let's add some length limits to this string, like [here](https://github.com/getsentry/relay/blob/ab2a3dca7a3daaecfcdf097d898c766fd6b69476/relay-general/src/protocol/event.rs#L381). Monitor slugs may contain non-ASCII characters, if this isn't the case for `environment` consider also adding [`allow_chars`](https://github.com/getsentry/relay/blob/ab2a3dca7a3daaecfcdf097d898c766fd6b69476/relay-general/src/protocol/event.rs#L368).

iker-barriocanal (approved):

relay-monitors/src/lib.rs
evanpurkhiser: Sounds good. Thanks for linking this

evanpurkhiser: Will merge and update this

evanpurkhiser: Actually, it looks like events uses a completely separate serialization system? I'm not familiar with `metastructure`

evanpurkhiser: Going to merge this as is and revisit validation

PR #2028, opened 2023-04-13
feat(kafka): Cache schemas for validation

jjbayer:
Add a caching layer to https://github.com/getsentry/relay/pull/2013 such that the schema does not have to be parsed from disk for every message.

#skip-changelog

### DIFF
--- a/relay-kafka/src/producer/mod.rs
+++ b/relay-kafka/src/producer/mod.rs
@@ -1,3 +1,4 @@
+use std::cell::RefCell;
 use std::collections::{BTreeMap, HashMap};
 use std::fmt;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


untitaker (approved):

relay-kafka/src/producer/mod.rs
```diff
@@ -284,6 +290,7 @@ impl KafkaClientBuilder {
     pub fn build(self) -> KafkaClient {
         KafkaClient {
             producers: self.producers,
+            schema_validator: Validator::default().into(),
```
untitaker, line 296:
```suggestion
            #[cfg(debug_assertions)]
            schema_validator: Validator::default().into(),
```

jjbayer: Thanks!

PR #2029, opened 2023-04-13
ref(buffering): Split queries into separate module

olksdr:
Create separate module which provides functions to build and in `insert` case also execute the sql queries. 
Also fixing and adding some more docs and notes. 

#skip-changelog

### DIFF
--- a/relay-server/src/sql/mod.rs
+++ b/relay-server/src/sql/mod.rs
@@ -42,6 +42,13 @@ pub fn delete<'a>(key: QueueKey) -> Query<'a, Sqlite, SqliteArguments<'a>> {
         .bind(key.sampling_key.to_string())
 }
 
+/// Creates a query which fetches the current page count and page size.
+///
+/// This info used to calculate the current allocated database size.
+pub fn current_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
+    sqlx::query("pragma page_count; pragma page_size;")
+}
+
 /// Descibes the chunk item which is handled by insert statement.
 type ChunkItem = (QueueKey, Vec<u8>, i64);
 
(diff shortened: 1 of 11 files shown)


relay-server/src/sql/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+/// This info used to calculate the current allocated database size.
+pub fn current_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
+    sqlx::query("pragma page_count; pragma page_size;")
+}
+
+/// Descibes the chunk item which is handled by insert statement.
+type ChunkItem = (QueueKey, Vec<u8>, i64);
+
+/// Creates an INSERT query for the chunk of provided data.
+pub fn insert_with_builder<'a>(
```
jjbayer: AFAICT this is not used outside of this module.

```suggestion
fn insert_with_builder<'a>(
```

relay-server/src/sql/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    builder.build()
+}
+
+/// Creates INSERT statements from the stream and execute them on provided database pool.
+///
+/// This function internally will split the provided stream into chunks and will prepare the
+/// insert statement for each chunk.
+pub async fn insert_with_exec(
```
jjbayer: nit: The fact that this function uses `execute` seems like an implementation detail. For the public interface name, I would call this `insert_chunks` or similar.

relay-server/src/sql/mod.rs
jjbayer, line 1: This module seems to be specific to `project_buffer`, so why not move `project_buffer.rs -> project_buffer/mod.rs`, and move this file to `project_buffer/sql.rs`?

jjbayer (approved):
Couple of nits but nothing blocking.

olksdr:
@jjbayer  I've renamed few things now. Could you, please, take another look and tell me what you think?

jjbayer (approved):
Thanks, LGTM!

PR #2031, opened 2023-04-13
feat(replays): lower max replay recording segment size to 10 MiB

JoshFerge:
- [x] Update the config to include two separate values for maximum replay size: one for compressed size (called `max_replay_compressed`) and one for uncompressed size `max_replay_uncompressed_size`. The original value now represents the uncompressed size (`100 MiB`), which is utilized during processing. The new value, `max_replay_compressed`, is applied during the ingest stage.
- [x] Replays formerly relied on the maximum attachment limit of 100 MiB. Now, they use the new max_replay_compressed_size (`10 MiB`) setting for check_envelope_size_limits.

We are making this change as we have done analysis around our payload sizes, and if replays are above this size there is usually something off with them (mass mutation bugs, accidental inclusion of files, old SDK versions, etc.)


bruno-garcia (approved):
RSLGTM!

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ Metrics:
 - The `route` tag of request metrics uses the route pattern instead of schematic names. There is an exact replacement for every previous route. For example, `"store-default"` is now tagged as `"/api/:project_id/store/"`.
 - Statsd metrics `event.size_bytes.raw` and `event.size_bytes.uncompressed` have been removed.
 
+Replays:
+
+- Lower max compressed replay recording segment size to 10 MiB. ([#2031](https://github.com/getsentry/relay/pull/2031))
+
```
olksdr: It doesn't seem like a breaking change to me.

You just changing the defaults and the configuration itself will be done somewhere else, which can still contain the old numbers.  And also can be rolled out completely unrelated to this PR.  From the Relay POV nothing breaking is happening.

relay-config/src/config.rs
```diff
@@ -542,8 +542,10 @@ struct Limits {
     max_api_chunk_upload_size: ByteSize,
     /// The maximum payload size for a profile
     max_profile_size: ByteSize,
-    /// The maximum payload size for a replay.
-    max_replay_size: ByteSize,
```
olksdr, line 546: Can we keep old config  and use `max_replay_size` as incoming check, the one is currently called `max_replay_compressed_size`, which clearly signals what the check is for, and if this is really necessary also use `serde` alias to rename it to the new name?

What do you think?

relay-config/src/config.rs
```diff
@@ -1822,9 +1825,14 @@ impl Config {
         self.values.limits.max_profile_size.as_bytes()
     }
 
-    /// Returns the maximum payload size for a replay.
-    pub fn max_replay_size(&self) -> usize {
-        self.values.limits.max_replay_size.as_bytes()
+    /// Returns the maximum payload size for an compressed replay.
```
olksdr:
```suggestion
    /// Returns the maximum payload size for a compressed replay.
```

olksdr (changes requested):
I left few questions, before approving this. 

Also you like to get @jjbayer opinion on the naming here.

relay-config/src/config.rs
jjbayer: The new naming looks OK to me, but I would add the old name as an `alias` as @olksdr suggests, such that this is not a breaking change.

CHANGELOG.md
jjbayer: +1 on moving this line to `** Internal **`.

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ Metrics:
 - The `route` tag of request metrics uses the route pattern instead of schematic names. There is an exact replacement for every previous route. For example, `"store-default"` is now tagged as `"/api/:project_id/store/"`.
 - Statsd metrics `event.size_bytes.raw` and `event.size_bytes.uncompressed` have been removed.
 
+Replays:
+
+- Lower max compressed replay recording segment size to 10 MiB. ([#2031](https://github.com/getsentry/relay/pull/2031))
```
```suggestion
- Lower default max compressed replay recording segment size to 10 MiB. ([#2031](https://github.com/getsentry/relay/pull/2031))
```

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ Metrics:
 - Pin schemars dependency to un-break schema docs generation. ([#2014](https://github.com/getsentry/relay/pull/2014))
 - Remove global service registry. ([#2022](https://github.com/getsentry/relay/pull/2022))
 - Apply schema validation to all topics in local development. ([#2013](https://github.com/getsentry/relay/pull/2013))
+- Lower default max compressed replay recording segment size to 10 MiB. ([#2031](https://github.com/getsentry/relay/pull/2031))
```
olksdr: Please, merge master in and move this record to `Unreleased` section.

olksdr (commented):
Once the changelog updated , this should be good to go.

jjbayer (approved):
See @olksdr 's changelog comments, otherwise LGTM!

JoshFerge:
@bmckerry has increased the max message size of our kafka topic, this and https://github.com/getsentry/relay/pull/2032 should be good to go! can merge once approved. can deploy this one first, then 2032 subsequently.

olksdr (approved):

PR #2032, opened 2023-04-13
ref(replays): increase chunking limit to 15MB

JoshFerge:
We were previously chunking kafka messages on the replay recording topic to 1 MiB. We are aiming to eliminate this chunking as it requires us to maintain a redis cluster / constrains us to in-order processing. Other infra at Sentry has successfully experimented with larger kafka messages, so we feel this is safe. 

With https://github.com/getsentry/relay/pull/2031 this should pretty much mean we no longer chunk any messages. We can remove the chunking logic once this is confirmed after deploy.

- [x] Increase chunking limit from 1MiB to 15 MiB. 

We chose 15MiB as the limit, as we're eventually going to combine the replay event with the replay recording, and will want some extra headroom for this in addition to our 10 MiB recording limit.

Depends on ops increasing our kafka payload size limit first.


relay-server/src/actors/store.rs
```diff
@@ -664,8 +664,8 @@ impl StoreService {
         // message because we can achieve better parallelism when dealing with a single
         // message.
 
-        // Max message size is 1MB.
-        let max_message_size = 1000 * 1000;
+        // Max message size is 15MB.
+        let max_message_size = 1000 * 150000;
```
olksdr: Shouldn't this be:
```suggestion
        let max_message_size = 1000 * 15000;
```

if you want to have 15mb?

olksdr (changes requested):

relay-server/src/actors/store.rs
olksdr: Could we also introduce the configuration option for this and not hardcode it?

relay-config/src/config.rs
```diff
@@ -1833,6 +1836,13 @@ impl Config {
     pub fn max_replay_uncompressed_size(&self) -> usize {
         self.values.limits.max_replay_uncompressed_size.as_bytes()
     }
+    /// Returns the maximum message size for an uncompressed replay.
+    /// this is greater than max_replay_compressed_size because
+    /// it can include additional metadata about the replay in
+    /// addition to the recording.
```
olksdr:
```suggestion
    }
    
    /// Returns the maximum message size for an uncompressed replay.
    ///
    /// This is greater than max_replay_compressed_size because
    /// it can include additional metadata about the replay in
    /// addition to the recording.
```

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ Metrics:
 - Remove global service registry. ([#2022](https://github.com/getsentry/relay/pull/2022))
 - Apply schema validation to all topics in local development. ([#2013](https://github.com/getsentry/relay/pull/2013))
 - Lower default max compressed replay recording segment size to 10 MiB. ([#2031](https://github.com/getsentry/relay/pull/2031))
+- Increase chunking limit to 15MB for replay recordings. ([#2032](https://github.com/getsentry/relay/pull/2032))
```
olksdr: Please, merge master in, and move this record to the `Unreleased` section.

olksdr (approved):

PR #2034, opened 2023-04-14
feat(replays): Key-based scrubbing

### LINKED ISSUE #2007: Session Replay: Key-based PII scrubbing
The Replay team has started to ingest fetch/xhr request and response bodies (and later on headers) on the Sentry org. As you can imagine, a big concern with this feature will be PII. Assuming the fetch response body is JSON, we will send the body as JSON rather than stringifying it. You can see an example payload here:

```json
{
    "type": 5,
    "timestamp": 1680009712.507,
    "data": {
        "tag": "performanceSpan",
        "payload": {
            "op": "resource.fetch",
            "description": "https://api2.amplitude.com/2/httpapi",
            "startTimestamp": 1680009712.507,
            "endTimestamp": 1680009712.671,
            "data": {
                "method": "POST",
                "statusCode": 200,
                "request": {
                    "size": 2949,
                    "body": {
                        "api_key": "foobar",
                        "events": "[...]",
                        "options": {
                            "min_id_length": 1
                        }
                    }
                },
                "response": {
                    "size": 94,
                    "body": {
                        "code": 200,
                        "server_upload_time": 1680009712652,
                        "payload_size_bytes": 2949,
                        "events_ingested": 5
                    }
                }
            }
        }
    }
}
```

Relay is already scrubbing replay events where `type == 5`, however we would want the scrubber to be contextually aware so that it is able to scrub values based on keys. In the example above we have `data.payload.data.request.body.api_key`, whose value (`foobar`) we would like to be scrubbed.

jjbayer:
Instead of scrubbing all string values in type 5 replay payloads and pretending that their key is `""`, make the scrubber aware of the current path in the payload, and restrict scrubbinh to a predefined set of fields (as we do for error events with `pii=true`).

By passing the current path into the PII processor, default key-based rules such as the [`PASSWORD_KEY_REGEX`](https://github.com/getsentry/relay/blob/3a81590c33b196c32c0c546ef93c7db9b1614ddd/relay-general/src/pii/regexes.rs#L275-L279) are now enabled as well.

In order to use the existing PII matchers, I had to modify `ProcessingState` to allow taking ownership of its parent.

Fixes https://github.com/getsentry/relay/issues/2007.


relay-general/src/processor/attrs.rs
```diff
@@ -384,7 +410,7 @@ impl<'a> fmt::Display for PathItem<'a> {
 /// the path items in the processing state and check whether a selector matches.
 #[derive(Debug, Clone)]
 pub struct ProcessingState<'a> {
-    parent: Option<&'a ProcessingState<'a>>,
+    parent: Option<BoxCow<'a, ProcessingState<'a>>>,
```
jjbayer, line 418: In regular event scrubbing, every state will only hold a reference to its parent (as it was before). But in Replay scrubbing, where we do not call `process_*` recursively, we will hold a single `ProcessingState` instance, which owns its parents, grandparents, etc.

relay-general/src/processor/attrs.rs
```diff
@@ -514,6 +559,18 @@ impl<'a> ProcessingState<'a> {
...
+    ///
+    /// - Returns `Ok(None)` if the current state is the root.
+    /// - Returns `Err(self)` if the current state does not own the parent state.
+    pub fn try_into_parent(self) -> Result<Option<Self>, Self> {
+        match self.parent {
+            Some(BoxCow::Borrowed(_)) => Err(self),
+            Some(BoxCow::Owned(parent)) => Ok(Some(*parent)),
+            None => Ok(None),
+        }
+    }
```
jjbayer, line 577: This function is used to pop a state from the stack.

relay-general/src/processor/attrs.rs
```diff
@@ -443,14 +469,33 @@ impl<'a> ProcessingState<'a> {
...
+    ) -> Self {
+        let depth = self.depth + 1;
+        ProcessingState {
+            parent: Some(BoxCow::Owned(self.into())),
+            path_item: Some(PathItem::OwnedKey(key)),
+            attrs,
+            value_type: value_type.into_iter().collect(),
+            depth,
+        }
+    }
```
jjbayer, line 502: This function is used to push a state to the stack.

relay-replays/src/recording.rs
```diff
@@ -213,6 +217,29 @@ where
     D::custom(s.to_string())
 }
 
+/// Config that will be applied in addition to any dynamic config added via `new()`.
+static STATIC_PII_CONFIG: Lazy<PiiConfig> = Lazy::new(|| {
+    let filter_paths = [vec![
+        "data", "payload", "data", "request", "body", "api_key",
+    ]];
```
jjbayer: @billyvg I only added the path you mentioned in the issue for now. If you want to add more paths, this is the place to do so.

relay-replays/src/recording.rs
```diff
@@ -562,4 +605,52 @@ mod tests {
...
+        let scrubbed_result = std::str::from_utf8(&transcoded).unwrap();
+        let scrubbed: serde_json::Value = serde_json::from_str(scrubbed_result).unwrap();
+
+        // Normal fields are not scrubbed:
+        assert_eq!(scrubbed[0]["data"]["payload"]["data"]["method"], "POST");
+
+        // `api_key` is caught by default scrubbing rules:
+        assert_eq!(
+            scrubbed[0]["data"]["payload"]["data"]["request"]["body"]["api_key"],
+            "[Filtered]"
```
jjbayer, line 647: @billyvg this test illustrates that even without loading the new `STATIC_PII_CONFIG`, `api_key` is now filtered because the scrubber has now become aware of keys, and `api_key` is listed as a secret key in the default data scrubbing config:

https://github.com/getsentry/relay/blob/1cfd522d43ce55cc63606bf5d2874d59bcd45963/relay-general/src/pii/regexes.rs#L275-L277

This also means that from now on, these keys will be scrubbed anywhere in the payload (`passwd`, `auth`, `token`, ...). Is there a risk that this will break replays functionally?

relay-replays/src/recording.rs
billyvg: @jjbayer Does this still work if the path is: `"data", "payload", "data", "request", "body"`

`request.body` is going to contain arbitrary data, so we won't know exact paths to the keys that need to be scrubbed.

relay-replays/src/recording.rs
billyvg: Oh ok, this answers my first comment if I'm understanding correctly. There is a risk if this is applied to all payloads (less risk if it's only where `type == 5`, but can still cause unwanted issues). Is it possible to only scrub for these keys within a path?

e.g. if in `STATIC_PII_CONFIG`:

```
let filter_paths = [vec![
        "data", "payload", "data", "request",
    ]];
```

I would want this assertion to be true:

```
assert_eq!(
            scrubbed[0]["data"]["payload"]["data"]["request"]["body"]["api_key"],
            "[Filtered]"
```

relay-replays/src/recording.rs
```diff
@@ -68,31 +68,72 @@ impl From<serde_json::Error> for ParseRecordingError {
...
+        self.path.push(key.to_owned());
+        let field_attrs = if scrub_at_path(&self.path) {
+            &FIELD_ATTRS_PII_TRUE
+        } else {
+            &FIELD_ATTRS_PII_FALSE
+        };
+
+        self.state = std::mem::take(&mut self.state).enter_owned(
+            key.to_owned(),
+            Some(Cow::Borrowed(field_attrs)),
```
jjbayer, line 130: `field_attrs` now depends on whether the field is in `PII_FIELDS`, just like it depends on the `pii` attribute in `Annotated` schemas.

relay-replays/src/recording.rs
```diff
@@ -562,4 +605,45 @@ mod tests {
...
+            .scrub_replay(payload.as_slice(), &mut transcoded)
+            .unwrap();
+
+        let scrubbed_result = std::str::from_utf8(&transcoded).unwrap();
+        let scrubbed: serde_json::Value = serde_json::from_str(scrubbed_result).unwrap();
+
+        // Normal fields are not scrubbed:
+        assert_eq!(scrubbed[0]["data"]["payload"]["data"]["method"], "POST");
+
+        // `api_key` is caught by custom rule:
```
jjbayer:
```suggestion
```

relay-replays/src/recording.rs
jjbayer: @billyvg this is what I now implemented using `PII_FIELDS`. The selector is currently `vec!["data", "payload", "data"]`, but feel free to change that.

relay-replays/src/recording.rs
```diff
@@ -68,31 +68,72 @@ impl From<serde_json::Error> for ParseRecordingError {
...
-        Some(Cow::Owned(FieldAttrs::new().pii(Pii::True))),
-        Some(ValueType::String),
-    )
+/// Paths to fields on which datascrubbing rules should be applied.
+///
+/// This is equivalent to marking a field as `pii = true` in an `Annotated` schema.
+static PII_FIELDS: Lazy<[Vec<&str>; 2]> = Lazy::new(|| {
+    [
+        vec!["data", "payload", "description"],
+        vec!["data", "payload", "data"],
```
jjbayer: @billyvg this is where you can add sensitive fields, or make them more precise.

relay-replays/src/snapshots/relay_replays__recording__tests__scrub_pii_key_based_edge_cases.snap
jjbayer, line 1: @billyvg I added a payload with some unrealistic edge cases to illustrate where exactly data is scrubbed now. See `rrweb-request-edge-cases.json` for the input data.

relay-replays/src/snapshots/relay_replays__recording__tests__scrub_pii_key_based_edge_cases.snap
```diff
@@ -0,0 +1,45 @@
+---
+source: relay-replays/src/recording.rs
+expression: scrubbed
+---
+[
+  {
+    "data": {
+      "payload": {
+        "data": {
+          "method": "POST [ip]",
```
jjbayer, line 10: An IP appearing in `method` is scrubbed, because `data.payload.data` is in `PII_FIELDS`.

relay-replays/src/snapshots/relay_replays__recording__tests__scrub_pii_key_based_edge_cases.snap
```diff
@@ -0,0 +1,45 @@
...
+          "request": {
+            "body": {
+              "api_key": "[Filtered]",
```
jjbayer, line 13: `api_key` is scrubbed because `data.payload.data` is in `PII_FIELDS`, and `api_key` is in [`PASSWORD_KEY_REGEX`](https://github.com/getsentry/relay/blob/258783fc348c7920cd3651734688ba95d12f5968/relay-general/src/pii/regexes.rs#L275-L279).

relay-replays/src/snapshots/relay_replays__recording__tests__scrub_pii_key_based_edge_cases.snap
```diff
@@ -0,0 +1,45 @@
...
+            "body": {
+              "code": 200,
+              "events_ingested": 5,
+              "payload_size_bytes": 2949,
+              "server_upload_time": 1680009712652,
+            },
+            "size": 94,
+          },
+          "statusCode": 200,
+          "token": "[Filtered]",
```
jjbayer, line 33: `token` is also in [PASSWORD_KEY_REGEX](https://github.com/getsentry/relay/blob/258783fc348c7920cd3651734688ba95d12f5968/relay-general/src/pii/regexes.rs#L275-L279).

relay-replays/src/snapshots/relay_replays__recording__tests__scrub_pii_key_based_edge_cases.snap
```diff
@@ -0,0 +1,45 @@
...
+        },
+        "description": "https://domain.com/api [Filtered] [ip]",
```
jjbayer, line 35: Scrubbed because `data.payload.description` is in `PII_FIELDS`.

relay-replays/src/snapshots/relay_replays__recording__tests__scrub_pii_key_based_edge_cases.snap
```diff
@@ -0,0 +1,45 @@
...
+        "endTimestamp": 1680009712.671,
+        "op": "resource.fetch 4111-1111-1111-1111",
```
jjbayer, line 37: If `op` contains a credit card number, it is not scrubbed because `op` is not in `PII_FIELDS`.

relay-replays/src/snapshots/relay_replays__recording__tests__scrub_pii_key_based_edge_cases.snap
```diff
@@ -0,0 +1,45 @@
...
+        "startTimestamp": 1680009712.507,
+      },
+      "tag": "performanceSpan 127.0.0.1",
```
jjbayer, line 40: If `tag` contains an IP address, it is not scrubbed because `tag` is not in `PII_FIELDS`.

relay-replays/src/recording.rs
jjbayer: See newest changes.

relay-general/src/processor/attrs.rs
```diff
@@ -328,47 +328,73 @@ impl Default for FieldAttrs {
...
+    Owned(Box<T>),
+}
+
+impl<T> Deref for BoxCow<'_, T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        match self {
+            BoxCow::Borrowed(slf) => slf,
+            BoxCow::Owned(slf) => slf.deref(),
```
iker-barriocanal: `slf` may be confusing, could we rename it? I think `contained` could be a better option.

relay-general/src/processor/attrs.rs
iker-barriocanal: > we will hold a single ProcessingState instance, which owns its parents, grandparents, etc.

Will this be the top-level `ProcessingState`? Could we add a comment in `ProcessingState` to note this github comment, since it's not trivial to figure it out?

relay-replays/src/recording.rs
```diff
@@ -68,31 +68,72 @@ impl From<serde_json::Error> for ParseRecordingError {
...
+    ]
 });
 
+/// Returns `True` if the given path should be treated as `pii = true`.
+fn scrub_at_path(path: &Vec<String>) -> bool {
+    PII_FIELDS
+        .iter()
+        .any(|p| p.iter().zip(path).all(|(k1, k2)| k1 == k2))
```
iker-barriocanal: > If either iterator returns [None](https://doc.rust-lang.org/std/option/enum.Option.html#variant.None), [next](https://doc.rust-lang.org/std/iter/trait.Iterator.html#tymethod.next) from the zipped iterator will return [None](https://doc.rust-lang.org/std/option/enum.Option.html#variant.None).

From the [docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.zip); so IIUC, a path will match with any of its subpaths with the same root. Is it possible to use something like `contains`?

relay-replays/src/recording.rs
iker-barriocanal, line 130: Q: this sets the `field_attrs`, but there may be additional fields that are being overwritten and are now lost; is this correct? In other words, should we overwrite or merge the attributes? It's also possible this question does not make sense at all.

iker-barriocanal (changes requested):
Are we planning to consider `pii = maybe` on the recordings?

Blocking because of a potential bug with `zip`.

relay-replays/src/recording.rs
jjbayer: Great catch!

relay-replays/src/recording.rs
jjbayer: Previously, we just used the static `STRING_STATE` here, which hard-coded its field attrs to `FieldAttrs::new().pii(Pii::True)` ([link](https://github.com/getsentry/relay/pull/2034/files/f67bef1b33b86a7db0acb204f3d66597e704d59b#diff-37853ccc35bc20c05e8c5a3f04679a82d62c0c1efb959f8db3e42049c97b6b16L74)).

Since replays are schemaless, all we care about for the purpose of data scrubbing is toggling the pii field attr.

relay-general/src/processor/attrs.rs
```diff
@@ -514,6 +559,18 @@ impl<'a> ProcessingState<'a> {
         }
     }
 
+    /// Return the contained parent state.
```
olksdr:
```suggestion
    /// Returns the contained parent state.
```

relay-replays/src/recording.rs
```diff
@@ -68,31 +68,72 @@ impl From<serde_json::Error> for ParseRecordingError {
...
-static STRING_STATE: Lazy<ProcessingState> = Lazy::new(|| {
-    ProcessingState::root().enter_static(
-        "",
-        Some(Cow::Owned(FieldAttrs::new().pii(Pii::True))),
-        Some(ValueType::String),
-    )
+/// Paths to fields on which datascrubbing rules should be applied.
+///
+/// This is equivalent to marking a field as `pii = true` in an `Annotated` schema.
+static PII_FIELDS: Lazy<[Vec<&str>; 2]> = Lazy::new(|| {
```
olksdr: Maybe it would be better to move this to the top of the file and put just after imports? 

Also does it make sense to use `Vec` here?  Maybe just switch to another array instead for now? Since we have the same number of parts in the arrays, like
```suggestion
static PII_FIELDS: Lazy<[[&str;3]; 2]> = Lazy::new(|| {
```

relay-replays/src/recording.rs
olksdr: @iker-barriocanal  can you give an example what you mean with your comment?  From what I can see this should work. 

This will match of course on the shorter paths, like if the `path` will have only  2 parts provided but our `PII_FIELDS` fields have 3. 

That's what you meant? 
It's easy to unit-test this function.

iker-barriocanal: > This will match of course on the shorter paths, like if the `path` will have only  2 parts provided but our `PII_FIELDS` fields have 3. 

That's what I meant. Paths should fully match to scrub them, with no extra items on any side.

billyvg (commented):
✅ ✅ ✅ This looks great from our end!

This is out of scope for this PR, but just wanted to confirm if custom scrubbing rules will also get applied to `PII_FIELDS`?

jjbayer:
> This is out of scope for this PR, but just wanted to confirm if custom scrubbing rules will also get applied to `PII_FIELDS`?

@billyvg, yes, advanced data scrubbing rules should work as well, although I have not tested it.

relay-replays/src/recording.rs
```diff
@@ -195,6 +245,9 @@ where
             if helper.ty == Self::SENTRY_EVENT_TYPE {
                 seq.serialize_element(&ScrubbedValue(raw, self.scrubber.clone()))
                     .map_err(s2d)?;
+                // `pop_path` calls should have reset the scrubber's state, but force a
+                // reset here just to be sure:
+                self.scrubber.borrow_mut().reset();
```
iker-barriocanal: Should we add some observability here to see if there's a bug in the previous code, or an edge-case we haven't considered? Referring to the state not being reset by `Transform`.

relay-general/src/processor/attrs.rs
```diff
@@ -328,47 +328,73 @@ impl Default for FieldAttrs {
...
+            PathItem::OwnedKey(s) => f.pad(s.as_str()),
             PathItem::Index(val) => write!(f, "{val}"),
         }
     }
 }
 
+/// Like [`std::borrow::Cow`], but with a boxed value.
+///
+/// This is useful for types that contain themselves, where otherwise the layout of the type
+/// cannot be computed, for example
```
iker-barriocanal, line 375: I think we discussed we didn't use an `Rc` because only two items were using it (the node itself and the child, so only one extra). I think this is worth mentioning in the docstrings `//` (not necessarily docs `///`, your call) to understand the reason for this design in the future.

relay-replays/src/recording.rs
```diff
@@ -68,31 +91,58 @@ impl From<serde_json::Error> for ParseRecordingError {
...
+            Some(Cow::Borrowed(field_attrs)),
+            Some(ValueType::String), // Pretend everything is a string.
+        )
+    }
+
+    fn pop_path(&mut self) {
+        if let Ok(Some(parent)) = std::mem::take(&mut self.state).try_into_parent() {
+            self.state = parent;
+        }
+        self.path.pop();
```
iker-barriocanal: `pop` returns `Option<T>`. Should we log an error if `pop` returns `None`? It means we tried to walk up to a parent while being in the root, which should never happen.

relay-replays/src/recording.rs
```diff
@@ -562,4 +619,61 @@ mod tests {
...
+            .unwrap();
+
+        let scrubbed_result = std::str::from_utf8(&transcoded).unwrap();
+        let scrubbed: serde_json::Value = serde_json::from_str(scrubbed_result).unwrap();
+
+        insta::assert_ron_snapshot!(scrubbed);
+    }
+
+    #[test]
+    fn test_scrub_at_path() {
```
iker-barriocanal, line 669: Good test.

iker-barriocanal (approved):
The PR looks good to me. I've left some comments about adding observability and logging errors, your call to add them.

olksdr (approved):
shipit!

relay-replays/src/recording.rs
jjbayer: Moved it to the top. I'd like to keep them `Vec`s because the path length varies, it's a coincidence that they both have length 3 at the moment.

relay-replays/src/recording.rs
jjbayer: Added debug asserts in two places now, + an error log on the caller side.

relay-general/src/processor/attrs.rs
```diff
@@ -410,9 +410,11 @@ impl<T> Deref for BoxCow<'_, T> {
...
 #[derive(Debug, Clone)]
 pub struct ProcessingState<'a> {
-    // In event scrubbing, every state will only hold a reference to its parent.
-    // In Replay scrubbing, where we do not call process_* recursively,
-    // we will hold a single ProcessingState instance, which owns its parents, grandparents, etc.
+    // In event scrubbing, every state holds a reference to its parent.
+    // In Replay scrubbing, we do not call `process_*` recursively,
+    // but instead hold a single `ProcessingState` that represents the current item.
+    // This item owns its parent (plus ancestors) exclusively, which is why we use `BoxCow` here
+    // rather than `Rc` / `Arc`.
```
iker-barriocanal, line 417: 👍

relay-replays/src/recording.rs
```diff
@@ -106,7 +106,11 @@ struct ScrubberTransform<'a> {
 }
 
 impl ScrubberTransform<'_> {
-    fn reset(&mut self) {
+    fn ensure_empty(&mut self) {
+        if !self.path.is_empty() || self.state.depth() > 0 {
+            debug_assert!(false, "ScrubberTransform not empty");
+            relay_log::error!("ScrubberTransform not empty");
```
iker-barriocanal, line 112: Should we add the project key to the error?

iker-barriocanal (approved):

PR #2035, opened 2023-04-14
ref(normalization): Always mark scrubbed transactions as sanitized

iker-barriocanal:
Currently, Relay only marks scrubbed transactions as sanitized with the feature flag enabled. Since we've GA-d this, there's no need to keep the feature flag.

The feature flag is `organizations:transaction-name-mark-scrubbed-as-sanitized`.

#skip-changelog

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -13,10 +13,6 @@ use crate::types::{Annotated, Meta, ProcessingAction, ProcessingResult, Remark, 
 /// Configuration around removing high-cardinality parts of URL transactions.
 #[derive(Clone, Debug, Default)]
 pub struct TransactionNameConfig<'r> {
-    /// True if transaction names scrubbed by regex patterns should be marked as [`TransactionSource::Sanitized`].
-    ///
-    /// Transaction names modified by clusterer rules are always marked as such.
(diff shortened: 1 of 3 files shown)


relay-general/src/store/transactions/processor.rs
```diff
@@ -13,10 +13,6 @@ use crate::types::{Annotated, Meta, ProcessingAction, ProcessingResult, Remark,
 /// Configuration around removing high-cardinality parts of URL transactions.
 #[derive(Clone, Debug, Default)]
 pub struct TransactionNameConfig<'r> {
-    /// True if transaction names scrubbed by regex patterns should be marked as [`TransactionSource::Sanitized`].
-    ///
-    /// Transaction names modified by clusterer rules are always marked as such.
-    pub mark_scrubbed_as_sanitized: bool,
     /// Rules for identifier replacement that were discovered by Sentry's transaction clusterer.
     pub rules: &'r [TransactionNameRule],
```
iker-barriocanal, line 17: Even if there's only `rules` left in the config, I'm planning to leave it as-is for now, and remove the config struct in a follow-up next week.

jjbayer (approved):
LGTM. If this causes problems for customers we can always reintroduce the flag.

olksdr (approved):
🚀

PR #2039, opened 2023-04-17
instr(statsd): Low-cardinality metric name tag

### LINKED ISSUE #1837: Stop tagging statsd metrics by metric name
We currently set the metric name on at least two statsd metrics:

https://github.com/getsentry/relay/blob/8f490056800d879903c669f7838f00b97fd72bf2/relay-metrics/src/aggregation.rs#L1706

This tag is no longer low-cardinality because of custom measurements.

Possible solution: Make metric name an `enum` with an `Other(String)` variant. Map that variant to `"other"` for the statsd tag.

jjbayer:
Instead of allowing _any_ metric name as a statsd tag, limit it to a predefined set of well known names.

Fixes #1837.

#skip-changelog

### DIFF
--- a/relay-metrics/src/statsd.rs
+++ b/relay-metrics/src/statsd.rs
@@ -148,3 +148,33 @@ impl GaugeMetric for MetricGauges {
         }
     }
 }
+
+/// Returns a low-cardinality metric name for use as a tag key on statsd metrics.
+///
+/// In order to keep this low-cardinality, we only enumerate a handful of well-known, high volume
+/// names. The rest gets mapped to "other".
+pub fn metric_name_tag<S: AsRef<str>>(value: S) -> &'static str {
+    let metric_name = value.as_ref();
+    if let Some(known_name) = [
+        "c:sessions/session@none",
+        "s:sessions/user@none",
+        "s:sessions/error@none",
+        "d:transactions/duration@millisecond",
+        "s:transactions/user@none",
+        "c:transactions/count_per_root_project@none",
+    ]
+    .into_iter()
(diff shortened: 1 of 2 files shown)


relay-metrics/src/statsd.rs
```diff
@@ -148,3 +148,33 @@ impl GaugeMetric for MetricGauges {
...
+        "s:sessions/error@none",
+        "d:transactions/duration@millisecond",
+        "s:transactions/user@none",
+        "c:transactions/count_per_root_project@none",
+    ]
+    .into_iter()
+    .find(|s| s == &metric_name)
+    {
+        return known_name;
+    }
```
olksdr, line 170: Why not just use `contains` method on slice here?
```suggestion
    let metric_name = value.as_ref();
    if [
        "c:sessions/session@none",
        "s:sessions/user@none",
        "s:sessions/error@none",
        "d:transactions/duration@millisecond",
        "s:transactions/user@none",
        "c:transactions/count_per_root_project@none",
    ].contains(&metric_name)
    {
        return metric_name;
    }
```

olksdr (approved):
I have one questions , otherwise lgtm!

relay-metrics/src/statsd.rs
olksdr: you might have to look into lifetimes, but it should also work just fine.

jjbayer: I first used `.contains()`, but `metric_name` does not have `'static` lifetime, which is why I return a reference to the string literal instead.

PR #2040, opened 2023-04-17
fix(feat-flags): Fix serialization of unknown feature flags

iker-barriocanal:
Currently, unknown feature flags that end up in the `Unknown` bucket are serialized as `"Unknown"`, so the actual feature flags don't reach downstream relays. Since `serde` doesn't yet support types with `"other"` fields, I've manually implemented the required serialization and deserialization methods.

As a result of this change, feature flags that have been removed in relay and are no longer required will still be sent to downstream relays. This is a trade-off we must accept to address the issue above, and this new issue should be resolved by not adding the feature flags to project configs in Sentry in the first place.

### DIFF
--- a/relay-dynamic-config/src/feature.rs
+++ b/relay-dynamic-config/src/feature.rs
@@ -1,30 +1,62 @@
+use std::borrow::Cow;
+
 use serde::{Deserialize, Serialize};
 
 /// Features exposed by project config.
-#[derive(Clone, Copy, Debug, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Hash)]
 pub enum Feature {
     /// Enables ingestion and normalization of profiles.
-    #[serde(rename = "organizations:profiling")]
     Profiling,
     /// Enables ingestion of Session Replays (Replay Recordings and Replay Events).
-    #[serde(rename = "organizations:session-replay")]
     SessionReplay,
     /// Enables data scrubbing of replay recording payloads.
-    #[serde(rename = "organizations:session-replay-recording-scrubbing")]
     SessionReplayRecordingScrubbing,
     /// Enables device.class synthesis
     ///
     /// Enables device.class tag synthesis on mobile events.
-    #[serde(rename = "organizations:device-class-synthesis")]
     DeviceClassSynthesis,
     /// Unused.
     ///
     /// This used to control the initial experimental metrics extraction for sessions and has been
     /// discontinued.
-    #[serde(rename = "organizations:metrics-extraction")]
     Deprecated1,
-
     /// Forward compatibility.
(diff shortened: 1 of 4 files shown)


relay-dynamic-config/src/feature.rs
```diff
@@ -1,30 +1,62 @@
...
+        match feature_name.trim() {
+            "organizations:profiling" => Ok(Feature::Profiling),
+            "organizations:session-replay" => Ok(Feature::SessionReplay),
+            "organizations:session-replay-recording-scrubbing" => {
+                Ok(Feature::SessionReplayRecordingScrubbing)
+            }
+            "organizations:device-class-synthesis" => Ok(Feature::DeviceClassSynthesis),
+            "organizations:metrics-extraction" => Ok(Feature::Deprecated1),
+            _ => Ok(Feature::Unknown(feature_name.to_string())),
+        }
```
jjbayer: nit: save some characters with `Ok(match ...)`

relay-dynamic-config/src/feature.rs
```diff
@@ -1,30 +1,62 @@
...
+    Unknown(String),
+}
+
+impl<'de> Deserialize<'de> for Feature {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let feature_name = Cow::<str>::deserialize(deserializer)?;
+        match feature_name.trim() {
```
```suggestion
        match feature_name {
```

If sentry sends something like `"  organizations:profiling   "`, that's a bug we don't have to be robust against.

relay-dynamic-config/src/feature.rs
```diff
@@ -1,30 +1,62 @@
...
     ///
     /// Enables device.class tag synthesis on mobile events.
-    #[serde(rename = "organizations:device-class-synthesis")]
     DeviceClassSynthesis,
     /// Unused.
     ///
     /// This used to control the initial experimental metrics extraction for sessions and has been
     /// discontinued.
-    #[serde(rename = "organizations:metrics-extraction")]
     Deprecated1,
```
jjbayer, line 22: Why not remove `Deprecated1`?

jjbayer (approved):

relay-dynamic-config/src/feature.rs
iker-barriocanal: I was planning to do that in a follow-up PR, to keep the PR scope small and focused. No reason besides that.

relay-dynamic-config/src/feature.rs
iker-barriocanal: The compiler doesn't like matching `Cow`s and `str`s, so I added `trim` to get the `str`. I changed this to getting the reference instead -- no important downsides, and should be able to spot potential bugs on the project config generation code in sentry.

jjbayer: Great!

PR #2041, opened 2023-04-17
ref(statsd): Use contains and `&str` directly

olksdr:
Simplify the metric tag name match:
* use `&str` directly
* use `contains` method on the slice instead of `iter` and `find` methods combined

@jjbayer what do you think about this change? 

#skip-changelog


jjbayer (approved):
Yep, thanks for simplifying!

iker-barriocanal (approved):

PR #2044, opened 2023-04-18
fix(build): Add Rust target explicitly

olksdr:
In this change 
* add the Rust target explicitly 
* remove depreciation warning  by adding `--force-non-host` to install toolchain on host with different architecture 

#skip-changelog


jjbayer (approved):

iker-barriocanal (commented):
TIL, good stuff!

PR #2047, opened 2023-04-19
ref(dynamic-sampling): Split function for sampling match

iambriccardo:
This PR is a preparation for the refactoring of dynamic sampling. It contains the splitting of the matching function in order to decouple `ProjectState` and `SamplingConfig`.


relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -50,21 +50,15 @@ fn check_unsupported_rules(
 
 /// Gets the sampling match result by creating the merged configuration and matching it against
 /// the sampling configuration.
-fn get_sampling_match_result(
+pub fn merge_configs_and_match(
```
iker-barriocanal: If you think about the semantical usage of this function, there're two steps that we're doing here. It'd be great if we could split this function into the merging and matching functions. Could we do this in a follow-up PR?

iker-barriocanal: Do we need this to be public?
```suggestion
fn merge_configs_and_match(
```

iker-barriocanal (approved):
Thank you for taking the time to improve the code quality, and splitting up the original PR!

Please see the comment on the `pub` before merging, and consider continuing the refactor in follow-up PRs.

relay-server/src/utils/dynamic_sampling.rs
iambriccardo: Ehm, they are already split up in lines 66 and 67. This function orchestrates their behavior, this is why it is named like this.

iambriccardo: Thanks!

iker-barriocanal: I was thinking something like the following:

```rs
fn get_sampling_match() {
  merge_configs();
  match_configs();
}
```

iambriccardo: I don't think that the complexity of the code requires the creation of a third function encapsulating that behavior. If it was more complex I would have agreed. On the other hand, if the encapsulation follows some strict Relay guidelines I am happy to adapt.

PR #2048, opened 2023-04-19
feat(protocol): Support old effective-directory format

### LINKED ISSUE #1973: Add support for CSP reports with extended `effective-directive`
Consider this CSP violation report:
```
{"csp-report":{"document-uri":"REDACTED","referrer":"REDACTED","violated-directive":"script-src 'report-sample' 'strict-dynamic' 'unsafe-eval' 'nonce-random'","original-policy":"base-uri 'self'; object-src 'none'; script-src 'report-sample' 'strict-dynamic' 'unsafe-eval' 'nonce-random'","blocked-uri":"REDACTED","effective-directive":"script-src 'report-sample' 'strict-dynamic' 'unsafe-eval' 'nonce-random'"}}
```

Currently, Relay drops those with such error:
```
failed to extract security report: invalid security report
  caused by: invalid value: string "script-src 'report-sample' 'strict-dynamic' 'unsafe-eval' 'nonce-random'", expected a csp directive at line 1 column 729
```

Apparently, this has something to do with older browsers before CSPv3 where `effective-directive` and `violated-directive` contained a longer substring of CSP. Check this excerpt from [csplite.com](https://csplite.com/csp66/#violated-directive):
> According to recomendations for [CSP 1](https://csplite.com/csp1/#violated-directive) and [CSP 2](https://csplite.com/csp2/#violation-report-violated-directive), in the violated-directive (unlike [effective-directive](https://csplite.com/csp66/#effective-directive)), not only the name of the directive is sent, but also a set of its rules, for example:
> `"violated-directive":"script-src 'unsafe-inline' 'unsafe-eval' 'self' https://site.com ..."`
> In [CSP 3](https://csplite.com/csp3/#violation-effective-directive) this behavior has been changed - now always [violated-directive = effective-directive](https://csplite.com/csp3/#deprecated-serialize-violation), and it is more does not contain violated policy rules:
> > [Note CSP 3](https://csplite.com/csp3/#violated-effective). Both [effectiveDirective](https://csplite.com/csp66/#effective-directive) and [violatedDirective](https://csplite.com/csp66/#violated-directive) have [the same value](https://csplite.com/csp3/#violated-effective) (meaning the name of the directive). This is intentional to maintain backwards compatibility.

My proposal is to add support for such cases. Seems like `effective-directive` is a searchable field (i.e. we can have a filter on the issues page like `effective-directive:img-src`), so I guess it should be of low cardinality. To keep cardinality low, we can do the following:
- save the original value of `effective-directive` CSP field to `effective-directive-full` (or something similar), make it unsearchable;
- save the first keyword of `effective-directive` CSP field to `effective-directive`.

That is, for the CSP violation above, the results would look like:
```
effective-directive: script-src
effective-directive-full: script-src 'report-sample' 'strict-dynamic' 'unsafe-eval' 'nonce-random'
```

TBS1996:
from [csplite.com](https://csplite.com/csp66/#violated-directive):

> According to recomendations for [CSP 1](https://csplite.com/csp1/#violated-directive) and [CSP 2](https://csplite.com/csp2/#violation-report-violated-directive), in the violated-directive (unlike [effective-directive](https://csplite.com/csp66/#effective-directive)), not only the name of the directive is sent, but also a set of its rules, for example:
> "violated-directive":"script-src 'unsafe-inline' 'unsafe-eval' 'self' https://site.com ..."

Currently when we parse `effective-directive` or `violated-directive` we assume that it only contains the CSP-dir, so if its set of rules are there as in version 1 and 2, a parsing error will occur. This PR will attempt to extract the CSP-dir if the full format is used. 
#1973


TBS1996:
from the issue:



> My proposal is to add support for such cases. Seems like effective-directive is a searchable field (i.e. we can have a filter on the issues page like effective-directive:img-src), so I guess it should be of low cardinality. To keep cardinality low, we can do the following:
> 
> save the original value of effective-directive CSP field to effective-directive-full (or something similar), **make it unsearchable**;


do I need to do something to make it unsearchable, or will a new field like this one be unsearchable by default?

relay-general/src/protocol/security_report.rs
```diff
@@ -183,7 +183,7 @@ fn normalize_uri(value: &str) -> Cow<'_, str> {
 #[serde(rename_all = "kebab-case")]
 struct CspRaw {
     #[serde(skip_serializing_if = "Option::is_none")]
-    effective_directive: Option<CspDirective>,
+    effective_directive: Option<String>,
```
TBS1996, line 186: Change it to a string so it won't be implicitly parsed immediatly into CSP-directive, but instead we can handle the cases of the old format.

relay-general/src/protocol/security_report.rs
```diff
@@ -216,22 +216,34 @@ impl CspRaw {
...
         // refs: https://bugzil.la/1192684#c8
 
-        if let Some(directive) = self.effective_directive {
-            return Ok(directive);
+        if let Some(directive) = &self.effective_directive {
+            // In C2P1 and CSP2, violated_directive and possibly effective_directive might contain
+            // more information than just the CSP-directive. For that reason, we try to split it
+            // and we return the original value as well so as to not lose any information.
+            if let Ok(parsed_directive) = directive
+                .split_once(' ')
```
TBS1996, line 228: so 'effective_directive' isn't supposed to have the old longer format even in csp1 and csp2, although we have a customer who populates this field with that format anyway, so i applied the same logic to it

relay-general/src/protocol/security_report.rs
```diff
@@ -216,22 +216,34 @@ impl CspRaw {
         "self".to_string()
     }
 
-    fn effective_directive(&self) -> Result<CspDirective, InvalidSecurityError> {
+    fn effective_directive(&self) -> Result<(CspDirective, String), InvalidSecurityError> {
```
TBS1996: possible alternative: return an Option<String> with the Some-variant only when the longer format is present

jjbayer (approved):
I'm not sure if saving the `effective_directive_full` field is worth the effort, we could also just drop the original value, but I guess it does no harm. Note that the new field will probably be normalized away on sentry-side until librelay gets an update (you can verify that by adding a test to https://github.com/getsentry/sentry/blob/f2d9ea5a430b0205cd5769bc0c02920fa449556b/tests/sentry/event_manager/interfaces/test_csp.py).

TBS1996:
ok, I just had it because alex suggested it, if it'll get normalized anyway then i prefer to not have it to keep it simple, ive removed it now. Although i'll ask him if he had any thoughts on this

PR #2049, opened 2023-04-19
test (relay-sampling): Add test to demonstrate more complex example

andriisoldatenko:
This PR add test to emulate what we have right now at Sentry, and it demonstrates that we want to put factor rule to the top.

#skip-changelog

### DIFF
--- a/relay-sampling/src/lib.rs
+++ b/relay-sampling/src/lib.rs
@@ -1935,11 +1935,7 @@ mod tests {
 
         for (rule_test_name, expected, condition) in conditions.iter() {
             let failure_name = format!("Failed on test: '{rule_test_name}'!!!");
-            assert!(
-                condition.matches(&dsc, None) == *expected,
-                "{}",
-                failure_name
-            );
+            assert_eq!(condition.matches(&dsc, None), *expected, "{}", failure_name);
         }
     }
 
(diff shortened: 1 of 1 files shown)


iambriccardo (approved):

relay-sampling/src/lib.rs
```diff
@@ -2770,6 +2758,100 @@ mod tests {
...
     }
 
+    #[test]
+    /// Tests that the multi-matching works for a mixture of trace and transaction
+    /// 1. Rule is healthcheck bias
+    /// 2. Rule is dev env bias
+    /// 3. Tx rebasing rule part a
+    /// 4. Tx rebasing rule part b
+    /// 5. Factor bias ( to fix rate )
+    /// 6. Uniform sample rate
```
```suggestion
    /// Tests that the multi-matching works for a mixture of trace and transaction.
    ///
    /// 1. Rule is health-check bias
    /// 2. Rule is dev env bias
    /// 3. Tx rebasing rule part a
    /// 4. Tx rebasing rule part b
    /// 5. Factor bias ( to fix rate )
    /// 6. Uniform sample rate
```

relay-sampling/src/lib.rs
```diff
@@ -2770,6 +2758,100 @@ mod tests {
...
+    fn test_match_against_rules_with_multiple_event_types_and_matches() {
+        let healthcheck_rule_id = 1002;
```
olksdr: This `healthcheck_rule_id` var is used in one place, it's fine to hardcode this number in the rule id below (especially in tests).

olksdr (approved):
:shipit:

relay-sampling/src/lib.rs
andriisoldatenko: @olksdr thanks! fixed in 82c7da46c

PR #2052, opened 2023-04-19
chore(spooler): Add errors per operation and increase robustness 

olksdr:
Adding more defined and separate errors to each operations to have better observability in what part of the spooling could fail and why. 

Also include few protective checks, in order to limit the DB file hits when not necessary. 

#skip-changelog

### DIFF
--- a/relay-server/src/actors/spooler/mod.rs
+++ b/relay-server/src/actors/spooler/mod.rs
@@ -854,6 +854,7 @@ mod tests {
             "buffer.envelopes_mem:2000|h",
             "buffer.reads:1|c",
             "buffer.reads:1|c",
+            "buffer.disk_size:24576|h",
             "buffer.envelopes_mem:0|h",
         ]
         "###);
--- a/relay-server/src/actors/spooler/mod.rs
+++ b/relay-server/src/actors/spooler/mod.rs
@@ -38,9 +38,20 @@ pub enum BufferError {
     #[error("failed to get the size of the buffer on the filesystem")]
     DatabaseFileError(#[from] std::io::Error),
 
-    /// Describes the errors linked with the `Sqlite` backed buffer.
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/spooler/mod.rs
```diff
@@ -271,7 +285,8 @@ impl BufferService {
                 .max_connections(config.spool_envelopes_max_connections())
                 .min_connections(config.spool_envelopes_min_connections())
                 .connect_with(options)
-                .await?;
+                .await
+                .map_err(BufferError::SetupFailed)?;
```
jjbayer, line 289: Could `setup()` return the `db`? Then this second intitialization would not be necessary. Or do we need different options here than in `setup()`?

relay-server/src/actors/spooler/sql.rs
```diff
@@ -46,7 +46,9 @@ pub fn delete<'a>(key: QueueKey) -> Query<'a, Sqlite, SqliteArguments<'a>> {
 ///
 /// This info used to calculate the current allocated database size.
 pub fn current_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
-    sqlx::query("pragma page_count; pragma page_size;")
+    sqlx::query(
+        "SELECT page_count * page_size as size FROM pragma_page_count(), pragma_page_size();",
```
jjbayer, line 50: Nice!

jjbayer (approved):

relay-server/src/actors/spooler/mod.rs
olksdr: yeah, so we need different options for migrations and then usage. And migration must run in separate dedicated pool before everything else, to make sure that all the things setup properly.

PR #2054, opened 2023-04-20
feat(profiles): Emit "accepted" outcomes for profiles filtered by sampling

jjbayer:
To facilitate a billing model that is consistent between transactions and profiles, #2051 introduced a new data category for profiles, such that

```
processed profiles = indexed profiles + sampled profiles  # "sampled" as in dropped by dynamic sampling
```

just like it is already the case for transactions:

```
processed transactions = indexed transactions + sampled transactions
```

In other words, "processed profiles" should count _all_ valid, non rate limited profiles seen by Relays, even if they were dropped by dynamic sampling.

## Difference between transactions and profiles

For transactions, we extract _metrics_ before dynamic sampling, and those metrics are what we rate limit and eventually log as "accepted" for the "processed" transaction data category (`DataCategory.Transaction`).

For profiles, we do not extract metrics (yet), so the outcomes for the "processed" profile data category have to be calculated in a different way.

## How this PR achieves this goal

1. In processing Relays, if an envelope still contains profiles _after_ dynamic sampling, log an `Accepted` outcome for the "processed" category. By restricting this to processing Relays, we can be sure that every profile is only counted once.
2. Also in processing Relays, shortly before reporting outcomes to kafka,
2.1. translate `Filtered` outcomes with reason `Sampled:` to an `Accepted` outcome. This counts all profiles dropped by dynamic sampling, regardless of where the dynamic sampling took place (external relay, pop relay, processing relay).
2.2. Also send the original `Filtered` outcome, but with data category `DataCategory.TransactionIndexed`.

By adding up the counts of these two disjoint sets, we should correctly count all profiles regardless of whether they were sampled or not.

## Alternative proposal (rejected for now)

In order to _actually_ line up behavior of transactions and profiles, we could start extracting a simple counter metric for profiles before dynamic sampling, and let that metric represent `DataCategory.Profile` -- this would mean that rate limits are applied to the metric, and the accepted outcome would be emitted from the `billing_metrics_consumer`, just like we do for the `DataCategory.Transactions`. See [this internal doc](https://www.notion.so/sentry/Implementation-Concerns-412caf18c2f04f579bb551b98c9dad8c) for more.

## Why the currrent approach was chosen

I personally believe that the alternative proposal described above would be more correct and easier to maintain, but:

1. By containing all new logic to processing relays, we will correctly count "processed" profiles even if they were dropped by dynamic sampling outdated external relays.
3. In a wider sense, by containing all new logic to processing relays, we can iterate faster (deploy hotfixes, etc.) without having to worry about the behavior of external relays (which we cannot update).
4. This change is easy to revert if needed. The alternative solution would be distributed across nodes, from external relays to sentry.


relay-server/src/actors/processor.rs
```diff
@@ -1070,6 +1070,51 @@ impl EnvelopeProcessorService {
...
+    /// enforce rate limits and emit outcomes based on the collect profile metric, as we do for
+    #[cfg(feature = "processing")]
+    fn count_processed_profiles(&self, state: &mut ProcessEnvelopeState) {
+        let profile_count: usize = state
+            .managed_envelope
+            .envelope()
+            .items()
+            .filter(|item| item.ty() == &ItemType::Profile)
+            .map(|item| item.quantity())
+            .sum();
```
phacops, line 1104: Technically, the SDKs will only send 1 profile per envelope.

phacops (commented):
LGTM

phacops:
It'd be good to deploy https://github.com/getsentry/sentry/pull/47703 before this PR is deployed to make sure we don't double count indexed profiles as processed.

olksdr (approved):
lgtm. Let's test it in prod 😄

relay-server/src/actors/processor.rs
```diff
@@ -1070,6 +1070,51 @@ impl EnvelopeProcessorService {
...
+    ///
+    /// NOTE: Instead of emitting a [processed](`DataCategory::Profile`) outcome here,
+    /// we could also do it in sentry, in the same place where the [indexed](`DataCategory::ProfileIndexed`)
+    /// outcome is logged. We do it here to be consistent with profiles that are dropped by dynamic sampling,
+    /// which also count as "processed" even though they did not pass through the `process_profiles` step yet.
+    ///
+    ///
+    /// In the future, we might actually extract metrics from profiles before dynamic sampling,
+    /// like we do for transactions. At that point, this code should be removed, and we should
+    /// enforce rate limits and emit outcomes based on the collect profile metric, as we do for
```
olksdr, line 1094: The sentence ends here kind of leaving you hanging at the end.
```suggestion
    /// enforce rate limits and emit outcomes based on the collect profile metric, as we do for transactions
```

jjbayer: Thanks!

relay-server/src/actors/processor.rs
jjbayer: Makes sense, but Relay does not enforce that AFAIK, so its safer to handle the case where there's multiple profiles.

jjbayer:
> lgtm. Let's test it in prod 😄

@olksdr I do hope the integration test covers the most important use case, let me know if you think we need more.

jjbayer:
> It'd be good to deploy [getsentry/sentry#47703](https://github.com/getsentry/sentry/pull/47703) before this PR is deployed to make sure we don't double count indexed profiles as processed.

@phacops yes, let's coordinate when to do the deploy!

phacops (approved):

PR #2055, opened 2023-04-21
ref(spooler): Add state into the buffer service

olksdr:
This is an initial refactor of the `BufferService` with introduced `BufferState`. If the on-disk spool file configuration option provided the state is switched between:
* `MemoryFileStandby` - use memory for buffering and when the max limit reached, switch to `File` and continue spooling to and reading from disk 
* `MemoryFileRead` - the file size hit the limit, and we switched to spooling to in-memory buffer only, but still continue to read from disk, when disk is empty we switch state to `MemoryFileStandby` or to `File`, if the there is more space for in-memory buffer to fit onto the disk
* `File` - use disk for spooling to and reading from, when the disk is full we switch state to `MemoryFileRead`

If there is no on-disk spooling, only 1 state `Memory`  will be always used.

This separation makes possible to better reason about what parts of the system currently involved, where the spool is located. Avoiding unnecessary accesses to the system, which is not used at the moment (e.g. if the in-memory buffer used, in normal circumstances there is no need to hit the disk all the time).

In terms of functionality almost nothing should change, except as mentioned above, depending on the state some parts of the buffer / spooler won't be used till the required state is reached, and I only changed test to accommodate this (see  test `metrics_work `).   This is mostly about the organisation , introducing helper structures and shuffling code around. 

### State changes can be represented as following: 

```mermaid
%%{init: {'theme':'forest'}}%%
stateDiagram-v2
direction LR

MemoryFileStandby --> File : Memory is full 
File --> MemoryFileStandby : Disk is empty
File --> MemoryFileRead : Disk is full
MemoryFileRead --> File : Disk has enough \n space for in-memory buffer
MemoryFileRead --> MemoryFileStandby: Disk is empty

```

#skip-changelog


relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,330 @@ impl BufferService {
...
+    /// Only disk used for read/write operations.
+    File { disk: OnDisk },
+}
+
+impl BufferState {
+    /// Creates the initial [`BufferState`] depending on the provided config.
+    ///
+    /// If the on-disk spool is configured, the `BufferState::MemoryDiskStandby` will be returned or
+    /// `BufferState::Memory` otherwise.
+    fn init(max_memory_size: usize, on_disk: Option<OnDisk>) -> Self {
```
jjbayer: nit:

```suggestion
    fn new(max_memory_size: usize, on_disk: Option<OnDisk>) -> Self {
```

relay-server/src/actors/spooler/sql.rs
```diff
@@ -51,6 +51,26 @@ pub fn current_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
     )
 }
 
+/// Creates the query to select only 1 record from the database.
```
iker-barriocanal: It's not a `SELECT * ...`, so we should add that to the docstrings:
```suggestion
/// Creates the query to select only 1 record's `received_at` from the database.
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -179,148 +176,105 @@ impl FromMessage<RemoveMany> for Buffer {
...
+        );
+
+        count
+    }
 
-            service.spool_config = Some(spool_config);
+    /// Dequeues the envelopes from the in-memory buffer and send them to provided `sender`.
+    fn dequeue(&mut self, keys: &Vec<QueueKey>, sender: mpsc::UnboundedSender<ManagedEnvelope>) {
+        for key in keys {
+            for value in self.buffer.remove(key).unwrap_or_default() {
```
iker-barriocanal: This makes it easier to understand it:
```suggestion
            for envelope in self.buffer.remove(key).unwrap_or_default() {
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -179,148 +176,105 @@ impl FromMessage<RemoveMany> for Buffer {
...
+        relay_statsd::metric!(
+            histogram(RelayHistograms::BufferEnvelopesMemory) = self.used_memory as f64
+        );
+
+        count
+    }
 
-            service.spool_config = Some(spool_config);
+    /// Dequeues the envelopes from the in-memory buffer and send them to provided `sender`.
+    fn dequeue(&mut self, keys: &Vec<QueueKey>, sender: mpsc::UnboundedSender<ManagedEnvelope>) {
```
iker-barriocanal, line 220: The `dequeue` and `remove` methods pretty much share the same API. Could we unify them?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -179,148 +176,105 @@ impl FromMessage<RemoveMany> for Buffer {
...
+                sender.send(value).ok();
+            }
         }
+        relay_statsd::metric!(
+            histogram(RelayHistograms::BufferEnvelopesMemory) = self.used_memory as f64
+        );
+    }
 
-        Ok(service)
+    /// Enqueues the envelope into in-memory buffer.
```
iker-barriocanal:
```suggestion
    /// Enqueues the envelope into the in-memory buffer.
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -179,148 +176,105 @@ impl FromMessage<RemoveMany> for Buffer {
...
-    max_disk_size: usize,
-    max_memory_size: usize,
-    is_disk_full: bool,
-}
-
-/// [`Buffer`] interface implementation backed by SQLite.
-#[derive(Debug)]
-pub struct BufferService {
-    services: Services,
+struct InMemory {
```
iker-barriocanal, line 181:
```suggestion
struct InMemoryBuffer {
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -179,148 +176,105 @@ impl FromMessage<RemoveMany> for Buffer {
...
 
-    /// Estimates the db size by multiplying `page_count * page_size`.
-    async fn estimate_buffer_size(db: &Pool<Sqlite>) -> Result<i64, BufferError> {
-        let size: i64 = sql::current_size()
-            .fetch_one(db)
-            .await
-            .and_then(|r| r.try_get(0))
-            .map_err(BufferError::FileSizeReadFailed)?;
+    /// Returns `true` if the in-memory buffer is full, `false` otherwise.
+    fn is_full(&self) -> bool {
```
iker-barriocanal, line 242: ∞ do we need a `does_fit` method before adding a new one?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -179,148 +176,105 @@ impl FromMessage<RemoveMany> for Buffer {
...
+        self.used_memory >= self.max_memory_size
+    }
+}
 
-        relay_statsd::metric!(histogram(RelayHistograms::BufferDiskSize) = size as u64);
-        Ok(size)
+/// The configuration which describes the on-disk [`BufferState`].
+#[derive(Debug)]
+struct OnDisk {
```
iker-barriocanal, line 249:
```suggestion
struct OnDiskBuffer {
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -179,148 +176,105 @@ impl FromMessage<RemoveMany> for Buffer {
...
-    async fn do_spool(
-        db: &Pool<Sqlite>,
-        buffer: BTreeMap<QueueKey, Vec<ManagedEnvelope>>,
+    async fn spool(
+        &self,
+        buffer: &mut BTreeMap<QueueKey, Vec<ManagedEnvelope>>,
+        used_memory: &mut usize,
     ) -> Result<(), BufferError> {
+        let buffer = std::mem::take(buffer);
+        *used_memory = 0;
```
iker-barriocanal: > Returns an error if the spooling failed, and the number of spooled envelopes on success.

If spooling failed, or partially failed (e.g. only spooled half the buffer), `used_memory` isn't 0.

iker-barriocanal: Should we check if there's available space on the disk? What if there's only space for the first item in the buffer? We're not checking that.

relay-server/src/actors/spooler/mod.rs
iker-barriocanal, line 249: Both buffers share some behavior and APIs. Have you considered creating a trait?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,330 @@ impl BufferService {
...
+    /// Dequeues the envelopes from the on-disk spool and send them to the provided `sender`.
+    ///
+    /// The keys for which the envelopes could not be fetched, send back to `ProjectCache` to merge
+    /// back into index.
+    async fn dequeue(
+        &self,
+        project_key: ProjectKey,
+        keys: &mut Vec<QueueKey>,
+        sender: mpsc::UnboundedSender<ManagedEnvelope>,
+        services: &Services,
```
iker-barriocanal, line 392: Is there any reason why `services` can't be provided to the buffer, so that we can remove this parameter?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,330 @@ impl BufferService {
...
+        .map_err(BufferError::InsertFailed)?;
+
+        relay_statsd::metric!(counter(RelayCounters::BufferWrites) += 1);
+        Ok(())
+    }
+}
+
+/// The state which defines the [`BufferService`] behaviour.
+#[derive(Debug)]
+enum BufferState {
```
iker-barriocanal, line 467: Even with the PR description, I find these names confusing. What do you think about the following suggestions?

- `MemoryOnly`.
- `MemoryFirstSpoolStandby`.
- `MemoryFirstSpoolRead`.
- `SpoolOnly`.

iker-barriocanal (dismissed):
I did a partial review and added some suggestions. In general, the PR is quite big, and splitting it into multiple and smaller PRs would help review it.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,330 @@ impl BufferService {
...
-        // Refresh DB state only if we get the proper reading on the file.
-        if let Ok(estimated_size) = Self::estimate_buffer_size(db).await {
-            *is_disk_full = (estimated_size as usize) >= *max_disk_size;
+        match self.state {
+            BufferState::Memory { ref mut ram } => ram.enqueue(key, managed_envelope),
+            // Will buffer till either there free space on disk, and the in-memory buffer can be
+            // spooled, or there is OOM.
+            // NOTE: Right now the number of envelopes is still restricted by
+            // `cache.envelope_buffer_size` and they will be dropped once the limit is reached.
+            BufferState::MemoryFileRead {
```
jjbayer: nit: I would place the `MemoryFileStandby` branch before the `MemoryFileRead` branch, to resemble the time line of "normal operation" -> "spooling to disk" -> "unspool".

relay-server/src/actors/spooler/mod.rs
jjbayer: I agree with the sentiment, but I find the suggestions just as confusing. Maybe naming is not the problem? E.g., maybe `Memory` and `Standby` don't have to be separate states?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,330 @@ impl BufferService {
...
+                ram: in_memory,
+                disk: on_disk,
+            },
+            None => Self::Memory { ram: in_memory },
+        }
+    }
+
+    /// Creates `BufferState::MemoryDiskStandby` from provided disk reference and maximum in-memory
+    /// buffer size.
+    fn to_memory(max_memory_size: usize, disk: &OnDisk) -> Self {
```
jjbayer: Regardless of what names we choose for the states, can we align the names of the `to_` methods with the state names? In this case, that would be `to_memory_file_standby`.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -525,33 +730,35 @@ impl BufferService {
...
+                }
+            }
+            BufferState::File { ref disk } => {
+                disk.dequeue(project_key, &mut keys, sender, &self.services)
+                    .await;
+
+                // If there is nothing left on the disk, just swithc to in-memory buffer.
+                if disk.is_empty().await.unwrap_or_default() {
+                    self.state =
+                        BufferState::to_memory(self.config.spool_envelopes_max_memory_size(), disk);
```
jjbayer: This and similar logic (check some condition -> transition state) seems to be duplicated in a few places. Can we instead have an `update_state(&mut self)` method where all transitions are handled? Beside reducing code duplication, this method would then compactly represent the entire state machine, which makes understanding it more easy.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,330 @@ impl BufferService {
...
+                    self.state = BufferState::to_file(disk);
+                }
+            }
+            BufferState::File { ref disk } => {
+                disk.enqueue(key, managed_envelope).await?;
+
+                // The disk is full, switch state to `BufferState::Memory` and try to buffer as
+                // much as we can before hitting OOM.
+                if disk.is_full().await.unwrap_or_default() {
+                    self.state = BufferState::to_memory_file_ro(
```
jjbayer: I'm questioning whether we actually need the `MemoryFileRead` state. This state is so extreme that we might as well just start dropping events. That would simplify the state machine a lot. But I might be missing something.

jjbayer (changes requested):
Pressing "Request changes" because of open questions.

relay-server/src/actors/spooler/mod.rs
olksdr: Isn't it a different api? I'm not sure how would you like to unify them? 
I could theoretically make a more generic function, which might take a closure to do some action, but it might even make the core a bit more complex.

relay-server/src/actors/spooler/mod.rs
olksdr: I don't think we need one

relay-server/src/actors/spooler/mod.rs
olksdr: What would the benefit of the trait be here? I do not think we need a defined interface right now. 
I also didn't want to enforce `async` on non-async functions, and write the trait with async functions, which also would make this code a bit more complex. 
But if there is a strong opinion that we have to have  an interface/trait, I can definitely do it.

relay-server/src/actors/spooler/mod.rs
olksdr: I can switch the condition , to check before `enqueue`, right now it's another way around. 
But also in case of errors while spooling we will lose the entire buffer or the items which cannot be spooled

relay-server/src/actors/spooler/mod.rs
olksdr: Only that it used in one place, which makes it easier to pass to function in one place only.

relay-server/src/actors/spooler/mod.rs
olksdr: I find the suggestions even more confusing and would prefer to have shorter names. 
`Memory` here is only when the disk is not configured, which makes it easier to work with. If the `Standby` state is re-used, I will have to wrap the disk in `Option`, which also will have to be treated differently?

relay-server/src/actors/spooler/mod.rs
olksdr: With `init` I wanted to  show that this is the initial state. But maybe `new` would work here as well

relay-server/src/actors/spooler/mod.rs
olksdr: So we want to make sure we spool to disk as much as we can, and then also use memory as much as we can before start rejecting incoming events.  `MemoryFileRead` signals us that the disk is full and can only be read from, but we still can try to fill up the memory, which right now is guarded by `cache.envelope_buffer_size`, but it will get deprecated and we will actually use the new config from `spool.envelopes.max_memory_size`. 
This is basically also a preparation for it. 
Does it makes sense or maybe I think about this somewhat wrong?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,330 @@ impl BufferService {
...
+            BufferState::Memory { ref mut ram } => ram.enqueue(key, managed_envelope),
+            // Will buffer till either there free space on disk, and the in-memory buffer can be
+            // spooled, or there is OOM.
+            // NOTE: Right now the number of envelopes is still restricted by
+            // `cache.envelope_buffer_size` and they will be dropped once the limit is reached.
+            BufferState::MemoryFileRead {
+                ref mut ram,
+                ref disk,
+            } => {
+                ram.enqueue(key, managed_envelope);
```
jjbayer, line 681: What happens when `ram.is_full()` in this case?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,345 @@ impl BufferService {
...
+    }
+
+    /// Becomes a different state, depdending on the current state and the current conditions of
+    /// underlying spool.
+    async fn update(&mut self, config: &Arc<Config>) -> Result<(), BufferError> {
+        match self {
+            BufferState::Memory { .. } => (),
+            BufferState::MemoryFileStandby { ram, disk } => {
+                if ram.is_full() {
+                    disk.spool(&mut ram.buffer, &mut ram.used_memory).await?;
```
jjbayer: nit: It would be nice if `update(self, config: ...)` consumed the state and returned a new state. Then you could pass `ram` by value into `spool`.

Added bonus: You might not even need the `to_` methods any longer, because you could pass the internals (in this case `disk`) into the new variant by value. E.g. in this case

`return BufferState::File { disk }`

But it would require implementing `Default` on `BufferState` so you can `std::mem::take` the whole thing, so I'm not sure if it's worth it.

relay-server/src/actors/spooler/mod.rs
jjbayer: Isn't `used_memory` dropped shortly after calling `spool`, because after `spool` we transition into disk mode?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,345 @@ impl BufferService {
...
+                    *self = BufferState::to_file(disk);
+                }
+            }
+            BufferState::File { disk } => {
+                if disk.is_empty().await.unwrap_or_default() {
```
jjbayer: Possible follow-up: As discussed offline, a `disk.is_below_low_watermark()` might be better here, to escape from disk mode if for some reason the disk never becomes fully empty.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,345 @@ impl BufferService {
...
+            disk.db.clone(),
+            disk.buffer_guard.clone(),
+            disk.max_disk_size,
+        );
+        BufferState::MemoryFileStandby { ram, disk }
+    }
+
+    /// Creates `BufferState::MemoryDiskRead` from provided disk reference and maximum in-memory
+    /// buffer size.
+    fn to_memory_file_ro(max_memory_size: usize, disk: &OnDisk) -> Self {
```
jjbayer:
```suggestion
    fn to_memory_file_read(max_memory_size: usize, disk: &OnDisk) -> Self {
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,345 @@ impl BufferService {
...
+        relay_statsd::metric!(counter(RelayCounters::BufferWrites) += 1);
+        Ok(())
+    }
+}
+
+/// The state which defines the [`BufferService`] behaviour.
+#[derive(Debug)]
+enum BufferState {
+    /// Only memory buffer is enabled.
+    Memory { ram: InMemory },
```
jjbayer: nit:
```suggestion
    Memory(InMemory),
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,345 @@ impl BufferService {
...
+    ///
+    /// The disk will be used when the memory limit will be hit.
+    MemoryFileStandby { ram: InMemory, disk: OnDisk },
+
+    /// THe disk is full and we use the memory to save incoming envelopes as much as we can. But
+    /// the disk still used to read from.
+    MemoryFileRead { ram: InMemory, disk: OnDisk },
+
+    /// Only disk used for read/write operations.
+    File { disk: OnDisk },
```
jjbayer: nit: I would call this variant `Disk`. The fewer different words, the better :)

```suggestion
    Disk(OnDisk),
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,345 @@ impl BufferService {
...
+        let mut state = BufferState::new(config.spool_envelopes_max_memory_size(), on_disk_state);
+
+        // When the old db file is picked up and it is not empty, we can also try to read from it
+        // on dequeue requests.
+        if let BufferState::MemoryFileStandby { disk, .. } = &state {
+            if !disk.is_empty().await.unwrap_or_default() {
+                state =
+                    BufferState::to_memory_file_ro(config.spool_envelopes_max_memory_size(), disk)
+            }
+        }
```
jjbayer: nit: Any reason this is not in `BufferState::new()`?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,345 @@ impl BufferService {
...
+            | BufferState::MemoryFileRead { ref mut ram, .. } => {
+                ram.enqueue(key, managed_envelope);
+                self.state.update(&self.config).await?;
+            }
+            BufferState::File { ref disk } => {
+                disk.enqueue(key, managed_envelope).await?;
+                self.state.update(&self.config).await?;
+            }
         }
+        Ok(())
```
jjbayer: nit: 
```suggestion
            }
            BufferState::File { ref disk } => {
                disk.enqueue(key, managed_envelope).await?;
            }
        }
        self.state.update(&self.config).await
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -525,33 +745,25 @@ impl BufferService {
...
+                self.state.update(&self.config).await?;
+            }
+            BufferState::MemoryFileRead {
+                ref mut ram,
+                ref disk,
+            } => {
+                ram.dequeue(&keys, sender.clone());
+                disk.dequeue(project_key, &mut keys, sender, &self.services)
+                    .await;
+                self.state.update(&self.config).await?;
```
jjbayer: nit: Same as above, could call `self.state.update()` after the `match` statement.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -501,18 +394,345 @@ impl BufferService {
...
+    Memory { ram: InMemory },
+
+    /// The memory buffer is used, but also disk spool is configured.
+    ///
+    /// The disk will be used when the memory limit will be hit.
+    MemoryFileStandby { ram: InMemory, disk: OnDisk },
+
+    /// THe disk is full and we use the memory to save incoming envelopes as much as we can. But
+    /// the disk still used to read from.
+    MemoryFileRead { ram: InMemory, disk: OnDisk },
```
jjbayer, line 478: As discussed offline, I think we should remove this state in a follow-up PR for the sake of simplicity. To me, two states seem to be enough:

1. Normal Operation a.k.a. `MemoryFileStandby`. Use a "small" in-memory buffer for fast performance.
2. Incident Mode a.k.a. Cannot Flush a.k.a. `File`: Use a large on-disk buffer. Sacrifice performance in order to save as much data as possible.

When even the disk buffer becomes full, we are in such an extreme state (were not able to resolve the incident after many hours or even days), that we might as well give up and start dropping events.

jjbayer (approved):
Left a couple of comments, none of them are blockers.

relay-server/src/actors/spooler/mod.rs
olksdr: yeah, you're right. Will remove this.

relay-server/src/actors/spooler/mod.rs
olksdr: Will address this in follow-up

relay-server/src/actors/spooler/mod.rs
olksdr: Yeah, in my first iteration, that's how it worked, i'm ok with returning this behaviour back.

relay-server/src/actors/spooler/mod.rs
olksdr: will address in the follow-up

relay-server/src/actors/spooler/mod.rs
olksdr: no particular reason, just did not want to make this method `async`, but I guess it's ok. will change.

olksdr (commented):
@jjbayer Addressed most of your comments, also will address some of them in the follow-ups as we discussed.

PR #2056, opened 2023-04-21
fix(profiles): Log ProfileIndexed outcome for invalid profiles

jjbayer:
Improvement to #2054:

Before a profile reaches the `process_profile` stage, we count it as `Profile:Accepted`. So if `process_profile` _fails_ with an `Invalid` outcome, it is incorrect to log it in the same data category, because it would double count within that category. Log it as `ProfileIndexed` instead.


tests/integration/test_outcome.py
```diff
@@ -1128,3 +1128,89 @@ def make_envelope(transaction_name, num_profiles):
...
+            "category": 6,  # Profile
+            "key_id": 123,
+            "org_id": 1,
+            "outcome": 0,  # Accepted
+            "project_id": 42,
+            "quantity": 1,
+            "source": "processing-relay",
+        },
+        {
+            "category": 11,  # ProfileIndexed
```
jjbayer, line 1197: Before the change in `envelope.rs`, this would report as `"category": 6,  # Profile`.

relay-server/src/envelope.rs
```diff
@@ -550,7 +550,11 @@ impl Item {
             ItemType::Metrics | ItemType::MetricBuckets => None,
             ItemType::FormData => None,
             ItemType::UserReport => None,
-            ItemType::Profile => Some(DataCategory::Profile),
+            ItemType::Profile => Some(if indexed {
+                DataCategory::ProfileIndexed
+            } else {
+                DataCategory::Profile
+            }),
```
jjbayer, line 557: `indexed` depends on whether metrics were extracted from a transaction. A profile counts towards `Profile` until metrics are extracted from the corresponding transaction. After that, it counts as `ProfileIndexed`.

phacops (approved):

olksdr (approved):

PR #2059, opened 2023-04-24
feat(filter): Add document_uri csp filter

### LINKED ISSUE #2037: Expand CSP filtering options
Our CSP filter works on blocked_uri and source_file: https://github.com/getsentry/relay/blob/1bd6764397a9a251d31ee1bd1b68f49269d1b895/relay-filter/src/csp.rs#L10-L34 

A customer has requested to add `document_uri` to filter, which can improve the CSP filters UX. One option is to allow users to choose which CSP fields they want to use with the filters.

Originally requested in: https://getsentry.atlassian.net/browse/FEEDBACK-1070

TBS1996:
#2037 
 simply adds document_uri to the list of fields the filter will search in.


jjbayer (approved):
From what I understand from https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP, this makes sense.

PR #2061, opened 2023-04-24
fix(profiles): Log sampling outcome as ProfileIndexed

jjbayer:
Improvement to https://github.com/getsentry/relay/pull/2054:

Since #2054 already introduced a translation from `Profile` to `ProfileIndexed` for sampling outcomes, this PR does not change anything functionally. It is merely an attempt to be consistent with #2056 and #2060, such that we consider the `metrics_extracted` flag any time we generate a profiling outcome.

### DIFF
--- a/relay-server/src/utils/managed_envelope.rs
+++ b/relay-server/src/utils/managed_envelope.rs
@@ -312,7 +312,11 @@ impl ManagedEnvelope {
         if self.context.summary.profile_quantity > 0 {
             self.track_outcome(
                 outcome,
-                DataCategory::Profile,
+                if self.context.summary.event_metrics_extracted {
+                    DataCategory::ProfileIndexed
+                } else {
+                    DataCategory::Profile
+                },
                 self.context.summary.profile_quantity,
             );
         }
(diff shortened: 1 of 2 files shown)


relay-server/src/utils/managed_envelope.rs
```diff
@@ -312,7 +312,11 @@ impl ManagedEnvelope {
         if self.context.summary.profile_quantity > 0 {
             self.track_outcome(
                 outcome,
-                DataCategory::Profile,
+                if self.context.summary.event_metrics_extracted {
+                    DataCategory::ProfileIndexed
+                } else {
+                    DataCategory::Profile
+                },
```
jjbayer, line 319: I did not add a separate integration test for this change, because https://github.com/getsentry/relay/blob/51888358224631ac085b71bb5dd49e87fab0fb9d/tests/integration/test_outcome.py#L981 already covers it. That test failed after adding the change in this file, and succeeded again after also making the change in `relay-server/src/actors/outcome.rs`.

olksdr (approved):

PR #2063, opened 2023-04-24
ref(replays): remove chunking logic for large recordings

JoshFerge:
After merging https://github.com/getsentry/relay/pull/2032, we can confirm via our metrics that we no longer hit this logic, as we reject payloads over 10 MiB and set our chunking limit to 15 MiB. 

_#skip-changelog_


jjbayer (approved):
In a follow-up PR, you could remove all the `*NotChunked` and `*_not_chunked` suffixes (for now it's probably good to keep them, so we don't get naming collisions with statsd metric names during rollout).

JoshFerge:
going ahead and merging this. will resolve your comments @jjbayer in a follow-up PR. thanks!

PR #2066, opened 2023-04-25
ref(dynamic-sampling): Refactor dynamic sampling code into the relay-sampling crate

### LINKED ISSUE #46742: Refactor dynamic sampling crate in Relay
**Description**

The Relay implementation of dynamic sampling is divided across multiple files which creates a problem.

With this ticket, we would like to refactor the code and bridge utils/dynamic_sampling with relay_sampling.

**Acceptance Criteria**

The dynamic sampling implementation is moved to utils/dynamic_sampling with relay_sampling.

iambriccardo:
This PR refactors the dynamic sampling logic into the `relay-sampling` crate in order to move most of the business logic out of `relay-server`. The main goal of this refactor is to have the bare minimum logic in `relay-server` which will only bridge the `ProjectState` which is not propagated to `relay-sampling`.

_This PR only moves functions and tests around, no functional changes are implemented._

Closes: https://github.com/getsentry/sentry/issues/46742


relay-sampling/src/lib.rs
```diff
@@ -878,11 +878,32 @@ impl Display for MatchedRuleIds {
     }
 }
 
+/// Checks whether unsupported rules result in a direct keep of the event or depending on the
+/// type of Relay an ignore of unsupported rules.
+fn check_unsupported_rules(
+    processing_enabled: bool,
+    sampling_config: &SamplingConfig,
+    root_sampling_config: Option<&SamplingConfig>,
+) -> Option<()> {
```
jjbayer: nit: I know this function was only moved, but maybe for a follow-up PR: I find a return type of `Option<()>` hard to read, I would prefer `Result<(), ()>`, and then map it to option with `.ok()` on the caller side.

jjbayer (approved):
I did not go through the changes line-by-line, but assuming that you only moved code around, this PR looks good to me!

Could you add a note to the PR description making clear that this only moves some functions and tests, without any functional changes?

relay-sampling/src/lib.rs
iambriccardo: I can do that! Thanks for the suggestion

PR #2069, opened 2023-04-26
feat(metrics): Extract metrics from spans

iker-barriocanal:
Internal-only mock-up to extract metrics from spans.

Left in the PR: apply this change only internally, with the current code metrics are extracted for all orgs.

#skip-changelog


relay-server/src/metrics_extraction/transactions/.mod.rs.pending-snap
olksdr, line 1: You might want to remove  or approve this file.

relay-metrics/src/protocol.rs
```diff
@@ -179,6 +179,8 @@ pub enum MetricNamespace {
     Sessions,
     /// Metrics extracted from transaction events.
     Transactions,
+    /// TODO(iker): docs
+    Spans,
```
olksdr, line 183: I think it's worth adding some docs in this PR, which does not have to be long.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -230,6 +231,10 @@ pub fn extract_transaction_metrics(
         sampling_metrics,
     )?;
 
+    if project_id.value() == 1 {
+        extract_span_metrics(aggregator_config, event, project_metrics)?;
```
olksdr, line 235: Why do we extract spans from within the `extract_transaction_metrics` function? 
Does it make sense actually to set the extraction up already in `relay-server/src/actors/processor.rs`, which mostly just calling same function just from the different place. 

The reason for this could be just to set up not a bit more which does not have to be moved in the later PRs.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -375,6 +380,76 @@ fn extract_transaction_metrics_inner(
     Ok(())
 }
 
+fn extract_span_metrics(
```
olksdr, line 386: Some docs on this function would be great

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -375,6 +380,76 @@ fn extract_transaction_metrics_inner(
...
+    };
+
+    // Validate the transaction event against the metrics timestamp limits. If the metric is too
+    // old or too new, we cannot extract the metric and also need to drop the transaction event
+    // for consistency between metrics and events.
+    if !aggregator_config.timestamp_range().contains(&timestamp) {
+        relay_log::debug!("event timestamp is out of the valid range for metrics");
+        return Err(ExtractMetricsError::InvalidTimestamp);
+    }
+
```
olksdr, line 413: This part (all the validation thingy) could be refactored in the separate helper function, since it's already used in the `extract_transaction_metrics_inner` in the same way

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -375,6 +380,76 @@ fn extract_transaction_metrics_inner(
...
+        let mut shared_tags = BTreeMap::new();
+
+        if let Some(transaction_name) = event.transaction.value() {
+            shared_tags.insert("transaction".to_owned(), transaction_name.to_owned());
+        }
+        // TODO(iker): must add shared tags: environment, op, release
+
+        for annotated_span in spans {
+            if let Some(span) = annotated_span.value_mut() {
+                // TODO(iker): add span-specific tags here
```
olksdr: For tags we could use something similar to `extract_universal_tags`, to prepare the common tags for each incoming event.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -375,6 +380,76 @@ fn extract_transaction_metrics_inner(
...
+
+                // Add to the span all the tags that exist in the metrics
+                let span_tags = span.tags.get_or_insert_with(BTreeMap::new);
+                for (key, value) in &shared_tags {
+                    // NOTE(iker): if the tag already existed before, we are overwriting it
+                    span_tags.insert(
+                        key.to_owned(),
+                        Annotated::new(JsonLenientString::from(value.to_owned())),
+                    );
+                }
```
olksdr: Instead of this you might be able to use [`extend`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.extend)

something like 
```suggestion
                // Add to the span all the tags that exist in the metrics
                let mut span_tags = span.tags.get_or_insert_with(BTreeMap::new);
                span_tags.extend(shared_tags);
```
Did not test it , but I guess it should work.

olksdr (changes requested):
There are few things which would be great to address in this PR, and some we can leave for followups. 
Also left few questions.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -230,6 +231,10 @@ pub fn extract_transaction_metrics(
         sampling_metrics,
     )?;
 
+    if project_id.value() == 1 {
```
jan-auer: Please make this a feature flag right away and set this on sentry side.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -375,6 +380,76 @@ fn extract_transaction_metrics_inner(
...
+            shared_tags.insert("transaction".to_owned(), transaction_name.to_owned());
+        }
+        // TODO(iker): must add shared tags: environment, op, release
+
+        for annotated_span in spans {
+            if let Some(span) = annotated_span.value_mut() {
+                // TODO(iker): add span-specific tags here
+
+                if let Some(user) = event.user.value() {
+                    if let Some(value) = get_eventuser_tag(user) {
```
jan-auer: The value of the event's user doesn't change between spans, so this can be lifted out of the loop. Or alternatively be guarded so that it only emits once on the first span.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: The type of the value of `BTreeMap` is `String`, and `span_tags`'s value is of type `Annotated<JsonLenientString>`, so afaik it's not possible to extend it.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: Yes, that's an improvement I decided to leave out of scope for now, and that we can iterate on later.

relay-server/src/metrics_extraction/transactions/mod.rs
olksdr: It's possible `map` the data and use that iterator to extend the `span_tags` in one go, which also will be must prettier and faster.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: I'd like to do this in the processor. There's still a function in there that needs refactoring to not be tied up with the transaction metrics.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: Yes, I agree. I think there's some refactoring that must happen from this PR.

relay-server/src/metrics_extraction/transactions/mod.rs
olksdr: I have to agree with @jan-auer  on this. Please, add the feature flag into this PR, it's a very minimal change, but will greatly improve the quality of life right away.

relay-server/src/metrics_extraction/transactions/mod.rs
olksdr: I still think this looks better 😄 , which also could be done only once, before the metrics are generated and the tags then populated in all the future metric instances. 
```suggestion
                // Add to the span all the tags that exist in the metrics
                let span_tags = span.tags.get_or_insert_with(BTreeMap::new);
                span_tags.extend(
                    shared_tags
                        .clone()
                        .into_iter()
                        .map(|(k, v)| (k, Annotated::new(JsonLenientString::from(v)))),
                );

                
```

iker-barriocanal: Looks better, thank you! Done in [00fdeac](https://github.com/getsentry/relay/pull/2069/commits/00fdeacc95ca81455552756c3eed5f3e04771147).

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -375,6 +380,79 @@ fn extract_transaction_metrics_inner(
...
+                            MetricUnit::None,
+                            MetricValue::set_from_str(&value),
+                            timestamp,
+                            shared_tags.clone(),
+                        ));
+                    }
+                }
+
+                // Add to the span all the tags that exist in the metrics
+                let span_tags = span.tags.get_or_insert_with(BTreeMap::new);
```
jjbayer: Best check this with @mitsuhiko, but I feel it would be safer to put a new data bag called `metrics_tags` or similar on the span. I don't know what `Span::tags` is currently used for, but using a different data bag would help prevent any potential conflicts.

iker-barriocanal: After a chat with Jan, this is out of scope for now.

relay-dynamic-config/src/feature.rs
```diff
@@ -15,6 +15,8 @@ pub enum Feature {
     ///
     /// Enables device.class tag synthesis on mobile events.
     DeviceClassSynthesis,
+    /// Enables metric extraction from spans.
+    SpanMetricExtraction,
```
olksdr: Do we want to align this name with the feature name? 
```suggestion
    SpanMetricsExtraction,
```

olksdr (approved):

iker-barriocanal:
Merging this as-is to have the structure on master, but note that the user metric is still incomplete.

PR #2070, opened 2023-04-26
feat(dynamic-sampling): Add support for `trace.replay_id` in matching rules

### LINKED ISSUE #47769: Replay Bias
Ensure that transactions with a Replay ID on the DSC are sampled so that the trace view for any given Replay is complete.
- Jernej asked us to give heads-up in #discuss-ingest once we start working on this and give a little context to the team

iambriccardo:
This PR adds the ability to read the `replay_id` field in the `DynamicSamplingContext`.

Closes: https://github.com/getsentry/sentry/issues/47769

### DIFF
--- a/relay-sampling/src/lib.rs
+++ b/relay-sampling/src/lib.rs
@@ -316,27 +316,6 @@ impl NotCondition {
     }
 }
 
-/// Has condition combinator.
-///
-/// Creates a condition that is true when a specific field is present.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct HasCondition {
-    field: String,
-}
-
-impl HasCondition {
-    fn supported(&self) -> bool {
-        true
-    }
-
-    fn matches<T>(&self, value: &T, ip_addr: Option<IpAddr>) -> bool
-    where
-        T: FieldValueProvider,
-    {
-        !value.get_value(&self.field).is_null()
-    }
-}
-
 /// A condition from a sampling rule.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase", tag = "op")]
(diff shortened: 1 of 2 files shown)


relay-sampling/src/lib.rs
```diff
@@ -316,6 +316,27 @@ impl NotCondition {
     }
 }
 
+/// Has condition combinator.
+///
+/// Creates a condition that is true when a specific field is present.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct HasCondition {
+    field: String,
+}
```
jjbayer: AFAIK This should also work with a composite `{"not": {"eq": null}}` condition. The reason why I would prefer that is that it is already supported by external relays.

jjbayer (changes requested):

relay-sampling/src/lib.rs
iambriccardo: Semantically it is equivalent yes, I can for sure change that. Thanks for the heads up!

relay-sampling/src/lib.rs
```diff
@@ -856,6 +856,10 @@ impl FieldValueProvider for DynamicSamplingContext {
                 None => Value::Null,
                 Some(ref s) => s.as_str().into(),
             },
+            "trace.replay_id" => match self.replay_id {
+                None => Value::Null,
+                Some(ref s) => Value::String(s.to_string()),
```
jjbayer, line 861: nit: For consistency with the code above

```suggestion
                Some(ref s) => s.as_str().into(),
```

jjbayer (approved):

relay-sampling/src/lib.rs
iambriccardo: I have tried that but `Uuid` has no method `as_str()`.

PR #2071, opened 2023-04-26
fix(profiles): Only use `ProfileIndexed` after counting as `Profile`

jjbayer:
In #2056, #2060 and #2061, I wrongly assumed that dropped profiles should always be counted as `ProfileIndexed` after metrics extraction and dynamic sampling, in order to not double-count towards `Profile`. This is wrong, because as the PR description of #2054 clearly states, profiles that are not dropped by dynamic sampling are only counted as accepted in processing relays, _after_ dynamic sampling and rate limiting:

> In processing Relays, if an envelope still contains profiles after dynamic sampling, log an Accepted outcome for the "processed" category. By restricting this to processing Relays, we can be sure that every profile is only counted once.

Instead of checking the `metrics_extracted` flag, introduce a new flag that explicitly states whether a profile was already counted towards `DataCategory::Profile` or not, and evaluate that flag instead of `metrics_extracted`.

### DIFF
--- a/relay-server/src/envelope.rs
+++ b/relay-server/src/envelope.rs
@@ -481,6 +481,13 @@ pub struct ItemHeaders {
     #[serde(default, skip_serializing_if = "is_false")]
     metrics_extracted: bool,
 
+    /// Internal flag to signal that the profile has been counted toward `DataCategory::Profile`.
+    ///
+    /// If `true`, outcomes for this item should be reported as `DataCategory::ProfileIndexed`
+    /// instead.
+    #[serde(skip)]
+    profile_counted_as_processed: bool,
+
     /// Other attributes for forward compatibility.
     #[serde(flatten)]
     other: BTreeMap<String, Value>,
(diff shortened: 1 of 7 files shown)


relay-server/src/envelope.rs
```diff
@@ -550,7 +558,9 @@ impl Item {
             ItemType::Metrics | ItemType::MetricBuckets => None,
             ItemType::FormData => None,
             ItemType::UserReport => None,
-            ItemType::Profile => Some(if indexed {
+            // For profiles, do not used the `indexed` flag, because it depends
```
olksdr, line 561:
```suggestion
            // For profiles, do not use the `indexed` flag, because it depends
```

phacops (approved):

olksdr (approved):

relay-server/src/envelope.rs
jjbayer: Thanks, will address in follow-up.

PR #2073, opened 2023-04-27
feat(crons): Pass-through minimal `sdk` field

evanpurkhiser:
We would like to collect more information about where cron checkins are coming from. Since we're not using `metastructure` yet we can't use the `ClientSdkInfo` struct 

https://github.com/getsentry/relay/blob/3ffdaf1c108ab44d1fea6f73874ddee351975e5d/relay-general/src/protocol/clientsdk.rs#L18-L65

#skip-changelog

### DIFF
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -100,6 +100,17 @@ struct MonitorConfig {
     timezone: Option<String>,
 }
 
+// XXX(epurkhiser): This is a duplicate of the ClientSdkInfo that is part of the relay-general
+// crate. Until we're able to migrate the checkin payload over to it's own protocol using
+// metastructure we'll need to have this duplicated here
+//
+/// The SDK Interface describes the Sentry SDK and its configuration used to capture and transmit an event.
+#[derive(Debug, Deserialize, Serialize)]
+struct MinimalClientSdkInfo {
+    pub name: String,
+    pub version: String,
+}
+
 /// The monitor check-in payload.
 #[derive(Debug, Deserialize, Serialize)]
 struct CheckIn {
@@ -112,6 +123,10 @@ struct CheckIn {
 
     /// Status of this check-in. Defaults to `"unknown"`.
     status: CheckInStatus,
+
+    /// monitor configuration to support upserts.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    sdk: Option<MinimalClientSdkInfo>,
 
     /// The environment to associate the check-in with
     #[serde(default, skip_serializing_if = "Option::is_none")]
@@ -176,6 +191,10 @@ mod tests {
   "check_in_id": "a460c25ff2554577b920fcfacae4e5eb",
   "monitor_slug": "my-monitor",
   "status": "in_progress",
+  "sdk": {
+    "name": "sentry.rust",
+    "version": "1.0.0"
+  },
   "environment": "production",
   "duration": 21.0
 }"#;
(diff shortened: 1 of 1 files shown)


relay-monitors/src/lib.rs
```diff
@@ -100,6 +100,17 @@ struct MonitorConfig {
...
 
+// XXX(epurkhiser): This is a duplicate of the ClientSdkInfo that is part of the relay-general
+// crate. Until we're able to migrate the checkin payload over to it's own protocol using
+// metastructure we'll need to have this duplicated here
+//
+/// The SDK Interface describes the Sentry SDK and its configuration used to capture and transmit an event.
+#[derive(Debug, Deserialize, Serialize)]
+struct MinimalClientSdkInfo {
+    pub name: String,
+    pub version: String,
```
jjbayer, line 111: Are we sure these two values will always be there? If not, I would make them optional: 

```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub name: Option<String>,
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub version: Option<String>,
```

relay-monitors/src/lib.rs
```diff
@@ -100,6 +100,17 @@ struct MonitorConfig {
...
+}
```
jjbayer, line 112: This is not strictly necessary, but if you want outdated external Relays to forward other attributes later on, it comes in handy:

```suggestion

    /// Other attributes for forward compatibility.
    #[serde(flatten)]
    other: BTreeMap<String, Value>,
}
```

relay-monitors/src/lib.rs
```diff
@@ -113,6 +124,9 @@ struct CheckIn {
     /// Status of this check-in. Defaults to `"unknown"`.
     status: CheckInStatus,
 
+    /// monitor configuration to support upserts.
+    sdk: MinimalClientSdkInfo,
```
jjbayer: Old SDKs won't send this, right? So I think this needs to be optional, otherwise deserialization will fail. Make sure to also add a test case to verify this.

```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    sdk: Option<MinimalClientSdkInfo>,
```

jjbayer (changes requested):

relay-monitors/src/lib.rs
evanpurkhiser: Old SDKs don’t support sending monitor checkins however

relay-monitors/src/lib.rs
evanpurkhiser: Does this mean they could shove as much as they want into the payload now however?

relay-monitors/src/lib.rs
evanpurkhiser: Both of these fields are required in the `ClientSdkInfo` struct

https://github.com/getsentry/relay/blob/3ffdaf1c108ab44d1fea6f73874ddee351975e5d/relay-general/src/protocol/clientsdk.rs#L19-L38

relay-monitors/src/lib.rs
jan-auer: For now this would be an overkill. Best to be added when moving to meta structure.

relay-monitors/src/lib.rs
jan-auer: Did we not release an SDK version already with this? If not we can skip this, but if we did, we should make this optional.

jan-auer (approved):

relay-monitors/src/lib.rs
evanpurkhiser: Chatted w/ @jan-auer I'll go ahead and make this optional

jjbayer (approved):

PR #2074, opened 2023-04-27
feat(metrics): Extract more tags for span metrics

iker-barriocanal:
Follow-up to https://github.com/getsentry/relay/pull/2069.

Adds more tags for the extracted metrics from spans.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -411,37 +411,45 @@ fn extract_span_metrics(
         return Err(ExtractMetricsError::InvalidTimestamp);
     }
 
+    if event.spans.value().is_none() {
+        return Ok(());
+    }
+
+    // Collect the shared tags for all the metrics and spans on this transaction
+    let mut shared_tags = BTreeMap::new();
+
+    if let Some(environment) = event.environment.as_str() {
+        shared_tags.insert("environment".to_owned(), environment.to_owned());
+    }
+
+    if let Some(transaction_name) = event.transaction.value() {
+        shared_tags.insert("transaction".to_owned(), transaction_name.to_owned());
+    }
+
+    if let Some(trace_context) = get_trace_context(event) {
+        if let Some(op) = extract_transaction_op(trace_context) {
+            shared_tags.insert("transaction.op".to_owned(), op);
+        }
+
+        shared_tags.insert(
+            "transaction.status".to_owned(),
+            extract_transaction_status(trace_context).to_string(),
+        );
+    }
+
+    // server_name is extracted into an event tag during light_normalization
+    if let Some(event_tags) = event.tags.value() {
+        if let Some(server_name) = event_tags.get("server_name") {
+            shared_tags.insert("domain".to_owned(), server_name.to_owned());
+        }
+    }
+
+    // Some more tags are required, but need SDK work first:
+    // - span_status
+    // - action
+    // - event platform (e.g. MySQL, Redis).
+
     if let Some(spans) = event.spans.value_mut() {
-        // Collect the shared tags for all the metrics and spans on this transaction
-        let mut shared_tags = BTreeMap::new();
-
-        if let Some(environment) = event.environment.as_str() {
-            shared_tags.insert("environment".to_owned(), environment.to_owned());
-        }
-        if let Some(transaction_name) = event.transaction.value() {
-            shared_tags.insert("transaction".to_owned(), transaction_name.to_owned());
-        }
-        if let Some(trace_context) = get_trace_context(event) {
-            if let Some(op) = extract_transaction_op(trace_context) {
-                shared_tags.insert("transaction.op".to_owned(), op);
-            }
-
-            shared_tags.insert(
-                "transaction.status".to_owned(),
-                extract_transaction_status(trace_context).to_string(),
-            );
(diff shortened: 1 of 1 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -411,37 +411,45 @@ fn extract_span_metrics(
         return Err(ExtractMetricsError::InvalidTimestamp);
     }
 
-    if let Some(_spans) = event.spans.value_mut() {
-        // Collect the shared tags for all the metrics and spans on this transaction
-        let mut shared_tags = BTreeMap::new();
+    if event.spans.value().is_none() {
+        return Ok(());
+    }
```
jjbayer, line 416: Couldn't we put a guard like this at the beginning of the function:

```rust
let Some(spans) = event.spans.value() else { return Ok(()); }
```

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -455,8 +463,36 @@ fn extract_span_metrics(
...
+                        None
+                    };
+                    if let Some(module) = span_module {
+                        span_tags.insert("span.module".to_owned(), module.to_owned());
+                    }
+                }
+
+                if let Some(description) = span.description.value() {
+                    // TODO(iker): needs sanitization
+                    span_tags.insert("span.description".to_owned(), description.to_owned());
```
jjbayer: Where are these `span_tags` used?

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -411,37 +411,45 @@ fn extract_span_metrics(
...
-        // - action
-        // - event platform (e.g. MySQL, Redis).
+    }
+
+    // Some more tags are required, but need SDK work first:
+    // - span_status
+    // - action
+    // - event platform (e.g. MySQL, Redis).
 
+    if let Some(spans) = event.spans.value_mut() {
```
jjbayer: Do you need this mutable?

jjbayer (changes requested):

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: This one is tricky. Since we'll need mutable references to the spans, but we need to extract metrics from the event first, we can't hold a mutable reference to the spans and a reference to the event. I'll try to clean this up next week.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: The user metric above is logged once per transaction. This is incorrect as per the current requirements, and metrics will be logged per span. This is WIP coming in a follow-up PR.

relay-server/src/metrics_extraction/transactions/mod.rs
olksdr: @jjbayer  is right, following  
```rust
let Some(spans) = event.spans.value() else { return Ok(())};
```
works and looks much better with early return

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -411,37 +411,45 @@ fn extract_span_metrics(
...
-            shared_tags.insert("environment".to_owned(), environment.to_owned());
-        }
-        if let Some(transaction_name) = event.transaction.value() {
-            shared_tags.insert("transaction".to_owned(), transaction_name.to_owned());
-        }
-        if let Some(trace_context) = get_trace_context(event) {
-            if let Some(op) = extract_transaction_op(trace_context) {
-                shared_tags.insert("transaction.op".to_owned(), op);
-            }
+    // Collect the shared tags for all the metrics and spans on this transaction
```
olksdr:
```suggestion
    // Collect the shared tags for all the metrics and spans on this transaction.
```

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -411,37 +411,45 @@ fn extract_span_metrics(
...
-                shared_tags.insert("domain".to_owned(), server_name.to_owned());
-            }
+
+        shared_tags.insert(
+            "transaction.status".to_owned(),
+            extract_transaction_status(trace_context).to_string(),
+        );
+    }
+
+    // server_name is extracted into an event tag during light_normalization
```
olksdr:
```suggestion
    // The `server_name` is extracted into an event tag during light normalization. 
```

olksdr:
Where is this branch is merged to? Why not `master` ?

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: It compiles now because the spans aren't a mutable reference. This will change when we add metric tags to the spans, and will not work then. Line 429 needs a reference to the event, which is not possible if prior to that we get a mutable reference to the spans.

I'll clean the code up when we unblock other teams.

iker-barriocanal:
@olksdr I've stacked the PR to make the diff smaller. Both PRs, this one and the base, are required.

iker-barriocanal:
Closing in favor of https://github.com/getsentry/relay/pull/2080.

PR #2076, opened 2023-04-27
fix(ci): Skip flaky integration test

jjbayer:
#skip-changelog


olksdr (approved):

iker-barriocanal (approved):
Could we create a ticket to prioritize and keep track of fixing this, if there isn't one already?

jjbayer:
> Could we create a ticket to prioritize and keep track of fixing this, if there isn't one already?

Good point. https://github.com/getsentry/relay/issues/2078

PR #2077, opened 2023-04-27
instr(server): More details on unexpected envelope drop

jjbayer:
Get more insights into dropped envelopes, see https://github.com/getsentry/team-ingest/issues/112.

It should be enough to deploy this on Canary without merging.

#skip-changelog

### DIFF
--- a/relay-server/src/utils/managed_envelope.rs
+++ b/relay-server/src/utils/managed_envelope.rs
@@ -290,7 +290,26 @@ impl ManagedEnvelope {
         let handling = Handling::from_outcome(&outcome);
         match handling {
             Handling::Success => relay_log::debug!("dropped envelope: {}", outcome),
-            Handling::Failure => relay_log::error!("dropped envelope: {}", outcome),
+            Handling::Failure => {
+                let summary = &self.context.summary;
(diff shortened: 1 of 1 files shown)


relay-server/src/utils/managed_envelope.rs
```diff
@@ -290,7 +290,26 @@ impl ManagedEnvelope {
...
+                        scope.set_tag("has_sessions", summary.session_quantity > 0);
+                        scope.set_tag("has_profiles", summary.profile_quantity > 0);
+                        scope.set_tag("has_replays", summary.replay_quantity > 0);
+                        scope.set_tag("has_checkins", summary.checkin_quantity > 0);
+                        scope.set_tag("event_category", format!("{:?}", summary.event_category));
+                        scope.set_extra("cached_summary", format!("{:?}", summary).into());
+                        scope.set_extra(
+                            "recomputed_summary",
+                            format!("{:?}", EnvelopeSummary::compute(self.envelope())).into(),
+                        );
```
jjbayer, line 308: If the error is a false alarm, it might be because the envelope summary is not up-to-date.

olksdr (approved):

PR #2080, opened 2023-04-28
feat(metrics): Extract more metrics from spans

iker-barriocanal:
Follow-up to https://github.com/getsentry/relay/pull/2069.

Adds the remaining metrics with their tags.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -390,10 +390,12 @@ fn extract_span_metrics(
 ) -> Result<(), ExtractMetricsError> {
     // TODO(iker): measure the performance of this whole method
 
+    let Some(spans) = event.spans.value() else { return Ok(())};
+
     if event.ty.value() != Some(&EventType::Transaction) {
         return Ok(());
     }
-    let (Some(_), Some(&end)) = (event.start_timestamp.value(), event.timestamp.value()) else {
+    let (Some(&start), Some(&end)) = (event.start_timestamp.value(), event.timestamp.value()) else {
         relay_log::debug!("failed to extract the start and the end timestamps from the event");
         return Err(ExtractMetricsError::MissingTimestamp);
     };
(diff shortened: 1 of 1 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -411,29 +413,104 @@ fn extract_span_metrics(
...
+                    span_tags.insert("module".to_owned(), module.to_owned());
+                }
+
+                let action = match span_module {
+                    Some("http") => span
+                        .data
+                        .value()
+                        // TODO(iker): some SDKs extract this as method
+                        .and_then(|v| v.get("http.method"))
+                        .and_then(|db_op| db_op.as_str()),
```
jjbayer:
```suggestion
                        .and_then(|method| method.as_str()),
```

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -411,29 +413,104 @@ fn extract_span_metrics(
...
+                    _ => None,
+                };
+
+                if let Some(act) = action {
+                    span_tags.insert("action".to_owned(), act.to_owned());
+                }
+
+                // TBD product requirements, adding it as a placeholder
+                span_tags.insert("domain".to_owned(), "todo".to_owned());
```
jjbayer: Does it even make sense to set the tag if we do not know the value yet?

jjbayer (approved):
LGTM

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: I discussed this with Jan, and he suggested adding it as a placeholder. I don't think it really helps, so I removed it; we could apply the same reasoning for the other unimplemented tags I assume.

TBS1996:
would it make sense to also make a `enum SpanMetric` in the way we do with `TrasnactionMetric` and `SessionMetric`? https://github.com/getsentry/relay/blob/dda830c29fbfd9d5cf79efcad59dfd68f7ce1bfd/relay-server/src/metrics_extraction/transactions/types.rs#L13

iker-barriocanal:
@TBS1996 Yes, it does. That is a follow-up step after we're no longer a blocker.

PR #2082, opened 2023-04-28
feat(general): Catch invalid metric names early

### LINKED ISSUE #1788: Catch invalid metric name early
Invalid measurement names and or values should be caught by event parsing or event normalization, not metrics aggregation. See

https://sentry.my.sentry.io/organizations/sentry/issues/391393

TBS1996:
Catch invalid measurement names early on (in light_normalize) instead of later at MetricsAggregation

#skip-changelog

#1788


relay-general/src/store/transactions/processor.rs
```diff
@@ -396,6 +397,20 @@ impl Processor for TransactionsProcessor<'_> {
...
     }
 
+    fn process_measurements(
+        &mut self,
+        value: &mut crate::protocol::Measurements,
+        _meta: &mut Meta,
+        _state: &ProcessingState<'_>,
+    ) -> ProcessingResult {
+        let measurements = &mut value.0;
+        let huge_number: usize = todo!();
```
TBS1996: want to get max length from aggregatorconfig, since that's how it's done in `fn validate_metric_name`, not sure how this can be done, especially within a trait method like this

relay-general/src/store/normalize.rs
```diff
@@ -693,6 +693,8 @@ pub struct LightNormalizationConfig<'a> {
     pub received_at: Option<DateTime<Utc>>,
     pub max_secs_in_past: Option<i64>,
     pub max_secs_in_future: Option<i64>,
+    pub max_metric_name_length: Option<usize>,
+    pub fixed_metric_name_length: Option<usize>,
```
TBS1996: theyre optional because values aren't available when light_normalize_config is called from relay::cabi crate

relay-general/src/store/transactions/processor.rs
```diff
@@ -396,6 +407,45 @@ impl Processor for TransactionsProcessor<'_> {
...
+        &mut self,
+        value: &mut crate::protocol::Measurements,
+        _meta: &mut Meta,
+        _state: &ProcessingState<'_>,
+    ) -> ProcessingResult {
+        let measurements = &mut value.0;
+
+        measurements.retain(|key, val| {
+            if !is_valid_metric_name(key) {
+                relay_log::error!("Measurement name contains invalid characters");
```
TBS1996: is this the right way to log? do we even wanna log it (and if so, as an error) ?

relay-general/src/store/transactions/processor.rs
```diff
@@ -396,6 +407,45 @@ impl Processor for TransactionsProcessor<'_> {
...
+                return false;
+            }
+
+            let name_len = key.len();
+
+            let unit_len = val
```
TBS1996: which scenarios is the unit value missing from Measurement? and in those scenarios, do we want to still keep the measurement or should we change this to return false?

relay-system/Cargo.toml
```diff
@@ -14,8 +14,8 @@ futures = "0.3"
 once_cell = "1.13.1"
 relay-log = { path = "../relay-log" }
 relay-statsd = { path = "../relay-statsd" }
-tokio = { version = "1.0", features = ["macros", "signal"] }
+tokio = { version = "1.0", features = ["macros", "signal", "full"] }
```
TBS1996: for some reason, I was unable to run unit tests individually without changing this (compilation error), but i still could do `make test-rust-all` without problem

TBS1996: so i'll remove the dependency later when i figure out what the deal is

relay-general/src/store/normalize.rs
```diff
@@ -714,8 +716,11 @@ pub fn light_normalize_event(
         // (internally noops for non-transaction events).
         // TODO: Parts of this processor should probably be a filter so we
         // can revert some changes to ProcessingAction)
-        let mut transactions_processor =
-            transactions::TransactionsProcessor::new(config.transaction_name_config);
+        let mut transactions_processor = transactions::TransactionsProcessor::new(
+            config.transaction_name_config,
+            config.max_metric_name_length,
+            config.fixed_metric_name_length,
```
iker-barriocanal: Do you think it's worth having a config structure for metric names, as we do for transaction names?

relay-general/src/store/transactions/processor.rs
iker-barriocanal: I think you can avoid the length logic by adding `max_chars` to the metastructure, see [example](https://github.com/getsentry/relay/blob/4e50f9e90ab7c2a21ca1778ddf2a5357e5ce75bf/relay-general/src/protocol/replay.rs#L211) (or look for other examples too).

relay-general/src/store/transactions/processor.rs
iker-barriocanal: If there's an error processing a measurement, this function should return an `Err(ProcessingAction::`.

relay-general/src/store/transactions/processor.rs
```diff
@@ -525,6 +583,108 @@ mod tests {
...
+
+        event.measurements = Annotated::from(Measurements(measurements));
+
+        let mut processor = TransactionsProcessor {
+            max_metric_name_len: Some(50),
+            fixed_metric_len: Some(20),
+            ..Default::default()
+        };
+
+        assert_eq!(event.measurements.0.clone().unwrap().0.len(), 1);
```
iker-barriocanal: We should try to simplify this expression a bit. I think you may be confused because of the `Annotated` API. Have a look at the `value()` function to get rid of `.0`.

relay-server/src/actors/processor.rs
```diff
@@ -59,12 +59,42 @@ use crate::envelope::{AttachmentType, ContentType, Envelope, Item, ItemType};
...
     self, get_sampling_key, log_transaction_name_metrics, ChunkedFormDataAggregator, FormDataIter,
     ItemAction, ManagedEnvelope, SamplingResult,
 };
+use once_cell::sync::Lazy;
+
+/// In 'fn process_measurements' we want to check if a measurement name is too long, to do that we
+/// check with a maximum length provided by the config, but that length represents the entire MRI name,
+/// and in process_measurements we only have access to the name part, and the unit part, so here
+/// we calculate the length of the remainder part. which isn't changing so we only need to do it once.
+static FIXED_MEASUREMENT_LEN: Lazy<usize> = Lazy::new(|| {
```
iker-barriocanal: I'm confused by the purpose of this variable. If it was used for local testing purposes, we should get rid of it before landing this change.

relay-system/Cargo.toml
iker-barriocanal: Let's remove it from the diff. If this change exists at this point, the intention is to land it. If you use it for local testing, don't stage it with git.

As for the compilation error you were getting, what is the error?

relay-metrics/src/aggregation.rs
```diff
@@ -1931,7 +1931,7 @@ impl AggregatorService {
         } = msg;
         for metric in metrics {
             if let Err(err) = self.insert(project_key, metric) {
-                relay_log::error!("failed to insert mertrics: {}", LogError(&err));
+                relay_log::error!("failed to insert metrics: {}", LogError(&err));
```
iker-barriocanal: nit: I think we should leave these changes out of scope for PRs that have a different goal and work on other parts.

relay-general/src/store/transactions/processor.rs
iker-barriocanal: It's also hard to know what asserting to `1` means. Does that mean it's ok? Or that the length is 1? If it's not ok, what's the error, or why is it failing?

iker-barriocanal (changes requested):

relay-general/src/store/transactions/processor.rs
TBS1996: good points, I changed to value() and made a comment on what the 1 and 0 represents

relay-server/src/actors/processor.rs
TBS1996: It's not for testing, but i agree it's confusing (thats why i made big description). Basically the value we use to check if a metric name is too long actually represents the entire MRI string, but in the process function we only have access to the name and the unit, you gotta also know how big the other parts are, so this function basically tries to compute the size of MRI without unit and name for measurements.

i could have hardcoded it but that would make maintaining harder in case we change something

TBS1996: I'd prefer to have this in the same file as `process_measurements` but for that i might have to move a lot of stuff around as it has dependencies on `relay-server` which can't be a dependency of `relay-general`

relay-general/src/store/transactions/processor.rs
TBS1996: hmm how would it work since it's looping over a lot of values, do you mean return Err immediatly if a measurement is deleted or in the end if any invalid ones are found?

relay-general/src/store/transactions/processor.rs
```diff
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 
 use relay_common::SpanStatus;
+use relay_metrics::is_valid_metric_name;
```
TBS1996: is it bad to import another crate just for this small function?

relay-general/src/store/transactions/processor.rs
TBS1996: interesting, im not familiar with it, does it allow to change it from config like the current version?

relay-metrics/src/aggregation.rs
TBS1996: ok changed it back, but then should I create a separate PR just for tiny typos?

relay-system/Cargo.toml
TBS1996: so when i do `make test-rust-all` it works as expected, when i do `run tests` within `processor.rs` I get a lot of errors from `relay-system` relating to tokio not finding certain objects

excerpt:

```
   Compiling relay-system v23.4.0 (/Users/tor/prog/rust/relay/relay-system)
error[E0432]: unresolved import `tokio::runtime::Runtime`
  --> relay-system/src/service.rs:12:5
   |
12 | use tokio::runtime::Runtime;
   |     ^^^^^^^^^^^^^^^^^^^^^^^ no `Runtime` in `runtime`
   |
help: consider importing this variant instead
   |
12 | use relay_log::protocol::Context::Runtime;
   |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error[E0432]: unresolved imports `tokio::sync::mpsc`, `tokio::sync::oneshot`
  --> relay-system/src/service.rs:13:19
   |
13 | use tokio::sync::{mpsc, oneshot};
   |                   ^^^^  ^^^^^^^ no `oneshot` in `sync`
   |                   |
   |                   no `mpsc` in `sync`
   |
help: consider importing this module instead
   |
13 | use tokio::sync::{std::sync::mpsc;
   |                   ~~~~~~~~~~~~~~~~

error[E0425]: cannot find function `spawn` in crate `tokio`
   --> relay-system/src/controller.rs:165:16
    |
165 |         tokio::spawn(monitor_shutdown(shutdown_timeout));
    |                ^^^^^ not found in `tokio`
    |
help: consider importing this function
    |
1   | use std::thread::spawn;
    |
help: if you import `spawn`, refer to it directly
    |
165 -         tokio::spawn(monitor_shutdown(shutdown_timeout));
165 +         spawn(monitor_shutdown(shutdown_timeout));
    |

error[E0603]: module `sync` is private
   --> relay-system/src/controller.rs:5:12
    |
5   | use tokio::sync::watch;
    |            ^^^^ private module
    |
```

relay-general/src/store/transactions/processor.rs
jjbayer: This would log an error for _us_, but we actually want to inform the user about this. So best write to `Meta`, like we do here: https://github.com/getsentry/relay/blob/0575289402d1cac2c29485a7dee8fc151ce39014/relay-general/src/store/normalize.rs#L295-L306

relay-general/src/store/transactions/processor.rs
```diff
@@ -396,6 +407,45 @@ impl Processor for TransactionsProcessor<'_> {
...
+
+            if let (Some(fixed_len), Some(max_len), Some(unit_len)) =
+                (self.fixed_metric_len, self.max_metric_name_len, unit_len)
+            {
+                if (name_len + unit_len + fixed_len) > max_len {
+                    relay_log::error!(
+                        "Measurement name is too long({}/{})",
+                        name_len + unit_len + fixed_len,
+                        max_len
+                    );
```
jjbayer: See comment about `Meta`.

jjbayer (changes requested):

relay-general/src/store/normalize.rs
TBS1996: hm yeah i agree, good idea, made it happen now

relay-general/src/store/transactions/processor.rs
TBS1996: ok, I changed it now and tried to make it similar to your example there, is it now what you have in mind?

relay-general/src/store/transactions/processor.rs
```diff
@@ -396,6 +414,53 @@ impl Processor for TransactionsProcessor<'_> {
...
+                    removed_measurements
+                        .insert(key.clone(), Annotated::new(std::mem::take(measurement)));
+                    return false;
+                }
+            }
+            true
+        });
+
+        if !removed_measurements.is_empty() {
+            meta.add_error(Error::invalid("Invalid measurement names"));
```
jjbayer: nit: It would be nice to get a more specific error message when the name is too long, e.g.

```rust
format!("Measurement name too long: name + unit > {}", limit)
```

relay-general/src/store/transactions/processor.rs
jjbayer: Better to move that function to `relay-common` IMO.

relay-general/src/store/transactions/processor.rs
```diff
@@ -396,6 +414,53 @@ impl Processor for TransactionsProcessor<'_> {
...
+
+            let name_len = key.len();
+            let unit_len = measurement.unit.value().map(|val| val.to_string().len());
+
+            if let (Some(fixed_len), Some(max_len), Some(unit_len)) = (
+                self.measurement_config.fixed_length,
+                self.measurement_config.max_len,
+                unit_len,
+            ) {
+                if (name_len + unit_len + fixed_len) > max_len {
```
jjbayer: There's no need to keep adding the `fixed_len` in every evaluation, better to move it to the right side of the equation:

```suggestion
                if (name_len + unit_len) > name_unit_max_len {
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -396,6 +414,53 @@ impl Processor for TransactionsProcessor<'_> {
...
+            };
+
+            if !is_valid_metric_name(key) {
+                removed_measurements
+                    .insert(key.clone(), Annotated::new(std::mem::take(measurement)));
+                return false;
+            }
+
+            let name_len = key.len();
+            let unit_len = measurement.unit.value().map(|val| val.to_string().len());
```
jjbayer: Measurements without unit get the unit `@none` in metrics MRI, so

```suggestion
            let unit_len = measurement.unit.value().unwrap_or(&MetricUnit::None).to_string().len();
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -17,15 +20,30 @@ pub struct TransactionNameConfig<'r> {
     pub rules: &'r [TransactionNameRule],
 }
 
+#[derive(Default)]
+pub struct MeasurementNameConfig {
+    /// How long the name-part of the MRI can be.
+    pub max_len: Option<usize>,
+    /// The length of a measurement MRI without the unit and the name.
+    pub fixed_length: Option<usize>,
```
```suggestion
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -17,15 +20,30 @@ pub struct TransactionNameConfig<'r> {
     pub rules: &'r [TransactionNameRule],
 }
 
+#[derive(Default)]
+pub struct MeasurementNameConfig {
+    /// How long the name-part of the MRI can be.
+    pub max_len: Option<usize>,
```
jjbayer: No need to keep track of two numbers, a single budget for name + unit should be enough:

```suggestion
    /// Combined maximum length of the measurement name + unit.
    pub max_name_unit_len: Option<usize>,
```

relay-server/src/actors/processor.rs
```diff
@@ -2270,6 +2300,8 @@ impl EnvelopeProcessorService {
                 received_at: Some(state.managed_envelope.received_at()),
                 max_secs_in_past: Some(self.config.max_secs_in_past()),
                 max_secs_in_future: Some(self.config.max_secs_in_future()),
+                max_metric_name_length: Some(self.config.aggregator_config().max_name_length),
+                fixed_metric_name_length: Some(*FIXED_MEASUREMENT_LEN),
```
jjbayer: See comments above.
```suggestion
                max_name_and_unit_length: Some(self.config.aggregator_config().max_name_length - *FIXED_MEASUREMENT_LEN),
                fixed_metric_name_length: Some(*FIXED_MEASUREMENT_LEN),
```

relay-system/Cargo.toml
jjbayer: I tried to reproduce this problem locally but was able to run individual tests just fine without the `"full"`, so this must be a local issue. Please remove from diff before merging, as @iker-barriocanal suggested.

jjbayer (changes requested):

relay-general/src/store/transactions/processor.rs
TBS1996: makes sense, I've now added `max_name_and_unit_len` to `lightnormalizationconfig` so it gets calculated a lot earlier.

relay-system/Cargo.toml
TBS1996: resolved it on my end now

relay-common/src/lib.rs
```diff
@@ -23,3 +23,17 @@ pub use crate::glob::*;
...
+/// underscores, slashes and periods.
+pub fn is_valid_metric_name(name: &str) -> bool {
+    let mut iter = name.as_bytes().iter();
+    if let Some(first_byte) = iter.next() {
+        if first_byte.is_ascii_alphabetic() {
+            return iter.all(|b| b.is_ascii_alphanumeric() || matches!(b, b'.' | b'_' | b'/'));
+        }
+    }
+    false
+}
```
TBS1996: simply moved to `relay-common` as suggested, however I don't know where in relay-common to put it. is `lib.rs` fine? should I make a new module?

relay-general/src/store/transactions/processor.rs
TBS1996, line 1: "accidentally changed" it when reverting a change i made, it makes more sense this way, but it does add more diffs

relay-server/src/actors/processor.rs
```diff
@@ -2259,6 +2291,12 @@ impl EnvelopeProcessorService {
     ) -> Result<(), ProcessingError> {
         let request_meta = state.managed_envelope.envelope().meta();
         let client_ipaddr = request_meta.client_addr().map(IpAddr::from);
+        let max_metric_name_and_unit_len = {
+            let max_mri_len = self.config.aggregator_config().max_name_length as isize;
+            // Length of mri without the parts 'name' and 'unit'.
+            let fixed_len = *FIXED_MEASUREMENT_LEN;
+            Some(max_mri_len - fixed_len)
```
TBS1996: alternatively I can have it return a None value if max_mri_len is smaller than fixed_len. At least now i've changed it to an isize so it won't panic

relay-general/src/store/normalize.rs
```diff
@@ -3007,4 +3036,67 @@ mod tests {
...
+
+        remove_invalid_measurements(
+            &mut measurements,
+            &mut meta,
+            &measurements_config,
+            max_name_and_unit_len,
+        );
+
+        // Verifies that this measurement has not been dropped after processing.
+        assert_eq!(measurements.len(), usize::from(should_keep));
```
TBS1996: i thought `if should_keep {1} else {0}` is more clear but clippy got mad

relay-general/src/store/normalize.rs
```diff
@@ -273,9 +274,22 @@ fn remove_invalid_measurements(
...
         let unit = measurement.unit.value().unwrap_or(&MetricUnit::None);
 
+        if let Some(max_name_and_unit_len) = max_name_and_unit_len {
+            let max_name_len = max_name_and_unit_len - unit.to_string().len() as isize;
+
+            if name.len() as isize > max_name_len {
+                return false;
+            }
+        }
+
```
TBS1996: not sure how to handle having multiple possible errors.

`meta.add_error` is a vector, while `meta.set_original_value` is a single field. if i push various errors to meta, how can i associate them with the correct `original_value`?

TBS1996: maybe i could rename `removed_measurements` to `excessive measurements`, and use `original_value` only for the ones removed because they were too many, while for the invalid_names ones i can provide more detailed error messages like `"metric name: '{}' contains invalid characters"` and `"metric name too long. {}/{}"`

relay-general/src/store/transactions/processor.rs
TBS1996: will do, when figured out how to deal with errors (as described down the page)

tests/integration/test_store.py
```diff
@@ -220,7 +220,7 @@ def test_store_buffer_size(mini_sentry, relay):
             relay.send_event(project_id, {"message": "pls ignore"})
         pytest.raises(queue.Empty, lambda: mini_sentry.captured_events.get(timeout=1))
 
-        for (_, error) in mini_sentry.test_failures:
+        for _, error in mini_sentry.test_failures:
```
TBS1996: I didn't want to do this change, but it happens automatically every time i press save

relay-general/src/store/normalize.rs
```diff
@@ -293,15 +321,17 @@ fn remove_invalid_measurements(
         }
 
         // Retain payloads in _meta just for excessive custom measurements.
-        if let Some(measurement) = value.value_mut().take() {
-            removed_measurements.insert(name.clone(), Annotated::new(measurement));
-        }
+        excessive_measurements = true;
```
TBS1996: since we can no longer rely on `removed_metrics` representing only excessive measurements, i made this flag instead. I don't add the error directly as then we won't get multiple of the same type.

relay-general/src/store/normalize.rs
```diff
@@ -293,15 +321,17 @@ fn remove_invalid_measurements(
         }
 
         // Retain payloads in _meta just for excessive custom measurements.
-        if let Some(measurement) = value.value_mut().take() {
```
TBS1996, line 296: I don't understand the code that was there before, we already checked if value.value() returns the Some variant, why do it again to get value_mut when we could have done it in the start of the function? (which i changed it to do now)

relay-common/src/lib.rs
iker-barriocanal: Not strongly opinionated here, but I'd create a `metrics.rs` and put it there.

relay-general/src/store/normalize.rs
```diff
@@ -264,18 +264,45 @@ fn remove_invalid_measurements(
...
 
     measurements.retain(|name, value| {
-        let measurement = match value.value() {
+        let measurement = match value.value_mut() {
             Some(m) => m,
             None => return false,
         };
+
+        if !is_valid_metric_name(name) {
+            meta.add_error(Error::invalid(format!(
```
iker-barriocanal, line 285: Q: why are we using `Error::invalid`? Why not [`ErrorKind::InvalidData`](https://github.com/getsentry/relay/blob/b902222580675e20110b12c935ff6d71b8dbd0ea/relay-general/src/types/meta.rs#L144) instead?

relay-general/src/store/normalize.rs
iker-barriocanal: Note the `take` here -- your code is equivalent, but you take the value in a different way.

relay-general/src/store/normalize.rs
```diff
@@ -293,15 +320,16 @@ fn remove_invalid_measurements(
...
-        if let Some(measurement) = value.value_mut().take() {
-            removed_measurements.insert(name.clone(), Annotated::new(measurement));
-        }
+        removed_measurements.insert(name.clone(), Annotated::new(std::mem::take(measurement)));
 
         false
     });
 
-    if !removed_measurements.is_empty() {
+    if custom_measurements_count >= measurements_config.max_custom_measurements {
```
iker-barriocanal: Being at the max should not be a problem?
```suggestion
    if custom_measurements_count > measurements_config.max_custom_measurements {
```

relay-metrics/src/aggregation.rs
```diff
@@ -2969,7 +2969,8 @@ mod tests {
             metric_name: "c:transactions/long_name_a_very_long_name_its_super_long_really_but_like_super_long_probably_the_longest_name_youve_seen_and_even_the_longest_name_ever_its_extremly_long_i_cant_tell_how_long_it_is_because_i_dont_have_that_many_fingers_thus_i_cant_count_the_many_characters_this_long_name_is".to_owned(),
             tags: BTreeMap::new(),
         };
-        let validation = AggregatorService::validate_bucket_key(long_metric, &aggregator_config);
+        let validation: Result<BucketKey, AggregateMetricsError> =
```
iker-barriocanal: Why this change?

relay-general/src/store/normalize.rs
```diff
@@ -3010,4 +3053,67 @@ mod tests {
...
+        let measurement = Measurement {
+            value: Annotated::new(420.69),
+            unit: Annotated::new(MetricUnit::Duration(DurationUnit::MilliSecond)),
+        };
+        let should_keep = false;
+
+        test_invalid_measurement(name, measurement, should_keep);
+    }
+
+    fn test_invalid_measurement(name: &str, measurement: Measurement, should_keep: bool) {
```
iker-barriocanal: This approach is not bad, but I think it's harder to understand than building a payload and passing it directly, such as [here](https://github.com/getsentry/relay/blob/b597efdbabaf47fae561e0a6b4e5aabe4dade683/relay-general/src/store/normalize.rs#L1952).

If you feel like trying something new and want to play around a bit with a macro, see how we do with transaction names: [this](https://github.com/getsentry/relay/blob/fcbd996ace227a9fe9c69b736db91482e5e178d5/relay-general/src/store/transactions/processor.rs#L1996) macro, and its usage [here](https://github.com/getsentry/relay/blob/fcbd996ace227a9fe9c69b736db91482e5e178d5/relay-general/src/store/transactions/processor.rs#LL2037C28-L2037C28).

relay-general/src/store/normalize.rs
```diff
@@ -3010,4 +3053,67 @@ mod tests {
...
+        let mut meta = Meta::default();
+        let measurements_config = MeasurementsConfig {
+            max_custom_measurements: 1,
+            ..Default::default()
+        };
+
+        // Checks that there is 1 measurement before processing.
+        assert_eq!(measurements.len(), 1);
+
+        remove_invalid_measurements(
```
iker-barriocanal, line 3104: Tests must not use part of the code that you're testing. Hardcoding the output in tests it's easy to understand and more solid than using the code.

relay-server/src/actors/processor.rs
```diff
@@ -2270,6 +2306,7 @@ impl EnvelopeProcessorService {
                 received_at: Some(state.managed_envelope.received_at()),
                 max_secs_in_past: Some(self.config.max_secs_in_past()),
                 max_secs_in_future: Some(self.config.max_secs_in_future()),
+                max_metric_name_and_unit_len,
```
iker-barriocanal: Can we use the same format as the other properties?

relay-server/src/actors/processor.rs
```diff
@@ -59,13 +59,45 @@ use crate::envelope::{AttachmentType, ContentType, Envelope, Item, ItemType};
...
 };
 
+/// The length of a full measurement MRI, minus the name and the unit. This length is the same
+/// for every measurement-mri.
+///
+/// In 'fn process_measurements' we want to check if a measurement name is too long, to do that we
+/// check with a maximum length provided by the config, but that length represents the entire MRI name,
+/// and in process_measurements we only have access to the name part, and the unit part, so here
+/// we calculate the length of the remainder part. which isn't changing so we only need to do it once.
+static FIXED_MEASUREMENT_LEN: Lazy<isize> = Lazy::new(|| {
```
iker-barriocanal: Let's get rid of this. We can go through the path of making it configurable, or just hardcode it. This is harder to understand (let's optimize for ease of reading and understanding), and this logic may introduce bugs.

To configure it, we can do this dynamically from the server and get the information from the project config (like the `project_state.has_feature` examples below), or statically (the envelope processor has a property of relay config). Let's try to take the dynamic approach since it's easier to configure and update without new deployments.

iker-barriocanal (changes requested):

iker-barriocanal (dismissed):
Blocking for the configuration of the maximum length, and the improvement on tests.

relay-general/src/store/normalize.rs
TBS1996: that sounds more right yeah, i'll change it. I used iinvalid because that was used in the already present error message but i see that it might not fit here as much

TBS1996: hmm although it seems you lose some information then, like you can put InvalidData and write the original value, but you don't get the exact reason why it's invalid?

relay-general/src/store/normalize.rs
TBS1996: if custom_measurements_count == max_custom_measurements, it will not pass the `if-block` at line 317 so it will be removed, thats why I made it like this, however I realize now, if it reaches the same num as max_cusotm_measurements on the final iteration, it will think it removed a measurement even when it didn't.

so I'll change this to rather be a flag

relay-server/src/actors/processor.rs
TBS1996: If we wanna hardcode it then we need to make sure the fixed part of measurement will not change in the future, since that would introduce bugs, about configuring it, we already configure how long an MRI but we still need to subtract the fixed part to calculate the name. 

we could do as jorris suggested, hardcore it in production code, but in tests ensure the FIXED_MEASUREMENT_LEN matches with the hardcoded value, I'm not sure if that would be simpler or not

relay-server/src/actors/processor.rs
```diff
@@ -59,13 +59,45 @@ use crate::envelope::{AttachmentType, ContentType, Envelope, Item, ItemType};
...
     ItemAction, ManagedEnvelope, SamplingResult,
 };
 
+/// The length of a full measurement MRI, minus the name and the unit. This length is the same
+/// for every measurement-mri.
+///
+/// In 'fn process_measurements' we want to check if a measurement name is too long, to do that we
+/// check with a maximum length provided by the config, but that length represents the entire MRI name,
+/// and in process_measurements we only have access to the name part, and the unit part, so here
+/// we calculate the length of the remainder part. which isn't changing so we only need to do it once.
```
jjbayer: This might help to understand what's going on:

```suggestion
/// we calculate the length of the remainder part. which isn't changing so we only need to do it once:
///
///     len("d:transactions/measurements.foobar@none") - len("foobar") - len("none") == len("d:transactions/measurements.@")
```

relay-server/src/actors/processor.rs
jjbayer: This might be a slightly better name, but not sure.

```suggestion
static MEASURENT_MRI_OVERHEAD: Lazy<isize> = Lazy::new(|| {
```

jjbayer: @iker-barriocanal we considered several options here but this one seems the most robust.

1. If you hard-code this value, you need a unit test to detect a change in the naming schema.
2. If you make this value configurable, you allow the possibility of accepting measurements that do not fit into an MRI. IMO this value is a derived value and should not be configurable.

tests/integration/test_store.py
jjbayer: `git checkout master tests/integration/test_store.py`

relay-general/src/store/normalize.rs
```diff
@@ -3010,4 +3053,68 @@ mod tests {
...
+        let measurement = Measurement {
+            value: Annotated::new(420.69),
+            unit: Annotated::new(MetricUnit::Duration(DurationUnit::MilliSecond)),
+        };
+
+        let is_dropped = test_is_measurement_dropped(name, measurement);
+        assert_eq!(is_dropped, true);
+    }
+
+    fn test_is_measurement_dropped(name: &str, measurement: Measurement) -> bool {
```
jjbayer: Our actual test cases have the `test_` prefix, so I would omit it for the helper function.

```suggestion
    fn measurement_is_dropped(name: &str, measurement: Measurement) -> bool {
```

relay-general/src/store/normalize.rs
```diff
@@ -3010,4 +3053,68 @@ mod tests {
...
+    #[test]
+    fn test_drops_measurements_with_invalid_characters() {
+        let name = "i æm frøm nørwåy";
+        let measurement = Measurement {
+            value: Annotated::new(420.69),
+            unit: Annotated::new(MetricUnit::Duration(DurationUnit::MilliSecond)),
+        };
+
+        let is_dropped = test_is_measurement_dropped(name, measurement);
+        assert_eq!(is_dropped, true);
```
```suggestion
        assert!(measurement_is_dropped(name, measurement));
```
(same for other test cases)

relay-general/src/store/normalize.rs
```diff
@@ -264,18 +264,45 @@ fn remove_invalid_measurements(
     measurements: &mut Measurements,
     meta: &mut Meta,
     measurements_config: &MeasurementsConfig,
+    max_name_and_unit_len: Option<isize>,
```
jjbayer: Let's move this into `MeasurementsConfig`.

jjbayer (changes requested):
Looks good in general, just a few remarks.

relay-server/src/actors/processor.rs
iker-barriocanal: nit: `usize` instead?

iker-barriocanal: >1. If you hard-code this value, you need a unit test to detect a change in the naming schema.

Why? We could just hardcode a test that fails if the naming schema changes.

>2. If you make this value configurable, you allow the possibility of accepting measurements that do not fit into an MRI.

IIUC, that assumes the config value is incorrect.

>IMO this value is a derived value and should not be configurable.

I think I agree with that. I still don't understand the reason to not hardcode it though.

jjbayer: > We could just hardcode a test that fails if the naming schema changes.

Yeah, that's what I mean. Something like `assert_eq!(FIXED_MEASUREMENT_LEN, len(mri) - len(name) - len(unit))`. This would also be fine with me. The one thing we should prevent is hardcoding this value without any checks if it corresponds to reality.

relay-general/src/store/normalize.rs
TBS1996: I tried that but it's behind an Arc in `ProjectState` so I can't mutate it, is there a way to go around that?

relay-server/src/actors/processor.rs
TBS1996: that would cause a panic if the user configures the max mri length to be too small when subtracting

TBS1996: @iker-barriocanal  the config value is the entire MRI, which is already configured, but we try to derive the name and unit max len from the max mri len. You can't have both be configurable because one is a subset of the other

iker-barriocanal: > You can't have both be configurable because one is a subset of the other

Taking that as a requirement, I think you can make it configurable?

relay-common/src/lib.rs
TBS1996: ok changed to your suggesiton

relay-server/src/actors/processor.rs
TBS1996: you can make the max full mri length configurable, or the name of mri configurable, but not both. Currently full mri length is already configurable so it makes sense to just keep it like that currently. 

we cannot make both configurable because it's like an equation. 

`mri_len == name_len + unit_len + fixed_len`

if we make both mri_len and name_len configurable then this equation cannot be guaranteed to be valid

relay-general/src/store/normalize.rs
```diff
@@ -292,16 +328,19 @@ fn remove_invalid_measurements(
             return true;
         }
 
+        too_many_measurements = true;
```
jjbayer: Instead of toggling a flag here, let's add an error `Too many measurements: {name}`, such that list of errors matches up exactly with the list of removed measurements.

relay-general/src/store/normalize.rs
```diff
@@ -264,18 +270,48 @@ fn remove_invalid_measurements(
...
+            removed_measurements.insert(name.clone(), Annotated::new(std::mem::take(measurement)));
+            return false;
+        }
+
         // TODO(jjbayer): Should we actually normalize the unit into the event?
         let unit = measurement.unit.value().unwrap_or(&MetricUnit::None);
 
+        if let Some(max_mri_len) = max_mri_len {
+            let max_name_and_unit_len =
+                max_mri_len as isize - MeasurementsConfig::MEASUREMENT_MRI_OVERHEAD;
```
jjbayer: I think I already wrote this in an earlier PR review: there's no need to do this subtraction for every measurement. Just compute `max_name_and_unit_len` once and put it in the `LightNormalizationConfig`.

relay-general/src/store/normalize.rs
```diff
@@ -264,18 +270,48 @@ fn remove_invalid_measurements(
...
+            let max_name_len = max_name_and_unit_len - unit.to_string().len() as isize;
+
+            if name.len() as isize > max_name_len {
```
jjbayer: All these casts make the code less readable. I remember talking about this, but why do we need `isize` again? I think we should make all numbers `usize` and use `saturating_sub` instead.

tests/integration/test_store.py
jjbayer: bump

relay-general/src/store/normalize.rs
jjbayer: You're right, `MeasurementsConfig` is part of the project config protocol, so let's not change it.

jjbayer (changes requested):

relay-general/src/store/normalize.rs
TBS1996: ok, do we then want to also set original value,  or is the name in the error message enough?

relay-general/src/store/normalize.rs
TBS1996: true, was an oversight in refactoring when I thought I wouldn't have to pass in an additional argument, fixed it now

relay-general/src/store/normalize.rs
jjbayer: Would also set the original value

relay-general/src/store/normalize.rs
```diff
@@ -292,16 +325,13 @@ fn remove_invalid_measurements(
             return true;
         }
 
-        // Retain payloads in _meta just for excessive custom measurements.
-        if let Some(measurement) = value.value_mut().take() {
-            removed_measurements.insert(name.clone(), Annotated::new(measurement));
-        }
+        meta.add_error(Error::invalid(format!("too many measurements: {}", name)));
```
```suggestion
        meta.add_error(Error::invalid(format!("Too many measurements: {}", name)));
```

relay-server/src/actors/processor.rs
```diff
@@ -3544,4 +3559,35 @@ mod tests {
...
+                unit,
+                tags,
+            };
+
+            let metric: Metric = measurement.into_metric(UnixTimestamp::now());
+
+            metric.name.len() - unit.to_string().len() - name.len()
+        };
+
+        assert_eq!(hardcoded_value, derived_value);
```
```suggestion
        assert_eq!(hardcoded_value, derived_value, "Update `MEASUREMENT_MRI_OVERHEAD` if the naming scheme changed.");
```

jjbayer (approved):

PR #2084, opened 2023-05-02
build: Use jemalloc on Linux

jan-auer:
Enables `jemalloc` as global allocator for Linux builds. This should reduce memory fragmentation and reclaim unused memory quicker, which is expected to lower Relay's overall memory footprint. 

Relay's processing profile frequently moves memory between threads and also spawns threads on the fly. To allow the allocator to reclaim this memory efficiently, we also enable the `background_threads` option. See https://github.com/jemalloc/jemalloc/issues/1128#issuecomment-378438622 for a mention of this.

See https://github.com/getsentry/relay/pull/344
See https://github.com/getsentry/relay/pull/1817


jjbayer (approved):
We should probably start off the deploy with a Canary deploy.

PR #2087, opened 2023-05-02
ref(spans): Use transactions namespace with `span` prefix

iker-barriocanal:
The storage side is not ready yet to handle a new namespace, so we will be using the transactions namespace for this first iteration. To differentiate metrics and tags from both namespaces, the `spans.` prefix is added to metric names and tag keys from spans.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -435,7 +435,7 @@ fn extract_span_metrics(
             let mut span_tags = shared_tags.clone();
 
             if let Some(span_op) = span.op.value() {
-                span_tags.insert("op".to_owned(), span_op.to_owned());
+                span_tags.insert("spans.op".to_owned(), span_op.to_owned());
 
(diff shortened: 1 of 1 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -805,27 +805,27 @@ mod tests {
...
-                name: "d:spans/duration@millisecond",
+                name: "d:transactions/spans.duration@millisecond",
                 value: Distribution(
                     59000.0,
                 ),
                 timestamp: UnixTimestamp(1619420400),
                 tags: {
                     "environment": "fake_environment",
-                    "op": "react.mount",
+                    "spans.op": "react.mount",
```
jjbayer: nit: I would use `span.` as prefix instead of `spans.`, because the information extracted comes from a single span. E.g. we have a tag called `transaction.op` two lines below, so it makes sense to use `span.op` here.

Feel free to ignore if this naming scheme was agreed upon with other teams.

jjbayer (approved):

PR #2088, opened 2023-05-02
feat(filter): Filter webkit error

TBS1996:
https://github.com/getsentry/sentry-javascript/issues/7993

The added regex pattern should filter out the webkit issues linked from here: https://github.com/getsentry/sentry-javascript/issues/7993#issuecomment-1569792418

The source is likely from safari extensions.


TBS1996:
~blocked on: https://github.com/getsentry/sentry-javascript/discussions/5875#discussioncomment-5781699~

no longer blocked: https://github.com/getsentry/sentry-javascript/issues/7993#issuecomment-1548640463

relay-filter/src/browser_extensions.rs
```diff
@@ -39,7 +39,8 @@ static EXTENSION_EXC_VALUES: Lazy<Regex> = Lazy::new(|| {
         # See: https://forum.sentry.io/t/error-in-raven-js-plugin-setsuspendstate/481/
         plugin\.setSuspendState\sis\snot\sa\sfunction|
         # Chrome extension message passing failure
-        Extension\scontext\sinvalidated
+        Extension\scontext\sinvalidated|
+        webkit-masked-url:
```
olksdr, line 43: Do we want to have also `hidden` in there as requested in the linked issue? Or having only `webkit-masked-url` should be fine?

olksdr (approved):

relay-filter/src/browser_extensions.rs
TBS1996: the way `hidden` is included seems a bit too irregular on the events here: https://github.com/getsentry/sentry-javascript/issues/7993#issuecomment-1569792418

whereas almost all have in common `webkit-masked-url:` 

if i want to include hidden then i think the pattern would be something like `webkit-masked-url:.*hidden`, but I'm not sure if theres a point to this because I don't think we will have false positives with the way this PR does it

PR #2090, opened 2023-05-03
ref(spans): Add more tags to span metrics

iker-barriocanal:
Add`span.domain` and `span.status_code` tags to metrics.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -479,8 +479,8 @@ fn extract_span_metrics(
                         Some((_method, url)) => {
                             let url = SchemeDomainPort::from(url);
                             match (url.domain, url.port) {
-                                (Some(dom), Some(port)) => Some(format!("{}:{}", dom, port)),
-                                (Some(dom), None) => Some(dom),
+                                (Some(domain), Some(port)) => Some(format!("{}:{}", domain, port)),
+                                (Some(domain), None) => Some(domain),
                                 _ => None,
                             }
                         }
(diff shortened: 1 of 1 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -2,6 +2,7 @@ use std::collections::BTreeMap;
 
 use relay_common::{DurationUnit, EventType, MetricUnit, SpanStatus, UnixTimestamp};
 use relay_dynamic_config::{AcceptTransactionNames, TaggingRule, TransactionMetricsConfig};
+use relay_filter::csp::SchemeDomainPort;
```
jjbayer, line 5: You could probably also use `url::Url` to reduce cross-dependencies.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -472,6 +473,23 @@ fn extract_span_metrics(
...
                     span_tags.insert("span.action".to_owned(), act.to_owned());
                 }
+
+                if span_op == "http.client" {
+                    let domain = match span.description.value().and_then(|v| v.split_once(' ')) {
+                        Some((_method, url)) => {
+                            let url = SchemeDomainPort::from(url);
+                            match (url.domain, url.port) {
+                                (Some(dom), Some(port)) => Some(format!("{}:{}", dom, port)),
+                                (Some(dom), None) => Some(dom),
```
jjbayer: nit: 
```suggestion
                                (Some(domain), Some(port)) => Some(format!("{}:{}", domain, port)),
                                (Some(domain), None) => Some(domain),
```

jjbayer (approved):
Would be great to have some unit tests for these.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: TIL, thanks! I'll do that in a follow-up once we've rolled this out.

PR #2091, opened 2023-05-03
feat(dynamic-sampling): Add possibility to run dynamic sampling from `sentry-relay`

iambriccardo:
This PR adds a new function to `sentry-relay` called `run_dynamic_sampling` which will run dynamic sampling on a two `SamplingConfig`s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@
 - Scrub sensitive keys (`passwd`, `token`, ...) in Replay recording data. ([#2034](https://github.com/getsentry/relay/pull/2034))
 - Add document_uri to csp filter. ([#2059](https://github.com/getsentry/relay/pull/2059))
 - Store `geo.subdivision` of the end user location. ([#2058](https://github.com/getsentry/relay/pull/2058))
-- Add new FFI function for running dynamic sampling. ([#2091](https://github.com/getsentry/relay/pull/2091))
 
 **Internal**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Scrub sensitive keys (`passwd`, `token`, ...) in Replay recording data. ([#2034](https://github.com/getsentry/relay/pull/2034))
 - Add document_uri to csp filter. ([#2059](https://github.com/getsentry/relay/pull/2059))
 - Store `geo.subdivision` of the end user location. ([#2058](https://github.com/getsentry/relay/pull/2058))
+- Add new FFI function for running dynamic sampling. ([#2091](https://github.com/getsentry/relay/pull/2091))
 
 **Internal**:
 
--- a/py/sentry_relay/processing.py
+++ b/py/sentry_relay/processing.py
@@ -274,14 +274,12 @@ def run_dynamic_sampling(sampling_config, root_sampling_config, dsc, event):
     assert isinstance(dsc, string_types)
     assert isinstance(event, string_types)
 
-    return json.loads(
-        decode_str(
-            rustcall(
-                lib.run_dynamic_sampling,
-                encode_str(sampling_config),
-                encode_str(root_sampling_config),
-                encode_str(dsc),
-                encode_str(event),
-            )
-        )
+    result_json = rustcall(
+        lib.run_dynamic_sampling,
+        encode_str(sampling_config),
+        encode_str(root_sampling_config),
+        encode_str(dsc),
+        encode_str(event),
     )
+
+    return json.loads(decode_str(result_json, free=True))
--- a/py/sentry_relay/processing.py
+++ b/py/sentry_relay/processing.py
@@ -28,6 +28,7 @@ __all__ = [
     "validate_sampling_condition",
     "validate_sampling_configuration",
     "validate_project_config",
+    "run_dynamic_sampling",
 ]
 
 
@@ -262,3 +263,25 @@ def validate_project_config(config, strict: bool):
     error = decode_str(raw_error, free=True)
     if error:
         raise ValueError(error)
+
+
+def run_dynamic_sampling(sampling_config, root_sampling_config, dsc, event):
+    """
+    Runs dynamic sampling on an event and returns the merged rules together with the sample rate.
+    """
+    assert isinstance(sampling_config, string_types)
+    assert isinstance(root_sampling_config, string_types)
+    assert isinstance(dsc, string_types)
+    assert isinstance(event, string_types)
+
+    return json.loads(
+        decode_str(
+            rustcall(
+                lib.run_dynamic_sampling,
+                encode_str(sampling_config),
+                encode_str(root_sampling_config),
+                encode_str(dsc),
+                encode_str(event),
+            )
+        )
+    )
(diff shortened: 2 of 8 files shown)


relay-cabi/src/processing.rs
```diff
@@ -321,3 +325,72 @@ pub unsafe extern "C" fn relay_validate_project_config(
...
+#[relay_ffi::catch_unwind]
+pub unsafe extern "C" fn run_dynamic_sampling(
+    sampling_config: *const RelayStr,
+    root_sampling_config: *const RelayStr,
+    dsc: *const RelayStr,
+    event: *const RelayStr,
+) -> RelayStr {
+    let sampling_config = serde_json::from_str::<SamplingConfig>((*sampling_config).as_str())?;
+    let root_sampling_config =
+        serde_json::from_str::<SamplingConfig>((*root_sampling_config).as_str())?;
```
iambriccardo: I have tried to use the `?` operator for curiosity and I don't understand why the compiler allows me to return a `Result` when the type is `RelayStr`.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Scrub sensitive keys (`passwd`, `token`, ...) in Replay recording data. ([#2034](https://github.com/getsentry/relay/pull/2034))
 - Add document_uri to csp filter. ([#2059](https://github.com/getsentry/relay/pull/2059))
 - Store `geo.subdivision` of the end user location. ([#2058](https://github.com/getsentry/relay/pull/2058))
+- Add new FFI function for running dynamic sampling. ([#2091](https://github.com/getsentry/relay/pull/2091))
```
jan-auer: Please move this to `py/CHANGELOG.md`

py/sentry_relay/processing.py
```diff
@@ -262,3 +263,25 @@ def validate_project_config(config, strict: bool):
     error = decode_str(raw_error, free=True)
     if error:
         raise ValueError(error)
+
+
+def run_dynamic_sampling(sampling_config, root_sampling_config, dsc, event):
```
jan-auer, line 268: Should we take an entire envelope as interface, instead? In Relay's pipeline, DS also reads information from other envelope headers.

py/sentry_relay/processing.py
```diff
@@ -262,3 +263,25 @@ def validate_project_config(config, strict: bool):
...
+        decode_str(
+            rustcall(
+                lib.run_dynamic_sampling,
+                encode_str(sampling_config),
+                encode_str(root_sampling_config),
+                encode_str(dsc),
+                encode_str(event),
+            )
+        )
+    )
```
jan-auer: Please add `free=True` to `decode_str`, since the owned string returned by FFI needs to be released. You can make this more readable like so:

```suggestion
    result_json = rustcall(
        lib.run_dynamic_sampling,
        encode_str(sampling_config),
        encode_str(root_sampling_config),
        encode_str(dsc),
        encode_str(event),
    )

    return json.loads(decode_str(result_json, free=True))
```

py/tests/test_processing.py
```diff
@@ -283,3 +283,268 @@ def test_validate_project_config():
     with pytest.raises(ValueError) as e:
         sentry_relay.validate_project_config(json.dumps(config), strict=True)
     assert str(e.value) == 'json atom at path ".foobar" is missing from rhs'
+
+
+def test_run_dynamic_sampling_with_valid_params_and_match():
```
jan-auer, line 288: These tests are quite elaborate and will have to change every time we change something about sampling internals, particularly because they check the contents of the returned rules. Should we instead make more general checks, such as `assert "merged_sampling_configs" in result` and leave the rest to the DS unit tests in Rust?

relay-cabi/src/processing.rs
```diff
@@ -3,13 +3,15 @@
 #![deny(unused_must_use)]
 #![allow(clippy::derive_partial_eq_without_eq)]
 
+use chrono::Utc;
 use std::cmp::Ordering;
 use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::slice;
 
+use crate::core::{RelayBuf, RelayStr};
```
jan-auer: Please order imports according to the [style guide](https://develop.sentry.dev/rust/#import-order). This concerns the `chrono` and `crate` imports.

relay-cabi/src/processing.rs
```diff
@@ -321,3 +325,72 @@ pub unsafe extern "C" fn relay_validate_project_config(
...
+    root_sampling_config: *const RelayStr,
+    dsc: *const RelayStr,
+    event: *const RelayStr,
+) -> RelayStr {
+    let sampling_config = serde_json::from_str::<SamplingConfig>((*sampling_config).as_str())?;
+    let root_sampling_config =
+        serde_json::from_str::<SamplingConfig>((*root_sampling_config).as_str())?;
+    // We can optionally accept a dsc and event.
+    let dsc = serde_json::from_str::<DynamicSamplingContext>((*dsc).as_str());
+    let event = serde_json::from_str::<EphemeralEvent>((*event).as_str());
```
jan-auer: Instead of parsing some synthetic `EphemeralEvent`, you should just load an actual `Annotated<Event>` directly:

```suggestion
    let event = Annotated::<Event>::from_json((*event).as_str())?;
    let event = event.value().ok_or(<some error>)?;
```

relay-cabi/src/processing.rs
```diff
@@ -321,3 +325,72 @@ pub unsafe extern "C" fn relay_validate_project_config(
...
+/// Runs dynamic sampling given the sampling config, root sampling config, dsc and event.
+///
+/// Returns the sampling decision containing the sample_rate and the list of matched rule ids.
+#[no_mangle]
+#[relay_ffi::catch_unwind]
+pub unsafe extern "C" fn run_dynamic_sampling(
+    sampling_config: *const RelayStr,
+    root_sampling_config: *const RelayStr,
+    dsc: *const RelayStr,
+    event: *const RelayStr,
```
jan-auer: You can pass all of these as references and then don't have to worry about dereferencing the pointer below. This is something we've not done consistently in the codebase, but we should move towards:

```suggestion
    sampling_config: &RelayStr,
    root_sampling_config: &RelayStr,
    dsc: &RelayStr,
    event: &RelayStr,
```

jan-auer (changes requested):

py/sentry_relay/processing.py
iambriccardo: I wanted to keep it stupid simple for now, in case we want, we can extend it to have an envelope as well.

py/tests/test_processing.py
iambriccardo: That's a very good idea yes!

relay-cabi/src/processing.rs
```diff
@@ -3,6 +3,8 @@
 #![deny(unused_must_use)]
 #![allow(clippy::derive_partial_eq_without_eq)]
 
+use anyhow::Context;
+use chrono::Utc;
```
jan-auer: Please move these two imports to the next import group (above `once_cell`).

jan-auer (approved):

PR #2092, opened 2023-05-03
ref(dynamic-config): Remove transaction metrics allowlist from Sentry

### LINKED ISSUE #1632: Remove transaction metrics allowlist from Sentry
Once Relay does not use the old configuration format anymore, remove it from project configs.

This means we can also completely remove the metrics allow list set up there.

**NOTE:** Outdated external Relays will not recognize the new config format, so when Sentry stops sending the metrics allowlist, they will stop sending metrics altogether, while still setting `metricsExtracted=1`. The way around this is bumping the `VERSION` of the metrics extraction config.

TBS1996:
#1632 

With this change, old external relays will have to be upgraded in order to extract metrics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 **Internal**:
 
+- Remove transaction metrics allowlist. ([#2092](https://github.com/getsentry/relay/pull/2092))
 - Include unknown feature flags in project config when serializing it. ([#2040](https://github.com/getsentry/relay/pull/2040))
 - Copy transaction tags to the profile. ([#1982](https://github.com/getsentry/relay/pull/1982))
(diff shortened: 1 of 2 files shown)


jjbayer (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 **Internal**:
 
+- Remove transaction metrics allowlist. ([#2092](https://github.com/getsentry/relay/pull/2092))
```
iker-barriocanal, line 20: Should we add this to a `breaking changes` section in the changelog?

iker-barriocanal (approved):

CHANGELOG.md
TBS1996: I asked Jan and got a no

PR #2093, opened 2023-05-03
instr(buffer): Count number of envelopes

jjbayer:
Bring back envelope counts as statsd metrics for the spooling envelope buffer:

1. `envelopes_mem_count`: Number of envelopes currently in memory.
2. `envelopes_written`, `envelopes_read`: Number of envelopes flushed to and from disk in any given time frame.
3. `envelopes_disk_count`: Number of envelopes on disk: Because running a `SELECT count(*)` query on a large table is expensive, we _only_ track this number if there is nothing in the database on startup. This should be the default use case, so in general, this metric will be available. If there is data in the db on startup, only `envelopes_written` / `envelopes_read` will be available.

**Discarded options for `envelopes_disk_count`**

* We considered running the `SELECT count(*)` query async on startup, but that might still lock the database table.
* We considered running the counting query in a `tokio::timeout`, but it is unclear if this would actually unlock the database, or only cancel the future while the query keeps running in the background.
* We could limit the max. number of rows in the counting query, such that the metric still works for small leftovers in the db, but not for large buffers encountered on start up. This should be easy to implement in a follow-up if the need for it occurs.
* We could get track this number in the db itself (in a separate table), and update it every time we spool or unspool, from within a database transaction. We decided this is too much engineering effort for just a metric, but might reconsider if it turns out we really need this metric in the edge case that there's something in the db on startup.

Fixes https://github.com/getsentry/team-ingest/issues/115

#skip-changelog

### DIFF
--- a/relay-server/src/actors/spooler/mod.rs
+++ b/relay-server/src/actors/spooler/mod.rs
@@ -264,11 +264,11 @@ struct OnDisk {
     db: Pool<Sqlite>,
     buffer_guard: Arc<BufferGuard>,
     max_disk_size: usize,
-    /// The number of items spooled to disk.
+    /// The number of items currently on disk.
     ///
-    /// This number can be negative when relay picks up an existing db and immediately starts
-    /// unspooling.
-    spool_count: isize,
+    /// We do not track the count when we encounter envelopes in the database on startup,
+    /// because counting those envelopes would risk locking the db for multiple seconds.
+    count: Option<u64>,
 }
 
 impl OnDisk {
@@ -301,11 +301,7 @@ impl OnDisk {
             .await
             .map_err(BufferError::InsertFailed)?;
 
-        self.spool_count += inserted as isize;
-
-        relay_statsd::metric!(
-            gauge(RelayGauges::BufferEnvelopesDiskCount) = self.spool_count as f64
-        );
+        self.track_count(inserted as i64);
 
         Ok(())
     }
@@ -323,10 +319,7 @@ impl OnDisk {
             count += result.rows_affected();
         }
 
-        self.spool_count -= count as isize;
-        relay_statsd::metric!(
-            gauge(RelayGauges::BufferEnvelopesDiskCount) = self.spool_count as f64
-        );
+        self.track_count(-(count as i64));
 
         Ok(count as usize)
     }
@@ -386,7 +379,7 @@ impl OnDisk {
                 return Ok(());
             }
 
-            let mut count = 0;
+            let mut count: i64 = 0;
             while let Some(envelope) = envelopes.next().await {
                 count += 1;
                 let envelope = match envelope {
@@ -398,10 +391,7 @@ impl OnDisk {
                             "failed to read the buffer stream from the disk: {}",
                             LogError(&err)
                         );
-                        self.spool_count -= count;
-                        relay_statsd::metric!(
-                            gauge(RelayGauges::BufferEnvelopesDiskCount) = self.spool_count as f64
-                        );
+                        self.track_count(-count);
                         return Err(key);
                     }
                 };
@@ -417,10 +407,7 @@ impl OnDisk {
                 }
             }
 
-            self.spool_count -= count;
-            relay_statsd::metric!(
-                gauge(RelayGauges::BufferEnvelopesDiskCount) = self.spool_count as f64
-            );
+            self.track_count(-count);
         }
     }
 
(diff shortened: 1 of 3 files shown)


relay-server/src/actors/spooler/mod.rs
```diff
@@ -209,8 +211,12 @@ impl InMemory {
             count += current_count;
             self.used_memory -= current_size;
         }
+        self.envelope_count -= count as isize;
         relay_statsd::metric!(
-            histogram(RelayHistograms::BufferEnvelopesMemory) = self.used_memory as f64
+            histogram(RelayHistograms::BufferEnvelopesMemoryBytes) = self.used_memory as f64
+        );
+        relay_statsd::metric!(
+            histogram(RelayHistograms::BufferEnvelopesMemoryCount) = self.envelope_count as u64
```
iker-barriocanal: Why not make it a counter instead? Is it because of potential negative counts?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -250,17 +264,22 @@ struct OnDisk {
     db: Pool<Sqlite>,
     buffer_guard: Arc<BufferGuard>,
     max_disk_size: usize,
+    /// The number of items spooled to disk.
+    ///
+    /// This number can be negative when relay picks up an existing db and immediately starts
+    /// unspooling.
```
iker-barriocanal: Even if `select count(*)` is expensive, I think we could consider it for startup cases, since it's only one query during startup so the overhead shouldn't be very big. We'd avoid reporting negative values in that case. What do you think?

iker-barriocanal (changes requested):
The PR looks good with the current approach, but I'd like to have another thought about the negative metric values.

relay-server/src/actors/spooler/mod.rs
jan-auer: This logs the number of envelopes in the buffer at a certain time. A counter is strictly incremental, so depending on how often you log the counter, you get different values. It's usually used for counting rather than for tracking the number of things.

@jjbayer I think you actually want to log a gauge here.

relay-server/src/actors/spooler/mod.rs
jan-auer: We should also test how fast it actually is with a completely filled database. There are conflicting statements on the internet about what the fastest way to count items is, but it seems that additional indexes or running `COUNT(_rowid_)` could be faster.

jjbayer: @olksdr and I did some research & experiment on this and ruled out counting on startup, see this discussion on a previous PR: https://github.com/getsentry/relay/pull/2011#discussion_r1158649001.

I did not try `count(_rowid_)`, but since it's a linear scan either way I cannot imagine it helps us much (see https://stackoverflow.com/a/18810481). If you think it's worth it, I can investigate some more time.

relay-server/src/actors/spooler/mod.rs
jjbayer: > I think you actually want to log a gauge here.

You're right. Shouldn't [`EnvelopeQueueSize`](https://github.com/getsentry/relay/blob/0575289402d1cac2c29485a7dee8fc151ce39014/relay-server/src/statsd.rs#L30-L44) also be a gauge then?

olksdr (approved):
It will take quite some time to run `count` on big files , since it's will do the sequential scan of the table in any case.

Otherwise lgtm!

jjbayer:
@iker-barriocanal I will add some more details to the PR description later, especially options considered.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -209,8 +211,12 @@ impl InMemory {
             count += current_count;
             self.used_memory -= current_size;
         }
+        self.envelope_count -= count as isize;
         relay_statsd::metric!(
-            histogram(RelayHistograms::BufferEnvelopesMemory) = self.used_memory as f64
+            histogram(RelayHistograms::BufferEnvelopesMemoryBytes) = self.used_memory as f64
+        );
+        relay_statsd::metric!(
+            gauge(RelayGauges::BufferEnvelopesMemoryCount) = self.envelope_count as u64
```
iker-barriocanal, line 219: We're casting an `isize` as `u64`. I think this panics on negative numbers. Should we use something like `u64::try_from`, and log an error if we couldn't parse it?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -457,9 +488,26 @@ impl OnDisk {
...
+    fn track_count(&mut self, increment: i64) {
+        // Track the number of envelopes read/written:
+        let metric = if increment < 0 {
+            RelayCounters::BufferEnvelopesRead
+        } else {
+            RelayCounters::BufferEnvelopesWritten
+        };
+        relay_statsd::metric!(counter(metric) += increment);
+
+        // If used, also track the total envelope count:
```
iker-barriocanal: What does "if used" mean?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -209,8 +211,12 @@ impl InMemory {
             count += current_count;
             self.used_memory -= current_size;
         }
+        self.envelope_count -= count as isize;
```
iker-barriocanal: What's the type of `count`? If it's `u64`, it may not fit in `isize`. Should we try the casting instead (comment below)?

iker-barriocanal (changes requested):
Blocking for questions on potential panics on type casting.

The approach makes sense, although I think the effort of having a table to track the envelopes in the database is worth the effort. Not strongly opinionated here though.

relay-server/src/actors/spooler/mod.rs
jjbayer: " as " wraps around (does not panic), but I changed the type to `usize` (see other comment).

relay-server/src/actors/spooler/mod.rs
jjbayer:
```suggestion
```

jjbayer: Will remove the comment, it does not add any information.

relay-server/src/actors/spooler/mod.rs
jjbayer: It has to be _very_ large to not fit into `isize`, but changed to `usize` and `saturating_sub` regardless.

iker-barriocanal (approved):

PR #2095, opened 2023-05-04
feat(normalization): Scrub span description URLs

iker-barriocanal:
Scrub identifiers from URLs in span description. The scrubbed description is available in the span as `data['description.scrubbed']`, and in case something was scrubbed, a `span.description` tag is added to span metrics. This only happens to these projects with the feature flag enabled.

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::collections::BTreeMap;
 
 use relay_common::SpanStatus;
 
@@ -468,16 +469,13 @@ fn scrub_span_description(span: &mut Span) -> Result<(), ProcessingAction> {
             return Ok(());
         };
         span.data
-            .value_mut()
-            .as_mut()
+            .get_or_insert_with(BTreeMap::new)
             // We don't care what the cause of scrubbing was, since we assume
             // that after scrubbing the value is sanitized.
-            .and_then(|data| {
-                data.insert(
-                    "description.scrubbed".to_owned(),
-                    Annotated::new(Value::String(new_desc.to_owned())),
-                )
-            });
+            .insert(
+                "description.scrubbed".to_owned(),
+                Annotated::new(Value::String(new_desc.to_owned())),
+            );
     }
 
     Ok(())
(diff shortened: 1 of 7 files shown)


relay-general/src/store/transactions/processor.rs
```diff
@@ -438,12 +444,45 @@ impl Processor for TransactionsProcessor<'_> {
...
+        };
+        span.data
+            .value_mut()
+            .as_mut()
+            // We don't care what the cause of scrubbing was, since we assume
+            // that after scrubbing the value is sanitized.
+            .and_then(|data| {
+                data.insert(
+                    "description.scrubbed".to_owned(),
+                    Annotated::new(Value::String(new_desc.to_owned())),
```
iker-barriocanal, line 476: I was having some issues to add the scrubbed description as the value here, and the dance around scrubbed values above is a result of this.

relay-general/src/store/transactions/processor.rs
```diff
@@ -2184,11 +2236,104 @@ mod tests {
...
+
+                process_value(
+                    &mut span,
+                    &mut TransactionsProcessor::new(TransactionNameConfig::default(), true),
+                    ProcessingState::root(),
+                )
+                .unwrap();
+
+                assert_eq!($input, span.value().unwrap().description.value().unwrap());
+                if $output == "" {
```
iker-barriocanal, line 2275: Ideally, this is `Option::None`, I'll address this in a future PR.

relay-general/src/store/transactions/processor.rs
```diff
@@ -438,12 +445,42 @@ impl Processor for TransactionsProcessor<'_> {
...
         Ok(())
     }
 }
 
+fn scrub_span_description(span: &mut Span) -> Result<(), ProcessingAction> {
+    if span.description.value().is_none() {
+        return Ok(());
+    }
+
+    // let mut scrubbed: Annotated<Value::String> = span.description.clone();
```
```suggestion
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -438,12 +445,42 @@ impl Processor for TransactionsProcessor<'_> {
...
+    let mut scrubbed = span.description.clone();
+
+    // Scrub URLs
+    if scrub_identifiers(&mut scrubbed)? {
```
jjbayer: nit: Should we restrict this to specific `span.op`s? It does not make sense to burn CPU cycles on this if we don't expect the description to be a URL.

jjbayer (approved):

relay-general/src/store/transactions/processor.rs
```diff
@@ -438,12 +445,42 @@ impl Processor for TransactionsProcessor<'_> {
...
+fn scrub_span_description(span: &mut Span) -> Result<(), ProcessingAction> {
+    if span.description.value().is_none() {
+        return Ok(());
+    }
+
+    let mut scrubbed = span.description.clone();
+
+    if let Some(is_url_like) = span.op.value().map(|op| op.starts_with("http")) {
+        if is_url_like && scrub_identifiers(&mut scrubbed)? {
+            let Some(new_desc) = scrubbed.value() else {
```
olksdr, line 467: you could use `.into_value()` here instead of the `.value()`, since it would give the the `Option<String>` and there won't be any need to to `new_desc.to_owned()` where you allocate another string again on line `476`

PR #2096, opened 2023-05-05
fix(buffer): Throttle dequeue

jjbayer:
There is a maximum limit on in-memory envelopes in Relay (`envelope_buffer_size`). Whenever the spooler loads envelopes from disk, we consume some of this budget. If we dequeue too eagerly, there's a chance that we exhaust the budget, which will lead to Relay dropping new envelopes coming in via web requests.

* To prevent this, wait until there is a `>= 50%` budget of permits available before dequeueing a batch from disk. 

Fixes https://github.com/getsentry/team-ingest/issues/114.

#skip-changelog


relay-server/src/actors/spooler/mod.rs
```diff
@@ -317,11 +318,14 @@ impl OnDisk {
...
-            services.test_store.clone(),
-        )?;
+        let managed_envelope = self
+            .buffer_guard
+            .enter(
+                envelope,
+                services.outcome_aggregator.clone(),
+                services.test_store.clone(),
+            )
+            .await?;
```
jjbayer: This now waits until a permit becomes available.

relay-server/src/utils/buffer.rs
```diff
@@ -53,16 +60,56 @@ impl BufferGuard {
     ///
     /// Returns `Ok(ManagedEnvelope)` on success, which internally holds a handle to the reserved
     /// resources. When the managed envelope is dropped, the slot is automatically reclaimed and can
-    /// be reused by a subsequent call to `enter`.
+    /// be reused by a subsequent call to `enter` or `try_enter`.
     ///
     /// If the buffer is full, this function returns `Err`.
-    pub fn enter(
+    pub fn try_enter(
```
iker-barriocanal: Is the async state the only difference with `enter`? If yes, I'd rename this to `enter_async` or similar.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -345,7 +349,19 @@ impl OnDisk {
             // to pick up.
             //
             // Right now we use 100 for batch size.
-            let mut envelopes = sql::delete_and_fetch(key, 100).fetch(&self.db).peekable();
+            let batch_size = 100;
```
iker-barriocanal, line 402: Should this be a property of `OnDisk`?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -345,7 +349,19 @@ impl OnDisk {
...
+
+            // Before querying the db, make sure that the buffer guard has enough availability:
+            let mut attempts = 1;
+            while self.buffer_guard.used() > self.buffer_guard.capacity() / 2 {
+                tokio::time::sleep(Duration::from_millis(100)).await;
+                attempts += 1;
+            }
```
iker-barriocanal: This is a potentially infinite loop. We should break the loop after reaching a specific number of attempts. API-wise, I think this number should be a parameter of the function.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -936,6 +956,94 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn dequeue_waits_for_permits() {
+        relay_test::setup();
+        let num_permits = 3;
+        let buffer_guard: Arc<_> = BufferGuard::new(num_permits).into();
+        let config: Arc<_> = Config::from_json_value(serde_json::json!({
```
iker-barriocanal, line 1006: Do we want to make the number of attempts to fetch from disk statically configurable?

iker-barriocanal (changes requested):

relay-server/src/utils/buffer.rs
jjbayer: This follows the naming of the underlying tokio primitive:

* https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire
* https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.try_acquire

IMO this distinction is slightly more informative than `enter_async`, which only tells you that it is an `async` function.

relay-server/src/actors/spooler/mod.rs
olksdr: I do not really like that we have the loop here running and burning the CPU, what if we use something like [`Notify`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) from within the `buffer_guard`, since it knows how many permits currently issued and dropped, we just wait for notification in this place, and continue once we are notified. 

something like this (pseudocode):
```rust
if self.buffer_guard.used() > self.buffer_guard.capacity() / 2 {
    self.buffer_guard.availability.notified().await;
}
```

And the metric will be not the number of attempts , but the time spent waiting on the notification.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -936,6 +956,94 @@ mod tests {
...
+            project_key,
+            keys: [key].into(),
+            sender: tx.clone(),
+        });
+        tokio::time::sleep(Duration::from_millis(50)).await;
+
+        // There is one permit left, but we only dequeue if we gave >= 50% capacity:
+        assert!(rx.try_recv().is_err());
+
+        // Freeing one permit gives us enough capacity
```
olksdr:
```suggestion
        // Freeing one permit gives us enough capacity:
```

olksdr (changes requested):

relay-server/src/actors/spooler/mod.rs
jjbayer: Comparing two numbers every 100 ms should not burn a lot of CPU cycles, so I believe it's good enough for our use case. But I agree that some kind of condition variable would be more elegant. Will look into it.

relay-server/src/actors/spooler/mod.rs
olksdr, line 402: nit: We might want to adjust the batch size (currently we hardcode `100`) in case if the configured buffer guards number is actually smaller than the batch.

olksdr (approved):
Let's merge this PR and follow up immediately with more refactoring.

relay-server/src/actors/spooler/mod.rs
jjbayer: Worse than that, this loop blocks message handling for this service, so we have to reconsider it. Best try to unspool once, else schedule a message in 100 ms.

iker-barriocanal: I'm not sure if that works -- the actor calling this method `delete_and_fetch` expects to fetch envelopes, so if you schedule it the caller actor may not be able to fetch and we'd just delete the messages. I'd have to think this through to evaluate what the best alternative is, but failing and letting the actor retry may be the best option.

olksdr: @jjbayer updated this code a bit, please, take a look. This works , or should work, as we discussed yesterday.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -277,6 +278,7 @@ impl InMemory {
 /// The configuration which describes the on-disk [`BufferState`].
 #[derive(Debug)]
 struct OnDisk {
+    attempts: usize,
```
jjbayer: nit 
```suggestion
    dequeue_attempts: usize,
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -380,14 +385,27 @@ impl OnDisk {
         services: &Services,
     ) -> Result<(), QueueKey> {
         loop {
+            // Before querying the db, make sure that the buffer guard has enough availability:
+            self.attempts += 1;
```
jjbayer: I intended this metric to be a separate number for each `QueueKey`. Now, it is a global counter, which may not matter because it does not depend on the queue key itself but rather on the availability of buffer permits.

If we want to make it per-queue-key, we'd have to store it with the `QueueKey` itself, which I think is not worth it.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -412,7 +430,7 @@ impl OnDisk {
                     }
                 };
 
-                match self.extract_envelope(envelope, services) {
+                match self.extract_envelope(envelope, services).await {
```
jjbayer: I think we should revert the entire change that introduced `tokio::sync::Semaphore` and awaitable permits. Since we do not want to block message handling when no permit is available, that entire change does not make sense anymore.

relay-server/src/statsd.rs
```diff
@@ -64,6 +64,11 @@ pub enum RelayHistograms {
     ///
     /// This metric is computed by multiplying `page_count * page_size`.
     BufferDiskSize,
+    /// Number of attempts needed to dequeue a batch of envelopes from disk.
+    ///
+    /// As long as there are enough permits in the [`crate::utils::BufferGuard`], this number should
+    /// always be one.
```
```suggestion
    /// Number of attempts needed to dequeue anything from disk.
    ///
    /// As long as there are enough permits in the [`crate::utils::BufferGuard`], this number should
    /// always be one.
```

relay-server/src/utils/buffer.rs
```diff
@@ -28,20 +30,25 @@ impl std::error::Error for BufferError {}
 /// into the processing pipeline, use [`BufferGuard::enter`].
 #[derive(Debug)]
 pub struct BufferGuard {
-    inner: Semaphore,
+    inner: Arc<tokio::sync::Semaphore>,
```
jjbayer: As mentioned above, I would revert back to the old Semaphore. Smaller PR and no more "blocking" `enter()`.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -384,6 +385,16 @@ impl OnDisk {
...
         loop {
+            // Before querying the db, make sure that the buffer guard has enough availability:
+            self.attempts += 1;
+            if self.buffer_guard.used() > self.buffer_guard.capacity() / 2 {
+                return Err(key);
+            }
+            relay_statsd::metric!(
+                histogram(RelayHistograms::BufferDequeueAttempts) = self.attempts as u64
+            );
+            self.attempts = 0;
```
iker-barriocanal: Why are we resetting the number of attempts here? Shouldn't this happen when the buffer guard is already taken (i.e. inside the `if` above)?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -461,9 +463,15 @@ impl OnDisk {
...
         }
-        services
-            .project_cache
-            .send(UpdateBufferIndex::new(project_key, unused_keys))
+        if !unused_keys.is_empty() {
+            let project_cache = services.project_cache.clone();
+            tokio::spawn(async move {
+                // Send the message back to `ProjectCache` with a bit of delay, to prevent the DDOS
+                // of our own services.
+                tokio::time::sleep(Duration::from_millis(100)).await;
```
iker-barriocanal, line 468: I may be missing something, but I don't think this helps much. The number of messages depends on the number of times threads call `dequeue`, and this has not changed. Now, they'll just arrive a bit later but at the same pace, so it'll only help for the first few messages to get a bit later. On the other hand, we are introducing potentially a lot of tasks that will live shortly before sending the message, and I'm not sure how much work other tasks can perform. Is it worth the extra overhead?

relay-server/src/actors/spooler/mod.rs
olksdr: Will fix

iker-barriocanal (changes requested):
Blocking for (the extra complexity of) introducing awaitable permits that block message handling. We should not block message handling.

relay-server/src/actors/spooler/mod.rs
olksdr: Yeah, I think it's fine to have a metrics on how many times we tried to dequeue, and this does not matter what key is used, if there are actually no permits to dequeue anything at all.

relay-server/src/actors/spooler/mod.rs
olksdr: done

relay-server/src/utils/managed_envelope.rs
```diff
@@ -158,7 +158,7 @@ impl ManagedEnvelope {
 
     /// Consumes itself returning the managed envelope.
     ///
-    /// This also releases the slot with [`SemaphorePermit`] and sets the internal context
+    /// This also releases the slot with [`OwnedSemaphorePermit`] and sets the internal context
```
```suggestion
    /// This also releases the slot with [`SemaphorePermit`] and sets the internal context
```

jjbayer (commented):
Looks good to me! Cannot formally approve.

relay-server/src/utils/managed_envelope.rs
olksdr: heh, seen this myself. Changed it already.  in [54dd9d3](https://github.com/getsentry/relay/pull/2096/commits/54dd9d36048537a64eacd4e94deb7828c5ae460b)

iker-barriocanal (approved):

PR #2103, opened 2023-05-08
fix(span-metrics): Scrub subdomains in metric tags

iker-barriocanal:
A dummy and straight-forward approach to scrub subdomains and reduce metric cardinality. It's a best effort.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -1,5 +1,6 @@
 use std::collections::BTreeMap;
 
+use itertools::Itertools;
 use relay_common::{DurationUnit, EventType, MetricUnit, SpanStatus, UnixTimestamp};
 use relay_dynamic_config::{AcceptTransactionNames, TaggingRule, TransactionMetricsConfig};
 use relay_filter::csp::SchemeDomainPort;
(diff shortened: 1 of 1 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -479,7 +479,23 @@ fn extract_span_metrics(
...
+                                            tokens.iter().rev().take(2).rev().collect();
+                                        Some(format!(
+                                            "*.{}.{}:{}",
+                                            extracted_domain[0], extracted_domain[1], port
+                                        ))
+                                    } else {
+                                        Some(format!("{}:{}", domain, port))
+                                    }
+                                }
                                 (Some(domain), None) => Some(domain),
```
jjbayer: Shouldn't we handle the case when there's no port?

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -479,7 +479,23 @@ fn extract_span_metrics(
...
+                                        // <name>.service.<cloud-region>.<whatever>
+                                        let extracted_domain: Vec<&&str> =
+                                            tokens.iter().rev().take(2).rev().collect();
+                                        Some(format!(
+                                            "*.{}.{}:{}",
+                                            extracted_domain[0], extracted_domain[1], port
+                                        ))
+                                    } else {
+                                        Some(format!("{}:{}", domain, port))
+                                    }
```
jjbayer: nit: Should be possible to do this without allocating a vector (and without the `else` branch), but I don't have the code for it ready so feel free to ignore.

jjbayer (changes requested):
Requesting changes because of the no port case

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -543,6 +542,33 @@ fn extract_span_metrics(
...
+    // this and consider cloud domains, e.g.
+    // <name>.service.<cloud-region>.<whatever>
+    let extracted_domain: Vec<&&str> = tokens.iter().rev().take(2).rev().collect();
+    if let Some(p) = port {
+        return Some(format!(
+            "*.{}.{}:{}",
+            extracted_domain[0], extracted_domain[1], p
+        ));
+    }
+    return Some(format!("*.{}.{}", extracted_domain[0], extracted_domain[1]));
```
iker-barriocanal: It'd be more readable to generate more than one string, but I think the code isn't that complicated to make it worth the performance overhead of creating more strings.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -543,6 +542,33 @@ fn extract_span_metrics(
...
+    }
+
+    // sth.subdomain.domain.tld => *.domain.tld
+    // TODO(iker): find a proper way to do
+    // this and consider cloud domains, e.g.
+    // <name>.service.<cloud-region>.<whatever>
+    let extracted_domain: Vec<&&str> = tokens.iter().rev().take(2).rev().collect();
+    if let Some(p) = port {
+        return Some(format!(
+            "*.{}.{}:{}",
```
jjbayer: This adds `*.` even if the domain looks like `domain.com`.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -543,6 +542,33 @@ fn extract_span_metrics(
...
+            extracted_domain[0], extracted_domain[1], p
+        ));
+    }
+    Some(format!("*.{}.{}", extracted_domain[0], extracted_domain[1]))
```
jjbayer: What do you think of this simplification?
```suggestion
    let mut parts = domain.rsplitn(3, '.');
    let tld = parts.next();
    let domain = parts.next();
    let prefix = parts.next().map(|_| "*");

    let mut replaced = prefix
        .iter()
        .chain(domain.iter())
        .chain(tld.iter())
        .join(".");
    if let Some(port) = port {
        replaced = format!("{}:{}", replaced, port);
    }
    Some(replaced)
```

jjbayer (changes requested):

jjbayer (approved):

PR #2104, opened 2023-05-08
fix(profiling): Return a proper outcome for processed profiles if a profile is invalid

phacops:
We want the profile to be parsed to then mark it as invalid if needed instead of marking it as accepted no matter if we discard it or not.

Now, on top of filtering profiles early if the flag is enabled, we also parse the metadata and discard them with the appropriate outcome if needed.

We're not parsing the profile data itself and that means an indexed profile can still be considered invalid even though we already sent an accepted outcome.

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -1042,18 +1042,14 @@ impl EnvelopeProcessorService {
     /// Remove profiles from the envelope if the feature flag is not enabled.
     fn filter_profiles(&self, state: &mut ProcessEnvelopeState) {
         let profiling_enabled = state.project_state.has_feature(Feature::Profiling);
-        let timestamp = state.managed_envelope.received_at();
-        let scoping = state.managed_envelope.scoping();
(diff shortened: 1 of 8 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -1042,8 +1042,29 @@ impl EnvelopeProcessorService {
...
+                            scoping,
+                            outcome: Outcome::Invalid(DiscardReason::Profiling(
+                                relay_profiling::discard_reason(err),
+                            )),
+                            event_id: None,
+                            remote_addr: None,
+                            category: DataCategory::Profile,
+                            quantity: 1,
+                        });
+                        ItemAction::DropSilently
```
```suggestion
                        ItemAction::Drop(Outcome::Invalid(DiscardReason::Profiling(
                            relay_profiling::discard_reason(err),
                        )))
```

jjbayer: Any reason not to use the `Drop(Outcome)` variant here?

jjbayer (approved):

viglia (approved):

PR #2106, opened 2023-05-09
feat(spans): Add span description tag on in-like sql queries

iker-barriocanal:
On `db.*`-like span statuses, scrub the span description looking for SQL queries and add the scrubbed strings as tags to span metrics. This PR only considers one use case of SQL queries: `IN ...`-like queries; new use cases will come in future PRs.

#skip-changelog

### DIFF
--- a/relay-general/src/store/regexes.rs
+++ b/relay-general/src/store/regexes.rs
@@ -50,3 +50,10 @@ pub static TRANSACTION_NAME_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     )
     .unwrap()
 });
+
+pub static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
+    // Slightly modified regex from
+    // https://github.com/getsentry/sentry/blob/244b33e44bbbfa0dd680f5a15053e2efaaf6fd65/src/sentry/spans/grouping/strategy/base.py#L132
+    Regex::new(r#" IN \((?P<in_condition_params>(%s|\$?\d+|\?)(\s*,\s*(%s|\$?\d+|\?))*)\)"#)
+        .unwrap()
+});
--- a/relay-general/src/store/regexes.rs
+++ b/relay-general/src/store/regexes.rs
@@ -54,7 +54,7 @@ pub static TRANSACTION_NAME_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
 /// Regex with multiple capture groups for SQL tokens we should scrub.
 ///
 /// Slightly modified from
-/// https://github.com/getsentry/sentry/blob/244b33e44bbbfa0dd680f5a15053e2efaaf6fd65/src/sentry/spans/grouping/strategy/base.py#L132
+/// <https://github.com/getsentry/sentry/blob/244b33e44bbbfa0dd680f5a15053e2efaaf6fd65/src/sentry/spans/grouping/strategy/base.py#L132>
 pub static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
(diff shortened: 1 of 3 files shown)


relay-general/src/store/regexes.rs
```diff
@@ -50,3 +50,10 @@ pub static TRANSACTION_NAME_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     )
     .unwrap()
 });
+
+pub static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
+    // Slightly modified regex from
+    // https://github.com/getsentry/sentry/blob/244b33e44bbbfa0dd680f5a15053e2efaaf6fd65/src/sentry/spans/grouping/strategy/base.py#L132
+    Regex::new(r#" IN \((?P<in_condition_params>(%s|\$?\d+|\?)(\s*,\s*(%s|\$?\d+|\?))*)\)"#)
+        .unwrap()
```
TBS1996:
```suggestion
    Regex::new(r#"(?i) IN \((?P<in_condition_params>(%s|\$?\d+|\?)(\s*,\s*(%s|\$?\d+|\?))*)\)"#)
        .unwrap()
```

this should make the lowercase unit test pass

relay-general/src/store/transactions/processor.rs
```diff
@@ -276,16 +278,25 @@ fn set_default_transaction_source(event: &mut Event) {
...
+}
+
+/// Normalize the given SQL-query-like string.
+fn scrub_sql_queries(string: &mut Annotated<String>) -> Result<bool, ProcessingAction> {
+    scrub_identifiers_with_regex(string, &SQL_NORMALIZER_REGEX)
+}
+
+fn scrub_identifiers_with_regex(
+    string: &mut Annotated<String>,
+    regex: &Lazy<Regex>,
```
olksdr: Can we not call it `regex` ? Since this is the name of the crate as well, which can be confusing. 
Maybe something like 
```suggestion
    re: &Lazy<Regex>,
```
Which usually the way regexes called in the code. Or `pattern` also will do.

relay-general/src/store/transactions/processor.rs
```diff
@@ -461,23 +472,34 @@ fn scrub_span_description(span: &mut Span) -> Result<(), ProcessingAction> {
...
+            return Ok(());
+        };
+        span.data
+            .get_or_insert_with(BTreeMap::new)
+            // We don't care what the cause of scrubbing was, since we assume
+            // that after scrubbing the value is sanitized.
+            .insert(
+                "description.scrubbed".to_owned(),
+                Annotated::new(Value::String(new_desc.to_owned())),
+            );
```
olksdr: I already left this comment in https://github.com/getsentry/relay/pull/2095#discussion_r1188215948, to repeat: 

* the `scrubbed` var is already cloned description
* you could use `.into_value()` here instead of the `.value()`
* since it would give the the `Option<String>` and there won't be any need to do `new_desc.to_owned()` where you allocate another string again on line `499`

relay-general/src/store/transactions/processor.rs
```diff
@@ -461,23 +472,34 @@ fn scrub_span_description(span: &mut Span) -> Result<(), ProcessingAction> {
...
+    if let Some(is_db_like) = span.op.value().map(|op| op.starts_with("db")) {
+        if is_db_like && scrub_sql_queries(&mut scrubbed)? {
+            did_scrub = true;
+        }
+    }
+
+    if did_scrub {
+        let Some(new_desc) = scrubbed.value() else {
+            return Ok(());
+        };
```
olksdr: There is also no need for early return here, the following code is super short and can be used directly in the `if` statement. 

```suggestion
        if let Some(new_desc) = scrubbed.into_value() {

```

olksdr (changes requested):

iker-barriocanal:
Rerequesting reviews after pushing the code. Let's give GitHub some time to show the new commits.

olksdr (approved):

PR #2107, opened 2023-05-10
feat(profiling): Fix regression on making release optional and copy dist from transaction

phacops:
This PR aims to fix the regression introduced right after we started making the release field option.

At the same time, `dist` should also be copied from the transaction and should also be optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,9 +31,8 @@
 - On Linux, switch to `jemalloc` instead of the system memory allocator to reduce Relay's memory footprint. ([#2084](https://github.com/getsentry/relay/pull/2084))
 - Scrub sensitive cookies `__session`. ([#2105](https://github.com/getsentry/relay/pull/2105)))
 - Parse profiles' metadata to check if it should be marked as invalid. ([#2104](https://github.com/getsentry/relay/pull/2104))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 - Parse profiles' metadata to check if it should be marked as invalid. ([#2104](https://github.com/getsentry/relay/pull/2104))
 - Set release as optional by defaulting to an empty string for profiles. ([#2098](https://github.com/getsentry/relay/pull/2098))
 - Accept source map debug images in debug meta for Profiling. ([#2097](https://github.com/getsentry/relay/pull/2097))
+- Fix regression on release being optional and copy missing dist field from transactions.([#2107](https://github.com/getsentry/relay/pull/2107))
```
jjbayer: Since #2098 and #2107 will go into the same release, I would summarize the change in a single line (with the PR links comma-separated).

jjbayer (approved):
The change by itself looks good to me, but what was the actual regression?

Zylphrex (approved):

phacops:
> The change by itself looks good to me, but what was the actual regression?

I made the release optional here https://github.com/getsentry/relay/commit/018ef16eb3338ede023a2f5f6a6a4c703cb6a64e. Two commits later, I reverted because of a bad merge https://github.com/getsentry/relay/commit/25557414f84a6563dc3c0627d79e5ad1111cf124.

PR #2108, opened 2023-05-10
chore(metrics): Add more tags to existing transaction metrics

olksdr:
Add following tags to transaction metrics:
* browser name
* os name
* geo country code

part of https://github.com/getsentry/team-ingest/issues/123

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -59,6 +59,36 @@ fn extract_http_method(transaction: &Event) -> Option<String> {
     Some(method.clone())
 }
 
+/// Extract the browser name from the [`Context::Browser`] context.
+fn extract_browser_name(event: &Event) -> Option<String> {
+    let contexts = event.contexts.value()?;
+    let browser = contexts.get("browser").and_then(Annotated::value);
+    if let Some(ContextInner(Context::Browser(browser_context))) = browser {
+        return browser_context.name.value().cloned();
+    }
+
+    None
+}
+
+/// Extract the OS name from the [`Context::Os`] context.
+fn extract_os_name(event: &Event) -> Option<String> {
+    let contexts = event.contexts.value()?;
+    let os = contexts.get("os").and_then(Annotated::value);
+    if let Some(ContextInner(Context::Os(os_context))) = os {
+        return os_context.name.value().cloned();
+    }
+
+    None
+}
+
+/// Extract the OS name from the [`Context::Os`] context.
+fn extract_geo_country_code(user: &User) -> Option<String> {
+    if let Some(geo) = user.geo.value() {
+        return geo.country_code.value().cloned();
+    }
+    None
+}
+
 fn is_low_cardinality(source: &TransactionSource, treat_unknown_as_low_cardinality: bool) -> bool {
     match source {
         // For now, we hope that custom transaction names set by users are low-cardinality.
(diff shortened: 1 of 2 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -189,6 +219,20 @@ fn extract_universal_tags(event: &Event, config: &TransactionMetricsConfig) -> C
...
+    }
+
+    if let Some(os_name) = extract_os_name(event) {
+        tags.insert(CommonTag::OsName, os_name);
+    }
+
+    if let Some(user) = event.user.value() {
+        if let Some(geo_country_code) = extract_geo_country_code(user) {
+            tags.insert(CommonTag::GeoCountryCode, geo_country_code);
+        }
```
jjbayer: nit: The other two `extract_` functions take the full event as parameter, even though they only need `event.contexts`. So I would change the signature of `extract_geo_country_code()` to take an `&Event` as well.

jjbayer (approved):
Please notify Search and Storage when this change goes out, so they know where the (potential) increase in cardinality comes from.

relay-server/src/metrics_extraction/transactions/mod.rs
olksdr: makes sense, done in [3fd3299](https://github.com/getsentry/relay/pull/2108/commits/3fd3299cfac6d03b0791cadcda69a38c6ac0c549)

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -59,6 +59,39 @@ fn extract_http_method(transaction: &Event) -> Option<String> {
...
+    let contexts = event.contexts.value()?;
+    let os = contexts.get("os").and_then(Annotated::value);
+    if let Some(ContextInner(Context::Os(os_context))) = os {
+        return os_context.name.value().cloned();
+    }
+
+    None
+}
+
+/// Extract the OS name from the [`Context::Os`] context.
```
matejminar: Just passing by and noticed that this is probably a bad copypaste.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -59,6 +59,39 @@ fn extract_http_method(transaction: &Event) -> Option<String> {
     Some(method.clone())
 }
 
+/// Extract the browser name from the [`Context::Browser`] context.
+fn extract_browser_name(event: &Event) -> Option<String> {
+    let contexts = event.contexts.value()?;
+    let browser = contexts.get("browser").and_then(Annotated::value);
+    if let Some(ContextInner(Context::Browser(browser_context))) = browser {
+        return browser_context.name.value().cloned();
+    }
```
matejminar, line 68: Out of curiosity, are these already sanitized/normalized, or can this yield 17 ways of saying "Chrome"?
Same question for OS.

olksdr: Yes, they supposed to be normalized.

relay-server/src/metrics_extraction/transactions/mod.rs
olksdr: Thanks! Fixed in [daf44e3](https://github.com/getsentry/relay/pull/2108/commits/daf44e3fc2ef907b38740775b98709e0b3bcf6d3)

iker-barriocanal (approved):

PR #2112, opened 2023-05-11
ref(spool): Simplify the spool states

olksdr:
Simplify the state machine and have only 2 transition, from memory to disk and from disk to memory. 

Right now when the disk is full we start dropping the incoming envelopes and report the error.

In the followup will integrate the health check to report this issue to the user more gracefully. 

related: https://github.com/getsentry/team-ingest/issues/110

#skip-changelog

### DIFF
--- a/relay-server/src/actors/spooler/mod.rs
+++ b/relay-server/src/actors/spooler/mod.rs
@@ -7,7 +7,7 @@ use futures::stream::{self, StreamExt};
 use relay_common::ProjectKey;
 use relay_config::Config;
 use relay_log::LogError;
-use relay_system::{Addr, Controller, FromMessage, Interface, Service};
+use relay_system::{Addr, Controller, FromMessage, Interface, Sender, Service};
 use sqlx::migrate::MigrateError;
 use sqlx::sqlite::{
     SqliteAutoVacuum, SqliteConnectOptions, SqliteJournalMode, SqlitePoolOptions, SqliteRow,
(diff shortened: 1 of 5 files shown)


relay-server/src/actors/spooler/mod.rs
```diff
@@ -673,14 +644,14 @@ impl BufferService {
...
-            // anything from the on-disk spool, we will eventually hit the
-            // `cache.envelope_buffer_size` limit or OOM, whatever comes first.
-            | BufferState::MemoryFileRead { ref mut ram, .. } => {
+            | BufferState::MemoryFileStandby { ref mut ram, .. } => {
                 ram.enqueue(key, managed_envelope);
             }
-            BufferState::Disk(ref disk) =>{
+            BufferState::Disk(ref disk) => {
+                // The disk is full, drop the incoming envelopes.
+                if disk.is_full().await? {
```
jjbayer, line 721: Idea for a potential follow-up: `is_full` is also called in `transition()` for every `handle_enqueue`, I wonder if we could cache it as a boolean flag on `OnDisk`, which gets updated on dequeue.

jjbayer (approved):

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: I'm not sure if I understand -- should `transition()` fail if we try to transition to `OnDisk` when the disk is full?

iker-barriocanal (approved):

relay-server/src/actors/spooler/mod.rs
olksdr: It will never happen, since we do not transition from Disk to Memory if the Disk has some data.

PR #2113, opened 2023-05-11
chore(spool): Integrate the health check into the spooler

olksdr:
Once the disk is full, we will report the unhealthy to the client, to signal that the Relay cannot accept any more envelopes.

related: https://github.com/getsentry/team-ingest/issues/110

#skip-changelog

### DIFF
--- a/relay-server/src/actors/health_check.rs
+++ b/relay-server/src/actors/health_check.rs
@@ -6,6 +6,7 @@ use relay_metrics::{AcceptsMetrics, Aggregator};
 use relay_statsd::metric;
 use relay_system::{Addr, AsyncResponse, Controller, FromMessage, Interface, Sender, Service};
 
+use crate::actors::project_cache::{ProjectCache, SpoolHealth};
 use crate::actors::upstream::{IsAuthenticated, IsNetworkOutage, UpstreamRelay};
 use crate::statsd::RelayGauges;
 
@@ -41,6 +42,7 @@ pub struct HealthCheckService {
     config: Arc<Config>,
     aggregator: Addr<Aggregator>,
     upstream_relay: Addr<UpstreamRelay>,
+    project_cache: Addr<ProjectCache>,
 }
 
 impl HealthCheckService {
@@ -51,12 +53,14 @@ impl HealthCheckService {
         config: Arc<Config>,
         aggregator: Addr<Aggregator>,
         upstream_relay: Addr<UpstreamRelay>,
+        project_cache: Addr<ProjectCache>,
     ) -> Self {
         HealthCheckService {
             is_shutting_down: AtomicBool::new(false),
             config,
             aggregator,
             upstream_relay,
+            project_cache,
         }
     }
 
@@ -85,8 +89,17 @@ impl HealthCheckService {
                     return false;
                 }
 
-                self.aggregator
+                if !self
+                    .aggregator
                     .send(AcceptsMetrics)
+                    .await
+                    .unwrap_or_default()
+                {
+                    return false;
+                }
+
+                self.project_cache
+                    .send(SpoolHealth)
                     .await
                     .unwrap_or_default()
             }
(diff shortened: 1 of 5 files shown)


relay-server/src/actors/spooler/mod.rs
```diff
@@ -719,12 +732,25 @@ impl BufferService {
         Ok(())
     }
 
+    async fn handle_health(&mut self, health: Health) -> Result<(), BufferError> {
+        match self.state {
+            // We do not check the ram, since we rely on `cache.envelope_buffer_size` to limit the
+            // memory usage and the number of the envelopes in the memory.
+            // Note: in the future we want to switch to `spool.envelopes.max_memory_size` option.
+            BufferState::Memory(_) | BufferState::MemoryFileStandby { .. } => health.0.send(true),
+            BufferState::Disk(ref disk) => health.0.send(disk.is_full().await.unwrap_or_default()),
```
jjbayer: If boolean `true` means healthy, we should send `!disk.is_full()` here. Then the question is, use `unwrap_or_default()` or `unwrap_or(true)` (depends on the behavior)?

As discussed, an integration test would be good.

jjbayer (changes requested):

relay-server/src/actors/spooler/mod.rs
olksdr: @jjbayer  fixed and added integration test in [9419c93](https://github.com/getsentry/relay/pull/2113/commits/9419c938c33b4bfade867e04d6d2dceb8eaa2172)

relay-server/src/actors/health_check.rs
```diff
@@ -85,10 +89,19 @@ impl HealthCheckService {
...
                 }
 
-                self.aggregator
+                if !self
+                    .aggregator
                     .send(AcceptsMetrics)
                     .await
                     .unwrap_or_default()
+                {
+                    return false;
```
iker-barriocanal, line 98: Does this trigger any alerts? Should OPS be aware of this change?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -719,12 +732,25 @@ impl BufferService {
         Ok(())
     }
 
+    async fn handle_health(&mut self, health: Health) -> Result<(), BufferError> {
+        match self.state {
+            // We do not check the ram, since we rely on `cache.envelope_buffer_size` to limit the
```
iker-barriocanal, line 737: nit
```suggestion
            // We do not check the RAM, since we rely on `cache.envelope_buffer_size` to limit the
```

tests/integration/test_healthchecks.py
```diff
@@ -97,3 +99,56 @@ def test_readiness_depends_on_aggregator_being_full(mini_sentry, relay):
...
+
+        # These events will consume all the disk sapce and we will report not ready.
+        relay.send_event(project_key)
+        relay.send_event(project_key)
+
+        response = wait_get(relay, "/api/relay/healthcheck/ready/")
+        assert response.status_code == 503
+    finally:
+        # Authentication failures would fail the test
+        mini_sentry.test_failures.clear()
```
iker-barriocanal, line 154: I don't really understand why we need this.

iker-barriocanal (approved):

tests/integration/test_healthchecks.py
jjbayer: @iker-barriocanal our testing harness has some magic which makes every test fail as soon as the relay instances logs an error to mini_sentry. So if you _expect_ error events, you have to explicitly clear this list.

@olksdr this would also work without a `try..finally` though, right?

jjbayer (approved):

tests/integration/test_healthchecks.py
olksdr: Yeah, I just re-used existing tests, to make them exactly same for uniformity.

relay-server/src/actors/health_check.rs
olksdr: The relay will behave as it did before, when the permits are exhausted, but in cleaner way.

Issue #1995, opened 2023-03-31
PII scrub bank accounts


iker-barriocanal:
Relay should be able to PII scrub bank accounts (e.g. `IBAN.....`).

TBS1996:
now that we're done with IBAN, I try to find other types of bank numbers we can scrub. BIC seemed logical, however, the format is too broad that I think we would have too many false positives.

do you have any thoughts on this? should we perhaps close this issue?

iker-barriocanal:
I don't think there's anything immediately actionable, so closing the issue. We can consider reopening it in the future or creating a new one if there're more requests.

PR #2117, opened 2023-05-12
feat(general): Scrub Iban as pii.

TBS1996:
https://github.com/getsentry/relay/issues/1995

i plan to make a future PR where I also scrub ibans with spaces in this format: `VG96 VPVG 0000 0123 4567 8901`

currently it scrubs any iban length from the shortest (norway) to the longest (russia). However, it does not check if the country-code matches the length. for example it will scrub an iban with the russian country code but the length of norway, despite that being invalid.

I'm looking for feedback on whether this approach is too aggressive, I might want to fine it more later to make the length of the regex match with the country code given.


relay-general/src/pii/regexes.rs
```diff
@@ -158,6 +159,17 @@ static EMAIL_REGEX: Lazy<Regex> = Lazy::new(|| {
     .unwrap()
 });
 
+static IBAN_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(
+        r#"(?x)
+            \b
+            [A-Z]{2}\d{2}[A-Z0-9]{11,29}
```
TBS1996: this regex matches any iban length from the shortest (norway) to the longest (russia). However, it does not check if the country-code matches the length. for example it will scrub an iban with the russian country code but the length of norway, despite that being invalid. 

I figured i'd wanna start out simple and maybe in the future add some additional logic.

relay-general/src/pii/builtin.rs
```diff
@@ -193,6 +194,33 @@ declare_builtin_rules! {
...
+    "@iban:filter" => RuleSpec {
+        ty: RuleType::Iban,
+        redaction: Redaction::Replace(ReplaceRedaction {
+            text: "[Filtered]".into(),
+        }),
+    };
+    "@iban:remove" => RuleSpec {
+        ty: RuleType::Iban,
+        redaction: Redaction::Remove
+    };
```
TBS1996, line 223: I looked at how creditcard is handled and I did the same

relay-general/src/pii/builtin.rs
```diff
@@ -779,6 +832,217 @@ mod tests {
...
+        assert_rule_not_applied!(
+            rule = "@iban";
+            input = "some iban: N19386011117945!";
+        );
+    }
+
+    #[test]
+    fn test_valid_iban_codes() {
+        // list taken from: `https://github.com/jschaedl/iban-validation/blob/master/tests/ValidatorTest.php`
+        // Todo: verify that it's okay to copy it.
```
TBS1996: I got this list from an MIT lisenced repository, is that enough to think of it as free to use, or do i need to also verify their source?

relay-general/src/pii/builtin.rs
```diff
@@ -36,6 +36,7 @@ declare_builtin_rules! {
     "@common" => RuleSpec {
         ty: RuleType::Multiple(MultipleRule {
             rules: vec![
+                "@iban".into(),
```
jjbayer, line 39: Please also add to `@common:filter`, which is the one used by default datascrubbing settings:

https://github.com/getsentry/relay/blob/dbb7569d74246ec5bbd7d10bfeab8e18f896500b/relay-general/src/pii/convert.rs#L66-L67

relay-general/src/pii/builtin.rs
```diff
@@ -779,6 +832,217 @@ mod tests {
...
+            "SK3112000000198742637541",
+            "SM86U0322509800000000270100",
+            "SV62CENR00000000000000700025",
+            "TL380080012345678910157",
+            "TN5910006035183598478831",
+            "TR330006100519786457841326",
+            "UA213223130000026007233566001",
+            "VA59001123000012345678",
+            "VG96VPVG0000012345678901",
+            "XK051212012345678906",
```
jjbayer: Let's replace the digits / letters with random characters just to make sure we're not publishing PII here.

relay-general/src/pii/regexes.rs
jjbayer: Should we make this case insensitive? We shouldn't let an IBAN slip through the cracks just because it's lowercase.
```suggestion
            [a-zA-Z]{2}\d{2}[a-zA-Z0-9]{11,29}
```

jjbayer (approved):
Please look at the two suggestions before merging.

relay-general/src/pii/regexes.rs
TBS1996: well theres a tradeoff where if it becomes too aggressive we'll start scrubbing actual useful information thats not iban, im not strongly opinionated where the balance should be

relay-general/src/pii/builtin.rs
TBS1996: ok, i did so with this script:

``` py
import random

def randomize_digits(input_string):
    output_string = ""
    for char in input_string:
        if char.isdigit():
            char = str(random.randint(0, 9))
        output_string += char
    return output_string
```

relay-general/src/pii/builtin.rs
```diff
@@ -54,6 +55,7 @@ declare_builtin_rules! {
         ty: RuleType::Multiple(MultipleRule {
             rules: vec![
                 "@creditcard:filter".into(),
+                "@iban:filter".into(),
```
TBS1996, line 58: add iban to common:filter

iker-barriocanal (approved):
> it does not check if the country-code matches the length

I think that's alright. Having a good RegEx to match valid IBANs can be complicated. I think this PR is good enough and I'd move forward with it, and improve the RegEx on user feedback.

PR #2118, opened 2023-05-12
feat(filter): Add a health check filter

RaduW:
This PR implements a filter for health-check events

Relates to https://github.com/getsentry/sentry/issues/49082


relay-filter/src/common.rs
```diff
@@ -127,6 +127,9 @@ pub enum FilterStatKey {
 
     /// Filtered due to invalid CSP policy.
     InvalidCsp,
+
+    /// Filtered due to the fact that is was a call to a health-check endpoint
```
iker-barriocanal: nit:
```suggestion
    /// Filtered due to the fact that it was a call to a health-check endpoint
```

relay-filter/src/health_check.rs
```diff
@@ -0,0 +1,162 @@
+//! Implements event filtering based on whether the endpoint called is a healthcheck endpoint
+//!
+//! If this filter is enabled messages to healthcheck endpoints will be filtered out
```
iker-barriocanal: nit:
```suggestion
//! Implements event filtering based on whether the endpoint called is a healthcheck endpoint.
//!
//! If this filter is enabled messages to healthcheck endpoints will be filtered out.
```

relay-filter/src/health_check.rs
```diff
@@ -0,0 +1,162 @@
...
+        ready|
+        heartbeat|
+        /health$|
+        /healthz$
+        "#,
+    )
+    .expect("Invalid healthcheck filter Regex")
+});
+
+fn matches(_event: &Event) -> bool {
```
iker-barriocanal: I think this is a cargo warning?
```suggestion
fn matches(event: &Event) -> bool {
```

relay-filter/src/health_check.rs
```diff
@@ -0,0 +1,162 @@
...
+    if let Some(transaction) = _event.transaction.value() {
+        HEALTH_CHECK_ENDPOINTS.is_match(transaction)
+    } else {
+        false
+    }
```
iker-barriocanal: I think we can simplify this a bit:
```suggestion
    _event.transaction.value().map_or(false, |transaction| {
        HEALTH_CHECK_ENDPOINTS.is_match(transaction)
    })
```

relay-filter/src/health_check.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/// Filters events for calls to healthcheck endpoints
```
iker-barriocanal: Should we explicitly mention that this only applies to transactions?

relay-filter/src/health_check.rs
```diff
@@ -0,0 +1,162 @@
...
+        healthy|
+        live|
+        ready|
+        heartbeat|
+        /health$|
+        /healthz$
+        "#,
+    )
+    .expect("Invalid healthcheck filter Regex")
+});
```
iker-barriocanal: How often will this list get updated? Are we planning on new inbound filters on transaction names?

The health check filter checks transaction names, there's nothing specific to health checks besides this pattern. However, if we need to change the pattern, it requires redeploying Relay. Could we benefit from this filter if we make it more generic, and accept a pattern from Sentry? This is the approach with the [client IPs inbound filter](https://github.com/getsentry/relay/blob/dd93441bbb607dba806c6f1ca0c03783e348f048/relay-filter/src/client_ips.rs#L43). This will give us more flexibility. What do you think?

iker-barriocanal (changes requested):
The PR looks good, but I want to clarify the configurability of the health check pattern from Sentry, see the comment below.

relay-filter/src/health_check.rs
RaduW: yep, remained _event from the initial commit when it was just a skeleton not using the event

relay-filter/src/health_check.rs
RaduW: I don't really know, we had a discussion and it was decided that we'd rather update Relay than add this to the configuration of each project ( I guess the implication is that it will not be updated very often).

relay-filter/src/health_check.rs
RaduW: ok, will do

iker-barriocanal (approved):
The config will live in Relay for now.

iambriccardo (approved):

relay-filter/src/config.rs
```diff
@@ -136,6 +136,20 @@ pub struct ErrorMessagesFilterConfig {
     pub patterns: GlobPatterns,
 }
 
+/// Configuration for the error messages filter.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct HealthCheckEndpointsFilterConfig {
+    /// List of healthcheck patterns that will be filtered.
```
iker-barriocanal: Is this solution generic for error messages (docs here), a generic solution for any pattern but only applying to transactions (what tests seem to be targeting), or only for health check endpoints (what the struct name indicates)?

relay-filter/src/health_check.rs
```diff
@@ -0,0 +1,154 @@
...
+        };
+
+        let filter_result = should_filter(&event, &config);
+        assert_eq!(
+            filter_result,
+            Err(FilterStatKey::HealthCheck),
+            "Event did not filter health check event"
+        )
+    }
+    #[test]
```
iker-barriocanal: nit:
```suggestion
    }

    #[test]
```

iker-barriocanal (dismissed):
Blocking for naming inconsistencies.

relay-filter/src/config.rs
RaduW: It was a copy and paste error ( the config is similar with the ErrorMessageFielterConfig), fixed comment.

relay-filter/src/health_check.rs
RaduW: thanks, fixed

relay-filter/src/config.rs
```diff
@@ -210,6 +224,13 @@ pub struct FiltersConfig {
     /// Configuration for the releases filter.
     #[serde(default, skip_serializing_if = "ReleasesFilterConfig::is_empty")]
     pub releases: ReleasesFilterConfig,
+
+    /// Configuration for the healthcheck filter.
+    #[serde(
+        default,
+        skip_serializing_if = "HealthCheckEndpointsFilterConfig::is_empty"
+    )]
+    pub health_check: HealthCheckEndpointsFilterConfig,
```
iker-barriocanal: This filter works for any pattern, not just health checks. Could we rename it to reflect it (e.g. `transaction_names`)? It's ok if we just use it for health checks, but once we introduce `health_check` we should keep it for backward compatibility.

iker-barriocanal (changes requested):

iker-barriocanal (approved):

PR #2121, opened 2023-05-15
feat(spans): Scrub bools in SQL queries in span descriptions

iker-barriocanal:
Scrub bools in span descriptions containing SQL queries. This is not only for the cardinality but also for a product requirement.

#skip-changelog

### DIFF
--- a/relay-general/src/store/regexes.rs
+++ b/relay-general/src/store/regexes.rs
@@ -62,7 +62,7 @@ pub static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     # Capture parameters in `IN` statements.
     ((?-x)IN \((?P<in>(%s|\$?\d+|\?)(\s*,\s*(%s|\$?\d+|\?))*)\)) |
     # Capture `SAVEPOINT` savepoints
-    ((?-x)SAVEPOINT (?P<savepoint>(?:(?:"[^"]+")|(?:`[^`]+`)|(?:[a-z]\w+))))
+    ((?-x)SAVEPOINT (?P<savepoint>(?:(?:"[^"]+")|(?:'[^']+')|(?:`[^`]+`)|(?:[a-z]\w+))))
     "#,
     )
     .unwrap()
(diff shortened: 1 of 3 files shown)


relay-general/src/store/regexes.rs
```diff
@@ -68,7 +68,9 @@ pub static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     ((?-x)(?P<single_quoted_strs>('(?:[^']|'')*?(?:\\'.*|[^']')))) |
     # Don't capture double-quoted strings (eg used for identifiers in PostgreSQL).
     # Capture numbers.
-    ((?-x)(?P<number>(-?\b(?:[0-9]+\.)?[0-9]+(?:[eE][+-]?[0-9]+)?\b)))
+    ((?-x)(?P<number>(-?\b(?:[0-9]+\.)?[0-9]+(?:[eE][+-]?[0-9]+)?\b))) |
+    # Capture booleans (as full tokens, not as substrings of other tokens).
+    ((?-x)(?P<bool>(\b(?:true|false)\b)))
```
olksdr, line 73: Looks good, but do we really have to scrub booleans?

olksdr (approved):

relay-general/src/store/regexes.rs
jjbayer, line 73: Is this case sensitive? Because

`select tRuE or FaLsE` is valid postgreSQL.

jjbayer: +1, booleans are not high cardinality, so why bother?

jjbayer (approved):

relay-general/src/store/regexes.rs
iker-barriocanal: Yes, the `i` modifier in the first line of the RegEx is for case insensitive. I added a test to ensure this in [a17ce05](https://github.com/getsentry/relay/pull/2121/commits/a17ce057c242072709f4b059a5bac555610ef936).

iker-barriocanal: In this case, scrubbing is not for high cardinality, but a product requirement. Updated the PR description.

PR #2122, opened 2023-05-15
test(spans): Don't scrub NULLs in SQL queries in span descriptions

iker-barriocanal:
Ensure `NULL`s in SQL queries in span descriptions are not scrubbed.

#skip-changelog

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -2339,8 +2339,8 @@ mod tests {
     span_description_test!(
         span_description_scrub_only_urllike_on_http_ops,
         "GET https://www.service.io/resources/01234",
-        "db.sql.query",
-        ""
+        "http.client",
+        "GET https://www.service.io/resources/*"
     );
 
     span_description_test!(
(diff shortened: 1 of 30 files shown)


relay-general/src/store/transactions/processor.rs
```diff
@@ -2568,4 +2568,11 @@ mod tests {
         "db.sql.query",
         ""
     );
+
+    span_description_test!(
+        span_description_dont_scrub_nulls,
+        "SELECT * FROM table WHERE deleted_at IS NULL",
```
jjbayer, line 2580: Question about placeholders in general: Since `*` is a [special character](https://www.postgresql.org/docs/current/sql-syntax-lexical.html) in SQL, should we use a different placeholder when we scrub something?

jjbayer (approved):

olksdr (approved):

relay-general/src/store/transactions/processor.rs
iker-barriocanal: Good comment! I addressed this in a follow-up: https://github.com/getsentry/relay/pull/2126.

PR #2126, opened 2023-05-15
fix(spans): Choose non-special character for scrubbed SQL descriptions

iker-barriocanal:
Follow-up to https://github.com/getsentry/relay/pull/2122#discussion_r1193714147.

`*` is a [special character](https://www.postgresql.org/docs/current/sql-syntax-lexical.html) in SQL, so we should have another way to reference to scrubbed values.

#skip-changelog


relay-general/src/store/transactions/processor.rs
```diff
@@ -317,11 +318,11 @@ fn scrub_identifiers_with_regex(
 
         // Sort by the capture end position.
         caps.sort_by_key(|(capture, _)| capture.end());
-        let mut changed = String::with_capacity(trans.len());
+        let mut changed = String::with_capacity(trans.len() + caps.len() * replacer.len());
```
iker-barriocanal, line 321: Not sure if this was necessary, but since the new string may now be larger than the previous one I wanted to make sure no unnecessary allocations are made.

jjbayer (approved):
LGTM!

PR #2127, opened 2023-05-16
feat(spans): Extract metrics on scrubbed cache descriptions

iker-barriocanal:
Scrubs span descriptions on cache-like spans (e.g. Redis) on extracted metrics.

#skip-changelog

### DIFF
--- a/relay-general/src/store/regexes.rs
+++ b/relay-general/src/store/regexes.rs
@@ -75,3 +75,16 @@ pub static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     )
     .unwrap()
 });
+
+pub static CACHE_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(
+        r#"(?xi)
+        # Don't scrub the first segment.
+        # Capture hex.
+        (([\s.+:/\-])+(?P<hex>[a-fA-F0-9]+\b)+) |
+        # Capture segments, in form of`:{hi}:`
+        (([\s.+:/\-])+(?P<segment>\{[^\}]*\})+)
+    "#,
+    )
+    .unwrap()
+});
--- a/relay-general/src/store/regexes.rs
+++ b/relay-general/src/store/regexes.rs
@@ -76,6 +76,11 @@ pub static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     .unwrap()
 });
 
+/// Regex with multiple capture groups for cache tokens we should scrub.
+///
+/// The regex attempts to identify all tokens based on hex chars and segments,
+/// excluding the first token. A segment is a string inside curly braces after a
+/// separator, for example `notsegment:{segment}:notsegment`.
 pub static CACHE_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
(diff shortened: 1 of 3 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -833,6 +833,19 @@ mod tests {
                         "db.system": "MyDatabase",
                         "db.operation": "SELECT"
                     }
+                },
+                {
+                    "description": "GET cache:user:{123}",
```
jjbayer, line 838: Where do the curly braces come from? Do SDKs send these?

relay-general/src/store/transactions/processor.rs
```diff
@@ -2582,4 +2594,11 @@ mod tests {
         "db.sql.query",
         ""
     );
+
+    span_description_test!(
+        span_description_scrub_cache,
+        "abc:12:{def}:{34}:{fg56}:EAB38:zookeeper",
+        "cache.get_item",
+        "abc:*:*:*:*:*:zookeeper"
+    );
```
jjbayer, line 2603: Would be great to have a negative test as well. That is, `op = cache...` but a none-matching description.

jjbayer (approved):
See comments below.

relay-general/src/store/regexes.rs
```diff
@@ -75,3 +75,16 @@ pub static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     )
     .unwrap()
 });
+
+pub static CACHE_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
```
olksdr, line 84: Please, add doc string for this regex.

relay-general/src/store/transactions/processor.rs
olksdr: +1

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: That's the key for the cache, not dependent on SDKs and dependent on users. I believe curly braces come from users potentially using them from variables, I'm not aware of any other reason.

relay-general/src/store/transactions/processor.rs
```diff
@@ -487,6 +493,12 @@ fn scrub_span_description(span: &mut Span) -> Result<(), ProcessingAction> {
         }
     }
 
+    if let Some(is_cache_like) = span.op.value().map(|op| op.starts_with("cache")) {
+        if is_cache_like && scrub_cache_keys(&mut scrubbed)? {
+            did_scrub = true;
+        }
+    }
```
jjbayer, line 500: This check and the one above should be `else if`: No need to check the prefix again if it already matched another one. You could also simplify to something like

```suggestion
    if span.op.value().map_or(false, |op| op.starts_with("cache")) {
        did_scrub |= scrub_cache_keys(&mut scrubbed)?;
    }
```

iker-barriocanal: Address this in [cfb8dbc](https://github.com/getsentry/relay/pull/2129/commits/cfb8dbcc4ecffd755b080b8bf484b4372af0cf3a) instead of this PR to avoid more merge conflicts.

PR #2128, opened 2023-05-16
feat(txnames): Add 'ready' flag to config

jjbayer:
Add a flag to project configs which indicates whether URL transaction names should always be considered `sanitized`, even if there are no clustering rules.

This PR does not yet _use_ the flag, it is just a prerequisite for merging https://github.com/getsentry/sentry/pull/48993.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Internal**:
+
+- Add `txNameReady` flag to project config. ([#2128](https://github.com/getsentry/relay/pull/2128))
+
 ## 23.5.0
 
 **Bug Fixes**:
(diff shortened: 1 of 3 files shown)


relay-dynamic-config/src/project.rs
```diff
@@ -123,4 +127,11 @@ pub struct LimitedProjectConfig {
     pub features: BTreeSet<Feature>,
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub tx_name_rules: Vec<TransactionNameRule>,
+    /// Whether or not a project is ready to mark all URL transactions as "sanitized".
+    #[serde(skip_serializing_if = "is_false")]
```
iker-barriocanal, line 131: nit: we can avoid creating the `is_false` function with [`Not::not`](https://doc.rust-lang.org/std/ops/trait.Not.html#tymethod.not), but I don't know how easy to understand it'd be.

iker-barriocanal (approved):

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 - Parse profiles' metadata to check if it should be marked as invalid. ([#2104](https://github.com/getsentry/relay/pull/2104))
 - Set release as optional by defaulting to an empty string and add a dist field for profiles. ([#2098](https://github.com/getsentry/relay/pull/2098), [#2107](https://github.com/getsentry/relay/pull/2107))
 - Accept source map debug images in debug meta for Profiling. ([#2097](https://github.com/getsentry/relay/pull/2097))
+- Add `txNameReady` flag to project config. ([#2128](https://github.com/getsentry/relay/pull/2128))
```
olksdr: Please, move this to the unreleased section,

olksdr (approved):

relay-dynamic-config/src/project.rs
jjbayer: TIL... but I'll leave it as-is. As you said, `is_false` seems easier to read than `Not::not`.

CHANGELOG.md
jjbayer: Thanks, I missed that!

PR #2133, opened 2023-05-17
feat(spool): Spool incoming envelopes proactively once the limit is reached

olksdr:
We should spool envelopes proactively if there is increase in the ingress traffic, and we cannot process them as fast anymore.

Even if all the caches are fresh, we still try to buffer all the incoming traffic if we over the 80 % of buffer guard usage.

fix: https://github.com/getsentry/team-ingest/issues/128

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_cache.rs
+++ b/relay-server/src/actors/project_cache.rs
@@ -745,7 +745,7 @@ impl ProjectCacheBroker {
         let key = QueueKey::new(own_key, sampling_key.unwrap_or(own_key));
 
         // If the more than 80% of capacity is exhausted, we immediately spool.
-        if self.buffer_guard.used() as f64 / self.buffer_guard.capacity() as f64 >= 0.8 {
+        if self.buffer_guard.is_over_high_watermark() {
             self.enqueue(key, context);
             return;
         }
--- a/relay-server/src/actors/project_cache.rs
+++ b/relay-server/src/actors/project_cache.rs
@@ -742,20 +742,16 @@ impl ProjectCacheBroker {
                 .filter(|st| !st.invalid())
         });
 
-        let key = QueueKey::new(own_key, sampling_key.unwrap_or(own_key));
-
-        // If the more than 80% of capacity is exhausted, we immediately spool.
-        if self.buffer_guard.is_over_high_watermark() {
-            self.enqueue(key, context);
-            return;
-        }
-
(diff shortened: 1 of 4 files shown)


relay-server/src/actors/project_cache.rs
```diff
@@ -741,13 +742,20 @@ impl ProjectCacheBroker {
                 .filter(|st| !st.invalid())
         });
 
+        let key = QueueKey::new(own_key, sampling_key.unwrap_or(own_key));
+
+        // If the more than 80% of capacity is exhausted, we immediately spool.
+        if self.buffer_guard.is_over_high_watermark() {
+            self.enqueue(key, context);
+            return;
+        }
```
jjbayer: nit: I would unite this check with the one below such that `enqueue` is only called on one line, i.e.

```rust
if ... && !self.buffer_guard.is_over_high_watermark() {
    return self.handle_processing(context);
}
self.enqueue(key, context);
```

relay-server/src/utils/buffer.rs
```diff
@@ -30,13 +30,33 @@ impl std::error::Error for BufferError {}
...
+            high_watermark: 0.8,
+            low_watermark: 0.5,
+        }
+    }
+
+    /// Returns `true` if the `BufferGuard` exhausted more permits then defined in
+    /// `high_watermark`.
+    #[inline]
+    pub fn is_over_high_watermark(&self) -> bool {
+        self.used() as f64 / self.capacity() as f64 >= self.high_watermark
```
jjbayer: nit: I would pu `self.used() as f64 / self.capacity() as f64` into a `usage()` function.

jjbayer (approved):

relay-server/src/actors/project_cache.rs
olksdr: done [b58dcb1](https://github.com/getsentry/relay/pull/2133/commits/b58dcb1176e9628a43b37da51db0e38e50d90721)

relay-server/src/utils/buffer.rs
olksdr: done [b58dcb1](https://github.com/getsentry/relay/pull/2133/commits/b58dcb1176e9628a43b37da51db0e38e50d90721)

jjbayer (approved):

PR #2134, opened 2023-05-17
ref: Replace log with tracing

jan-auer:
An internal refactor that uses `tracing` for all internal logging instead of the `log` crate. Since many dependencies still emit `log` records, we keep the subscriber's default integration which converts `log` records into tracing events. The only two dependencies that are now emitting tracing directly are `tower-http` and `rdkafka`, although the latter one doesn't add structural data.

The formatting of log output has changed slightly:
image

Note that this PR does not yet change emission of log records. In particular, errors are still tracked as regular message events. This is left for an immediate follow-up PR.

# Tasks

- [x] Initialize tracing subscribers
- [x] Export tracing types from `relay-log`
- [x] Initialize tracing for tests
- [x] Make the pretty log format more compact
- [x] Check fields of the JSON log format
- [x] Enable tracing in hyper/axum/axum-server
- [x] Check other crates for tracing support (e.g. `symbolic`)
- [x] Enable log integration for old crates
- [x] Check error reporting via Sentry

#skip-changelog


cleptric (approved):

relay-log/src/setup.rs
```diff
@@ -127,20 +166,22 @@ fn capture_native_envelope(data: &[u8]) {
...
-    builder
-        // Configure INFO as default for all third-party crates.
-        .filter_level(LevelFilter::Info)
-        // Actix-web has useful information on the debug stream, so allow this.
-        .filter_module("actix_web::pipeline", LevelFilter::Debug)
-        // Logs from sqlx are very spammy on INFO level, so configure a higher WARN level.
-        .filter_module("sqlx", LevelFilter::Warn)
-        .filter_module("trust_dns_proto", LevelFilter::Warn);
+fn get_default_filters() -> EnvFilter {
+    // Configure INFO as default, expect for crates that are very spammy on INFO level.
```
olksdr:
```suggestion
    // Configure INFO as default, except for crates that are very spammy on INFO level.
```

olksdr (approved):
:shipit: LGTM!

PR #2135, opened 2023-05-17
ref(crons): Remove SDK field from cron body

evanpurkhiser:
After a bit of discussion we think it doesn't make sense for this to be part of the body, since this information is already available in the envelope headers

This is a follow up from 753995836


evanpurkhiser:
Closing in favor of #2136

PR #2136, opened 2023-05-18
ref(crons): Remove SDK from check-in body + extract from header

evanpurkhiser:
After a bit of discussion we think it doesn't make sense for this to be
part of the body, since this information is already available in the
envelope headers.

Instead we pass the client SDK name in the kafka message (not the
payload).

This is a follow up from 753995836

#skip-changelog


jjbayer (approved):
Seems more consistent with what we do for sessions.

PR #2138, opened 2023-05-19
feat(spans): Apply transaction name rules to span descriptions

iker-barriocanal:
Applies transaction name rules to span descriptions with a URL-like pattern. Notes:

- We say a span description has a URL-like format when a span's operation starts with `http`.
- Transaction name rules are created on the assumption of a specific transaction name format. Span descriptions have a different format: there's an HTTP verb as a prefix of the string. Thus, I decided to adapt the rule (vs removing the prefix, for the edge cases) and add a leading `**` to match any leading substring (i.e. not only HTTP verbs). Since we're only applying this to URL-like transactions, we don't have to worry about other types.
- We'll consider in the future applying rules to resource-like span descriptions.

#skip-changelog

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -41,11 +41,11 @@ impl<'r> TransactionsProcessor<'r> {
         let mut span_desc_rules = None;
         if scrub_span_descriptions && !name_config.rules.is_empty() {
             span_desc_rules = Some(
-                (*(name_config.rules))
+                name_config
+                    .rules
                     .iter()
-                    .clone()
-                    .map(|tx_rule| SpanDescriptionRule::from(tx_rule))
-                    .collect::<Vec<SpanDescriptionRule>>(),
+                    .map(SpanDescriptionRule::from)
+                    .collect::<Vec<_>>(),
             );
         }
 
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -2832,6 +2832,19 @@ mod tests {
                         "status": "ok"
                     },
                     {
+                        "description": "POST http://example.com/remains/to-scrub/remains-too/1234567890",
+                        "op": "http.client",
+                        "parent_span_id": "8f5a2b8768cafb4e",
+                        "span_id": "bd2eb23da2beb450",
+                        "start_timestamp": 1597976393.4619668,
+                        "timestamp": 1597976393.4718769,
+                        "trace_id": "ff62a8b040f340bda5d830223def1d81",
+                        "status": "ok",
+                        "data": {
+                            "description.scrubbed": "POST http://example.com/remains/to-scrub/remains-too/*"
+                        }
+                    },
+                    {
                         "description": "GET http://example.com/another/url/is/intact",
                         "op": "http.client",
                         "parent_span_id": "8f5a2b8768cafb4e",
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -109,9 +109,7 @@ impl<'r> TransactionsProcessor<'r> {
         let previously_scrubbed =
             matches!(span.data.value(), Some(data) if data.get("description.scrubbed").is_some());
         if !previously_scrubbed {
-            dbg!(&span.description.clone().value());
             if let Some(description) = span.description.clone().value() {
-                dbg!(&description);
                 span.data
                     .value_mut()
                     .get_or_insert_with(BTreeMap::new)
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -32,21 +32,19 @@ pub struct TransactionNameConfig<'r> {
 #[derive(Default)]
 pub struct TransactionsProcessor<'r> {
     name_config: TransactionNameConfig<'r>,
-    span_desc_rules: Option<Vec<SpanDescriptionRule>>,
+    span_desc_rules: Vec<SpanDescriptionRule>,
     scrub_span_descriptions: bool,
 }
 
 impl<'r> TransactionsProcessor<'r> {
     pub fn new(name_config: TransactionNameConfig<'r>, scrub_span_descriptions: bool) -> Self {
-        let mut span_desc_rules = None;
+        let mut span_desc_rules = Vec::new();
         if scrub_span_descriptions && !name_config.rules.is_empty() {
-            span_desc_rules = Some(
-                name_config
-                    .rules
-                    .iter()
-                    .map(SpanDescriptionRule::from)
-                    .collect::<Vec<_>>(),
-            );
+            span_desc_rules = name_config
+                .rules
+                .iter()
+                .map(SpanDescriptionRule::from)
+                .collect::<Vec<_>>();
         }
 
         Self {
@@ -113,9 +111,9 @@ impl<'r> TransactionsProcessor<'r> {
             }
         }
 
-        let Some(rules) = &self.span_desc_rules else {
+        if self.span_desc_rules.is_empty() {
             return Ok(());
-        };
+        }
 
         // HACK(iker): work-around to scrub the description, in a
         // context-manager-like approach.
@@ -148,7 +146,7 @@ impl<'r> TransactionsProcessor<'r> {
             if let Some(description) = data.get_mut("description.scrubbed") {
                 description.apply(|name, meta| {
                     if let Value::String(s) = name {
-                        let result = rules.iter().find_map(|rule| {
+                        let result = self.span_desc_rules.iter().find_map(|rule| {
                             rule.match_and_apply(Cow::Borrowed(s))
                                 .map(|new_name| (rule.pattern.compiled().pattern(), new_name))
                         });
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -71,7 +71,7 @@ impl<'r> TransactionsProcessor<'r> {
         if let Some(info) = info.as_mut() {
             transaction.apply(|transaction, meta| {
                 let result = self.name_config.rules.iter().find_map(|rule| {
-                    rule.transaction_match_and_apply(Cow::Borrowed(transaction), info)
+                    rule.match_and_apply(Cow::Borrowed(transaction), info)
                         .map(|applied_result| (rule.pattern.compiled().pattern(), applied_result))
                 });
 
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -14,6 +14,7 @@ use crate::store::regexes::{
     CACHE_NORMALIZER_REGEX, RESOURCE_NORMALIZER_REGEX, SQL_NORMALIZER_REGEX,
     TRANSACTION_NAME_NORMALIZER_REGEX,
 };
+use crate::store::LazyGlob;
 use crate::types::{
     Annotated, Meta, ProcessingAction, ProcessingResult, Remark, RemarkType, Value,
 };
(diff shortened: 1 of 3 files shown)


relay-general/src/store/transactions/snapshots/relay_general__store__transactions__processor__tests__scrub_span_identifiers_and_apply_rules.snap
```diff
@@ -0,0 +1,75 @@
...
+    {
+      "timestamp": 1597976393.471877,
+      "start_timestamp": 1597976393.461967,
+      "description": "GET http://example.com/another/url/is/intact",
+      "op": "http.client",
+      "span_id": "bd2eb23da2beb451",
+      "parent_span_id": "8f5a2b8768cafb4e",
+      "trace_id": "ff62a8b040f340bda5d830223def1d81",
+      "status": "ok",
+      "data": {}
```
iker-barriocanal, line 57: I'll get rid of this in a future PR.

relay-general/src/store/transactions/rules.rs
```diff
@@ -111,7 +111,7 @@ pub struct TransactionNameRule {
 
 impl TransactionNameRule {
     /// Checks is the current rule matches and tries to apply it.
-    pub fn match_and_apply(
+    pub fn transaction_match_and_apply(
```
iker-barriocanal: I don't really like this (especially the code duplication), but it's the fastest way I could think of to make the `TransactionNameRule` generic enough for this PR without spending time in a refactor to make a clean interface.

relay-general/src/store/transactions/rules.rs
```diff
@@ -149,16 +167,21 @@ impl TransactionNameRule {
         }
     }
 
-    /// Returns `true` if the current rule pattern matches transaction, expected transaction
-    /// source, and not expired yet.
-    pub fn matches(&self, transaction: &str, info: &TransactionInfo) -> bool {
-        let now = Utc::now();
+    /// Returns `true` if the transaction info has a matching transaction source and
+    /// the transaction name matches the rule (see
+    /// [`TransactionNameRule::matches`]).
```
iker-barriocanal: I belive this link doesn't really work, but I spent too much time already and didn't manage to point to that method properly. Any ideas?

relay-general/src/store/transactions/snapshots/relay_general__store__transactions__processor__tests__scrub_span_identifiers_and_apply_rules.snap
```diff
@@ -0,0 +1,75 @@
...
+  ],
+  "_meta": {
+    "spans": {
+      "0": {
+        "data": {
+          "description.scrubbed": {
+            "": {
+              "rem": [
+                [
+                  "description.scrubbed:/remains/*/**",
```
iker-barriocanal: This doesn't really identify what span it applied to. Happy to update it, but I don't think this is really important right now.

relay-general/src/store/transactions/processor.rs
```diff
@@ -84,6 +85,94 @@ impl<'r> TransactionsProcessor<'r> {
...
+        // HACK(iker): work-around to scrub the description, in a
+        // context-manager-like approach.
+        //
+        // If data[description.scrubbed] isn't present, we want to scrub
+        // span.description. However, they have different types:
+        // Annotated<Value> vs Annotated<String>. The simplest and fastest
+        // solution I found is to add span.description to span.data if it
+        // doesn't exist already, scrub it, and remove it if we did nothing.
+        let previously_scrubbed =
+            matches!(span.data.value(), Some(data) if data.get("description.scrubbed").is_some());
```
```suggestion
        let previously_scrubbed = span.data.value().map(|d| d.get("description.scrubbed")).is_some();
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -84,6 +85,94 @@ impl<'r> TransactionsProcessor<'r> {
...
+        if !previously_scrubbed {
+            dbg!(&span.description.clone().value());
```
```suggestion
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -84,6 +85,94 @@ impl<'r> TransactionsProcessor<'r> {
...
+            if let Some(description) = span.description.clone().value() {
+                dbg!(&description);
```
```suggestion
```

relay-general/src/store/transactions/rules.rs
jjbayer: Does `Self` work?

relay-general/src/store/transactions/snapshots/relay_general__store__transactions__processor__tests__scrub_span_identifiers_and_apply_rules.snap
jjbayer: It has the `"0"` on line 59, doesn't that identify the span?

relay-general/src/store/transactions/rules.rs
jjbayer: You could get rid of code duplication by making `transaction_info` an optional argument to `match_and_apply` (and only checking the transaction source if it's `Some`).

relay-general/src/store/transactions/processor.rs
```diff
@@ -84,6 +85,94 @@ impl<'r> TransactionsProcessor<'r> {
...
+
+/// Applies the transaction rule to the string, no matter what the string format is.
+///
+/// Transaction rules are generated on the assumption that they are only
+/// applied to transactions, which have a specific format. Span
+/// descriptions, however, have a different format, including the start of
+/// the string. As a workaround, we add a leading `**` to omit that first
+/// part of the string.
+fn apply_tx_rule_to_string(string: Cow<String>, mut rule: TransactionNameRule) -> Option<String> {
+    rule.pattern = LazyGlob::new(format!("**{}", rule.pattern.raw));
```
jjbayer: This compiles the rule for every span, correct? Can we instead put another field `span_rules` on `name_config` or `self` which contains a separate set of `LazyGlob`s for spans?

jjbayer: Alternative: Don't use different rules, just do a `split(" ")` on the description, or even check for case insensitive `(get|put|post|...) ` explicitly. Similar to how we add-and-remove a trailing slash before scrubbing, we can normalize the span description before applying the rule.

jjbayer (changes requested):
Just a partial review, did not review everything yet.

relay-general/src/store/transactions/snapshots/relay_general__store__transactions__processor__tests__scrub_span_identifiers_and_apply_rules.snap
```diff
@@ -0,0 +1,75 @@
...
+  ],
+  "_meta": {
+    "spans": {
+      "0": {
+        "data": {
+          "description.scrubbed": {
+            "": {
+              "rem": [
+                [
+                  "description.scrubbed:**/remains/*/**",
```
iker-barriocanal, line 78: Note the leading `**` here. This is fine for now, once we have proper span description rules these will disappear.

relay-general/src/store/transactions/rules.rs
```diff
@@ -94,6 +94,70 @@ impl Default for RedactionRule {
...
+            pattern: LazyGlob::new(format!("**{}", value.pattern.raw)),
+            expiry: value.expiry,
+            scope: value.scope.clone(),
+            redaction: value.redaction.clone(),
+        }
+    }
+}
+
+impl SpanDescriptionRule {
+    pub fn match_and_apply(&self, mut string: Cow<String>) -> Option<String> {
```
iker-barriocanal, line 122: I think we could benefit from having a trait for these rules. I'll look into this at some point.

relay-general/src/store/transactions/processor.rs
```diff
@@ -29,13 +30,26 @@ pub struct TransactionNameConfig<'r> {
...
 pub struct TransactionsProcessor<'r> {
     name_config: TransactionNameConfig<'r>,
+    span_desc_rules: Option<Vec<SpanDescriptionRule>>,
     scrub_span_descriptions: bool,
 }
 
 impl<'r> TransactionsProcessor<'r> {
     pub fn new(name_config: TransactionNameConfig<'r>, scrub_span_descriptions: bool) -> Self {
+        let mut span_desc_rules = None;
+        if scrub_span_descriptions && !name_config.rules.is_empty() {
```
iker-barriocanal: I tried something like:
```rust
let span_desc_rules = if ... {
  // generate rules
} else {
  None
}
```
But could not deal with lifetimes. I think this is fine for now.

iker-barriocanal (commented):
Note that the logic around this will change a lot once we have proper span description rules in place.

relay-general/src/store/transactions/processor.rs
iker-barriocanal: I opted for the first version, since the second one would require more logic to also get rid of the subdomain properly.

relay-general/src/store/transactions/rules.rs
iker-barriocanal: I'll refactor this at some point to extract common functionality into a trait, and both the API and logic should get much easier to read and deal with.

relay-general/src/store/transactions/processor.rs
```diff
@@ -31,13 +32,26 @@ pub struct TransactionNameConfig<'r> {
...
+            span_desc_rules = Some(
+                (*(name_config.rules))
+                    .iter()
+                    .clone()
+                    .map(|tx_rule| SpanDescriptionRule::from(tx_rule))
+                    .collect::<Vec<SpanDescriptionRule>>(),
+            );
```
```suggestion
            span_desc_rules = Some(
                name_config
                    .rules
                    .iter()
                    .map(SpanDescriptionRule::from)
                    .collect::<Vec<_>>(),
            );
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -31,13 +32,26 @@ pub struct TransactionNameConfig<'r> {
 #[derive(Default)]
 pub struct TransactionsProcessor<'r> {
     name_config: TransactionNameConfig<'r>,
+    span_desc_rules: Option<Vec<SpanDescriptionRule>>,
```
jjbayer: What's the benefit of making this an option vs. just using an empty `Vec` to disable?

relay-general/src/store/transactions/processor.rs
```diff
@@ -2705,4 +2801,82 @@ mod tests {
...
+                    {
+                        "description": "POST http://example.com/remains/not-scrubbed-for-different-op/remains-too",
+                        "op": "db.sql.query",
+                        "parent_span_id": "8f5a2b8768cafb4e",
+                        "span_id": "bd2eb23da2beb452",
+                        "start_timestamp": 1597976393.4619668,
+                        "timestamp": 1597976393.4718769,
+                        "trace_id": "ff62a8b040f340bda5d830223def1d81",
+                        "status": "ok"
+                    }
```
jjbayer, line 2866: Should we add a test span here that already has a `description.scrubbed`? If I read the code correctly, the intention is to scrub that one again.

relay-general/src/store/transactions/rules.rs
```diff
@@ -12,7 +12,7 @@ use crate::utils::Glob;
 /// This allows to compile the Glob with internal regexes only then whent they are used.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct LazyGlob {
-    raw: String,
+    pub raw: String,
```
jjbayer: Why does `.raw` have to be pub?

relay-general/src/store/transactions/rules.rs
```diff
@@ -94,6 +94,70 @@ impl Default for RedactionRule {
...
+    /// Object containing transaction attributes the rules must only be applied to.
+    pub scope: RuleScope,
+    /// Object describing what to do with the matched pattern.
+    pub redaction: RedactionRule,
+}
+
+impl From<&TransactionNameRule> for SpanDescriptionRule {
+    fn from(value: &TransactionNameRule) -> Self {
+        SpanDescriptionRule {
+            pattern: LazyGlob::new(format!("**{}", value.pattern.raw)),
```
jjbayer, line 113: Let's see if `**` produces undesired cases because it is so permissive. Since the effects are contained in `description.scrubbed`, there is no risk in being aggressive.

jjbayer (approved):

PR #2139, opened 2023-05-22
feat(txnames): Use txNameReady flag

jjbayer:
Start using the flag introduced by https://github.com/getsentry/sentry/pull/48993 and https://github.com/getsentry/relay/pull/2128.

Once the clusterer has run ~10 times, trust that all the rules than can be discovered have been discovered, and mark all incoming URL transactions as sanitized.


olksdr (approved):
lgtm

iker-barriocanal (approved):

PR #2140, opened 2023-05-22
feat(spans): Add span metrics to span.data

iker-barriocanal:
Adds span metrics to `span.data`, even if metrics are extracted. This is useful to have metrics available when querying spans.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -8,7 +8,7 @@ use relay_general::protocol::{
     AsPair, Context, ContextInner, Event, TraceContext, TransactionSource, User,
 };
 use relay_general::store;
-use relay_general::types::Annotated;
+use relay_general::types::{Annotated, Value};
 use relay_metrics::{AggregatorConfig, Metric, MetricNamespace, MetricValue};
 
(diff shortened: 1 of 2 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -571,6 +571,14 @@ fn extract_span_metrics(
                 span_tags.insert("span.status_code".to_owned(), status_code.to_owned());
             }
 
+            // Even if we emit metrics, we want this info to be duplicated in every span.
+            for (key, value) in &span_tags {
+                span.data.get_or_insert_with(BTreeMap::new).insert(
+                    key.to_owned(),
+                    Annotated::from(Value::String(value.to_owned())),
+                );
+            }
```
olksdr: What do you think if we overwrite this a bit differently? Which I find a little bit more nicer, but it's up-to-you if you want to change. 

```suggestion
                        // Even if we emit metrics, we want this info to be duplicated in every span.
            span.data.get_or_insert_with(BTreeMap::new).extend(
                span_tags
                    .clone()
                    .into_iter()
                    .map(|(k, v)| (k, Annotated::from(Value::String(v)))),
            );

```

olksdr (approved):

PR #2141, opened 2023-05-22
Add `data` and `api_target` fields to `ResponseContext` and scrub `graphql` bodies.

marandaneto:
PR for https://github.com/getsentry/rfcs/blob/choregraphql-errors/text/0087-graphql-errors.md

* This PR adds a few fields to the event protocol.
  * `Request#api_target` which is the API target used for advanced data scrubbing, a marker of the type of the request.
  `Response#data` is similar to `Request#data`.
  `Response#inferred_content_type` is similar to `Request#inferred_content_type`.
* This PR implements the `process_event` for the PII processor scrubbing variables from a GraphQL error based on the `Request#api_target`.


jjbayer:
> * Do I need to `normalize` similarly to [request](https://github.com/getsentry/relay/blob/master/relay-general/src/store/normalize/request.rs)?
 
 Only if we allow SDKs to send multiple formats that have to be normalized into one. When in doubt, I would say no.
 
> * Do I need to `process_response` similarly to [request](https://github.com/getsentry/relay/blob/ca22143e815522b351da353187ea6dea522f25aa/relay-general/src/store/normalize.rs#L901C8-L912)? I believe so if the one above is also a yes.

This would be the function to hook into. It already takes care of normalizing cookies: https://github.com/getsentry/relay/blob/5d764fc1d7d877586193ba49fe56f278c2346461/relay-general/src/store/normalize/contexts.rs#L180-L200

> * Where `response.data` data scrubbing would go? Probably `PiiProcessor`?

Default scrubbing runs automatically because of `pii=true`. Any custom scrubbing would have to go into `PiiProcessor`, yes.

> * Is the new field auto-generated in `test_fixtures__event_schema.snap` or should I add it manually?

Use `cargo insta review` to review the changed snapshots:

```shell
cargo install cargo-insta
cargo insta test
cargo insta review
```

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,78 @@ impl<'a> Processor for PiiProcessor<'a> {
...
     }
+
+    fn process_event(
+        &mut self,
+        event: &mut Event,
+        _meta: &mut Meta,
+        state: &ProcessingState<'_>,
+    ) -> ProcessingResult {
+        // TODO: make keys lazy init
+        let mut keys = Vec::new();
```
marandaneto: Would make more sense to make this init lazy only if `graph_ql_error` is true?

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,78 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+
+        // collect the variables keys and scrub them out
+        if let Some(request) = event.request.value_mut() {
+            if let Some(Value::Object(data)) = request.data.value_mut() {
+                if let Some(Annotated(Some(Value::Bool(graph_ql_error)), _)) =
+                    request.other.get("graphql_error")
```
marandaneto: Still have to decided where to place the marker `graphql_error`

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,78 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+        Ok(())
+    }
+}
+
+fn scrub_data_from_response(keys: Vec<String>, data: &mut BTreeMap<String, Annotated<Value>>) {
+    for item in data.iter_mut() {
+        match item.1.value_mut() {
+            Some(Value::Object(item_data)) => {
+                // TODO: can we avoid cloning every time?
+                scrub_data_from_response(keys.clone(), item_data);
```
marandaneto: Is cloning the right thing to do?

relay-general/src/store/normalize/contexts.rs
```diff
@@ -177,7 +177,82 @@ fn normalize_os_context(os: &mut OsContext) {
     }
 }
 
+// TODO: extract urlencoded_from_str, parse_raw_response_data and normalize_response_data
+// can probably be reused in other places (original from normalize/request)
```
marandaneto: Where would make sense to put this?
This is now being used by `Request` and `Response`.

relay-general/tests/snapshots/test_fixtures__event_schema.snap
```diff
@@ -1,5 +1,6 @@
 ---
 source: relay-general/tests/test_fixtures.rs
+assertion_line: 109
```
marandaneto: This was added automatically?

relay-general/src/pii/processor.rs
jjbayer: When we do, please make sure it's typed, so we don't have to read from `request.other` (which is only for forwarding data).

relay-general/src/pii/processor.rs
jjbayer: Vectors are lazy by design (see [docs](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new)):

> The vector will not allocate until elements are pushed onto it.

But I think it's good practice to only initialize the key on line 222. You could move the entire "response" block in there, because there is no need to run that when `keys` is empty.

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,78 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                            if let Some(Annotated(Some(Value::Object(response_data)), _)) =
+                                data.get_mut("data")
+                            {
+                                scrub_data_from_response(keys.clone(), response_data);
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
jjbayer, line 263: To make it easier to spot what this code is about, I would put this entire block in a helper function called `scrub_graphql` or similar.

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,78 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+
+        // scrub PII from the data object if they match the variables keys
+        if let Some(contexts) = event.contexts.value_mut() {
+            if let Some(Context::Response(response)) = contexts.get_context_mut("response") {
+                if let Some(Value::Object(data)) = response.data.value_mut() {
+                    if let Some(Annotated(Some(Value::Bool(graph_ql_error)), _)) =
```
marandaneto: Probably just the `graphql_error` from the request is enough.

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,78 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+            }
+        }
+
+        event.process_child_values(self, state)?;
+
+        Ok(())
+    }
+}
+
+fn scrub_data_from_response(keys: Vec<String>, data: &mut BTreeMap<String, Annotated<Value>>) {
```
jjbayer: It should be possible to pass a `BTreeSet<&str>` instead of a `Vec<String>` here to avoid cloning.
```suggestion
fn scrub_data_from_response<'a>(keys: BTreeSet<&'a str>, data: &mut BTreeMap<String, Annotated<Value>>) {
```

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,78 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                    request.other.get("graphql_error")
+                {
+                    dbg!("graphql_error request {graphql_error}");
+                    if *graph_ql_error {
+                        if let Some(Annotated(Some(Value::Object(variables)), _)) =
+                            data.get_mut("variables")
+                        {
+                            for item in variables.iter_mut() {
+                                keys.push(item.0.to_string());
+                                item.1
```
jjbayer:
```suggestion
                            for(key, value) in variables.iter_mut() {
                                keys.push(key.to_string());
                                value
```

relay-general/src/pii/processor.rs
jjbayer: See comment above.

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,78 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                scrub_data_from_response(keys.clone(), item_data);
+            }
+            _ => {
+                if keys.contains(item.0) {
+                    item.1
+                        .set_value(Some(Value::String("[Filtered]".to_string())));
+                }
+            }
+        }
+    }
```
jjbayer, line 280: Do we want these changes to be conditional on user configuration? If so, we'll have to think of how to pass the configuration through:

image

relay-general/src/pii/processor.rs
marandaneto: If they are lazy it is alright, otherwise `keys` isn't initiated yet in L222, if I declare it in L222, then it's not accessible in the other scope.
In Kotlin you can do `keys.isInitialized` before accessing to be sure, is that possible in Rust? Probably using `Option` instead.

marandaneto: I used an `Option<BTreeSet<String>>` inited with `None`, let me know if there's a better way to avoid the `Option` or `BTreeSet` init overhead.

relay-general/src/pii/processor.rs
marandaneto: Since we are guarding against `graphql_error`, I believe not for now, we can add it later if needed.

relay-general/src/pii/processor.rs
marandaneto: Added `request.api_target` typed.

marandaneto:
@jjbayer this PR should be good to go, I've updated the PR description.
I'll test these changes locally with Client SDKs, so it's fine if we don't merge it right away.

relay-general/src/store/normalize/contexts.rs
```diff
@@ -177,7 +178,42 @@ fn normalize_os_context(os: &mut OsContext) {
...
+fn parse_raw_response_data(response: &mut ResponseContext) -> Option<(&'static str, Value)> {
+    let raw = response.data.as_str()?;
+
+    // TODO: Try to decode base64 first
+
+    if let Ok(value) = serde_json::from_str(raw) {
+        Some(("application/json", value))
+    } else {
+        url_encoding::encoded_from_str(raw)
+            .map(|value| ("application/x-www-form-urlencoded", value))
```
jjbayer: This content type makes sense as a fallback for requests, but it seems wrong for responses. Do HTTP responses actually contain urlencoded form data?

relay-general/src/protocol/request.rs
```diff
@@ -497,6 +497,15 @@ pub struct Request {
     #[metastructure(skip_serialization = "empty")]
     pub inferred_content_type: Annotated<String>,
 
+    /// The API target/specification that made the request.
+    ///
+    /// Values can be `graphql`, `rest`, etc.
+    ///
+    /// The data field should contain the request and response bodies based on its target specification.
+    ///
+    /// This information can be used to better data scrubbing and normalization.
```
jjbayer:
```suggestion
    /// This information can be used for better data scrubbing and normalization.
```

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,95 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+fn scrub_graphql(event: &mut Event) {
+    let mut keys: Option<BTreeSet<String>> = None;
+
+    let mut is_graphql = false;
+
+    // collect the variables keys and scrub them out.
+    if let Some(request) = event.request.value_mut() {
+        if let Some(Value::Object(data)) = request.data.value_mut() {
+            if let Some(api_target) = request.api_target.value() {
+                if api_target.to_lowercase().eq("graphql") {
```
jjbayer:
```suggestion
                if api_target.eq_ignore_ascii_case("graphql") {
```

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,95 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+
+        event.process_child_values(self, state)?;
+
+        Ok(())
+    }
+}
+
+/// Scrubs GraphQL variables from the event.
+fn scrub_graphql(event: &mut Event) {
+    let mut keys: Option<BTreeSet<String>> = None;
```
jjbayer: 1. Please use `&str` instead of `String`.
2. Why wrap this in an `Option`?

```suggestion
    let mut keys: Option<BTreeSet<&str>> = None;
```

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,95 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                if api_target.to_lowercase().eq("graphql") {
+                    is_graphql = true;
+                }
+            }
+
+            if is_graphql {
+                if let Some(Annotated(Some(Value::Object(variables)), _)) =
+                    data.get_mut("variables")
+                {
+                    let mut current_keys: BTreeSet<String> = BTreeSet::new();
```
jjbayer: If you make `keys` non-optional, you can insert straight into `keys` here, right?

jjbayer (changes requested):

relay-general/src/pii/processor.rs
marandaneto: 2. if you init the `BTreeSet` every time `process_event` gets called, you pay the overhead of initing a `BTreeSet` every time even if you don't need it.
The equivalent of a nullable variable in Rust is `Option` unless I'm missing something.

relay-general/src/pii/processor.rs
marandaneto: yes, it was like that before, but I wanted to optimize because of https://github.com/getsentry/relay/pull/2141#discussion_r1212903223

relay-general/src/store/normalize/contexts.rs
marandaneto: Based on a quick search, https://stackoverflow.com/questions/4007969/application-x-www-form-urlencoded-or-multipart-form-data but the response was still JSON.
My question is then, what would be the fallback if any? the original `response.data`?

relay-general/src/pii/processor.rs
marandaneto: https://doc.rust-lang.org/stable/std/collections/struct.BTreeSet.html#method.new

> Does not allocate anything on its own.

Removing the `Option`.

relay-general/src/store/normalize/contexts.rs
jjbayer: > My question is then, what would be the fallback if any? the original response.data?

Yes, I would just return `None` here if JSON parsing fails, and leave the `data` field untouched (same as what would happen now if parsing of form data fails).

jjbayer (approved):

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,85 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                }
+            }
+        }
+    }
+
+    // scrub PII from the data object if they match the variables keys.
+    if let Some(contexts) = event.contexts.value_mut() {
+        if let Some(Context::Response(response)) = contexts.get_context_mut("response") {
+            if let Some(Value::Object(data)) = response.data.value_mut() {
+                if is_graphql {
```
olksdr: Can we move this out to, maybe, line `243` and have an early return if the `is_graphql` is `false`.  This way you will have fewer nested functions and will communicate clearly that if it's not `graphql` we do not care

relay-general/src/protocol/contexts/response.rs
```diff
@@ -26,6 +29,17 @@ pub struct ResponseContext {
     /// HTTP response body size.
     pub body_size: Annotated<u64>,
 
+    /// Response data in any format that makes sense.
+    ///
+    /// SDKs should discard large and binary bodies by default. Can be given as string or
```
olksdr:
```suggestion
    /// SDKs should discard large and binary bodies by default. Can be given as a string or
```

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,85 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+    }
+}
+
+/// Scrubs GraphQL variables from the event.
+fn scrub_graphql(event: &mut Event) {
+    let mut keys: BTreeSet<&str> = BTreeSet::new();
+
+    let mut is_graphql = false;
+
+    // collect the variables keys and scrub them out.
```
olksdr:
```suggestion
    // Collect the variables keys and scrub them out.
```

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,85 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                    for (key, value) in variables.iter_mut() {
+                        keys.insert(key);
+                        value.set_value(Some(Value::String("[Filtered]".to_string())));
+                    }
+                }
+            }
+        }
+    }
+
+    // scrub PII from the data object if they match the variables keys.
```
olksdr:
```suggestion
    // Scrub PII from the data object if they match the variables keys.
```

relay-general/src/pii/processor.rs
```diff
@@ -195,6 +198,85 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+        if let Some(Context::Response(response)) = contexts.get_context_mut("response") {
+            if let Some(Value::Object(data)) = response.data.value_mut() {
+                if is_graphql {
+                    if let Some(Annotated(Some(Value::Object(graphql_data)), _)) =
+                        data.get_mut("data")
+                    {
+                        if !keys.is_empty() {
+                            scrub_graphql_data(&keys, graphql_data);
+                        } else {
+                            // if we don't have the variable keys, we scrub the whole data object
```
olksdr:
```suggestion
                            // If we don't have the variable keys, we scrub the whole data object
```

relay-general/src/store/normalize/contexts.rs
```diff
@@ -177,7 +177,39 @@ fn normalize_os_context(os: &mut OsContext) {
...
 }
 
+fn parse_raw_response_data(response: &mut ResponseContext) -> Option<(&'static str, Value)> {
+    let raw = response.data.as_str()?;
+
+    if let Ok(value) = serde_json::from_str(raw) {
+        Some(("application/json", value))
+    } else {
+        None
+    }
```
olksdr:
```suggestion

    serde_json::from_str(raw)
        .ok()
        .map(|value| ("application/json", value))

```

relay-general/src/pii/processor.rs
marandaneto: Good point, makes sense.

relay-general/src/store/normalize/contexts.rs
marandaneto: Otherwise, falls back to `None` automatically? Nice.

PR #2142, opened 2023-05-22
feat(spans): Use table name as domain in sql-like descriptions

iker-barriocanal:
The domain should represent the table name in SQL-like span descriptions. This PR adds this capability by extracting the table name with a RegEx. Note that this is a best effort and the RegEx may not cover all queries; this is ok for now, we'll improve in further iterations as needs evolve.

#skip-changelog


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -600,6 +611,41 @@ fn extract_span_metrics(
...
+
+/// only returns the first match
+fn sql_table_from_query(query: &str) -> Option<&str> {
+    let capture_names: Vec<_> = SQL_TABLE_EXTRACTOR_REGEX
+        .capture_names()
+        .flatten()
+        .collect();
+
+    for captures in SQL_TABLE_EXTRACTOR_REGEX.captures_iter(query) {
+        dbg!(&captures);
```
jjbayer:
```suggestion
        dbg!(&captures);
```

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -600,6 +611,41 @@ fn extract_span_metrics(
     Ok(())
 }
 
+static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r#"(?i)from(\s|\()+(?P<table>(\w+(\.\w+)*))(\s|\))+"#).unwrap());
```
jjbayer: Should we request whitespace or a newline before `from`? For example,

```rust
INSERT INTO from_date VALUES (1, 2, 3);
```

should not match the regex.

I would add a test for this to be robust against regressions.

jjbayer: Should we also consider other sql statements like INSERT? That is, `(from|into)`?

jjbayer: Does this match formatted multiline queries, for example

```sql
SELECT
    * 
FROM
    table_name
WHERE
    1
```

jjbayer (approved):
Approving to unblock, but please take a look at the comment about whitespace before `FROM`.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: Insert statement support added [here](https://github.com/getsentry/relay/pull/2142/commits/6c918bb9673ae78fae1298685bef2df3e7df0f2b). It does match multi-lines, added test [here](https://github.com/getsentry/relay/pull/2142/commits/6c918bb9673ae78fae1298685bef2df3e7df0f2b#diff-decea315cf714306f36169f6d773ba8a05f2f2fc841a7684da0c52968c9337c7R899) (I don't know how to format it properly so added `\n`s).

iker-barriocanal: It matches correctly, see [this test](https://github.com/getsentry/relay/pull/2142/commits/6c918bb9673ae78fae1298685bef2df3e7df0f2b#diff-decea315cf714306f36169f6d773ba8a05f2f2fc841a7684da0c52968c9337c7R885).

jjbayer (approved):

PR #2143, opened 2023-05-22
feat(replays): Enable pii scrubbing for urls field

### LINKED ISSUE #49495: Scrub sensitive query param values from Replay URLs
### Problem Statement

When Replays are captured and associated with a navigation event (by URL), we see in some cases that sensitive query param values (e.g., `jwt` or auth token) are displayed in full – either in a list view, ordered by URL, or on mouseover of some detail in a list of breadcrumbs.

In the sentry.client.config.js for our frontend (React - Next.js) app, we currently use the `beforeBreadcrumb` and `beforeSend` hooks to scrub sensitive values (including instances of a `jwt` param in URLs sent as part of a navigation or request breadcrumb, or in a stack trace). Is it possible to offer something like these hooks for Replays? Or some kind of config in the privacy & security settings that would address this? 

Please let me know if more info / context would be helpful; happy to supply what i can. Thanks very much!

----

Please see the attachments below for details:
 
1️⃣ **Entire `jwt` param visible on mouseover of URL in breadcrumb**
 
Screenshot 2023-05-11 at 10 15 20 PM (1)

2️⃣ **Example of `jwt` param obfuscated with `beforeBreadcrumb` or `beforeSend`**

Screen Shot 2023-05-11 at 5 15 41 PM

3️⃣ **Screen recording showing `jwt` revealed in full on mouseover**
https://github.com/getsentry/sentry/assets/3287153/3f405d17-51b5-495e-8c86-6191d80db9e3

### Solution Brainstorm

_No response_

### Product Area

Replays

cmanallen:
Fixes https://github.com/getsentry/sentry/issues/49495


iker-barriocanal (changes requested):
I think GitHub removed my previous comment 😕

Replays work a bit differently in Relay -- the payloads are quite big so we don't deserialize them into memory. To scrub a specific field, you should add the path to [this list](https://github.com/getsentry/relay/blob/b54a7dcae1b692fb170dc6b89bb77ca1df214e6d/relay-replays/src/recording.rs#L35) and it will automatically be scrubbed.

iker-barriocanal (approved):
I assumed this was for replay recordings, and not the metadata. LGTM!

PR #2144, opened 2023-05-22
build: Use sentry CLI v2

jjbayer:
We have recently encountered cases where the call to `sentry-cli` in the `create-sentry-release` script was killed by the OOM killer. Using a newer version of `sentry-cli` might help.

#skip-changelog


olksdr (approved):
:shipit:

PR #2145, opened 2023-05-22
feat: Add HTTP status code to the common tags on transactional metrics

olksdr:
This PR adds the HTTP status code extracted from the `Event` to the common tags on transactional metrics. 
The status code currently provided in different places by different SDKs and requires from us to make sure to cover most obvious places, like `breadcrumbs`, `span data`, `response context`, `span tags`.

Once the normalization on SDK side is done (tracking issue: https://github.com/getsentry/team-sdks/issues/9), the extraction on Relay side can be simplified as well. 
 
fix https://github.com/getsentry/team-ingest/issues/123

#skip-changelog


iker-barriocanal (approved):
Nice! Looking forward to agreeing on a single way on the SDK side 😄.

PR #2146, opened 2023-05-22
fix: Remove profiling feature flag

### LINKED ISSUE #2132: Profiles not recording when using Relay in proxy or static modes
### Environment

Laravel 8 PHP 7.4 using Relay in proxy/static mode.
 traces_sample_rate and profiles_sample_rate both set to 1.0

### Steps to Reproduce

1. Set up PHP profiling on existing project using sentry DSN, profiling, performance, issues were all working
2. Installed Relay in proxy mode
3. Issues and Performance continued to be recorded in sentry through Relay, Profiling stopped recording
4. Tried Relay in static mode same result
5. Tried disabling spike protection, same result

### Expected Result

Expected profiling to continue recording same as using the sentry DSN. Expected to see same number of profiles and transactions since profile sample rate set to 1.

### Actual Result

What actually happened. Maybe a screenshot/recording? Maybe some logs?

Issues and Performance continued to be recorded in sentry but Profiling stopped recording. Saw some profiles coming in but all were dropped because of spike protection even when spike protection disabled. Number of profiles was much lower than number of performance transactions

cleptric:
We forgot to remove this feature flag when profiling was GAed.

Closes #2132

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Bug Fixes**:
 
 - Sample only transaction events instead of sampling both transactions and errors. ([#2130](https://github.com/getsentry/relay/pull/2130))
+- Remove profiling feature flag. ([#2146](https://github.com/getsentry/relay/pull/2146))
 
 **Internal**:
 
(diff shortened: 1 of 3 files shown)


phacops (approved):

phacops:
This is important when running Relay as a proxy, since there's no project config, it filters all profiles at the proxy level. Since profiling is now GA, this can be safely removed.

olksdr (approved):

PR #2148, opened 2023-05-23
fix(dynamic-sampling): Fix missing sampling project state

iambriccardo:
This PR fixes a problem in which the sampling project state was not loaded for incoming items of type event.

### DIFF
--- a/relay-server/src/utils/dynamic_sampling.rs
+++ b/relay-server/src/utils/dynamic_sampling.rs
@@ -109,7 +109,7 @@ pub fn get_sampling_result(
 ///  - there is no [`DynamicSamplingContext`] in the envelope headers.
 ///  - there are no transactions or events in the envelope, since in this case sampling by trace is redundant.
 pub fn get_sampling_key(envelope: &Envelope) -> Option<ProjectKey> {
-    // If we the envelope item is not of type transaction or event, we will not return a sampling key
+    // If the envelope item is not of type transaction or event, we will not return a sampling key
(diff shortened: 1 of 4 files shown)


relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -107,13 +107,13 @@ pub fn get_sampling_result(
...
 ///  - there is no [`DynamicSamplingContext`] in the envelope headers.
-///  - there are no transactions in the envelope, since in this case sampling by trace is redundant.
+///  - there are no transactions or events in the envelope, since in this case sampling by trace is redundant.
 pub fn get_sampling_key(envelope: &Envelope) -> Option<ProjectKey> {
-    let transaction_item = envelope.get_item_by(|item| item.ty() == &ItemType::Transaction);
-
-    // if there are no transactions to sample, return here
-    transaction_item?;
-
+    // If we the envelope item is not of type transaction or event, we will not return a sampling key
```
iker-barriocanal: nit:
```suggestion
    // If the envelope item is not of type transaction or event, we will not return a sampling key
```

iker-barriocanal (approved):

jjbayer (approved):
This should be fine because we now explicitly check the event type before sampling:

https://github.com/getsentry/relay/blob/5d764fc1d7d877586193ba49fe56f278c2346461/relay-server/src/actors/processor.rs#L2252-L2254

Let's do a canary deploy and see if the number of cached project states increases drastically.

PR #2150, opened 2023-05-23
ref: Use structured logging through tracing

jan-auer:
Pass errors as field to the error macro so that the error, sources, and backtrace (in the future) can be handled directly by the logging subscriber and the sentry SDK. This improves the quality and grouping of errors to Sentry and makes for conciser error messages.

By convention, errors are now passed via an `error` field before the message. The message itself is not required, but encouraged if the error is too generic to be self-explanatory.

```rust
relay_log::error!(
    error = &error as &dyn Error,
    "descriptive message"
);
```

On the command line, this error will be shown as:

```
ERROR relay::module: descriptive message  error=source error message
```

In Sentry, the error is wrapped in a synthetic exception, with the title pointing at the source location. The actual error and its sources follow in a chained exception.

#skip-changelog


jan-auer:
The way errors are logged now, the main message (which has most information about the bug) does not contribute to the issue title or grouping. This is addressed in https://github.com/getsentry/sentry-rust/pull/585 and will require an SDK bump.

## Before

image

image

## After

image

image

jjbayer (approved):

PR #2151, opened 2023-05-24
ref: Use different error message for empty strings in schema processing

### LINKED ISSUE #2137: "expected a non-empty value" is a super ambiguous error message for empty strings
I recommend we rephrase this to something like "expected a non-empty string". This is what it looks like in the interface:

Screenshot 2023-05-19 at 15 12 10

Took me some JSON digging to understand what Sentry wants me to do.

https://github.com/getsentry/relay/blob/6f2946c479a44e96722c95ccf47b5985ce4e5dd5/relay-general/src/types/meta.rs#L314

lforst:
Fixes #2137

### DIFF
--- a/relay-general/src/store/schema.rs
+++ b/relay-general/src/store/schema.rs
@@ -13,7 +13,7 @@ impl Processor for SchemaProcessor {
         state: &ProcessingState<'_>,
     ) -> ProcessingResult {
         value_trim_whitespace(value, meta, state);
-        verify_value_nonempty(value, meta, state)?;
+        verify_value_nonempty_string(value, meta, state)?;
         verify_value_characters(value, meta, state)?;
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


relay-general/src/store/schema.rs
```diff
@@ -139,9 +155,35 @@ mod tests {
...
+                bar: Annotated::from_error(Error::expected("a non-empty string"), None),
+                bar2: Annotated::new(String::default())
+            })
+        );
+    }
+
     #[test]
     fn test_nonempty_string() {
-        assert_nonempty_base::<String>();
+        assert_nonempty_string::<String>();
```
jjbayer: Can we make `assert_nonempty_base` expect a `&str` with the expected error message instead of duplicating the function?

jjbayer (approved):
See comment about test, other than that, LGTM!

relay-general/src/store/schema.rs
lforst: Great feedback. Thanks! 29a689ef7cdfa2f3c3d05aff44626cb77a7067b6

PR #2155, opened 2023-05-25
instr(txnames): Tag 404s

jjbayer:
We now mark all URL transactions as sanitized after 10 clusterer runs (see https://github.com/getsentry/sentry/pull/48993), except 404s. Tag those explicitly on the metric so we can confirm that non-404s are now sanitized at 100%.

#skip-changelog

### DIFF
--- a/relay-server/src/utils/statsd.rs
+++ b/relay-server/src/utils/statsd.rs
@@ -51,12 +51,16 @@ where
     };
 
     let new_source = inner.get_transaction_source();
+    let is_404 = inner
+        .get_tag_value("http.status_code")
+        .map_or(false, |s| s == "404");
 
     relay_statsd::metric!(
         counter(RelayCounters::TransactionNameChanges) += 1,
         source_in = old_source.as_str(),
         changes = changes,
         source_out = new_source.as_str(),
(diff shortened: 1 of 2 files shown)


relay-server/src/utils/statsd.rs
```diff
@@ -51,12 +51,16 @@ where
...
+    let is_404 = inner
+        .get_tag_value("http.status_code")
+        .map_or(false, |s| s == "404");
 
     relay_statsd::metric!(
         counter(RelayCounters::TransactionNameChanges) += 1,
         source_in = old_source.as_str(),
         changes = changes,
         source_out = new_source.as_str(),
+        is_404 = if is_404 { "true" } else { "false" },
```
iker-barriocanal, line 63: Do we want to track in this metric how many 404 transactions are? If not, we could get rid of the tag and only emit the metric on non-404s.

iker-barriocanal (approved):

relay-server/src/utils/statsd.rs
jjbayer: Good point, might do that in a follow up.

PR #2156, opened 2023-05-25
feat(spans): Extract exclusive time metric

iker-barriocanal:
Extracts the exclusive time metric from spans. The exclusive time doesn't consider all cases, such as sub-transactions, but we accept this limitation for now with the goal to move forward.

#skip-changelog

### DIFF
--- a/relay-general/src/store/normalize.rs
+++ b/relay-general/src/store/normalize.rs
@@ -816,7 +816,7 @@ pub fn light_normalize_event(
         ); // Measurements are part of the metric extraction
         normalize_breakdowns(event, config.breakdowns_config); // Breakdowns are part of the metric extraction too
 
-        if config.light_normalize_spans {
+        if config.light_normalize_spans && event.ty.value() == Some(&EventType::Transaction) {
             // XXX(iker): span normalization runs in the store processor, but
             // the exclusive time is required for span metrics. Most of
             // transactions don't have many spans, but if this is no longer the
--- a/relay-general/src/store/normalize.rs
+++ b/relay-general/src/store/normalize.rs
@@ -1,4 +1,5 @@
 use std::collections::hash_map::DefaultHasher;
+use std::collections::BTreeSet;
 use std::hash::{Hash, Hasher};
 use std::mem;
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


jjbayer (approved):

relay-general/src/store/normalize.rs
```diff
@@ -814,6 +816,15 @@ pub fn light_normalize_event(
         ); // Measurements are part of the metric extraction
         normalize_breakdowns(event, config.breakdowns_config); // Breakdowns are part of the metric extraction too
 
+        if config.light_normalize_spans {
+            // XXX(iker): span normalization runs in the store processor, but
+            // the exclusive time is required for span metrics. Most of
+            // transactions don't have many spans, but if this is no longer the
+            // case and we roll this flag out for most projects, we may want to
+            // reconsider this approach.
+            spans::normalize_spans(event, &BTreeSet::from([SpanAttribute::ExclusiveTime]));
```
jjbayer, line 825: Should we also check that the event is a transaction, and the span attribute was actually requested in project configs?

See

https://github.com/getsentry/relay/blob/9dc2bcf145a598de7146f53df58dbf875a27a158/relay-general/src/store/normalize.rs#L197-L199

iker-barriocanal: Added check on transaction events [c5517da](https://github.com/getsentry/relay/pull/2156/commits/c5517daaba49abf320dd6cad14e9c8eceb746be4). Didn't add the other check because exclusive time must unconditionally be emitted as a metric and exist as a span property for the selected projects, so this seems the easiest way forward. Let me know if this may have negative and impactful consequences so that I reconsider this decision.

jjbayer: Should be fine, since `exclusive-time` is enabled by default:

https://github.com/getsentry/sentry/blob/e88987f897aa6b855dce0aca247c392d90793d85/src/sentry/projectoptions/defaults.py#L97-L98

If you want to be totally correct, you could do the intersection of `config.span_attributes` and `[SpanAttribute::ExclusiveTime]`, but it's probably not worth the effort.

iker-barriocanal: Yes, but I don't want to rely on that flag either (even if it's very unlikely to get changed).

PR #2157, opened 2023-05-25
feat: Add Sentry performance monitoring

jan-auer:
This is an early prototype to show performance monitoring instrumentations using
`tracing` spans and the Sentry SDK.

Tasks:

- [x] Add the tracing integration
- [x] Enable performance monitoring (trace sampling)
- [x] Expose a configuration field for the sample rate
- [x] Set defaults for the sample rate based on build type
- [x] Use [structured span ops](https://develop.sentry.dev/sdk/performance/span-operations/) and instead put function names into description (might need upstream fix)
- [x] Add tracing instrumentation to all extractors
- [x] Test performance overhead with sample rate of `0.0` in test infra

Possible followup:

 - Spawn connected transactions for background processing
 - Spawn transactions for all background jobs, such as project fetching
 - Contribute tracing instrumentation upstream to axum.
 - Lift span instrumentation into `relay-system` so that traces are
   automatically connected and message handlers (along with message variants,
   probably) are automatically created as spans.


relay-server/Cargo.toml
```diff
@@ -78,6 +78,10 @@ tower-http = { version = "0.4.0", default-features = false, features = ["catch-p
 url = { version = "2.1.1", features = ["serde"] }
 zstd = { version = "0.12.3", optional = true }
 
+
+# TODO(ja): Here
+tracing = "0.1"
```
jan-auer: Added a direct dependency on `tracing` here, because re-exporting the macro through `relay-log` isn't possible. Since we will need this dependency on `tracing` now, we could also stop the re-export of other tracing macros through `relay-log`, though those have been useful in the migration. I'm up for suggestions, defaulting to keeping everything as-is.

relay-server/src/endpoints/common.rs
```diff
@@ -348,6 +350,7 @@ pub async fn handle_envelope(
     let checked = state
         .project_cache()
         .send(CheckEnvelope::new(managed_envelope))
+        .instrument(tracing::debug_span!("ProjectCache::CheckEnvelope"))
```
jan-auer: Noted in the PR description for a follow-up: We have the opportunity to make `Addr::send` carry over spans or context automatically, so that we do not have to instrument ourselves. It's unclear what a suitable automatic `op` and `description` for these spans are, however.

relay-server/src/endpoints/common.rs
```diff
@@ -145,6 +146,25 @@ impl IntoResponse for BadStoreRequest {
     }
 }
 
+/// Adds the instrumentation layer on top of the [`bytes::Bytes`].
+#[derive(Debug)]
+pub(crate) struct BytesWrapper(pub(crate) Bytes);
```
jan-auer: What would you think if we move this into `relay_server::extractors`? We could have a module there with overrides of the built-in axum extractors.

_nit_: "Wrapper" is a rather generic name, I'm thinking something like `InstrumentedBytes`, though that's very long :)

relay-server/src/endpoints/attachments.rs
```diff
@@ -13,6 +13,25 @@ use crate::extractors::RequestMeta;
 use crate::service::ServiceState;
 use crate::utils;
 
+#[derive(Debug)]
+struct MultipartWrapper(Multipart);
```
jan-auer: I've had an open TODO for a while to ditch axum's Multipart and instead use `multer` directly. Since we now need to write a custom extractor, what if we include that change in this PR?

relay-server/src/endpoints/security_report.rs
```diff
@@ -25,19 +25,23 @@ struct SecurityReportParams {
     meta: RequestMeta,
     #[from_request(via(Query))]
     query: SecurityReportQuery,
-    body: BytesWrapper,
+    body: InstrumentedBytes,
```
jan-auer:
```suggestion
    #[from_request(via(InstrumentedBytes))]
    body: Bytes,
```

olksdr: nice! done in [fff712c](https://github.com/getsentry/relay/pull/2157/commits/fff712ccf76b76f3fcda4b7db7f67c0e6289ca1e)

relay-server/src/endpoints/attachments.rs
jan-auer: Moved to https://github.com/getsentry/relay/issues/2218

olksdr (approved):
:shipit:

PR #2160, opened 2023-05-30
fix: Disable `trust-dns` option

### LINKED ISSUE #2147: Panic on parsing resolv.conf
### Environment

Running docker image getsentry/relay:23.5.0  on nomad

Same thing on symbolicator

### Steps to Reproduce
resolv.conf
```
nameserver x.x.x.x
nameserver y.y.y.y
option rotate timeout:2
```

### Expected Result

running 

### Actual Result
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Builder, source: Custom { kind: Other, error: "error reading DNS system conf: Error parsing resolv.conf: InvalidDirective(2)" } }', relay-server/src/actors/upstream.rs:719:14
```

olksdr:
See my comment https://github.com/getsentry/relay/issues/2147#issuecomment-1568382277

Basically `resolv-conf`  which is used by `reqwset` through `trust-dns-resolver` crate supports a limited number of options and some of them , when not supported make the Relay panic, when the `Upstream` service is created. 

The upstream `reqwest` crate disables `trust-dns` by default, to avoid similar issues, see https://github.com/seanmonstar/reqwest/pull/437, which we might want to do as well. 

fix: https://github.com/getsentry/relay/issues/2147

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1109,18 +1109,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "enum-as-inner"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c9720bba047d567ffc8a3cba48bf19126600e249ab7f128e9233e6376976a116"
-dependencies = [
- "heck",
- "proc-macro2",
- "quote",
- "syn 1.0.109",
-]
-
-[[package]]
 name = "enum-primitive-derive"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


iker-barriocanal (approved):
Should we consider this a fix for the changelog?

olksdr:
> Should we consider this a fix for the changelog?

So, this is the questions. Should we even disable it? Or better yet to introduce the config option for users to be able to opt-out (or opt-in)?

olksdr:
closing in favour of https://github.com/getsentry/relay/pull/2164

PR #2161, opened 2023-05-30
feat(spans): Apply span description rules

iker-barriocanal:
Applies span description name rules to span descriptions.

Transaction name rules also apply to span descriptions for now, but this will soon not be the case. As a result, the avoidable clones in the transaction processor will also go away (this is a shortcut for now). On the other hand, we are still applying rules to `op=http*` spans, ignoring the rule's scope; this is another shortcut that will get addressed at a later time.

#skip-changelog


relay-general/src/store/transactions/processor.rs
```diff
@@ -37,14 +37,19 @@ pub struct TransactionsProcessor<'r> {
 }
 
 impl<'r> TransactionsProcessor<'r> {
-    pub fn new(name_config: TransactionNameConfig<'r>, scrub_span_descriptions: bool) -> Self {
-        let mut span_desc_rules = Vec::new();
+    pub fn new(
+        name_config: TransactionNameConfig<'r>,
+        scrub_span_descriptions: bool,
+        span_description_rules: Option<&Vec<SpanDescriptionRule>>,
```
olksdr, line 43: To avoid cloning the rules in this function, which is not really great, it would be better that the caller provides not an option, but already the reference (or maybe mutable reference) to the list of `SpanDescriptionRule` rules.  And then in this function you do not clone anything and just re-use already provided list as it is. 

What do you think about it?

olksdr (approved):
Approve to unblock, but I would like us to handle open TODO and cloning as soon as possible.

relay-general/src/store/transactions/processor.rs
iker-barriocanal: Discussed this -- decided to move forward and revisit this in the future.

PR #2162, opened 2023-05-31
instr(txname): Log sentry error for remaining URL transactions

jjbayer:
Temporarily log a sentry error for 1% of transactions that are still `source:url` after normalization to get some project IDs / event IDs to investigate.

Assumes that event IDs are randomly distributed.

#skip-changelog


olksdr (approved):
:shipit:

Issue #2158, opened 2023-05-26
[EPIC] Ensure processed transactions always exist for processed profiles


jjbayer:
Because of #2083, it can happen that a processed profile is counted as "accepted" while the corresponding processed transaction is rate limited:

```mermaid
sequenceDiagram
    Ingestion->>+Aggregator: transaction metric
    Ingestion->>+RateLimiter: profile
    RateLimiter->>+Outcomes: profile:accepted
    RateLimiter->>+RateLimiter: update rate limits
    Aggregator->>+RateLimiter: transaction metric
    RateLimiter->>+Outcomes: transaction:rate_limited
```

This violates the requirement that processed profiles are always dropped with their corresponding processed transactions.

A possible solution would be to change the way that we count processed profiles: Instead of counting them in two different places (see [PR description here](https://github.com/getsentry/relay/pull/2054)), add a `has_profile` tag to the transaction counter metric, and define

```
processed_profiles := count(processed_transactions) WHERE has_profile = true
```

We have to make sure that this way of counting does not produce more undesired edge cases though.

```[tasklist]
### Code changes
- [x] relay: Revert PRs around counting profiles. https://github.com/getsentry/relay/pull/2163
- [x] relay: Add `has_profile(s)` tag on `transaction.duration` metric. https://github.com/getsentry/relay/pull/2165
- [x] relay: Update metrics rate limiter to emit `rate_limited` outcome for profiles.
- [x] relay: Make profile category depend on `metrics_extracted` flag (either of the transaction or set it on the profile item itself). https://github.com/getsentry/relay/pull/2166
- [x] sentry: Update metrics billing consumer to emit `accepted` outcome for profiles. https://github.com/getsentry/sentry/pull/50047
```

jjbayer:
- The proposed solution requires that the relationship between transactions and profiles will always be `1 to 0..1`. Unless we encode the number of profiles into the tag value ("1", "2", ...).
- The proposed solution will mean we under-count when transaction metrics are extracted in external relays that do not set the `has_profile` tag yet. We could bump the metrics extraction version, but then we would under-count transactions that get sampled out in external relays (because of https://github.com/getsentry/relay/issues/1631).

jjbayer:
Rollout:

1. Update consumer observing `has_profiles`
2. Update Relays

PR #2165, opened 2023-05-31
ref(profiles): Count processed profiles with metrics

jjbayer:
As described in https://github.com/getsentry/relay/issues/2158, track the number of processed profiles by tagging the corresponding transaction metric.

Also: Do not accept multiple profiles per envelope.

**Not in this PR:**
* Distinguishing between `Profile` and `ProfileIndexed`
* Rate limiting processed profiles based on metrics

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -2729,7 +2729,6 @@ mod tests {
     use std::str::FromStr;
 
     use chrono::{DateTime, TimeZone, Utc};
-    use insta::assert_debug_snapshot;
     use similar_asserts::assert_eq;
 
     use relay_common::{DurationUnit, MetricUnit, Uuid};
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -13,6 +13,7 @@ use chrono::{DateTime, Duration as SignedDuration, Utc};
 use flate2::write::{GzEncoder, ZlibEncoder};
 use flate2::Compression;
 use once_cell::sync::OnceCell;
+use relay_profiling::ProfileError;
 use serde_json::Value as SerdeValue;
 use tokio::sync::Semaphore;
 
@@ -307,6 +308,9 @@ struct ProcessEnvelopeState {
 
     /// The managed envelope before processing.
     managed_envelope: ManagedEnvelope,
+
+    /// Whether there is a profiling item in the envelope.
+    has_profile: bool,
 }
 
 impl ProcessEnvelopeState {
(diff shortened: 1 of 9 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -1061,15 +1065,28 @@ impl EnvelopeProcessorService {
...
+            ItemType::Profile => {
+                if !found_profile {
+                    // Found first profile, keep it.
+                    found_profile = true;
+                } else {
+                    // We found a second profile, drop it.
+                    return ItemAction::Drop(Outcome::Invalid(DiscardReason::Profiling(
+                        relay_profiling::discard_reason(ProfileError::TooManyProfiles),
+                    )));
+                }
```
jjbayer: @phacops I added this logic as you suggested.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -358,15 +362,21 @@ pub fn extract_transaction_metrics(
...
 }
 
+struct ExtractInput<'a> {
+    aggregator_config: &'a AggregatorConfig,
+    config: &'a TransactionMetricsConfig,
+    event: &'a Event,
+    transaction_from_dsc: Option<&'a str>,
+    sampling_result: &'a SamplingResult,
+    has_profile: bool,
+}
```
jjbayer, line 372: The linter finally started complaining about the number of arguments to `extract_transaction_metrics_inner` so I created this struct.

jjbayer:
Letting one integration test fail deliberately. It will pass when https://github.com/getsentry/relay/pull/2166 is merged into this branch.

relay-server/src/actors/processor.rs
```diff
@@ -1061,15 +1065,28 @@ impl EnvelopeProcessorService {
...
-            ItemType::Profile => match relay_profiling::parse_metadata(&item.payload()) {
-                Ok(_) => ItemAction::Keep,
-                Err(err) => ItemAction::Drop(Outcome::Invalid(DiscardReason::Profiling(
-                    relay_profiling::discard_reason(err),
-                ))),
-            },
+            ItemType::Profile => {
+                if !found_profile {
+                    // Found first profile, keep it.
+                    found_profile = true;
```
iker-barriocanal: Relay also considers towards outcomes dropped profiles that could not parsed, even if they aren't dropped by dynamic sampling. https://github.com/getsentry/relay/pull/2054 mentions only "valid" profiles should be considered:

>In other words, "processed profiles" should count all valid, non rate limited profiles seen by Relays, even if they were dropped by dynamic sampling.

iker-barriocanal (approved):
Assuming the PR fulfills the intention in my comment below.

relay-server/src/actors/processor.rs
jjbayer: The wording in that PR description is too vague, it should say: "_accepted_ processed profiles should count all valid non rate limited profiles seen by Relays" .

A better way to think of it is two data streams splitting ways (see diagram [here](https://github.com/getsentry/relay/pull/2166)), and getting counted towards a different category regardless of whether they're accepted, rate limited, or invalid.

PR #2167, opened 2023-05-31
fix(spans): Consider quotes to extract the domain

iker-barriocanal:
Consider `"`s in an SQL query to extract the domain, aka table name.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -690,7 +690,7 @@ fn extract_span_metrics(
 }
 
 static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r#"(?i)(from|into)(\s|"|\()+(?P<table>(\w+(\.\w+)*))(\s|"|\))+"#).unwrap()
+    Regex::new(r#"(?i)(from|into)(\s|"|'|\()+(?P<table>(\w+(\.\w+)*))(\s|"|'|\))+"#).unwrap()
 });
 
 /// Returns the table in the SQL query, if any.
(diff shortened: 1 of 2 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -689,8 +689,9 @@ fn extract_span_metrics(
     Ok(())
 }
 
-static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r#"(?i)(from|into)(\s|\()+(?P<table>(\w+(\.\w+)*))(\s|\))+"#).unwrap());
+static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r#"(?i)(from|into)(\s|"|\()+(?P<table>(\w+(\.\w+)*))(\s|"|\))+"#).unwrap()
```
olksdr: Just a quick question, do we always have `""` around the table name , and no other types of quotes are used? Can it be there will be no quotes at all?

olksdr (approved):
left one question, otherwise lgtm

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: Considered single quotes in [49abd68](https://github.com/getsentry/relay/pull/2167/commits/49abd6878fba56c02c9ce2fa3588c3715fdafe9e) -- note that we are intentionally scrubbing strings within single quotes, so I'm not sure how helpful this can be.

It's possible there are no quotes at all, yes. `(\s|"|'|\()+` considers spaces, double and single quotes, and brackets (closing bracket in the closing group).

PR #2168, opened 2023-05-31
feat(profiling): Reject profiles longer than 30s

phacops:


relay-profiling/src/sample.rs
```diff
@@ -213,6 +213,13 @@ impl SampleProfile {
         true
     }
 
+    fn check_duration(&self) -> bool {
+        if let Some(sample) = &self.profile.samples.last() {
+            return sample.elapsed_since_start_ns <= 30_000_000_000;
```
Zylphrex: nit: Would be nice if we could use the const declared in the android module and do something like `THIRTY_SECONDS.as_nanos()`

Zylphrex (approved):

JonasBa:
sad

relay-profiling/src/lib.rs
```diff
@@ -114,6 +115,8 @@ use crate::extract_from_transaction::{extract_transaction_metadata, extract_tran
 pub use crate::error::ProfileError;
 pub use crate::outcomes::discard_reason;
 
+const MAX_PROFILE_DURATION: Duration = Duration::new(30, 0);
```
jjbayer: Readability nit:

```suggestion
const MAX_PROFILE_DURATION: Duration = Duration::from_secs(30);
```

relay-profiling/src/sample.rs
```diff
@@ -327,6 +337,10 @@ fn parse_profile(payload: &[u8]) -> Result<SampleProfile, ProfileError> {
         return Err(ProfileError::MalformedStacks);
     }
 
+    if !profile.is_below_max_duration() {
```
jjbayer: Any reason why the method returns `is_below` and not `is_above`?

jjbayer (approved):
nit: There's already a bunch of unit tests in this file, you could copy-paste one to verify the new behavior.

relay-profiling/src/sample.rs
phacops: No. Better to change though so I don't have the `!`.

PR #2169, opened 2023-05-31
feat(profiling): Drop profiles without a transaction in the same envelope

phacops:
This aims to drop profiles not bundled with a transaction in the same envelope.

Support for our legacy cocoa profiles is removed at the same time since they were sent only in a different envelope.

We'll merge this when we formally end of our beta.


jjbayer (approved):

relay-server/src/actors/processor.rs
```diff
@@ -1061,7 +1061,16 @@ impl EnvelopeProcessorService {
...
     fn filter_profiles(&self, state: &mut ProcessEnvelopeState) {
+        let transaction_count: usize = state
+            .managed_envelope
+            .envelope()
+            .items()
+            .filter(|item| item.ty() == &ItemType::Transaction)
+            .count();
+
         state.managed_envelope.retain_items(|item| match item.ty() {
+            // Drop profile without a transaction in the same envelope
```
```suggestion
            // Drop profile without a transaction in the same envelope.
```

olksdr (approved):

Zylphrex (approved):

PR #2170, opened 2023-05-31
feat(replay): combined envelope items

JoshFerge:
Implements the Combined Replay Envelope item as specified here: https://github.com/getsentry/rfcs/pull/82

- New ItemType
- New processing function that takes the two items, combines them in a MsgPack, and re-inserts the item into the envelope
    - This just uses msgpack keys and bytes as the values. Attempting to serialize the nested key/values of the messages proved problematic as `Annotated` cannot be easily serialized to messagepack.
- New logic in `store` to handle this new ItemType. It will emit to a topic already in use.

Other Changes:
- ReplayEvent can no longer "create an event" as this was putting it in a separate envelope for processing. The Envelope containing a replay should always contain an `event_id`, so this shouldn't be a problem.

Open Questions:

I'm not quite sure how to think about processing vs non-processing relays and this feature. This is something I could use ingests help on, and I have a commented out test for confirming this behavior once we know what we want. e.g.
- Should this only run in processing mode? If not, is there anything else we need to add here?


relay-server/src/actors/store.rs
```diff
@@ -952,6 +985,17 @@ struct ReplayRecordingChunkKafkaMessage {
...
+struct CombinedReplayEventAndRecordingKafkaMessage {
+    /// Raw event payload.
+    payload: Bytes,
+    /// The event id.
+    replay_id: EventId,
+    /// The project id for the current event.
+    project_id: ProjectId,
+    // Number of days to retain.
+    retention_days: u16,
+}
```
cmanallen, line 1011: This struct is missing fields.  It should match the fields from `ReplayRecordingNotChunkedKafkaMessage` and be versioned with `version: u8`.

```rust
#[derive(Debug, Serialize)]
struct CombinedReplayEventAndRecordingKafkaMessage {
    replay_id: EventId,
    key_id: Option<u64>,
    org_id: u64,
    project_id: ProjectId,
    received: u64,
    retention_days: u16,
    payload: Bytes,
    version: u8,
}
```

relay-server/src/actors/processor.rs
```diff
@@ -1304,6 +1304,46 @@ impl EnvelopeProcessorService {
...
+        let project_state = &state.project_state;
+        let combined_envelope_items =
+            project_state.has_feature(Feature::SessionReplayCombinedEnvelopeItems);
+
+        if combined_envelope_items {
+            // If this flag is enabled, combine both items into a single item,
+            // and remove the original items.
+            // The combined Item's payload is a MsgPack map with the keys
+            // "replay_event" and "replay_recording".
+            // The values are the original payloads of the items.
```
cmanallen: I didn't see the `replay_recording_headers` key mentioned.  It can always be added on a future payload version if needed.

This is how the headers are extracted currently:

```rust
let mut split = bytes.splitn(2, |b| b == &b'\n');
let header = split
  .next()
  .ok_or(ParseRecordingError::Message("no headers found"))?;

let body = match split.next() {
  Some(b"") | None => return Err(ParseRecordingError::Message("no body found")),
  Some(body) => body,
};
```

We can break it out into a function and you could re-use it in this method.  Returning `Result<Tuple<Headers, Body>, Error>` or similar. Error case could be handled the same way as `process_replays` by writing a discarded outcome.

cmanallen (commented):
I know you explicitly asked for no review which is why this isn't a review!  It just looks like one :P

These are just general comments from looking at the code.  No questions.  No calls for action.  Do your thing and let me know if you want me to help at all.

bruno-garcia:
Taking note that we'll rely on this change to have the data we need in order to create Dead Click events:

* https://github.com/getsentry/sentry/issues/48259

relay-server/src/actors/store.rs
JoshFerge: was WIP, this is fixed now.

relay-server/src/actors/processor.rs
JoshFerge: if you wanna take a crack at implementing this feel free, was having issues w/ borrowing and references when i tried. for now it is just included with the recording payload.

relay-server/src/actors/processor.rs
```diff
@@ -1304,6 +1304,55 @@ impl EnvelopeProcessorService {
...
+                {
+                    let mut data = Vec::new();
+                    let mut combined_item_payload = BTreeMap::new();
+
+                    combined_item_payload
+                        .insert("replay_event", replay_event_item.payload().to_vec());
+                    combined_item_payload
+                        .insert("replay_recording", replay_recording_item.payload().to_vec());
+
+                    if let Err(e) = rmp_serde::encode::write(&mut data, &combined_item_payload) {
```
JoshFerge: is this the best way to do the msgpack encoding?

relay-server/src/actors/processor.rs
```diff
@@ -1304,6 +1304,55 @@ impl EnvelopeProcessorService {
...
+            // The values are the original payloads of the items.
+            let envelope = &mut state.envelope_mut();
+            if let Some(replay_event_item) =
+                envelope.take_item_by(|item| item.ty() == &ItemType::ReplayEvent)
+            {
+                if let Some(replay_recording_item) =
+                    envelope.take_item_by(|item| item.ty() == &ItemType::ReplayRecording)
+                {
+                    let mut data = Vec::new();
+                    let mut combined_item_payload = BTreeMap::new();
```
JoshFerge: I wonder if I could use a struct instead of a BTree and if that would be preferred?

iker-barriocanal:
Hi @JoshFerge! I see https://github.com/getsentry/rfcs/pull/82 is not merged yet so I assume there's still work pending, such as [this question](https://github.com/getsentry/rfcs/pull/82#discussion_r1175426915). Could we first agree on the RFC before jumping into the implementation?

bruno-garcia:
> Hi @JoshFerge! I see https://github.com/getsentry/rfcs/pull/82 is not merged yet so I assume there's still work pending, such as [this question](https://github.com/getsentry/rfcs/pull/82#discussion_r1175426915). Could we first agree on the RFC before jumping into the implementation?

Iker the comment is about the sdk change and won't affect this PR. Can we get a hand here to help clarify Josh's questions?
We'll merge the rfc in parallel.

Specifically:

> Open Questions:

> I'm not quite sure how to think about processing vs non-processing relays and this feature. This is something I could use ingests help on, and I have a commented out test for confirming this behavior once we know what we want. e.g.

> Should this only run in processing mode? If not, is there anything else we need to add here?

relay-server/src/actors/processor.rs
cmanallen: I implemented it as a struct.  BTreeMap will have some amount of overhead.  Not enough for us to worry about but the structure was easy enough to represent and makes it more clear what the serialization output is.

relay-server/src/actors/processor.rs
cmanallen: Its a perfectly fine way to do it.  We can make it better though by pre-allocating memory in our `data` buffer (`Vec::with_capacity(some_number)`).  By doing this pushing bytes into the buffer is essentially free.

cmanallen:
@JoshFerge Updated the pull.  We're now splitting the headers and the body.  I've also added error handling around serialization failure.  There are comments which should explain the intent of the changes.  My thought is if the envelope combining process fails we should default to the old style (e.g. two, un-merged envelope items).

JoshFerge:
> Hi @JoshFerge! I see [getsentry/rfcs#82](https://github.com/getsentry/rfcs/pull/82) is not merged yet so I assume there's still work pending, such as [this question](https://github.com/getsentry/rfcs/pull/82#discussion_r1175426915). Could we first agree on the RFC before jumping into the implementation?

Sorry, I wasnt sure how to update RFCs based off of a discussion, but see this comment from me last week, guiding this implementation: https://github.com/getsentry/rfcs/pull/82#issuecomment-1563605953. i'll go ahead and update the main Doc.

relay-server/src/utils/sizes.rs
```diff
@@ -63,6 +63,9 @@ pub fn check_envelope_size_limits(config: &Config, envelope: &Envelope) -> bool
             ItemType::UserReport => (),
             ItemType::Metrics => (),
             ItemType::MetricBuckets => (),
+            // The Combined Replay Envelope isn't generated on the client so its size does not need
+            // to be checked.
+            ItemType::CombinedReplayEventAndRecording => (),
```
iker-barriocanal, line 68: Even if envelopes are generated in Relays, size requirements may vary and we don't have control over external relays, which could be running an older version and sending bigger envelopes.

relay-server/src/envelope.rs
```diff
@@ -743,13 +752,14 @@ impl Item {
             ItemType::RawSecurity => true,
             ItemType::UnrealReport => true,
             ItemType::UserReport => true,
-            ItemType::ReplayEvent => true,
+            ItemType::ReplayEvent => false,
```
iker-barriocanal, line 755: >ReplayEvent can no longer "create an event" as this was putting it in a separate envelope for processing. The Envelope containing a replay should always contain an event_id, so this shouldn't be a problem.

Should we expect older SDKs to continue sending `ReplayEvent`s? This change seems like a breaking change for users not upgrading their SDKs.

relay-dynamic-config/src/feature.rs
```diff
@@ -9,6 +9,9 @@ pub enum Feature {
     SessionReplay,
     /// Enables data scrubbing of replay recording payloads.
     SessionReplayRecordingScrubbing,
+    /// Enables combining session replay envelope item (Replay Recordings and Replay Events).
+    /// into one item.
+    SessionReplayCombinedEnvelopeItems,
```
iker-barriocanal, line 14: Is it currently possible to send non-combined events in different envelopes? How should relay handle that when this feature flag is enabled?

iker-barriocanal (commented):
About processing vs non-processing relays ([docs](https://docs.sentry.io/product/relay/modes/)), the general approach should be to not target processing actions to processing relays unless there's a reason not to. Non-processing relays include our internal PoP Relays, but also external relays run by customers (including sidecars).

I have only skimmed through the PR, leaving a few questions.

relay-dynamic-config/src/feature.rs
cmanallen: Envelopes will always arrive in an uncombined state.  These combined envelopes are forwarded to our Kafka consumer.  If for whatever reason Relay can't or chooses not to combine the envelopes the consumer process will know how to handle the uncombined envelopes.

JoshFerge: ReplayEvents and ReplayRecordings will always be in the same envelope, and this has always been the case.

relay-server/src/envelope.rs
JoshFerge: I don't believe this is a breaking change, as ReplayEvents and ReplayRecordings are always in the same envelope, which has an event_id envelope header.

relay-server/src/utils/sizes.rs
JoshFerge: gotcha, will add a size here.

relay-server/src/envelope.rs
```diff
@@ -150,6 +152,9 @@ impl fmt::Display for ItemType {
             Self::Profile => write!(f, "profile"),
             Self::ReplayEvent => write!(f, "replay_event"),
             Self::ReplayRecording => write!(f, "replay_recording"),
+            Self::CombinedReplayEventAndRecording => {
+                write!(f, "combined_replay_event_and_recording")
```
jjbayer, line 156: We tend to use short names for everything in our transport protocols, so I would also go for brevity here:

```suggestion
                write!(f, "replay_combined")
```

relay-server/src/actors/processor.rs
```diff
@@ -1267,6 +1268,95 @@ impl EnvelopeProcessorService {
         Ok(())
     }
 
+    fn process_replays_combine_items(
+        &self,
+        state: &mut ProcessEnvelopeState,
+    ) -> Result<(), ProcessingError> {
```
jjbayer, line 1279: If this function does not fail, no need to return a `Result`.

```suggestion
    )
```

relay-server/src/actors/processor.rs
```diff
@@ -1267,6 +1268,95 @@ impl EnvelopeProcessorService {
...
+        let project_state = &state.project_state;
+        let combined_envelope_items =
+            project_state.has_feature(Feature::SessionReplayCombinedEnvelopeItems);
```
jjbayer: nit: Making this a [guard](https://en.wikipedia.org/wiki/Guard_(computer_science)#Flatter_code_with_less_nesting) might improve readability

```suggestion
        if !project_state.has_feature(Feature::SessionReplayCombinedEnvelopeItems) { return }
```

relay-server/src/actors/processor.rs
```diff
@@ -1267,6 +1268,95 @@ impl EnvelopeProcessorService {
...
+        if combined_envelope_items {
+            // If this flag is enabled, combine both items into a single item,
+            // and remove the original items.
+            // The combined Item's payload is a MsgPack map with the keys
+            // "replay_event" and "replay_recording".
+            // The values are the original payloads of the items.
+            let envelope = &mut state.envelope_mut();
+            if let Some(replay_event_item) =
+                envelope.take_item_by(|item| item.ty() == &ItemType::ReplayEvent)
+            {
```
jjbayer: nit: Can also be a guard.

```suggestion
            let Some(replay_event_item) =
                envelope.take_item_by(|item| item.ty() == &ItemType::ReplayEvent) else { return }
```

relay-server/src/actors/processor.rs
```diff
@@ -1267,6 +1268,95 @@ impl EnvelopeProcessorService {
...
+                        // If the operation was successful we can create our new combined envelope
+                        // items and append it to the envelope stream.
+                        let mut combined_item =
+                            Item::new(ItemType::CombinedReplayEventAndRecording);
+                        combined_item.set_payload(ContentType::MsgPack, buffer);
+                        envelope.add_item(combined_item);
+                    }
+                } else {
+                    envelope.add_item(replay_event_item)
+                }
```
jjbayer: Instead of taking the `ReplayEvent` and putting it back when there is no recording, can we take the recording first, and only remove the `ReplayEvent` if it exists?

relay-server/src/actors/processor.rs
```diff
@@ -1267,6 +1268,95 @@ impl EnvelopeProcessorService {
...
+            {
+                if let Some(replay_recording_item) =
+                    envelope.take_item_by(|item| item.ty() == &ItemType::ReplayRecording)
+                {
+                    // Split the headers from the body so they can be stored on separate keys.
+                    // This was previously done in another processing step (`process_replays`). To
+                    // be here we know it must have been successful otherwise the envelope would
+                    // have been dropped.
+                    //
+                    // TODO: Should we handle this possible error state more gracefully?
```
jjbayer: Yes please.

relay-dynamic-config/src/feature.rs
jjbayer: > Envelopes will always arrive in an uncombined state.

@cmanallen If I understand the RFC correctly, SDKs will at some point send combined replay items though, right?

relay-server/src/actors/processor.rs
```diff
@@ -1267,6 +1268,95 @@ impl EnvelopeProcessorService {
...
+                    // Potentially following the error pattern established in the msgpack
+                    // serialization step below?
+                    let (headers, body) =
+                        split_headers_from_body(replay_recording_item.payload().into())
+                            .expect("This is pre-validated and should not fail.");
```
jjbayer: Instead of performing this split twice (once in `process_replays` and once here), I would unite the two functions and serialize either back into a recording item _or_ into a message pack item:

```mermaid
flowchart TD
    A[event, recording] -->|split| B[event, header, body]
    B -->|scrub| C[event, header, body]
    C --> D{combine?}
    D -->|yes| E[MsgPack: event, header, body]
    D -->|no| F[event, recording]
```

jjbayer (changes requested):
I'm unclear on whether or not Relay should be able to handle combined items coming from SDKs:

> We'd then modify the sentry-sdk to send combined item types going forward

([rfc](https://github.com/getsentry/rfcs/blob/main/text/0082-combined-replay-envelope-item.md))

> Envelopes will always arrive in an uncombined state.

Primary concern is: in this PR's implementation combined items will bypass data scrubbing, because they are not considered by `process_replays`. So if we expect SDKs to send combined items, we have to split and re-join these combined items in relay to do the scrubbing.

> Should this only run in processing mode? If not, is there anything else we need to add here?

I think it makes sense to run this **only** in processing mode for now, for two reasons:

1. In general, it is easier to iterate on experimental changes if they are not performed by external relays, which might run outdated versions.
2. More specifically, if we change the data scrubbing logic for replays, items combined by outdated external relays would bypass this updated logic (external relays play a role similar to SDKs in this regard).

To do so, wrap the call to `process_replays_combined_item` in a `if_processing!({` condition.

relay-dynamic-config/src/feature.rs
cmanallen: @jjbayer 

> SDKs will at some point send combined replay items

I believe that requirement has been dropped.  I'll let @JoshFerge clarify the point.  To my knowledge, the SDK will not be updated to use a combined format.

---

Also I may have misunderstood the above statements so I'll clarify quickly.

>  ...send non-combined events in different envelopes

As in one request per each envelope item?  @JoshFerge is correct.  We do not support this.  What we support is a single HTTP request with two distinct envelope items.

cmanallen:
> > So if we expect SDKs to send combined items, we have to split and re-join these combined items in relay to do the scrubbing.

> Is this true?  We call `process_replays` prior to `process_replays_combine_items`.  `process_replays` should mutate the envelope items prior to combination?

@cmanallen Prior to combination, yes. But if the SDK sends a `CombinedReplayEventAndRecording`, `process_replays` will ignore it because it is neither `ReplayEvent` nor `ReplayRecording`, right?

relay-dynamic-config/src/feature.rs
JoshFerge: Yes, we are not planning to change SDK behavior in regards to these envelope items in the foreseeable future. And yes, all replay requests from our SDKs are currently always sent in the same envelope.

JoshFerge: I'll update the RFC diagrams today to reflect this decision.

cmanallen:
> Prior to combination, yes. But if the SDK sends a CombinedReplayEventAndRecording, process_replays will ignore it because it is neither ReplayEvent nor ReplayRecording, right?

@jjbayer The SDK won't send that type so no worries there.

relay-server/src/actors/processor.rs
cmanallen: @jjbayer Whats the best way to emit outcomes for dropped items?  The only way I've found to do it is with `retain_items` but it seems like that method won't work for this use case.

jjbayer: You could manually call the function that `retain_items` calls: https://github.com/getsentry/relay/blob/c057e370834049843932c3d11ac4d74850aaeb1d/relay-server/src/utils/managed_envelope.rs#L231-L243

cmanallen: @jjbayer So having done this change (https://github.com/getsentry/relay/pull/2170/commits/e2782b85740a67413bc91edd477f479a118c0abc) I feel it was not worth it.  I'd rather eat the cost of reading <20 bytes.  In the original code, there may have been some memory concerns where we drop the merged value in exchange for the split values.  That could have had implications (since we're doing it twice) but I leave it to you to decide if its worth it.

relay-server/src/actors/processor.rs
cmanallen: If the replay-event doesn't exist then we would have to add the recording back as an envelope item.  There's no benefit to this swap.

relay-server/src/actors/processor.rs
cmanallen: @jjbayer I am out for a little while.  Josh Ferge will be re-taking over the PR so please direct feedback items to him.  Thank you.

relay-server/src/actors/processor.rs
```diff
@@ -1267,6 +1273,177 @@ impl EnvelopeProcessorService {
...
+        state: &mut ProcessEnvelopeState,
+    ) -> Result<(), ProcessingError> {
+        let project_state = state.project_state.clone();
+        let envelope = state.envelope_mut();
+        let meta = envelope.meta().clone();
+        let event_id = envelope.event_id();
+
+        // If combined envelope processing is not enabled default to the original workflow.
+        if !project_state.has_feature(Feature::SessionReplayCombinedEnvelopeItems) {
+            return self.process_replays(state);
```
jjbayer, line 1287: Looks like we're calling process_replays twice in this case? Should we remove the call to `process_replays` here:

https://github.com/getsentry/relay/blob/e2782b85740a67413bc91edd477f479a118c0abc/relay-server/src/actors/processor.rs#L2572-L2573

jjbayer (changes requested):
> So having done this change (https://github.com/getsentry/relay/commit/e2782b85740a67413bc91edd477f479a118c0abc) I feel it was not worth it. I'd rather eat the cost of reading <20 bytes.

@cmanallen I see your point that it does not improve readability, even though the data flow looks cleaner to me. Feel free to revert that change, as long as we don't `.unwrap()` the second split result.

Still pressing "Request changes" because it looks like `process_replays` and `process_replays_combine_items` contain duplicate code right now (e.g. `process_recording`).

TBS1996:
@JoshFerge Do you still want to work on this? If not we will close it.

cmanallen:
@TBS1996 Sorry we've let this sit for a little bit.  Josh and I will sync and see if we need to pursue this still.  We'll close it out if so otherwise we'll get it back up to date.

jan-auer:
Please feel free to reopen this, once you plan to pursue it. Thanks!

PR #2124, opened 2023-05-15
feat(server): Add support x-vercel forwarding

TBS1996:
#2094 

This PR ensures relay can also use the `x-vercel-forwarded-for` header, and preferring this when also `x-forwarded-for` is also present.

Made a small refactor for the sake of writing unit tests. 

Also fixed some linting errors that were introduced to relay 2 weeks ago, not sure how it passed CI.

PR #2171, opened 2023-06-01
feat(protocol): Add lock attribute to the frame protocol

romtsn:
This is necessary in addition to #2018 to highlight the offending frame in the UI, so the lock has to be also on the per-frame level.

Also fixed the changelog for #2124 as it sneaked into the previous release, although hasn't been part of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 **Features**:
 
 - Use different error message for empty strings in schema processing. ([#2151](https://github.com/getsentry/relay/pull/2151))
+- Add support for X-Vercel-Forwarded-For header. ([#2124](https://github.com/getsentry/relay/pull/2124))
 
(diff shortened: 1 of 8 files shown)


olksdr (approved):

iker-barriocanal:
@romtsn please don't forget to update develop docs!

romtsn:
> @romtsn please don't forget to update develop docs!

that's on my list, but thanks for the reminder nonetheless :)

romtsn:
@olksdr @iker-barriocanal hmm, did I also have to add an entry to `py/CHANGELOG` actually? I'd like to extract this field on python so I can use it on the frontend

iker-barriocanal:
@romtsn yep, you can refer to [relay changelog instructions](https://github.com/getsentry/relay#instructions-for-changelogs).

PR #2172, opened 2023-06-01
feat(profiling): Extract app identifier from app context

Zylphrex:
To improve the in app detection on profiles, we need the app identifier from the app context. Going forwards, when the app identifier exists on android profiles, only modules that start with the app identifier will be classified as in app.


iker-barriocanal (changes requested):
The PR looks good to me! Could we add some tests to ensure this works as intended?

relay-profiling/src/extract_from_transaction.rs
```diff
@@ -15,14 +15,11 @@ pub fn extract_transaction_metadata(event: &Event) -> BTreeMap<String, String> {
     if let Some(dist) = event.dist.as_str() {
         tags.insert("dist".to_owned(), dist.to_owned());
     }
-    if let Some(dist) = event.dist.value() {
```
phacops, line 18: I feel like I wanted to extract something else but never renamed it and I forgot what it was.

phacops (approved):

relay-profiling/src/extract_from_transaction.rs
```diff
@@ -94,10 +97,85 @@ fn extract_http_method(transaction: &Event) -> Option<String> {
...
+                "start_timestamp": "2011-05-02T17:40:36Z",
+            })
+            .into(),
+        );
+
+        let metadata = extract_transaction_metadata(&event.0.unwrap());
+
+        assert_eq!(
+            metadata,
+            BTreeMap::from([
```
iker-barriocanal: nit: if you want to check the whole object and not just a couple of keys, consider using a snapshot instead. See [`insta::assert_debug_snapshot!`](https://github.com/getsentry/relay/blob/ed39746ec4b6cee3066f19d0cae54198bbe32971/relay-filter/src/config.rs#L236) if you prefer an in-line snapshot or [`insta::assert_annotated_snapshot!`](https://github.com/getsentry/relay/blob/ed39746ec4b6cee3066f19d0cae54198bbe32971/relay-general/src/pii/convert.rs#L450) if you prefer to have it in another file. Another benefit of using insta snapshots is that you can run `cargo insta review` and update them automatically, more convenient than modifying keys manually.

iker-barriocanal (approved):

PR #2174, opened 2023-06-02
feat(spans): Extract span.action when no db.operation provided

iker-barriocanal:
We were extracting the `span.action` tag under the assumption `db.operation` was always in the payload. Sometimes this is not the case, so we extract it with a Regex as a workaround.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -696,6 +696,9 @@ fn extract_span_metrics(
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -689,6 +696,13 @@ fn extract_span_metrics(
     Ok(())
 }
 
+static SQL_ACTION_EXTRACTOR_REGEX: Lazy<Regex> =
```
olksdr, line 702: Maybe a short doc string will be good here.

olksdr (approved):

PR #2176, opened 2023-06-02
feat(spans): Extract `span.group` tag

iker-barriocanal:
Adds `span.group` tag to span metrics. The span group is calculated by taking the first 16 characters of the MD5 of the sanitized span description; thus, if the span description isn't sanitized, we don't extract it.

In order to calculate the MD5 hash, we introduce [the MD5 crate](https://github.com/stainless-steel/md5) as a new dependency ([licensed](https://github.com/stainless-steel/md5/blob/master/LICENSE.md) under both Apache 2.0 and MIT).

#skip-changelog


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -562,6 +563,12 @@ fn extract_span_metrics(
                     "span.description".to_owned(),
                     scrubbed_description.to_owned(),
                 );
+
+                let mut span_group = format!("{:?}", md5::compute(scrubbed_description));
```
jjbayer, line 582: Doesn't taking the _scrubbed_ description instead of the raw description make this group hash inconsistent with the group hash computed in sentry? Or does that also use the scrubbed version now?

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -562,6 +563,12 @@ fn extract_span_metrics(
...
+                if span_group.len() > 16 {
+                    span_group = span_group[..16].to_owned();
+                }
```
jjbayer: Both versions can panic in theory, but this should be safe because the hash won't contain non-ascii chars.

```suggestion
                span_group.truncate(16);
```

jjbayer (approved):

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: Span grouping takes the sanitized scrubbed description to apply the hashing algorithm in Sentry too. Note that with evolving span description rules, the hash could also vary.

iker-barriocanal:
Apparently, something went wrong when merging https://github.com/getsentry/relay/pull/2183 to master and master CI is now red. [864c684](https://github.com/getsentry/relay/pull/2176/commits/864c68424c60e142ab92554a566995a05dde7f53) in this PR should fix it.

PR #2178, opened 2023-06-02
Feat/profile rate limit buckets

jjbayer:
- [x] Remove profile tags and emit outcomes
- [ ] Count in redis.


jjbayer:
There is no product requirement for this and it adds a lot of complexity. We can always reopen when needed.

PR #2179, opened 2023-06-02
feat(spans): Extract `transaction.method` tag

iker-barriocanal:
Extract `transaction.method` tag with a regex from the transaction name, and add the tag to all span metrics.

#skip-changelog


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -719,6 +726,16 @@ fn sql_table_from_query(query: &str) -> Option<&str> {
     extract_captured_substring(query, &SQL_TABLE_EXTRACTOR_REGEX)
 }
 
+/// Regex with a capture group to extract the HTTP method from a string.
+static HTTP_METHOD_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r#"(?i)^(?P<method>(GET|HEAD|POST|PUT|DELETE|CONNECT|OPTIONS|TRACE|PATCH))\s"#)
```
iker-barriocanal, line 731: I've taken the list of these methods from [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods).

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -873,7 +890,7 @@ mod tests {
             "release": "1.2.3",
             "dist": "foo ",
             "environment": "fake_environment",
-            "transaction": "mytransaction",
+            "transaction": "GET /api/:version/users/",
```
iker-barriocanal, line 893: Note that these first metrics are transaction metrics, so `transaction.method` is expected to not exist in those.

olksdr (approved):

PR #2180, opened 2023-06-04
fix(kafka): Update default topic for transaction metrics

jjbayer:
In sentry, the default topic for transaction a.k.a. performance metrics is defined as

```python
KAFKA_INGEST_PERFORMANCE_METRICS = "ingest-performance-metrics"
```

([source](https://github.com/getsentry/sentry/blob/75833f69cea56d4d0f7c7dbde6b8026b1110376f/src/sentry/conf/server.py#L2925))

But Relay still sends all metrics to the same topic by default.

During a local end-to-end test, I noticed that the metrics indexer crashes when it receives buckets from multiple use cases on the same topic:

```python
            raise ValueError(
                f"The set of use_case_ids: {use_case_ids} maps to multiple metric path keys"
            )
 ```

([source](https://github.com/getsentry/sentry/blob/75833f69cea56d4d0f7c7dbde6b8026b1110376f/src/sentry/sentry_metrics/indexer/postgres/postgres_v2.py#L288-L290))

To make development setup easier, use the same default topic in Relay as in Sentry.

### DIFF
--- a/relay-kafka/src/config.rs
+++ b/relay-kafka/src/config.rs
@@ -89,7 +89,7 @@ pub struct TopicAssignments {
     /// Topic name for metrics extracted from sessions. Defaults to the assignment of `metrics`.
     pub metrics_sessions: Option<TopicAssignment>,
     /// Topic name for metrics extracted from transactions. Defaults to the assignment of `metrics`.
-    pub metrics_transactions: Option<TopicAssignment>,
+    pub metrics_transactions: TopicAssignment,
     /// Stacktrace topic name
     pub profiles: TopicAssignment,
     /// Replay Events topic name.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: relay-kafka/src/config.rs [hunk 1]
```diff
@@ -134,7 +132,7 @@ impl Default for TopicAssignments {
             sessions: "ingest-sessions".to_owned().into(),
             metrics: "ingest-metrics".to_owned().into(),
             metrics_sessions: None,
-            metrics_transactions: None,
+            metrics_transactions: "ingest-performance-metrics".to_owned().into(),
```
olksdr, line 135: This kind of makes sense. But I still would prefer that we do not set this up by default, and just provide the correct config. 
Since `Sentry` backend have such default already, I guess it's fine to have those too.

olksdr (approved):

iker-barriocanal (approved):

relay-kafka/src/config.rs
jjbayer: Confirmed offline that session and transaction metrics should never go into the same topic these days, not even in self hosted or dev environments. Therefore, I think we have to change the default to get a working out-of-the-box experience.

jjbayer: Will add a changelog entry though, in case anyone encounters problems with this change.

PR #2181, opened 2023-06-05
feat(spans): Extract `span.category` tag

iker-barriocanal:
Extracts `span.category` tag in spans from its operation. The mapping is available in [Sentry develop docs](https://develop.sentry.dev/sdk/performance/span-operations/). I decided to match the beginning of strings instead of hardcoding a list for simplicity and to match similar operations, we can revisit this decision at a later point if this causes any issues.

#skip-changelog


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -913,6 +1042,15 @@ mod tests {
                     "timestamp": 1597976302.0000000,
                     "trace_id": "ff62a8b040f340bda5d830223def1d81"
                 },
+                {
+                    "description": "<SomeUiRendering>",
+                    "op": "ui.react.render",
```
iker-barriocanal, line 1097: The negative test is the span right above this one. I didn't reuse it to have a negative test, and because changing the op (adding the `ui.` prefix) extracts a measurement less. We should revisit this edge case when we move to onlyspans.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -736,6 +740,131 @@ fn extract_captured_substring<'a>(string: &'a str, pattern: &'a Lazy<Regex>) ->
...
+            "ui.react"
+        } else if op.starts_with("ui.vue") {
+            "ui.vue"
+        } else if op.starts_with("ui.svelte") {
+            "ui.svelte"
+        } else if op.starts_with("ui.angular") {
+            "ui.angular"
+        } else if op.starts_with("ui.ember") {
+            "ui.ember"
+        }
```
olksdr: This can never happen, since on the line `765` you will have a match with `ui` only.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -736,6 +740,131 @@ fn extract_captured_substring<'a>(string: &'a str, pattern: &'a Lazy<Regex>) ->
...
+        // Serverless
+        // `http*` marked in Browser
+        // `grpc*` marked in Web server
+        else if op.starts_with("function.gpc") {
+            "function.grpc"
+        } else if op.starts_with("function.aws") {
+            "function.aws"
+        } else if op.starts_with("function.azure") {
+            "function.azure"
+        }
```
olksdr: Same here, on line `750` it will match the `function` first.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -736,6 +740,131 @@ fn extract_captured_substring<'a>(string: &'a str, pattern: &'a Lazy<Regex>) ->
     None
 }
 
+/// Returns the category of a span from its operation. The mapping is available in:
+/// <https://develop.sentry.dev/sdk/performance/span-operations/>
+fn span_op_to_category(op: &str) -> Option<&str> {
+    Some({
```
olksdr: It's not really readable, maybe better would be to assign these `if`s to the variable and return it at the end, like 

```rust

  let category = if op.starts_with("mark") {
...

  // at the end of the function 
  Some(category)
}
```

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -736,6 +740,131 @@ fn extract_captured_substring<'a>(string: &'a str, pattern: &'a Lazy<Regex>) ->
...
+            "view"
+        } else if op.starts_with("template") {
+            "template"
+        } else if op.starts_with("event") {
+            "event"
+        } else if op.starts_with("function.nextjs") {
+            "function.nextjs"
+        } else if op.starts_with("function.remix") {
+            "function.remix"
+        } else if op.starts_with("function") {
```
olksdr: you have already `function` on the line `750` which will get caught first

relay-server/src/metrics_extraction/transactions/mod.rs
olksdr: you also might want to do it a bit differently, make the inner `if`s, something like 
```rust
// e.g. first match function and only then you can check if the function match something else, otherwise return `function`
else if op.starts_with("function") {
       if op.starts_with("function.gpc") {
            "function.grpc"
        } else if op.starts_with("function.aws") {
            "function.aws"
        } else if op.starts_with("function.azure") {
            "function.azure"
        } else {
            "function"
       }
}

```

what do you think?

olksdr (changes requested):
There are few things I think should be fixed first.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: Nice! It also discovered a bug (see [ae68c82](https://github.com/getsentry/relay/pull/2181/commits/ae68c82364b7dacaf6fd7fe37ebbf54660a698b2)).

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -736,6 +740,138 @@ fn extract_captured_substring<'a>(string: &'a str, pattern: &'a Lazy<Regex>) ->
...
+        // Messages / queues
+        else if op.starts_with("topic") {
+            "topic"
+        } else if op.starts_with("queue") {
+            "queue"
+        }
+        //
+        // Unknown
+        else {
+            ""
```
olksdr: I think you still can keep `return None` here, it would be fine.  so you don't have to do another check at the end. 
And at the end just return `Some(category)`.

olksdr (approved):
approving to unblock fast

PR #2184, opened 2023-06-05
feat(ci): Add CI-check for pop-deploys 

TBS1996:
This should prevent deploying pop-relays before it has passed all the CI tests.

#skip-changelog


relay/src/cli.rs
```diff
@@ -110,7 +110,7 @@ pub fn extract_config_env_vars() -> OverridableConfig {
         id: env::var("RELAY_ID").ok(),
         public_key: env::var("RELAY_PUBLIC_KEY").ok(),
         secret_key: env::var("RELAY_SECRET_KEY").ok(),
-        outcome_source: None, //already extracted in params
```
iker-barriocanal: Why this change?

gocd/pipelines/relay-pop.yaml
```diff
@@ -23,7 +23,28 @@ pipelines:
...
+            type: manual
+          fetch_materials: true
+
+          jobs:
+            checks:
+              environment_variables:
+                # Required for checkruns.
+                GITHUB_TOKEN: "{{SECRET:[devinfra-github][token]}}"
+              timeout: 1800
+              elastic_profile_id: relay
```
iker-barriocanal: Should this be the same profile as processing relays?

relay/src/cli.rs
TBS1996: youre quick! just an artefact from accidentally editing master and failing to revert properly, ive fixed it

gocd/pipelines/relay-pop.yaml
TBS1996: It's now `relay-pop` to align with the other pop jobs

TBS1996: sorry for the noise! I should have made it draft

jjbayer (approved):
As discussed offline, another approach would be to make sure that PoPs can only deploy versions that have already been deployed to processing relays. But this should work too!

PR #2185, opened 2023-06-05
Revert "fix(spans): Extract span http_status tag from span tags (#2183)"

iker-barriocanal:
Creates too high cardinality.

#skip-changelog


jjbayer (approved):

iker-barriocanal:
I double-checked this revert PR is ok.

PR #2189, opened 2023-06-06
feat(spans): Extract `release` from the transaction to the span's databag

iker-barriocanal:
Currently, we don't extract the `release` in Relay and it's extracted in [Snuba](https://github.com/getsentry/snuba/blob/17d6322c64ca10ddd3c8c01c10c2ea6c96783a11/snuba/datasets/processors/spans_processor.py#L178) instead. This PR moves us one step closer to removing data extraction from Snuba and centralizing it in Relay.

~I'm not sure if metric extraction in Snuba runs before or after checking the cardinality limits, so I'll monitor closely the deployment of this PR.~ This is now extracted to `span.data` instead of metrics, so there's no impact on metrics cardinality.

#skip-changelog


jjbayer (approved):
The cardinality limiter runs in sentry before the data reach snuba, so this could definitely push it over the threshold again (if it's not already above the threshold).

olksdr (approved):

nikhars:
@iker-barriocanal - Is there a way to not extract metrics for these tags but extract the data and send it in spans data bag. That is the intended outcome I was looking for.

Basically, a tag is at the top level transaction. But it needs to be exploded to all individual spans beneath it. We don't need metrics for this tag. Tag only needs to be propagated to all children spans in the spans data bag.

If that is hard to do, I would say do not proceed with this. Snuba will continue to do what it does today.

olksdr (approved):

PR #2191, opened 2023-06-06
ref(metrics-extraction): Split span and transaction metric extraction

iker-barriocanal:
We've been plugging span extraction in `relay-server/src/metrics_extraction/transactions/mod.rs`. As the module's scope and responsibilities increase, splitting span and transaction responsibilities into two different modules helps to better understand the code and make more targeted tests, for example.

**Note**: this PR doesn't contain any functional changes, only a refactor moving code around.

Summary of changes, in the `relay-server` crate:
- Create module `metrics_extraction/performance`, with two submodules inside: `spans` and `transactions`.
- Delete former `metrics_extraction/transactions/mod.rs`. The contents are split into the above-mentioned `spans` and `transactions` modules.
- Shared functions between `metrics_extraction/performance/spans` and `metrics_extraction/performance/transactions` have been moved to `metrics_extraction/performance/utils.rs`. I'm not opinionated on the name of this file and am open to changing it.
- A single test snapshot has been moved, without any changes besides the path rename.
- Other minor import updates to reflect the new paths.

This PR is a first step to achieving a better responsibility split. More PRs will follow to improve this, such as splitting tests. I'm not including these in the current PR to minimize the number of changes and reduce risk.

#skip-changelog


relay-server/src/metrics_extraction/performance/mod.rs
olksdr, line 1: What the reason to create so deep module structure? 
Why don't just use `transactions` and `spans` modules in the root of `metrics_extractions` which makes perfect sense to me, and we do not create those very long module paths.

relay-server/src/actors/processor.rs
```diff
@@ -63,9 +63,9 @@ use crate::actors::project_cache::ProjectCache;
 use crate::actors::upstream::{SendRequest, UpstreamRelay};
 use crate::envelope::{AttachmentType, ContentType, Envelope, Item, ItemType};
 use crate::extractors::RequestMeta;
+use crate::metrics_extraction::performance::transactions::extract_transaction_metrics;
```
olksdr: `performance` here does not work very well. IMO we do not need this module at all

olksdr (changes requested):
Could we remove such deep-nested module structure? 

> More PRs will follow to improve this, such as splitting tests. I'm not including these in the current PR to minimize the number of changes and reduce risk.

You are already moving code around, what the risk would that be to also move the tests into appropriate modules?

relay-server/src/metrics_extraction/performance/mod.rs
iker-barriocanal: I created it to have shared functions that `sessions` don't need to access, but I agree it's too much complexity. Addressed in [0389e09](https://github.com/getsentry/relay/pull/2191/commits/0389e09e806986fc1f12091655a79172e4971fbc).

iker-barriocanal:
>You are already moving code around, what the risk would that be to also move the tests into appropriate modules?

In this case, it's not about moving tests, but about splitting the test that currently verifies both transaction and span metric extraction. Addressing this means splitting the test, and I want to avoid adding that in this PR because it's harder to identify missing cases (good luck finding that in the diff). Not strongly opinionated though, so I'm open to also addressing that in this PR. I think this approach reduces some risk. What do you think, does this sound reasonable?

olksdr (approved):
lgtm!

If taking caring of the tests is easier to do in the followup, i'm also ok with that. 

Thanks for taking care of this!

PR #2193, opened 2023-06-07
ref(spans): Introduce opinionated casing in span metrics

iker-barriocanal:
In the past, consistent and logical input in certain parameters was assumed. For example, an organization could send `SELECT` or `select` consistently, but combining them with other alternatives such as `sELEct` was not considered. The combination of these cases contributes to higher cardinality and degrades the product experience. In an attempt to improve both, this PR introduces opinionated casing for some fields. This decision is not locked-in and may change in the future as needs evolve.

The following cases are enforced:
- Transaction method, uppercase. Example: `GET`.
- Transaction operation, lowercase. Example: `measure`.
- Span operation, lowercase. Example: `ui.react.render`.
- Span action, uppercase for both HTTP and database spans. Examples: `POST`, `SELECT`.
- Span domain, lowercase for both HTTP and database spans. Examples: `mydomain`, `tablename`.
- Span system, lowercase. Example: `mysql`.

The following are NOT enforced:
- Environment, to reuse the same as in transaction data.
- Transaction name, since casing is important.
- Span description, since casing is important.
- Span group, the MD5 hash remains unchanged.

#skip-changelog


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -1404,8 +1410,8 @@ mod tests {
                     "os.name": "Windows",
                     "platform": "javascript",
                     "release": "1.2.3",
-                    "transaction": "GET /api/:version/users/",
-                    "transaction.op": "myop",
+                    "transaction": "gEt /api/:version/users/",
+                    "transaction.op": "mYOp",
```
iker-barriocanal, line 1414: Note that this test contains mixed metrics from transactions and spans. This PR only changes span metrics.

olksdr (commented):
It looks good, but why did you opt into use `to_ascii_` equivalents of the lower- and uppercase functions?  I do not see the reason or benefits in doing so. 
Could you, please, explain?

iker-barriocanal:
@olksdr you're right, there's no relevant reason to do so. Updated in [6ce20af](https://github.com/getsentry/relay/pull/2193/commits/6ce20af3b12a833434ebd20d97d0ecc0ab3cf41d). Thanks for pointing it out!

olksdr (approved):

PR #2195, opened 2023-06-09
feat(spans): Extract `span.status_code` tag

iker-barriocanal:
This tag was reverted in https://github.com/getsentry/relay/commit/ccd19b5b7b67b8ebb64a5560df565e89ac62bea4. This PR reintroduces the tag.

#skip-changelog

PR #2196, opened 2023-06-09
feat(spans): Extract transaction's HTTP status code into span tags as `http.status_code`

iker-barriocanal:
Each span has its own status code, but also information on the status code of the parent transaction they belong to. We're not looking for perfect results here and `extract_http_status_code` provides good enough values.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/spans/mod.rs
+++ b/relay-server/src/metrics_extraction/spans/mod.rs
@@ -1,5 +1,5 @@
-use crate::metrics_extraction::transactions::extract_http_status_code;
 use crate::metrics_extraction::transactions::types::ExtractMetricsError;
+use crate::metrics_extraction::utils::extract_http_status_code;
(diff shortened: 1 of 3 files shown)


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -1,3 +1,4 @@
+use crate::metrics_extraction::transactions::extract_http_status_code;
```
iker-barriocanal: The refactor of this method (moving it to `utils` module) is done in https://github.com/getsentry/relay/pull/2195, which should be merged before this PR. The reason I'm not stacking PRs is because it causes quite a lot of conflicts, that I don't consider worth resolving for a one-liner update.

olksdr (approved):

iker-barriocanal:
This is planned to be deployed on Monday, June 12.

PR #2197, opened 2023-06-09
feat(spans): Extract transaction `user` into span's databag

iker-barriocanal:
Follow-up to https://github.com/getsentry/relay/pull/2189. `user` is the last remaining field.

#skip-changelog


olksdr (approved):
:shipit:

PR #2198, opened 2023-06-09
test(document-pii): Fix flaky test

iker-barriocanal:
This test has resulted flaky in CI only on ubuntu machines (OSX and Windows seem ok), see [latest example](https://github.com/getsentry/relay/actions/runs/5204020888/jobs/9387669570#step:7:61). Sorting the file paths should remove the flakiness.

#skip-changelog

### DIFF
--- a/tools/document-pii/src/main.rs
+++ b/tools/document-pii/src/main.rs
@@ -176,7 +176,8 @@ mod tests {
     #[test]
     fn test_find_rs_files() {
         let rust_crate = PathBuf::from_slash(RUST_TEST_CRATE);
-        let rust_file_paths = find_rs_files(&rust_crate);
+        let mut rust_file_paths = find_rs_files(&rust_crate);
+        rust_file_paths.sort();
         insta::assert_debug_snapshot!(rust_file_paths);
     }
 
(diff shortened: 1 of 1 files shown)


tools/document-pii/src/main.rs
```diff
@@ -176,7 +176,8 @@ mod tests {
     #[test]
     fn test_find_rs_files() {
         let rust_crate = PathBuf::from_slash(RUST_TEST_CRATE);
-        let rust_file_paths = find_rs_files(&rust_crate);
+        let mut rust_file_paths = find_rs_files(&rust_crate);
+        rust_file_paths.sort_unstable();
```
iker-barriocanal, line 180: From [docs](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort):

>When applicable, unstable sorting is preferred because it is generally faster than stable sorting and it doesn’t allocate auxiliary memory. See [sort_unstable](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable).

Since there shouldn't be equal elements (it's files from the file system, in the end), I think unstable sorting is a better trade-off in this case.

olksdr (approved):

PR #2199, opened 2023-06-09
fix(ci): Auto-deploy pop after check

TBS1996:
#skip-changelog

auto-deploy pop if checks succeed


TBS1996:
close because of identical pr: #2200

PR #2201, opened 2023-06-12
instr(metrics): Add values to invalid metric errors

jjbayer:
When we fail to add a metric to the aggregator, log the value that caused the problem.

This should make it easier to review related error logs.

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregation.rs
+++ b/relay-metrics/src/aggregation.rs
@@ -873,19 +873,19 @@ impl From<AggregateMetricsErrorKind> for AggregateMetricsError {
 #[allow(clippy::enum_variant_names)]
 enum AggregateMetricsErrorKind {
     /// A metric bucket had invalid characters in the metric name.
-    #[error("found invalid characters")]
+    #[error("found invalid characters: {0}")]
     InvalidCharacters(String),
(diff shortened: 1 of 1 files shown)


relay-metrics/src/aggregation.rs
```diff
@@ -874,19 +874,19 @@ impl From<AggregateMetricsErrorKind> for AggregateMetricsError {
 enum AggregateMetricsErrorKind {
     /// A metric bucket had invalid characters in the metric name.
     #[error("found invalid characters")]
-    InvalidCharacters,
+    InvalidCharacters(String),
```
olksdr, line 877: in each of this cases you add, you might want to update also the macro like `#[error("found invalid characters: ({0})")]` to actually print the error with the wrong value. 

What do you think?

olksdr (approved):
I have one suggestion, otherwise LGTM

iker-barriocanal (approved):

relay-metrics/src/aggregation.rs
iker-barriocanal: I was wondering how the new error message would be printed.

jjbayer: Good one, that's what I actually wanted.

PR #2202, opened 2023-06-12
feat(general): Make selectorspec forward compatible

TBS1996:
#1751 

Should make selectorspec forward compatible by adding a fall-back enum variant

#skip-changelog

### DIFF
--- a/relay-general/src/processor/attrs.rs
+++ b/relay-general/src/processor/attrs.rs
@@ -717,7 +717,10 @@ impl<'a> Path<'a> {
             SelectorSpec::And(ref xs) => xs.iter().all(|x| self.matches_selector(x)),
             SelectorSpec::Or(ref xs) => xs.iter().any(|x| self.matches_selector(x)),
             SelectorSpec::Not(ref x) => !self.matches_selector(x),
-            SelectorSpec::Other(_) => false,
+            SelectorSpec::Other(_) => {
+                relay_log::warn!("Incoming selector is not supported");
+                false
+            }
         }
     }
 }
--- a/relay-general/src/processor/attrs.rs
+++ b/relay-general/src/processor/attrs.rs
@@ -717,6 +717,7 @@ impl<'a> Path<'a> {
             SelectorSpec::And(ref xs) => xs.iter().all(|x| self.matches_selector(x)),
             SelectorSpec::Or(ref xs) => xs.iter().any(|x| self.matches_selector(x)),
             SelectorSpec::Not(ref x) => !self.matches_selector(x),
+            SelectorSpec::Other(_) => false,
         }
     }
 }
--- a/relay-general/src/processor/selector.rs
+++ b/relay-general/src/processor/selector.rs
@@ -138,22 +138,24 @@ pub enum SelectorSpec {
     Or(Vec<SelectorSpec>),
     Not(Box<SelectorSpec>),
     Path(Vec<SelectorPathItem>),
+    Other(String),
 }
 
 impl fmt::Display for SelectorSpec {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
+        match self {
             SelectorSpec::And(ref xs) => {
                 for (idx, x) in xs.iter().enumerate() {
                     if idx > 0 {
                         write!(f, " && ")?;
                     }
 
-                    let needs_parens = match *x {
+                    let needs_parens = match x {
                         SelectorSpec::And(_) => false,
                         SelectorSpec::Or(_) => true,
                         SelectorSpec::Not(_) => false,
                         SelectorSpec::Path(_) => false,
+                        SelectorSpec::Other(_) => true,
                     };
 
                     if needs_parens {
@@ -180,6 +182,7 @@ impl fmt::Display for SelectorSpec {
                     SelectorSpec::And(_) => true,
                     SelectorSpec::Or(_) => true,
                     SelectorSpec::Not(_) => true,
+                    SelectorSpec::Other(_) => true,
                     SelectorSpec::Path(_) => false,
                 };
 
@@ -197,6 +200,7 @@ impl fmt::Display for SelectorSpec {
                     write!(f, "{item}")?;
                 }
             }
+            SelectorSpec::Other(s) => write!(f, "{s}")?,
         }
         Ok(())
     }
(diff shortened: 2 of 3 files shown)


TBS1996:
i see olek used relay_log::warn when the catch-all variant is used https://github.com/getsentry/relay/pull/1859/files

so perhaps I should do it too

relay-general/src/processor/selector.rs
```diff
@@ -138,22 +138,24 @@ pub enum SelectorSpec {
...
-        match *self {
+        match self {
             SelectorSpec::And(ref xs) => {
                 for (idx, x) in xs.iter().enumerate() {
                     if idx > 0 {
                         write!(f, " && ")?;
                     }
 
-                    let needs_parens = match *x {
+                    let needs_parens = match x {
```
TBS1996, line 153: I got borrow-checking errors if I didn't do this, something about the string in Other() being moved out, I don't know exactly how this works

relay-general/src/processor/selector.rs
```diff
@@ -138,22 +138,24 @@ pub enum SelectorSpec {
...
                         SelectorSpec::And(_) => false,
                         SelectorSpec::Or(_) => true,
                         SelectorSpec::Not(_) => false,
                         SelectorSpec::Path(_) => false,
+                        SelectorSpec::Other(_) => true,
```
TBS1996, line 158: figured it's better with too many parens than too few

relay-general/src/processor/attrs.rs
```diff
@@ -717,6 +717,7 @@ impl<'a> Path<'a> {
             SelectorSpec::And(ref xs) => xs.iter().all(|x| self.matches_selector(x)),
             SelectorSpec::Or(ref xs) => xs.iter().any(|x| self.matches_selector(x)),
             SelectorSpec::Not(ref x) => !self.matches_selector(x),
+            SelectorSpec::Other(_) => false,
```
TBS1996: I'm a bit unsure on this one, also, should I add the relay_warn log here, in the way that olek did here: https://github.com/getsentry/relay/pull/1859/files

relay-general/src/processor/selector.rs
```diff
@@ -294,10 +298,7 @@ fn handle_selector(pair: Pair<Rule>) -> Result<SelectorSpec, InvalidSelectorErro
         Rule::NotSelector => Ok(SelectorSpec::Not(Box::new(handle_selector(
             pair.into_inner().next().unwrap(),
         )?))),
-        rule => Err(InvalidSelectorError::UnexpectedToken(
-            format!("{rule:?}"),
-            "a selector",
-        )),
+        _ => Ok(SelectorSpec::Other(pair.as_str().to_string())),
```
TBS1996, line 298: so far unable to actually find a case where this code gets executed with unit tests, using the from_str method. 

it either gets parsed into a `SelectorSpec::Path`, or it returns `InvalidSelectorError::ParseError`

relay-general/src/processor/selector.rs
```diff
@@ -221,15 +225,12 @@ impl FromStr for SelectorSpec {
...
-                .map_err(|e| InvalidSelectorError::ParseError(Box::new(e)))?
-                .next()
-                .unwrap()
-                .into_inner()
-                .next()
-                .unwrap(),
-        )
+        let rule_pair = match SelectorParser::parse(Rule::RootSelector, s) {
+            Ok(mut rule_pairs) => rule_pairs.next().unwrap().into_inner().next().unwrap(),
+            Err(e) => return Ok(SelectorSpec::Other(e.to_string())),
```
TBS1996, line 230: I need some input on whether this logic makes sense at all,  as I wrote [here](https://github.com/getsentry/relay/pull/2202/files#r1226697680) i was unable to catch the invalid strings in the `handle_selector` function, so I tried this instead.

relay-general/src/processor/attrs.rs
iker-barriocanal: I think this is a good idea!

relay-general/src/processor/selector.rs
iker-barriocanal: What is the error, could you maybe add it here?

relay-general/src/processor/selector.rs
iker-barriocanal: I think having parenthesis is required if we think about forward compatibility.

relay-general/src/processor/selector.rs
iker-barriocanal: This logic seems different to me. Previously, errors were mapped to `InvalidSelectorError`, and in the new diff we are parsing the errors into an `Ok`.

relay-general/src/processor/selector.rs
iker-barriocanal: If it can return an `InvalidSeelctorError`, it seems this branch would catch it.

iker-barriocanal (changes requested):
The `Result` enum is useful because we can identify whether it's an error or not. If we convert the errors into `Ok` variants, we no longer have the chance to distinguish them.

To validate this logic works as intended, what do you think about adding some tests? They could help to ensure forward compatibility exists.

relay-general/src/processor/selector.rs
TBS1996: sure, here: 

```
error[E0507]: cannot move out of `self` as enum variant `Other` which is behind a shared reference
   --> relay-general/src/processor/selector.rs:146:15
    |
146 |         match *self {
    |               ^^^^^ help: consider borrowing here: `&*self`
...
203 |            SelectorSpec::Other(s) => write!(f, "{s}")?,
    |                                 -
    |                                 |
    |                                 data moved here
    |                                 move occurs because `s` has type `std::string::String`, which does not implement the `Copy` trait
 ```


i dont understand why it was dereferenced in the first place

File: relay-general/src/processor/selector.rs (hunk 5 above)
Comment (TBS1996) [reply relay-general/src/processor/selector.rs:230]: yep, but isn't this somewhat what we want? like if a relay is outdated, previously it'd make an error, but with this, it'll instead parse it into the other() variant ?

File: relay-general/src/processor/selector.rs (hunk 4 above)
Comment (TBS1996) [reply relay-general/src/processor/selector.rs:298]: I'm not sure what you mean here exactly 

About this `InvalidSelectError`, among the variants:

```
pub enum InvalidSelectorError {
    #[error("invalid selector: deep wildcard used more than once")]
    InvalidDeepWildcard,

    #[error("invalid selector: wildcard must be part of a path")]
    InvalidWildcard,

    #[error("invalid selector: {0}")]
    ParseError(Box<Error<Rule>>),

    #[error("invalid selector: invalid index")]
    InvalidIndex,

    #[error("invalid selector: unknown value")]
    UnknownType,

    #[error("parser bug: consumed {0} (expected {1})")]
    UnexpectedToken(String, &'static str),
}
```

the ones that to me seems relevant for forward compatiblity are: `ParseError`, `UnknownType`, and `UnexpextedToken`, if this is right we could remove them and instead put them all into `Ok(Other(...))`, or maybe even somehow also pass along the errors themselves?

TBS1996:
closing on reccomendation from jjbayer

PR #2203, opened 2023-06-12
chore(protocol): Add missing py/changelog entry for lock property

romtsn:
_#skip-changelog_


jjbayer (approved):
Python 3.8 tests are failing on multiple PRs, we're looking into it.

PR #2204, opened 2023-06-12
build(py): Move pytest.ini to tests directory

jjbayer:
Our CI uses an unpinned version of `pytest`. pytest 7.3.2 includes the following fix:

> [#10987](https://github.com/pytest-dev/pytest/issues/10987): [testpaths](https://docs.pytest.org/en/latest/reference/reference.html#confval-testpaths) is now honored to load root conftests.

([changelog](https://docs.pytest.org/en/latest/changelog.html))

In our case, this led to an import error, because the CI job running tests in `py/` does not install the requirements needed to configure our python integration tests. Moving the pytest.ini file to `tests/` fixes the issue.

#skip-changelog


olksdr (approved):
lgtm

iker-barriocanal (approved):
Thanks for doing this!

PR #2205, opened 2023-06-12
ref(spans): Move span metrics to spans namespace

iker-barriocanal:
Since the span namespace for metrics now exists, we're moving these metrics. The new strings were introduced in https://github.com/getsentry/sentry/pull/50718.

We also take the chance to remove the `spans.` prefix in MRIs. Span metric tags will still have the `spans.` prefix, since there already are some tags without the prefix created by some organizations.

#skip-changelog


olksdr (approved):
makes sense.

PR #2209, opened 2023-06-13
feat(general): Make selectorpathitem fwd compatible

TBS1996:
#1751 

#skip-changelog

### DIFF
--- a/relay-general/src/processor/selector.rs
+++ b/relay-general/src/processor/selector.rs
@@ -47,11 +47,12 @@ pub enum SelectorPathItem {
     Key(String),
     Wildcard,
     DeepWildcard,
+    Other(String),
 }
 
 impl fmt::Display for SelectorPathItem {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
+        match self {
             SelectorPathItem::Type(ty) => write!(f, "${ty}"),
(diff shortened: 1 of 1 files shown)


relay-general/src/processor/selector.rs
```diff
@@ -128,6 +130,7 @@ impl SelectorPathItem {
                 .key()
                 .map(|k| k.to_lowercase() == key.to_lowercase())
                 .unwrap_or(false),
+            (SelectorPathItem::Other(_), _) => false,
```
TBS1996: If there is an unknown selectorpathitem, should we assume it's pii = true or false ?

relay-general/src/processor/selector.rs
```diff
@@ -47,11 +47,12 @@ pub enum SelectorPathItem {
     Key(String),
     Wildcard,
     DeepWildcard,
+    Other(String),
 }
 
 impl fmt::Display for SelectorPathItem {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
+        match self {
```
TBS1996, line 55: im getting borrow check issues if i try to display a string when matching on '*self'

TBS1996:
This item doesn't seem to be implementing serialization/deserialization, isn't that a prerequisite for this change to be relevant?

TBS1996:
closing on suggestion from jjbayer, as it's only generated from code

PR #2210, opened 2023-06-13
feat(txname): Mark all URL transactions as sanitized

### LINKED ISSUE #2186: [metrics] Handling `transaction:<< unparameterized >>` better for new users
### Problem statement
We currently let through `unparameterized` still especially for new users since there is a warmup period before rules are built. 

>  at least 99% (TBD) of transaction names are parameterized when a customer first sets up Performance on all platforms 

While we are at a very high % for overall unparameterized sanitized, the majority of the remaining % unparameterized resides in our users first experience sending in transactions. We don't track user 'newness' in our metrics charts for transaction source transformation, but since the mechanism needs warming time it's likely that at that many of our browser frontends are seeing `<< unparameterized >>` as their first experience. Part of this is due to lack of integrations creating better transaction names in the sdk themselves, we also need to mitigate the current legitimate use case to improve the initial user experience since sdks adoption is quite slow and we don't have a perfect solution there either.

### Current state
This only applies for `transaction.source:url` and when the clusterer does not have rules built. 

```
start = [unparameterized]  // wait until clusterer is warmed up
[unparameterized] --brought cardinality down with clusterer --> [let_through]  
```

Before we make any decisions here, the first step should probably be improving the metrics. Adding `txNameReady` to the existing metrics so we can track what % are being let through before the 10h mark gets hit might be helpful since it's closer to 'new user' since it applies only before clusterer has run multiple times.

### Potential solutions
Sourced from discussion in slack.

- Add a set to allow through the first X transaction names. Doesn't need to be exact just needs to let through 
```
start = [let_through]   // Allow all new transactions in for accounts never seen by the clusterer
[let_through] --high cardinality detected--> [unparameterized]      // Add a check to limit an explosion of cardinality
[unparameterized] --brought cardinality down--> [let_through]
```
- Lower time to `txNameReady` (less clusterer runs required) or add more fine grain controls based on throughput. 
- Improve adoption of routers vs. raw urls (not relevant to this ticket but will have an impact on this %)

jjbayer:
In https://github.com/getsentry/relay/pull/2139, we transitioned to marking all URL transactions as "sanitized" once sentry provides the "ready" flag. However, this results in a bad onboarding experience for new projects because they still see `<< unparameterized >>` for the first ~10 hours.

This PR removes the "ready" condition, such that _all_ transactions with source URL will now get their transaction name added as a metric tag. Since we were already doing this for all but the newest projects, this should not add much cardinality after all. But we have to keep an eye on the cardinality limiter in the first few hours / days after deploying this.

Fixes https://github.com/getsentry/relay/issues/2186

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -24,8 +24,6 @@ use crate::types::{
 pub struct TransactionNameConfig<'r> {
     /// Rules for identifier replacement that were discovered by Sentry's transaction clusterer.
     pub rules: &'r [TransactionNameRule],
-    /// True if URL transactions should be marked as sanitized, even if there are no rules.
-    pub ready: bool,
 }
 
 /// Rejects transactions based on required fields.
@@ -471,18 +469,13 @@ impl Processor for TransactionsProcessor<'_> {
                 .set_value(Some("<unlabeled transaction>".to_owned()))
         }
 
-        // If the project is marked as 'ready', always set the transaction source to sanitized.
-        let mut mark_as_sanitized = self.name_config.ready;
-
         if matches!(
             event.get_transaction_source(),
             &TransactionSource::Url | &TransactionSource::Sanitized
         ) {
             // Normalize transaction names for URLs and Sanitized transaction sources.
             // This in addition to renaming rules can catch some high cardinality parts.
-            scrub_identifiers(&mut event.transaction)?.then(|| {
-                mark_as_sanitized = true;
-            });
+            scrub_identifiers(&mut event.transaction)?;
         }
 
         if !self.name_config.rules.is_empty() {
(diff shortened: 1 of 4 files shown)


relay-general/src/store/transactions/processor.rs
```diff
@@ -1808,14 +1806,7 @@ mod tests {
...
             &mut event,
-            &mut TransactionsProcessor::new(
-                TransactionNameConfig {
-                    rules: &[],
-                    ready: true,
-                },
-                false,
-                None,
-            ),
+            &mut TransactionsProcessor::new(TransactionNameConfig { rules: &[] }, false, None),
```
```suggestion
            &mut TransactionsProcessor::default(),
```
Any reason to not do this?

iker-barriocanal (approved):

olksdr (approved):

PR #2211, opened 2023-06-13
feat(spans): Forward already scrubbed SQL queries

iker-barriocanal:
In span metrics, we only add the span description tag if we managed to scrub the span description. In database spans, it's common for SDKs to already send scrubbed queries, and usually Relay won't do further scrubbing, resulting in the span description not being present. With this PR, Relay will check if the description looks like it has already been scrubbed, and extract the span description in these cases.

#skip-changelog


relay-general/src/store/transactions/processor.rs
```diff
@@ -449,6 +453,16 @@ fn scrub_identifiers_with_regex(
...
 }
 
+fn is_sql_query_scrubbed(query: &Annotated<String>) -> bool {
+    if let Some(is_match) = query
+        .value()
+        .map(|q| SQL_ALREADY_NORMALIZED_REGEX.is_match(q))
+    {
+        return is_match;
+    }
+    false
```
```suggestion
    query
        .value()
        .map_or(false, |q| SQL_ALREADY_NORMALIZED_REGEX.is_match(q))
```

jjbayer (approved):

olksdr (approved):

PR #2214, opened 2023-06-13
feat(inbound filters): Add `file:///` url to localhost inbound filter

### LINKED ISSUE #49519: Investigate server filter for `file://`
If we added a filter for urls starting with `file://` how may issues would be filtered and for how many projects / orgs.

This might be difficult to measure since its in tagstore.

lobsterkatie:
This adds a filter for events whose `url` value starts with `file://` to the existing localhost inbound filter, as such events similarly represent errors on pages opened locally.

Ref: https://github.com/getsentry/sentry/issues/49519

### DIFF
--- a/relay-filter/src/localhost.rs
+++ b/relay-filter/src/localhost.rs
@@ -25,6 +25,9 @@ pub fn matches(event: &Event) -> bool {
                 }
             }
         }
+        if url.scheme() == "file" {
+            return true;
+        }
     }
 
     false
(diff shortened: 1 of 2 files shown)


scttcper (approved):
lgtm

relay-filter/src/localhost.rs
```diff
@@ -25,6 +25,9 @@ pub fn matches(event: &Event) -> bool {
                 }
             }
         }
+        if url.as_str().starts_with("file://") {
```
jjbayer:
```suggestion
        if url.scheme() == "file" {
```

jjbayer (approved):

PR #2219, opened 2023-06-14
fix(spans): Use span's timestamps for span durations

iker-barriocanal:
To calculate a span's duration, we were using a transaction's start and end timestamps. This PR fixes this by using the span's timestamps.

#skip-changelog


jjbayer (approved):
Whoops

PR #2221, opened 2023-06-14
ref(spans): Extract transaction's HTTP method from transaction's request interface

iker-barriocanal:
It's common for transactions to have the HTTP method in the `request` interface and not in the transaction name. This PR considers that field on tag extraction, and prioritizes it over using a RegExp to extract it from the transaction name.

#skip-changelog


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -514,6 +514,9 @@ mod tests {
                     "name": "Windows"
                 }
             },
+            "request": {
+                "method": "post"
+            }
```
olksdr:
```suggestion
            },
```

olksdr (approved):

PR #2222, opened 2023-06-15
fix(docs): Fix filtering of "other"-fields

TBS1996:
There is a special case with the fields that have the attribute `additional_properties`, where we want to pass them through only if they also have the attribute `retain = "true"`. To support this, I refactored the tool, so that now it collects all of the attributes of all the matching pii-fields, and then filters it based on this logic. 

Also, the output is now reverted to json, and the output schema has a boolean for whether the field has `additional_properties` or not. 

Due to us now collecting all the attributes at the stage before the output, it makes this tool easier to use if we wanna document some other attribute in the future. 

#skip-changelog


tools/document-pii/src/snapshots/document_pii__tests__pii_retain_additional_properties_truth_table.snap
TBS1996, line 1: 3/4 fields are in the output, as predicted by the logic table in the "test_pii_retain_additional_properties_truth_table" function

tools/document-pii/src/pii_finder.rs
```diff
@@ -21,6 +21,29 @@ pub struct TypeAndField {
     pub field_ident: String,
 }
 
+#[derive(Ord, PartialOrd, PartialEq, Eq, Hash, Clone, Debug, Default)]
+pub struct FieldsWithAttribute {
+    pub type_and_fields: Vec<TypeAndField>,
+    pub attributes: BTreeMap<String, Option<String>>,
```
TBS1996, line 27: attribute are string -> Option\<String\> because some attributes are key-value pair while others just have the key. e.g `pii = "true"` vs `additional_properties`

tests/test_pii_docs/src/anothermod.rs
```diff
@@ -17,3 +17,14 @@ pub struct WeirdType {
...
+pub struct TruthTableTest {
+    #[metastructure(pii = "truth_table_test", retain = "true", additional_properties)]
+    foo: u32,
+    #[metastructure(pii = "truth_table_test", additional_properties)]
+    bar: u32,
+    #[metastructure(pii = "truth_table_test", retain = "true")]
+    baz: u32,
+    #[metastructure(pii = "truth_table_test")]
+    qux: u32,
+}
```
TBS1996, line 30: this represents all the combination with `retain = "true"` and `additional_properties`

tools/document-pii/src/main.rs
```diff
@@ -93,44 +85,62 @@ impl Cli {
...
+        };
+
+        output
+            .path
+            .push_str(&pii_type.type_and_fields[0].qualified_type_name);
+
+        let mut iter = pii_type.type_and_fields.iter().peekable();
+        while let Some(path) = iter.next() {
+            // If field has attribute "additional_properties" it means it's not a real field
+            // but represents unstrucutred data. So we remove it and pass the information as a boolean
```
jjbayer, line 123:
```suggestion
            // but represents unstructured data. So we remove it and pass the information as a boolean
```

tools/document-pii/src/main.rs
```diff
@@ -93,44 +85,62 @@ impl Cli {
...
+            // in order to properly document this fact in the docs.
+            if !(output.additional_properties && iter.peek().is_none()) {
+                output.path.push_str(&format!(".{}", path.field_ident));
+            }
```
jjbayer, line 127: Not adding anything to the path here could be problematic. If you have a struct like

```rust
struct Foo {
  outer: struct Outer {
    #[metastructure(additional_properties, retain = "true", pii ="true")]
    inner: Object<Value>,
  }
}
```

wouldn't this give you two identical paths?

```json
[
  {"path": "Foo.outer", "additional_properties": false},
  {"path": "Foo.outer", "additional_properties": true}
]
```

jjbayer: Might not be a problem if the client of this output knows how to interpret the "additional_properties" flag.

jjbayer (approved):

PR #2224, opened 2023-06-16
ref(spans): Introduce types for span metrics and tags

iker-barriocanal:
Span metrics and types were initially introduced as a hard-coded way of directly typing strings. To avoid mistyping in the future, this PR introduces types so that there's no need to write strings. The approach it's pretty much the same as with transaction metrics, but a bit simpler due to less strict requirements.

#skip-changelog


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -1,19 +1,24 @@
+use crate::metrics_extraction::spans::types::SpanMetric;
+use crate::metrics_extraction::spans::types::SpanTagKey;
```
olksdr: could we, please, combine these two into one import ?

olksdr (approved):

relay-server/src/metrics_extraction/spans/types.rs
```diff
@@ -0,0 +1,155 @@
...
+    },
+    Duration {
+        value: Duration,
+        tags: BTreeMap<SpanTagKey, String>,
+    },
+    ExclusiveTime {
+        value: f64,
+        tags: BTreeMap<SpanTagKey, String>,
+    },
+}
```
TBS1996, line 24: Just a thought.. would it make sense to do like this, since `tags` is shared ?

``` Rust
enum SpanMetricType{
  User(String),
  Duration(Duration),
  ExclusiveTime(f64),
}

struct SpanMetric{
  ty: SpanMetricType,
  tags: BTreeMap<SpanTagKey, String>
}

Comment (TBS1996) [reply relay-server/src/metrics_extraction/spans/types.rs:24]: I thought i'd try to reduce the duplicated logic in the match statements but maybe it ended up being more messy, so, your call

``` Rust

impl SpanMetricType {
    fn display(&self) -> &str {
        match self {
            Self::User(_) => "user",
            Self::Duration(_) => "duration",
            Self::ExclusiveTime(_) => "exclusive_time",
        }
    }
}

struct SpanMetric {
    ty: SpanMetricType,
    tags: BTreeMap<SpanTagKey, String>,
}

impl IntoMetric for SpanMetric {
    fn into_metric(self, timestamp: UnixTimestamp) -> Metric {
        let (unit, value) = match self.ty {
            SpanMetricType::User(user) => (MetricUnit::None, MetricValue::set_from_str(&user)),
            SpanMetricType::Duration(duration) => (
                MetricUnit::Duration(DurationUnit::MilliSecond),
                MetricValue::Distribution(relay_common::chrono_to_positive_millis(duration)),
            ),
            SpanMetricType::ExclusiveTime(exclusive_time) => (
                MetricUnit::Duration(DurationUnit::MilliSecond),
                MetricValue::Distribution(exclusive_time),
            ),
        };

        Metric::new_mri(
            MetricNamespace::Spans,
            self.ty.display(),
            unit,
            value,
            timestamp,
            span_tag_mapping_to_string_mapping(self.tags),
        )
    }
}
```

iker-barriocanal: Although this may work, I think the API of the current approach is simpler: one object vs two (in your suggestion). See how you'd interact with it:

```rs
// Current approach
SpanMetric::User {
    value: user_tag,
    tags: span_tags.clone()
}
.into_metric(timestamp);

// Suggested approach
SpanMetric {
    ty: SpanMetricType::User(user_tag),
    tags: span_tags.clone()
}
.into_metric(timestamp);
```

If/when we increase the number of metrics, we can revisit this and it may make sense at that point. I believe this simpler approach is better suited for now.

PR #2226, opened 2023-06-16
ref(spans): Allow localhost domains in domain tag

iker-barriocanal:
Allows `127.0.0.1` and `localhost` domains, to not be scrubbed for the `span.domain` tag. Note the span description will still be scrubbed.

#skip-changelog


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -474,6 +478,21 @@ fn normalize_domain(domain: &str, port: Option<&String>) -> Option<String> {
...
+fn normalized_domain_from_allowlist(domain: &str, port: Option<&String>) -> Option<String> {
+    for allowed in DOMAIN_ALLOW_LIST {
+        if domain == *allowed {
+            if let Some(p) = port {
+                return Some(format!("{}:{}", domain, p));
+            }
+            return Some(domain.to_owned());
+        }
+    }
+    None
```
olksdr: Let's do something shorter here:

```suggestion
        let domain = DOMAIN_ALLOW_LIST
        .iter()
        .find(|allowed| *allowed == &domain)?;
    port.map(|port| format!("{}:{}", domain, port))

```

olksdr (approved):
One suggestions, otherwise lgtm

relay-server/src/metrics_extraction/spans/mod.rs
olksdr: But I think you might just have to have a test before port mapping it is defined , otherwise it will return `None`.

so instead of `map` you might wanna do 
```rust
  if Some(port) = port {

    return Some(..)
  }

  return Some(domain)
}
```

Issue #2225, opened 2023-06-16
Relay appears to extract geo.country_code tag too early (before geo location)


RaduW:
### Environment

Latest version as of 16 June 2023

### Steps to Reproduce

1. Look at some transaction in discover ... observe that they have `user.geo.country_code` set to a non null value
2. Query for the transactions in Dashboards (uses MetricsEnhanced) and observe that they do not have anything set in the `geo.country_code` tag.

If you set manually the geo information on the client and send the transactions with the country_code already filled: 
```
scope.set_user({"id": 123, "email": "bob@sentry",
                            "geo": {"country_code": "AT", "city": "Wien", "subdivision": "1020", "region": "Wien"}})
``` 

And then look again in Dashboards you will find the country_code set.
So it appears that Relay correctly extracts the information if it is pre-existing but not if it fills it itself.
Perhaps the code that extracts the `geo.country_code` runs before the code that adds it based on the ip-address.

### Expected Result

Transaction metrics should have non empty `geo.country_code` tag.

### Actual Result

Unless you explicitly set the country_code manually in the transaction the metrics contain empty information for the tag
`geo.country_code`.

Note: I have talked with @olksdr and he has more context on this issue.

PR #2229, opened 2023-06-19
feat(geo): Add GeoIp lookup to light normalization

olksdr:
This change introduces the GeoIp lookup to light normalization process, which allows us to populate the `User` with correct geo information as soon as we can.  Which also allow to set a correct tag  in metrics extraction , if the metrics extraction is happening on our side (in the PoPs  in this case).

I still keep the GeoIp lookup in the store normalization, since  it's also used in few places, more specifically in `relay-cabi`, which exposed few different functions to the python lib. 

Lookup won't be triggered in any case if the Geo information is either provided by user or set on any of the steps. 

related to https://github.com/getsentry/relay/issues/2225


jjbayer (approved):

relay-server/src/actors/processor.rs
```diff
@@ -551,13 +551,13 @@ impl EnvelopeProcessorService {
         project_cache: Addr<ProjectCache>,
         upstream_relay: Addr<UpstreamRelay>,
     ) -> anyhow::Result<Self> {
+        let geoip_lookup = match config.geoip_path() {
```
iker-barriocanal, line 554: This relies on `processing` being active and requires a change in PoP configuration. Do PoP relays have the `processing` flag enabled in this case (and behave pretty much like processing relays)? If yes, have we identified the side effects of this config change? I'd like to avoid undesired side effects by PoPs attempting to behave like processing relays after this config change.

iker-barriocanal (approved):
I think this PR is a bit more risky than it may look like. Could we deploy this in a canary?

relay-server/src/actors/processor.rs
olksdr: you're right , the processing had to be set to enabled, otherwise it fails, but in [bc416d4](https://github.com/getsentry/relay/pull/2229/commits/bc416d4a283dc214012588bf1ef4495534839bd8) I make a change to make sure we can set the processing options even if the `processing.enabled` isn't set, which makes it easy to configure geoip db for the pops.

olksdr:
Re-requesting review , since added few more changes to the config part

iker-barriocanal (approved):
Nice!

PR #2230, opened 2023-06-20
instr(server): Count transactions by platform, source

jjbayer:
Collect counts of transactions per `(platform, source, contains_slashes)`. The goal of this metric is to verify a suspicion that most `unknown` transactions from javascript projects are URL transactions.

This metric replaces the now unused `transaction_source` metric, which is unused and produced some high-cardinality tags.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -1839,7 +1839,7 @@ impl EnvelopeProcessorService {
                     },
                     platform =
                         PlatformTag::from(event.platform.as_str().unwrap_or("other")).as_str(),
-                    contains_slashes = if event.transaction.as_str().unwrap_or("").contains('/') {
+                    contains_slashes = if event.transaction.as_str().unwrap_or_default().contains('/') {
                         "true"
                     } else {
                         "false"
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -25,12 +25,12 @@ use relay_filter::FilterStatKey;
 use relay_general::pii::{PiiAttachmentsProcessor, PiiConfigError, PiiProcessor};
 use relay_general::processor::{process_value, ProcessingState};
 use relay_general::protocol::Context::Trace;
-use relay_general::protocol::Contexts;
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -1832,10 +1832,18 @@ impl EnvelopeProcessorService {
                 let source = event.get_transaction_source();
 
                 metric!(
-                    counter(RelayCounters::EventTransactionSource) += 1,
-                    source = &source.to_string(),
-                    sdk = envelope.meta().client_name().unwrap_or("proprietary"),
-                    platform = event.platform.as_str().unwrap_or("other"),
+                    counter(RelayCounters::EventTransaction) += 1,
+                    source = match &source {
+                        TransactionSource::Other(_) => "other",
```
jjbayer, line 1837: Keep statsd cardinality low.

relay-server/src/actors/processor.rs
```diff
@@ -1832,10 +1832,18 @@ impl EnvelopeProcessorService {
...
+                        source => source.as_str(),
+                    },
+                    platform =
+                        PlatformTag::from(event.platform.as_str().unwrap_or("other")).as_str(),
```
jjbayer, line 1841: Keep statsd cardinality low.

relay-server/src/actors/processor.rs
```diff
@@ -1832,10 +1832,18 @@ impl EnvelopeProcessorService {
...
+                    contains_slashes = if event.transaction.as_str().unwrap_or("").contains('/') {
```
olksdr: Will `unwrap_or_default()` work here? Which also should give you an empty string. 

```suggestion
                    contains_slashes = if event.transaction.as_str().unwrap_or_default().contains('/') {
```

olksdr (approved):

iker-barriocanal (approved):

relay-server/src/actors/processor.rs
jjbayer: That's more characters 😉, but yeah makes sense!

PR #2236, opened 2023-06-22
feat(general): Keep stackframes closest to crash

### LINKED ISSUE #1980: Keep frames closest to crash
The `TrimmingProcessor` keeps the first 249 frames in a stack trace, plus the last frame in the stack trace. This means that we prioritize frames far away from the crash, and drop frames that are close to the crash:

> The list is ordered from caller to callee [...] The last frame is the one creating the exception.

https://develop.sentry.dev/sdk/event-payloads/stacktrace/#attributes

https://github.com/getsentry/relay/blob/35f5d8627b3de1749b25c31ff1b94c7a9833c6ad/relay-general/src/store/trimming.rs#L326-L339

Unless Relay inverts the entire list at some point before calling this function, this means that the frames closest to the crash are trimmed (except the one closest to it).

Consider trimming from the start instead.

TBS1996:
Keeps the 250 stackframes closest to the crash, instead of the earliest 249 + the latest one

references: #1980


relay-general/src/store/trimming.rs
```diff
@@ -326,17 +326,11 @@ fn trim_string(value: &mut String, meta: &mut Meta, max_chars: MaxChars) {
 }
 
 fn enforce_frame_hard_limit(frames: &mut Array<Frame>, meta: &mut Meta, limit: usize) {
-    // Trim down the frame list to a hard limit. Leave the last frame in place in case
-    // it's useful for debugging.
+    // Trim down the frame list to a hard limit. Prioritize the last frames.
     let original_length = frames.len();
-    if original_length >= limit {
+    if original_length > limit {
```
TBS1996, line 331: I'm not sure why the original implementation would trim it when the length was equal to the limit?

relay-general/src/store/trimming.rs
```diff
@@ -773,6 +767,32 @@ mod tests {
...
+            .apply(|f, m| {
+                enforce_frame_hard_limit(f, m, 3);
+                Ok(())
+            })
+            .unwrap();
+
+        // original_length isn't set, when limit is equal to length, as no trimming took place.
+        assert!(frames.meta().original_length().is_none());
+    }
+
```
TBS1996, line 795: Probably overkill, just checks that no trimming took place when limit is equal to length, unlike previous implementation

relay-general/src/store/trimming.rs
```diff
@@ -326,17 +326,11 @@ fn trim_string(value: &mut String, meta: &mut Meta, max_chars: MaxChars) {
...
-    if original_length >= limit {
+    if original_length > limit {
         meta.set_original_length(Some(original_length));
-
-        let last_frame = frames.pop();
-        frames.truncate(limit - 1);
-        if let Some(last_frame) = last_frame {
-            frames.push(last_frame);
-        }
+        let _ = frames.drain(0..original_length - limit);
```
TBS1996, line 333: the `_` is the values that are removed, not the remaining items (drain takes `&mut self`)

TBS1996:
I'll look into this sentence from referenced issue: `Unless Relay inverts the entire list at some point before calling this function` 

edit: couldn't find any reversing of the list by checking references to `rawstacktrace.frames`

jjbayer (approved):
LGTM. @Swatinem do you think we should keep the outermost frame in addition to the ~250 innermost frames, or does that just add confusion?

Swatinem (approved):
lgtm, and I would say it would be more confusing to keep a single frame. better to just make a clean cut

PR #2240, opened 2023-06-22
feat(spans): Span description fallback on unparameterized descriptions

iker-barriocanal:
Currently, unparameterized span descriptions are not extracted as metric tags. The number of tags without descriptions is bigger than expected, generating a poor UX. This PR adds a default span description so that it's not empty.

The default span description is built as follows: `{span.action} {span.domain}/<unparameterized>`, excluding the fields that don't exist.

#skip-changelog


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -261,6 +269,40 @@ pub(crate) fn extract_span_metrics(
     Ok(())
 }
 
+/// Returns the sanitized span description.
+///
+/// If a scrub description is provided, that's returned. If not, a new
+/// description is built for `http*` modules with the following format:
+/// `{action} {domain}/<unparameterized>`.
```
iker-barriocanal, line 276: I guess I could write a nice doctest here, but I didn't want to put that much effort.

olksdr (approved):

PR #2242, opened 2023-06-22
hot-fix, disable generation for new transaction tags

RaduW:
Temporarily disable generation of new transaction metric tags
They seem to cause too many buckets in production
disable htt.status_code, browser_name, os.name and geo.country_code they are creating too many buckets.

#skip-changelog


iker-barriocanal:
Do we still need this PR?

RaduW:
This was meant as a quick fix until we sort the problem with transactions being rejected because too many buckets were
created. The limit of number of buckets was increased.

PR #2243, opened 2023-06-23
ref(general): Split TransactionsProcessor into helper methods

jjbayer:
https://github.com/getsentry/team-ingest/issues/129 will require me to change the order of processing steps for transactions. To make that future change more readable, put processing steps into their own functions.

This PR does not contain any functional changes.

#skip-changelog


iker-barriocanal (approved):
Thanks for doing this!

olksdr (approved):
lgtm

PR #2245, opened 2023-06-23
ref(config): Combined interface for multiple scopes

jjbayer:
Fixes https://github.com/getsentry/team-ingest/issues/151.


TBS1996:
```Rust
    pub fn get_val<T>(
        &self,
        mut get_item: Box<dyn FnMut(&ProjectConfig) -> Option<&T>>,
    ) -> Option<&T> {
        if let Some(item) = get_item(&self.public_key) {
            return Some(item);
        }
        if let Some(item) = get_item(&self.project) {
            return Some(item);
        }
        if let Some(item) = get_item(&self.organization) {
            return Some(item);
        }
        if let Some(item) = get_item(&self.global) {
            return Some(item);
        }

        None
    }
````




do you think something like this would make sense for the values that this is relevant for? Or do we want to restrict it more so that it won't attempt to get it from global if it's not supposed to be in global?

PR #2246, opened 2023-06-23
fix(spans): Truncate span descriptions for span metrics

iker-barriocanal:
Metrics have tags, and their length is configured statically in the `AggregatorService`, with a [default value of 200 bytes](https://github.com/getsentry/relay/blob/e13d231441bbfe43ec7ca992a39eeb1beb1e92e3/relay-metrics/src/aggregation.rs#L994) for tag values. The metrics aggregator service [silently drops tags that exceed such limit](https://github.com/getsentry/relay/blob/dcad1bdad65ab0a7cf70692f0bd66fa3566b90a1/relay-metrics/src/aggregation.rs#L1629) (not ideal), which is often the case with some span descriptions. Truncating span description tags allows us to keep them.

This PR truncates the descriptions, and sets a `*` as the last character of the string if truncated. Note this only impacts span metrics (and corresponding values in `span.data`), leaving the descriptions in the original payload unmodified.

This change must only affect `span.description`, and not other tags that depend on it like `span.group`.

#skip-changelog

### DIFF
--- a/relay-server/Cargo.toml
+++ b/relay-server/Cargo.toml
@@ -30,8 +30,8 @@ axum = { version = "0.6.15", features = ["headers", "macros", "matched-path", "m
 axum-server = "0.4.7"
 backoff = "0.4.0"
 brotli = "3.3.4"
-bytes = { version = "1.4.0" }
 bytecount = "0.6.0"
+bytes = { version = "1.4.0" }
 chrono = { version = "0.4.24", default-features = false, features = ["clock", "std", "serde"] }
 data-encoding = "2.3.3"
 flate2 = "1.0.19"
--- a/relay-server/Cargo.toml
+++ b/relay-server/Cargo.toml
@@ -31,6 +31,7 @@ axum-server = "0.4.7"
 backoff = "0.4.0"
 brotli = "3.3.4"
 bytes = { version = "1.4.0" }
+bytecount = "0.6.0"
 chrono = { version = "0.4.24", default-features = false, features = ["clock", "std", "serde"] }
 data-encoding = "2.3.3"
 flate2 = "1.0.19"
--- a/relay-server/src/metrics_extraction/spans/mod.rs
+++ b/relay-server/src/metrics_extraction/spans/mod.rs
@@ -568,7 +568,14 @@ mod tests {
 
     use relay_metrics::AggregatorConfig;
 
-    use crate::metrics_extraction::spans::extract_span_metrics;
+    use crate::metrics_extraction::spans::{extract_span_metrics, truncate_string};
+
+    #[test]
+    fn test_truncate_string_no_panic() {
+        let string = "ÆÆÆÆÆÆ".to_owned();
+        let truncated = truncate_string(string, 3);
+        assert_eq!(truncated, "Æ*");
+    }
 
     macro_rules! span_transaction_method_test {
         // Tests transaction.method is picked from the right place.
--- a/relay-server/src/metrics_extraction/spans/mod.rs
+++ b/relay-server/src/metrics_extraction/spans/mod.rs
@@ -11,7 +11,6 @@ use once_cell::sync::Lazy;
 use regex::Regex;
 use relay_common::{EventType, UnixTimestamp};
 use relay_filter::csp::SchemeDomainPort;
-use relay_general::processor::MaxChars;
 use relay_general::protocol::Event;
 use relay_general::types::{Annotated, Value};
 use relay_metrics::{AggregatorConfig, Metric};
@@ -321,13 +320,22 @@ fn sanitized_span_description(
 /// doesn't use this type; this method uses it for simplicity, and because at
 /// the time of writing this the values are the same. If the limit this type
 /// provides is longer, this method may be ineffective.
-fn truncate_span_description(description: String) -> String {
-    let mut truncated = description;
-    if bytecount::num_chars(truncated.as_bytes()) > MaxChars::TagValue.limit() {
-        truncated.truncate(MaxChars::TagValue.limit() - 1);
-        truncated.push('*');
+fn truncate_span_description(mut description: String) -> String {
+    let max_limit = 200;
+
+    if bytecount::num_chars(description.as_bytes()) < max_limit {
+        return description;
     }
-    truncated
+
+    let mut cutoff = max_limit - 1; // Leave space for `*`
+
+    while cutoff > 0 && description.is_char_boundary(cutoff) {
+        cutoff -= 1;
+    }
+
+    description.truncate(cutoff);
+    description.push('*');
+    description
 }
 
 /// Regex with a capture group to extract the database action from a query.
--- a/relay-server/src/metrics_extraction/spans/mod.rs
+++ b/relay-server/src/metrics_extraction/spans/mod.rs
@@ -310,8 +310,8 @@ fn sanitized_span_description(
     Some(sanitized)
 }
 
-/// Trims the given string to ensure
-/// [`relay-metrics::AggregatorService`] doesn't drop it for being too long.
+/// Trims the given string with the given maximum bytes. Splitting only happens
+/// on char boundaries.
 ///
 /// If the string is short, it remains unchanged. If it's long, this method
 /// truncates it to the maximum allowed size and sets the last character to
--- a/relay-server/src/metrics_extraction/spans/mod.rs
+++ b/relay-server/src/metrics_extraction/spans/mod.rs
@@ -187,11 +187,17 @@ pub(crate) fn extract_span_metrics(
                 );
 
                 if let Some(scrubbed_desc) = sanitized_description {
-                    span_tags.insert(SpanTagKey::Description, scrubbed_desc.to_owned());
+                    // Truncating the span description's tag value is, for now,
+                    // a temporary solution to not get large descriptions dropped. The
+                    // group tag mustn't be affected by this, and still be
+                    // computed from the full, untruncated description.
 
-                    let mut span_group = format!("{:?}", md5::compute(scrubbed_desc));
+                    let mut span_group = format!("{:?}", md5::compute(&scrubbed_desc));
                     span_group.truncate(16);
                     span_tags.insert(SpanTagKey::Group, span_group);
+
+                    let truncated = truncate_span_description(scrubbed_desc);
+                    span_tags.insert(SpanTagKey::Description, truncated);
                 }
             }
 
@@ -282,8 +288,7 @@ fn sanitized_span_description(
     domain: Option<&str>,
 ) -> Option<String> {
     if let Some(scrubbed) = scrubbed_description {
-        let trimmed = truncate_span_description(scrubbed.to_owned());
-        return Some(trimmed);
+        return Some(scrubbed.to_owned());
     }
 
     if let Some(module) = module {
@@ -301,7 +306,6 @@ fn sanitized_span_description(
         sanitized.push_str(&format!("{domain}/"));
     }
     sanitized.push_str("<unparameterized>");
-    sanitized = truncate_span_description(sanitized);
 
     Some(sanitized)
 }
--- a/relay-server/src/metrics_extraction/spans/mod.rs
+++ b/relay-server/src/metrics_extraction/spans/mod.rs
@@ -11,6 +11,7 @@ use once_cell::sync::Lazy;
 use regex::Regex;
 use relay_common::{EventType, UnixTimestamp};
 use relay_filter::csp::SchemeDomainPort;
+use relay_general::processor::MaxChars;
 use relay_general::protocol::Event;
 use relay_general::types::{Annotated, Value};
 use relay_metrics::{AggregatorConfig, Metric};
@@ -281,7 +282,8 @@ fn sanitized_span_description(
     domain: Option<&str>,
 ) -> Option<String> {
     if let Some(scrubbed) = scrubbed_description {
-        return Some(scrubbed.to_owned());
+        let trimmed = truncate_span_description(scrubbed.to_owned());
+        return Some(trimmed);
     }
 
     if let Some(module) = module {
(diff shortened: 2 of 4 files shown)


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -299,10 +301,31 @@ fn sanitized_span_description(
...
+/// `*`. The maximum length is determined by [`MaxChars::TagValue`].
+///
+/// Note [`relay-metrics::AggregatorService`] is configured independently and
+/// doesn't use this type; this method uses it for simplicity, and because at
+/// the time of writing this the values are the same. If the limit this type
+/// provides is longer, this method may be ineffective.
+fn truncate_span_description(description: String) -> String {
+    let mut truncated = description;
+    if bytecount::num_chars(truncated.as_bytes()) > MaxChars::TagValue.limit() {
+        truncated.truncate(MaxChars::TagValue.limit() - 1);
```
iker-barriocanal: ~Using `MaxChars` is definitely not ideal here (as mentioned in the docstrings above), but I think it's better than hardcoding a value here. I believe building a proper solution to share the config with `AggregatorService` is out of scope for now.~ Fixed in https://github.com/getsentry/relay/pull/2246/commits/bef33de62b75e92b23c2e8a4907f85ae992f8bdb.

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -186,11 +187,17 @@ pub(crate) fn extract_span_metrics(
                 );
 
                 if let Some(scrubbed_desc) = sanitized_description {
-                    span_tags.insert(SpanTagKey::Description, scrubbed_desc.to_owned());
+                    // Truncating the span description's tag value is, for now,
+                    // a temporary solution to not get large descriptions dropped. The
+                    // group tag mustn't be affected by this, and still be
+                    // computed from the full, untruncated description.
```
iker-barriocanal, line 192: I don't really like adding a product requirement in a comment like this, so I'm open to removing it. This case is also covered in the integration test below, but I wouldn't rely too much on that because it depends a lot on the string to find MD5 hashing conflicts.

relay-server/src/metrics_extraction/spans/mod.rs
TBS1996, line 1: thoughts on these tests ? 

``` Rust
    #[test]
    fn test_no_panic_truncate() {
        let mut description = "".to_string();
        let chars = vec!['@', 'ي', 'Æ', '中', 'a'];

        let exceeding_limit = {
            let limit = MaxChars::TagValue.limit();
            limit + 50
        };

        for i in 0..exceeding_limit {
            let ch = chars[i % chars.len()];
            description.push(ch);
            truncate_span_description(description.clone());
        }
    }

    #[test]
    fn test_correct_bytes_limit() {
        let limit = MaxChars::TagValue.limit();

        let mut input = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut a".to_string();
        assert_eq!(input.len(), limit);
        let output = truncate_span_description(input.clone());
        // If input doesn't exceed limit, no change in input vs output.
        assert_eq!(input, output);

        input.push('@');
        assert_eq!(input.len(), limit + 1);
        let output = truncate_span_description(input.clone());
        assert_ne!(input, output);

        // Output should be equal to the limit.
        assert_eq!(output.len(), limit);
        assert_eq!(output.chars().last().unwrap(), '*');
    }
 ```

I guess your bytecount import avoids panic by always finding the char boundaries, so I made a test to verify this, using non-ascii letters that vary in byte lengths. 

I guess the second test is superflous since you test it in python, I'm a bit biased to rust unit testing 😅

File: relay-server/src/metrics_extraction/spans/mod.rs [hunk 3]
```diff
@@ -303,6 +310,26 @@ fn sanitized_span_description(
...
+///
+/// If the description is short, it remains unchanged. If it's long, this method
+/// truncates it to the maximum allowed length and sets the last character to
+/// `*`. The maximum length is determined by [`MaxChars::TagValue`].
+///
+/// Note [`relay-metrics::AggregatorService`] is configured independently and
+/// doesn't use this type; this method uses it for simplicity, and because at
+/// the time of writing this the values are the same. If the limit this type
+/// provides is longer, this method may be ineffective.
+fn truncate_span_description(description: String) -> String {
```
TBS1996: Could it be better to take a mutable reference as an argument ? wrt allocations it won't matter if the caller has only a &str of course, but in the cases where it's an owned type then you can avoid allocating.

for the &str ones, is it possible to utilize a Cow maybe?

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -186,11 +187,17 @@ pub(crate) fn extract_span_metrics(
                 );
 
                 if let Some(scrubbed_desc) = sanitized_description {
-                    span_tags.insert(SpanTagKey::Description, scrubbed_desc.to_owned());
+                    // Truncating the span description's tag value is, for now,
+                    // a temporary solution to not get large descriptions dropped. The
```
TBS1996, line 190: what would the permanent solution be ?

relay-server/src/metrics_extraction/spans/mod.rs
iker-barriocanal: I made the test in Python because I was looking for an integration test. A unit test doesn't show the issue here: the problem where the tags are dropped, `AggregatorService`, and where the fix is, here, are far enough in the relay processing pipeline for a unit test. Does this make sense?

I hope that at some point we redesign integration tests to fix some of the issues we have, and that may be a good chance to write them in Rust (and have a single language in the codebase).

relay-server/src/metrics_extraction/spans/mod.rs
iker-barriocanal: At this point, unknown. We will need more product iterations to see what the best action is; maybe we want to increase the limit of the values instead.

relay-server/src/metrics_extraction/spans/mod.rs
iker-barriocanal: If the caller has a `String`, my understanding is there are no additional allocations because it passes the ownership of the variable. If my understanding is correct, then `Cow` doesn't help here because there aren't additional allocations. Does this make sense, or am I missing something?

TBS1996: ah yeah I think youre right, my bad. Since you're directly passing the same variable then I think theres no new allocations as you said.

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -303,6 +310,26 @@ fn sanitized_span_description(
...
+/// truncates it to the maximum allowed length and sets the last character to
+/// `*`. The maximum length is determined by [`MaxChars::TagValue`].
+///
+/// Note [`relay-metrics::AggregatorService`] is configured independently and
+/// doesn't use this type; this method uses it for simplicity, and because at
+/// the time of writing this the values are the same. If the limit this type
+/// provides is longer, this method may be ineffective.
+fn truncate_span_description(description: String) -> String {
+    let mut truncated = description;
+    if bytecount::num_chars(truncated.as_bytes()) > MaxChars::TagValue.limit() {
```
jjbayer: The metrics aggregator seems to limit the number of bytes in the string, so why are we checking the number of _characters_ here instead? https://github.com/getsentry/relay/blob/83a435e0754df148184c4ad642fb8a33eba3a469/relay-metrics/src/aggregation.rs#L1629

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -303,6 +310,26 @@ fn sanitized_span_description(
...
+        truncated.truncate(MaxChars::TagValue.limit() - 1);
```
jjbayer: `truncate` takes a number of bytes as input, but on line 326, we interpret `MaxChars::TagValue.limit()` as a number of characters. But what's worse is that `truncate` [might panic](https://doc.rust-lang.org/std/string/struct.String.html#panics-3), as @TBS1996 said.

You could use a function like this to truncate safely: https://github.com/getsentry/relay/blob/83a435e0754df148184c4ad642fb8a33eba3a469/relay-monitors/src/lib.rs#L145-L149

jjbayer (changes requested):

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -303,6 +312,28 @@ fn sanitized_span_description(
...
+///
+/// If the description is short, it remains unchanged. If it's long, this method
+/// truncates it to the maximum allowed size and sets the last character to
+/// `*`.
+fn truncate_span_description(mut description: String, max_bytes: usize) -> String {
+    if bytecount::num_chars(description.as_bytes()) < max_bytes {
+        return description;
+    }
+
+    let mut cutoff = max_bytes - 1; // Leave space for `*`
```
iker-barriocanal, line 328: I'm not sure if extracting this into a variable makes the function easier to understand, open to suggestions.

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -303,6 +312,28 @@ fn sanitized_span_description(
...
 }
 
+/// Trims the given span description to ensure
+/// [`relay-metrics::AggregatorService`] doesn't drop it for being too long.
+///
+/// If the description is short, it remains unchanged. If it's long, this method
+/// truncates it to the maximum allowed size and sets the last character to
+/// `*`.
+fn truncate_span_description(mut description: String, max_bytes: usize) -> String {
+    if bytecount::num_chars(description.as_bytes()) < max_bytes {
```
jjbayer:
```suggestion
    if description.len() <= max_bytes {
```

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -303,6 +312,28 @@ fn sanitized_span_description(
...
+        return description;
+    }
+
+    let mut cutoff = max_bytes - 1; // Leave space for `*`
+
+    while cutoff > 0 && description.is_char_boundary(cutoff) {
```
jjbayer: We want the boundary check negated, right? For ASCII-strings, `is_char_boundary` is always true, so this would erase the string.

```suggestion
    while cutoff > 0 && !description.is_char_boundary(cutoff) {
```

jjbayer (changes requested):

relay-server/src/metrics_extraction/spans/mod.rs
TBS1996: can't you just do description.len() ? that will return the length of the string in bytes

https://doc.rust-lang.org/std/string/struct.String.html#method.len

TBS1996: oops, didn't see jorris had already commented

relay-server/Cargo.toml
```diff
@@ -31,6 +31,7 @@ axum-server = "0.4.7"
 backoff = "0.4.0"
 brotli = "3.3.4"
 bytes = { version = "1.4.0" }
+bytecount = "0.6.0"
```
olksdr, line 33: Could we ,please, sort this and put `bytecount` before `bytes`? 😄

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -303,6 +312,28 @@ fn sanitized_span_description(
...
+        cutoff -= 1;
+    }
+
+    description.truncate(cutoff);
+    description.push('*');
+    description
+}
```
jjbayer, line 337: Also, it would be nice to have a unit test that proves that special characters are truncated correctly (like @TBS1996 suggested, but I would target a specific case that would panic with a simple truncate).

TBS1996: tried the fuzzy method and it fails :( 

```Rust
    #[test]
    fn test_no_panic_truncate() {
        let mut description = "".to_string();
        let chars = vec!['@', 'ي', 'Æ', '中', 'a'];

        for i in 0..200 {
            let ch = chars[i % chars.len()];
            description.push(ch);
        }

        for max_bytes in 10..100 {
            truncate_span_description(description.clone(), max_bytes);
        }
    }
   ```

File: relay-server/src/metrics_extraction/spans/mod.rs [hunk 12]
```diff
@@ -303,6 +310,28 @@ fn sanitized_span_description(
...
 }
 
+/// Trims the given string with the given maximum bytes. Splitting only happens
+/// on char boundaries.
+///
+/// If the string is short, it remains unchanged. If it's long, this method
+/// truncates it to the maximum allowed size and sets the last character to
+/// `*`.
+fn truncate_string(mut string: String, max_bytes: usize) -> String {
+    if string.len() < max_bytes {
```
jjbayer: Still think this should be 

```suggestion
    if string.len() <= max_bytes {
```

That is what "max" means, and it would also be consistent with https://github.com/getsentry/relay/blob/0638c8c403e4178dc302791932d067a738aa3bd2/relay-metrics/src/aggregation.rs#L1629

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -539,7 +568,14 @@ mod tests {
 
     use relay_metrics::AggregatorConfig;
 
-    use crate::metrics_extraction::spans::extract_span_metrics;
+    use crate::metrics_extraction::spans::{extract_span_metrics, truncate_string};
+
+    #[test]
+    fn test_truncate_string_no_panic() {
+        let string = "ÆÆÆÆÆÆ".to_owned();
+        let truncated = truncate_string(string, 4);
```
jjbayer: Would be nice to repeat this test with limit = 3 to see what happens if the limit is not on the string boundary, and with limits `string.len() - 1`, `string.len()`, `string.len() + 1` to cover the edge cases.

jjbayer (approved):
Approving to unblock, but I still have some comments :)

PR #2248, opened 2023-06-23
Introduce relay pipedream

mattgauntseo-sentry:
This introduces a new set of GoCD pipelines generated from jsonnet.

The pipelines produced are:

- deploy-relay
- deploy-us
- deploy-monitor

`deploy-relay` is a simple pipeline that triggers a deployment of relay but doesn't actually deploy anything.
`deploy-us` is the equivalent and deploy-relay and deploy-relay-pops that currently exist. The only difference is that the pops pipeline has been turned into a stage. `deploy-monitor` is a new pipeline that deploys to a single-tenant style environment.

#skip-changelog

### DIFF
--- a/gocd/README.md
+++ b/gocd/README.md
@@ -1,0 +1,14 @@
+# Relay Pipelines
+
+Relay is in the process of moving to a set of rendered jsonnet pipelines.
+
+## Jsonnet
+
+You can render the jsonnet pipelines by running:
+
+```
+make gocd
+```
+
+This will clean, fmt, lint and generate the GoCD pipelines to
+`./gocd/generated-pipelines`.
--- a/gocd/README.md
+++ b/gocd/README.md
@@ -12,3 +12,7 @@ make gocd
 
 This will clean, fmt, lint and generate the GoCD pipelines to
 `./gocd/generated-pipelines`.
+
+
+The Relay pipelines are using the https://github.com/getsentry/gocd-jsonnet
+libraries to generate the pipeline for each region.
--- a/gocd/README.md
+++ b/gocd/README.md
@@ -31,7 +31,7 @@ The `gocd/templates/relay.jsonnet` file is the entry point for the
 relay pipelines.
 
 `gocd/templates/libs/*.libsonnet` define the pipeline behaviors for
-deploy relay and relay-pops. These libraries are use to create a
+deploy relay and relay-pops. These libraries are used to create a
 GoCD pipeline, following the same naming as the
 [GoCD yaml pipelines](https://github.com/tomzo/gocd-yaml-config-plugin#readme).
 
--- a/gocd/README.md
+++ b/gocd/README.md
@@ -16,3 +16,44 @@ This will clean, fmt, lint and generate the GoCD pipelines to
 
 The Relay pipelines are using the https://github.com/getsentry/gocd-jsonnet
 libraries to generate the pipeline for each region.
+
+## Files
+
+Below is a description of the directories in the `gocd/` directory.
+
+### `gocd/templates/`
+
+These are a set of jsonnet and libsonnet files which are used
+to generate the relay pipelines. This avoids duplication across
+our GoCD pipeline files as we deploy to multiple regions.
+
+The `gocd/templates/relay.jsonnet` file is the entry point for the
+relay pipelines.
+
+`gocd/templates/libs/*.libsonnet` define the pipeline behaviors for
+deploy relay and relay-pops. These libraries are use to create a
+GoCD pipeline, following the same naming as the
+[GoCD yaml pipelines](https://github.com/tomzo/gocd-yaml-config-plugin#readme).
+
+`gocd/templates/bash/*.sh` are shell scripts that are inlined in the
+result pipelines. This seperation means syntax highlighting and
+extra tooling works for relay's bash scripts.
+
+`gocd/templates/jsonnetfile.json` and `gocd/templates/jsonnetfile.lock.json`
+are used by [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler#readme), a package manager for jsonnet.
+
+You can update jsonnet dependencies by runnning `jb update`.
+
+### `gocd/generated-pipelines/`
+
+The current setup of GoCD at sentry is only able to look for
+yaml pipelines in a repo, so relay has to commit the generated
+pipelines.
+
+The dev-infra team is working on a GoCD plugin that will accept
+the jsonnet directly, removing the need for the generated-pipelines.
+
+### `gocd/pipelines/`
+
+These are the original relay pipelines and will be used until we move
+to the jsonnet pipelines.
--- a/gocd/generated-pipelines/relay-next-monitor.yaml
+++ b/gocd/generated-pipelines/relay-next-monitor.yaml
@@ -5,7 +5,7 @@
          "environment_variables": {
             "SENTRY_REGION": "monitor"
          },
-         "group": "relay-next",
+         "group": "relay-next-region-deployments",
          "lock_behavior": "unlockWhenFinished",
          "materials": {
             "relay_repo": {
--- a/gocd/generated-pipelines/relay-next-monitor.yaml
+++ b/gocd/generated-pipelines/relay-next-monitor.yaml
@@ -1,0 +1,144 @@
+{
+   "format_version": 10,
+   "pipelines": {
+      "region-deploy-relay-next-monitor": {
+         "environment_variables": {
+            "SENTRY_REGION": "monitor"
+         },
+         "group": "relay-next",
+         "lock_behavior": "unlockWhenFinished",
+         "materials": {
+            "relay_repo": {
+               "branch": "master",
+               "destination": "relay",
+               "git": "git@github.com:getsentry/relay.git",
+               "shallow_clone": true
+            },
+            "upstream_pipeline": {
+               "pipeline": "region-deploy-relay-next-us",
+               "stage": "pipeline-complete"
+            }
+         },
+         "stages": [
+            {
+               "ready": {
+                  "jobs": {
+                     "ready": {
+                        "tasks": [
+                           {
+                              "exec": {
+                                 "command": true
+                              }
+                           }
+                        ]
+                     }
+                  }
+               }
+            },
+            {
+               "wait": {
+                  "approval": {
+                     "type": "manual"
+                  },
+                  "jobs": {
+                     "wait": {
+                        "tasks": [
+                           {
+                              "exec": {
+                                 "command": true
+                              }
+                           }
+                        ]
+                     }
+                  }
+               }
+            },
+            {
+               "checks": {
+                  "fetch_materials": true,
+                  "jobs": {
+                     "checks": {
+                        "elastic_profile_id": "relay",
+                        "environment_variables": {
+                           "GITHUB_TOKEN": "{{SECRET:[devinfra-github][token]}}"
+                        },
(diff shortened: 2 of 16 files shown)


gocd/generated-pipelines/relay-next-monitor.yaml
olksdr, line 1: Do we have to check in all the generated json? 
Is it possible to get actual yaml here?

olksdr (commented):
To be honest it's quite difficult to review, there are lots of files, and some of them not really obvious (without any docs):
 
* a lot of generated files - do we have to have those checked into the repo? 
* lack of docs in jsonnet libs files - is it possible to add some top level documentation what the purpose of the file / lib?
* also migrating from 3 yaml files to bunch of json/jsonnet, shell and yaml
* It is difficult to spot any changes in the generated files, if something goes wrong - I there is a mistake in the jsonnet and it produces bunch of json , it still can be valid json, but won't be doing what was planned. 
* will be there any docs linked , so ingest team can contribute to the project's pipelines, or infra-dev-team will fully support all and any requests for changes?

How does it make this setup easier to use if we still check in the generated json? Why then not directly use yaml as it was done before?

gocd/generated-pipelines/relay-next-monitor.yaml
mattgauntseo-sentry: We can convert to yaml if you prefer - give me a sec.

mattgauntseo-sentry:
@olksdr If you prefer to check in the yaml as before that works for me.

The reason against it is the follow:

1. There will be a great deal of duplication for each pipeline of s4s, US, EU and each customer deployment.
2. Chaining the deployments together end-to-end may be error prone (albeit that will be a one time issue).

Regarding your questions:

> * a lot of generated files - do we have to have those checked into the repo? 

At the moment we do need the generated files as gocd expects the yaml. We are in the process of trying to build a GoCD plugin that will take the jsonnet directly.

> * lack of docs in jsonnet libs files - is it possible to add some top level documentation what the purpose of the file / lib?

Ofc, add some here: https://github.com/getsentry/gocd-jsonnet/pull/9

> * also migrating from 3 yaml files to bunch of json/jsonnet, shell and yaml

Yeah I get you, if its not desirable, I'm happy to just move forward with the yaml files. Pulling out the bash is nice for editing since you get syntax highlighting, which you don't when its in the yaml directly.

> * It is difficult to spot any changes in the generated files, if something goes wrong - I there is a mistake in the jsonnet and it produces bunch of json , it still can be valid json, but won't be doing what was planned. 

GoCD will not accept changes if they are invalid, and secondly the main benefit of jsonnet is it cuts down on duplication, which I'd argue will be more error prone once we add all regions.

> * will be there any docs linked , so ingest team can contribute to the project's pipelines, or infra-dev-team will fully support all and any requests for changes?

The end goal will be for the ingest team to own these pipelines, so if the desirable state is to have just the yaml files, I'm happy to add them.

> How does it make this setup easier to use if we still check in the generated json? Why then not directly use yaml as it was done before?

The setup will be easier by removing duplication, if you disagree, I'm happy to make smaller PRs along the lines of:

1. Copy existing relay pipeslines into `relay-next` group
2. Make the changes that result in pipelines above

Once that's down and s4s it working correctly, I'll add the additional pipelines for ST customers.

mattgauntseo-sentry:
@olksdr if this is more desirable please let me know: https://github.com/getsentry/relay/pull/2253

gocd/generated-pipelines/relay-next-us.yaml
```diff
@@ -0,0 +1,175 @@
...
+                    /devinfra/scripts/k8s/k8stunnel
+
+                    /devinfra/scripts/k8s/k8s-deploy.py \
+                        --label-selector="service=relay,deploy_if_production=true" \
+                        --image="us.gcr.io/sentryio/relay:${GO_REVISION_RELAY_REPO}" \
+                        --container-name="relay"
+              timeout: 1200
+      - deploy-pops:
+          fetch_materials: true
+          jobs:
```
olksdr, line 83:
```suggestion
      - deploy-pops:
          approval:
            type: manual
          fetch_materials: true
          jobs:
```

Is it possible to add here `approve` to `manual`? Or this step by default will have to be approved?

gocd/generated-pipelines/relay-next-us.yaml
```diff
@@ -0,0 +1,175 @@
...
+          approval:
+            type: manual
+          jobs:
+            wait:
+              tasks:
+                - exec:
+                    command: true
+      - checks:
+          fetch_materials: true
+          jobs:
```
olksdr, line 21: And if checks are passing we want processing relay (which is `deploy-production` step) proceed automatically.

olksdr (changes requested):
Thanks for taking the time and adding the `yaml` generated files and some docs.

There are still few things I would like to clarify before approving this PR:

For out production we usually have following workflow:
* checks passing -> we want relay-production to be deployed
* than we usually take some time to check if the rollout succeeds, check the errors etc. 
* only then we want to manually trigger POPs deployment, 
 
And this is crucial we keep this workflow for now.

----

To address some of your comments:

> @olksdr If you prefer to check in the yaml as before that works for me.
> 
> The reason against it is the follow:
> 
> 1. There will be a great deal of duplication for each pipeline of s4s, US, EU and each customer deployment.
> 2. Chaining the deployments together end-to-end may be error prone (albeit that will be a one time issue).

I'm not against the automation or generating the jobs. I'm looking forward for making our lives easier here, and have  fewer misconfigurations. But I'm also would like to have less complex system, so more people could understand and contribute to. 

Maybe in the README you added it's worth mentioning that generated code does not have to be changed manually, etc. and rely fully on the jsonnet configs. 

> At the moment we do need the generated files as gocd expects the yaml. We are in the process of trying to build a GoCD plugin that will take the jsonnet directly.

Makes sense. Thanks for clarification! 
 
> > * lack of docs in jsonnet libs files - is it possible to add some top level documentation what the purpose of the file / lib?
> 
> Ofc, add some here: [getsentry/gocd-jsonnet#9](https://github.com/getsentry/gocd-jsonnet/pull/9)

Thanks for this! It makes it easier to understand what's going on. 

> GoCD will not accept changes if they are invalid, and secondly the main benefit of jsonnet is it cuts down on duplication, which I'd argue will be more error prone once we add all regions.

Will it also show where the error occurred? And how this is related to the jsonned libs?  
The problem with this workflow is that we won't know that something goes wrong till we merge it into `master` and only then we see something does not work (if we even spot the error). Is that right? 
Is it possible to add tests to the jsonnet and ensure that we produce the valid GoCD pipelines? 


> The end goal will be for the ingest team to own these pipelines, so if the desirable state is to have just the yaml files, I'm happy to add them.

It's ok if the files are generated as I mentioned above, but we have to make sure the team understands the process and how to make the changes to pipelines. 

Of course it would be great to have only one set of files to be checked in. It would be definitely great to see that plugging for GoCD you are working on.

gocd/generated-pipelines/relay-next-us.yaml
mattgauntseo-sentry: This was added as a precaution while initially rolling this out, but removed as it should be safe.

gocd/generated-pipelines/relay-next-us.yaml
mattgauntseo-sentry: Sorry that I missed the manual approval in relay-pops.

I've added an extra stage after deploy-production, `progress-to-pops` that is a manual approval step. This will be a manual approval for each region, which I'm assuming is the best option for now, however if you wanted to to just be for US deployments let me know and I can make that change.

mattgauntseo-sentry:
You are right regarding that the issues will only surface when committed to master, but that will hopefully not be the case for much longer, Josh is working on using the GoCD API to validate the pipeline against the GoCD service, this should catch a range of issues that the current GoCD CLI can't catch.

Regarding the quality of the errors, it doesn't link bank to the jsonnet, it'll only refer to the generated yaml, but so far I've found it fairly easy to trace that back to the jsonnet.

I can add some tests that work in a similar fashion to the [gocd-jsonnet repo](https://github.com/getsentry/gocd-jsonnet) where the test cases compare generated yaml to a set of golden files, this is largely helpful for diff'ing changes locally, but they end up being similar to what you'd see in a PR.

mattgauntseo-sentry:
@olksdr is there more you would like changed here?

I've updated relay after getting feedback from other services that have added pipedreams (snuba, super big consumers and symbolicator)

olksdr (approved):
From my side it look good and we can try those out. 
I would like to see how they look in GoCD and then maybe we will have to make some adjustments. 

If someone from @getsentry/owners-ingest want to have another look, it would be also nice.

gocd/templates/libs/relay.libsonnet
```diff
@@ -0,0 +1,91 @@
...
+          },
+        },
+      },
+    },
+
+    {
+      'progress-to-pops': {
+        approval: {
+          type: 'manual',
+          allow_only_on_success: true,
```
iker-barriocanal, line 74: Does a manual approval mean that someone needs to click to proceed with the deployment? My understanding is that the goal was to do this automatically.

gocd/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+These are a set of jsonnet and libsonnet files which are used
+to generate the relay pipelines. This avoids duplication across
+our GoCD pipeline files as we deploy to multiple regions.
+
+The `gocd/templates/relay.jsonnet` file is the entry point for the
+relay pipelines.
+
+`gocd/templates/libs/*.libsonnet` define the pipeline behaviors for
+deploy relay and relay-pops. These libraries are use to create a
```
iker-barriocanal:
```suggestion
deploy relay and relay-pops. These libraries are used to create a
```

gocd/README.md
iker-barriocanal, line 1: Thanks for writing this!

gocd/generated-pipelines/relay-next-monitor.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+                    /devinfra/scripts/k8s/k8s-deploy.py \
+                        --label-selector="service=relay,deploy_if_production=true" \
+                        --image="us.gcr.io/sentryio/relay:${GO_REVISION_RELAY_REPO}" \
+                        --container-name="relay"
+              timeout: 1200
+      - progress-to-pops:
+          approval:
+            allow_only_on_success: true
+            type: manual
```
iker-barriocanal, line 73: Will this pause the deployment pipeline until manual confirmation?

iker-barriocanal (approved):

gocd/templates/libs/relay.libsonnet
mattgauntseo-sentry: That is up to the folks owning the pipeline.

Currently relay-pops is a manual pipeline and this preserves that.

I think it's worth keeping for now to be as close to the current setup and then review it once everyone is more comfortable with pipedream deployments.

iker-barriocanal: Sounds good to me!

gocd/generated-pipelines/relay-next-monitor.yaml
mattgauntseo-sentry: Sorry I missed this comment - Yes it will, so the process for now will be:

1. Manually start deploy-relay
2. US relay is deployed
3. Manually start deploy-pops for US
4. US relay pops is deployed
5. s4s relay is deployed
6. Manually start deploy-pops for s4s
7. s4s relay pops is deployed

PR #2250, opened 2023-06-26
feat(txname): Apply transaction rules to legacy SDKs

jjbayer:
https://github.com/getsentry/sentry/pull/51437 expanded the scope of transaction name clustering to transactions from old SDKs that do not send a `source` annotation yet. This PR enables the application of rules to this type of transactions in Relay.

Because there is a potential of triggering the cardinality limiter, I gated this new behavior with a feature flag.

The condition for applying clusterer rules has now become more complicated, so I decided to remove / ignore the `scope` configuration flag entirely.

### DIFF
--- a/relay-dynamic-config/src/feature.rs
+++ b/relay-dynamic-config/src/feature.rs
@@ -51,6 +51,7 @@ impl Serialize for Feature {
             }
             Feature::DeviceClassSynthesis => "organizations:device-class-synthesis",
             Feature::SpanMetricsExtraction => "projects:span-metrics-extraction",
+            Feature::NormalizeLegacyTransactions => "organizations:normalize-legacy-transactions",
             Feature::Unknown(s) => s,
         })
     }
--- a/relay-dynamic-config/src/feature.rs
+++ b/relay-dynamic-config/src/feature.rs
@@ -15,6 +15,8 @@ pub enum Feature {
     DeviceClassSynthesis,
     /// Enables metric extraction from spans.
     SpanMetricsExtraction,
+    /// Apply transaction normalization rules to transactions from legacy SDKs.
+    NormalizeLegacyTransactions,
     /// Forward compatibility.
     Unknown(String),
 }
--- a/relay-dynamic-config/src/feature.rs
+++ b/relay-dynamic-config/src/feature.rs
@@ -51,7 +51,9 @@ impl Serialize for Feature {
             }
             Feature::DeviceClassSynthesis => "organizations:device-class-synthesis",
             Feature::SpanMetricsExtraction => "projects:span-metrics-extraction",
-            Feature::NormalizeLegacyTransactions => "organizations:normalize-legacy-transactions",
+            Feature::NormalizeLegacyTransactions => {
+                "organizations:transaction-name-normalize-legacy"
+            }
             Feature::Unknown(s) => s,
         })
     }
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -2036,6 +2036,131 @@ mod tests {
     }
 
     #[test]
+    fn test_normalize_twice() {
+        // Simulate going through a chain of relays.
+        let json = r#"
+        {
+            "type": "transaction",
+            "transaction": "/foo/rule-target/user/123/0/",
+            "transaction_info": {
+              "source": "url"
+            },
+            "timestamp": "2021-04-26T08:00:00+0100",
+            "start_timestamp": "2021-04-26T07:59:01+0100",
+            "contexts": {
+                "trace": {
+                    "trace_id": "4c79f60c11214eb38604f4ae0781bfb2",
+                    "span_id": "fa90fdead5f74053",
+                    "op": "rails.request"
+                }
+            }
+        }
+        "#;
+
+        let rules = vec![TransactionNameRule {
+            pattern: LazyGlob::new("/foo/*/user/*/**".to_string()),
+            expiry: Utc::now() + Duration::hours(1),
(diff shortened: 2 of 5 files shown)


relay-dynamic-config/src/feature.rs
```diff
@@ -15,6 +15,8 @@ pub enum Feature {
     DeviceClassSynthesis,
     /// Enables metric extraction from spans.
     SpanMetricsExtraction,
+    /// Apply transaction normalization rules to transactions from legacy SDKs.
+    NormalizeLegacyTransactions,
```
jjbayer, line 19: This feature has not been enabled by sentry yet, so merging this PR should not cause any functional changes.

relay-general/src/store/transactions/rules.rs
```diff
@@ -176,25 +161,18 @@ pub struct TransactionNameRule {
     pub pattern: LazyGlob,
     /// Date time when the rule expires and it should not be applied anymore.
     pub expiry: DateTime<Utc>,
-    /// Object containing transaction attributes the rules must only be applied to.
-    #[serde(default)]
-    pub scope: TransactionNameRuleScope,
```
jjbayer, line 181: The condition for applying clustering rules has changed from `source == "url"` to `source in ["url", "sanitized", null]`. Instead of extending the config protocol to support lists, I suggest we drop this field from the config and let Relay decide when to apply these rules.

relay-general/src/store/transactions/rules.rs
```diff
@@ -225,14 +203,8 @@ impl TransactionNameRule {
...
-    fn matches(&self, transaction: &str, info: &TransactionInfo) -> bool {
-        let now = Utc::now();
-        info.source
-            .value()
-            .map(|s| s == &self.scope.source)
-            .unwrap_or_default()
-            && self.expiry > now
-            && self.pattern.compiled().is_match(transaction)
+    fn matches(&self, transaction: &str) -> bool {
+        self.expiry > Utc::now() && self.pattern.compiled().is_match(transaction)
```
jjbayer, line 207: Not checking the `source` also means that we're now applying sanitation rules to transactions that were already marked as sanitized. This should be fine, since normalizing a string like `/foo/*/bar/` with rule `/foo/*/**` will result in `/foo/*/bar/` again.

relay-general/src/store/transactions/processor.rs
```diff
@@ -532,14 +541,13 @@ impl Processor for TransactionsProcessor<'_> {
                 .set_value(Some("<unlabeled transaction>".to_owned()))
         }
 
+        set_default_transaction_source(event);
```
jjbayer, line 549: After setting the default transaction source, we know that `source:unknown` means low-cardinality and `source:null` means high-cardinality (see function itself).

relay-general/src/store/transactions/rules.rs
jjbayer: Note that old Relays will still deserialize the missing scope as `{"source": "url"}` because of the `#[serde(default)]` attribute, so this change is backward compatible.

relay-general/src/store/transactions/processor.rs
```diff
@@ -184,23 +182,34 @@ impl<'r> TransactionsProcessor<'r> {
...
-        if matches!(
-            event.get_transaction_source(),
-            &TransactionSource::Url | &TransactionSource::Sanitized
-        ) {
+        let source = event
+            .transaction_info
+            .value()
+            .and_then(|i| i.source.value());
+
+        // Treat the transaction a a URL in the following cases:
```
olksdr:
```suggestion
        // Treat the transaction as an URL in the following cases:
```

relay-general/src/store/transactions/rules.rs
olksdr: Is there a unit test for to make sure it behaves as we expect for sanitized transactions? 
Maybe I'm thinking about this wrong, but when the new pattern comes in, which will the current url, it will get replaced and already sanitized value will be saved in meta info.

olksdr (approved):
Some questions, but overall lgtm

relay-general/src/store/transactions/processor.rs
```suggestion
        // Treat the transaction as a URL in the following cases:
```
?

relay-general/src/store/transactions/processor.rs
```diff
@@ -184,23 +182,34 @@ impl<'r> TransactionsProcessor<'r> {
...
+        // 1. It is marked with source:url
+        // 2. It is marked with source:sanitized, in which case we run normalization again.
+        // 3. It has no source attribute because it's from an old SDK version,
+        //    but it contains slashes and we expect it to be high-cardinality
+        //    based on the sdk information (see `set_default_transaction_source`).
```
iker-barriocanal: nit:
```suggestion
        //    based on the SDK information (see `set_default_transaction_source`).
```

relay-general/src/store/transactions/processor.rs
```diff
@@ -184,23 +182,34 @@ impl<'r> TransactionsProcessor<'r> {
...
+        let treat_as_url = matches!(
```
iker-barriocanal: nit: do you think extracting these conditions into a separate method would ease the readability of this method?

iker-barriocanal (approved):

relay-general/src/store/transactions/rules.rs
jjbayer: Good point, will add some tests.

relay-general/src/store/transactions/processor.rs
jjbayer: Yep, done.

PR #2251, opened 2023-06-26
fix(filter): Added is_enabled field to transactions name filter

RaduW:
This PR adds the is_enabled field to the TransactionsName configuration.

The flag is added in order to make it easier to work with it in Sentry (since all inbound filters have it).

### DIFF
--- a/relay-filter/src/config.rs
+++ b/relay-filter/src/config.rs
@@ -138,10 +138,12 @@ pub struct ErrorMessagesFilterConfig {
 
 /// Configuration for transaction name filter.
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct IgnoreTransactionsFilterConfig {
     /// List of patterns for ignored transactions that should be filtered.
     pub patterns: GlobPatterns,
     /// True if the filter is enabled
+    #[serde(default)]
     pub is_enabled: bool,
 }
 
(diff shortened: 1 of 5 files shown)


relay-filter/src/transaction_name.rs
```diff
@@ -18,6 +18,10 @@ pub fn should_filter(
     event: &Event,
     config: &IgnoreTransactionsFilterConfig,
 ) -> Result<(), FilterStatKey> {
+    if !config.is_enabled || config.patterns.is_empty() {
```
jjbayer: This is the same check as the `is_empty` method, right?

```suggestion
    if config.is_empty() {
```

relay-filter/src/config.rs
```diff
@@ -141,12 +141,14 @@ pub struct ErrorMessagesFilterConfig {
 pub struct IgnoreTransactionsFilterConfig {
     /// List of patterns for ignored transactions that should be filtered.
     pub patterns: GlobPatterns,
+    /// True if the filter is enabled
+    pub is_enabled: bool,
```
jjbayer, line 147: There is a tiny chance that somebody updated their external Relay since this filter was introduced. These old Relays would ignore this flag until they update again. But given the short time frame and the fact that there was no calendar release in between, I think we can ignore this problem.

jjbayer (approved):

olksdr (approved):
What @jjbayer said.
Otherwise lgtm

relay-filter/src/transaction_name.rs
RaduW: of course, you are right ! Thanks.

iker-barriocanal (approved):
Thanks for getting to this!

Issue #2237, opened 2023-06-22
Add metrics extraction to project config


jan-auer:
To extract any sort of metrics from events, Relay needs a generic specification in project configs that configures the value, tags, and conditions under which metrics are to be extracted.

## Spec

The WIP specification is located in this [**Notion document**](https://www.notion.so/sentry/Metrics-Extraction-Project-Configs-87d8f1e9dc4a422196904ecafc53caf5). 

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/relay/pull/2252
- [ ] https://github.com/getsentry/relay/pull/2255
```

PR #2252, opened 2023-06-26
feat(dynamic_config): Add config for dynamic metrics extraction

jan-auer:
Adds configuration for dynamic metrics extraction to project configuration. This is inactive at this moment, as actual extraction will be implemented in a follow-up.

Ref https://github.com/getsentry/relay/issues/2237

### DIFF
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -192,24 +192,39 @@ pub struct TagSpec {
     pub key: String,
 
     /// TODO(ja): Doc
-    #[serde(flatten)]
-    pub expr: TagExpression,
+    #[serde(default)]
+    pub field: Option<String>,
+
+    /// TODO(ja): Doc
+    #[serde(default)]
+    pub value: Option<String>,
 
     /// TODO(ja): Doc
     #[serde(default)]
     pub condition: Option<RuleCondition>,
 }
 
+impl TagSpec {
+    /// TODO(ja): Doc
+    pub fn expression(&self) -> TagExpression<'_> {
+        if let Some(ref field) = self.field {
+            TagExpression::Field(field)
+        } else if let Some(ref value) = self.value {
+            TagExpression::Value(value)
+        } else {
+            TagExpression::Unknown
+        }
+    }
+}
+
 /// TODO(ja): Doc
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
-pub enum TagExpression {
+#[derive(Clone, Debug, PartialEq)]
+pub enum TagExpression<'a> {
     /// TODO(ja): Doc
-    Value(String),
+    Value(&'a str),
     /// TODO(ja): Doc
-    Field(String),
+    Field(&'a str),
     /// TODO(ja): Doc
-    #[serde(other)]
     Unknown,
 }
 
@@ -222,41 +237,20 @@ mod tests {
     fn parse_tag_spec_value() {
         let json = r#"{"key":"foo","value":"bar"}"#;
         let spec: TagSpec = serde_json::from_str(json).unwrap();
-
-        let expected = TagSpec {
-            key: "foo".to_owned(),
-            expr: TagExpression::Value("bar".to_owned()),
-            condition: None,
-        };
-
-        assert_eq!(spec, expected);
+        assert_eq!(spec.expression(), TagExpression::Value("bar"));
     }
 
     #[test]
     fn parse_tag_spec_field() {
         let json = r#"{"key":"foo","field":"bar"}"#;
         let spec: TagSpec = serde_json::from_str(json).unwrap();
-
-        let expected = TagSpec {
-            key: "foo".to_owned(),
-            expr: TagExpression::Field("bar".to_owned()),
-            condition: None,
-        };
-
-        assert_eq!(spec, expected);
+        assert_eq!(spec.expression(), TagExpression::Field("bar"));
     }
 
     #[test]
     fn parse_tag_spec_unsupported() {
         let json = r#"{"key":"foo","somethingNew":"bar"}"#;
         let spec: TagSpec = serde_json::from_str(json).unwrap();
-
-        let expected = TagSpec {
-            key: "foo".to_owned(),
-            expr: TagExpression::Unknown,
-            condition: None,
-        };
-
-        assert_eq!(spec, expected);
+        assert_eq!(spec.expression(), TagExpression::Unknown);
     }
 }
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -2,6 +2,7 @@
 
 use std::collections::BTreeSet;
 
+use relay_common::DataCategory;
 use relay_sampling::RuleCondition;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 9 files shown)


relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
         self.version > 0 && self.version <= TRANSACTION_EXTRACT_VERSION
     }
 }
+
+/// Configuration for generic extraction of metrics from all data categories.
+#[derive(Clone, Default, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MetricExtractionConfig {
+    /// Versioning of metrics extration. Relay skips extraction if the version is not supported.
```
olksdr:
```suggestion
    /// Versioning of metrics extraction. Relay skips extraction if the version is not supported.
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+    /// condition will be applied.
+    #[serde(default)]
+    pub tags: Vec<TagSpec>,
+}
+
+/// Configuration for a tag to add to a metric.
+///
+/// Tags values can be static if defined through `value` or dynamically queried from the payload if
+/// defined through `field`. These two options are mutually exclusive, behavior is undefined if both
+/// are specified.
```
olksdr, line 229: So this is the false statement here, if both specified we always return `field` as a `source` on line 255.
Or will be there more code which will make this behavior undefined ?

olksdr (approved):
Left few questions , otherwise LGTM

relay-dynamic-config/src/metrics.rs
jan-auer: My intent was to leave behavior undefined on the protocol level (hence the doc comment), but of course we need to do *something*. At this point, I'd like to leave our options open to throw an error, use both as a fallback, or just have one take precedence over the other. Either way, at this point we do not expect Sentry to emit entries with both set.

relay-dynamic-config/src/project.rs
```diff
@@ -91,6 +97,7 @@ impl Default for ProjectConfig {
             breakdowns_v2: None,
             session_metrics: SessionMetricsConfig::default(),
             transaction_metrics: None,
+            metric_extraction: Default::default(),
```
iker-barriocanal, line 100: Q: how does `default` behave here with `ErrorBoundary<MetricExtractionConfig>`? Does it create a default `MetricExtractionConfig`?

Does wrapping it with `Option` help to differentiate if the config was bogus or was not provided at all? `Option<ErrorBoundary<MetricExtractionConfig>>`.

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+    ///
+    /// These tags add further tags to a range of metrics. If some metrics already have a matching
+    /// tag extracted, the existing tag is left unchanged.
+    #[serde(default)]
+    pub tags: Vec<TagMapping>,
+}
+
+impl MetricExtractionConfig {
+    /// Returns `true` if metric extraction is disabled.
+    pub fn is_empty(&self) -> bool {
```
iker-barriocanal: nit: do you think `disabled` and `empty` is confusing on a config?
```suggestion
    /// Returns `true` if metric extraction is disabled.
    pub fn is_disabled(&self) -> bool {
```

iker-barriocanal: nit2: I prefer `is_enabled`.

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+        self.version == 0 || (self.metrics.is_empty() && self.tags.is_empty())
```
iker-barriocanal: Are we postponing the minimum version in Relay for now?

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+    }
+}
+
+/// Specification for a metric to extract from some data.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MetricSpec {
+    /// Category of data to extract this metric for.
+    pub category: DataCategory,
```
iker-barriocanal, line 163: I'm not sure how I feel about reusing `DataCategory`, but I guess it's fine.

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+
+    /// Identifier of the metric to extract.
```
iker-barriocanal: nit: same as in [here](https://github.com/getsentry/relay/blob/ee77573331057763ad97e6ba784b5654fb38c7ae/relay-metrics/src/aggregation.rs#L729).
```suggestion
    /// The MRI (metric resource identifier).
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+    ///
+    /// How the value is treated depends on the metric type:
+    ///
+    /// - **Counter** metrics are a special case, since the default product counters do not count
+    ///   any specific field but rather the occurrence of the event. As such, there is no value
+    ///   expression, and the field is set to `None`. Semantics of specifying remain undefined at
+    ///   this point.
+    /// - **Distribution** metrics require a numeric value. If the value at the specified path is
+    ///   not numeric, metric extraction will be skipped.
+    /// - **Set** metrics are require a string value, which is then emitted into the set as unique
```
iker-barriocanal:
```suggestion
    /// - **Set** metrics require a string value, which is then emitted into the set as unique
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+    ///   value. Insertion of numbers and other types is undefined.
+    ///
+    /// If the field does not exist, extraction is skipped.
+    #[serde(default)]
+    pub field: Option<String>,
+
+    /// An optional condition to apply before extraction.
```
iker-barriocanal: nit:
```suggestion
    /// An optional condition to meet before extraction.
```
I think `apply` may lead to some ambiguity, while `meet` doesn't.

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+    /// condition will be applied.
+    #[serde(default)]
+    pub tags: Vec<TagSpec>,
+}
+
+/// Mapping between extracted metrics and additional tags to extract.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TagMapping {
+    /// A list of metrics identifiers (MRI) to apply tags to.
```
iker-barriocanal: nit: skipping `resource` may be confusing when relating to `MRI` due to the missing `R`.
```suggestion
    /// A list of metric resource identifiers (MRI) to apply tags to.
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+#[serde(rename_all = "camelCase")]
+pub struct TagSpec {
+    /// The key of the tag to extract.
+    pub key: String,
+
+    /// Path to a field containing the tag's value.
+    ///
+    /// Mutually exclusive from `value`.
+    #[serde(default)]
+    pub field: Option<String>,
```
iker-barriocanal: Is it worth mentioning `FieldValueProvider`, as we do in `MetricSpec.field` above?

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+
+    /// Literal value of the tag.
+    ///
+    /// Mutually exclusive from `field`.
+    #[serde(default)]
+    pub value: Option<String>,
+
+    /// An optional condition to apply before extraction.
```
iker-barriocanal: Same suggestion as above:
```suggestion
    /// An optional condition to meet before extraction.
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+
+impl TagSpec {
+    /// Returns the source of tag values, either literal or a field.
+    pub fn source(&self) -> TagSource<'_> {
+        if let Some(ref field) = self.field {
+            TagSource::Field(field)
+        } else if let Some(ref value) = self.value {
+            TagSource::Literal(value)
+        } else {
+            TagSource::Unknown
```
iker-barriocanal, line 267: What do you think about logging an error in this case, at least for the first few iterations?

iker-barriocanal (approved):
LGTM. Mostly comments on docs and a few non-urgent questions.

relay-dynamic-config/src/metrics.rs
```diff
@@ -124,3 +125,178 @@ impl TransactionMetricsConfig {
...
+    /// possible to list tag keys multiple times, each with different conditions. The first matching
+    /// condition will be applied.
+    #[serde(default)]
+    pub tags: Vec<TagSpec>,
+}
+
+/// Mapping between extracted metrics and additional tags to extract.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TagMapping {
```
iker-barriocanal, line 209: I don't really like `TagMapping`, but don't really have better alternatives. What do you think about `SharedTags`?

relay-dynamic-config/src/project.rs
jan-auer: `Default` is implemented in the same PR by invoking `MetricExtractionConfig::default()`. I prefer to avoid an option here since it's just not needed -- the default state is disabled, so the option is redundant.

relay-dynamic-config/src/metrics.rs
jan-auer: This is part of the spec.

relay-dynamic-config/src/metrics.rs
jan-auer: If we make changes during our first iterations, this error would also show up for customers if they run a Relay with this intermediate version. Since this is future-proofing, it's better if we keep this silent. There can be other tests that ensure our config roundtrips in Sentry.

relay-dynamic-config/src/metrics.rs
jan-auer: We could do `SharedTagMapping`, would that be better?

Note: I might merge this PR sooner, but we can follow up with some cleanup.

PR #2253, opened 2023-06-26
Introduce relay pipedream

mattgauntseo-sentry:
This introduces a new set of GoCD pipelines.

The pipelines produced are:

- deploy-relay
- deploy-us
- deploy-monitor

`deploy-relay` is a simple pipeline that triggers a deployment of relay but doesn't actually deploy anything.
`deploy-us` is the equivalent and deploy-relay and deploy-relay-pops that currently exist. The only difference is that the pops pipeline has been turned into a stage. `deploy-monitor` is a new pipeline that deploys to a single-tenant style environment.

Note: This is labeled as relay-next as I don't expect the ingest team to use it quite yet and I'll work with the team when its ready to test.

### DIFF
--- a/gocd/pipelines/relay-next-monitor.yaml
+++ b/gocd/pipelines/relay-next-monitor.yaml
@@ -1,0 +1,103 @@
+format_version: 10
+pipelines:
+  region-deploy-relay-next-monitor:
+    environment_variables:
+      SENTRY_REGION: monitor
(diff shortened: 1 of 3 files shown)


mattgauntseo-sentry:
The more I think about this, the more issues I can see, this really should follow the jsonnet usage.

PR #2255, opened 2023-06-27
ref(sampling): Remove unused custom rule condition

jan-auer:
The initial implementation of rule evaluation for dynamic sampling was
supposed to support all inbound filters. Since those filters have more
complex conditions, such as `is_legacy_browser` or `is_web_crawler`,
virtual fields and custom conditions were added.

These have since been unused, so this PR removes them without
replacement. This further allows to remove the `client_ip` from the rule
matcher interface.

#skip-changelog


olksdr (approved):
I like red PRs!  :shipit:

PR #2256, opened 2023-06-27
feat(config): Only propagate known feature flags

jjbayer:
This PR
* reverts https://github.com/getsentry/relay/pull/2040/,
* ensures that feature flags are only propagated to downstream Relays if they are known,
* removes unknown features on deserialization,
* fixes a bug in https://github.com/getsentry/relay/pull/2250, which did not deserialize a new feature flag correctly.

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/feature.rs
+++ b/relay-dynamic-config/src/feature.rs
@@ -1,60 +1,79 @@
-use std::borrow::Cow;
+use std::collections::BTreeSet;
 
 use serde::{Deserialize, Serialize};
 
 /// Features exposed by project config.
-#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Hash)]
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub enum Feature {
     /// Enables ingestion of Session Replays (Replay Recordings and Replay Events).
+    #[serde(rename = "organizations:session-replay")]
     SessionReplay,
     /// Enables data scrubbing of replay recording payloads.
+    #[serde(rename = "organizations:session-replay-recording-scrubbing")]
     SessionReplayRecordingScrubbing,
     /// Enables device.class synthesis
     ///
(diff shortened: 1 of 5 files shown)


relay-dynamic-config/src/feature.rs
```diff
@@ -1,60 +1,79 @@
...
+
+    /// Deprecated, still forwarded for older downstream Relays.
+    #[serde(rename = "organizations:transaction-name-mark-scrubbed-as-sanitized")]
+    Deprecated1,
+    /// Deprecated, still forwarded for older downstream Relays.
+    #[serde(rename = "organizations:transaction-name-normalize")]
+    Deprecated2,
+    /// Deprecated, still forwarded for older downstream Relays.
+    #[serde(rename = "organizations:profiling")]
+    Deprecated3,
```
jjbayer, line 34: Now that deprecated features are not forwarded as `Unknown(String)`, we have to list them explicitly (taken from [sentry](https://github.com/getsentry/sentry/blob/7ef1718552effcdf2de5f664882fcb395841ef9f/src/sentry/relay/config/__init__.py#L50-L60)). Listing them explicitly will also encourage us to clean up every once in a while.

relay-dynamic-config/src/feature.rs
```diff
@@ -1,60 +1,79 @@
...
-            "organizations:session-replay" => Feature::SessionReplay,
-            "organizations:session-replay-recording-scrubbing" => {
-                Feature::SessionReplayRecordingScrubbing
-            }
-            "organizations:device-class-synthesis" => Feature::DeviceClassSynthesis,
-            "projects:span-metrics-extraction" => Feature::SpanMetricsExtraction,
-            _ => Feature::Unknown(feature_name.to_string()),
-        })
+        let mut set = BTreeSet::<Feature>::deserialize(deserializer)?;
+        set.remove(&Feature::Unknown);
```
jjbayer, line 58: Without this custom deserializer, we would serialize `["Unknown"]`.

olksdr (approved):

PR #2257, opened 2023-06-27
feat(metrics): Implement generic metric extraction for transactions

jan-auer:
Adds an initial version of generic metrics extraction based on the
previously added schema (#2252). It is limited to transaction events and
evaluates conditions with the fields supported by `FieldValueProvider`.

Since generic metric extraction can operate on all item types, the a new
top-level processing function is added that runs before the more
specific transaction metric extraction. In the future, we will convert
conditional tagging logic and even transaction metric extraction
configuration to the new format when the project config is read, so that
the old functions can be removed.

Since the new function supports only transactions, it reuses the
existing `transaction_metrics_extracted` flag from the processing state.
This is subject to change in a follow-up.

Closes https://github.com/getsentry/relay/pull/2257

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -64,8 +64,6 @@ use crate::actors::project_cache::ProjectCache;
 use crate::actors::upstream::{SendRequest, UpstreamRelay};
 use crate::envelope::{AttachmentType, ContentType, Envelope, Item, ItemType};
 use crate::extractors::RequestMeta;
-use crate::metrics_extraction::sessions::extract_session_metrics;
-use crate::metrics_extraction::transactions::extract_transaction_metrics;
 use crate::metrics_extraction::transactions::types::ExtractMetricsError;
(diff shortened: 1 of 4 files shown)


olksdr (approved):

relay-server/src/actors/processor.rs
```diff
@@ -2138,6 +2132,32 @@ impl EnvelopeProcessorService {
...
+    fn extract_metrics(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
+        let config = match state.project_state.config.metric_extraction {
+            ErrorBoundary::Ok(ref config) if config.is_enabled() => config,
+            _ => return Ok(()),
+        };
+
+        if let Some(event) = state.event.value() {
+            // XXX: Since we only support transactions, we can skip generic metric extraction from
+            // events if transaction metrics have been extracted before.
+            if !state.transaction_metrics_extracted {
```
iker-barriocanal, line 2150: I guess this doesn't consider external relays for now, but I think that's ok.

PR #2258, opened 2023-06-27
ref(filter): Extract URL path into `SchemaDomainPort`

iker-barriocanal:
`SchemaDomainPort` is a structure to extract the schema, domain, and port from a URL. This is used for filtering, but also to extract the domain from a URL in span metrics extraction. Adding the path to the structure means we have all the sections of the URL in it, easing the normalization and string rebuilding.

#skip-changelog

### DIFF
--- a/relay-filter/src/csp.rs
+++ b/relay-filter/src/csp.rs
@@ -58,6 +58,8 @@ pub struct SchemeDomainPort {
     pub domain: Option<String>,
     /// The port of the url.
     pub port: Option<String>,
+    /// The path of the url.
+    pub path: Option<String>,
 }
 
 impl From<&str> for SchemeDomainPort {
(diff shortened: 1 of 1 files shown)


relay-filter/src/csp.rs
```diff
@@ -47,7 +47,7 @@ pub fn should_filter(event: &Event, config: &CspFilterConfig) -> Result<(), Filt
 
 /// A pattern used to match allowed paths.
 ///
-/// Scheme, domain and port are extracted from an url,
+/// Scheme, domain, port and the remaining path are extracted from an URL,
```
iker-barriocanal, line 50: Not sure if I should modify the name of the struct, I think it's not necessary but open to other ideas.

iker-barriocanal (commented):
I believe this is the best way to separate the components of a URL in Relay. If this is not the case and there are other alternatives, we may not need this PR.

relay-filter/src/csp.rs
olksdr: So I would rename this. It becomes a `URI` actually at this point, and maybe that's the name I would use. Or something like `PatternUri` 🤷

olksdr (approved):

iker-barriocanal:
I believe the [`url` crate](https://docs.rs/url/2.4.0/url/) (existing dependency in Relay) solves this issue and thus we no longer need this PR. I'll reopen it in the future if this no longer holds.

Issue #2239, opened 2023-06-22
Support common alerting fields in FieldValueProvider


jan-auer:
Implement support for more fields in the `FieldValueProvider` so that alert and dashboard metrics can specify them in conditions. This issue focuses on the most commonly used fields:

```[tasklist]
### Fields
- [x] measurements
- [x] breakdowns (in-product: spans.*)
- [x] transaction duration
- [x] core attributes: dist, extra.*,
- [x] user.{email, username, ip}
- [x] release.version
- [x] generic tags
```
```[tasklist]
### PRs
- [ ] https://github.com/getsentry/relay/pull/2259
- [ ] https://github.com/getsentry/relay/pull/2276
- [ ] https://github.com/getsentry/relay/pull/2293
```

PR #2259, opened 2023-06-27
feat(protocol): Add access to more fields via FieldValueProvider

jan-auer:
Adds additional fields to the hard-coded field value provider implementation on `Event`. See #2239 for more information.

### DIFF
--- a/relay-general/src/protocol/event.rs
+++ b/relay-general/src/protocol/event.rs
@@ -579,6 +579,28 @@ impl Event {
 
         "unknown"
     }
+
+    /// Returns extra data at the specified path.
+    ///
+    /// The path is evaluated recursively where each path component is joined by a period (`"."`).
+    /// Periods in extra keys are not supported.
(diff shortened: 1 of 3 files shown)


relay-general/src/protocol/event.rs
```diff
@@ -579,6 +579,28 @@ impl Event {
...
     }
+
+    /// Returns extra data at the specified path.
+    ///
+    /// The path is evaluated recursively where each path component is joined by a period (`"."`).
+    /// Periods in extra keys are not supported.
+    pub fn extra_at(&self, path: &str) -> Option<&Value> {
+        let mut path = path.split('.');
+
+        // Get the top-level item explicitly since, since those have a different type
```
iker-barriocanal:
```suggestion
        // Get the top-level item explicitly, since those have a different type
```

iker-barriocanal (approved):

PR #2260, opened 2023-06-27
fix(spans): Don't fallback span description for React spans

iker-barriocanal:
https://github.com/getsentry/relay/pull/2240 introduces a fallback for span descriptions for `http` modules. However, it's applied to all spans that don't fall in `db` or `cache` modules, such as React transactions. This PR fixes that issue.

#skip-changelog


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -295,6 +295,8 @@ fn sanitized_span_description(
         if !module.starts_with("http") {
             return None;
         }
+    } else {
+        return None;
     }
```
olksdr, line 296: It would be shorter and better instead of:

```rust
if let Some(module) = module {
        if !module.starts_with("http") {
            return None;
        }
    } else {
        return None;
    }
```

to try to write something like:

```rust
      if !module?.starts_with("http") {
          return None;
      }
```
I guess it should work.

Notice `?` after `module`, which will try to unwrap the module or return `None`.

olksdr (changes requested):

olksdr (approved):

PR #2261, opened 2023-06-27
ref(spans): Split tag extraction into modules

iker-barriocanal:
NB: this PR will fail until https://github.com/getsentry/relay/pull/2240 is merged, which contains a bug fix discovered in the refactor.

---

Currently, span tag extraction happens in a single method. This means there are too many things going on, making the method very hard to maintain and extend; for example, it's complex to move tag extraction because the order is determined by ownership rules. In an attempt to make the code easier to both maintain and extend, this PR splits this method into several methods in different modules; not only increasing the readability but allowing its extension.

There's some work duplicated as a result of this split, but regex-related work still runs once. Thus, there should be no performance overhead.

There are no functional changes in this PR.

#skip-changelog


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -1,21 +1,22 @@
...
-use relay_filter::csp::SchemeDomainPort;
 use relay_general::protocol::Event;
 use relay_general::types::{Annotated, Value};
 use relay_metrics::{AggregatorConfig, Metric};
 use std::collections::BTreeMap;
 
+mod cache_tags;
+mod common_tags;
+mod db_tags;
+mod http_tags;
```
iker-barriocanal, line 19: I don't really like these module names, but I can't think of better suggestions.

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -108,114 +109,21 @@ pub(crate) fn extract_span_metrics(
...
+                    BTreeMap::new()
                 };
 
-                if let Some(module) = span_module {
-                    span_tags.insert(SpanTagKey::Module, module.to_owned());
-                }
-
-                // TODO(iker): we're relying on the existance of `http.method`
-                // or `db.operation`. This is not guaranteed, and we'll need to
-                // parse the span description in that case.
```
iker-barriocanal, line 127: The second part of this comment, re `db.operation`, is stale and has been removed.

relay-server/src/metrics_extraction/spans/mod.rs
olksdr: I do not really see the reason to do it in first place. 
In my opinion the good separation would be to extract all tags-related code into `tags.rs` module and that's it. Keep it simple, also all code concerning tags is together.

iker-barriocanal: If I understand correctly, your suggestion is to move these into the same file?

olksdr: Exactly! And we might have to rename some of the functions to make it more obvious what's extracted.

iker-barriocanal:
I believe we don't really need this.

PR #2264, opened 2023-06-28
ref(py): Modernize python syntax in relay

asottile-sentry:


jan-auer (approved):
Thanks a lot, @asottile-sentry!

py/CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
+This release requires Python 3.8 or later.
```
asottile-sentry, line 5: this was already true I believe -- I just made it explicit

PR #2265, opened 2023-06-29
feat(spans): Emit exclusive time metric without transaction name

iker-barriocanal:
The transaction name tag is not always used and creates too high cardinality. This PR extracts a new metric without this tag.

#skip-changelog


olksdr (approved):

relay-server/src/metrics_extraction/spans/types.rs
```diff
@@ -52,6 +56,14 @@ impl IntoMetric for SpanMetric {
                 timestamp,
                 span_tag_mapping_to_string_mapping(tags),
             ),
+            SpanMetric::ExclusiveTimeLight { value, tags } => Metric::new_mri(
+                namespace,
+                "exclusive_time_light",
```
iker-barriocanal, line 61: Open to set another name too!

TBS1996:
although it doesn't concern this PR directly, what do you think about adding some documentation on the types you added recently like `SpanMetric` and `SpanTagKey` ?

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -255,6 +255,16 @@ pub(crate) fn extract_span_metrics(
                     }
                     .into_metric(timestamp),
                 );
+
+                let mut reduced_tags = span_tags.clone();
+                reduced_tags.remove(&SpanTagKey::Transaction);
+                metrics.push(
+                    SpanMetric::ExclusiveTimeLight {
+                        value: *exclusive_time,
+                        tags: reduced_tags.clone(),
```
jjbayer: nit: Unnecessary clone?

jjbayer (approved):

iker-barriocanal:
@TBS1996 I'm open to it! What do you think it's the most confusing part of these metrics? Or, what would be the most useful part I could describe?

TBS1996:
It's not that theyre particularly confusing, it's more about in general documenting types to make it faster to grasp whatever part of the codebase youre looking at ^^

PR #2266, opened 2023-06-29
ref(dynamic-config): Skip serializing empty metric extraction fields

jan-auer:
When the schema for metric extraction was added, we forgot to skip
serializing empty fields. This PR adds all fields that are supposed to
be omitted if they are empty.

The fields in `TagMapping` are not skipped since there is currently no
point in a mapping without either.

#skip-changelog


TBS1996 (approved):

olksdr (approved):
lgtm

iker-barriocanal (approved):
🚀

PR #2267, opened 2023-06-29
instr(metrics): Add separate buckets for aggregated spans metrics observability

iker-barriocanal:
Currently, we don't have any observability around how many span metrics we emit, and these go into the `"other"` bucket. This PR adds this observability.

#skip-changelog

### DIFF
--- a/relay-metrics/src/statsd.rs
+++ b/relay-metrics/src/statsd.rs
@@ -174,5 +174,12 @@ pub fn metric_name_tag(value: &str) -> &str {
         return "d:transactions/measurements.*";
     }
 
+    if value.starts_with("s:spans/") {
+        return "s:spans/";
+    }
+    if value.starts_with("d:spans/") {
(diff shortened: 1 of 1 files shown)


relay-metrics/src/statsd.rs
```diff
@@ -174,5 +174,12 @@ pub fn metric_name_tag(value: &str) -> &str {
         return "d:transactions/measurements.*";
     }
 
+    if value.starts_with("s:spans/") {
+        return "s:spans/";
+    }
+    if value.starts_with("d:spans/") {
+        return "d:spans/";
+    }
```
iker-barriocanal, line 182: We have several distribution metrics. I decided to not add the full MRIs to the list above to control the cost further, but I'll revisit this decision if we need more details.

olksdr (approved):

jjbayer (approved):

PR #2268, opened 2023-06-29
chore(metrics): Normalize `db.redis` operation

olksdr:
This change add normalization to the `db.redis` operation, reusing regexes which were done for `cache` op.
Also make sure not to run SQL normalization for this op. 

fixes: https://github.com/getsentry/team-ingest/issues/155

#skip-changelog


jjbayer (approved):

relay-general/src/store/transactions/processor.rs
```diff
@@ -620,8 +620,8 @@ fn scrub_span_description(span: &mut Span) -> Result<(), ProcessingAction> {
 
     let did_scrub = match span.op.value() {
         Some(op) if op.starts_with("http") => scrub_identifiers(&mut scrubbed)?,
-        Some(op) if op.starts_with("db") => scrub_sql_queries(&mut scrubbed)?,
-        Some(op) if op.starts_with("cache") => scrub_cache_keys(&mut scrubbed)?,
+        Some(op) if op.starts_with("cache") || op == "db.redis" => scrub_redis_keys(&mut scrubbed)?,
+        Some(op) if op.starts_with("db") && op != "db.redis" => scrub_sql_queries(&mut scrubbed)?,
```
iker-barriocanal, line 624: Is `&& op != "db.redis"` needed, considering only the first matching arm is used?

iker-barriocanal (approved):

relay-general/src/store/transactions/processor.rs
olksdr: I wanted to be super sure, if code changes, it will be obvious that we do not normalize SQL where we cannot do anything anyway.

PR #2270, opened 2023-06-30
instr(server): Measure ratio between project keys, projects and orgs

jjbayer:
Add two statsd metrics to keep track of the ratio between cached organizations, projects, and project keys (DSNs). This lets us estimate how much memory we could save by pushing parts of the config into the project and organization scope.

ref: https://github.com/getsentry/team-ingest/issues/132

#skip-changelog

### DIFF
--- a/relay-server/src/statsd.rs
+++ b/relay-server/src/statsd.rs
@@ -8,8 +8,8 @@ pub enum RelayGauges {
     /// The ratio between project keys (DSNs) and project IDs, i.e. the average number of keys
     /// per project.
     ProjectCacheKeysPerProject,
-    /// The ratio between project IDs and organization IDS, i.e. the average number of projects per
-    /// org.
+    /// The ratio between project IDs and organization IDs, i.e. the average number of projects per
+    /// organization.
     ProjectCacheProjectsPerOrg,
     /// The number of items currently in the garbage disposal queue.
     ProjectCacheGarbageQueueSize,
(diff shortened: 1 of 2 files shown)


TBS1996 (approved):
lgtm

tangential but what's our policy on  nighly-only experimental api? https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter

relay-server/src/statsd.rs
```diff
@@ -5,6 +5,12 @@ pub enum RelayGauges {
     /// The state of Relay with respect to the upstream connection.
     /// Possible values are `0` for normal operations and `1` for a network outage.
     NetworkOutage,
+    /// The ratio between project keys (DSNs) and project IDs, i.e. the average number of keys
+    /// per project.
+    ProjectCacheKeysPerProject,
+    /// The ratio between project IDs and organization IDS, i.e. the average number of projects per
```
olksdr:
```suggestion
    /// The ratio between project IDs and organization IDs, i.e. the average number of projects per
```

relay-server/src/statsd.rs
```diff
@@ -5,6 +5,12 @@ pub enum RelayGauges {
...
+    /// org.
```
olksdr:
```suggestion
    /// organization.
```

olksdr (approved):

iker-barriocanal (approved):

jjbayer:
> lgtm
> 
> tangential but what's our policy on nighly-only experimental api? https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter

@TBS1996 I feel more comfortable using `stable` and making one exception here than switching to nightly entirely.

PR #2272, opened 2023-06-30
ref(txnames): Treat legacy without `/` as low cardinality

jjbayer:
Since https://github.com/getsentry/relay/pull/2250, we treat transactions without `transaction_info.source` from old SDKs as URLs if they look like URLs, that is: If it contains slashes, we apply all known clustering rules to the transaction name and then declare it as `sanitized`.

There is still a small group of transactions that ends up without a transaction tag on performance metrics, even though it _probably_ does not produce high cardinality: Transactions without `transaction_info.source` from old SDKs that _do not_ contain slashes.

Overview of transaction categories (by `transaction_info.source`) and whether we extract the transaction metrics tag for them:

* Has source
  * ✅ low cardinality (`route`, `component`, `custom`, ...)
  * high cardinality
    * ✅ `url` -> now all marked as `sanitized`
    * ❌ _other_ -> still marked as unparameterized. An unknown transaction source is an SDK bug, so this is fine.
* No source (legacy SDK)
  * ✅ `unknown` - assumed to be low cardinality by `is_high_cardinality_sdk` heuristic.
  * ✅ `None`
    * ✅ Looks like URL (contains `/`) -> marked as `sanitized`.
    * ❌ **Does not look like URL -> handled in this PR**
    
This PR amends the `is_high_cardinality_sdk` heuristic to mark legacy transactions without slashes as `source:unknown`, which makes metrics extraction assume low cardinality and extract the original transaction name as tag.

This PR also contains some cleanup work to simplify the control flow around `unknown` and `null` sources.

### DIFF
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -94,7 +94,7 @@ pub enum AcceptTransactionNames {
 
 impl Default for AcceptTransactionNames {
     fn default() -> Self {
-        Self::Strict
+        Self::ClientBased
     }
 }
 
@@ -108,8 +108,10 @@ pub struct TransactionMetricsConfig {
     pub extract_custom_tags: BTreeSet<String>,
     /// Deprecated in favor of top-level config field. Still here to be forwarded to external relays.
     pub custom_measurements: CustomMeasurementConfig,
-    /// Defines whether URL transactions should be considered low cardinality.
-    pub accept_transaction_names: AcceptTransactionNames,
+    /// Deprecated. Defines whether URL transactions should be considered low cardinality.
+    /// Keep this around for external Relays.
+    #[serde(rename = "acceptTransactionNames")]
+    pub deprecated1: AcceptTransactionNames,
 }
 
 impl TransactionMetricsConfig {
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -79,17 +79,16 @@ pub struct CustomMeasurementConfig {
 /// The version is an integer scalar, incremented by one on each new version.
 const TRANSACTION_EXTRACT_VERSION: u16 = 1;
 
-/// Defines whether URL transactions should be considered low cardinality.
+/// Deprecated. Defines whether URL transactions should be considered low cardinality.
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub enum AcceptTransactionNames {
-    /// For some SDKs, accept all transaction names, while for others, apply strict rules.
-    ClientBased,
-
     /// Only accept transaction names with a low-cardinality source.
-    /// Any value other than "clientBased" will be interpreted as "strict".
-    #[serde(other)]
     Strict,
+
+    /// For some SDKs, accept all transaction names, while for others, apply strict rules.
+    #[serde(other)]
+    ClientBased,
 }
 
 impl Default for AcceptTransactionNames {
--- a/relay-general/src/protocol/event.rs
+++ b/relay-general/src/protocol/event.rs
@@ -14,7 +14,7 @@ use crate::protocol::{
     Breadcrumb, Breakdowns, ClientSdkInfo, Contexts, Csp, DebugMeta, Exception, ExpectCt,
     ExpectStaple, Fingerprint, Hpkp, LenientString, Level, LogEntry, Measurements, Metrics,
     RelayInfo, Request, Span, Stacktrace, Tags, TemplateInfo, Thread, Timestamp, TransactionInfo,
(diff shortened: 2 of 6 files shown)


relay-dynamic-config/src/metrics.rs
```diff
@@ -79,22 +79,21 @@ pub struct CustomMeasurementConfig {
...
+
+    /// For some SDKs, accept all transaction names, while for others, apply strict rules.
+    #[serde(other)]
+    ClientBased,
 }
 
 impl Default for AcceptTransactionNames {
     fn default() -> Self {
-        Self::Strict
+        Self::ClientBased
```
jjbayer, line 96: This flag has been hard coded as `clientBased` by sentry for a while now, time to make it the default and deprecate.

https://github.com/getsentry/sentry/blob/49ce035653ddc9ce9e523f9710bb66fd6aa39a8a/src/sentry/relay/config/__init__.py#L645

relay-general/src/protocol/event.rs
```diff
@@ -538,13 +538,6 @@ pub struct Event {
 }
 
 impl Event {
-    pub fn get_transaction_source(&self) -> &TransactionSource {
-        self.transaction_info
-            .value()
-            .and_then(|info| info.source.value())
-            .unwrap_or(&TransactionSource::Unknown)
-    }
```
jjbayer: This function does not distinguish between `None` and `Some(Unknown)`, even though it makes a semantic difference -> better delete it.

relay-general/src/store/transactions/processor.rs
```diff
@@ -409,20 +406,29 @@ pub fn is_high_cardinality_sdk(event: &Event) -> bool {
...
             .source
             .set_value(Some(TransactionSource::Unknown));
     }
 }
 
+fn is_high_cardinality_transaction(event: &Event) -> bool {
+    let transaction = event.transaction.as_str().unwrap_or_default();
+    // We treat transactions from legacy SDKs as URLs if they contain slashes.
+    // Otherwise, we assume low cardinality.
+    transaction.contains('/') && is_high_cardinality_sdk(event)
```
jjbayer, line 429: This is the functional change of this PR.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -71,57 +71,55 @@ fn extract_geo_country_code(event: &Event) -> Option<String> {
...
             return None;
         }
     };
 
-    // In client-based mode, handling of "unknown" sources depends on the SDK name.
-    // In strict mode, treat "unknown" as high cardinality.
-    let treat_unknown_as_low_cardinality = matches!(
-        accept_transaction_names,
-        AcceptTransactionNames::ClientBased
-    ) && !store::is_high_cardinality_sdk(event);
```
jjbayer, line 121: No need to call `is_high_cardinality_sdk` again (already done in normalization). Just rely on the value of `event.transaction_info.source`.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -71,57 +71,55 @@ fn extract_geo_country_code(event: &Event) -> Option<String> {
     None
 }
 
-fn is_low_cardinality(source: &TransactionSource, treat_unknown_as_low_cardinality: bool) -> bool {
+fn is_low_cardinality(source: Option<&TransactionSource>) -> bool {
     match source {
         // For now, we hope that custom transaction names set by users are low-cardinality.
-        TransactionSource::Custom => true,
+        Some(TransactionSource::Custom) => true,
```
olksdr: I'm thinking if instead of wrapping all this in `Some` we could use early return? 

```rust

let Some(source) = source else { return false; };

```

and then just match `source` as it was before.

olksdr (approved):
Overall LGTM,  can be tested 😄 
Left only 1 suggestion for readability.

tests/integration/test_outcome.py
```diff
@@ -1211,10 +1211,12 @@ def make_envelope(transaction_name):
...
 
-    metrics = metrics_by_name(metrics_consumer, 3)
-    assert (
-        metrics["d:transactions/duration@millisecond"]["tags"]["has_profile"] == "true"
-    )
+    metrics = [
+        m
+        for m in metrics_consumer.get_metrics()
+        if m["name"] == "d:transactions/duration@millisecond"
+    ]
```
iker-barriocanal, line 1218: Should we verify the length of the metrics is the expected one?

iker-barriocanal (approved):

tests/integration/test_outcome.py
jjbayer: done

relay-server/src/metrics_extraction/transactions/mod.rs
jjbayer: done

PR #2273, opened 2023-07-03
fix(server): Skip dynamic sampling if relay doesn't support incoming metrics extraction version

### LINKED ISSUE #1631: Dynamic sampling should respect metrics extraction version
For transaction metrics extraction, we have a versioning mechanism that works as follows: When sentry bumps the version number in the `transactionMetrics` part of the project config, old Relays which do not yet support this version will stop extracting metrics.

In practice, this means that customer Relays will stop extracting metrics, deferring the work to our PoP Relays instead. But these customer Relays will still drop events in dynamic sampling, under the assumption that transaction metrics have already been extracted.

→ Dynamic sampling should also be skipped if the `transactionMetrics.version` goes out of sync. This does mean increased load on our PoP Relays, though.

TBS1996:
#1631 

> For transaction metrics extraction, we have a versioning mechanism that works as follows: When sentry bumps the version number in the transactionMetrics part of the project config, old Relays which do not yet support this version will stop extracting metrics.
> 
> In practice, this means that customer Relays will stop extracting metrics, deferring the work to our PoP Relays instead. But these customer Relays will still drop events in dynamic sampling, under the assumption that transaction metrics have already been extracted.
> 
> → Dynamic sampling should also be skipped if the transactionMetrics.version goes out of sync. This does mean increased load on our PoP Relays, though.


tests/integration/test_outcome.py
```diff
@@ -1074,7 +1077,7 @@ def test_graceful_shutdown(relay, mini_sentry):
         "key_id": 123,
         "outcome": 1,
         "reason": "Sampled:1",
-        "category": 2,
+        "category": 9,
```
TBS1996, line 1082: datacategory changed from `Transaction` to `TransactionIndexed`

is this a problem?

tests/integration/test_filters.py
```diff
@@ -243,6 +243,7 @@ def test_ignore_transactions_filters_are_applied(
     project_id = 42
     project_config = mini_sentry.add_full_project_config(project_id)
     filter_settings = project_config["config"]["filterSettings"]
+    del project_config["config"]["transactionMetrics"]
```
TBS1996: not adding the transactionMetrics config to the basic_project_config made a lot of tests fail, adding it made another set of tests fail, manually deleting it for the latter set is what ended up working.  it feels a bit strange way to do it but let me know your thoughts please

jjbayer (changes requested):
Pressing "Request changes" because deleting `transactionMetrics` config in the test looks wrong to me. I will leave a more thorough review later.

relay-server/src/actors/processor.rs
```diff
@@ -2888,6 +2895,81 @@ mod tests {
...
+                id: Annotated::new(EventId::new()),
+                ty: Annotated::new(EventType::Transaction),
+                transaction: Annotated::new("testing".to_owned()),
+                ..Event::default()
+            };
+
+            let mut project_state = state_with_rule_and_condition(
+                Some(0.0),
+                RuleType::Transaction,
+                SamplingMode::Received,
```
jjbayer: nit: This `SamplingMode` parameter distracts from understanding the intention of the test, can we remove it from the helper function's signature?

CHANGELOG.md
```diff
@@ -5,8 +5,10 @@
 **Internal**:
 
 - Add support for `sampled` field in the DSC and error tagging. ([#2290](https://github.com/getsentry/relay/pull/2290))
+- Skip dynamic sampling if relay doesn't support incoming metrics extraction version. ([#2273](https://github.com/getsentry/relay/pull/2273))
```
```suggestion
- Skip dynamic sampling if relay does not support metrics extraction configuration. ([#2273](https://github.com/getsentry/relay/pull/2273))
```

I would move this to the Bug Fixes section.

tests/integration/test_outcome.py
jjbayer: It actually makes sense to report `Indexed` for sampled transactions, but I don't understand why this changed now. Could you please try removing all the `del`s as suggested above and then see what happens?

tests/integration/test_filters.py
jjbayer: I would prefer additive construction: remove this field from the `basic_project_config`, remove all the `del` statements, and then use a helper function `make_config` that adds the `transactionMetrics` config in the tests that need it.

relay-server/src/actors/processor.rs
TBS1996: `state_with_rule_and_condition` is an old helper though, should i make a new one just to not include `SamplingMode`?

jjbayer: I would just remove  that argument from the existing helper function. It is only called from two locations anyway.

tests/integration/fixtures/mini_sentry.py
```diff
@@ -168,6 +172,24 @@ def full_project_config(self, project_id, dsn_public_key=None, extra=None):
...
+
+        if extra:
+            extra += transaction_metrics_config
+        else:
+            extra = transaction_metrics_config
+
+        return self.add_full_project_config(
+            project_id, dsn_public_key=dsn_public_key, extra=extra
+        )
+
```
jjbayer:
```suggestion
```

tests/integration/test_dynamic_sampling.py
```diff
@@ -467,6 +472,8 @@ def test_multi_item_envelope(mini_sentry, relay, rule_type, event_factory):
 
     # create a basic project config
     config = mini_sentry.add_basic_project_config(project_id)
+    config["config"]["transactionMetrics"] = {"version": 1}
+    # config["config"]["transactionMetricsConfig"] = {"version": 1}
```
jjbayer:
```suggestion
```

tests/integration/test_outcome.py
```diff
@@ -745,6 +747,7 @@ def test_outcome_to_client_report(relay, mini_sentry):
             }
         ],
     }
+    project_config["config"]["transactionMetrics"] = {"version": 1}
```
jjbayer: Already set on line 734.

```suggestion
```

tests/integration/test_outcome.py
jjbayer: It makes sense: We pick either category based on the presence of config:

https://github.com/getsentry/relay/blob/7b34191316429a2061ed83b128fd3f35ea45d30e/relay-server/src/utils/rate_limits.rs#L460-L463

jjbayer (approved):
Please take a look at remaining comments before merging.

PR #2274, opened 2023-07-03
instr(project-cache): Track accessed keys

jjbayer:
Temporary metric:

Gauge the number of unique project keys being read each second.

This will help us decide whether it makes sense to keep all project configs in-memory an uncompressed at all times, or if parts of them may be kept in a more high-latency storage (e.g. disk, compressed memory).

ref: https://github.com/getsentry/team-ingest/issues/132

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_cache.rs
+++ b/relay-server/src/actors/project_cache.rs
@@ -1,6 +1,7 @@
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::error::Error;
 use std::sync::Arc;
+use std::time::Duration;
 
 use relay_common::ProjectKey;
 use relay_config::{Config, RelayMode};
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/project_cache.rs
```diff
@@ -446,6 +447,35 @@ impl Services {
...
+    fn new() -> Self {
+        Self {
+            keys: Default::default(),
+            last_reset: Instant::now(),
+        }
+    }
+
+    fn track(&mut self, project_key: ProjectKey) {
+        let now = Instant::now();
+        if (now - self.last_reset) > Duration::from_secs(1) {
```
olksdr: should we use [`elapsed`](https://doc.rust-lang.org/stable/std/time/struct.Instant.html#method.elapsed) instead here, which  returns duration you could compare?

olksdr (approved):

relay-server/src/actors/project_cache.rs
```diff
@@ -446,6 +447,34 @@ impl Services {
...
+        Self {
+            keys: Default::default(),
+            last_reset: Instant::now(),
+        }
+    }
+
+    fn track(&mut self, project_key: ProjectKey) {
+        if self.last_reset.elapsed() > Duration::from_secs(1) {
+            let count = self.keys.len();
+            // When we get < 1 access per second, this will log results from the past.
```
iker-barriocanal, line 468: Assuming <1 access per second happens when `keys` is empty, why does this log results from the past?

PR #2275, opened 2023-07-03
fix(config): Keep separate project config states per configured public key

### LINKED ISSUE #2120: Multiple Public Keys Per Project
### Environment

- Docker image
  - `getsentry/relay:latest`
  - Digest: `c5e282591c38`

### Steps to Reproduce

1. Setup a project with multiple public keys.

``` json
{
    "slug": "example",
    "publicKeys": [
        {
            "publicKey": "1234567890abcdefghijklmnopqrstuv",
            "isEnabled": true
        },
        {
            "publicKey": "zyxwvutsrqponmlkjihgfedcba098765",
            "isEnabled": true
        }
    ]
}
```
2. Send a test event
``` bash
export SENTRY_DSN='http://zyxwvutsrqponmlkjihgfedcba098765@example.org/1234567'
sentry-cli send-event -m 'A test event'
```
3. Observe the error in the logs
``` logs
public key mismatch on state zyxwvutsrqponmlkjihgfedcba098765
```

### Expected Result

The relay accepts all valid public keys.

### Actual Result

The relay accepts only the first public key configured.

### Insights

The error is thrown only in one line - when checking the request.

https://github.com/getsentry/relay/blob/075170c9f8c4dbe7ff66779f51f2c2ce2687b8e1/relay-server/src/actors/project.rs#L321-L324

The function `self.is_matching_key` compares the keys while using the function `get_public_key_config` to get the key configuration.

https://github.com/getsentry/relay/blob/075170c9f8c4dbe7ff66779f51f2c2ce2687b8e1/relay-server/src/actors/project.rs#L219-L222

The function `get_public_key_config` has all public key configurations available, but just returns the first key.

https://github.com/getsentry/relay/blob/075170c9f8c4dbe7ff66779f51f2c2ce2687b8e1/relay-server/src/actors/project.rs#L142-L144

olksdr:
Makes sure to support the following configuration for Relay in `static` or `proxy` mode described in the docs:  https://docs.sentry.io/product/relay/projects/

Makes sure we distinguish the config associated with configured public key. 

fix: https://github.com/getsentry/relay/issues/2120

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_local.rs
+++ b/relay-server/src/actors/project_local.rs
@@ -107,9 +107,11 @@ async fn load_local_states(
             }
         }
 
-        let arc = Arc::new(sanitized);
-        for key in &arc.public_keys {
-            states.insert(key.public_key, arc.clone());
+        // Keep a separate project state per key.
+        let keys = std::mem::take(&mut sanitized.public_keys);
+        for key in keys {
+            sanitized.public_keys = smallvec::smallvec![key.clone()];
+            states.insert(key.public_key, Arc::new(sanitized.clone()));
         }
     }
 
--- a/relay-server/src/actors/project_local.rs
+++ b/relay-server/src/actors/project_local.rs
@@ -249,4 +249,37 @@ mod tests {
             project_key,
         )
     }
+
+    #[tokio::test]
+    async fn test_multi_pub_static_config() {
+        let temp = tempfile::tempdir().unwrap();
+
+        let tmp_project_file = "111111.json";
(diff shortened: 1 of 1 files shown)


iker-barriocanal (approved):
Do you think it's worth testing this?

olksdr:
> Do you think it's worth testing this?

Yeah, why not. 
Added a test for this case in [699f93f](https://github.com/getsentry/relay/pull/2275/commits/699f93f6931633517103cb1ea85454a630f8966d)

jjbayer (approved):

PR #2276, opened 2023-07-03
feat(metrics): Support access to structured release fields

jan-auer:
Add support for virtual version fields to `FieldValueProvider`, which makes them
accessible to dynamic sampling and generic metrics extraction.

SemVer comparisons are not yet possible on the release version.

### DIFF
--- a/relay-general/src/protocol/event.rs
+++ b/relay-general/src/protocol/event.rs
@@ -568,7 +568,7 @@ impl Event {
             .unwrap_or(false)
     }
 
-    /// Return the identifier of the client SDK if available.
+    /// Returns the identifier of the client SDK if available.
     ///
     /// Sentry's own SDKs use a naming schema prefixed with `sentry.`. Defaults to `"unknown"`.
     pub fn sdk_name(&self) -> &str {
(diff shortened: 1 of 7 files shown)


relay-general/src/protocol/event.rs
```diff
@@ -538,28 +538,39 @@ pub struct Event {
...
 
+    /// Returns `true` if [`modules`](Self::modules) contains the given module.
     pub fn has_module(&self, module_name: &str) -> bool {
         self.modules
             .value()
             .map(|m| m.contains_key(module_name))
             .unwrap_or(false)
     }
 
+    /// Return the identifier of the client SDK if available.
```
olksdr:
```suggestion
    /// Returns the identifier of the client SDK if available.
```

relay-general/src/protocol/event.rs
```diff
@@ -570,6 +581,9 @@ impl Event {
         "unknown"
     }
 
+    /// Return the version of the client SDK if available.
```
olksdr:
```suggestion
    /// Returns the version of the client SDK if available.
```

olksdr (approved):

jan-auer:
I will actually put a bit more work into this right away and make sure that we parse the correct build hash always.

iker-barriocanal (commented):
Thanks for adding documentation!

PR #2279, opened 2023-07-04
build: Update testing dependencies for python 3.11

jan-auer:
#skip-changelog


iker-barriocanal (approved):

.python-version
```diff
@@ -0,0 +1 @@
+3.8.12
```
asottile-sentry: you should be able to use sentry's pypi -- we have wheels for confluent-kafka up to 3.11

`--index-url https://pypi.devinfra.sentry.io/simple`

asottile-sentry (changes requested):

.python-version
asottile-sentry: actually you already have that -- what error message are you hitting? https://github.com/getsentry/relay/blob/5e161d17f9b5af3468b1b044cf60a356cad63ab2/requirements-dev.txt#L1

jan-auer: That's there already: https://github.com/getsentry/relay/blob/7bcda764a0c31b42d172cf8afa1a991cb3c69da0/requirements-dev.txt#L1

However, running with python 3.11.3, we receive this failure:

```
.venv/bin/pip install -U -r requirements-dev.txt
Looking in indexes: https://pypi.devinfra.sentry.io/simple
[...]
ERROR: Could not find a version that satisfies the requirement confluent-kafka==1.9.2 (from versions: 2.1.1)
ERROR: No matching distribution found for confluent-kafka==1.9.2
```

For comparison, when I pin to 3.8, it installs just fine:

```
pip install -r requirements-dev.txt --no-cache-dir
[...]
Collecting confluent-kafka==1.9.2
  Downloading https://pypi.devinfra.sentry.io/wheels/confluent_kafka-1.9.2-cp38-cp38-macosx_11_0_arm64.whl (2.5 MB)
     |████████████████████████████████| 2.5 MB 28.2 MB/s
```

Do we need to upgrade that dependency, instead?

asottile-sentry: yep!  sentry uses `2.1.1` for instance (I think snuba team just standardized on this version)

asottile-sentry (approved):

PR #2281, opened 2023-07-05
feat(kafka): Add support for custom headers in Kafka messages

### LINKED ISSUE #2233: Encode namespace into kafka message header for metrics
via @john-z-yang 

The generic metrics platform would like to implement a kill switch that shuts down certain metrics from a specific namespace (aka use case ID).

Since all (generic) metrics are sent over the same Kafka topic, to determine the namespace of a metric requires us to parse the message, which is still computationally expensive.

We would like to have a lighter weight mechanism to do this. We propose that the ingest platform encode the namespace into the message header so that we do not need to parse the entire message.

olksdr:
Make sure the `MetricKafkaMessage` set the `namespace` header to distinguish between different metrics streams.

fix: https://github.com/getsentry/relay/issues/2233

#skip-changelog


relay-server/src/actors/store.rs
```diff
@@ -953,6 +959,8 @@ struct MetricKafkaMessage {
     org_id: u64,
     project_id: ProjectId,
     name: String,
+    #[serde(skip)]
+    headers: BTreeMap<String, String>,
```
jjbayer: nit: I would find it slightly cleaner to make the `KafkaMessage` type a struct with fields `headers` and a `payload` enum, but up to you.

jjbayer (approved):
Could you add a few assertions to the metrics integration tests that these headers are now set correctly for every namespace? Apart from that, LGTM.

relay-server/src/actors/store.rs
olksdr: I like the idea to extract headers from the message, but also would like each type of `KafkaMessage` have its own set of headers. Refactored it a little bit to make the intention a little bit clearer.

jjbayer (approved):

PR #2282, opened 2023-07-05
Simplify relay deployment label selector

mattgauntseo-sentry:
Removing the deploy_if_production label should not make a difference with the saas deployments from what I can see in the ops repo and it'll make the saas deployment the same as s4s and customer deployments.

I can keep the label selector as is for saas and update it for other deployments if preferred but I believe this is the better approach.

#skip-changelog

### DIFF
--- a/gocd/generated-pipelines/relay-next-monitor.yaml
+++ b/gocd/generated-pipelines/relay-next-monitor.yaml
@@ -63,7 +63,7 @@ pipelines:
                     /devinfra/scripts/k8s/k8stunnel
 
                     /devinfra/scripts/k8s/k8s-deploy.py \
-                        --label-selector="service=relay,deploy_if_production=true" \
+                        --label-selector="service=relay" \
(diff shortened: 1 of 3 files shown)


olksdr (approved):
LGTM !
But I don't have enough context why this label was added in first place. 

I would suggest to get approval from someone from SRE team as well.  And maybe test it before taking those pipelines in production.

mattgauntseo-sentry:
@olksdr Will do - plan is to test each pipedream flow completely before moving forward with it as the primary deployment pipeline.

rgibert (approved):
This looks good to me, @tonyo added these for Relay canaries in Freight so may have more insight.   I don't see Relay canary deploys in GoCD, did they go away or are still to be implemented?

tonyo:
In Freight `deploy_if_*` labels were used to make sure that the canary deployment was updated when production was deployed (technically speaking, "canary" and "production" were two separate deployments).
Assuming that relay canaries are still supported, and that the Ingest team is aware of how things work today -- I'm 👍

PR #2284, opened 2023-07-06
fix(filter): Apply transaction name filter to transaction events

jan-auer:
The transaction name filter is used to drop health check and ping transaction
events. The original PR missed to check for the event type and also dropped
errors. This patch adds a check for the event type and lets all errors through.

### DIFF
--- a/relay-filter/src/transaction_name.rs
+++ b/relay-filter/src/transaction_name.rs
@@ -63,6 +63,7 @@ mod tests {
         for name in transaction_names {
             let event = Event {
                 transaction: Annotated::new(name.into()),
+                ty: Annotated::new(EventType::Transaction),
                 ..Event::default()
             };
             assert!(matches(&event, &patterns), "Did not match `{name}`")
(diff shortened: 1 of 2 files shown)


relay-filter/src/transaction_name.rs
```diff
@@ -2,18 +2,24 @@
...
+use relay_common::EventType;
 use relay_general::protocol::Event;
 
 use crate::{FilterStatKey, GlobPatterns, IgnoreTransactionsFilterConfig};
 
 fn matches(event: &Event, patterns: &GlobPatterns) -> bool {
+    if event.ty.value() != Some(&EventType::Transaction) {
+        return false;
+    }
+
```
olksdr, line 14: Could we add the unit test at least to make sure this won't happen again?

RaduW (approved):

relay-filter/src/transaction_name.rs
RaduW: Will add test, will also fix the currently failing tests (that do not have the event type set)

olksdr (approved):

rcuhljr:
Thank you! Much appreciate the fix as I encountered this in the wild.

PR #2286, opened 2023-07-06
fix(spans): Run SQL scrubber even if already parameterized

jjbayer:
In https://github.com/getsentry/relay/pull/2211, we fixed the problem that noop-scrubbing led to an empty span description:

> In database spans, it's common for SDKs to already send scrubbed queries, and usually Relay won't do further scrubbing, resulting in the span description not being present.

But that change had the side effect that queries partially scrubbed on the client side were not scrubbed any further in relay. With this PR, we restore best-effort scrubbing on those, for two reasons:

* Catch additional parameters that the SDK did not mark as such.
* Better aggregation by replacing `IN (%s, %s, %s)` (variable list length) with `IN (%s)`.

#skip-changelog


relay-general/src/store/transactions/processor.rs
```diff
@@ -2836,6 +2835,13 @@ mod tests {
         "select count() from table where id in (%s)"
     );
 
+    span_description_test!(
+        span_description_scrub_mixed,
+        "UPDATE foo SET a = %s, b = log(e + 5) * 600 + 12345 WHERE true",
+        "db.sql.query",
+        "UPDATE foo SET a = %s, b = log(e + %s) * %s + %s WHERE %s"
```
jjbayer, line 2842: This was taken from a real-world example. If the incoming query is only partially parameterized, IMO we should still attempt to scrub all values from the query.

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -964,20 +964,6 @@ mod tests {
                         "db.operation": "SELECT"
                     }
                 },
-                {
-                    "description": "SELECT 'TABLE'.'col' FROM 'TABLE' WHERE 'TABLE'.'col' = %s",
```
jjbayer, line 968: Removed this example because it is AFAIK uncommon to wrap identifiers in single quotes in SQL.

olksdr (approved):

PR #2287, opened 2023-07-06
fix(span-metrics): Extract DELETE and UPDATE actions for SQL

shruthilayaj:
We want to extract these actions as well
since we're presenting these options as
a filter in the database view

#skip-changelog


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -348,7 +348,7 @@ fn truncate_string(mut string: String, max_bytes: usize) -> String {
 ///
 /// Currently, we're only interested in either `SELECT` or `INSERT` statements.
```
iker-barriocanal:
```suggestion
/// Currently, we're only interested in `SELECT`, `INSERT`, `DELETE` and `UPDATE` statements.
```

iker-barriocanal (approved):
LGTM, thanks for working on Relay!

What do you think of adding a test? Span tests are at the bottom of this file, in `test_extract_span_metrics`. Duplicating one of the spans and updating the description should update two snapshots: one for indexed spans, and one for metrics.

PR #2289, opened 2023-07-07
chore(criterion): Update `criterion` to the current latest: 0.5.1

olksdr:
Bump `criterion` dependency to `0.5` which also removed `atty` crate  addressing https://github.com/getsentry/relay/security/dependabot/66

#skip-changelog


iker-barriocanal (approved):
I haven't checked the `criterion`'s changelog to check for breaking changes.

PR #2290, opened 2023-07-07
feat(dynamic-sampling): Update error tagging by considering new field in DSC

### LINKED ISSUE #52420: Implement updated error tagging
We agreed with SDK teams that they will send us in the DSC a new field called `sampled` that will signal whether the head of the trace was kept or dropped by client side sampling. We would have to support in in Relay error tagging.

iambriccardo:
This PR adds the new `sampled` check for the dsc.

The new logic will work as follows:
* If DSC has no `sampled` field, tagging will not be performed.
* If DSC has `sampled = true`, tagging will be performed and the value will depend on the sampling decision.
* If DSC has `sampled = false`, tagging will be performed and the value will be `false`.

Closes https://github.com/getsentry/sentry/issues/52420


relay-server/src/actors/processor.rs
```diff
@@ -3344,6 +3363,65 @@ mod tests {
...
+        envelope.add_item(mocked_error_item());
+        let sampling_project_state = project_state_with_single_rule(0.0);
+        let new_envelope = process_envelope_with_root_project_state(
+            envelope,
+            Some(Arc::new(sampling_project_state)),
+        );
+        let event = extract_first_event_from_envelope(new_envelope);
+
+        assert!(event.contexts.value().is_none());
+
```
TBS1996: this test is getting very long, what do you think about splitting it ?

iambriccardo: I had two ways of proceeding, clustering it by areas or individual tests. Which proposal would you have?

relay-sampling/src/lib.rs
```diff
@@ -1165,6 +1165,43 @@ mod sample_rate_as_string {
...
+        } else if value == "false" {
+            Ok(Some(false))
+        } else {
+            Err(serde::de::Error::custom(
+                "the `sampled` value can only be `true` or `false`",
+            ))
+        }
+    }
+
+    pub fn serialize<S>(value: &Option<bool>, serializer: S) -> Result<S::Ok, S::Error>
```
jan-auer: - Let's support deserializing from both `bool` and `string`, then always serialize with the default serializer.
- If we want to be extra lenient, we could ignore all errors and make them `None`.

relay-server/src/actors/processor.rs
```diff
@@ -2299,18 +2299,39 @@ impl EnvelopeProcessorService {
...
-            utils::get_sampling_result(
-                self.config.processing_enabled(),
-                None,
-                Some(root_project_state),
-                Some(dsc),
-                None,
-            )
+            // In case the dsc doesn't have a transaction, we don't want to tag it. This can happen
+            // since in tracing without performance we can have a trace that is not made up of
+            // transactions.
```
jan-auer: I'm not clear on the reasoning behind this, could you elaborate? It looks like we should primarily look at the `sampled` flag from the client.

relay-sampling/src/lib.rs
jan-auer: There's a way to deserialize this without going through `Value` by implementing a [visitor](https://serde.rs/impl-deserialize.html#the-visitor-trait). I can help with that tomorrow. 

As for serialization, I'd still ask to serialize directly as boolean. You can remove `fn serialize` entirely and just define `deserialize_with`.

relay-sampling/src/lib.rs
```diff
@@ -1287,7 +1330,9 @@ impl DynamicSamplingContext {
 
         let contexts = event.contexts.value()?;
         let context = contexts.get(TraceContext::default_key())?.value()?;
-        let Context::Trace(ref trace) = context.0 else { return None };
+        let Context::Trace(ref trace) = context.0 else {
```
jan-auer: Sorry for causing conflicts through https://github.com/getsentry/relay/pull/2296. Once you merge or rebase master, this would be removed.

relay-sampling/src/lib.rs
```diff
@@ -3219,6 +3281,80 @@ mod tests {
...
+        let json = r#"
+        {
+            "trace_id": "00000000-0000-0000-0000-000000000000",
+            "public_key": "abd0f232775f45feab79864e580d160b",
+            "user_id": "hello",
+            "sampled": true
+        }
+        "#;
+        let dsc = serde_json::from_str::<DynamicSamplingContext>(json).unwrap();
+        insta::assert_ron_snapshot!(dsc, @r###"
```
jan-auer: Instead of RON, could you assert JSON here? This should roundtrip cleanly, since we require the upstream to parse this again.

relay-server/src/actors/processor.rs
```diff
@@ -2284,61 +2284,33 @@ impl EnvelopeProcessorService {
...
-
-        // We want to get the specific trace context, or we want to create it in case
-        // it is not there.
-        let context =
-            contexts.get_or_insert_with(TraceContext::default_key(), || Trace(Box::default()));
+        // We want to get the trace context, in which we will inject the `sampled` field.
+        let context = event
+            .contexts
+            .get_or_insert_with(Contexts::new)
+            .get_or_insert_with(TraceContext::default_key(), || Trace(Box::default()));
```
jan-auer: Also here please check for the upstream diff, since there's a new `Contexts::insert_or_default()` and the import on enum variant `Trace` has been removed as per style guide.

relay-server/src/actors/processor.rs
```diff
@@ -3343,8 +3322,51 @@ mod tests {
         if let Trace(context) = trace_context {
             assert!(context.sampled.value().unwrap())
         }
+    }
+
+    #[tokio::test]
+    async fn test_error_is_not_tagged_if_inputs_are_invalid() {
```
jan-auer: Should such tests rather be unit tests on one of the util functions? It's great to have a few tests covering the full call hierarchy, though the differences between these tests actually occur within `is_trace_fully_sampled`, if I read this correctly.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -101,6 +101,51 @@ pub fn get_sampling_result(
     SamplingResult::determine_from_sampling_match(sampling_result)
 }
 
+pub fn is_trace_fully_sampled(
```
jan-auer, line 107: Could you add a comment describing the intended outside visible behavior? This would in part pick up on the code comments written below.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -101,6 +101,51 @@ pub fn get_sampling_result(
...
+    root_project_state: Option<&ProjectState>,
+    dsc: Option<&DynamicSamplingContext>,
+) -> Option<bool> {
+    // We want to run dynamic sampling only if we have a root project state and a dynamic
+    // sampling context.
+    //
+    // In reality the dynamic sampling logic supports optional root state and dsc but it will
+    // return keep. In our case having a keep in case of none root state and dsc will be
+    // a problem, since in reality we can't infer anything without trace metadata.
+    let sampling_result = if let (Some(root_project_state), Some(dsc)) = (root_project_state, dsc) {
```
jan-auer: _nit_: This function could become more readable if written with early returns. Even though it is just a minor style change, the cases and control flow discussed in the comment demonstrate that we need to take great care here.


```rust
let dsc = dsc?;
let root_project_state = root_project_state?;

// If the sampled field is not set, we prefer to not tag the error since we have no clue on
// whether the head of the trace was kept or dropped on the client side.
// If the head of the trace was dropped on the client we will immediately mark
// the trace as not fully sampled.
if !dsc.sampled? {
  return Some(false);
}

let sampling_result = get_sampling_result(...);
// rest here...
```

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -101,6 +101,51 @@ pub fn get_sampling_result(
...
+                true => get_sampling_result(
+                    processing_enabled,
+                    None,
+                    Some(root_project_state),
+                    Some(dsc),
+                    None,
+                ),
+                // If the head of the trace was dropped on the client we will immediately mark
+                // the trace as not fully sampled.
+                false => SamplingResult::Drop(MatchedRuleIds(vec![])),
```
jan-auer: This constructs a SamplingResult only to convert it back to a boolean `false` later. You could map this directly here, and map in the other branch alone. This would also qualify for another early return.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -101,6 +101,51 @@ pub fn get_sampling_result(
     SamplingResult::determine_from_sampling_match(sampling_result)
 }
 
+pub fn is_trace_fully_sampled(
+    processing_enabled: bool,
```
jan-auer, line 108: For a follow-up PR: The `processing_enabled` flag doesn't really belong here, and if we trace it in, then it's just to emit an error warning. Could we look at an alternative way or log this unconditionally?

relay-server/src/actors/processor.rs
iambriccardo: Yep, I didn't refactor them since i wanted to wait for your ok first

relay-server/src/utils/dynamic_sampling.rs
iambriccardo: Yeah much better, I personally don't like the usage of `!` and `?` in the same expression but it's a matter of personal taste.

relay-server/src/utils/dynamic_sampling.rs
iambriccardo: We would have to really try to lift that out and see how we can implement sampling either without or with a semantically more meaningful value.

jan-auer (approved):

PR #2291, opened 2023-07-07
chore(deps): Update `terminal_size`

iambriccardo:
Update `terminal_size` to the latest one which also removes old versions `rustix` `linux-raw-sys` `errno`.

#skip-changelog


olksdr (changes requested):
What is the reason for this update? 

Instead we might actually be able to update `terminal_size` which in return should bump `rustix` to `0.37.x`

iambriccardo:
> What is the reason for this update?
> 
> Instead we might actually be able to update `terminal_size` which in return should bump `rustix` to `0.37.x`

Had a chat with Jan since I got `error[E0554]: #![feature] may not be used on the stable release channel` on my machine when compiling with the latest stable Rust version. So we tried to update and it works and he said that updating it should be good.

olksdr:
Please, run 
```bash
> cargo update -p terminal_size
```

to update `terminal_size` instead and please, test it on your machine. And then also update this PR and we will merge the changes.

olksdr (approved):
lgtm

PR #2292, opened 2023-07-07
ref(spans): Don't extract `domain` for Redis spans

iker-barriocanal:
Redis doesn't have the concept of tables, and the extraction of `domain` is unreliable producing unexpected values. This PR removes this tag for now, but we may revisit this decision in the future.

#skip-changelog


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -1012,6 +1014,17 @@ mod tests {
                     "status": "ok",
                     "data": {}
                 },
+                {
+                    "description": "SET 'aaa:bbb:123:zzz' '{\"from json\": \"no\"}'",
```
iker-barriocanal, line 1018: Without the change above, Relay would extract `span.domain=json` from this description.

olksdr (approved):

relay-server/src/metrics_extraction/spans/mod.rs
olksdr: So that means the regex is quite wrong?
I would expect it to extract `SET` or what kind of `action` would it be from the product point of view?

iker-barriocanal: Oops, I meant `domain`. Updated the PR title, description, and message above to reflect it.

PR #2293, opened 2023-07-10
feat(sampling): Add more fields to FieldValueProvider

jan-auer:
Adds more fields to the implementation of `FieldValueProvider for Event`. This
includes previously missed contexts as well as request and response attributes.

Note that `contexts.response.status_code` is not yet aligned with metrics
extraction, as it does not perform a fully normalized lookup. This will be
addressed in a follow-up.

#skip-changelog

### DIFF
--- a/relay-common/src/constants.rs
+++ b/relay-common/src/constants.rs
@@ -316,6 +316,31 @@ pub enum SpanStatus {
     Unauthenticated = 16,
 }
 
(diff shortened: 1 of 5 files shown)


olksdr (approved):

iker-barriocanal (commented):
Thanks for putting in the effort to improve the documentation!

PR #2294, opened 2023-07-10
feat(project_cache): Avoid evicting projects while they cannot be fetched

olksdr:
Handle the eviction of the cached projects differently when they cannot be fetched. 

Let's say:
* there is ongoing incident
* temporary network issues

So, we want to keep those projects and linked envelopes longer. 

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project.rs
+++ b/relay-server/src/actors/project.rs
@@ -448,7 +448,7 @@ impl Project {
         }
     }
 
-    /// Return `true` if the backoff currently in progress.
+    /// Return `true` if the backoff is currently in progress.
     pub fn backoff_started(&self) -> bool {
         self.backoff.started()
     }
--- a/relay-server/src/actors/project.rs
+++ b/relay-server/src/actors/project.rs
@@ -448,6 +448,11 @@ impl Project {
         }
     }
 
+    /// Return `true` if the backoff currently in progress.
+    pub fn backoff_started(&self) -> bool {
+        self.backoff.started()
+    }
+
     /// Returns the next attempt `Instant` if backoff is initiated, or None otherwise.
     pub fn next_fetch_attempt(&self) -> Option<Instant> {
         self.next_fetch_attempt
--- a/relay-server/src/actors/project.rs
+++ b/relay-server/src/actors/project.rs
@@ -536,7 +536,7 @@ impl Project {
         if let Some(next_attempt_at) = self.next_fetch_attempt {
             return next_attempt_at <= Instant::now();
         }
-        // Initiate a first attemt.
+        // Initiate a first attempt.
         self.backoff.next_backoff();
         true
     }
@@ -965,7 +965,7 @@ mod tests {
 
         // This tests that we actually initiate the backoff and the backoff mechanism works:
         // * first call to `update_state` with invalid ProjectState starts the backoff, but since
-        //   it's the first attemt, we get Duration of 0.
+        //   it's the first attempt, we get Duration of 0.
         // * second call to `update_state` here will bumpt the `next_backoff` Duration to somehing
         //   like ~ 1s
         // * and now, by calling `fetch_state` we test that it's a noop, since if backoff is active
--- a/relay-server/src/actors/project.rs
+++ b/relay-server/src/actors/project.rs
@@ -532,10 +532,13 @@ impl Project {
     }
 
     /// Returns `true` if backoff expired and new attempt can be triggered.
-    fn can_fetch(&self) -> bool {
-        self.next_fetch_attempt
-            .map(|next_attempt_at| next_attempt_at <= Instant::now())
-            .unwrap_or(true)
+    fn can_fetch(&mut self) -> bool {
+        if let Some(next_attempt_at) = self.next_fetch_attempt {
+            return next_attempt_at <= Instant::now();
+        }
+        // Initiate a first attemt.
+        self.backoff.next_backoff();
+        true
     }
 
     /// Triggers a debounced refresh of the project state.
@@ -557,7 +560,7 @@ impl Project {
 
         if should_fetch {
             channel.no_cache(no_cache);
-            let attempts = self.backoff.attempt() + 1;
+            let attempts = self.backoff.attempt();
             relay_log::debug!(
                 "project {} state requested {attempts} times",
                 self.project_key
@@ -715,7 +718,7 @@ impl Project {
         // If the state is still invalid, return back the taken channel and schedule state update.
         if state.invalid() {
             self.state_channel = Some(channel);
-            let attempts = self.backoff.attempt() + 1;
+            let attempts = self.backoff.attempt();
             relay_log::debug!(
                 "project {} state requested {attempts} times",
                 self.project_key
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/project_cache.rs
```diff
@@ -467,6 +467,10 @@ struct ProjectCacheBroker {
 }
 
 impl ProjectCacheBroker {
+    /// Hours to wait till the [`ProjectState`] is evictd from the cache if the current incident is
```
```suggestion
    /// Hours to wait till the [`ProjectState`] is evicted from the cache if the current incident is
```

relay-server/src/actors/project_cache.rs
```diff
@@ -467,6 +467,10 @@ struct ProjectCacheBroker {
...
+    /// ongoing, e.g. backoff is initiated and new proejct state cannot be fetched.
+    const INCIDENT_EVICTON_TIMEOUT: u32 = 12;
```
iker-barriocanal: This is a personal preference, but I would word this around the specific problem instead of "incident". Since this issue is specific to the network `NETWORK_BACKOFF_EVICTION_TIME` may be more suitable (you may also want to keep it shorter 😄).

relay-server/src/actors/project.rs
```diff
@@ -448,6 +448,11 @@ impl Project {
         }
     }
 
+    /// Return `true` if the backoff currently in progress.
```
iker-barriocanal:
```suggestion
    /// Return `true` if the backoff is currently in progress.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -526,11 +530,16 @@ impl ProjectCacheBroker {
...
 
-        let expired = self
-            .projects
-            .drain_filter(|_, entry| entry.last_updated_at() + delta <= eviction_start);
+        let expired = self.projects.drain_filter(|_, entry| {
+            // If for some reasons the project cannot be updated, we avoid evicting its envelopes
+            // for the next 12 hours.
+            if entry.backoff_started() {
+                delta *= Self::INCIDENT_EVICTON_TIMEOUT
+            }
```
iker-barriocanal: How does this behave when a project isn't updated the first time because the response didn't get in time?

iker-barriocanal (changes requested):
Should we get into "incident mode" if Relay makes a single request that doesn't get a timely response? What do you think if we consider a threshold of requests and/or time before getting into "incident mode"?

Should the "incident mode" be part of the state of Relay and not a project? What do you think of the following suggestions?
- If Relay can't fetch any project config for >30 minutes, projects of new envelopes should already be in "incident mode" even if no requests were made yet.
- If Relay doesn't manage to get the config for a single project (and manages to get it for the rest), adding an exponential back-off may not help.

I think we should consider testing some use cases before we move on with updating the back-off.

olksdr:
> Should we get into "incident mode" if Relay makes a single request that doesn't get a timely response? What do you think if we consider a threshold of requests and/or time before getting into "incident mode"?

In my opinion it does not make sense to complicate the setup. This patch only introduces following 
* if the project state cannot be fetched , do not GC it and all the spooled envelopes
* only if the state can be fetched and was not used / updated as before we will GC it


> Should the "incident mode" be part of the state of Relay and not a project? What do you think of the following suggestions?
> 
> * If Relay can't fetch any project config for >30 minutes, projects of new envelopes should already be in "incident mode" even if no requests were made yet.
> * If Relay doesn't manage to get the config for a single project (and manages to get it for the rest), adding an exponential back-off may not help.
> 
> I think we should consider testing some use cases before we move on with updating the back-off.

Each project is independent here, and can happen that a batch of the projects could not be fetched at this moment, but another batch could. And so removing those staled projects does not make sense yet, but also it's not a full outage either. We basically keep spooled envelopes till we fetch any states (it can be either deleted or missing or valid one), but the backoff will be reset and we fallback to the normal mode and the project gets evicted as it's now.

relay-server/src/actors/project_cache.rs
```diff
@@ -530,15 +526,14 @@ impl ProjectCacheBroker {
     /// [`super::project_upstream`].
     fn evict_stale_project_caches(&mut self) {
         let eviction_start = Instant::now();
-        let mut delta = 2 * self.config.project_cache_expiry() + self.config.project_grace_period();
+        let delta = 2 * self.config.project_cache_expiry() + self.config.project_grace_period();
 
+        // Remove only unsed project states.
```
```suggestion
        // Remove only unused project states.
```

iker-barriocanal (approved):
🚀

relay-server/src/actors/project_cache.rs
```diff
@@ -528,9 +528,13 @@ impl ProjectCacheBroker {
         let eviction_start = Instant::now();
         let delta = 2 * self.config.project_cache_expiry() + self.config.project_grace_period();
 
-        let expired = self
-            .projects
-            .drain_filter(|_, entry| entry.last_updated_at() + delta <= eviction_start);
+        // Remove only unused project states.
+        //
+        // If the project cannot be fetched because of the network outage, etc., still keep it
+        // and all spooled envleopes if any.
```
```suggestion
        // and all spooled envelopes if any.
```

relay-server/src/actors/project.rs
```diff
@@ -527,10 +532,13 @@ impl Project {
...
     /// Returns `true` if backoff expired and new attempt can be triggered.
-    fn can_fetch(&self) -> bool {
-        self.next_fetch_attempt
-            .map(|next_attempt_at| next_attempt_at <= Instant::now())
-            .unwrap_or(true)
+    fn can_fetch(&mut self) -> bool {
+        if let Some(next_attempt_at) = self.next_fetch_attempt {
+            return next_attempt_at <= Instant::now();
+        }
+        // Initiate a first attemt.
```
```suggestion
        // Initiate a first attempt.
```

jjbayer (approved):
Any way we can test this within the PR? An integration test in `test_projectconfigs.py`, maybe?

olksdr:
> Any way we can test this within the PR? An integration test in `test_projectconfigs.py`, maybe?

yeah. was planning to add some tests, will do it in this PR then.

olksdr:
@joris I've added a unit test in [`3e51e6c`](https://github.com/getsentry/relay/pull/2294/commits/3e51e6c8b0cbd9da0a9a8c6e4d5bf3448d0e4681) to actually show use-case for this new patch and how it would work in case of the problem with fetching the project config.

relay-server/src/actors/project_cache.rs
```diff
@@ -1106,4 +1116,54 @@ mod tests {
...
+        tokio::time::sleep(Duration::from_secs(2)).await;
+
+        // One of the project will be removed.
+        broker.evict_stale_project_caches();
+
+        // The `project1` is removed, since the state was never requested, so there is no attempts there
+        // either, and we know it's definitely expired. But we have tried to fetch `project2` and
+        // failed so far, so the backoff is in progress.
+        assert_eq!(broker.projects.len(), 1);
+        assert!(broker.projects.get(&key2).is_some());
```
iker-barriocanal, line 1167: I'm a bit confused by this. Where have we tried requesting project 2 but not project 1?

olksdr: It's few lines above , line `1152` - this is were the `project2` was requested.

iker-barriocanal: Oh dang, I missed it. Thanks for pointing it out!

iker-barriocanal (approved):

PR #2297, opened 2023-07-11
ref(server): Change internal flagging for metric extraction

jan-auer:
Restructures the signatures of the metric extraction functions so that there are
distinct top-level functions that can be called and tested independently of each
other. These functions are:

 - generic metric extraction
 - transaction metric extraction
 - conditional tagging for transaction metrics
 - span metric extraction

All of these functions are called from a single place `extract_metrics` in the
envelope processor. The transaction metrics flag is checked a single time.

#skip-changelog
Closes https://github.com/getsentry/team-ingest/issues/163


relay-server/src/actors/processor.rs
```diff
@@ -2064,83 +2057,74 @@ impl EnvelopeProcessorService {
...
-        }
+    /// Extract metrics from all envelope items.
+    ///
+    /// Caveats:
+    ///  - This functionality is incomplete. At this point, extraction is implemented only for
+    ///    transaction events.
+    fn extract_metrics(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
+        // TODO: Make span metrics extraction immutable
+        if let Some(event) = state.event.value_mut() {
+            if state.transaction_metrics_extracted {
```
jjbayer: Should we go ahead and rename this field to `event_metrics_extracted`?

relay-server/src/actors/processor.rs
```diff
@@ -2064,83 +2057,74 @@ impl EnvelopeProcessorService {
...
-    ///  - This functionality is incomplete. At this point, extraction is implemented only for
-    ///    transaction events.
-    ///  - This function must run BEFORE `extract_transaction_metrics` to avoid double-extraction.
-    fn extract_metrics(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
-        let config = match state.project_state.config.metric_extraction {
-            ErrorBoundary::Ok(ref config) if config.is_enabled() => config,
-            _ => return Ok(()),
-        };
+                    state.extracted_metrics.extend(extracted);
+                    state.transaction_metrics_extracted |= true;
```
jjbayer: I think this makes sense in preparation of replacing transaction metrics extraction (partially) into `extract_event_metrics`. But we have to be careful not to enable `extract_event_metrics` without `extract_transaction_metrics` (e.g. in self-hosted).

jjbayer (approved):
LGTM. Integration test fails because we now set the flag even if only `extract_event_metrics` runs.

relay-server/src/actors/processor.rs
jan-auer: Yeah, let's go ahead with this.

relay-server/src/actors/processor.rs
jan-auer: Makes sense, this is a great callout. Currently, we have this coupling in Sentry [here](https://github.com/getsentry/sentry/blob/27ba845bb6b9e4f5169d0513966ab479435e2495/src/sentry/relay/config/__init__.py#L376). Metric extraction config is only generated if transaction metrics are generated.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -231,194 +229,177 @@ fn extract_universal_tags(event: &Event, config: &TransactionMetricsConfig) -> C
...
-            has_profile,
-        },
-        project_metrics,
-        sampling_metrics,
-    )?;
-
-    if extract_spans_metrics {
-        extract_span_metrics(aggregator_config, event, project_metrics)?;
-    }
+/// TODO(ja): Doc
```
olksdr, line 232: Are you going to address this TODO in the followup ?

relay-server/src/actors/processor.rs
```diff
@@ -220,16 +221,8 @@ fn has_unprintable_fields(event: &Annotated<Event>) -> bool {
...
-struct ExtractedMetrics {
-    /// Metrics associated with the project that the transaction belongs to.
-    project_metrics: Vec<Metric>,
-    /// Metrics associated with the sampling project (a.k.a. root or head project)
-    /// which started the trace. See [`ProcessEnvelopeState::sampling_project_state`].
-    sampling_metrics: Vec<Metric>,
-}
-
 impl ExtractedMetrics {
+    // TODO(ja): Move
```
olksdr, line 225: Is this TODO for the followup?

olksdr (commented):
There are still few TODOs left, like adding docs or moving things around. 
Are you going to address those in the followups?

PR #2300, opened 2023-07-11
feat(services): Create service for global configs

iker-barriocanal:
Creates a new service to handle Relay global configs. This service is a WIP and only implements the global config forwarding functionality to the project cache service (which doesn't do anything with the config). Future PRs will implement the config fetching and application in the project cache service, where the `TODO` placeholders are. PRs with functional changes will also contain tests.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -1,0 +1,60 @@
+use std::sync::Arc;
+
+use relay_dynamic_config::GlobalConfig;
+use relay_system::{Addr, Interface, Service};
+use tokio::sync::mpsc;
+
+use crate::actors::project_cache::ProjectCache;
+
+// No messages are accepted for now.
+#[derive(Debug)]
+pub enum GlobalConfigMessage {}
+
+impl Interface for GlobalConfigMessage {}
+
+/// Helper type to forward global config updates.
+pub struct UpdateGlobalConfig {
+    global_config: Arc<GlobalConfig>,
+}
+
+/// Service implementing the [`GlobalConfig`] interface.
+///
+/// The service is responsible to fetch the global config appropriately and
+/// forward it to the services that require it.
+#[derive(Debug)]
+pub struct GlobalConfigService {
+    project_cache: Addr<ProjectCache>,
+}
+
+impl GlobalConfigService {
+    pub fn new(project_cache: Addr<ProjectCache>) -> Self {
+        GlobalConfigService { project_cache }
+    }
+
+    /// Forwards the given global config to the services that require it.
+    fn update_global_config(&mut self, new_config: UpdateGlobalConfig) {
+        self.project_cache.send(new_config.global_config);
+    }
+}
+
+impl Service for GlobalConfigService {
+    type Interface = GlobalConfigMessage;
+
+    fn spawn_handler(mut self, _rx: relay_system::Receiver<Self::Interface>) {
+        tokio::spawn(async move {
+            relay_log::info!("global config started");
+
+            let (_config_tx, mut config_rx) = mpsc::unbounded_channel();
+            loop {
+                tokio::select! {
+                    biased;
+
+                    Some(message) = config_rx.recv() => self.update_global_config(message),
+                    // TODO(iker): request global config
+                    else => break,
+                }
+            }
+            relay_log::info!("global config stopped");
+        });
+    }
+}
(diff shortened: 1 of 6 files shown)


relay-server/src/actors/global_config.rs
```diff
@@ -0,0 +1,60 @@
...
+            let (_config_tx, mut config_rx) = mpsc::unbounded_channel();
+            loop {
+                tokio::select! {
+                    biased;
+
+                    Some(message) = config_rx.recv() => self.update_global_config(message),
+                    // TODO(iker): request global config
+                    else => break,
+                }
+            }
```
jjbayer, line 56: What other service will hold `_config_tx`? Wouldn't this be a message that we define in `UpdateGlobalConfig`?

On a higher level, what's the purpose of a service that only forwards a message? Couldn't whoever sends the update request send it straight to the `ProjectCache`?

Or will this service fetch the project config in a cron-job-like manner, and then call `update_global_config` itself?

TL;DR I don't really understand what's going on in this PR. Could you provide more information on the design, or add a bit more functionality to the PR so that it explains the design itself?

relay-server/src/actors/global_config.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+impl Service for GlobalConfigService {
+    type Interface = GlobalConfigMessage;
+
+    fn spawn_handler(mut self, _rx: relay_system::Receiver<Self::Interface>) {
+        tokio::spawn(async move {
+            relay_log::info!("global config started");
+
+            let (_config_tx, mut config_rx) = mpsc::unbounded_channel();
```
iker-barriocanal, line 47: The idea was to fetch the global configs and send them through this channel, so that fetching and forwarding to other services was done in two separate steps. It turns out this service is simpler than that, so the [next PR](https://github.com/getsentry/relay/pull/2308) removes this channel.

relay-server/src/actors/global_config.rs
iker-barriocanal: Added a small clarification in https://github.com/getsentry/relay/pull/2300/files#r1262827148, let me know if it's still unclear. My idea was to separate the work into fetching and forwarding the global config, and this PR does the forwarding.

relay-server/src/actors/global_config.rs
```diff
@@ -0,0 +1,60 @@
...
+use relay_system::{Addr, Interface, Service};
+use tokio::sync::mpsc;
+
+use crate::actors::project_cache::ProjectCache;
+
+// No messages are accepted for now.
+#[derive(Debug)]
+pub enum GlobalConfigMessage {}
+
+impl Interface for GlobalConfigMessage {}
```
olksdr, line 13: You do not have to write this, as there an implementation of the `Interface` for services without messages, see 
https://github.com/getsentry/relay/blob/22a026f4720f964e33144c7fb78b6ba4a6b1bfee/relay-system/src/service.rs#L119-L120
The interface in your service will be just a simple unit `()`

relay-server/src/actors/global_config.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Helper type to forward global config updates.
+pub struct UpdateGlobalConfig {
+    global_config: Arc<GlobalConfig>,
```
olksdr, line 17: Why does this have to be `Arc`?

relay-server/src/actors/global_config.rs
```diff
@@ -0,0 +1,60 @@
...
+
+impl GlobalConfigService {
+    pub fn new(project_cache: Addr<ProjectCache>) -> Self {
+        GlobalConfigService { project_cache }
+    }
+
+    /// Forwards the given global config to the services that require it.
+    fn update_global_config(&mut self, new_config: UpdateGlobalConfig) {
+        self.project_cache.send(new_config.global_config);
+    }
```
olksdr, line 37: I already mentioned this. But what if we turn this around and let `ProjectCache` actually pull the config from this service, instead of the `GlobalConfig` being pulled? This way `ProjectCache` only pull the config it needs one.

olksdr (changes requested):
Overall it's quote difficult to review PR like this, which does not have any purpose. I would prefer to get bigger PR but with working solution in one place. This allows get the entire picture of what's happening and why. 

Looking at this PR, it feels like we do not need this service at all, and we could let the ProjectCache request the global config directly without needing all this boilerplate

TBS1996:
i resumed on your work here: https://github.com/getsentry/relay/pull/2320 😃

iker-barriocanal:
Closing in favor of https://github.com/getsentry/relay/pull/2320.

PR #2301, opened 2023-07-11
Update jsonnet libs and add rollback pipeline

mattgauntseo-sentry:
This PR does the following:

- Updates the jsonnet libs to new version (there is a cleanup on the libs themselves, hence the directory change)
- The order of the pipedream has changed so s4s is deployed before US
- A rollback pipeline is being added which will hopefully make it faster and easier to rollback pipedream

#skip-changelog

### DIFF
--- a/gocd/generated-pipelines/relay-next-monitor.yaml
+++ b/gocd/generated-pipelines/relay-next-monitor.yaml
@@ -7,8 +7,8 @@ pipelines:
     group: relay-next
     lock_behavior: unlockWhenFinished
     materials:
-      deploy-relay-next-us-pipeline-complete:
-        pipeline: deploy-relay-next-us
+      deploy-relay-next-pipeline-complete:
+        pipeline: deploy-relay-next
         stage: pipeline-complete
       relay_repo:
         branch: master
--- a/gocd/generated-pipelines/relay-next-us.yaml
+++ b/gocd/generated-pipelines/relay-next-us.yaml
@@ -1,14 +1,14 @@
 format_version: 10
 pipelines:
   deploy-relay-next-us:
-    display_order: 1
+    display_order: 3
     environment_variables:
       SENTRY_REGION: us
     group: relay-next
     lock_behavior: unlockWhenFinished
     materials:
-      deploy-relay-next-pipeline-complete:
-        pipeline: deploy-relay-next
+      deploy-relay-next-monitor-pipeline-complete:
+        pipeline: deploy-relay-next-monitor
         stage: pipeline-complete
       relay_repo:
         branch: master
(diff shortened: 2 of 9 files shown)


gocd/generated-pipelines/relay-next-us.yaml
```diff
@@ -1,14 +1,14 @@
 format_version: 10
 pipelines:
   deploy-relay-next-us:
-    display_order: 1
+    display_order: 3
```
olksdr, line 4: How did this get generated? Does it come from `libs` you updated to?

gocd/generated-pipelines/relay-next-monitor.yaml
```diff
@@ -7,8 +7,8 @@ pipelines:
     group: relay-next
     lock_behavior: unlockWhenFinished
     materials:
-      deploy-relay-next-us-pipeline-complete:
-        pipeline: deploy-relay-next-us
+      deploy-relay-next-pipeline-complete:
+        pipeline: deploy-relay-next
```
olksdr, line 11: also here, does this come from `libs`?

gocd/templates/relay.jsonnet
```diff
@@ -12,6 +12,10 @@ local pipedream_config = {
       destination: 'relay',
     },
   },
+  rollback: {
+    material_name: 'relay_repo',
+    stage: 'deploy-production',
+  },
```
olksdr, line 18: Only this this like adding new rollback generated pipeline.

olksdr (approved):
Left few questions to the generated files. 

Do those changes come from the new version of `libs` you upgraded to? Asking because there is no. changes in the local `realy.{jsonnet,libsonnet}` files but there suddenly changes  to `display-order` or renaming the materials and pipelines.

gocd/generated-pipelines/relay-next-us.yaml
mattgauntseo-sentry: yes it comes from the libs. The order becomes:

1. deploy-<service>
2. rollback-<service>
3. deploy-<service>-<first region>
4. deploy-<service>-<second region> 
5. .....

gocd/generated-pipelines/relay-next-monitor.yaml
mattgauntseo-sentry: Yeah, as mentioned in the PR description, the order of the pipedream is changed and this is how GoCD manages the ordering - via dependencies on pipelines.

gocd/templates/relay.jsonnet
mattgauntseo-sentry: Yep, these values are from the `gocd/templates/libs/relay.libsonnet` file, the `relay_repo` is the material name used in each repo for github.com/getsentry/relay, which the rollback pipeline used to lookup the SHA we want to rollback to and the `stage` is the name of the stage the rollback pipeline will re-run during a rollback.

I'm putting together some docs on pipedream and rollbacks here: https://www.notion.so/sentry/Pipedreams-in-GoCD-with-Jsonnet-430f46b87fa14650a80adf6708b088d9?pvs=4#0fdf3335c8154303ab9dda719f44c697

Not complete, but I'm intending to link to the docs in the relay.jsonnet file when switching over to pipedream next week.

PR #2304, opened 2023-07-13
ref(spans): Move tag extraction to normalization

jjbayer:
In preparation of making metrics extraction completely configurable, move all modifications of the event payload to the normalization stage.

This PR has a large diff but the only functional changes are:

* We now use `Url::parse` instead of `SchemaDomainPort` to parse URLs, to prevent adding `relay-filter` as a dependency to `relay-general`.
* Because normalization runs in every relay in a chain, span tags in payloads can theoretically diverge from the ones in metrics if the metrics were extracted by an older relay.

ref: https://github.com/getsentry/team-ingest/issues/142, https://github.com/getsentry/team-ingest/issues/137


relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -49,190 +46,28 @@ pub(crate) fn extract_span_metrics(
...
-        databag.insert(SpanTagKey::User, user);
-    }
-
-    // Collect the shared tags for all the metrics and spans on this transaction.
-    let mut shared_tags = BTreeMap::new();
-
-    if let Some(environment) = event.environment.as_str() {
-        shared_tags.insert(SpanTagKey::Environment, environment.to_owned());
-    }
+    // TODO: do not extract Release and User tags.
```
jjbayer:
```suggestion
```

relay-cabi/src/processing.rs
```diff
@@ -132,8 +132,9 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
         transaction_name_config: Default::default(), // only supported in relay
         is_renormalize: config.is_renormalize.unwrap_or(false),
         device_class_synthesis_config: false, // only supported in relay
-        scrub_span_descriptions: false,
+        enrich_spans: false,
```
jjbayer, line 135: This flag now controls both span description scrubbing and tag extraction.

relay-general/Cargo.toml
```diff
@@ -18,6 +18,7 @@ enumset = "1.0.4"
 hmac = "0.12.1"
 itertools = "0.10.5"
 maxminddb = "0.23.0"
+md5 = "0.7.0"
```
jjbayer, line 21: Necessary to compute the group hash of a span.

relay-server/src/metrics_extraction/spans/snapshots/relay_server__metrics_extraction__spans__tests__extract_span_metrics.snap
```diff
@@ -320,13 +320,13 @@ expression: event.value().unwrap().spans
                 "http",
             ),
             "span.description": String(
-                "POST *.domain.tld:targetport/<unparameterized>",
+                "POST http://*/api/hi",
```
jjbayer, line 323: This is a side effect of changing the test input: `"targetport"` is not a valid port parsable by `Url`, so I replaced it with 1234. Now, the entire port + domain get scrubbed by `scrub_identifiers`. I will follow up on this in a later PR.

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -0,0 +1,693 @@
...
+
+use crate::macros::derive_fromstr_and_display;
+use crate::protocol::{Event, Span, TraceContext};
+use crate::store::utils::{
+    extract_http_status_code, extract_transaction_op, get_eventuser_tag, http_status_code_from_span,
+};
+use crate::types::Annotated;
+// use relay_filter::csp::SchemeDomainPort;
+// use relay_metrics::{AggregatorConfig, Metric};
+use std::collections::BTreeMap;
```
jan-auer: _nit_: Sort imports as per style guide.

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -0,0 +1,693 @@
+//! Logic for persisting items into `span.data` fields.
+//! These are then used for metrics extraction.
+
+use itertools::Itertools;
```
jan-auer, line 5: _nit_: What is itertools used for? I wasn't able to spot it while scanning through the file and wonder if we can go without.

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -0,0 +1,693 @@
...
+
+    let mut sanitized = String::new();
+
+    if let Some(transaction_method) = action {
+        sanitized.push_str(&format!("{transaction_method} "));
+    }
+    if let Some(domain) = domain {
+        sanitized.push_str(&format!("{domain}/"));
+    }
+    sanitized.push_str("<unparameterized>");
```
jan-auer: _nit_: This can be written without the intermediate allocations.

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -0,0 +1,693 @@
...
+fn span_op_to_category(op: &str) -> Option<&str> {
+    let category =
+        // General
+        if op.starts_with("mark") {
+            "mark"
+        }
+        //
+        // Browser
+        // `ui*` mapped in JS frameworks
+        else if op.starts_with("pageload") {
```
jan-auer, line 402: _nit_: As per convention, our span ops are always separated with a `'.'`. This means you can call split_once and then match on it. 

I'm pretty sure we have similar code somewhere else already, probably in breakdown computation.

relay-general/src/store/normalize.rs
```diff
@@ -745,12 +746,37 @@ pub struct LightNormalizationConfig<'a> {
     pub transaction_name_config: TransactionNameConfig<'a>,
     pub is_renormalize: bool,
     pub device_class_synthesis_config: bool,
-    pub scrub_span_descriptions: bool,
+    pub enrich_spans: bool,
     pub light_normalize_spans: bool,
+    pub max_tag_value_length: usize, // TODO: move span related fields into separate config.
```
iker-barriocanal: nit: `max_tag_value_size`, since the limit is in bytes.

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -0,0 +1,693 @@
...
+use url::Url;
+
+use crate::macros::derive_fromstr_and_display;
+use crate::protocol::{Event, Span, TraceContext};
+use crate::store::utils::{
+    extract_http_status_code, extract_transaction_op, get_eventuser_tag, http_status_code_from_span,
+};
+use crate::types::Annotated;
+// use relay_filter::csp::SchemeDomainPort;
+// use relay_metrics::{AggregatorConfig, Metric};
```
iker-barriocanal:
```suggestion
```

jan-auer (approved):
metrics extraction should now work with a plain `&Event` instead of an `&mut Event`, could you check?

relay-general/src/store/transactions/processor.rs
```diff
@@ -24,29 +25,32 @@ pub struct TransactionNameConfig<'r> {
 pub struct TransactionsProcessor<'r> {
     name_config: TransactionNameConfig<'r>,
     span_desc_rules: Vec<SpanDescriptionRule>,
-    scrub_span_descriptions: bool,
+    enrich_spans: bool,
+    max_tag_value_length: usize,
```
iker-barriocanal: I think this should also be `_size`?

iker-barriocanal (commented):
>We now use Url::parse instead of SchemaDomainPort to parse URLs
🚀

relay-general/src/store/normalize.rs
```diff
@@ -745,12 +746,37 @@ pub struct LightNormalizationConfig<'a> {
...
+            breakdowns_config: Default::default(),
+            normalize_user_agent: Default::default(),
+            transaction_name_config: Default::default(),
+            is_renormalize: Default::default(),
+            device_class_synthesis_config: Default::default(),
+            enrich_spans: Default::default(),
+            light_normalize_spans: Default::default(),
+            max_tag_value_length: usize::MAX,
+            span_description_rules: Default::default(),
+            geoip_lookup: Default::default(),
```
olksdr: Would it make sense to write something like 
```suggestion
            max_tag_value_length: usize::MAX,
            ..Default::default()

```

relay-general/src/store/normalize/span/description.rs
```diff
@@ -10,7 +10,9 @@ use crate::store::regexes::{
 use crate::store::{scrub_identifiers, scrub_identifiers_with_regex, SpanDescriptionRule};
 use crate::types::{Annotated, ProcessingAction, ProcessingResult, Remark, RemarkType, Value};
 
-// TODO: docs
+/// Attempt to replace identifiers in the span description with placeholders.
```
```suggestion
/// Attempts to replace identifiers in the span description with placeholders.
```

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -0,0 +1,693 @@
...
+            shared_tags
+                .clone()
+                .into_iter()
+                .chain(tags.into_iter())
+                .map(|(k, v)| (k.to_string(), Annotated::new(v.into()))),
+        );
+    }
+}
+
+/// Extract tags shared by every span.
```
olksdr:
```suggestion
/// Extracts tags shared by every span.
```

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -0,0 +1,693 @@
...
+    }
+
+    if let Some(transaction_http_status_code) = extract_http_status_code(event) {
+        tags.insert(SpanTagKey::HttpStatusCode, transaction_http_status_code);
+    }
+
+    tags
+}
+
+/// Write fields into [`Span::data`].
```
olksdr:
```suggestion
/// Writes fields into [`Span::data`].
```

olksdr (approved):
:shipit:

relay-general/src/store/normalize/span/tag_extraction.rs
jjbayer: It's used to `.join()` an iterator [into a string](https://github.com/getsentry/relay/pull/2304/files#diff-53f81fab03a0fb9fa97597b16f42b28ec40c153b47ee77f03b67580f89081302R550).

relay-general/src/store/normalize/span/tag_extraction.rs
jjbayer: Will do this in a follow-up.

relay-general/src/store/normalize.rs
jjbayer: `..Default::default()` takes the values from the containing type's default, not the field's default, so unfortunately not.

PR #2308, opened 2023-07-13
feat(global-config): Request Relay global config from upstream

iker-barriocanal:
Requires https://github.com/getsentry/sentry/pull/52421.

---

Requests Relay's global config from upstream periodically, every few seconds.

### Approaches

1. The global config service accepts config update requests. The project cache service requests the global config to the project config service periodically, and the global config service fetches it and forwards it to the project cache service. Creating new services that require the global config would also need to send messages to request the global config.
2. The global config service proactively fetches the global config from upstream. The project cache service receives the new global config when it's updated. Creating new services that require the global config should be provided during global config service initialization.

Beyond what the above points mention, I don't see any major pros/cons of either approach. For Relay instances that don't receive much traffic, option 1 will generate more network requests than option 2. On the other hand, option 2 is technically more complex to implement, so I've decided to go with option 2; I'm happy to update the implementation if there's something important I'm missing.

It's possible that option 2 also allows to only forward the global config once if it won't be modified, in cases where the config is defined statically in a local configuration file (e.g. for proxy relays). Note that I haven't thought about this in detail.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -83,13 +83,19 @@ impl From<GetGlobalConfigResponse> for UpdateGlobalConfig {
 pub struct GlobalConfigService {
     project_cache: Addr<ProjectCache>,
     upstream_relay: Addr<UpstreamRelay>,
+
+    config_update_tx: mpsc::UnboundedSender<UpdateGlobalConfig>,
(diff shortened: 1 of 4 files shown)


iker-barriocanal:
I'll include this in https://github.com/getsentry/relay/pull/2300, to be easier to review.

PR #2309, opened 2023-07-13
fix(crons): Limit environment names on check-ins to 64 chars

rjo100:
Restricts max_chars on environment names.

Helps close SENTRY-1351

#skip-changelog


jjbayer (changes requested):
The `metastructure` attribute only works in combination with `Annotated`. I think we need to do some manual work here, similar to `trim_slug`.

rjo100:
is there any way to directly import this value instead? can add in the copycat `trim_slug` after
https://github.com/getsentry/relay/blob/3ec78a1fb4b687dbf5c7c77149126bb1b4e69ebc/relay-general/src/processor/attrs.rs#L134

jjbayer:
> is there any way to directly import this value instead? can add in the copycat `trim_slug` after
> 
> https://github.com/getsentry/relay/blob/3ec78a1fb4b687dbf5c7c77149126bb1b4e69ebc/relay-general/src/processor/attrs.rs#L134

You can either add `relay-general` to this crate's Cargo.toml and then use `MaxChars::Environment.limit()`, or (to avoid an additional dependency), define a new constant in `relay-common` and import it from both here and  `MaxChars::limit()`.

relay-monitors/src/lib.rs
```diff
@@ -160,6 +167,10 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
         return Err(ProcessCheckInError::EmptySlug);
     }
 
+    if check_in.environment.len() > ENVIRONMENT_LENGTH {
```
jjbayer: Here I think you want the number of chars, not the number of bytes. Rust is nasty.

```suggestion
    if check_in.environment.chars().count() > ENVIRONMENT_LENGTH {
```

relay-crash/sentry-native
jjbayer, line 1: Could you check why this submodule was updated?

jjbayer (changes requested):
Almost there, see two remaining comments.

relay-crash/sentry-native
rjo100: Oh hm didn't mean to do that

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add support for `sampled` field in the DSC and error tagging. ([#2290](https://github.com/getsentry/relay/pull/2290))
 - Move span tag extraction from metrics to normalization. ([#2304](https://github.com/getsentry/relay/pull/2304))
+- Limit environment names on check-ins to 64 chars. ([#2309](https://github.com/getsentry/relay/pull/2309))
```
olksdr: Could you, please, move this to `Unreleased` section of the changelog

relay-monitors/src/lib.rs
```diff
@@ -32,6 +35,10 @@ pub enum ProcessCheckInError {
     /// Monitor slug was empty after slugification.
     #[error("the monitor slug is empty or invalid")]
     EmptySlug,
+
+    /// Environment name was invalid.environment.
```
olksdr:
```suggestion
    /// Environment name was invalid.
```

olksdr (commented):
Left few suggestions. 

I'm not sure we want to rely on `relay-general` crate though, it's quite big with a lot of deps as well, and it would be better to keep the `relay-monitors` somewhere with smaller footprint.

CHANGELOG.md
rjo100: Yep, got moved during a bad merge

rjo100:
I really want to only import the `environment` max chars length to keep everything synced up to one variable as it represents the same object. I can manually copy the `64` limit over but would prefer not to.

jjbayer:
> I really want to only import the `environment` max chars length to keep everything synced up to one variable as it represents the same object. I can manually copy the `64` limit over but would prefer not to.

How about we define a constant in `relay-common` and import it from both places (`relay-general` and `relay-monitors`)?

rjo100:
I think I'm just going to hardcode `64` in `relay-monitors` and update [this comment](https://github.com/getsentry/sentry/blob/7eef24a0858eb2d1868b323f26d0998dcd8cc94a/src/sentry/models/environment.py#L56) in the `sentry` repo mention both instances

rjo100:
alright I can't figure out how to make this work. will wait for comments

jjbayer (approved):
LGTM as-is.

PR #2318, opened 2023-07-17
fix(spans): Extract table name for UPDATE queries

jjbayer:
Previously, only `SELECT`, `DELETE` and `INSERT` SQL queries got a `domain` tag.

ref: [internal issue](https://www.notion.so/sentry/Most-UPDATE-queries-have-no-table-a1eb863e5a8048538338b95a874ae460?pvs=4)

#skip-changelog


olksdr (approved):

wmak (commented):
Should we add `DELETE` too?

jjbayer:
> Should we add `DELETE` too?

@wmak `DELETE` was already covered by the `FROM` keyword. Added a unit test to verify.

PR #2319, opened 2023-07-17
feat(spans): Strip quotes and prefix from SQL table and column names

jjbayer:
Attempt to shorten normalized SQL span descriptions to increase the information we can fit into the tag limit.

The downside of this is that it creates ambiguous references to columns and tables, but we're assuming this does not impact usability in a negative way.

For example,

```sql
SELECT "table"."foo", "table"."bar" from "table"

-- becomes

SELECT foo, bar from table
```

#skip-changelog


relay-general/src/store/normalize/span/description.rs
```diff
@@ -101,7 +105,16 @@ pub(crate) fn scrub_span_description(
...
     mark_as_scrubbed |= scrub_identifiers_with_regex(string, &SQL_NORMALIZER_REGEX, "%s")?;
-    mark_as_scrubbed |= scrub_identifiers_with_regex(string, &SQL_COLLAPSE_REGEX, "%s")?;
+    mark_as_scrubbed |= scrub_identifiers_with_regex(string, &SQL_COLLAPSE_PLACEHOLDERS, "%s")?;
+
+    if let Some(s) = string.as_str() {
+        match SQL_COLLAPSE_ENTITIES.replace_all(s, "$entity_name") {
+            Cow::Borrowed(_unchanged) => {
+                dbg!(_unchanged);
+            }
+            Cow::Owned(changed) => string.set_value(dbg!(Some(changed))),
```
jjbayer:
```suggestion
            Cow::Borrowed(_unchanged) => {}
            Cow::Owned(changed) => string.set_value(Some(changed))
```

olksdr (approved):

PR #2321, opened 2023-07-17
fix(profiling): Filter idle thread at the edge per thread

Zylphrex:
When filtering idle threads, it needs to be done on a per-thread basis otherwise short-lived threads that have only idle samples will persist in the profile.

### DIFF
--- a/relay-profiling/src/sample.rs
+++ b/relay-profiling/src/sample.rs
@@ -13,6 +13,12 @@ use crate::MAX_PROFILE_DURATION;
 
 const MAX_PROFILE_DURATION_NS: u64 = MAX_PROFILE_DURATION.as_nanos() as u64;
 
+#[derive(Debug)]
+struct Range {
+    start: usize,
+    end: usize,
+}
+
 #[derive(Debug, Serialize, Deserialize, Clone)]
 struct Frame {
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


relay-profiling/src/sample.rs
```diff
@@ -249,23 +249,55 @@ impl SampleProfile {
...
+                        true => false,
+                        false => {
+                            active_threads.insert(sample.thread_id);
+                            true
+                        }
+                    },
+                    None => false,
+                }
+            })
+            .rev()
```
Zylphrex: Not 100% sure what's going on here but if I changed the order here from

```rust
.drain(..).filter().rev().collect();

.drain(..).filter().rev().collect();
```

to

```rust
.drain(..).filter().collect();

.drain(..).rev().filter().rev().collect();
```

The second filter no longer iterates the samples in reversed order.

Zylphrex: updated to not rely on the filter order

phacops (approved):

relay-profiling/src/sample.rs
```diff
@@ -249,23 +255,39 @@ impl SampleProfile {
...
-            self.profile.samples.truncate(end + 1);
         }
+
+        self.profile.samples = self
+            .profile
+            .samples
+            .drain(..)
+            .enumerate()
+            .filter(|(i, sample)| {
+                if let Some(range) = active_ranges.get_mut(&sample.thread_id) {
```
jjbayer:
```suggestion
                if let Some(range) = active_ranges.get(&sample.thread_id) {
```

relay-profiling/src/sample.rs
```diff
@@ -13,6 +13,12 @@ use crate::MAX_PROFILE_DURATION;
 
 const MAX_PROFILE_DURATION_NS: u64 = MAX_PROFILE_DURATION.as_nanos() as u64;
 
+#[derive(Debug)]
+struct Range {
+    start: usize,
+    end: usize,
+}
```
jjbayer: You could use https://doc.rust-lang.org/std/ops/struct.Range.html for this. If you prefer to keep a right-inclusive range (`[a, b]`), I would rename its fields to `first` and `last`, to emphasize that its last element is included.

jjbayer (approved):
No blockers, but please take a look at the comments.

PR #2324, opened 2023-07-18
feat(spans): Simplify scrubbed description for HTTP spans

jjbayer:
Simplify parameterization of http.client span descriptions. Strip out everything except the HTTP method, hostname and port (unless 80 or 443). Wildcard everything but the hostname and TLD e.g., `POST api.github.com/repos?param=value` becomes `POST *.github.com` and `POST http://127.0.0.1:10006/discover/snql` becomes `POST https://127.0.0.1:10006`.

This is a temporary solution to bring cardinality to an acceptable level. We might return to a smarter solution, including automatic URL clustering, later.

ref: [internal issue](https://www.notion.so/sentry/Stop-extracting-span-group-metric-tag-for-API-module-spans-302c00f115ff44b5a2e24a9ca500cc28?pvs=4)

#skip-changelog


relay-general/src/store/normalize/span/description.rs
```diff
@@ -112,11 +110,52 @@ fn is_sql_query_scrubbed(query: &Annotated<String>) -> bool {
...
-fn scrub_http(string: &mut Annotated<String>) -> Result<bool, ProcessingAction> {
-    let mut scrubbed =
-        scrub_identifiers_with_regex(string, &TRANSACTION_NAME_NORMALIZER_REGEX, "*")?;
-    scrubbed |= scrub_identifiers_with_regex(string, &HTTP_SCRUB_ADDITIONAL, "*")?;
-    Ok(scrubbed)
+fn scrub_http(string: &mut String) -> bool {
+    let Some((method, url)) = string.split_once(' ') else { return false };
+    let Ok(url) = Url::parse(url) else { return false };
+    let Some(domain) = url.domain() else { return false };
+    let Some(domain) = normalize_domain(domain, url.port()) else { return false};
```
```suggestion
    let Some(domain) = normalize_domain(domain, url.port()) else { return false };
```

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -219,10 +218,19 @@ pub(crate) fn extract_tags(span: &Span, config: &Config) -> BTreeMap<SpanTagKey,
...
-            span.description
-                .value()
-                .and_then(|url| domain_from_http_url(url))
-                .map(|d| d.to_lowercase())
+            // HACK: Parse the normalized description to get the normalized domain.
+            scrubbed_description
+                .and_then(|d| d.split_once(' '))
+                .and_then(|(_, d)| Url::parse(d).ok())
+                .and_then(|url| {
+                    url.domain().map(|d| {
```
olksdr, line 226: This code repeats a bit in `scrub_http`  function. Maybe it makes sense to create a smallish helper util for this. But it's also fine to keep it this way.

olksdr (approved):

relay-general/src/store/normalize/span/tag_extraction.rs
jjbayer: I see what you mean. Ideally we would parse the URL only once and then reuse that materialized field, but because description scrubbing runs before tag extraction, I saw no obvious way of doing this.

The two functions have a few differences (one normalizes the domain, the other does not, one adds the scheme, one does not), so I didn't manage to isolate the commonalities into a helper function.

PR #2325, opened 2023-07-18
fix(spans): Keep minimal description for relative paths

jjbayer:
Follow-up to https://github.com/getsentry/relay/pull/2324:

Relative paths like `GET /api/endpoint` should become `GET *`.

#skip-changelog


olksdr (approved):
🚢  it!

gggritso (approved):
👍🏻

PR #2327, opened 2023-07-19
docs: Be very clear that the `.relay` folder is in the project root

evanpurkhiser:
I was stuck forever before realizing that this folder is actually supposed to be in my local checkout of relay.

#skip-changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ make test-python
 
 To develop Relay with an existing Sentry devserver, self-hosted Sentry
 installation, or Sentry SaaS, configure the upstream to the URL of the Sentry
-server in `.relay/config.yml`. For example, in local development set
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -212,7 +212,7 @@ make test-python
 
 To develop Relay with an existing Sentry devserver, self-hosted Sentry
 installation, or Sentry SaaS, configure the upstream to the URL of the Sentry
-server in `.relay/config.yml`. For example, in local development set
+server in `./.relay/config.yml`. For example, in local development set
```
olksdr: Maybe it's better to be even more explicit, e.g.: 
```suggestion
server in `.relay/config.yml` in the project root directory. For example, in local development set
```

evanpurkhiser: Sounds good

olksdr (approved):

PR #2330, opened 2023-07-19
feat(spans): Collapse SQL columns in SELECT

jjbayer:
Replace lists of column names in SQL `SELECT` queries with `..`, e.g. 

```sql
SELECT .. FROM table WHERE x = %s
```

This only works for simple lists of column names, not for aliases, aggregations, etc.

#skip-changelog


gggritso (commented):
👍🏻 let's try it

olksdr (approved):

PR #2331, opened 2023-07-19
fix(spans): Fix SQL table extraction for nested selects

jjbayer:
ref: [internal issue](https://www.notion.so/sentry/Nested-Selects-cause-the-table-to-be-select-7a7424d8511846ae895ef092b6342c50?pvs=4)

#skip-changelog


gggritso (approved):
Awesome!

olksdr (approved):

olksdr (approved):

PR #2336, opened 2023-07-20
ref(spans): Simplify description scrubbing

jjbayer:
Instead of using `Annotated` and the custom `scrub_identifiers_with_regex`, work with plain `String` / `&str` and use `Regex::replace_all`.

No functional changes except a small accidental bugfix.

#skip-changelog

### DIFF
--- a/relay-general/src/store/normalize/span/description.rs
+++ b/relay-general/src/store/normalize/span/description.rs
@@ -74,7 +74,7 @@ pub(crate) fn scrub_span_description(span: &mut Span, rules: &Vec<SpanDescriptio
     let scrubbed = span
         .op
         .as_str()
-        .and_then(|op| op.split_once('.'))
+        .map(|op| op.split_once('.').unwrap_or((op, "")))
         .and_then(|(op, sub)| match (op, sub) {
             ("http", _) => scrub_http(description),
             ("cache", _) | ("db", "redis") => scrub_redis_keys(description),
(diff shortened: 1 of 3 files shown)


relay-general/src/store/normalize/span/description.rs
```diff
@@ -561,7 +556,7 @@ mod tests {
         span_description_dont_scrub_double_quoted_strings_format_mysql,
         r#"SELECT * from table WHERE sku = \"foo\""#,
         "db.sql.query",
-        ""
+        "SELECT * from table WHERE sku = foo"
```
jjbayer, line 560: Should be marked as scrubbed, because something was changed. This was wrong before.

relay-general/src/store/normalize/span/description.rs
```diff
@@ -22,7 +22,7 @@ static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     Regex::new(
         r#"(?xi)
         # Capture `SAVEPOINT` savepoints.
-        ((?-x)SAVEPOINT (?P<savepoint>(?:(?:"[^"]+")|(?:'[^']+')|(?:`[^`]+`)|(?:[a-z]\w+)))) |
+        ((?-x)(?P<pre>SAVEPOINT )(?P<savepoint>(?:(?:"[^"]+")|(?:'[^']+')|(?:`[^`]+`)|(?:[a-z]\w+)))) |
```
jjbayer, line 25: `Regex::replace_all` always matches the entire expression, so we need capture groups for the substrings that we want to keep.

olksdr (approved):

PR #2340, opened 2023-07-21
instr(metrics): Measure avg. bucket size

jjbayer:
Add a metric to measure how many elements there are on average in a metrics bucket when it is flushed.

This will help us decide whether increasing the bucket interval for span metrics (and possible performance metrics).

#skip-changelog

### DIFF
--- a/relay-metrics/src/statsd.rs
+++ b/relay-metrics/src/statsd.rs
@@ -1,5 +1,7 @@
 use relay_statsd::{CounterMetric, GaugeMetric, HistogramMetric, SetMetric, TimerMetric};
 
+use crate::BucketValue;
+
 /// Set metrics for Relay Metrics.
 pub enum MetricSets {
     /// Count the number of unique buckets created.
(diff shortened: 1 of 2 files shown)


relay-metrics/src/statsd.rs
```diff
@@ -153,16 +162,17 @@ impl GaugeMetric for MetricGauges {
...
         "c:sessions/session@none",
         "s:sessions/user@none",
         "s:sessions/error@none",
         "d:transactions/duration@millisecond",
         "s:transactions/user@none",
         "c:transactions/count_per_root_project@none",
     ]
-    .contains(&value)
+    .into_iter()
+    .find(|x| x == &value)
```
jjbayer, line 175: This was refactored in https://github.com/getsentry/relay/pull/2041, but using `find` instead of `contains` allows us to return a static str and prevents allocations on the caller side.

relay-metrics/src/statsd.rs
```diff
@@ -183,3 +193,13 @@ pub fn metric_name_tag(value: &str) -> &str {
 
     "other"
 }
+
+/// Returns the metric type for use as a tag key on statsd metrics.
+pub(crate) fn metric_type_tag(value: &BucketValue) -> &'static str {
```
TBS1996, line 198: iiuc, this is useful for the cases where the metric name tag resolves to "other"? as otherwise the first letter would give the type?

TBS1996 (approved):
looks good as far as i can tell

relay-metrics/src/statsd.rs
jjbayer: Yep!

relay-metrics/src/statsd.rs
TBS1996: would unwrap_or("other") work too ?

jjbayer: Would you assign that to `value`? Because then you cannot compare with other values anymore (see comparisions below).

PR #2341, opened 2023-07-21
feat(metrics): Route to different aggregators

jjbayer:
Place a router in front of the `relay_metrics::Aggregator` service, to allow different aggregator configurations for different use cases. This will allow us to

1. change the bucket interval for span metrics from 10 seconds to 1 minute to improve aggregation,
1. (potentially) set a different maximum length for tag values for span metrics.

Old:

```mermaid
flowchart LR
    Service -->|metrics| Aggregator
 ```
 
 New:
 
 ```mermaid
 flowchart LR
    Service -->|metrics| Router
    Router -->|metrics| agg_d["Aggregator (default)"]
    Router -->|metrics| agg_1["Aggregator (spans)"]
```


relay-metrics/src/router.rs
```diff
@@ -0,0 +1,240 @@
...
+    where
+        Aggregator: FromMessage<M>,
+    {
+        let (project_key, metrics) = message.into_parts();
+        let metrics_by_namespace = metrics.into_iter().group_by(|m| {
+            MetricResourceIdentifier::parse(m.name())
+                .map(|mri| mri.namespace)
+                .ok()
+        });
+        // TODO: Parse MRI only once, move validation from Aggregator here.
```
jjbayer, line 208: Will take care of this in a follow-up.

relay-metrics/src/aggregation.rs
```diff
@@ -2007,7 +2035,6 @@ impl Service for AggregatorService {
         tokio::spawn(async move {
             let mut ticker = tokio::time::interval(FLUSH_INTERVAL);
             let mut shutdown = Controller::shutdown_handle();
-            relay_log::info!("aggregator started");
 
             // Note that currently this loop never exists and will run till the tokio runtime shuts
```
TBS1996:
```suggestion
            // Note that currently this loop never exits and will run till the tokio runtime shuts
```

relay-metrics/src/aggregation.rs
```diff
@@ -1811,37 +1828,41 @@ impl AggregatorService {
...
-                            bucket,
-                        });
-
-                    false
-                } else {
-                    true
-                }
-            });
-        });
+        relay_statsd::metric!(
```
TBS1996, line 1831: really wish github would show it in another color if only the indentation changed

relay-metrics/src/router.rs
```diff
@@ -0,0 +1,240 @@
...
+                    for req in requests {
+                        accepts &= req.await.unwrap_or_default();
+                    }
+                    sender.send(accepts);
+                });
+            }
+            Aggregator::InsertMetrics(msg) => self.handle_metrics(msg),
+            Aggregator::MergeBuckets(msg) => self.handle_metrics(msg),
+            #[cfg(test)]
+            Aggregator::BucketCountInquiry(_, _sender) => (), // not supported
```
TBS1996, line 194: I suppose this is fine because the tests will simply use the aggregator service directly?

TBS1996 (approved):
Reading through it, it makes sense to me and seems pretty clever. Although I'm not that confident to know I didn't miss any details since most my effort was on understanding it as a whole

relay-metrics/src/router.rs
jjbayer: Yeah it's not ideal, but acceptable IMO because this message is only used in one test.

PR #2342, opened 2023-07-21
fix(metrics): Move `TrimmingProcessor` to light normalization

jjbayer:
Because the `TrimmingProcessor` is not part of light normalization, we extract the untrimmed transaction name from the payload, which then gets dropped by the aggregator.

Solution: Call `TrimmingProcessor` in `light_normalize_event`.

After deploying this, we have to keep an eye on the `event.processing_time` metric.

Fixes https://github.com/getsentry/team-ingest/issues/187

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Bug Fixes**:
+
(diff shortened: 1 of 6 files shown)


olksdr (approved):
Light normalization will soon replace everything

jjbayer:
> Light normalization will soon replace everything

Yep I would like to revert the split between light normalization and store normalization eventually!

PR #2345, opened 2023-07-24
fix(crons): Avoid responding with a Event ID, 202 instead

evanpurkhiser:
This is confusing as this is NOT the ID of the check-in, we don't know
what that will be since sentry will process the check-in.

Instead we can respond with a 202 Accepted

> The HyperText Transfer Protocol (HTTP) 202 Accepted response status
> code indicates that the request has been accepted for processing, but
> the processing has not been completed; in fact, processing may not have
> started yet. The request might or might not eventually be acted upon, as
> it might be disallowed when processing actually takes place.

#skip-changelog


jjbayer (approved):
Could you add a test case to verify?

PR #2346, opened 2023-07-24
ref(crons): Consistently use the name 'monitors'

evanpurkhiser:
We use monitors internally to refer to the crons product. Let's keep
doing that within relay as well.

#skip-changelog


jjbayer (approved):

iker-barriocanal (approved):
Thanks for doing this!

PR #2347, opened 2023-07-24
Rename monitor to s4s and add first customer

mattgauntseo-sentry:
This switches the name of monitor to s4s (just easier to understand for sentaurs).

The `customer-1` pipeline is the first customer pipeline being added to relay, this should follow the same process as s4s.

I've also removed the rollback pipeline as it doesn't work how relay team would like it to and in its current form it doesn't work with manual steps for pops.

#skip-changelog


mattgauntseo-sentry:
NOTE: The error in the validate deployment pipelines is a known issue and I need to investigate it.

olksdr (approved):

gocd/generated-pipelines/rollback-relay.yaml
jjbayer, line 1: Why is this deleted?

gocd/templates/jsonnetfile.json
```diff
@@ -8,7 +8,7 @@
           "subdir": "libs"
         }
       },
-      "version": "v1.1.9"
+      "version": "v1.2.2"
```
jjbayer, line 11: Does this version version bump cause `relay-customer-1.yaml` to be generated?

jjbayer (commented):
Would this be an independent pipeline, or would it be automatically triggered on every Relay deploy like S4S?

gocd/templates/jsonnetfile.json
mattgauntseo-sentry: yes it does

gocd/generated-pipelines/rollback-relay.yaml
mattgauntseo-sentry: This is because the rollback pipeline doesn't work with the relay pipedream because of the manual promotion to pops step.

The built in pipeline works by re-running the primary deployment stage in each region, however, if you have a manual approval step, that doesn't work because it'll deploy-relay but GoCD will not offer the promote-to-pops step.

Further - when asking about this in the slack channel, it seemed like a rollback would first deploy to pops and then deploy relay (i.e. operate in reverse)

PR #2350, opened 2023-07-25
feat(spans): Extract spans from transactions

jjbayer:
Extract spans from transactions and publish them on a new Kafka topic `ingest-spans`. Protected by a feature flag.

Notes:

* Spans are extracted only in processing relays. If an SDK or downstream Relay provides `Span` items in an envelope, they are deleted unconditionally.
* The extraction happens after all regular transaction processing, for example metrics extraction and dynamic sampling. If we want to split the sampling logic in the future, we will have to move the extraction to an earlier stage.


relay-config/src/config.rs
```diff
@@ -582,6 +584,7 @@ impl Default for Limits {
             max_api_file_upload_size: ByteSize::mebibytes(40),
             max_api_chunk_upload_size: ByteSize::mebibytes(100),
             max_profile_size: ByteSize::mebibytes(50),
+            max_span_size: ByteSize::mebibytes(1),
```
jjbayer, line 587: TBD, we could go lower, e.g. 100 KiB.

relay-server/src/actors/store.rs
```diff
@@ -990,6 +1022,18 @@ struct CheckInKafkaMessage {
...
+struct SpanKafkaMessage {
+    /// Raw span payload.
+    payload: Bytes,
+    /// Time at which the span was received by Relay.
+    start_time: u64,
+    /// The project id for the current span.
+    project_id: ProjectId,
+    // Number of days to retain.
+    retention_days: u16,
+}
```
jjbayer: This schema was copy-pasted from events and is 100% open still.

1. Do we want msgpack wrapping a JSON payload, or just a flat JSON message? Or something else entirely?
2. If we're not going for an opaque `payload` field, what fields should this message have?
3. Do we even need the `start_time`, `project_id`, and `retention_days` fields?

relay-server/src/actors/processor.rs
```diff
@@ -2232,6 +2233,36 @@ impl EnvelopeProcessorService {
...
+            _ => ItemAction::Keep,
+        });
+
+        if !state
+            .project_state
+            .has_feature(Feature::ExtractStandaloneSpans)
+        {
+            return;
+        };
+
```
jjbayer, line 2261: Should probably check for `event.type == transaction` here.

relay-server/src/actors/store.rs
phacops: 1. The simpler the better. If we can have one message encoded as msgpack, I'll take it (so we only have to decode once and not decode the msgpack from Kafka then the payload as JSON).
2. Can we have a field `span` which would embed the span structure we receive?
3. `start_time`is necessary to monitor the latency of the pipeline so I'd leave it, `project_id` will likely be useful later on and where does `retention_days` come from? It's likely something we can and will fetch in `sentry`.

relay-config/src/config.rs
phacops: I'd leave it high for now, not sure how big spans will be (very likely way lower than 1 MiB, even 100 KiB would be surprising), we'll lower it later.

relay-config/src/config.rs
```diff
@@ -542,6 +542,8 @@ struct Limits {
     max_api_chunk_upload_size: ByteSize,
     /// The maximum payload size for a profile
     max_profile_size: ByteSize,
+    /// The maximum payload size for a span
```
iker-barriocanal: nit:
```suggestion
    /// The maximum payload size for a span.
```

relay-server/src/actors/processor.rs
iker-barriocanal: Line 2250 returns early if the event isn't a transaction, does that cover the need?

relay-server/src/actors/store.rs
jjbayer: @phacops Refactored as you suggested.

relay-server/src/actors/processor.rs
```diff
@@ -2237,6 +2238,43 @@ impl EnvelopeProcessorService {
         Ok(())
     }
 
+    #[cfg(feature = "processing")]
+    fn extract_spans(&self, state: &mut ProcessEnvelopeState) {
+        // For now, drop any spans submitted by the SDK.
+        state.managed_envelope.retain_items(|item| match item.ty() {
+            ItemType::Span => ItemAction::DropSilently,
+            _ => ItemAction::Keep,
+        });
```
iker-barriocanal, line 2247: I don't know how much load standalone spans in envelopes will represent to consider moving this logic to PoPs.

relay-server/src/actors/store.rs
```diff
@@ -1041,6 +1093,7 @@ impl Message for KafkaMessage {
             Self::ReplayEvent(message) => message.replay_id.0,
             Self::ReplayRecordingNotChunked(_message) => Uuid::nil(), // Ensure random partitioning.
             Self::CheckIn(_message) => Uuid::nil(),
+            Self::Span(_) => Uuid::nil(), // TODO
```
iker-barriocanal: Q: how does a nil partition key interact with the forwarding?

iker-barriocanal (approved):
Nicee

olksdr (approved):

relay-server/src/actors/processor.rs
jjbayer: There should not be any, so for now this is just a precaution.

relay-server/src/actors/store.rs
jjbayer: `nil` means random partitioning (see lines above), which should be good for now.

```suggestion
            Self::Span(_) => Uuid::nil(), // random partitioning
```

PR #2352, opened 2023-07-25
Add second customer

mattgauntseo-sentry:
#skip-changelog

### DIFF
--- a/gocd/generated-pipelines/relay-customer-2.yaml
+++ b/gocd/generated-pipelines/relay-customer-2.yaml
@@ -1,0 +1,109 @@
+format_version: 10
+pipelines:
+  deploy-relay-customer-2:
+    display_order: 5
+    environment_variables:
+      SENTRY_REGION: customer-2
+    group: relay
+    lock_behavior: unlockWhenFinished
+    materials:
+      deploy-relay-customer-1-pipeline-complete:
+        pipeline: deploy-relay-customer-1
+        stage: pipeline-complete
+      relay_repo:
+        branch: master
+        destination: relay
+        git: git@github.com:getsentry/relay.git
+        shallow_clone: true
+    stages:
+      - checks:
+          fetch_materials: true
+          jobs:
+            checks:
+              elastic_profile_id: relay
+              environment_variables:
+                GITHUB_TOKEN: '{{SECRET:[devinfra-github][token]}}'
+              tasks:
+                - script: |
+                    ##!/bin/bash
+
+                    /devinfra/scripts/checks/githubactions/checkruns.py \
+                        getsentry/relay \
+                        "${GO_REVISION_RELAY_REPO}" \
+                        "Integration Tests" \
+                        "Test (macos-latest)" \
+                        "Test (windows-latest)" \
+                        "Test All Features (ubuntu-latest)" \
+                        "Push GCR Docker Image"
+              timeout: 1800
+      - deploy-production:
+          fetch_materials: true
+          jobs:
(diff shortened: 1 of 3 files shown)


joshuarli (approved):

jjbayer (approved):

gocd/generated-pipelines/relay-customer-2.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+                    /devinfra/scripts/k8s/k8s-deploy.py \
+                        --label-selector="service=relay-pop" \
+                        --image="us.gcr.io/sentryio/relay:${GO_REVISION_RELAY_REPO}" \
+                        --container-name="relay"
+              timeout: 1200
+      - pipeline-complete:
+          approval:
+            allow_only_on_success: true
+            type: success
```
iker-barriocanal, line 104: Are deployments on further customers blocked if a deployment on a customer doesn't succeed?

gocd/generated-pipelines/relay-customer-2.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+                    /devinfra/scripts/k8s/k8s-deploy.py \
+                        --label-selector="service=relay" \
+                        --image="us.gcr.io/sentryio/relay:${GO_REVISION_RELAY_REPO}" \
+                        --container-name="relay"
+              timeout: 1200
+      - progress-to-pops:
+          approval:
+            allow_only_on_success: true
+            type: manual
```
iker-barriocanal, line 73: Does this require manual approval to progress processing -> pops for each customer?

gocd/generated-pipelines/relay-customer-2.yaml
iker-barriocanal, line 1: What is the difference between the customer 2 and customer 1 deployment scripts? Is there any chance we can have a single script for all customers?

iker-barriocanal (commented):
Leaving a couple of questions since I'm missing some context.

gocd/generated-pipelines/relay-customer-2.yaml
mattgauntseo-sentry: yes it is - the idea being that with each deploy we should have increasing confidence that the deploy is safe / going to succeed.

gocd/generated-pipelines/relay-customer-2.yaml
mattgauntseo-sentry: yes exactly. This is how it was originally configured for the US deployment and I think we discussed it in previous PRs and this was desirable to keep.

If you'd like to change this behavior happy to discuss ideas (i.e. if you only care about this manual approval for US deployments I'm happy to add a PR to do that)

gocd/generated-pipelines/relay-customer-2.yaml
mattgauntseo-sentry: All of these changes are generated from a jsonnet file (see gocd/templates/relay.jsonnet).

The reason we are adding the files in this commit is because GoCD expects the generated yaml to import and create the pipeline.

In the near future (we are hoping to deploy today/tomorrow) we'll have a GoCD plugin that can read jsonnet files, after that the `gocd/generated-pipelines` will be deleted.

As for "why is customer-2.yaml being generated" - if you look at the jsonnetfile.json file, it's a version bump of a library, this library determines what regions are deployed and where, in this case, v1.2.2 was [s4s, us, customer-1] and in v1.2.3 the deployments are [s4s, us, customer-1, customer-2].

PR #2357, opened 2023-07-27
feat(spans): Normalize PHP SQL placeholders

jjbayer:
E.g. 

```sql
SELECT x FROM y WHERE (z = :c0 AND w = :c1) LIMIT 1
```

becomes 

```sql
SELECT x FROM y WHERE (z = %s AND w = %s) LIMIT %s
```

#skip-changelog


iker-barriocanal (approved):

gggritso (commented):
Sweet!

PR #2358, opened 2023-07-27
Normalize app start spans ops and measurements names

krystofwoldrich:
- Related to https://github.com/getsentry/sentry-react-native/pull/2855

Fixes wrong span and measurement names coming from older RN SDK releases.


relay-general/src/store/normalize.rs
```diff
@@ -254,6 +255,22 @@ impl<'a> NormalizeProcessor<'a> {
...
+                if let Some(op) = span.op.value() {
+                    if op == "app_start_cold" {
+                        span.op.set_value(Some("app.start.cold".to_string()));
+                    } else if op == "app_start_warm" {
+                        span.op.set_value(Some("app.start.warm".to_string()));
+                    }
+                }
+            }
+        }
+    }
```
krystofwoldrich: This might be expensive for all events, should we guard it only for the specific RN SDK version? Would that be ok?

relay-general/src/store/normalize.rs
```diff
@@ -865,6 +892,10 @@ pub fn light_normalize_event(
         ); // Measurements are part of the metric extraction
         normalize_breakdowns(event, config.breakdowns_config); // Breakdowns are part of the metric extraction too
 
+        if event.ty.value() == Some(&EventType::Transaction) {
+            normalize_app_start_spans(event);
+        }
```
```suggestion
```

Could you please move this into the `if config.light_normalize_spans` block? "Light" normalization is only necessary for metrics extraction, which is currently still feature-flagged for spans.

relay-general/src/store/normalize.rs
jjbayer: Sounds good. `semver` is already an indirect [dependency](https://github.com/getsentry/relay/blob/562b794c36d78a60874851fcb986ad137ba1baed/Cargo.lock#L3751-L3752) so you could use it to compare versions. Though what versions do we actually want to match on? Maybe a simple `startswith` or a regex would also work?

krystofwoldrich: I'll check, likely startsWith will do the job.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Normalize old RN SDK app start time measurements and spans. ([#2358](https://github.com/getsentry/relay/pull/2358))
```
olksdr: Could we, please, use the full name of the SDK? Since it's not really clear what `RN`  stands for?

```suggestion
- Normalize old React Native SDK app start time measurements and spans. ([#2358](https://github.com/getsentry/relay/pull/2358))
```

relay-general/src/store/normalize.rs
```diff
@@ -254,6 +255,33 @@ impl<'a> NormalizeProcessor<'a> {
...
 }
 
+fn normalize_app_start_spans(event: &mut Event) {
+    if !event.sdk_name().eq("sentry.javascript.react-native")
+        || !(event.sdk_version().starts_with("4.4")
+            || event.sdk_version().starts_with("4.3")
+            || event.sdk_version().starts_with("4.2")
+            || event.sdk_version().starts_with("4.1")
+            || event.sdk_version().starts_with("4.0")
+            || event.sdk_version().starts_with('3'))
```
olksdr, line 268: Just to double check, these are only the effected versions you would like to normalize?

relay-general/src/store/normalize.rs
```diff
@@ -254,6 +255,33 @@ impl<'a> NormalizeProcessor<'a> {
     }
 }
 
+fn normalize_app_start_spans(event: &mut Event) {
```
olksdr, line 261: As a small thing it would be great to add docs for the newly added function.

relay-general/src/store/normalize.rs
```diff
@@ -375,6 +403,15 @@ fn get_metric_measurement_unit(measurement_name: &str) -> Option<MetricUnit> {
     }
 }
 
+fn normalize_app_start_measurements(measurements: &mut Measurements) {
```
olksdr, line 415: Also here docs are missing.

olksdr (changes requested):
Overall looks good to me. 

It would be great to add some docs to the newly added functions what they do and the reason behind it.

relay-general/src/store/normalize.rs
krystofwoldrich: Yes. Only these.

relay-general/src/store/normalize.rs
krystofwoldrich: Thanks, added.

relay-general/src/store/normalize.rs
krystofwoldrich: Thanks, added.

relay-general/src/store/normalize.rs
```diff
@@ -254,6 +255,39 @@ impl<'a> NormalizeProcessor<'a> {
     }
 }
 
+/// Replaces snake_case app start spans op with dot.case op.
+/// This is done for the affected RN SDK versions (from 3 to 4.4).
+/// Because we have to iterate over the spans we should do it only
+/// for the affected events.
```
olksdr:
```suggestion
/// Replaces snake_case app start spans op with dot.case op.
///
/// This is done for the affected React Native SDK versions (from 3 to 4.4).
```

relay-general/src/store/normalize.rs
```diff
@@ -375,6 +409,18 @@ fn get_metric_measurement_unit(measurement_name: &str) -> Option<MetricUnit> {
     }
 }
 
+/// Replaces dot.case app start measurements keys with snake_case keys.
+/// The dot.case app start measurements keys are treated as custom measurements.
+/// The snake_case is the key expected by the Sentry UI to aggregate and display in graphs.
```
olksdr:
```suggestion
/// Replaces dot.case app start measurements keys with snake_case keys.
///
/// The dot.case app start measurements keys are treated as custom measurements.
/// The snake_case is the key expected by the Sentry UI to aggregate and display in graphs.
```

olksdr (approved):

PR #2359, opened 2023-07-27
fix(spans): SQL comments & whitespace

jjbayer:
For example,

```sql
    select * -- some comment
            from (
                x
            )
 ```

becomes

```sql
select * from (x)
```

#skip-changelog


relay-general/src/store/normalize/span/description.rs
```diff
@@ -534,28 +537,28 @@ mod tests {
 
     span_description_test!(
         span_description_dont_scrub_double_quoted_strings_format_postgres,
-        r#"SELECT * from \"table\" WHERE sku = %s"#,
+        r#"SELECT * from "table" WHERE sku = %s"#,
```
jjbayer, line 548: `span_description_test!` previously expected JSON encoded strings, hence the escape characters. Now it takes pure strings.

relay-general/src/store/normalize/span/description.rs
```diff
@@ -13,6 +13,8 @@ use crate::store::span::tag_extraction::HTTP_METHOD_EXTRACTOR_REGEX;
 use crate::store::SpanDescriptionRule;
 use crate::types::{Annotated, ProcessingResult, Remark, RemarkType, Value};
 
+static SQL_COMMENTS: Lazy<Regex> = Lazy::new(|| Regex::new(r"--.*(?P<newline>\n)").unwrap());
```
iker-barriocanal, line 17: nit: should we add docstrings?

relay-general/src/store/normalize/span/description.rs
```diff
@@ -36,6 +38,13 @@ static SQL_NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     .unwrap()
 });
 
+/// Removes extra whitespace and newlines.
+static SQL_WHITESPACE: Lazy<Regex> = Lazy::new(|| Regex::new(r"(\s*\n\s*)|(\s\s+)").unwrap());
+
+/// Removes whitespace around parentheses.
+static SQL_PARENS: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"((?P<pre>\()\s+)|(\s+(?P<post>\)))").unwrap());
```
iker-barriocanal, line 47: Q: do we need the capture group names?

iker-barriocanal (approved):

relay-general/src/store/normalize/span/description.rs
jjbayer, line 17:
```suggestion
/// Removes SQL comments starting with "--".
static SQL_COMMENTS: Lazy<Regex> = Lazy::new(|| Regex::new(r"--.*(?P<newline>\n)").unwrap());
```

relay-general/src/store/normalize/span/description.rs
jjbayer: `scrub_sql_queries` uses them for replacement, so I would say yes.

PR #2320, opened 2023-07-17
feat(server): Flesh out global config service

TBS1996:
This PR introduces a new global config service to fetch global configs from upstream. The global config is a project-independent config for Relay that may include configuration for Relays or specific to projects. The service is responsible for fetching the config from upstream regularly in order to have an up-to-date version, currently set to the arbitrary number 10 to have a good balance of the latest version without smashing upstream.

Global configs are served from the same endpoint as project configs for practical reasons (e.g. no need to add new routes to proxy allow-lists) with a new parameter, as in the following example. The endpoint version is currently a version above the project configs while we're in development, these two will match once the initiative is closed.

The global config service supports two ways to fetch the global config to the rest of Relay: serving one-off requests and subscribing for continued updates. In this PR, the envelope processor service (subscribed) and the endpoint (one-off requests, to serve the configs downstream) are the only entities using the global config service.

### Example request

```txt
/api/0/relays/projectconfigs/?version=4
{"publicKeys": [1, 2, 3], "global": true}
```

### Future work
This PR also introduces unfinished work that needs further discussion, brought out of the scope of this PR:
- [ ] https://github.com/getsentry/relay/pull/2411
- [ ] https://github.com/getsentry/relay/pull/2445

PR #2363, opened 2023-07-30
ref(server): Internal Arc for EnvelopeProccesorService

TBS1996:
This refactor is needed for #2320, made it a separate PR to make it easier to read the diffs on that branch

We want to make envelopeprocessorservice easily cloneable, so we put most of the data it contains behind an arc. This allows us to remove the outer arc in the spawn_handler, so that we can mutate the service. Which will be needed for global config update.

#skip-changelog

### DIFF
--- a/relay-general/src/store/normalize/span/tag_extraction.rs
+++ b/relay-general/src/store/normalize/span/tag_extraction.rs
@@ -124,9 +124,9 @@ fn extract_shared_tags(event: &Event) -> BTreeMap<SpanTagKey, String> {
             .and_then(|r| r.method.value())
             .map(|m| m.to_uppercase());
 
-        if let Some(transaction_method) = transaction_method_from_request
-            .or(http_method_from_transaction_name(transaction_name).map(|m| m.to_uppercase()))
-        {
+        if let Some(transaction_method) = transaction_method_from_request.or_else(|| {
+            http_method_from_transaction_name(transaction_name).map(|m| m.to_uppercase())
+        }) {
             tags.insert(SpanTagKey::TransactionMethod, transaction_method);
         }
     }
@@ -213,7 +213,7 @@ pub(crate) fn extract_tags(span: &Span, config: &Config) -> BTreeMap<SpanTagKey,
             _ => None,
         };
 
-        if let Some(act) = action.clone() {
+        if let Some(act) = action {
             span_tags.insert(SpanTagKey::Action, act);
         }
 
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -520,8 +520,13 @@ impl FromMessage<RateLimitFlushBuckets> for EnvelopeProcessor {
 /// Service implementing the [`EnvelopeProcessor`] interface.
 ///
 /// This service handles messages in a worker pool with configurable concurrency.
+#[derive(Clone)]
 pub struct EnvelopeProcessorService {
     config: Arc<Config>,
```
jjbayer: `config` can also move into `inner`. Only the new global config (which is not part of this PR) will be outside of the `Arc`.

relay-server/src/actors/processor.rs
```diff
@@ -541,6 +546,7 @@ impl EnvelopeProcessorService {
         project_cache: Addr<ProjectCache>,
         upstream_relay: Addr<UpstreamRelay>,
     ) -> Self {
+        // TODO(tor): Replace with state enum (`init`, `ready`) and do not process anything while global_config is undefined.
```
jjbayer: Does this comment make sense in this PR?
```suggestion
```

relay-server/src/actors/processor.rs
```diff
@@ -520,8 +520,13 @@ impl FromMessage<RateLimitFlushBuckets> for EnvelopeProcessor {
 /// Service implementing the [`EnvelopeProcessor`] interface.
 ///
 /// This service handles messages in a worker pool with configurable concurrency.
+#[derive(Clone)]
 pub struct EnvelopeProcessorService {
     config: Arc<Config>,
+    inner: Arc<InnerEnvelopeProcessorService>,
+}
+
+struct InnerEnvelopeProcessorService {
```
jjbayer: nit: Since this is a private type, I would be fine with a shorter name, e.g. `InnerProcessor`.

relay-general/src/store/normalize/span/tag_extraction.rs
jjbayer, line 1: Changes in this file are unrelated.

jjbayer (approved):
See comments.

relay-server/src/actors/processor.rs
TBS1996: I wasn't sure why but this is what jan was suggesting when all paired 🤔

relay-general/src/store/normalize/span/tag_extraction.rs
TBS1996: true but my linter complained about them so might as well change it now. seems it suddenly got the ability to spot unneeded clones

relay-server/src/actors/processor.rs
TBS1996: ok nvm, was a misunderstanding, changed it now

jjbayer:
Also please capitalize the PR title 😄

PR #2366, opened 2023-08-01
chore(lint): Fix all beta lint warnings

olksdr:
Using combination of `cargo fix` and some of the manual changes, to make sure `Relay` can be compiled with `stable` and `beta` toolsets.

#skip-changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = ["relay", "relay-*", "tools/*"]
 default-members = ["relay"]
+resolver = "2"
 
 [profile.dev]
 # Debug information slows down the build and increases caches in the
--- a/relay-server/src/actors/project_upstream.rs
+++ b/relay-server/src/actors/project_upstream.rs
@@ -257,6 +257,10 @@ impl UpstreamProjectSourceService {
         let nocache_batches = channels.nocache_channels.into_iter().chunks(batch_size);
 
         let mut requests = vec![];
+        // The `nocache_batches.into_iter()` still must be called here, since compiler produces the
+        // error: `that nocache_batches is not an iterator`.
+        // Since `IntoChunks` is not an iterator itself but only implements `IntoIterator` trait.
+        #[allow(clippy::useless_conversion)]
         for channels_batch in cache_batches.into_iter().chain(nocache_batches.into_iter()) {
             let mut channels_batch: ProjectStateChannels = channels_batch.collect();
             for channel in channels_batch.values_mut() {
--- a/relay-server/src/actors/upstream.rs
+++ b/relay-server/src/actors/upstream.rs
@@ -139,7 +139,7 @@ impl UpstreamRequestError {
     fn is_network_error(&self) -> bool {
         match self {
             Self::SendFailed(_) => true,
-            Self::ResponseError(code, _) => matches!(code.as_u16(), 502 | 503 | 504),
+            Self::ResponseError(code, _) => matches!(code.as_u16(), 502..=504),
             Self::Http(http) => http.is_network_error(),
             _ => false,
         }
--- a/tools/document-metrics/Cargo.toml
+++ b/tools/document-metrics/Cargo.toml
@@ -14,7 +14,7 @@ clap = { version = "4.1.4", features = ["derive"] }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.55"
 serde_yaml = "0.9.17"
-syn = "1.0.39"
+syn = { version = "1.0.39", features = ["full"] }
 
 [dev-dependencies]
 insta = "1.19.0"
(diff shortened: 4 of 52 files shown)


Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 members = ["relay", "relay-*", "tools/*"]
 default-members = ["relay"]
+resolver = "2"
```
iker-barriocanal, line 4: We pretty much use the 2021 Rust edition, which according to [docs](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html) uses `resolver=2` by default:

> Starting in Rust 2021, this will be the default. That is, writing edition = "2021" in Cargo.toml will imply resolver = "2".

Am I missing something? Or what are the benefits of typing it explicitly?

relay-server/src/actors/project_upstream.rs
```diff
@@ -257,6 +257,10 @@ impl UpstreamProjectSourceService {
         let nocache_batches = channels.nocache_channels.into_iter().chunks(batch_size);
 
         let mut requests = vec![];
+        // The `nocache_batches.into_iter()` still must be called here, since compiler produces the
+        // error: `that nocache_batches is not an iterator`.
+        // Since `IntoChunks` is not an iterator itself but only implements `IntoIterator` trait.
+        #[allow(clippy::useless_conversion)]
```
iker-barriocanal, line 263: 👍

relay-server/src/actors/upstream.rs
```diff
@@ -139,7 +139,7 @@ impl UpstreamRequestError {
     fn is_network_error(&self) -> bool {
         match self {
             Self::SendFailed(_) => true,
-            Self::ResponseError(code, _) => matches!(code.as_u16(), 502 | 503 | 504),
+            Self::ResponseError(code, _) => matches!(code.as_u16(), 502..=504),
```
iker-barriocanal, line 142: I believe this is actually harder to read than the previous diff state. Do you think the new version is more readable?

iker-barriocanal (approved):

Cargo.toml
olksdr: Because of the warning 
```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

relay-server/src/actors/upstream.rs
olksdr: That's what `lint` wants. For me personally it does not matter, both are equally ok.

tools/document-metrics/Cargo.toml
```diff
@@ -14,7 +14,7 @@ clap = { version = "4.1.4", features = ["derive"] }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.55"
 serde_yaml = "0.9.17"
-syn = "1.0.39"
+syn = { version = "1.0.39", features = ["full"] }
```
jjbayer, line 17: Why this?

Cargo.toml
jjbayer: Would it be easier to upgrade `relay-kafka` to `edition = 2021`? Or would that cause other problems?

olksdr: `relay-kafka` is upgraded, see the changes in `relay-kafka/Cargo.toml`

tools/document-metrics/Cargo.toml
olksdr: This tool failed to compile, same locally, on the branch, it could not find `syn::Pat` enum, and according to the docs https://docs.rs/syn/latest/syn/enum.Pat.html this is available when feature `full` is enabled.

Cargo.toml
jjbayer: > virtual workspaces default to `resolver = "1"

Sorry, should have read this. I thought the fallback to resolver = 1 happened because of the kafka crate.

jjbayer (approved):

Cargo.toml
iker-barriocanal: Also missed it hehe. Is this happening due to external crates using older editions?

olksdr: See the [`docs on resolvers`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions)

olksdr: @iker-barriocanal  also the docs you posted are describing exactly the same https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html#details

iker-barriocanal: > If you are using a virtual workspace, you will still need to explicitly set the resolver field in the [workspace] definition if you want to opt-in to the new resolver.

Should have read the whole section 😄

tools/document-metrics/Cargo.toml
iker-barriocanal: The resolver bump introduces some [feature unification](https://doc.rust-lang.org/cargo/reference/resolver.html#features) that I haven't read, but may be caused by that?

Zylphrex (approved):

PR #2367, opened 2023-08-01
feat(spans): Collapse list of SQL columns with aliases

jjbayer:
We already scrubbed list of columns in SQL columns such as

```sql
SELECT a, b, c FROM table
-- becomes
SELECT .. FROM table
```

Also collapse if columns have an alias, e.g.

```sql
SELECT table.a AS table__a, table.b AS table__b, table.c AS table__c FROM table
-- becomes
SELECT .. FROM table
```

#skip-changelog

### DIFF
--- a/relay-general/src/store/normalize/span/description.rs
+++ b/relay-general/src/store/normalize/span/description.rs
@@ -62,8 +62,12 @@ static SQL_COLLAPSE_PLACEHOLDERS: Lazy<Regex> = Lazy::new(|| {
 });
 
 /// Collapse simple lists of columns in select.
+/// For example,
+///   SELECT a, b FROM x -> SELECT .. FROM x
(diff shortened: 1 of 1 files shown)


relay-general/src/store/normalize/span/description.rs
```diff
@@ -62,8 +62,12 @@ static SQL_COLLAPSE_PLACEHOLDERS: Lazy<Regex> = Lazy::new(|| {
 });
 
 /// Collapse simple lists of columns in select.
+/// For example,
```
```suggestion
/// For example:
```

relay-general/src/store/normalize/span/description.rs
```diff
@@ -701,6 +705,14 @@ mod tests {
         "SELECT DISTINCT .. FROM table WHERE %s"
     );
 
+    span_description_test!(
+        span_description_collapse_columns_with_as,
+        // Simple lists of columns will be collapsed
```
```suggestion
        // Simple lists of columns will be collapsed.
```

olksdr (approved):
🚢  it!

PR #2370, opened 2023-08-01
[WIP] feat(relay): Build Relay for ARM

iker-barriocanal:
WIP, created PR to run CI to test.


Dav1dde:
@iker-barriocanal what do you think, let's close this, it's stale and already linked to the issue for future reference?

iker-barriocanal:
Closed the PR but keeping the branch alive, as I'm not sure if it's needed to keep the diff (don't want to lose that).

PR #2371, opened 2023-08-01
fix: Don't reject entire envelope

jjbayer:
### Legal Boilerplate

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -2355,6 +2355,7 @@ impl EnvelopeProcessorService {
                 let outcome = Outcome::FilteredSampling(rule_ids.clone());
                 state.managed_envelope.retain_items(|item| {
                     if feature_enabled && item.ty() == &ItemType::Profile {
+                        item.set_header("sampled", false);
                         ItemAction::Keep
                     } else {
                         ItemAction::Drop(outcome.clone())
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -2351,9 +2351,19 @@ impl EnvelopeProcessorService {
             SamplingResult::Drop(rule_ids)
                 if state.event_type() == Some(EventType::Transaction) =>
             {
-                state
-                    .managed_envelope
-                    .reject(Outcome::FilteredSampling(rule_ids.clone()));
+                let feature_enabled = false;
```
jjbayer, line 2354: @viglia replace this by a feature declared in the enum below:

https://github.com/getsentry/relay/blob/1adb61adc8c19322cacef75203195690f06575ae/relay-dynamic-config/src/feature.rs#L7

relay-server/src/actors/processor.rs
```diff
@@ -2351,9 +2351,20 @@ impl EnvelopeProcessorService {
...
+                let outcome = Outcome::FilteredSampling(rule_ids.clone());
+                state.managed_envelope.retain_items(|item| {
+                    if feature_enabled && item.ty() == &ItemType::Profile {
+                        item.set_header("sampled", false);
```
jjbayer, line 2358: @viglia you can access this header through `item.get_header()` in https://github.com/getsentry/relay/blob/a4a6922815903f14528ffb5ca7437573c9192ee5/relay-server/src/actors/store.rs#L617 and add the flag to the kafka message.

jjbayer:
@viglia this is the filter that needs tweaking to let through unsampled profiles: https://github.com/getsentry/relay/blob/a4a6922815903f14528ffb5ca7437573c9192ee5/relay-server/src/actors/processor.rs#L1070

PR #2355, opened 2023-07-26
ref(spans): Use generic metric extraction

jjbayer:
We now have a generic configuration language for metrics extraction. Let's use it for spans.

In this PR, we keep a hardcoded config for span metrics extraction. At a later point we can move this configuration to sentry.

#skip-changelog

PR #2373, opened 2023-08-02
instr(spans): Measure span metrics extraction time

jjbayer:
In preparation of https://github.com/getsentry/relay/pull/2355, measure how much time span metrics extraction currently costs.

#skip-changelog


olksdr (approved):

iker-barriocanal (commented):
The cost of extracting span metrics consists of the cost of normalizing the fields + the cost of extracting these fields into actual metrics. This PR only considers the latter, should we also consider the former?

jjbayer:
> The cost of extracting span metrics consists of the cost of normalizing the fields + the cost of extracting these fields into actual metrics. This PR only considers the latter, should we also consider the former?

@iker-barriocanal that would make sense. I introduced this metric purely to do a before / after comparison of a refactor of the extraction part, so for that purpose it is enough.

Issue #2083, opened 2023-05-02
More indexed transactions than processed


jjbayer:
In a test project, we see that "transactions" (a.k.a. transaction metrics) are limited by spike protection, but the corresponding "indexed transactions" (a.k.a. transaction payloads) are not:

| processed  | indexed |
| ------------- | ------------- |
| ![Screenshot 2023-05-02 at 08 20 52](https://user-images.githubusercontent.com/1565449/235595893-47592d9e-2d0e-4367-9b2d-9fcdb881ebdd.png)  | ![Screenshot 2023-05-02 at 08 21 18](https://user-images.githubusercontent.com/1565449/235595913-8f6bf458-5d03-4c72-bb41-5a06e2ca0b22.png) |

([link](https://sentry.io/_admin/customers/profiling-ga-test/?dataType=indexed_transactions&end=2023-05-02T02%3A17%3A26&start=2023-05-01T12%3A22%3A13))

This also means that profiles are not limited by transaction spike protection.
```[tasklist]
### Delay metrics extraction
- [x] Skip metrics extraction in non-processing Relays when sampling decision is "keep": https://github.com/getsentry/relay/pull/2562
- [x] Bump metrics extraction version so outdated external Relays stop extracting for "keep" as well
```

```[tasklist]
### Rate limit metrics before aggregation
- [ ] https://github.com/getsentry/relay/pull/2574
- [ ] https://github.com/getsentry/relay/pull/2588
- [ ] https://github.com/getsentry/relay/pull/2652
```

jjbayer:
cc @phacops

jjbayer:
Tracking the profiles-specific issue here now because it has higher priority than the inconsistency between indexed and processed transactions: https://github.com/getsentry/relay/issues/2158

jernejstrasner:
@jjbayer the Profiling specific case doesn't fix this one fully, does it?

jjbayer:
> the Profiling specific case doesn't fix this one fully, does it?

@jernejstrasner no, https://github.com/getsentry/relay/issues/2158 only guarantees consistency between processed profiles and processed transactions.

jjbayer:
More real world examples here (all spike protection): https://www.notion.so/sentry/More-indexed-than-processed-10acca5514324738a8c7587f9b93788d

jjbayer:
[...]

PR #2374, opened 2023-08-02
feat(profiling): add support to ingest non-sampled profiles

viglia:
This will let us use data of non sampled profiles to enhance current aggregation (slowest function) with more data.

Currently under feature flag for experimentation with the `Sentry` org.

Feature flag definition `getsentry` [here](https://github.com/getsentry/getsentry/pull/11335) and `sentry` [here](https://github.com/getsentry/sentry/pull/54023).

⚠  Please, do not merge this before the appropriate `vroom` [PR](https://github.com/getsentry/vroom/pull/300) is merged.


relay-server/src/actors/processor.rs
```diff
@@ -1066,8 +1066,18 @@ impl EnvelopeProcessorService {
...
+            // but we still want to keep the profile for aggregations (slowest functions, etc.)
+            ItemType::Profile
+                if (transaction_count == 0)
+                    && item
+                        .get_header("sampled")
+                        .and_then(Value::as_bool)
+                        .unwrap_or(true) =>
+            {
+                ItemAction::DropSilently
+            }
```
Zylphrex: We **never** want to keep any profiles that came in their own envelope. This includes the old cocoa SDK which we do not want to support. So let's make sure to scope this to just profiles that came in the same envelope as the corresponding transaction.

viglia: @Zylphrex for profiles that get sampled we still do this check.

We only do this for profiles that were non sampled due to `DS`.

@jjbayer was explaining to me (IIUC) that there is a chain of operations on the envelope as it gets processed, so if a envelope comes with transaction + profile, and then the transaction is not sampled due to DS, then the output would be the envelope without the transaction.

Gotta check with @jjbayer if there is another way to check wether the `transaction_count==0` because of DS or because it was not sent over to begin with.

phacops: This `transaction_count` is counting the number of transaction in the envelope as it was received. So, if it's 0, it's not because of DS and we want to discard the profile no matter what.

relay-server/src/actors/processor.rs
```diff
@@ -2346,14 +2346,28 @@ impl EnvelopeProcessorService {
...
-                state
-                    .managed_envelope
-                    .reject(Outcome::FilteredSampling(rule_ids.clone()));
+                let unsampled_profiles_enabled =
+                    project_state.has_feature(Feature::IngestUnsampledProfiles);
+
+                let outcome = Outcome::FilteredSampling(rule_ids.clone());
+                state.managed_envelope.retain_items(|item| {
+                    if unsampled_profiles_enabled && item.ty() == &ItemType::Profile {
+                        item.set_header("sampled", false);
```
jjbayer: To reduce processing cost, could we skip [`process_profiles`](https://github.com/getsentry/relay/blob/13d9953162673ab546f10108d3bd87ecba718602/relay-server/src/actors/processor.rs#L1121) for profiles with `sampled: false`?

jjbayer (commented):
We definitely need some tests for this, ideally with a chain of relays to see if the item header is propagated correctly. See for example

https://github.com/getsentry/relay/blob/13d9953162673ab546f10108d3bd87ecba718602/tests/integration/test_dynamic_sampling.py#L560-L590

relay-server/src/actors/processor.rs
viglia: @jjbayer I don't think so. 

If the `feature-flag` is enabled for a given org, and we want to send non-sampled profiles over, we still need to check that those are proper profiles (respect max size, profile payload properly formed, etc.)

viglia:
> This issue has gone three weeks without activity. In another week, I will close it.
> 
> But! If you comment or otherwise update it, I will reset the clock, and if you remove the label `Waiting for: Community`, I will leave it alone ... forever!
> 
> "A weed is but an unloved flower." ― _Ella Wheeler Wilcox_ 🥀

Please, leave it open. This is currently blocked as we're waiting for some data to do the cost/benefit analysis, which in turns will only be available once https://github.com/getsentry/relay/issues/2083 will be fixed (plus 2 more weeks to collect new data).

PR #2375, opened 2023-08-02
Fix build error on the platforms where sentry-native is not being built

mkhon:
The error is:

```
error: couldn't read .../build/relay-crash-20eaf6600db2deb6/out/bindings.rs: No such file or directory (os error 2)
```

The problem is that bindings.rs is not generated on such platforms.

### Legal Boilerplate

#skip-changelog


iker-barriocanal:
Hi @mkhon, thanks for contributing! When is `sentry-native` not built? I've tried replicating the error by cloning the repo on an empty directory and just running `make` sets everything up for [setting up and running](https://github.com/getsentry/relay#building-and-running) Relay. Could you provide some more details of your setup?

mkhon:
It is not built on unsupported platforms (not MacOS or Linux), but the relay on these platforms still can be built (and not generating bindings.rs is the only issue on such platforms)

iker-barriocanal:
What platforms were you targeting? I wonder if it's worth adding them to the CI workflows too.

mkhon:
We are building relay on FreeBSD (yuck!)

iker-barriocanal (approved):
LGTM! Could you add a changelog entry? See instructions in https://github.com/getsentry/relay#instructions-for-changelogs? Thanks for your contribution!

iker-barriocanal:
Hi @mkhon, getting back to this. The compilation feature flag `crash-handler` decides whether sentry native is compiled, so disabling that feature flag should skip building that piece and thus not erroring. This is documented in the [README](https://github.com/getsentry/relay?tab=readme-ov-file#features). Let me know if you have any questions!

PR #2376, opened 2023-08-02
ci: only render gocd pipelines if relevant files are changed

joshuarli:
This saves ~2 minutes every time which would help relieve github free runner queue pressure.

#skip-changelog

### DIFF
--- a/.github/workflows/validate-pipelines.yml
+++ b/.github/workflows/validate-pipelines.yml
@@ -9,20 +9,6 @@ on:
   pull_request:
 
 jobs:
-    render:
-        name: Render GoCD Pipelines with Jsonnet
-        runs-on: ubuntu-latest
-        steps:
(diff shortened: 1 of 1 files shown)


olksdr (approved):

.github/workflows/validate-pipelines.yml
```diff
@@ -40,6 +26,22 @@ jobs:
                   - 'gocd/**'
                   - '.github/workflows/validate-pipelines.yml'
 
+    render:
+        if: needs.files-changed.outputs.gocd == 'true'
```
iker-barriocanal, line 30: Q: how does this know what files generate output?

joshuarli: getsentry/paths-filter `with.filters.gocd` - see right above

PR #2380, opened 2023-08-03
ref(server): Do not cache envelope summary

jjbayer:
We currently keep a cached envelope summary on `ManagedEnvelope`, which is not updated after the event item has been popped from the envelope for processing.

This has the advantage that the summary still "contains" an event while it has been temporarily removed from the envelope, but it causes the problem that we cannot update it _during_ processing without "losing" the event for the purpose of outcome reporting.

This PR attempts to remove all caching of the envelope summary, while still reporting correct outcomes for events.

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -1720,6 +1720,10 @@ impl EnvelopeProcessorService {
         };
 
         state.event = event;
+        if let Some(category) = state.event_category() {
+            state.managed_envelope.assume_event(category);
+        }
+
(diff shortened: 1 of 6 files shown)


Dav1dde:
@jjbayer 👀

jjbayer:
The idea behind this PR is still valid, but we might want to introduce [structured envelopes](https://github.com/getsentry/team-ingest/blob/feat/structured-envelopes/rfcs/0001-structured-items.md) first, so that it is valid to compute a summary even after the event has been parsed.

PR #2381, opened 2023-08-03
ref: Reduce metrics bucket size to 5MB

jan-auer:
The maximum bucket size for metrics messages was set to 50MB, which is large for
upstream web requests and larger than the usual maximum message size configured
on our Kafka brokers. As a more sensible default, this PR uses 5MB.

#skip-changelog


nikhars (approved):
🎉

olksdr (approved):
LGTM

It's fine to change default, but last time I checked we never ever send the messages even larger than 1 mb, all of them just 10s of kb.

jjbayer (approved):

PR #2385, opened 2023-08-04
feat(spans): Add `SAVEPOINT` and `SET` sql db actions

AbhiPrasad:
Add support for parsing `SET`, `SAVEPOINT`, `RELEASE SAVEPOINT`, and `ROLLBACK TO SAVEPOINT` db sql actions.

Adds table test to validate this.

ref https://www.notion.so/sentry/Improve-DB-Module-dropdowns-bb8b9776b7194cdb99ef26273aa452c7?d=f373818634b74aa58e0f00ac68250767#460bfb7733244d43a510edb778192572

#skip-changelog

### DIFF
--- a/relay-general/src/store/normalize/span/tag_extraction.rs
+++ b/relay-general/src/store/normalize/span/tag_extraction.rs
@@ -332,7 +332,7 @@ fn truncate_string(mut string: String, max_bytes: usize) -> String {
 }
 
 const ALLOWED_SQL_ACTION: &str =
-    "SELECT|INSERT|DELETE|UPDATE|SAVEPOINT|RELEASE SAVEPOINT|ROLLBACK TO SAVEPOINT";
+    "SELECT|INSERT|DELETE|UPDATE|SET|SAVEPOINT|RELEASE SAVEPOINT|ROLLBACK TO SAVEPOINT";
 
 /// Regex with a capture group to extract the database action from a query.
 ///
@@ -585,16 +585,14 @@ mod tests {
                 r#"DELETE FROM "sentry_groupinbox" WHERE "sentry_groupinbox"."id" IN (%s)"#,
                 "DELETE",
             ),
+            (r#"SET search_path TO my_schema, public"#, "SET"),
             (r#"SAVEPOINT %s"#, "SAVEPOINT"),
             (r#"RELEASE SAVEPOINT %s"#, "RELEASE SAVEPOINT"),
             (r#"ROLLBACK TO SAVEPOINT %s"#, "ROLLBACK TO SAVEPOINT"),
         ];
 
-        for (query, expected) in test_cases.iter() {
-            assert_eq!(
-                sql_action_from_query(query.to_owned()).unwrap(),
-                expected.to_owned()
-            )
+        for (query, expected) in test_cases.into_iter() {
+            assert_eq!(sql_action_from_query(query).unwrap(), expected)
         }
     }
 }
--- a/relay-general/src/store/normalize/span/tag_extraction.rs
+++ b/relay-general/src/store/normalize/span/tag_extraction.rs
@@ -331,11 +331,16 @@ fn truncate_string(mut string: String, max_bytes: usize) -> String {
     string
 }
 
+const ALLOWED_SQL_ACTION: &str =
+    "SELECT|INSERT|DELETE|UPDATE|SAVEPOINT|RELEASE SAVEPOINT|ROLLBACK TO SAVEPOINT";
+
 /// Regex with a capture group to extract the database action from a query.
 ///
-/// Currently, we're only interested in `SELECT`, `INSERT`, `DELETE` and `UPDATE` statements.
-static SQL_ACTION_EXTRACTOR_REGEX: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r#"(?i)(?P<action>(SELECT|INSERT|DELETE|UPDATE))"#).unwrap());
+/// Currently we have an explicit allow-list of database actions considered important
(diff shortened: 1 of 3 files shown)


relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -331,11 +331,16 @@ fn truncate_string(mut string: String, max_bytes: usize) -> String {
     string
 }
 
+const ALLOWED_SQL_ACTION: &str =
+    "SELECT|INSERT|DELETE|UPDATE|SAVEPOINT|RELEASE SAVEPOINT|ROLLBACK TO SAVEPOINT";
```
gggritso: Can we add `SET` to the list?

gggritso (approved):
Awesome!

relay-general/src/store/normalize/span/tag_extraction.rs
AbhiPrasad: yes good call, let me find a good test query

AbhiPrasad: [faa19e2](https://github.com/getsentry/relay/pull/2385/commits/faa19e2432128f1f1a3a84c2bf38f584dfb5d0a0)

AbhiPrasad:
btw to ingest team - I noticed that we were skipped changelog for these spans changes, any reason for that?

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -331,11 +331,16 @@ fn truncate_string(mut string: String, max_bytes: usize) -> String {
...
 /// Regex with a capture group to extract the database action from a query.
 ///
-/// Currently, we're only interested in `SELECT`, `INSERT`, `DELETE` and `UPDATE` statements.
-static SQL_ACTION_EXTRACTOR_REGEX: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r#"(?i)(?P<action>(SELECT|INSERT|DELETE|UPDATE))"#).unwrap());
+/// Currently we have an explicit allow-list of database actions considered important
+static SQL_ACTION_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
+    let action_regex: String = format!("(?i)(?P<action>({}))", ALLOWED_SQL_ACTION);
+    Regex::new(&action_regex).unwrap()
+});
```
olksdr: I would rather have the all defined `allowed` action directly in the regex, which personally for me is easier to grasp, instead of jumping from one variable to another. 

Is there a reason why this was split? Can we merge everything in one line of the `Regex::new` again?

Changing the docs on the `SQL_ACTION_EXTRACTOR_REGEX` would be enough, or even rename it to something including `_ALLOWED` if that makes the regex usage even clearer.

olksdr (commented):
Left few questions, otherwise it's good. 
Thanks also for adding a test! 👍

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -331,11 +331,16 @@ fn truncate_string(mut string: String, max_bytes: usize) -> String {
...
 
+const ALLOWED_SQL_ACTION: &str =
+    "SELECT|INSERT|DELETE|UPDATE|SET|SAVEPOINT|RELEASE SAVEPOINT|ROLLBACK TO SAVEPOINT";
+
 /// Regex with a capture group to extract the database action from a query.
 ///
-/// Currently, we're only interested in `SELECT`, `INSERT`, `DELETE` and `UPDATE` statements.
-static SQL_ACTION_EXTRACTOR_REGEX: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r#"(?i)(?P<action>(SELECT|INSERT|DELETE|UPDATE))"#).unwrap());
+/// Currently we have an explicit allow-list of database actions considered important
```
iker-barriocanal: nit:
```suggestion
/// Currently we have an explicit allow-list of database actions considered important.
```

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -560,4 +565,34 @@ mod tests {
...
+                r#"DELETE FROM "sentry_groupinbox" WHERE "sentry_groupinbox"."id" IN (%s)"#,
+                "DELETE",
+            ),
+            (r#"SET search_path TO my_schema, public"#, "SET"),
+            (r#"SAVEPOINT %s"#, "SAVEPOINT"),
+            (r#"RELEASE SAVEPOINT %s"#, "RELEASE SAVEPOINT"),
+            (r#"ROLLBACK TO SAVEPOINT %s"#, "ROLLBACK TO SAVEPOINT"),
+        ];
+
+        for (query, expected) in test_cases.into_iter() {
```
iker-barriocanal: I believe you don't need this, as the compiler handles it:
```suggestion
        for (query, expected) in test_cases {
```

iker-barriocanal (approved):

relay-general/src/store/normalize/span/tag_extraction.rs
AbhiPrasad: > Is there a reason why this was split

For git blame reasons - so that it becomes easier to tell if it was just the regex changing, or if it was functionality in `SQL_ACTION_EXTRACTOR_REGEX` changing.

If you feel strongly though I can change it back. Thoughts?

AbhiPrasad: I ended up inlining the regex in [d17203c](https://github.com/getsentry/relay/pull/2385/commits/d17203c9559e353b57463cec8611245a9ade169e), so changed.

PR #2387, opened 2023-08-07
ref(metrics): Move shared tagging to metric extraction

jan-auer:
Instead of inlining shared tag definitions into individual metrics' extraction
configuration, run a single extraction pass after extracting all metrics from a
payload instance.

Follow-up to https://github.com/getsentry/relay/pull/2355
Ref https://github.com/getsentry/team-ingest/issues/142

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/generic.rs
+++ b/relay-server/src/metrics_extraction/generic.rs
@@ -88,11 +88,11 @@ where
     }
 
     for mapping in &config.tags {
-        let mut tags_opt = None;
+        let mut lazy_tags = None;
 
         for metric in &mut metrics {
             if mapping.matches(&metric.name) {
-                let tags = tags_opt.get_or_insert_with(|| extract_tags(instance, &mapping.tags));
+                let tags = lazy_tags.get_or_insert_with(|| extract_tags(instance, &mapping.tags));
                 metric.tags.extend(tags.clone());
             }
         }
--- a/relay-server/src/metrics_extraction/generic.rs
+++ b/relay-server/src/metrics_extraction/generic.rs
@@ -57,7 +57,7 @@ where
         return metrics
     };
 
-    for metric_spec in config.metrics() {
+    for metric_spec in &config.metrics {
         if metric_spec.category != instance.category() {
             continue;
         }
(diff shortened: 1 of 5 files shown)


olksdr (approved):

relay-server/src/metrics_extraction/generic.rs
```diff
@@ -87,13 +87,24 @@ where
         });
     }
 
+    for mapping in &config.tags {
+        let mut tags_opt = None;
+
+        for metric in &mut metrics {
+            if mapping.matches(&metric.name) {
+                let tags = tags_opt.get_or_insert_with(|| extract_tags(instance, &mapping.tags));
```
iker-barriocanal: Are we doing the `get_or_insert_with` here to compute the tags lazily, in case no tag mapping matches? I'm not strongly opinionated, but I think extracting tags proactively could make this easier to read.

relay-server/src/metrics_extraction/spans/mod.rs
```diff
@@ -17,9 +16,9 @@ use crate::statsd::RelayTimers;
 /// This configuration is temporarily hard-coded here. It will later be moved to project config
 /// and be provided by the upstream.
 static SPAN_EXTRACTION_CONFIG: Lazy<MetricExtractionConfig> = Lazy::new(|| {
-    MetricExtractionConfig::new(
```
iker-barriocanal, line 20: I'm not sure if I prefer the constructor here.

iker-barriocanal (approved):

relay-server/src/metrics_extraction/generic.rs
jan-auer: That's the reason. There's a good chance that none of the extracted metrics matches, and in that case we won't compute the tags.
I'm happy to rewrite this for better readability, what if I call it `lazy_tags` instead?

relay-server/src/metrics_extraction/spans/mod.rs
jan-auer: I removed it for two reasons:
 1. This is a data struct that already had at least one public field and has to have a public signature. In that case, we can construct directly.
 2. The usage (right in the line below) had comments to name the fields. That indicates that the constructor has a bad signature.

relay-server/src/metrics_extraction/generic.rs
iker-barriocanal: Sounds good to me!

PR #2388, opened 2023-08-07
ref(metrics): Migrate conditional tagging

jan-auer:
This PR replaces the previously existing metric "conditional tagging" with the more generic metric extraction module. This introduces three changes to metric extraction: 

 1. Project configuration automatically converts conditional tagging settings to generic metric extraction settings. The old settings remain in the project config struct, however, so that they can be served to downstream Relays that are not yet updated. 
 2. The conditional tagging module is removed along and is no longer called from the envelope processor.
 3. Transaction metric extraction calls a temporarily exposed function from generic metric extraction and passes along the tagging configuration. This makes extraction and tagging a single step, which is in line with generic metric extraction.
 4. Inside `EnvelopeProcessor::extract_metrics`, the generic metric extraction config is now fetched at the very top in preparation of making this the only configuration. The plan is to normalize all other configs in sanitation, so that the envelope processor only has to know about the generic mechanism.

## TODOs

- [x] Convert from legacy to new config in project config sanitation
- [x] Remove legacy code
- [x] Temporarily apply tagging to legacy transaction metrics
- [x] Move tests

## Follow-up 

- Move transaction metric extraction to generic metric extraction in the same way as it was done for spans.
- Move the default configuration for transaction and span extraction to project config sanitation. This way, there is no more specific business logic for metric extraction in the EnvelopeProcessor.
- Clean up the metric extraction module.

#skip-changelog


relay-dynamic-config/src/project.rs
```diff
@@ -80,6 +80,28 @@ pub struct ProjectConfig {
...
 }
 
+impl ProjectConfig {
+    /// Validates fields in this project config and removes values that are partially invalid.
+    pub fn sanitize(&mut self) {
+        self.quotas.retain(Quota::is_valid);
+
+        // NOTE: This clones the rules so that they remain in the project state for old Relays that
+        // do not support generic metrics extraction. Once the migration is complete, this can be
+        // removed with a version bump of the transaction metrics config
```
olksdr:
```suggestion
        // removed with a version bump of the transaction metrics config.
```

olksdr (approved):
lgtm

PR #2391, opened 2023-08-08
chore(relay-redis): Add more redis pool configuration options

olksdr:
Also decrease the default timeouts to smaller numbers:

* `max_lifetime` - 5 min
* `idle_timeout` - 1 min
* `connection_timeout` - 5 sec

Also on each connection we set:
* `read_timeout` - 3 sec
* `write_timeout` 3 sec

Which will give fail and produce errors faster when the Redis cluster is unavailable.

These timeouts are still a bit conservative , but this change also gives now a possibility to configure those values in the config file, without re-compiling the `Relay`.

fix https://github.com/getsentry/team-ingest/issues/186

#skip-changelog


beezz (approved):

relay-quotas/src/redis.rs
```diff
@@ -695,7 +695,7 @@ mod tests {
 
         let rate_limiter = build_rate_limiter();
         let mut client = rate_limiter.pool.client().expect("get client");
-        let mut conn = client.connection();
+        let mut conn = client.connection().expect("Redis connection");
```
iker-barriocanal, line 698: Should we set a more meaningful error message? For example:
```suggestion
        let mut conn = client.connection().expect("Establish Redis connection");
```

relay-redis/src/config.rs
```diff
@@ -4,19 +4,61 @@ const fn default_max_connections() -> u32 {
...
+    ///
+    /// Calls to `Pool::get` will wait this long for a connection to become available before returning an error.
+    #[serde(default = "default_connection_timeout")]
+    pub connection_timeout: u64,
+    /// Sets the maximum lifetime of connections in the pool, in seconds.
+    #[serde(default = "default_max_lifetime")]
+    pub max_lifetime: u64,
+    /// Sets the idle timeout used by the pool, in seconds.
+    #[serde(default = "default_idle_timeout")]
+    pub idle_timeout: u64,
```
iker-barriocanal, line 44: What does the pool use this timeout for?

relay-redis/src/config.rs
```diff
@@ -115,3 +159,30 @@ server: "redis://127.0.0.1:6379"
         }
     }
 }
+
+#[test]
+fn test_redis_cluster_nodes_opts() {
```
iker-barriocanal: Is there any reason to not add a module for tests?

relay-redis/src/config.rs
```diff
@@ -115,3 +159,30 @@ server: "redis://127.0.0.1:6379"
...
+            cluster_nodes,
+            options,
+        } => {
+            assert_eq!(options.max_connections, 24);
+            assert_eq!(options.connection_timeout, 5);
+            assert_eq!(options.read_timeout, 10);
+            assert_eq!(options.write_timeout, 3);
+            assert!(cluster_nodes.contains(&String::from("redis://127.0.0.2:6379")));
+        }
+        e => panic!("Expected RedisConfig::SingleWithOpts but got {e:?}"),
```
iker-barriocanal: Since this is a test, what do you think about `unwrap`ing instead of this explicit panic?

iker-barriocanal (approved):

relay-redis/src/config.rs
olksdr: just wrong copy/paste, fixed it now

PR #2394, opened 2023-08-08
feat: add cross-origin-resource-policy response header

mdtro:
Fix for https://github.com/getsentry/sentry/issues/41225. 

Browsers are starting to require the [CORP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Resource-Policy) header it seems. Relay sets `Access-Control-Allow-Origin: *`, but this is not sufficient for those that have stricter headers set:

```
cross-origin-embedder-policy: require-corp
cross-origin-opener-policy: same-origin
```

`tower_http`'s CORS module does not support CORP/COOP/COEP so I just added the static header on the response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 23.7.3
+## Unreleased
 
 **Features**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 23.7.3
+
+**Features**:
+
+- Add `Cross-Origin-Resource-Policy` HTTP header to responses. ([#2394](https://github.com/getsentry/relay/pull/2394))
+
+
 ## 23.7.2
 
 **Features**:
--- a/relay-server/src/actors/server.rs
+++ b/relay-server/src/actors/server.rs
@@ -76,7 +76,7 @@ impl Service for HttpServer {
                 HeaderValue::from_static(constants::SERVER),
             ))
             .layer(SetResponseHeaderLayer::overriding(
-                HeaderName::from_static("Cross-Origin-Resource-Policy"),
+                HeaderName::from_static("cross-origin-resource-policy"),
                 HeaderValue::from_static("cross-origin"),
             ))
             .layer(NewSentryLayer::new_from_top())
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 23.7.3
+
+**Features**:
+
+- Add `Cross-Origin-Resource-Policy` HTTP header to responses. ([#2394](https://github.com/getsentry/relay/pull/2394))
+
+
```
```suggestion
## Unreleased

**Features**:

- Add `Cross-Origin-Resource-Policy` HTTP header to responses. ([#2394](https://github.com/getsentry/relay/pull/2394))


```

relay-server/src/actors/server.rs
```diff
@@ -75,6 +75,10 @@ impl Service for HttpServer {
                 header::SERVER,
                 HeaderValue::from_static(constants::SERVER),
             ))
+            .layer(SetResponseHeaderLayer::overriding(
+                HeaderName::from_static("Cross-Origin-Resource-Policy"),
```
olksdr:
```suggestion
                HeaderName::from_static("cross-origin-resource-policy"),
```

olksdr (changes requested):
It also would be great to add a test (an integration test) to make sure the relay responds with the correct headers.

mdtro:
> It also would be great to add a test (an integration test) to make sure the relay responds with the correct headers.

@olksdr Thank you for the suggestion. I wasn't too sure where to add them, but I was able to track down where the other CORS headers are tested as well.

Please let me know if the added test is sufficient. :)

tests/integration/test_basic.py
```diff
@@ -214,3 +214,23 @@ def test_compression(mini_sentry, relay, content_encoding):
...
+        "cross-origin",
+    ],
+)
+def test_corp_response_header(mini_sentry, relay, cross_origin_resource_policy):
+    project_id = 42
+    mini_sentry.add_basic_project_config(project_id)
+    relay = relay(mini_sentry)
+
+    response = relay.post(
+        "/api/42/store/?sentry_key=%s" % mini_sentry.get_dsn_public_key(project_id),
```
olksdr: I think you could write it something like this

```suggestion
        f"/api/42/store/?sentry_key={mini_sentry.get_dsn_public_key(project_id)}",
```

olksdr (approved):

tests/integration/test_basic.py
mdtro: I like this as well. I was following what looked like convention from the rest of the tests. Happy to accept this suggestion. :)

PR #2395, opened 2023-08-09
ref(globconfig): Set empty global config

iker-barriocanal:
Setting a defined list of keys for the global config requires further consideration of each steps, which currently slows down the iteration velocity. Simplifying this config postpones the considerations for later steps, and should help bring the velocity up again.

### Why is this PR useful?

The global config has two fields: `measurements` and `metric_conditional_tagging`. The latter is going to be removed soon as part of the generic metric extraction initiative so it will never be shipped as part of the global config, thus we're removing it. The former doesn't have this issue, but it adds extra steps that we're not interested in having at this point to try to ship a bit faster.

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -1,7 +1,4 @@
-use relay_general::store::MeasurementsConfig;
 use serde::{Deserialize, Serialize};
-
-use crate::TaggingRule;
 
 /// A dynamic configuration for all Relays passed down from Sentry.
 ///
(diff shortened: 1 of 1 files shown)


olksdr (commented):
What is the reason for this PR? 

The keys do not mater at all now, they are not used anyway. 
And also they optional, and if not send over wire then will be `None`, which also kind of does not do anything. 

In another PR https://github.com/getsentry/sentry/pull/53773 the config is not even sent. 

I would suggest just do not spread you attentions on this kind of PRs, and try to finish PR which is currently open https://github.com/getsentry/relay/pull/2320 and come back the global config format when it will actually get used.

iker-barriocanal:
@olksdr excellent question! I've updated the PR description to reflect it.

olksdr (approved):

PR #2396, opened 2023-08-09
feat(spans): Improve span data http method extraction

AbhiPrasad:
Check `http.request.method` and `method` fields when trying to extract `span.action` from an http span.

As a we start rolling out https://github.com/getsentry/team-sdks/issues/19 to more SDKs, we updated the span data field for HTTP methods from `http.method` to `http.request.method`. This is to match the [new OpenTelemetry conventions](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/http/#common-attributes).

image

Not sure how to best test this, feedback appreciated!

#skip-changelog


iker-barriocanal (commented):
The code looks good!

Adding new spans to [this test](https://github.com/getsentry/relay/blob/95cb257a5fdac22e71d3ab175653a3d4cac4d511/relay-server/src/metrics_extraction/spans/mod.rs#L116) and running it should update a couple of snapshots and the tag should be available in there.

AbhiPrasad:
Added snapshot tests with d0de3571221cd9fcbb4b38414d7cf89c3e84a905

I also changed `db.system` to be `postgresql` in the snapshot test, this should reflect real world usage a little more closely.

relay-general/src/store/normalize/span/tag_extraction.rs
```diff
@@ -200,15 +200,18 @@ pub(crate) fn extract_tags(span: &Span, config: &Config) -> BTreeMap<SpanTagKey,
             .and_then(|data| data.get("description.scrubbed"))
             .and_then(|value| value.as_str());
 
-        // TODO(iker): we're relying on the existance of `http.method`
+        // TODO(iker): we're relying on the existance of `http.method`/`http.response.method`
```
olksdr: Can we remove this TODO now and just put in the doc string the places we actually look in (and maybe why? E.g. open telemetry scheme etc.), which in my opinion makes more sense.

olksdr (approved):

iker-barriocanal (approved):
🚀

relay-general/src/store/normalize/span/tag_extraction.rs
AbhiPrasad: Yes I can amend this, good call.

PR #2397, opened 2023-08-10
ref(server): Simplify span metric extraction module

jan-auer:
Refactors the span metric extraction module without changing its behavior:

 - Removes `Result` from the function signature since the function is
   infallible.
 - Inline the implementation of the function to prepare for moving it into a
   larger event extraction function
 - Move the module file one level up from `spans/mod.rs` to `spans.rs`.
 - Remove a redundant test assertion on immutable data. The test kept a snapshot
   on spans, but the function does not mutate spans.

#skip-changelog


olksdr (approved):
lgtm

PR #2398, opened 2023-08-10
ref(metrics): Move span metric config to project config

jan-auer:
This PR adds a top-level function to extract metrics from all event
types. This comprises transaction metrics as well as span metrics.
Internally, it delegates most of the heavy-lifting to the generic metric
extraction function. The intent is to create one top-level extraction
function for every envelope item type that is capable of traversing data
and instantiating `FieldValueProvider`s.

All event-specific logic is therefore moved to the events module. Most
importantly, this includes the implementations of `Extractable`, but
also the logic for traversing through spans.

The configuration for span metric extraction is now located in the
project config (`relay-dynamic-config`). It is added automatically if
the project has the span extraction feature enabled.

#skip-changelog


relay-dynamic-config/src/defaults.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let config = project_config
+        .metric_extraction
+        .get_or_insert_with(MetricExtractionConfig::empty);
+
+    if config._span_metrics_extended {
+        return;
+    }
+
+    config.metrics.extend([
```
jan-auer, line 28: This used to be a static global, but since we now normalize existing config, a function was more convenient. Note that this code will move to Sentry backend in the immediate future.

To do this, we will:
- Move these specs to Sentry backend
- Add the config in Sentry backend
- Disable the feature flag for Relay
- Bump transaction metric extraction version so that Relays forward extractions if they do not support the new config yet.

relay-dynamic-config/src/metrics.rs
```diff
@@ -156,6 +158,16 @@ pub struct MetricExtractionConfig {
...
+
+    /// This config has been extended with default span metrics.
+    ///
+    /// Relay checks for the span extraction flag and adds built-in metrics and tags to this struct.
+    /// If the flag is `true`, this has already happened and should not be repeated.
+    ///
+    /// This is a temporary flag that will be removed once the transaction metric extraction version
+    /// is bumped to `2`.
+    #[serde(default)]
+    pub _span_metrics_extended: bool,
```
jan-auer, line 170: Similar to the above, this flag ensures that we do not add the same metric spec twice when we propagate configs from processing to PoP Relays. This will be removed as soon as the default is removed from Relay.

relay-dynamic-config/src/metrics.rs
```diff
@@ -353,19 +366,40 @@ mod tests {
...
+    if config._conditional_tags_extended {
+        return;
     }
+
+    config.tags.extend(TaggingRuleConverter {
+        rules: rules.iter().cloned().peekable(),
+        tags: Vec::new(),
+    });
+
+    config._conditional_tags_extended = true;
```
jan-auer, line 396: The contents of this functions have mostly moved from the caller. However this line is new, this was a bug in the original PR.

relay-server/src/actors/processor.rs
```diff
@@ -2082,7 +2082,7 @@ impl EnvelopeProcessorService {
             }
 
             if let Some(config) = config {
-                let metrics = crate::metrics_extraction::generic::extract_metrics(event, config);
+                let metrics = crate::metrics_extraction::event::extract_metrics(event, config);
```
jan-auer, line 2085: Note how there is now just one entry point for extracting metrics from an event: The event function. The idea is that there is one top-level extraction function for every data category that is able to traverse data internally. These functions will mostly need to forward to the generic extract function.

I'm planning a follow-up that makes the generic function a provided trait method to solidify this design.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -1,109 +1,66 @@
...
+}
+
+/// Extract metrics from an [`Event`].
+///
+/// The event must have a valid timestamp; if the timestamp is missing or invalid, no metrics are
+/// extracted. Timestamp and clock drift correction should occur before metrics extraction to ensure
+/// valid timestamps.
+///
+/// If this is a transaction event with spans, metrics will also be extracted from the spans.
+pub fn extract_metrics(event: &Event, config: &MetricExtractionConfig) -> Vec<Metric> {
```
jan-auer, line 45: This function and module is now responsible for extracting metrics from all event types. It mostly delegates to generic functionality, but also handles spans. That's also why the implementation of the `Extractable` trait has moved here. 

The `generic` module is now fully generic.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -543,7 +508,15 @@ mod tests {
         )
         .unwrap();
 
-        let metrics = extract_metrics(event.value_mut().as_mut().unwrap());
+        // Create a project config with the relevant feature flag. Sanitize to fill defaults.
+        let mut project = ProjectConfig {
+            features: [Feature::SpanMetricsExtraction].into_iter().collect(),
+            ..ProjectConfig::default()
+        };
+        project.sanitize();
```
jan-auer, line 516: Note how we also test that feature flag + sanitize + extraction all works together.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -1,109 +1,66 @@
...
+    }
 
+    fn timestamp(&self) -> Option<UnixTimestamp> {
+        self.timestamp
+            .value()
+            .and_then(|ts| UnixTimestamp::from_datetime(ts.0))
+    }
+}
+
+/// Extract metrics from an [`Event`].
```
olksdr:
```suggestion
/// Extracts metrics from an [`Event`].
```

olksdr (approved):
🚢  it!

PR #2400, opened 2023-08-11
ref: Faster case-insensitive comparison

TBS1996:
This method avoids extra allocations when comparing two strings. 

#skip-changelog

### DIFF
--- a/relay-general/src/pii/config.rs
+++ b/relay-general/src/pii/config.rs
@@ -35,7 +35,7 @@ pub struct LazyPattern {
 
 impl PartialEq for LazyPattern {
     fn eq(&self, other: &Self) -> bool {
-        self.raw.to_lowercase() == other.raw.to_lowercase()
+        self.raw.eq_ignore_ascii_case(&other.raw)
     }
 }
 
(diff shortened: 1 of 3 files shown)


iker-barriocanal (changes requested):
From `eq_ignore_ascii_case` docs:

> Same as to_ascii_lowercase(a) == to_ascii_lowercase(b), but without allocating and copying temporaries.

From [`to_ascii_lowercase` docs](https://doc.rust-lang.org/std/primitive.str.html#method.to_ascii_lowercase): 

> Returns a copy of this string where each character is mapped to its ASCII lower case equivalent.

and from [`to_lowercase` docs](https://doc.rust-lang.org/std/primitive.str.html#method.to_lowercase):

> Since some characters can expand into multiple characters when changing the case, this function returns a [String](https://doc.rust-lang.org/std/string/struct.String.html) instead of modifying the parameter in-place.

It seems to me this is not equivalent: `to_lowercase` converts to lowercase, while `to_ascii_lowercase` parses them to ASCII's lowercase equivalent. I assume it's possible for multiple unicode characters to map to the same ASCII character, making this not behave as expected. Is this accurate?

TBS1996:
hmm youre right, i could go on with just the headers line perhaps, since there shouldn't be non-ascii there, and since it's so frequently used there could be some nice gain there. assuming no header will be like `usÉr-ageNt` lol

PR #2405, opened 2023-08-16
ref(server): Refactor get_transaction_name in metric extraction

jan-auer:
Internal refactor that separates tracking a statsd metric for unparameterized
transaction names from the logic of extracting a transaction name from the
event. This will help with moving this functionality to `FieldValueProvider`.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -22,6 +22,10 @@ use relay_general::store::utils::{
 
 pub mod types;
 
+/// Placeholder for transaction names in metrics that is used when SDKs are likely sending high
+/// cardinality data such as raw URLs.
+const PLACEHOLDER_UNPARAMETERIZED: &str = "<< unparameterized >>";
+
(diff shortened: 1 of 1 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -55,11 +59,7 @@ fn extract_geo_country_code(event: &Event) -> Option<String> {
     geo.country_code.value().cloned()
 }
 
-fn is_low_cardinality(source: Option<&TransactionSource>) -> bool {
-    // `None` is used to mark a legacy SDK that does not send the transaction name,
-    // and we assume sends high-cardinality data. See `is_high_cardinality_transaction`.
```
jjbayer, line 60: Should we restore this comment somewhere?

jjbayer (approved):

relay-server/src/metrics_extraction/transactions/mod.rs
jan-auer: I thought about this, too. Since this just references the other function, and that function is pretty clear on what it does, I chose to remove it entirely.

PR #2406, opened 2023-08-16
ref(metrics): Validate transaction timestamps in light normalization

jan-auer:
As opposed to errors, transactions have a valid timestamp range defaulting to
(-5d..+1h). If transactions with an end timestamp outside of this range are
ingested, they are dropped as invalid transactions.

The reason for this check are performance metrics: Sentry metrics cannot be
backdated more than five days, and normalizing the transaction timestamp like
for errors creates strange artifacts in metrics.

However, this check was originally implemented in metric extraction code. In
this PR, it is moved to light normalization, which already performs a series of
validations of transaction events.

#skip-changelog

### DIFF
--- a/relay-general/src/store/transactions/processor.rs
+++ b/relay-general/src/store/transactions/processor.rs
@@ -548,6 +548,25 @@ mod tests {
     }
 
     #[test]
+    fn test_discards_when_timestamp_out_of_range() {
+        let mut event = new_test_event();
+
+        let processor = &mut TransactionsProcessor::new(
+            TransactionNameConfig::default(),
+            false,
+            None,
+            Some(UnixTimestamp::now()..UnixTimestamp::now()),
+        );
+
+        assert!(matches!(
+            process_value(&mut event, processor, ProcessingState::root()),
+            Err(ProcessingAction::InvalidTransaction(
+                "timestamp is out of the valid range for metrics"
+            ))
+        ));
+    }
+
+    #[test]
     fn test_replace_missing_timestamp() {
         let span = Span {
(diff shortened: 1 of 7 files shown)


jjbayer (approved):
I would restore the removed unit test, other than that LGTM!

olksdr (approved):

relay-general/src/store/transactions/processor.rs
```diff
@@ -138,62 +142,67 @@ impl<'r> TransactionsProcessor<'r> {
...
-            "trace context is missing trace_id",
-        ));
-    }
+                if let Some(ref range) = self.timestamp_range {
+                    let Some(timestamp) = UnixTimestamp::from_datetime(end.into_inner()) else {
+                        return Err(ProcessingAction::InvalidTransaction("invalid unix timestamp"));
+                    };
+                    if !range.contains(&timestamp) {
+                        return Err(ProcessingAction::InvalidTransaction(
+                            "timestamp is out of the valid range for metrics",
```
iker-barriocanal, line 190: nit: I'd remove the reason ("...for metrics"), if the metrics requirements change it's easy to forget about this and have an inaccurate error message.

iker-barriocanal (approved):

relay-general/src/store/transactions/processor.rs
jan-auer: This range is taken directly from the metrics aggregator config, that's why this message mentions metrics.

jan-auer:
Still need to fix integration tests before this can merge. 

**Edit**: Rerunning tests, they pass consistently on my machine.

PR #2407, opened 2023-08-17
feat(spans): Add event ID to extracted spans

jjbayer:
True standalone spans will not have an associated event ID. But because the current spans dataset expects it, add it to the kafka message of spans extracted from transactions for now.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/store.rs
+++ b/relay-server/src/actors/store.rs
@@ -205,6 +205,7 @@ impl StoreService {
                 ItemType::Span => self.produce_span(
                     scoping.organization_id,
                     scoping.project_id,
+                    event_id,
                     start_time,
                     item,
                 )?,
@@ -743,6 +744,7 @@ impl StoreService {
         &self,
         organization_id: u64,
         project_id: ProjectId,
+        event_id: Option<EventId>,
         start_time: Instant,
         item: &Item,
     ) -> Result<(), StoreError> {
@@ -761,6 +763,7 @@ impl StoreService {
             project_id,
             start_time: UnixTimestamp::from_instant(start_time).as_secs(),
             span,
+            event_id,
         });
 
         self.produce(KafkaTopic::Spans, organization_id, message)?;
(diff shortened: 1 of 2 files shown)


olksdr (approved):
lgtm!

Quick question: will consumers process these message with new field correctly?

relay-server/src/actors/store.rs
```diff
@@ -1032,12 +1035,17 @@ struct CheckInKafkaMessage {
...
     span: serde_json::Value,
     /// Time at which the span was received by Relay.
     start_time: u64,
     /// The project id for the current span.
     project_id: ProjectId,
+    /// The event id for the current span.
+    ///
+    /// Once spans are truly standalone, this field will be omitted.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    event_id: Option<EventId>,
```
iker-barriocanal, line 1048: Q: is this an option to make the transition to no event_id smoother?

iker-barriocanal (approved):

jjbayer:
> lgtm!
> 
> Quick question: will consumers process these message with new field correctly?

@olksdr yep AFAIK the [consumer](https://github.com/getsentry/sentry/blob/4349a91fcac183474020f850a241d5ef22ce52f4/src/sentry/spans/consumers/process/factory.py#L36) ignores fields it does not know, but I'll let @phacops confirm.

relay-server/src/actors/store.rs
jjbayer: That was the idea, yes. `handle_store_envelope` produces kafka messages for span items in an envelope regardless of whether the envelope contains an event.

phacops (approved):

phacops:
Indeed, we sort of ignore the field right now because we're not even looking at the right place for it.

https://github.com/getsentry/sentry/blob/8e90edebdd5775824ae4880f5e5fdd28a012036a/src/sentry/spans/consumers/process/factory.py#L48

PR #2410, opened 2023-08-17
ref(globconfig): Instrument global config service

iker-barriocanal:
This PR adds two metrics to the service to improve observability:
- Request duration, measuring all upstream request durations.
- Fetch successes, measuring the number of successful responses from the server. This metric ignores send errors, as the goal is to ensure global configs are received ok.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -19,7 +19,6 @@ use relay_system::{Addr, AsyncResponse, FromMessage, Interface, Service};
 use reqwest::Method;
 use serde::{Deserialize, Serialize};
 use tokio::sync::{mpsc, watch};
-use tokio::time::Instant;
 
 use crate::actors::upstream::{
     RequestPriority, SendQuery, UpstreamQuery, UpstreamRelay, UpstreamRequestError,
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -14,14 +14,17 @@ use std::sync::Arc;
 
 use relay_config::Config;
 use relay_dynamic_config::GlobalConfig;
+use relay_statsd::metric;
 use relay_system::{Addr, AsyncResponse, FromMessage, Interface, Service};
 use reqwest::Method;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/global_config.rs
```diff
@@ -187,11 +190,15 @@ impl GlobalConfigService {
...
         tokio::spawn(async move {
+            let request_start = Instant::now();
+
             let query = GetGlobalConfig::new();
             let res = upstream_relay.send(SendQuery(query)).await;
             // Internal forwarding should only fail when the internal receiver
             // is closed.
             internal_tx.send(res).await.ok();
+
+            metric!(timer(RelayTimers::GlobalConfigRequestDuration) = request_start.elapsed());
```
jjbayer: You can wrap the code you want to time instead:

```rust
  metric!(timer(RelayTimers::Whatever), {
      // Your code here. 
  });

**Review by jjbayer — APPROVED** [review]

File: relay-server/src/statsd.rs [hunk 2]
```diff
@@ -316,6 +316,8 @@ pub enum RelayTimers {
     OutcomeAggregatorFlushTime,
     /// Time in milliseconds spent on parsing, normalizing and scrubbing replay recordings.
     ReplayRecordingProcessing,
+    /// Total time spent to send a request to upstream and handle the response.
```
olksdr: Looking at the code, it's not really true. This only count the send request and receive the response, but not handle it.

relay-server/src/actors/global_config.rs
```diff
@@ -204,13 +211,21 @@ impl GlobalConfigService {
...
         match result {
-            Ok(Ok(config)) => match config.global {
-                Some(global_config) => {
-                    // Forwarding only fails when no subscribers.
-                    self.global_config_watch.send(Arc::new(global_config)).ok();
+            Ok(Ok(config)) => {
+                let mut success = false;
+                match config.global {
+                    Some(global_config) => {
+                        // Forwarding only fails when no subscribers.
```
olksdr: nit: it's not really forwarding but notification of the subscribers.

PR #2411, opened 2023-08-17
ref(globconfig): Extract subscription mechanism from processor

iker-barriocanal:
Follow-up to https://github.com/getsentry/relay/pull/2320. Extract the global config subscription mechanism from the processor to the service initialization.

Currently, the processor subscribes to the global config service by sending a message during its startup, and returning early if this fails finalizing the processor but allowing the rest of the services to continue to work. When subscribing to the global config fails, we're interested in failing as early as possible to actually enforce the dependency on that service. Additionally, we remove any future logic to retry subscribing which is too much complexity. On the other hand, Relay startup can take a bit longer and there's another moving piece that can fail during this time.

Note this is just a solution to the problem described above, and may not be the best. I'm open to ideas and suggestions to iterate on this approach or proceed with a different one.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -16,9 +16,9 @@ use flate2::Compression;
 use once_cell::sync::OnceCell;
 use relay_profiling::ProfileError;
 use serde_json::Value as SerdeValue;
+use tokio::sync::watch::Receiver;
 use tokio::sync::Semaphore;
 
-use crate::actors::global_config::{GlobalConfigManager, Subscribe};
 use crate::metrics_extraction::transactions::{ExtractedMetrics, TransactionExtractor};
 use crate::service::ServiceError;
 use relay_auth::RelayVersion;
@@ -525,6 +525,7 @@ impl FromMessage<RateLimitFlushBuckets> for EnvelopeProcessor {
 /// This service handles messages in a worker pool with configurable concurrency.
 #[derive(Clone)]
 pub struct EnvelopeProcessorService {
+    global_config_subscription: Receiver<Arc<GlobalConfig>>,
     global_config: Arc<GlobalConfig>,
     inner: Arc<InnerProcessor>,
 }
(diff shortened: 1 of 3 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -569,13 +569,13 @@ impl EnvelopeProcessorService {
             config,
             envelope_manager,
             project_cache,
-            global_config,
```
iker-barriocanal, line 572: This isn't strictly required for this PR. I deleted it because it was causing trouble with types and it's currently not used anywhere, so it's very hard to predict how it'll be used at this point. We can reintroduce it when we use it.

relay-server/src/service.rs
```diff
@@ -133,6 +133,7 @@ impl ServiceState {
 
         let global_config =
             GlobalConfigService::new(config.clone(), upstream_relay.clone()).start();
+        let global_config_subscription = global_config.send(Subscribe).await?;
```
olksdr, line 136: This way you expose the internal of the global config, which does not seem like a good idea. Also, it's already starting in the. correct order, first `GlobalConfigService` and then `EnveloperProcessortService` which subscribes to the `GlobalConfigService`, which already should cover the dependency you're talking about.

olksdr (changes requested):
I do not think these changes are necessary.  

Have you observed the problem or could reproduce in tests, that subscription is failing in the current setup? 
The current order of the service as they start, already ensures that they will work correctly and subscription will succeed. The one thing is needed though is a proper comment on why the global config service should always start first.

TBS1996:
oops, i had started working on this one too https://github.com/getsentry/relay/pull/2414

closed mine, anyway. you don't really need to make it async, if you create the channel in the servicestate::start function and pass it in (as done in my pr)

iker-barriocanal:
Postponing this for a later time.

iker-barriocanal:
Closing in favor of https://github.com/getsentry/relay/pull/2501.

PR #2413, opened 2023-08-17
fix(spans): Collapse lists of SQL columns

jjbayer:
Always collapse lists of columns, not just if they are delimited by `SELECT` and `WHERE`.

ref: [internal issue](https://www.notion.so/sentry/Collapse-lists-of-columns-regardless-of-position-in-query-ea33ab4d885d415fad77e63c90dcf474?pvs=4)

#skip-changelog


gggritso (approved):
👍🏻

olksdr (approved):

PR #2414, opened 2023-08-17
ref(server): Make glob config receiver in servicestate

TBS1996:
Instead of subscribing to the global config service from within envelopeprocessorservice, just make it in the servicestate.  

it should be faster, and we also don't need to handle the case where sending and receiving the subscriber fails.

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -16,9 +16,9 @@ use flate2::Compression;
 use once_cell::sync::OnceCell;
 use relay_profiling::ProfileError;
 use serde_json::Value as SerdeValue;
+use tokio::sync::watch::Receiver;
 use tokio::sync::Semaphore;
 
-use crate::actors::global_config::{GlobalConfigManager, Subscribe};
 use crate::metrics_extraction::transactions::{ExtractedMetrics, TransactionExtractor};
 use crate::service::ServiceError;
 use relay_auth::RelayVersion;
@@ -526,6 +526,7 @@ impl FromMessage<RateLimitFlushBuckets> for EnvelopeProcessor {
 #[derive(Clone)]
 pub struct EnvelopeProcessorService {
     global_config: Arc<GlobalConfig>,
+    global_config_receiver: Receiver<Arc<GlobalConfig>>,
     inner: Arc<InnerProcessor>,
 }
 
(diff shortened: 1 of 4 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -526,14 +526,14 @@ impl FromMessage<RateLimitFlushBuckets> for EnvelopeProcessor {
 #[derive(Clone)]
 pub struct EnvelopeProcessorService {
     global_config: Arc<GlobalConfig>,
+    global_config_receiver: Receiver<Arc<GlobalConfig>>,
```
TBS1996, line 529: it feels a bit ugly to have both of thes, although if we simply remove the global_config, we might end up having different global_configs for different parts of the processing.

relay-server/src/actors/processor.rs
```diff
@@ -2753,22 +2753,14 @@ impl Service for EnvelopeProcessorService {
         tokio::spawn(async move {
             let semaphore = Arc::new(Semaphore::new(thread_count));
 
-            let Ok(mut subscription) = self.inner.global_config.send(Subscribe).await else {
-                // TODO(iker): we accept this sub-optimal error handling. TBD
-                // the approach to deal with failures on the subscription
-                // mechanism.
-                relay_log::error!("failed to subscribe to GlobalConfigService");
-                return;
-            };
```
TBS1996, line 2762: removing this is the main benefit of this pr

relay-server/src/actors/processor.rs
```diff
@@ -2753,22 +2753,14 @@ impl Service for EnvelopeProcessorService {
...
                 let next_msg = async { tokio::join!(rx.recv(), semaphore.clone().acquire_owned()) };
 
                 tokio::select! {
                    biased;
 
                     // TODO(iker): deal with the error when the sender of the channel is dropped.
-                    _ = subscription.changed() => self.global_config = subscription.borrow().clone(),
+                    _ = self.global_config_receiver.changed() => self.global_config = self.global_config_receiver.borrow().clone(),
                     (Some(message), Ok(permit)) = next_msg => {
                         let service = self.clone();
```
TBS1996, line 2765: i was a bit worried about the cloning of the receiver, but according to the source code it seems trivial.  
 
we could avoid it, perhaps by making a broker with the global_config and the `inner` field, and moving all the methods to it. this could then be cloned separately from the receiver.

relay-server/src/service.rs
```diff
@@ -131,8 +132,10 @@ impl ServiceState {
         let outcome_aggregator =
             OutcomeAggregator::new(&config, outcome_producer.clone()).start_in(&outcome_runtime);
 
+        let (global_config_watch, global_config_receiver) = watch::channel(Arc::default());
```
TBS1996, line 135: im wondering now, why would we have a subscription message at all, couldn't every service needing a subscirption from any other service receive it in servicestate::start? this would avoid any complexities with messages and async code

olksdr:
@iker-barriocanal  @TBS1996  Is this PR about the same as https://github.com/getsentry/relay/pull/2411?

PR #2415, opened 2023-08-18
feat(server): Merge measurementsconfig from global config

TBS1996:
First step of actually using the global config.

This PR adds `MeasurementsConfig` to the `GlobalConfig` and adds the logic for merging it with the `MeasurementsConfig` from the `ProjectConfig`. 

Since the measurementsconfig is used in light normalization, and adding iterators to struct fields can be tricky, we have a new struct `DynamicMeasurementConfig` which holds an optional refernce to both Measurementconfigs, and we use that in `LightNormalizationConfig`.


relay-dynamic-config/src/project.rs
```diff
@@ -90,6 +91,13 @@ impl ProjectConfig {
         metrics::convert_conditional_tagging(self);
         defaults::add_span_metrics(self);
     }
+
+    /// Retrieves the measurements from project config and falls back on global config if missing.
+    pub fn measurements<'a>(&'a self, global_config: &'a Arc<GlobalConfig>) -> &MeasurementsConfig {
```
TBS1996: having a reference to an Arc feels pretty ugly, but I did it to keep consistent since the field value is an Option<MeasurementsConfig>

should I perhaps instead make the field value an option<arc<x>> ? or is there a better way altogether

relay-general/src/store/normalize.rs
```diff
@@ -62,6 +71,46 @@ pub struct MeasurementsConfig {
     max_custom_measurements: usize,
 }
 
+impl Default for MeasurementsConfig {
+    fn default() -> Self {
+        use MetricUnit::*;
```
TBS1996: i know glob imports are frowned upon, but I think it makes sense here for the sake of readability

relay-dynamic-config/src/global.rs
```diff
@@ -1,9 +1,55 @@
...
+            {"name": "stall_percentage", "unit": "ratio"},
+            {"name": "stall_total_time", "unit": "millisecond"},
+            {"name": "ttfb.requesttime", "unit": "millisecond"},
+            {"name": "ttfb", "unit": "millisecond"},
+            {"name": "time_to_full_display", "unit": "millisecond"},
+            {"name": "time_to_initial_display", "unit": "millisecond"}
+        ],
+        "maxCustomMeasurements": 10
+       }
+    }"#;
```
TBS1996: this is copied from the sentry side, i we should have a way to make sure it's synced

relay-dynamic-config/src/global.rs
```diff
@@ -1,9 +1,55 @@
...
 /// Values shared across all projects may also be included here, to keep
 /// [`ProjectConfig`](crate::ProjectConfig)s small.
-#[derive(Default, Clone, Debug, Serialize, Deserialize)]
+#[derive(Default, Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(default, rename_all = "camelCase")]
-pub struct GlobalConfig {}
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct GlobalConfig {
+    /// Configuration for measurements normalization.
+    pub measurements: MeasurementsConfig,
```
iker-barriocanal: Let's make all fields optional for future compatibility (e.g. we decide to discontinue this key in the future).

relay-dynamic-config/src/global.rs
iker-barriocanal: The follow-up https://github.com/getsentry/team-ingest/issues/193 should do it.

relay-dynamic-config/src/project.rs
```diff
@@ -90,6 +91,13 @@ impl ProjectConfig {
...
+        self.measurements
+            .as_ref()
+            .unwrap_or(&global_config.measurements)
```
iker-barriocanal: The accessor method should return a combination of both fields, not one or the other. It's valid for the global configs to have certain parameters and the project config to overwrite or extend these.

iker-barriocanal (changes requested):

relay-dynamic-config/src/project.rs
iker-barriocanal: Let's also document that `measurements` shouldn't be used directly and we should use the accessor instead.

TBS1996: >The accessor method should return a combination of both fields, not one or the other. It's valid for the global configs to have certain parameters and the project config to overwrite or extend these.

im not sure how this makes sense in this regards, it's not like it's key-value pair in this context, rather name + unit. Would it make sense to make them into hashsets though? so if the project config has more builtin measurements then itd append to the global config ones?   
 
thoughts, @jjbayer ? (since you're at work today)

jjbayer: You can probably have an iterator that does not emit duplicates. See for example https://github.com/getsentry/relay/pull/2245/files#diff-8bec88aaf27ba7c7052845939014ef1114dddb12da7b045fd81ec5e258334327R94-R99

TBS1996: some thoughts here..

the measurements is directly taken to the `lightnormalizationconfig`, so we'd have to store it as a `Box<dyn Iterator<Item = BuiltinMeasurementKey>>`, atm its just used once, so it would work, but i thought maybe its better to just store a vector for easier maintainability (like if we wanna iterate multiple times). 
 
if it's not acceptable to create a new vector then i could make it an iterator as described, or maybe theres a better way (?)

relay-dynamic-config/src/global.rs
```diff
@@ -1,9 +1,63 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(default, rename_all = "camelCase")]
-pub struct GlobalConfig {}
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct GlobalConfig {
+    /// Configuration for measurements normalization.
+    pub measurements: Option<MeasurementsConfig>,
+}
+
+impl Default for GlobalConfig {
```
iker-barriocanal: By default, the global config should be empty and we should be able to achieve this without implementing `Default` manually.

We're trying to guarantee forward compatibility so removing the key at some point should be a NOOP for measurements, and adding a default measurement config prevents that. Wrapping the config in an `Option` has pretty much the same reasoning.

relay-dynamic-config/src/global.rs
```diff
@@ -1,9 +1,63 @@
...
+    fn default() -> Self {
+        Self {
+            measurements: Some(MeasurementsConfig::default()),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
iker-barriocanal: IIUC the test below only needs `GlobalConfig`, wdyt about just importing it?

relay-dynamic-config/src/project.rs
```diff
@@ -90,6 +93,35 @@ impl ProjectConfig {
         metrics::convert_conditional_tagging(self);
         defaults::add_span_metrics(self);
     }
+
+    /// Combines the measurementsconfig from global config and project config.
```
iker-barriocanal: nit: could we link to the structures in the docs here?

relay-dynamic-config/src/project.rs
```diff
@@ -90,6 +93,35 @@ impl ProjectConfig {
...
+    pub fn measurements<'a>(
+        &'a self,
+        global_config: &'a Arc<GlobalConfig>,
+    ) -> Option<MeasurementsConfig> {
+        match (
+            self.measurements.as_ref(),
+            global_config.measurements.as_ref(),
+        ) {
+            (None, None) => None,
+            (None, Some(glob_measurements_config)) => Some(glob_measurements_config.clone()),
```
iker-barriocanal: Is this `clone` allocating new memory? The goal of having this accessor method is to provide a way to access measurements from both project and global configs without allocating new memory.

iker-barriocanal (changes requested):

relay-dynamic-config/src/project.rs
TBS1996: oh yeah, good call!

relay-dynamic-config/src/project.rs
TBS1996: yep, it is! I wrote my thoughts on it here: https://github.com/getsentry/relay/pull/2415#discussion_r1308348181   
 
I'll make it now without new allocations, so we can decide how to balance performance vs simplicity

TBS1996: these lifetimes issues driving me crazy

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -349,7 +351,7 @@ fn remove_invalid_measurements(
         }
 
         // Check if this is a builtin measurement:
-        for builtin_measurement in &measurements_config.builtin_measurements {
+        for builtin_measurement in &builtin_measurement_keys {
```
TBS1996: changed to a vec because this loop is called on each iteration, so the iterator would be exhausted after the first one. is there a better way to do it ? @iker-barriocanal

tests/integration/test_envelope.py
```diff
@@ -200,7 +200,6 @@ def test_normalize_measurement_interface(
         "inp": {"unit": "millisecond", "value": 100.14},
         "fp": {"unit": "millisecond", "value": None},
         "lcp": {"unit": "millisecond", "value": 420.69},
-        "missing_value": None,
```
TBS1996, line 203: Previously, this test would send an event through without a MeasuremntConfig, which would result in `fn remove_invalid_measurements` not being called, but now that we will get a measurementconfig from globalconfig, this function is called, and this line is dropped completely. To me this seems fine but please have a look that it's fine for me to edit the expected test output here.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +930,48 @@ impl Default for LightNormalizationConfig<'_> {
...
+    pub fn builtin_measurement_keys(&'a self) -> impl Iterator<Item = &'a BuiltinMeasurementKey> {
+        let from_project = self.project.iter().flat_map(|c| &c.builtin_measurements);
+        let from_global = self.global.iter().flat_map(|c| &c.builtin_measurements);
+
+        from_project.chain(from_global).unique()
+    }
+
+    /// Gets the minimum value of the max amount of custom measurements from project and global level
+    /// [`MeasurementsConfig`] if at least one is available.
+    pub fn max_custom_measurements(&'a self) -> Option<&'a usize> {
```
TBS1996: @iker-barriocanal  just a nit but does it make sense to return a reference to a primitive?

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -44,23 +45,33 @@ mod request;
...
 #[serde(default, rename_all = "camelCase")]
 pub struct MeasurementsConfig {
     /// A list of measurements that are built-in and are not subject to custom measurement limits.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    builtin_measurements: Vec<BuiltinMeasurementKey>,
+    pub builtin_measurements: Vec<BuiltinMeasurementKey>,
 
-    /// The maximum number of measurements allowed per event that are not known measurements.
-    max_custom_measurements: usize,
+    /// The maximum number of custom measurements allowed per event.
```
iker-barriocanal: I think the previous doc was more clear to understand.

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: I've been thinking about it and can't find any better solutions, unfortunately. Let's see if other reviewers have better suggestions.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +940,48 @@ impl Default for LightNormalizationConfig<'_> {
...
+    /// Constructor for [`DynamicMeasurementsConfig`].
+    pub fn new(
+        project: Option<&'a MeasurementsConfig>,
+        global: Option<&'a MeasurementsConfig>,
+    ) -> Self {
+        DynamicMeasurementsConfig { project, global }
+    }
+
+    /// Returns an iterator over the union of [`BuiltinMeasurementKey`]s from both the global and
+    /// project-level [`MeasurementsConfig`]s.
```
iker-barriocanal: Let's mention that the project config goes first, as that prioritization is relevant and may impact the outcome.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +940,48 @@ impl Default for LightNormalizationConfig<'_> {
...
+    pub fn builtin_measurement_keys(&'a self) -> impl Iterator<Item = &'a BuiltinMeasurementKey> {
+        let from_project = self.project.iter().flat_map(|c| &c.builtin_measurements);
+        let from_global = self.global.iter().flat_map(|c| &c.builtin_measurements);
+
+        from_project.chain(from_global).unique()
+    }
+
+    /// Gets the minimum value of the max amount of custom measurements from project and global level
+    /// [`MeasurementsConfig`] if at least one is available.
```
iker-barriocanal: Let's mention that if both exist, the most restrictive among both is taken (i.e. the minimum of the two). This may impact the outcome of light normalization.

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: I thought about this when I was coding, and I think it doesn't matter at all. The reason I added the reference is that I didn't see any reason not to, and seemed "more correct".

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -1416,6 +1480,60 @@ mod tests {
...
+
+        // If just global or just project is available, return their max custom measurements value.
+        {
+            let dynamic_config = DynamicMeasurementsConfig::new(Some(&proj), None);
+
+            assert_eq!(dynamic_config.max_custom_measurements().unwrap(), &3);
+
+            let dynamic_config = DynamicMeasurementsConfig::new(None, Some(&glob));
+            assert_eq!(dynamic_config.max_custom_measurements().unwrap(), &4);
+        }
```
iker-barriocanal: Q: is there a reason to make this a block?

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -2764,7 +2882,12 @@ mod tests {
         }))
         .unwrap();
 
-        normalize_measurements(&mut event, Some(&config), None);
+        let config = DynamicMeasurementsConfig {
+            project: Some(&config),
```
iker-barriocanal: The config passed to the project is the project-specific measurement config, could we rename that variable to make it easier to understand?

relay-base-schema/src/metrics.rs
```diff
@@ -25,7 +25,7 @@ pub fn is_valid_metric_name(name: &str) -> bool {
 /// measurements.
 ///
 /// Units and their precisions are uniquely represented by a string identifier.
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Default)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Default, PartialOrd)]
```
iker-barriocanal: Q: what do we need these `PartialOrd` for?

tests/integration/fixtures/mini_sentry.py
```diff
@@ -316,9 +318,7 @@ def get_project_config():
         version = flask_request.args.get("version")
 
         if version == "4" and flask_request.json.get("global"):
-            # There are no fields in global config for now
-            global_ = {}
-
+            global_ = GLOBAL_CONFIG
```
iker-barriocanal, line 319: Why did we decide to make the global config a constant there, over e.g. returning it from a function?

tests/integration/test_envelope.py
iker-barriocanal: Where was the `missing_value` entry added?

relay-dynamic-config/src/global.rs
```diff
@@ -1,9 +1,45 @@
...
+#[cfg(test)]
+mod tests {
+
+    use relay_base_schema::metrics::MetricUnit;
+    use relay_event_normalization::{BuiltinMeasurementKey, MeasurementsConfig};
+
+    use super::GlobalConfig;
+
+    #[test]
+    fn test_deserialize_global_config() {
```
iker-barriocanal: Let's make the whole roundtrip and ensure serialization is also correct and no data is lost.

iker-barriocanal (changes requested):
The PR looks solid, well done!

tests/integration/test_envelope.py
TBS1996: here: https://github.com/getsentry/relay/pull/2415/files/4cda1ea5f666d7ffe285863cee15a2958e856984#diff-12f50ce86ac73bfd9ac75e584534e50a3e3da9011a4f63ca9ee96740f1ed69b6R184

relay-event-normalization/src/normalize/mod.rs
TBS1996: ok, tried to rewrite it to be more clear!

relay-event-normalization/src/normalize/mod.rs
TBS1996: its to make it more clear that the comment above it is referring to whats inside the block, is there a better way to do this?

relay-base-schema/src/metrics.rs
TBS1996: oh damn, i needed them for a different way i tried of looking for unique elements, but currently i dont need, thanks for pointing this out!

tests/integration/fixtures/mini_sentry.py
TBS1996: is it preferred to use a function? If it's totally static wouldn't it be good to just have it as a variable?

relay-dynamic-config/src/global.rs
TBS1996: but this does test the roundtrip, although the function name wasnt updated to reflect it, updating that now!

relay-dynamic-config/src/project.rs
```diff
@@ -48,6 +48,7 @@ pub struct ProjectConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub dynamic_sampling: Option<SamplingConfig>,
     /// Configuration for measurements.
+    /// NOTE: do not access directly, use DynamicMeasurementsConfig.
```
TBS1996: Can't link to it directly without adding a `use crate.. dynamicmeasurementsconfig` statement. But when I do that, I get the linter error that the use statement is unused. any way to go around that?

relay-event-normalization/src/normalize/mod.rs
TBS1996: ok changed it to be more conventional

tests/integration/test_envelope.py
iker-barriocanal: It seems GitHub doesn't show it to me... could you tell me what file is it and I'll look for it?

relay-dynamic-config/src/global.rs
```diff
@@ -1,9 +1,45 @@
+use relay_event_normalization::MeasurementsConfig;
 use serde::{Deserialize, Serialize};
 
 /// A dynamic configuration for all Relays passed down from Sentry.
 ///
 /// Values shared across all projects may also be included here, to keep
 /// [`ProjectConfig`](crate::ProjectConfig)s small.
-#[derive(Default, Clone, Debug, Serialize, Deserialize)]
+#[derive(Default, Clone, Debug, Serialize, Deserialize, PartialEq)]
```
iker-barriocanal, line 8: Do we need this `PartialEq`?

relay-dynamic-config/src/project.rs
iker-barriocanal: Does the linter complain if you fully qualify the import in the docstring? That is, instead of importing it with `use` and then linking it, adding the full path in the link directly.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -44,23 +45,33 @@ mod request;
 mod stacktrace;
 
 /// Defines a builtin measurement.
-#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Hash, Eq)]
```
iker-barriocanal, line 47: Do we need this `PartialEq`?

tests/integration/fixtures/mini_sentry.py
iker-barriocanal: Sounds good to me, no strong opinions on that. I think we should move it to `mini_sentry` though.

iker-barriocanal (commented):
The PR looks good to me! I'd still like to double-check the missing `missing_value` in the test.

tests/integration/test_envelope.py
TBS1996: ah, this link should work:  https://github.com/getsentry/relay/blob/master/tests/integration/test_envelope.py#L184

as you can see it's just something they sent within the same test, so the value is kept when theres no measurementsconfig available, but now that we fetch from global, it will be removed due to this condition: https://github.com/getsentry/relay/blob/master/relay-event-normalization/src/normalize/mod.rs#L456

relay-dynamic-config/src/project.rs
TBS1996: nice tip, thx!

relay-dynamic-config/src/global.rs
TBS1996: im using it for the unit test, i could do
 ```rust
    impl PartialEq for GlobalConfig {
        fn eq(&self, other: &Self) -> bool {
            self.measurements == other.measurements
        }
    }
``` 

within the test module instead, although i feel adding a derive is easier

File: relay-event-normalization/src/normalize/mod.rs (hunk 23 above)
Comment (TBS1996) [reply relay-event-normalization/src/normalize/mod.rs:47]: im using it for unit test as well, and also for the partialeq of globalconfig

File: relay-event-normalization/src/normalize/mod.rs [hunk 24]
```diff
@@ -1488,13 +1489,13 @@ mod tests {
 
         let proj = MeasurementsConfig {
             builtin_measurements: vec![foo.clone(), bar.clone()],
-            max_custom_measurements: 3,
+            max_custom_measurements: 4,
```
iker-barriocanal, line 1490: Does the test pass with 3? I wonder where the 3 -> 4 change comes from.

TBS1996: yeah it also passes with 3.  
 
in this test, we insert 4 measurements, but the output should only be 3 because of duplicates. if we had the max custom measurements to 3, then we cant be sure if the output is 3 because we remove duplicates (as we wanna test) or because it removes excessive measurements (which is tested elsewhere), does this make sense?

iker-barriocanal: Seems reasonable.

iker-barriocanal (approved):
This looks good to me, let's see what others think.

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@
 - Prevent sporadic data loss in `EnvelopeProcessorService`. ([#2454](https://github.com/getsentry/relay/pull/2454))
 - Prevent panic when android trace contains invalid start time. ([#2457](https://github.com/getsentry/relay/pull/2457))
 
+**Features**:
```
jan-auer: There's a duplicate `Features` section, let's move the entry up to the existing one.

relay-dynamic-config/Cargo.toml
```diff
@@ -15,6 +15,7 @@ default = []
 [dependencies]
 anyhow = { workspace = true }
 assert-json-diff = "2.0.2"
+itertools = { workspace = true }
```
jan-auer: It looks like this dependency is not actually used, can we remove it?

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +941,48 @@ impl Default for LightNormalizationConfig<'_> {
...
+    }
+
+    /// Gets the max custom measurements value from the [`MeasurementsConfig`] from project level or
+    /// global level. If both of them are available, it will choose the most restrictive.
+    pub fn max_custom_measurements(&'a self) -> Option<&'a usize> {
+        match (&self.project, &self.global) {
+            (None, None) => None,
+            (None, Some(global)) => Some(&global.max_custom_measurements),
+            (Some(project), None) => Some(&project.max_custom_measurements),
+            (Some(project), Some(global)) => Some(min(
```
jan-auer: Ideally, remove the import and qualify `std::cmp::min` directly or use the `min` method: `Some(a.min(b))`

relay-event-normalization/src/normalize/mod.rs
jan-auer: One way you could solve this is by passing the view type in here. You've already created `DynamicMeasurementsConfig` and expose iterators. So you can pass this in and iterate them explicitly here.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +941,48 @@ impl Default for LightNormalizationConfig<'_> {
...
             enable_trimming: false,
+            dynamic_measurements_config: None,
+        }
+    }
+}
+
+/// Container for global and project level [`MeasurementsConfig`]. The purpose is to handle
+/// the merging logic.
+#[derive(Clone, Debug)]
+pub struct DynamicMeasurementsConfig<'a> {
```
jan-auer, line 941: _nit_: What would the "Dynamic" part in the name refer to? Isn't this rather a "merged" or "combined" measurements config?

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +941,48 @@ impl Default for LightNormalizationConfig<'_> {
...
+        DynamicMeasurementsConfig { project, global }
+    }
+
+    /// Returns an iterator over the union of [`BuiltinMeasurementKey`]s from both the global and
+    /// project-level [`MeasurementsConfig`]s, with the elements from project config coming first.
+    pub fn builtin_measurement_keys(&'a self) -> impl Iterator<Item = &'a BuiltinMeasurementKey> {
+        let from_project = self.project.iter().flat_map(|c| &c.builtin_measurements);
+        let from_global = self.global.iter().flat_map(|c| &c.builtin_measurements);
+
+        from_project.chain(from_global).unique()
```
jan-auer: The `unique` is a rather expensive operation since it needs to collect all elements. Could you instead do this with a manual `filter` function?

Basically, iterate all entries from one list, and then while iterating the ones from the second one, skip those that are in the first.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -847,9 +870,10 @@ pub struct LightNormalizationConfig<'a> {
 
     /// Configuration for measurement normalization in transaction events.
     ///
-    /// If provided, normalization truncates custom measurements and adds units of known built-in
-    /// measurements.
-    pub measurements_config: Option<&'a MeasurementsConfig>,
+    /// Has an optional [`MeasurementsConfig`] from both the project and the global level.
+    /// If at least one is provided, then normalization will truncate custom measurements
+    /// and add units of known built-in measurements.
+    pub dynamic_measurements_config: Option<DynamicMeasurementsConfig<'a>>,
```
jan-auer: Could we call this field just `measurements`? Both the `dynamic` and `config` name components do not help to increase clarity.

jan-auer (commented):
Preliminary review.

relay-dynamic-config/Cargo.toml
TBS1996: we use it on `.unique()`, but i'll try to remove that function anyway like you mentioned, in which case i'll remove the dependency

relay-event-normalization/src/normalize/mod.rs
TBS1996: The crate is called `relay-dynamic-config` so it seems to fit the naming convention, this is the measurements part of the dynamic config.

relay-event-normalization/src/normalize/mod.rs
TBS1996: https://github.com/getsentry/relay/pull/2415/files#diff-f09c042e62d13a4f85957ebfe2ee58d702960ac82b49955f0ef65d7001f09d50R956-R972  
 
Would this be better? 

also, a thought, maybe in the future we could cache it in the `ProjectState` struct, like when we fetch it we create the list of unique measurement keys there, so it could be re-used many times. Although then it'd have to handle updating the cache whenever there's a new global config.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -966,10 +956,19 @@ impl<'a> DynamicMeasurementsConfig<'a> {
...
     /// project-level [`MeasurementsConfig`]s, with the elements from project config coming first.
     pub fn builtin_measurement_keys(&'a self) -> impl Iterator<Item = &'a BuiltinMeasurementKey> {
-        let from_project = self.project.iter().flat_map(|c| &c.builtin_measurements);
-        let from_global = self.global.iter().flat_map(|c| &c.builtin_measurements);
+        let from_project = self
+            .project
+            .iter()
+            .flat_map(|c| &c.builtin_measurements)
+            .collect::<Vec<_>>();
+        let from_project_clone = from_project.clone();
```
iker-barriocanal: Do we need two vectors? The `contains` below is O(n), could we have an iterator and a set? We could use the set for checking duplicates and the iterator to chain the other one.

TBS1996: I did try with a set but ran into many problems, and it's also not clear that its actually better. the amount of elements are likely quite small so the overhead of a hashset might be more significant than the time it takes to iterate through it. perhaps it's different with a btreeset

iker-barriocanal: This seems to work:

```rust
    pub fn builtin_measurement_keys(&'a self) -> impl Iterator<Item = &'a BuiltinMeasurementKey> {
        let from_project = self.project.iter().flat_map(|c| &c.builtin_measurements);
        let mut project_set = HashSet::<&BuiltinMeasurementKey>::new();
        project_set.extend(from_project.clone());

        let from_global = self
            .global
            .iter()
            .flat_map(|c| &c.builtin_measurements)
            .filter(move |&k| !project_set.contains(k));

        from_project.into_iter().chain(from_global)
    }
```

iker-barriocanal: Resolving this comment as the size of the collection is not expected to be high enough for the performance benefit.

relay-event-normalization/src/normalize/mod.rs
jan-auer: Instead of collecting to a vec, just call the iterator every time here.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +931,57 @@ impl Default for LightNormalizationConfig<'_> {
...
+    }
+
+    /// Returns an iterator over the union of [`BuiltinMeasurementKey`]s from both the global and
+    /// project-level [`MeasurementsConfig`]s, with the elements from project config coming first.
+    pub fn builtin_measurement_keys(&'a self) -> impl Iterator<Item = &'a BuiltinMeasurementKey> {
+        let from_project = self
+            .project
+            .iter()
+            .flat_map(|c| &c.builtin_measurements)
+            .collect::<Vec<_>>();
```
jan-auer: Please avoid collecting into a vector here; this method can be implemented without allocating.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +931,57 @@ impl Default for LightNormalizationConfig<'_> {
...
+impl<'a> DynamicMeasurementsConfig<'a> {
+    /// Constructor for [`DynamicMeasurementsConfig`].
+    pub fn new(
+        project: Option<&'a MeasurementsConfig>,
+        global: Option<&'a MeasurementsConfig>,
+    ) -> Self {
+        DynamicMeasurementsConfig { project, global }
+    }
+
+    /// Returns an iterator over the union of [`BuiltinMeasurementKey`]s from both the global and
```
jan-auer: Please rephrase the doc comments on this type; see the [coding guidelines](https://develop.sentry.dev/rust/#writing-doc-comments).

jan-auer (changes requested):

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -918,6 +930,59 @@ impl Default for LightNormalizationConfig<'_> {
...
+        global: Option<&'a MeasurementsConfig>,
+    ) -> Self {
+        DynamicMeasurementsConfig { project, global }
+    }
+
+    /// Returns an iterator over the merged builtin measurement keys.
+    ///
+    /// Items from the project config are prioritized over global config, and
+    /// there are no duplicates.
+    pub fn builtin_measurement_keys(&'a self) -> impl Iterator<Item = &'a BuiltinMeasurementKey> {
```
jan-auer: Optional: The correct signature for this would be:

```suggestion
    pub fn builtin_measurement_keys(&self) -> impl Iterator<Item = &'a BuiltinMeasurementKey> + '_ {
```

relay-event-normalization/src/normalize/mod.rs
jan-auer: In general, it's better to return `usize` directly here since that removes an indirection and allows for further optimization. You can usually opt to return `Copy` types directly, unless they are extremely large. Particularly with `usize`, the reference is just an obstacle.

jan-auer (approved):

PR #2416, opened 2023-08-18
ci(gocd): Deploy PoPs automatically

jan-auer:
With Relay deployments now split across 10 pipelines, the manual approval step
for PoPs has become a frequent friction point and is often forgotten. In favor
of a fully automated rollout pipeline, PoPs should now be deployed
automatically, followed by all regions.

#skip-changelog


jjbayer (approved):
This gives us one less checkpoint to verify if everything is still healthy, so we should probably make a habit out of doing a Canary deploy before deploying the rest of the pods.

iker-barriocanal (approved):
YASS

olksdr (changes requested):
if I remember correctly you will have to check in the generated pipeline for now as well. Otherwise this won't work

olksdr (approved):

PR #2417, opened 2023-08-18
feat(spans): Add a span for the transaction itself, add Segment ID

jjbayer:
In https://github.com/getsentry/relay/pull/2350 we set up the extraction of spans from transactions & publishing on a dedicated spans topic.

This PR contains two fixes:

* The initial PR did not emit a span for the transaction itself, which is conceptually also a span.
* Add a `segment_id` (= `span_id` of the transaction) and an `is_segment` flag to every span.

#skip-changelog


relay-general/src/protocol/span.rs
```diff
@@ -38,7 +39,16 @@ pub struct Span {
     #[metastructure(required = "true")]
     pub trace_id: Annotated<TraceId>,
 
-    /// The status of a span
+    /// A unique identifier for a segment within a trace (8 byte hexadecimal string).
+    ///
+    /// For spans embedded in transactions, the `segment_id` is the `span_id` of the containing
+    /// transaction.
+    pub segment_id: Annotated<SpanId>,
```
jjbayer, line 46: See See https://github.com/getsentry/sentry-kafka-schemas/blob/ba47924510aa9b1e6d1ec67e69f40e2948a365b9/schemas/snuba-spans.v1.schema.json#L33-L36

relay-server/src/actors/processor.rs
```diff
@@ -2252,20 +2254,38 @@ impl EnvelopeProcessorService {
             return;
         };
 
-        // Extract.
-        let Some(spans) = state.event.value().and_then(|e| e.spans.value()) else { return };
-        for span in spans {
+        // Extract transaction as span.
```
```suggestion
        // Extract transaction as a span.
```

relay-server/src/actors/processor.rs
```diff
@@ -2252,20 +2254,38 @@ impl EnvelopeProcessorService {
...
+                // as part of normalization once standalone spans reach wider adoption.
+                let mut span = span.clone();
+                let Some(inner_span) = span.value_mut() else { continue };
+                inner_span.segment_id = transaction_span.segment_id.clone();
+                inner_span.is_segment = Annotated::new(false);
+                add_span(span);
+            }
         }
+
+        // Add transaction span as envelope item.
```
```suggestion
        // Add transaction span as an envelope item.
```

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -526,7 +526,7 @@ mod tests {
         };
 
         let extracted = extractor.extract(event.value().unwrap()).unwrap();
-        insta::assert_debug_snapshot!(event.value().unwrap().spans, @r#"
+        insta::assert_debug_snapshot!(event.value().unwrap().spans, @r###"
```
olksdr: Do we have to have 3 `###` here?  The new `beta` rust linter does not like it and the new release won't either and will make your CI fail, since we treat warnings as errors.  It would be good idea to keep this to the linter likings 😄 , one `#` where it's possible

olksdr (approved):

relay-server/src/metrics_extraction/transactions/mod.rs
jjbayer: `insta` did that when it updated the snapshot. Manually reverted now.

relay-general/src/protocol/span.rs
```diff
@@ -59,6 +69,29 @@ pub struct Span {
...
+impl From<&Event> for Span {
+    fn from(event: &Event) -> Self {
+        let mut span = Self {
+            timestamp: event.timestamp.clone(),
+            start_timestamp: event.start_timestamp.clone(),
+            is_segment: Some(true).into(),
+            ..Default::default()
+        };
+
+        if let Some(trace_context) = event.context::<TraceContext>().cloned() {
```
iker-barriocanal, line 81: Even with this condition, an error event will also be converted into a span. Should we use [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) instead of `From`, and error if the event isn't a transaction?

jjbayer: I considered that, but then thought it's fine to leave this responsibility to the caller, since there is no formal reason why a conversion from error to span should fail. It might be a good safeguard though, let me reconsider.

iker-barriocanal: > since there is no formal reason why a conversion from error to span should fail

Shouldn't that case be disallowed, by the definition of an error? I think leaving this responsibility to the caller is okay, but it makes it easier to introduce bugs IMO. I'll let you decide.

PR #2418, opened 2023-08-18
ci: Remove generated pipelines and update gocd-jsonnet

mattgauntseo-sentry:
This PR does the following:

- update gocd-jsonnet lib (new feature that allows excluding a region in case of an incident)
- removed the generated pipelines because GoCD can use the jsonnet file directly now
- deletes the old pipeline as it's not needed anymore
- moves the experimental pipeline to the end of the pipeline list (let me know if you'd prefer to keep it at the start)

#skip-changelog


gocd/README.md
```diff
@@ -2,18 +2,25 @@
 
 Relay is in the process of moving to a set of rendered jsonnet pipelines.
```
olksdr: I think this can be removed?

olksdr (approved):

olksdr:
I would still keep the experimental in the beginning of the list, since this is the first option we often want to deploy and see if things work before everything else.

mattgauntseo-sentry:
Going to merge after hackweek

PR #2420, opened 2023-08-21
fix(server): Drop runtimes in sync context on error

### LINKED ISSUE #2427: Improve error message on invalid config
When starting internal services fails (e.g. because of faulty config), created tokio runtimes are dropped in an asynchronous context, leading to a panic with the following error message:

> thread 'main' panicked at 'Cannot drop a runtime in a context where blocking is not allowed.
This happens when a runtime is dropped from within an asynchronous context.'

This makes it hard to debug the actual problem, especially in production (see [INC-486](https://getsentry.atlassian.net/browse/INC-486)).

[INC-486]: https://getsentry.atlassian.net/browse/INC-486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jjbayer:
Tokio runtimes should not be dropped from within an asynchronous context. In the success case, this was previously guaranteed by explicitly dropping the service state. When starting the service state failed (for example because of a faulty kafka config), relay would still panic. This PR prevents the panic by keeping the runtimes alive outside of the service state.

Before:

```
thread 'main' panicked at 'Cannot drop a runtime in a context where blocking is not allowed.
This happens when a runtime is dropped from within an asynchronous context.'
```

After:

```
error=failed to create kafka producer: invalid kafka config
error.sources=[Client config error: No such configuration property ...
```

ref: [sentry issue](https://sentry.my.sentry.io/organizations/sentry/issues/399521), 

#skip-changelog

Fixes https://github.com/getsentry/relay/issues/2427


relay-server/src/actors/store.rs
```diff
@@ -56,12 +55,8 @@ impl Producer {
...
             .filter(|t| **t != KafkaTopic::Outcomes || **t != KafkaTopic::OutcomesBilling)
         {
-            let kafka_config = &config
-                .kafka_config(*topic)
-                .map_err(|_| ServiceError::Kafka)?;
-            client_builder = client_builder
-                .add_kafka_topic_config(*topic, kafka_config)
-                .map_err(|_| ServiceError::Kafka)?
+            let kafka_config = &config.kafka_config(*topic)?;
+            client_builder = client_builder.add_kafka_topic_config(*topic, kafka_config)?;
```
jjbayer, line 63: This change helps to get a more specific error when the kafka config cannot be loaded. For example,

```
error=failed to create kafka producer: invalid kafka config
error.sources=[Client config error: No such configuration property: "foo" foo value]
```

instead of just

```
error=could not initialize kafka producer
```

relay-server/src/lib.rs
```diff
@@ -286,25 +298,34 @@ pub fn run(config: Config) -> anyhow::Result<()> {
     let config = Arc::new(config);
     relay_log::info!("relay server starting");
 
-    // Create the main tokio runtime that all services run in.
+    // Creates the main runtime.
```
jjbayer, line 289:
```suggestion
    // Create the main runtime.
```

jan-auer (approved):
Beautiful <3

PR #2421, opened 2023-08-21
Implement Network Error Logging

oioki:
The first part of the NEL ([Network Error Logging](https://developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging)) implementation.

* Adds a new `Nel` EventType
* Adds a new `Nel` protocol with minimal validation (i.e. only specific fields are captured)
* Adds a new `/api/:project_id/nel/` endpoint for NEL reports ingestion
  * simple content type validation
  * splits payload into separate envelopes (a single HTTP NEL request could contain several independent reports)
* `user.ip_address` field is set to the IP address of the request (real user's IP address)
* An event is enriched with browser information derived from the request's `User-Agent` header

Related PRs:
https://github.com/getsentry/sentry/pull/55135
https://github.com/getsentry/sentry-kafka-schemas/pull/177


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Features**:
 
 - Scrub span descriptions with encoded data images. ([#2560](https://github.com/getsentry/relay/pull/2560))
+- Added new endpoint for NEL (Network Error Logging) reports ingestion. ([#2421](https://github.com/getsentry/relay/pull/2421))
```
```suggestion
- Added support for NEL (Network Error Logging) reports ingestion. ([#2421](https://github.com/getsentry/relay/pull/2421))
```

py/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Added new endpoint for NEL (Network Error Logging) reports ingestion. ([#2421](https://github.com/getsentry/relay/pull/2421))
```
olksdr:
```suggestion
- Added support for NEL (Network Error Logging) reports ingestion. ([#2421](https://github.com/getsentry/relay/pull/2421))
```

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,302 @@
...
+};
+#[cfg(feature = "jsonschema")]
+use relay_jsonschema_derive::JsonSchema;
+use relay_protocol::{Annotated, Empty, FromValue, IntoValue};
+use thiserror::Error;
+
+/// The NEL parsing errors.
+#[derive(Debug, Error)]
+pub enum NelError {
+    /// Unexpected format
```
olksdr:
```suggestion
    /// Unexpected format.
```

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,302 @@
...
+    #[error("unexpected format")]
+    InvalidNel,
+    /// Incoming Json is unparsable.
+    #[error("incoming json is unparsable")]
+    InvalidJson(#[from] serde_json::Error),
+}
+
+/// Inner (useful) part of a NEL report.
+///
+/// See `Nel` for meaning of fields.
```
olksdr:
```suggestion
/// See [`Nel`] for fields documentation. 
```

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,302 @@
...
+    sampling_fraction: Option<f64>,
+    server_ip: Option<String>,
+    status_code: Option<u64>,
+    #[serde(rename = "type")]
+    ty: Option<String>,
+}
+
+/// Defines external, RFC-defined schema we accept, while `Nel` defines our own schema.
+///
+/// See `Nel` for meaning of fields.
```
olksdr:
```suggestion
/// See [`Nel`] for fields documentation.
```

relay-server/src/actors/processor.rs
```diff
@@ -1721,6 +1751,13 @@ impl EnvelopeProcessorService {
                     );
                     error
                 })?
+        } else if let Some(item) = nel_item {
+            relay_log::trace!("processing nel report");
+            self.event_from_nel_report(item, envelope.meta())
+                .map_err(|error| {
+                    relay_log::error!(error = &error as &dyn Error, "failed to extract nel report");
```
```suggestion
                    relay_log::error!(error = &error as &dyn Error, "failed to extract NEL report");
```

relay-server/src/endpoints/nel.rs
```diff
@@ -0,0 +1,67 @@
+//! Endpoints for Network Error Logging reports.
```
olksdr:
```suggestion
//! Endpoint for Network Error Logging (NEL) reports.
//!
//! It split list of incoming events from the envelope into separate envelope with 1 item inside.
//! Which later get failed by the service infrastructure. 
```

olksdr (commented):
Added few suggestions to the comments and docs for now.

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,302 @@
...
+    age: Option<i64>,
+    #[serde(rename = "type")]
+    ty: Option<String>, // always "network-error"
+    url: Option<String>,
+    user_agent: Option<String>,
+    body: NelBodyRaw,
+}
+
+impl NelReportRaw {
+    fn into_protocol(self) -> Nel {
```
jjbayer: I see we do the same for security reports, but why are we deserializing into `NewlReportRaw` and then manually transforming into `Nel`? Is there no way to parse `Nel` directly?

olksdr: Removed `Raw` types and use `Nel` directly.

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,249 @@
...
+    Timestamp,
+};
+#[cfg(feature = "jsonschema")]
+use relay_jsonschema_derive::JsonSchema;
+use relay_protocol::{Annotated, Empty, FromValue, IntoValue};
+use thiserror::Error;
+
+/// The NEL parsing errors.
+#[derive(Debug, Error)]
+pub enum NelError {
```
jjbayer: nit: The Relay code-base is inconsistent about this, but in principle we [want](https://develop.sentry.dev/rust/#naming) to [avoid redundant prefixes](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-%5Brfc-356%5D). 

That is, `nel::Error` instead of `nel::NelError`, `nel::Body` instead of `nel::NelBody`, etc.

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,249 @@
...
+/// Models the content of a NEL report.
+///
+/// NOTE: This is the structure used inside the Event (serialization is based on Annotated
+/// infrastructure). We also use a version of this structure to deserialize from raw JSON
+/// via serde.
+///
+/// See <https://w3c.github.io/network-error-logging/>
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct Nel {
```
jjbayer: nit: See prefix comment above.

```suggestion
pub struct Report {
```

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,249 @@
...
+                        Annotated::new("server_ip".to_string()),
+                        Annotated::new(server_ip.to_string()),
+                    ))));
+                }
+            }
+
+            // Set the timestamp on the event when it actually occured.
+            let now: DateTime<Utc> = Utc::now();
+            let event_time = now.sub(Duration::milliseconds(*nel.age.value().unwrap_or(&0)));
+            event.timestamp = Annotated::new(Timestamp::from(event_time));
```
jjbayer: `.sub()` might panic.

```suggestion
            if let Some(event_time) = now.checked_sub_signed(Duration::milliseconds(*nel.age.value().unwrap_or(&0))) {
                event.timestamp = Annotated::new(Timestamp::from(event_time));
            }
```

relay-server/src/endpoints/nel.rs
```diff
@@ -0,0 +1,70 @@
...
+    let items: Value =
+        serde_json::from_slice(&params.body).map_err(BadStoreRequest::InvalidJson)?;
+
+    // Iterate only if the body contains the list of the items.
+    for item in items.as_array().into_iter().flatten() {
+        let mut envelope = Envelope::from_request(Some(EventId::new()), params.meta.clone());
+        let mut report_item = Item::new(ItemType::Nel);
+        report_item.set_payload(ContentType::Json, item.to_owned().to_string());
+        envelope.add_item(report_item);
+        common::handle_envelope(&state, envelope).await?;
```
jjbayer:
```suggestion
        // Every NEL report maps to one error event, and we can only have one error event per envelope.
        common::handle_envelope(&state, envelope).await?;
```

jjbayer: We can can respond faster if we `join!` on `handle_envelope` futures simultaneously instead of awaiting them in sequence.

jjbayer (approved):
Looks good to me, but please take a look at my suggestions before merging!

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,251 @@
...
+    InvalidNel,
+    /// Incoming Json is unparsable.
+    #[error("incoming json is unparsable")]
+    InvalidJson(#[from] serde_json::Error),
+}
+
+/// Generated network error report (NEL).
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct Body {
```
iker-barriocanal: > Adds a new Nel protocol with minimal validation (i.e. only specific fields are captured)

Not capturing the rest of the fields doesn't allow forward compatibility, and will require upgrading external Relays to get them in our infra. Is this ok?

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// The elapsed number of milliseconds between the start of the resource
+    /// fetch and when it was completed or aborted by the user agent.
+    pub elapsed_time: Annotated<u64>,
```
iker-barriocanal: As per [Relay guidelines on documentation](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#987c0416a9f54674999eed59eecbf846), could you add a single line of the description and include the rest of the information in a separate paragraph?

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,251 @@
...
+                            body.phase.as_str().unwrap_or(""),
+                            body.ty.as_str().unwrap_or(""),
+                        )
+                    }
+                }));
+
+                event.request = Annotated::new(nel.get_request());
+                event.logger = Annotated::from("nel".to_string());
+
+                // Exrtact common tags.
```
iker-barriocanal:
```suggestion
                // Extract common tags.
```

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,251 @@
...
+            "url": "http://example.com/",
+            "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36"
+        }"#;
+
+        let mut event = Event::default();
+        NetworkReport::apply_to_event(json.as_bytes(), &mut event).unwrap();
+
+        // mock timestamp because it is actually dynamic and depend on current time and "age" field
+        event.timestamp =
+            Annotated::new(Utc.with_ymd_and_hms(2023, 10, 5, 0, 0, 0).unwrap().into());
```
iker-barriocanal: nit: I think you can add this time to the json above?

iker-barriocanal (approved):
🚀

relay-server/src/utils/rate_limits.rs
```diff
@@ -93,6 +93,7 @@ fn infer_event_category(item: &Item) -> Option<DataCategory> {
         ItemType::Event => Some(DataCategory::Error),
         ItemType::Transaction => Some(DataCategory::Transaction),
         ItemType::Security | ItemType::RawSecurity => Some(DataCategory::Security),
+        ItemType::Nel => None,
```
jan-auer: `From<EventType> for DataCategory` returns `DataCategory::Error`, so we should probably also return the same here. That would mean NEL reports will be considered an error event for all intents and purposes, including rate limiting and billing. Considering that size limiting also counts NEL as events, that would be the most coherent option.

relay-server/src/endpoints/nel.rs
```diff
@@ -0,0 +1,75 @@
...
+    let handlers = items.as_array().into_iter().flatten().map(|item| {
+        let mut envelope = Envelope::from_request(Some(EventId::new()), params.meta.clone());
+        let mut report_item = Item::new(ItemType::Nel);
+        report_item.set_payload(ContentType::Json, item.to_owned().to_string());
+        envelope.add_item(report_item);
+        common::handle_envelope(&state, envelope)
+    });
+
+    // Check the results of the handled envelopes.
+    for handle_result in futures::future::join_all(handlers).await {
```
jan-auer: Since this fans out, could we please apply a hard limit on the length on `handlers` to avoid extremely large arrays?

I wonder if it's necessary to produce these concurrently at all here, since this will issue a _lot_ of redundant requests to the project cache and rate limiter. Temporarily, that would be fine, but we should rather process all of the reports in batch.

relay-server/src/endpoints/nel.rs
```diff
@@ -0,0 +1,75 @@
...
+    state: ServiceState,
+    mime: Mime,
+    params: NelReportParams,
+) -> Result<impl IntoResponse, BadStoreRequest> {
+    if !is_nel_mime(mime) {
+        return Ok(StatusCode::UNSUPPORTED_MEDIA_TYPE.into_response());
+    }
+
+    let items: Value =
+        serde_json::from_slice(&params.body).map_err(BadStoreRequest::InvalidJson)?;
```
jan-auer: This is unfortunate, since it requires rather expensive operations directly in the web handler. Usually, our endpoints will try to not parse (large!) payloads, and instead buffer bytes. Then, all the parsing happens asynchronously in the `EnvelopeProcessor`. 

The expensive operations here are:
- First, we need to parse a `Value`, which needs a lot of allocations
- Then, we stringify it just to add it to an envelope item
- inside `future::join_all`, we collect all items in another large list and potentially await a large number of futures

Would it be possible to instead collect the raw data all into a single item, and then split them in the `EnvelopeProcessor`? If that is not possible at all, consider using [`RawValue`](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) like we do for Replays to parse the individual items without a roundtrip.

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,251 @@
...
+            }
+            Some(_) | None => Annotated::empty(),
+        };
+        Request {
+            url: self.url.clone(),
+            headers,
+            ..Request::default()
+        }
+    }
+    pub fn apply_to_event(data: &[u8], event: &mut Event) -> Result<(), ReportError> {
```
jan-auer: Modelling this after CSP unfortunately means quite a bit of normalization logic is now placed in the `event-schema` crate. Could we move this to `event-normalization`, instead?

relay-event-schema/src/protocol/event.rs
```diff
@@ -448,6 +448,10 @@ pub struct Event {
     #[metastructure(omit_from_schema)] // we only document error events for now
     pub csp: Annotated<Csp>,
 
+    /// NEL (Network Error Logging) reports.
+    #[metastructure(omit_from_schema)] // we only document error events for now
+    pub nel: Annotated<NetworkReport>,
```
jan-auer: What's the rationale of making this a top-level field rather than a context?

Separately: During normalization, many of the fields will be cloned into standard places in the event payload. Do we have to store the full original report here, or could we store just the remaining fields that are NEL-specific? That would also allow us to _move_ the other fields.

relay-event-schema/src/protocol/nel.rs
jan-auer: We should add the usual `other` field here:

```
#[metastructure(additional_properties, pii = "maybe")]
pub other: Object<Value>,
```

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,251 @@
...
+        let mut nel: Annotated<NetworkReport> =
+            Annotated::from_json_bytes(data).map_err(ReportError::InvalidJson)?;
+
+        if let Some(nel) = nel.value_mut() {
+            if let Some(body) = nel.body.value() {
+                event.logentry = Annotated::new(LogEntry::from({
+                    if nel.ty.value().map_or("", |v| v.as_str()) == "http.error" {
```
jan-auer:
```suggestion
                    if nel.ty.as_str() == Some("http.error") {
```

jan-auer (commented):
Preliminary review, excluding the exact normalization logic.

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,251 @@
...
+                    }
+                }));
+
+                event.request = Annotated::new(nel.get_request());
+                event.logger = Annotated::from("nel".to_string());
+
+                // Exrtact common tags.
+                let tags = event.tags.get_or_insert_with(Tags::default);
+                if let Some(ref method) = body.method.value() {
+                    tags.push(Annotated::new(TagEntry::from_pair((
```
jan-auer: Please try to avoid creating tags. Instead, we have two better options:
- Put these fields into a `context`, where they will be automatically indexed by Snuba.
- Explicitly index fields from the NEL structure in case we're keeping it.

Tags are today meant for user-supplied data.

olksdr:
Addressed more feedback and specifically from @jan-auer:
* Created the NEL context
* Added some missing fields to existing contexts to support this new protocol
* Moved big chunk of the code into normalization
* Also tried to make the endpoint more performant and do the minimal deserialization and quickly proceed with processing of the incoming events.

oioki:
A few suggestions:
1. Copy `body.elapsed_time` to "nel" context. It's really valuable field, for example it shows how long it took till the server failed the connection or DNS failed.
2. The `origin` in the `response` context is a bit misleading. It is coming from `body.server_ip` which means the IP address of the server where the request was sent. Can we move this to "nel" context as `server_ip`?
3. Copy `body.referrer` to "request.headers" context.
4. I noticed Relay produces two `type` fields in the "nel" context, neither of which are useful (first is always `network-error`, and second `nel` is coming from us). I guess we can remove both of them since we already store valuable `body.type` in the `logentry.logformatted`:
```
"nel":{
  "age":31042,
  "type":"network-error",
  "type":"nel"
}
```
5. I'm not a big fan of splitting data to several contexts. Can we store the entire `body` of NEL report in the "nel" context, including `phase`, `sampling_fraction`, `type`? In my opinion, this would make it easier for user to understand these values by referring to them exactly as in the NEL spec.

relay-event-normalization/src/nel.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use chrono::{DateTime, Duration, Utc};
+use relay_event_schema::protocol::{
+    Contexts, Event, HeaderName, HeaderValue, Headers, LogEntry, NelContext, NetworkReportRaw,
+    PairList, Request, ResponseContext, Timestamp,
+};
+use relay_protocol::Annotated;
+
+/// Normalizes the event using the provided [`NetworkReportRaw`].
+pub fn normalize(event: &mut Event, nel: Annotated<NetworkReportRaw>) {
```
jjbayer: I would move this function to `relay-server`, just to consistent with other cases in which we copy information from one envelope item to another. See for example: https://github.com/getsentry/relay/blob/4ed7d5fcec0caa3ca5db9f616527dcd612bfd5e4/relay-server/src/utils/unreal.rs#L266

relay-server/src/endpoints/nel.rs
```diff
@@ -45,18 +45,17 @@ async fn handle(
         return Ok(StatusCode::UNSUPPORTED_MEDIA_TYPE.into_response());
     }
 
-    let items: Value =
+    let items: Vec<Box<RawValue>> =
```
jjbayer: Why does `RawValue` need to be boxed here? Could We do a `Vec<&RawValue>` instead?

py/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `scraping_attempts` field to the event schema. ([#2575](https://github.com/getsentry/relay/pull/2575))
```
jjbayer, line 18: thanks

relay-event-normalization/src/nel.rs
```diff
@@ -0,0 +1,77 @@
...
+    };
+    let Some(body) = nel.body.value() else {
+        return;
+    };
+
+    event.logentry = Annotated::new(LogEntry::from({
+        if nel.ty.value().map_or("", |v| v.as_str()) == "http.error" {
+            format!(
+                "{} / {} ({})",
+                body.phase.as_str().unwrap_or(""),
```
jjbayer: Not sure if this happens in the wild, but should we write some placeholder here?
```suggestion
                body.phase.as_str().unwrap_or("<unknown-phase>"),
```

relay-event-normalization/src/nel.rs
```diff
@@ -0,0 +1,77 @@
...
+use relay_event_schema::protocol::{
+    Contexts, Event, HeaderName, HeaderValue, Headers, LogEntry, NelContext, NetworkReportRaw,
+    PairList, Request, ResponseContext, Timestamp,
+};
+use relay_protocol::Annotated;
+
+/// Normalizes the event using the provided [`NetworkReportRaw`].
+pub fn normalize(event: &mut Event, nel: Annotated<NetworkReportRaw>) {
+    // If the incoming NEL report is empty or it contains an empty body, just exit.
+    let Some(nel) = nel.value() else {
```
jjbayer: I see there's a couple of `.clone()` happening below, could we prevent those by calling `into_value()` here?

relay-event-normalization/src/nel.rs
```diff
@@ -0,0 +1,77 @@
...
+    response_context.origin = body.server_ip.clone();
+    response_context.status_code = body
+        .status_code
+        .clone()
+        .map_value(|v| u64::try_from(v).unwrap_or(0));
+
+    // Set the timestamp on the event when it actually occurred.
+    let now: DateTime<Utc> = Utc::now();
+    if let Some(event_time) =
+        now.checked_sub_signed(Duration::milliseconds(*nel.age.value().unwrap_or(&0)))
```
jjbayer: If `nel.age.value()` is `None`, do we really want to set event.timestamp to the current time?

relay-event-schema/src/protocol/contexts/nel.rs
```diff
@@ -0,0 +1,54 @@
...
+/// W3C Editor's Draft: <https://w3c.github.io/network-error-logging/>
+/// MDN: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging>
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct NelContext {
+    /// The age of the report since it got collected and before it got sent.
+    pub age: Annotated<i64>,
+    /// The type of the report.
+    #[metastructure(field = "type")]
+    pub ty: Annotated<String>,
```
jjbayer: nit: If we make `type` an enum with an `other(String)` variant, we can probably shave off a string allocation for most use cases:

> The type of the network error may be one of the following pre-defined values from the specification, but browsers can add and send their own error types

https://developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging#dns.unreachable

relay-server/src/endpoints/common.rs
```diff
@@ -291,14 +308,14 @@ fn queue_envelope(
...
         )?;
-
-        // Update the old context after successful forking.
-        managed_envelope.update();
         state
             .project_cache()
             .send(ValidateEnvelope::new(event_context));
     }
 
+    // Update the old context before continue with the source envelope.
```
jjbayer:
```suggestion
    // Update the old context before continuing with the source envelope.
```

relay-server/src/envelope.rs
```diff
@@ -1080,6 +1080,38 @@ impl Envelope {
...
+        self.items = own_items;
+
+        for item in split_items {
+            envelopes.push(Box::new(Envelope {
+                headers: self.headers.clone(),
+                items: smallvec![item],
+            }))
+        }
+
+        envelopes
```
jjbayer: nit: `split_by` and `split_all_by` could use a common helper method `split_off_items` that returns `split_items`.

relay-server/src/utils/rate_limits.rs
```diff
@@ -93,7 +93,7 @@ fn infer_event_category(item: &Item) -> Option<DataCategory> {
         ItemType::Event => Some(DataCategory::Error),
         ItemType::Transaction => Some(DataCategory::Transaction),
         ItemType::Security | ItemType::RawSecurity => Some(DataCategory::Security),
-        ItemType::Nel => None,
+        ItemType::Nel => Some(DataCategory::Error),
```
jjbayer, line 96: Good catch. Should we add a test that shows that `Nel` reports are removed when there is an error quota?

relay-server/src/endpoints/nel.rs
olksdr: Used `RawValue` here as suggested. But still had to deserialize into `Vec` to get the items out of the payload

relay-server/src/endpoints/nel.rs
olksdr: yeah. you're right, changed it

relay-event-normalization/src/nel.rs
olksdr: makes sense to me, moved this into `utils` in `relay-server`

relay-event-normalization/src/nel.rs
olksdr: yeah, used `into_value()` here

relay-event-normalization/src/nel.rs
olksdr: So, yeah. it would mean that the event happen just right now. Which can really happen

relay-server/src/envelope.rs
olksdr: Will look into it

py/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Add `scraping_attempts` field to the event schema. ([#2575](https://github.com/getsentry/relay/pull/2575))
+- Added support for NEL (Network Error Logging) reports ingestion. ([#2421](https://github.com/getsentry/relay/pull/2421))
```
jjbayer: nit

```suggestion
- Add context for NEL (Network Error Logging) reports to the event schema. ([#2421](https://github.com/getsentry/relay/pull/2421))
```

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// For forward compatibility.
+    #[metastructure(additional_properties, pii = "maybe")]
+    pub other: Object<Value>,
+}
+
+/// Models the content of a NEL report.
+///
+/// NOTE: This is the structure used inside the Event (serialization is based on Annotated
+/// infrastructure). We also use a version of this structure to deserialize from raw JSON
+/// via serde.
```
jjbayer: Is this still true?

relay-server/src/envelope.rs
```diff
@@ -1047,32 +1054,72 @@ impl Envelope {
...
-    /// with all items that return `true`. Items that return `false` remain in this envelope.
+    /// Splits off the items from the envelope using provided predicates.
     ///
-    /// The returned envelope assumes the same headers.
-    pub fn split_by<F>(&mut self, mut f: F) -> Option<Box<Self>>
+    /// First predicate is the the additional condition on the count of found items by second
+    /// predicate.
+    fn split_off_items<C, F>(&mut self, cond: C, mut f: F) -> Option<SmallVec<[Item; 3]>>
     where
+        C: Fn(usize) -> bool,
```
jjbayer, line 1094: Sorry, I missed that the condition for returning `None` is different for the two functions., this makes the helper function less helpful than I thought. Does the condition have to be different though?

jjbayer (approved):
LGTM. Do nel reports respect rate limits for errors now? Can we test that? See https://github.com/getsentry/relay/pull/2421/files#r1360800171.

Also, [this convo](https://github.com/getsentry/relay/pull/2421/files#r1360774789) is still open, but feel free to ignore if you think it's not worth it.

relay-server/src/envelope.rs
olksdr: yeah, the condition are different and must stay this way

relay-event-schema/src/protocol/nel.rs
olksdr: 👍  nice catch, removed .

olksdr:
> LGTM. Do nel reports respect rate limits for errors now? Can we test that? See https://github.com/getsentry/relay/pull/2421/files#r1360800171.
> 
> Also, [this convo](https://github.com/getsentry/relay/pull/2421/files#r1360774789) is still open, but feel free to ignore if you think it's not worth it.

@jjbayer  yeah. I will still take time and look into adding a test.

relay-server/src/utils/mod.rs
```diff
@@ -27,6 +28,7 @@ pub use self::metrics_rate_limits::*;
 pub use self::multipart::*;
 #[cfg(feature = "processing")]
 pub use self::native::*;
+pub use self::nel::*;
```
iker-barriocanal: Should we just export the items we need?

relay-event-schema/src/protocol/contexts/mod.rs
```diff
@@ -19,6 +20,7 @@ pub use cloud_resource::*;
 pub use device::*;
 pub use gpu::*;
 pub use monitor::*;
+pub use nel::*;
```
iker-barriocanal, line 23: Same here wrt star exports.

iker-barriocanal (approved):
Really good stuff 🚀

relay-event-schema/src/protocol/contexts/nel.rs
```diff
@@ -0,0 +1,60 @@
+#[cfg(feature = "jsonschema")]
+use relay_jsonschema_derive::JsonSchema;
+use relay_protocol::{Annotated, Empty, FromValue, IntoValue, Object, Value};
+
+use crate::processor::ProcessValue;
+
+/// Contains the unique to NEL reports information.
+///
+/// NEL is a browser feature that allows reporting of failed network requests from the client side.
```
jan-auer:
```suggestion
/// Contains NEL report information.
///
/// Network Error Logging (NEL) is a browser feature that allows reporting of failed network
/// requests from the client side. See the following resources for more information:
///
/// - [W3C Editor's Draft](https://w3c.github.io/network-error-logging/)
/// - [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging)
```

relay-event-schema/src/protocol/contexts/nel.rs
```diff
@@ -0,0 +1,60 @@
...
+/// W3C Editor's Draft: <https://w3c.github.io/network-error-logging/>
+/// MDN: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging>
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct NelContext {
+    /// If request failed, the type of its network error. If request succeeded, "ok".
+    pub error_type: Annotated<String>,
+    /// Server IP where the requests was sent to.
+    #[metastructure(pii = "true")]
```
jan-auer: - Should we rather mark that as `pii = "maybe"`? If that is the target server IP, it is still sensitive information but does not identify users.
- Could we use the `IpAddr` type here?

relay-event-schema/src/protocol/contexts/nel.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub server_ip: Annotated<String>,
+    /// The time between the start of the resource fetch and when it was completed or aborted.
+    pub elapsed_time: Annotated<u64>,
+    /// If request failed, the phase of its network error. If request succeeded, "application".
+    pub phase: Annotated<String>,
```
jan-auer: Could we introduce an enumeration for the known phases, with an `Other(String)` variant as fallback? According to [the specification](https://w3c.github.io/network-error-logging/#network-errors), the known phases are `dns`, `connection`, and `application`.

relay-event-schema/src/protocol/contexts/nel.rs
```diff
@@ -0,0 +1,60 @@
...
+/// MDN: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging>
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct NelContext {
+    /// If request failed, the type of its network error. If request succeeded, "ok".
+    pub error_type: Annotated<String>,
+    /// Server IP where the requests was sent to.
+    #[metastructure(pii = "true")]
+    pub server_ip: Annotated<String>,
+    /// The time between the start of the resource fetch and when it was completed or aborted.
```
jan-auer: Could you please document, which unit this time is specified in?

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,128 @@
+//! Contains definitions for the Network Error Logging (NEL) interface.
+//!
+//! NEL is a browser feature that allows reporting of failed network requests from the client side.
+//! W3C Editor's Draft: <https://w3c.github.io/network-error-logging/>
+//! MDN: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging>
```
jan-auer: Consider linking to the NEL context type for these docs, that way they don't have to be maintained in two places when we add more information to them.

relay-event-schema/src/protocol/nel.rs
```diff
@@ -0,0 +1,128 @@
...
+use relay_jsonschema_derive::JsonSchema;
+use relay_protocol::{Annotated, Empty, FromValue, IntoValue, Object, Value};
+use thiserror::Error;
+
+/// The NEL parsing errors.
+#[derive(Debug, Error)]
+pub enum NetworkReportError {
+    /// Unexpected format.
+    #[error("unexpected format")]
+    InvalidNel,
```
jan-auer: It looks like the `InvalidNel` error variant is not instantiated anywhere, and the only other variant covers deserialization. Could we simplify this by calling `Annotated::from_json_bytes(value).map_err(ProcessingError::InvalidJson)?` directly in `event_from_nel_item` like the other deserializing functions do?

relay-event-schema/src/protocol/request.rs
```diff
@@ -444,6 +444,9 @@ pub struct Request {
     /// HTTP request method.
     pub method: Annotated<String>,
 
+    /// HTTP protocol.
+    pub protocol: Annotated<String>,
```
jan-auer, line 448: Could we introduce an enum for the well-known protocols? Particularly, it would be nice to normalize the casing of the protocols, so that both `https` and `HTTPS` can be sent in, but they are always stored in the same notation.

relay-server/src/endpoints/nel.rs
```diff
@@ -0,0 +1,69 @@
...
+async fn handle(
+    state: ServiceState,
+    mime: Mime,
+    params: NelReportParams,
+) -> Result<impl IntoResponse, BadStoreRequest> {
+    if !is_nel_mime(mime) {
+        return Ok(StatusCode::UNSUPPORTED_MEDIA_TYPE.into_response());
+    }
+
+    let items: Vec<&RawValue> =
```
jan-auer, line 48: In practice, allocating a vector here should be fine. However, since there's currently no limit placed on the number of NEL reports in an envelope - neither in the endpoint nor where it gets sent into validation (`common.rs`) -- let's please be sure we have enough visibility into this endpoint to add more safety measures in the future.

If this turns out to be a problem, we can easily make this a streaming deserializer.

relay-server/src/utils/nel.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use chrono::{DateTime, Duration, Utc};
+use relay_event_schema::protocol::{
+    Contexts, Event, HeaderName, HeaderValue, Headers, LogEntry, NelContext, NetworkReportRaw,
+    Request, ResponseContext, Timestamp,
+};
+use relay_protocol::Annotated;
+
+/// Enriches the event with new values using the provided [`NetworkReportRaw`].
+pub fn enrich_nel_event(event: &mut Event, nel: Annotated<NetworkReportRaw>) {
```
jan-auer: Could this function be moved into `relay-normalize`? 

Either way works, as there's currently no clear criteria for where normalization is called and where the code is located. Particularly, this is not part of the current "store normalizer" but rather called during deserialization like for security reports.

However, in an effort to reduce the amount of schema-specific business logic that resides in `relay-server`, a better place for these sorts of functions would be the normalize crate.

jan-auer (approved):
Thank you, this looks great and is in shippable state.

See a number of smaller comments attached, and feel free to merge without another review from me.

relay-server/src/utils/nel.rs
```diff
@@ -0,0 +1,87 @@
...
+    let status_code = body
+        .status_code
+        .map_value(|v| u64::try_from(v).unwrap_or(0));
+    if status_code.value().unwrap_or(&0) > &0 {
+        let response_context = contexts.get_or_default::<ResponseContext>();
+        response_context.status_code = status_code;
+    }
+
+    // Set the timestamp on the event when it actually occurred.
+    let now: DateTime<Utc> = Utc::now();
```
jan-auer: Could we use the receive timestamp from the envelope instead of `Utc::now()`? Relay might have delayed processing the item for various reasons, but the receive timestamp is guaranteed to be as accurate as possible.

relay-event-schema/src/protocol/contexts/nel.rs
olksdr: done

relay-event-schema/src/protocol/contexts/nel.rs
olksdr: done

relay-event-schema/src/protocol/contexts/nel.rs
olksdr: done

relay-event-schema/src/protocol/nel.rs
olksdr: done

relay-event-schema/src/protocol/nel.rs
olksdr: done

relay-event-schema/src/protocol/request.rs
olksdr: I could not find the list of the protocols to be used here. The doc is not really clear on this. 
I can postpone this to later PRs

relay-server/src/utils/nel.rs
olksdr: Yeah, I can move it to normalize. It was initially there, but then was moved back to the server utils.

PR #2424, opened 2023-08-24
ref: Split relay-general into smaller crates

jan-auer:
This patch splits the `relay-general` crate into smaller successors with
distinct responsibilities. The smaller crates offer a larger public API
surface of previously internal utilities. This means more documentation
and the ability to write doc tests for normalization and PII scrubbing.

## Background

The `relay-general` crate (historically the successor to
`sentry-marshal`) for a long time has combined several concerns into a
single crate:

- Basic types for building JSON-based schemas, also called "the
  protocol"
- Derives for traits related to the above
- A custom derive to genrate JSON schema from protocol types
- Definitions of the event schema
- A visitor trait and visitor, called "processor" that is both generic
  over protocol types, as well as specific to the event schema
- Event normalization, including UA parsing and GeoIP
- PII scrubbing
- Release health, client reports, and replay schema types

Due to the way the traits and derives were written, new types using the
existing facilities had to be added to the `relay-general` crate. This
creates slow response times and leads to suboptimal code organization
with intermixed concerns.

## Changes

The majority of the diff moves files to new crates, moves code into
other files, or updates imports. However, there are also other notable
changes:

- The derives for `JsonSchema` and `ProcessValue` are now separated into
  their own crates, while the basic protocol traits (`FromValue`,
  `IntoValue`, and `Empty`) remain in a single derive. They still share
  the same `metastructure` annotations, but have separated code bases
  now.
- Derive crates are now all top-level; there are no more nested crates
  in the workspace.
- `#![warn(missing_docs)]` is now enabled in nearly all crates and stubs
  for missing docs have been added.
- Several unit tests had to be moved to crate-level tests, so that they
  are compiled outside of the crate and can make use of our own derives.

Further code-level changes include:

- All `Cargo.toml` files have been reformatted using BetterTOML's
  defaults. Crate dependencies with multiple feature flags now wrap.
- Code formatting with 1.72.0 has been applied, which includes
  formatting support for `let else`.
- JSONSchema validation in `relay-kafka` is now behind a feature flag
  instead of debug assertions so that the `jsonschema` dependency can be
  excluded from production builds.

## Future

- This refactor provides the foundation for implementing fully isolated
  domain logic in separate crates comprising schema, normalization, and
  configuration in a colocated way.
- The multiple glob implementations can be unified into a single glob
  builder and globset matcher that is used for filters, dynamic
  sampling, metric extraction, and the C-ABI.
- Separate a generic protocol visitor that can traverse through any
  protocol struct and visit basic data types from specific visitors for
  schema types. This is currently combined in the `ProcessValue` trait,
  but should be separated.
- Adopt Cargo [`workspace.dependencies`](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table)
  for common crates like `uuid`, `chrono`, and `insta` to avoid
  duplicate declarations or re-exports through utility crates as much as
  possible.

#skip-changelog


r
olksdr, line 1: I thinks this file should be removed 😄

relay/src/main.rs
```diff
@@ -33,28 +33,36 @@
...
 //!  - [`relay-crash`]: Crash reporting for the Relay server.
-//!  - [`relay-monitors`]: Monitors protocol and processing for Sentry.
 //!  - [`relay-dynamic-config`]: Dynamic configuration passed from Sentry.
+//!  - [`relay-event-derive`]: Derive for visitor traits on the Event schema.
+//!  - [`relay-event-normalization`]: Event normalization and processing.
+//!  - [`relay-event-schema`]: Event schema (Error, Transaction, Security) and types for event processing.
 //!  - [`relay-ffi`]: Utilities for error handling in FFI bindings.
 //!  - [`relay-ffi-macros`]: Macros for error handling in FFI bindings.
 //!  - [`relay-filter`]: Inbound data filters.
-//!  - [`relay-general`]: Event protocol, normalization and data scrubbing.
```
olksdr, line 45: The `relay-general` was removed from there, but we still have it. Should this be added back?

relay/src/main.rs
```diff
@@ -69,28 +77,36 @@
...
 //! [`relay-crash`]: ../relay_crash/index.html
-//! [`relay-monitors`]: ../relay_monitors/index.html
 //! [`relay-dynamic-config`]: ../relay_dynamic_config/index.html
+//! [`relay-event-derive`]: ../relay_event_derive/index.html
+//! [`relay-event-normalization`]: ../relay_event_normalization/index.html
+//! [`relay-event-schema`]: ../relay_event_schema/index.html
 //! [`relay-ffi`]: ../relay_ffi/index.html
 //! [`relay-ffi-macros`]: ../relay_ffi_macros/index.html
 //! [`relay-filter`]: ../relay_filter/index.html
-//! [`relay-general`]: ../relay_general/index.html
```
olksdr, line 81: Also here `relay-general` is removed

olksdr (approved):

relay/src/main.rs
jan-auer: Where do you see it? The was split up and has been removed entirely.

olksdr: yeah, you're right!  Sorry for the confusion. The Github UI made me think it's still there

relay-event-normalization/src/schema.rs
```diff
@@ -119,56 +119,57 @@ fn verify_value_characters(
...
-    fn test_nonempty_array() {
-        assert_nonempty_base::<Array<u64>>("a non-empty value");
-    }
+    use super::*;
 
-    #[test]
-    fn test_nonempty_object() {
-        assert_nonempty_base::<Object<u64>>("a non-empty value");
-    }
+    // TODO(ja): Enable this test
```
jan-auer, line 132: This test is disabled since it needs to derive `ProcessValue`, but it also needs access to the `SchemaProcessor`. This is currently not possible and requires a larger refactor of the trait and the derive, which would be too much changes for this already large PR.

relay-event-normalization/src/trimming.rs
```diff
@@ -719,46 +716,47 @@ mod tests {
...
-
-        // Ensure stack does not leak with newtypes
-        assert!(processor.bag_size_state.is_empty());
-
-        assert_eq!(
-            value.to_json().unwrap(),
-            r#"{"inner":{"inner":{"inner":"hi"}}}"#
-        );
-    }
+    // TODO(ja): Enable this test
```
jan-auer, line 719: Same as with the other disabled test.

mitsuhiko:
image

viglia (approved):

relay-common/src/lib.rs
```diff
@@ -8,20 +8,12 @@
...
 mod macros;
 
-mod constants;
-mod glob;
-mod metrics;
-mod project;
-mod time;
+pub mod glob;
+pub mod glob2;
+pub mod glob3;
```
jan-auer, line 13: All glob implementations have moved into `relay-common` now and can be deduplicated in a followup.

relay-common/src/glob.rs
```diff
@@ -41,105 +37,6 @@ fn translate_pattern(pat: &str, options: GlobOptions) -> Option<Regex> {
         .ok()
 }
 
-fn translate_codeowners_pattern(pattern: &str) -> Option<Regex> {
```
jan-auer, line 44: This does not belong into `relay-common` as a shared utility crate and now resides in `relay-cabi` directly, since it was added to Relay with the primary purpose of exposing it to Sentry.

relay-event-normalization/src/normalize/contexts.rs
```diff
@@ -37,6 +38,46 @@ static OS_UNAME_REGEX: Lazy<Regex> = Lazy::new(|| {
 static RUNTIME_DOTNET_REGEX: Lazy<Regex> =
     Lazy::new(|| Regex::new(r"^(?P<name>.*) (?P<version>\d+\.\d+(\.\d+){0,2}).*$").unwrap());
 
+/// A hashmap that translates from the android model to the more human-friendly product-names.
+/// E.g. NE2211 -> OnePlus 10 Pro 5G
+static ANDROID_MODEL_NAMES: Lazy<HashMap<&'static str, &'static str>> = Lazy::new(|| {
```
jan-auer, line 43: This declaration has moved from the schema to normalization.

relay-pii/src/snapshots/relay_pii__convert__tests__sanitize_http_body.snap
```diff
@@ -7,7 +7,6 @@ expression: data.value().unwrap().request
     "email": "zzzz@gmail.com",
     "password": "[Filtered]"
   },
-  "inferred_content_type": "application/json",
```
jan-auer, line 10: This test no longer runs normalization before testing PII scrubbing, hence the `inferred_content_type` property is missing from the output.

relay-event-schema/src/processor/funcs.rs
```diff
@@ -0,0 +1,54 @@
+use relay_protocol::{Annotated, IntoValue, Meta};
+
+use crate::processor::{
+    ProcessValue, ProcessingAction, ProcessingResult, ProcessingState, Processor,
+};
+
+/// Modifies this value based on the action returned by `f`.
+#[inline]
+pub fn apply<T, F, R>(v: &mut Annotated<T>, f: F) -> ProcessingResult
```
jan-auer, line 9: This used to be a method on `Annotated`. Since the signature requires types from processors that aren't available to `relay-protocol`, the method had to be moved to a utility function in the schema.

Originally, `ProcessingAction` was supposed to be a simple enum declaring whether or not a value should be removed as part of applying a transform function on it. At some point, the `InvalidTransaction` variant got added to it, probably mistaking this enum for an error. That variant has since stuck, and more logic has built on that new model. It's now too much effort to change this back.

relay-event-schema/src/processor/impls.rs
```diff
@@ -254,3 +258,50 @@ process_tuple!(T1, T2, T3, T4, T5, T6, T7, T8, T9);
 process_tuple!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10);
 process_tuple!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11);
 process_tuple!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12);
+
+impl ProcessValue for Value {
```
jan-auer, line 262: This was previously derived and is now implemented manually. The manual implementation is slightly more performant as it avoids cloning the processing state when traversing into the branches of the `Value` type.

relay-event-schema/src/protocol/constants.rs
```diff
@@ -19,3 +20,6 @@ pub const VALID_PLATFORMS: &[&str] = &[
     "python",
     "ruby",
 ];
+
+/// The latest version of the protocol.
+pub const PROTOCOL_VERSION: u16 = 7;
```
jan-auer, line 25: Vendored from `sentry-types`.

relay-common/src/lib.rs
jjbayer: See issue https://github.com/getsentry/relay/issues/2426.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -782,26 +786,97 @@ fn normalize_user_geoinfo(geoip_lookup: &GeoIpLookup, user: &mut User) {
...
+    ///
+    /// This is similar to `transaction_name_config`, but applies to span descriptions.
     pub span_description_rules: Option<&'a Vec<SpanDescriptionRule>>,
+
+    /// An initialized GeoIP lookup.
     pub geoip_lookup: Option<&'a GeoIpLookup>,
+
+    /// When `Some(true)`, individual parts of the event payload is trimmed to a maximum size.
+    ///
+    /// See the event schema for size declarations.
```
jjbayer, line 879: Thanks for writing all these docs!

PR #2425, opened 2023-08-24
ref(statsd): Use statsdproxy to pre-aggregate metrics in-memory

untitaker:
use experimental statsdproxy hackweek project to aggregate counters and
gauges (i.e. the "easy stuff") in memory before sending it over the UDP
buffer.

We use the same code in rust consumers to pre-aggregate metrics. The
performance improvement is a wash (neither improves nor degrades perf),
but it should load on veneur, so it may still amount to cost savings.

Arpad has a kind of pre-aggregation that results in actual cost savings
within the application itself, in the future we may replace statsdproxy
with that.


relay-config/src/config.rs
```diff
@@ -497,6 +497,10 @@ struct Metrics {
     ///
     /// Defaults to `true`.
     buffering: bool,
+    /// Emitted metrics will be aggregated to optimize bandwidth.
+    ///
+    /// Defaults to `false`.
```
Dav1dde: Why do we not enable it by default?

relay-statsd/src/lib.rs
```diff
@@ -247,7 +250,24 @@ pub fn init<A: ToSocketAddrs>(
     let socket = UdpSocket::bind("0.0.0.0:0").unwrap();
     socket.set_nonblocking(true).unwrap();
 
-    let statsd_client = if buffering {
+    let statsd_client = if aggregating {
+        let host = host.to_socket_addrs().unwrap().next().unwrap();
+        let statsdproxy_sink = StatsdProxyMetricSink::new(move || {
```
Dav1dde, line 255: How does this sink behave compared to our buffered config, I assume the aggregator also immediately accepts, aggregates in the background and then uses a buffered UDP sink? Depending on how the aggregator flushes a buffered UDP sink could still be beneficial.

relay-config/src/config.rs
Dav1dde: Also should probably document that it takes precedence over `buffering`.

untitaker: I think we can. Originally this PR was set out to be a bit more conservative in how this would be rolled out, but now that it's live in Rust consumers we can also just enable it unconditionally. In that case, would we want to keep the other flags (`buffering` etc) too, or just remove all of those different configs?

relay-statsd/src/lib.rs
untitaker: it does both aggregation and buffering. the buffering (in `Upstream`) is not optional or configurable

relay-config/src/config.rs
Dav1dde: I think we should just roll it out with enabled by default (since we only expect upsides from this and are confident it works) and use canary to monitor it a bit.

If everything looks good, go to SaaS but keep the option for a while so we can go back if something looks fishy.

Later merge the aggregation option with the buffering option.

What do you think?

relay-statsd/src/lib.rs
Dav1dde: The `Upstream` itself is not buffered though, right?

https://github.com/getsentry/statsdproxy/blob/0b6d207423ad979b571b3807837142ce3cebaeb8/src/middleware/upstream.rs#L22-L35

relay-config/src/config.rs
untitaker: yeah sounds good, for now i made the minimal change and just flipped the default. if all goes well i think both options (aggregation _and_ buffering) should go away, to be honest. i don't see the value in having that be configurable

untitaker:
@Dav1dde some additional context that i forgot to share: we use statsdproxy in rust consumers to send data to DDM.

the way this works is that we pre-aggregate using statsdproxy, then multiplex to the rust SDK, in order to offset the performance overhead that the rust SDK has.

I think long-term statsdproxy is not the right abstraction for this, and in fact @Swatinem is already working on what I think could be a replacement for all of this. but in the short-term this would allow you to dogfood DDM in relay with minimal overhead (and no code locations). take a look at `statsd.rs` in snuba if you're interested.

Dav1dde:
> the way this works is that we pre-aggregate using statsdproxy, then multiplex to the rust SDK, in order to offset the performance overhead that the rust SDK has.

Thanks, that seems like a good approach and something we wanted to do anyways.

Dav1dde (approved):
Let's try it.

We need to test this properly on Canary and S4S first, please don't merge if you don't have enough time to do that.

If you want I can pick this up and do the rollout sometime beginning of next week.

relay/src/setup.rs
```diff
@@ -69,6 +69,7 @@ pub fn init_metrics(config: &Config) -> Result<()> {
         &addrs[..],
         default_tags,
         config.metrics_buffering(),
+        config.metrics_aggregation(),
```
jjbayer, line 72: nit: With this number of arguments, it might be nice to pass a `StatsdConfig` object instead. Not a blocker though.

relay-statsd/src/lib.rs
```diff
@@ -247,7 +250,24 @@ pub fn init<A: ToSocketAddrs>(
...
+            let next_step = statsdproxy::middleware::upstream::Upstream::new(host)
+                .expect("failed to create statsdproxy metric sink");
+            statsdproxy::middleware::aggregate::AggregateMetrics::new(
+                AggregateMetricsConfig {
+                    aggregate_gauges: true,
+                    aggregate_counters: true,
+                    flush_interval: 1,
+                    flush_offset: 0,
+                    max_map_size: None,
```
jjbayer, line 264: Should these be configurable?

jjbayer (approved):

relay/src/setup.rs
Dav1dde: Plan is to get rid of the options all together: https://github.com/getsentry/relay/pull/2425#discussion_r1499565181

What do you think?

jjbayer: OK!

relay-statsd/src/lib.rs
untitaker: they should probably not have been options to begin with tbh

PR #2433, opened 2023-08-28
fix(spans): Strip more table prefixes

jjbayer:
Strip prefixes of table names even if they are not quotes, e.g.

```
SELECT a.b FROM a
-- becomes
SELECT b FROM a
```

ref: [internal issue](https://www.notion.so/sentry/Fix-normalization-issues-discovered-in-5-rollout-65335d3ac4ab4eabb886faa318edad3b?pvs=4)

#skip-changelog


gggritso (approved):
Makes sense!

TBS1996 (approved):

relay-event-normalization/src/normalize/span/description/sql.rs
```diff
@@ -383,6 +392,13 @@ mod tests {
         "SELECT .. FROM table WHERE %s"
     );
 
+    scrub_sql_test!(
+        span_description_collapse_columns_with_as_without_quotes,
+        // Simple lists of columns will be collapsed.
+        r#"SELECT myfield1, a.b AS a__b, another_field as bar FROM table WHERE %s"#,
+        "SELECT .. FROM table WHERE %s"
```
iker-barriocanal, line 399: Is there a test validating the use case in the PR description? How does stripping work on table names with prefixes _and_ `AS`, where column names aren't collapsed?

jjbayer: Good catch, the example in the PR description is actually not covered by this PR. Updated the description now. We only remove ` AS ..` if there's a list of columns.

PR #2434, opened 2023-08-28
fix(replays): Cap segment_id size to u16

cmanallen:
`Annotated<u16>` is not possible so I'm manually checking if the u64 type is within the bounds of a u16.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/replay.rs
+++ b/relay-event-normalization/src/replay.rs
@@ -337,7 +337,7 @@ mod tests {
         // Does not fit within a u16.
         let replay_id =
             Annotated::new(EventId("52df9022835246eeb317dbd739ccd059".parse().unwrap()));
-        let segment_id: Annotated<u64> = Annotated::new(65536);
+        let segment_id: Annotated<u64> = Annotated::new(u16::MAX + 1 as u64);
         let mut replay = Annotated::new(Replay {
             replay_id,
             segment_id,
--- a/relay-event-normalization/src/replay.rs
+++ b/relay-event-normalization/src/replay.rs
@@ -337,7 +337,7 @@ mod tests {
         // Does not fit within a u16.
         let replay_id =
             Annotated::new(EventId("52df9022835246eeb317dbd739ccd059".parse().unwrap()));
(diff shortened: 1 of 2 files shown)


relay-event-normalization/src/replay.rs
```diff
@@ -44,6 +44,17 @@ pub fn validate(replay: &mut Replay) -> Result<(), ReplayError> {
         .value()
         .ok_or_else(|| ReplayError::InvalidPayload("missing replay_id".to_string()))?;
 
+    let segment_id = replay
+        .segment_id
+        .value()
+        .ok_or_else(|| ReplayError::InvalidPayload("missing segment_id".to_string()))?;
+
+    if segment_id > &65535 {
```
jjbayer:
```suggestion
    let segment_id = *replay
        .segment_id
        .value()
        .ok_or_else(|| ReplayError::InvalidPayload("missing segment_id".to_string()))?;

    if segment_id > u16::MAX as u64 {
```

jjbayer (approved):
@jan-auer IMO it would be more elegant to implement `FromValue` etc. for `u16`, but I'm not sure whether we want to keep the list of base types clean (i.e. `f64`, `i64`, `u64`).

relay-event-normalization/src/replay.rs
```diff
@@ -321,6 +332,31 @@ mod tests {
         assert!(replay_value.urls.value().unwrap().len() == 100);
     }
 
+    #[test]
+    fn test_validate_u16_segment_id() {
+        // Does not fit within a u16.
+        let replay_id =
+            Annotated::new(EventId("52df9022835246eeb317dbd739ccd059".parse().unwrap()));
+        let segment_id: Annotated<u64> = Annotated::new(65536);
```
iker-barriocanal: nit (not sure if syntax is correct):
```suggestion
        let segment_id: Annotated<u64> = Annotated::new(u16::MAX + 1 as u64);
```
and `u16::MAXX` in the number below too.

iker-barriocanal (approved):

PR #2437, opened 2023-08-29
ref(protocol): Move `FieldValueProvider` to a new `Getter` trait

jan-auer:
The `FieldValueProvider` trait was originally introduced to abstract data access
for dynamic sampling rules. Since then, the same rules are being used for
conditional metric tagging and metric extraction.

Since accessing fields in data instances by path is a more generic concern, the
trait is now moved to `relay_protocol` and renamed to `Getter`. It retains its
single `get_value` method, but will be extended by more functionality for arrays
and objects in the near future.

To simplify the implementation of the trait, the `get_value` method now returns
an option. This option is `None` if either the path does not match any known
field, or if the field is an empty option.

To avoid redundant allocations, `get_value` no longer returns an owned
`serde_json::Value`. Instead, there is a new `Val` type: The borrowed variant of
`Value`. For now, the only data actually borrowed are strings, however, there
are stubs to support arrays and objects.

#skip-changelog

### DIFF
--- a/relay-protocol/src/value.rs
+++ b/relay-protocol/src/value.rs
@@ -1,6 +1,7 @@
 use std::collections::BTreeMap;
 use std::{fmt, str};
 
+use relay_common::uuid::Uuid;
 #[cfg(feature = "jsonschema")]
 use schemars::{gen::SchemaGenerator, schema::Schema, JsonSchema};
 use serde::de::{Deserialize, MapAccess, SeqAccess, Visitor};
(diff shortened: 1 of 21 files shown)


relay-protocol/src/value.rs
```diff
@@ -334,3 +335,172 @@ where
...
+/// Borrowed version of [`Array`].
+#[derive(Debug, Clone, Copy)]
+pub struct Arr<'a> {
+    _phantom: std::marker::PhantomData<&'a ()>,
+}
+
+/// Borrowed version of [`Object`].
+#[derive(Debug, Clone, Copy)]
+pub struct Obj<'a> {
+    _phantom: std::marker::PhantomData<&'a ()>,
```
jjbayer, line 348: This phantom causes a change in behavior that I'm not 100% sure about. E.g.  for a span `{"data": {"foo": {"bar": 1}}}`, `span.data.foo` would give you an `Object { "bar": Number(1) }`, now it returns this opaque type. Not sure if this has any practical consequences.

We could instead return `None` for object values, to make it clear that getters for objects and arrays are not yet implemented.

jan-auer: We couldn't extract such a value before anyway, since the only data types we could work with were the primitive ones. You can see evidence for that because now you cannot access array or object types anymore, but no code broke for this. 

For metric extraction from transactions, we will have to implement support for arrays, but that wasn't there before either.

jjbayer (approved):

PR #2438, opened 2023-08-29
ref(sampling): Split the sampling crate into submodules

jan-auer:
Refactor that splits the relay-sampling crate into several submodules. The
entire crate was implemented and tested in a single file, which exceeded a
reasonable size.

As part of this refactor, missing documentation is added to types that were
previously undocumented.

Most of the tests still reside in the top-level library source file, since they
test e2e sampling functionality. If the API surface of the crate is changed,
these tests can also be moved to a better location.

#skip-changelog


relay-sampling/src/utils.rs
```diff
@@ -0,0 +1,4 @@
+/// Returns `true` if this value is equal to `Default::default()`.
+pub fn is_default<T: Default + PartialEq>(t: &T) -> bool {
```
iker-barriocanal, line 2: nit: could we benefit from inlining this method?

relay-sampling/src/lib.rs
```diff
@@ -1,51 +1,57 @@
...
-//! data that will be used in `relay_server` to perform the sampling decision.
 //!
-//! # Sampling Determinism
-//! The concept of determinism is extremely important for sampling. We want to be able to make the
-//! a deterministic sampling decision for a wide variety of reasons, including:
-//! - Across a **chain of Relays** (e.g., we don't want to drop an event that was retained by a previous
-//! Relay and vice-versa).
-//! - Across **transactions of the same trace** (e.g., we want to be able to sample all the transactions
-//! of the same trace, even though some exceptions apply).
+//! - The incoming data is received by Relay.
```
iker-barriocanal, line 32: nit: should this be an ordered list?

relay-sampling/src/lib.rs
```diff
@@ -1,51 +1,57 @@
...
+//! - Relay resolves the [`SamplingConfig`] of the project to which the incoming data belongs.
+//!   Additionally, it tries to resolve the configuration of the project that started the trace, the
+//!   so-called "root project".
+//! - The two [`SamplingConfig`]s are merged together and dynamic sampling evaluates a result. The
+//!   algorithm goes over each rule and compute either a factor or sample rate based on the
+//!   value of the rule.
```
iker-barriocanal, line 38: Mentioning sampling config for the first, configuration for the second, and then referring to the sampling config for both is a bit confusing. What do you think of the following?
```suggestion
//! - Relay resolves the [`SamplingConfig`] of both the project to which the incoming data belongs
//!   and the project that started the trace, the so-called "root project".
//! - The two [`SamplingConfig`]s are merged together and dynamic sampling evaluates a result. The
//!   algorithm goes over each rule and compute either a factor or sample rate based on the
//!   value of the rule.
```

relay-sampling/src/lib.rs
```diff
@@ -1,51 +1,57 @@
...
-//! for the random number generator (e.g., all transactions with the same trace id will have the same
-//! random number being generated). _Since we allow the matching of both transaction and trace rules, we might
-//! end up in cases in which we perform inconsistent trace sampling but this is something we decided
-//! to live with as long as there are no big implications on the product._
+//! - Run dynamic sampling repeatedly over a **chain of Relays** (e.g., we don't want to drop an
+//! event that was retained by a previous Relay and vice-versa).
+//! - Run dynamic sampling independently on **transactions of the same trace** (e.g., we want to be
+//! able to sample all the transactions of the same trace, even though some exceptions apply).
+//!
+//! In order to perform deterministic sampling, Relay uses teh trace ID or event ID as the seed for
```
iker-barriocanal, line 53:
```suggestion
//! In order to perform deterministic sampling, Relay uses the trace ID or event ID as the seed for
```

jjbayer (approved):

relay-sampling/src/condition.rs
```diff
@@ -0,0 +1,414 @@
...
+/// more expensive to construct and check, so preferrably use [`EqCondition`] when no wildcard
+/// matching is needed.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct GlobCondition {
+    /// Path of the field that should match the value.
+    pub name: String,
+    /// A list of glob patterns to check.
+    ///
+    /// Note that this cannot be a single value, it must be a list of values.
+    pub value: GlobPatterns,
```
iker-barriocanal, line 130: Shouldn't single value / list of values be enforced in `GlobPatterns`?

relay-sampling/src/condition.rs
```diff
@@ -0,0 +1,414 @@
...
+
+impl RuleCondition {
+    /// Returns a condition that matches everything.
+    pub fn all() -> Self {
+        Self::And(AndCondition { inner: Vec::new() })
+    }
+
+    /// Checks if Relay supports this condition (in other words if the condition had any unknown configuration
+    /// which was serialized as "Unsupported" (because the configuration is either faulty or was created for a
+    /// newer relay that supports some other condition types)
```
iker-barriocanal, line 245:
```suggestion
    /// Checks if Relay supports this condition. In other words, if the condition had any unknown configuration
    /// which was serialized as "Unsupported", because the configuration is either faulty or was created for a
    /// newer relay that supports some other condition types.
```
? (I find two nested parenthesis hard to follow)

relay-sampling/src/dsc.rs
```diff
@@ -0,0 +1,683 @@
...
+    #[serde(default)]
+    pub environment: Option<String>,
+    /// The name of the transaction extracted from the `transaction` field in the starting
+    /// transaction.
+    ///
+    /// Set on transaction start, or via `scope.transaction`.
+    #[serde(default)]
+    pub transaction: Option<String>,
+    /// The sample rate with which this trace was sampled in the client. This is a float between
+    /// `0.0` and `1.0`.
```
iker-barriocanal, line 45: Both inclusive?

relay-sampling/src/dsc.rs
```diff
@@ -0,0 +1,683 @@
...
+    #[serde(
+        default,
+        with = "sample_rate_as_string",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub sample_rate: Option<f64>,
+    /// The user specific identifier (e.g. a user segment, or similar created by the SDK from the
+    /// user object).
```
iker-barriocanal, line 53:
```suggestion
    /// The user-specific identifier (e.g. a user segment, or similar created by the SDK from the
    /// user object).
```

relay-sampling/src/config.rs
```diff
@@ -0,0 +1,455 @@
...
+
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+
+use crate::condition::RuleCondition;
+use crate::utils;
+
+/// Represents the dynamic sampling configuration available to a project.
+///
+/// Note: This comes from the organization data
```
iker-barriocanal, line 13:
```suggestion
/// Note: this comes from the organization data.
```

relay-sampling/src/config.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+}
+
+/// A range of time.
+///
+/// The time range should be applicable between the start time, inclusive, and
+/// end time, exclusive. There aren't any explicit checks to ensure the end
+/// time is equal to or greater than the start time; the time range isn't valid
+/// in such cases.
+#[derive(Debug, Clone, Copy, Default, Serialize, Deserialize)]
```
iker-barriocanal, line 149: Is this specific to the config module? What do you think about moving it to utils?

PR #2439, opened 2023-08-29
chore(self-hosted): Dogfood self-hosted e2e test action

hubertdeng123:


.github/workflows/ci.yml
```diff
@@ -465,7 +465,7 @@ jobs:
         # Skip for dependabot or if it's a fork as the image cannot be uploaded to ghcr since this test attempts to pull
         # the image from ghcr
         if: "!github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'"
-        uses: getsentry/action-self-hosted-e2e-tests@f45ef07793b2cc805a9a9401819f486da449a90a
+        uses: getsentry/action-self-hosted-e2e-tests@03010bd2963edc1f47b6e5e03167a4bc1433ea36
```
jjbayer, line 468: @hubertdeng123 as far as I can see, this will remove the docker-compose logs from the action output. Don't we want the logs to be visible in both the sentry error and the github action?

jjbayer (changes requested):

.github/workflows/ci.yml
hubertdeng123: We've found that docker compose logs were not very useful to us. Anyhow, we actually decided that we'll be removing the self-hosted e2e test suite since it's not a required job and it doesn't seem to be providing much value in other repos beyond `self-hosted`

PR #2440, opened 2023-08-30
fix(dynamic-config): Expose metric extraction to all Relays

jan-auer:
Metric extraction configuration was not added to `LimitedProjectConfig`. For
this reason, external Relays would not see specs for on-demand metric
extraction. All traffic ingested through Relay PoPs or external Relays therefore
was missing on-demand metrics.


jjbayer (approved):
Wondering how we could make it easier to catch bugs like this. Maybe we should invert the structs into `ProjectConfig` (for the limited config) and `FullProjectConfig`, where `ProjectConfig` is `serde(flatten)`ed into `FullProjectConfig`?

iker-barriocanal (approved):

PR #2444, opened 2023-08-30
feat(spans): Detect multiple SQL table names

jjbayer:
Previously, we extracted the `span.domain` for SQL spans by looking for the first table name we could find, e.g. `SELECT FROM x` or `UPDATE x = ...`. 

With this change the domain becomes a sorted, comma-separated list of table names, for example for

```sql
SELECT * FROM table1 INNER JOIN table2 ON table2_id = table2.id
```

the span domain becomes `table1,table2`.

This allows us to search for all queries that involve a certain table.

**NOTE:** To reliably find all table names involved in a query, this PR introduces a dependency on `sqlparser`. When parsing the SQL query **fails** (for example because the input query is already truncated) we **fall back** to the previous behavior of extracting a a single table name via regex.

ref: [internal issue](https://www.notion.so/sentry/Concatenate-all-table-names-found-in-a-query-together-21872abfe2664dccb00eeffd4b37e1a4?pvs=4)

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/tag_extraction.rs
+++ b/relay-event-normalization/src/normalize/span/tag_extraction.rs
@@ -275,7 +275,7 @@ pub(crate) fn extract_tags(span: &Span, config: &Config) -> BTreeMap<SpanTagKey,
         } else if span_op.starts_with("db") {
             span.description
                 .value()
-                .and_then(|query| sql_table_from_query(system, query))
+                .and_then(|query| sql_tables_from_query(system, query))
         } else {
             None
         };
@@ -359,10 +359,8 @@ static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
 static SQL_PLACEHOLDER_REGEX: Lazy<Regex> =
     Lazy::new(|| Regex::new(r"(?:\?+|\$\d+|%(?:\(\w+\))?s|:\w+)").unwrap());
 
-/// Returns the table in the SQL query, if any.
-///
-/// If multiple tables exist, only the first one is returned.
-fn sql_table_from_query(db_system: Option<&str>, query: &str) -> Option<String> {
+/// Returns a sorted, comma-separated list of SQL tables, if any.
+fn sql_tables_from_query(db_system: Option<&str>, query: &str) -> Option<String> {
     use sqlparser::ast::{ObjectName, Visit, Visitor};
     use sqlparser::dialect::GenericDialect;
     use std::ops::ControlFlow;
@@ -604,7 +602,7 @@ mod tests {
     fn extract_table_select() {
         let query = r#"SELECT * FROM "a.b" WHERE "x" = 1"#;
         assert_eq!(
-            sql_table_from_query(Some("postgresql"), query).unwrap(),
+            sql_tables_from_query(Some("postgresql"), query).unwrap(),
             "b"
         );
     }
@@ -612,14 +610,14 @@ mod tests {
     #[test]
     fn extract_table_select_nested() {
         let query = r#"SELECT * FROM (SELECT * FROM "a.b") s WHERE "x" = 1"#;
-        assert_eq!(sql_table_from_query(None, query).unwrap(), "b");
+        assert_eq!(sql_tables_from_query(None, query).unwrap(), "b");
     }
 
     #[test]
     fn extract_table_multiple() {
         let query = r#"SELECT * FROM a JOIN t.c ON c_id = c.id JOIN b ON b_id = b.id"#;
         assert_eq!(
-            sql_table_from_query(Some("postgresql"), query).unwrap(),
+            sql_tables_from_query(Some("postgresql"), query).unwrap(),
             "a,b,c"
         );
     }
@@ -628,7 +626,10 @@ mod tests {
     fn extract_table_multiple_mysql() {
         let query =
             r#"SELECT * FROM a JOIN `t.c` ON /* hello */ c_id = c.id JOIN b ON b_id = b.id"#;
-        assert_eq!(sql_table_from_query(Some("mysql"), query).unwrap(), "a,b,c");
+        assert_eq!(
+            sql_tables_from_query(Some("mysql"), query).unwrap(),
+            "a,b,c"
+        );
     }
 
     #[test]
@@ -657,7 +658,7 @@ WHERE (
 LIMIT 1
             "#;
         assert_eq!(
-            sql_table_from_query(Some("postgresql"), query).unwrap(),
+            sql_tables_from_query(Some("postgresql"), query).unwrap(),
             "sentry_environmentrelease,sentry_grouprelease,sentry_release_project"
         );
     }
@@ -665,14 +666,14 @@ LIMIT 1
     #[test]
     fn extract_table_delete() {
         let query = r#"DELETE FROM "a.b" WHERE "x" = 1"#;
-        assert_eq!(sql_table_from_query(None, query).unwrap(), "b");
+        assert_eq!(sql_tables_from_query(None, query).unwrap(), "b");
     }
 
     #[test]
     fn extract_table_insert() {
         let query = r#"INSERT INTO "a" ("x", "y") VALUES (%s, %s)"#;
         assert_eq!(
-            sql_table_from_query(Some("postgresql"), query).unwrap(),
+            sql_tables_from_query(Some("postgresql"), query).unwrap(),
             "a"
         );
     }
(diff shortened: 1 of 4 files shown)


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -177,6 +178,16 @@ fn extract_shared_tags(event: &Event) -> BTreeMap<SpanTagKey, Value> {
...
     let mut span_tags: BTreeMap<SpanTagKey, Value> = BTreeMap::new();
+
+    let system = span
+        .data
+        .value()
+        .and_then(|v| v.get("db.system"))
+        .and_then(|system| system.as_str());
+    if let Some(sys) = system {
+        span_tags.insert(SpanTagKey::System, sys.to_lowercase().into());
+    }
```
jjbayer, line 192: This block was moved up from line 294.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -264,8 +275,7 @@ pub(crate) fn extract_tags(span: &Span, config: &Config) -> BTreeMap<SpanTagKey,
         } else if span_op.starts_with("db") {
             span.description
                 .value()
-                .and_then(|query| sql_table_from_query(query))
-                .map(|t| t.to_lowercase())
```
jjbayer, line 268: `to_lowercase` happens inside the function now, to prevent an additional string allocation.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -354,11 +355,60 @@ static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
...
-///
-/// If multiple tables exist, only the first one is returned.
-fn sql_table_from_query(query: &str) -> Option<&str> {
-    extract_captured_substring(query, &SQL_TABLE_EXTRACTOR_REGEX)
+/// Used to replace placeholders (parameters) by dummy values such that the query can be parsed.
+static SQL_PLACEHOLDER_REGEX: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"(?:\?+|\$\d+|%(?:\(\w+\))?s|:\w+)").unwrap());
+
+/// Returns a sorted, comma-separated list of SQL tables, if any.
+fn sql_tables_from_query(db_system: Option<&str>, query: &str) -> Option<String> {
```
iker-barriocanal, line 366: Is there any reason to add the structs, imports, etc inside the function?

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -354,11 +355,60 @@ static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
...
+    match sqlparser::parser::Parser::parse_sql(&*dialect, &parsable_query) {
+        Ok(ast) => {
+            let mut visitor = RelationsVisitor {
+                relations: Default::default(),
+            };
+            ast.visit(&mut visitor);
+            let mut s = String::with_capacity(visitor.relations.iter().map(String::len).sum());
+            for (i, name) in visitor.relations.into_iter().enumerate() {
+                if i == 0 {
+                    s = name;
```
iker-barriocanal: Doesn't `s = name` drop the buffer you allocate on line 397?

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -354,11 +355,60 @@ static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
...
+        }
+    }
+
+    // See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#notes-and-well-known-identifiers-for-dbsystem
+    //     https://docs.rs/sqlparser/latest/sqlparser/dialect/fn.dialect_from_str.html
+    let dialect = db_system
+        .and_then(sqlparser::dialect::dialect_from_str)
+        .unwrap_or_else(|| Box::new(GenericDialect {}));
+
+    let parsable_query = SQL_PLACEHOLDER_REGEX.replace_all(query, "1");
```
iker-barriocanal, line 374: Q: is the reason you replace placeholders with `1`s to remove tokens before you traverse the AST?

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -354,11 +355,60 @@ static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
...
+    match sqlparser::parser::Parser::parse_sql(&*dialect, &parsable_query) {
+        Ok(ast) => {
+            let mut visitor = RelationsVisitor {
+                relations: Default::default(),
+            };
+            ast.visit(&mut visitor);
```
iker-barriocanal, line 381: Does the order of the tables in the output depend on the implementation details of `visit`? Should we sort the table names before adding them to a string? Should we even care about it?

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -354,11 +355,60 @@ static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
...
+                } else {
+                    write!(&mut s, ",{name}").ok();
+                }
+            }
+            (!s.is_empty()).then_some(s)
+        }
+        Err(e) => {
+            relay_log::debug!("Failed to parse SQL: {e}");
+            extract_captured_substring(query, &SQL_TABLE_EXTRACTOR_REGEX).map(str::to_lowercase)
+        }
```
iker-barriocanal, line 395: Do you think adding this limitation to the PR description can be helpful in the future?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -354,11 +355,60 @@ static SQL_TABLE_EXTRACTOR_REGEX: Lazy<Regex> = Lazy::new(|| {
...
+        type Break = ();
+
+        fn pre_visit_relation(&mut self, relation: &ObjectName) -> ControlFlow<Self::Break> {
+            if let Some(name) = relation.0.last() {
+                let last = name.value.split('.').last().unwrap_or(&name.value);
+                self.relations.insert(last.to_lowercase());
+            }
+            ControlFlow::Continue(())
+        }
+    }
```
TBS1996: not that im against it, but im a bit curious on this pattern. I didn't see before having structs and impl blocks inside of a function, is this idiomatic ?

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: I wanted the Visitor to be a private type of the function, since it is the only one that uses it. But I agree that it's not great for readability, so I'll move them out of the function.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: The language allows it, and I like how it keeps the types private. But I'll move them out (see comment above).

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: sqlparser cannot handle placeholders like `%s`. Added a comment now.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: The order is maintained by the `BTreeSet` (added a comment to clarify). We definitely care because having two different tag values containing the same tables produces needless cardinality.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: Good catch!

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: Yes, done!

gggritso (approved):
👍🏻 LGTM!

relay-event-normalization/src/normalize/span/tag_extraction.rs
iker-barriocanal: Oh I missed that! Good stuff.

relay-event-normalization/src/normalize/span/tag_extraction.rs
iker-barriocanal: If visibility is the only concern, what about encapsulating it in a module?

iker-barriocanal (approved):
🚀

wmak (approved):

PR #2445, opened 2023-08-30
fix(globalconfig): Shutdown the service on Relay shutdown

iker-barriocanal:
When Relay gets a `SIGTERM` and network access is disabled for production pods, the global config service continues to request configs from upstream until Relay gets a `SIGINT` or `SIGQUIT`. As a result, the service can't make network requests and emits error metrics, which can be confused with actual errors.

This PR introduces the shutdown control in the service, shutting it down with Relay and avoiding unnecessary network requests and error metrics. Once the global config service has shut down updates from its subscription will fail, so the processor has been updated accordingly to continue processing the remaining envelopes and avoid data loss.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Filter out exceptions originating in Safari extensions. ([#2408](https://github.com/getsentry/relay/pull/2408))
 - Fixes the `TraceContext.status` not being defaulted to `unknown` before the new metrics extraction pipeline. ([#2436](https://github.com/getsentry/relay/pull/2436))
 - Support on-demand metrics for alerts and widgets in external Relays. ([#2440](https://github.com/getsentry/relay/pull/2440))
+- Shut the global config service down when Relay is shutting down. ([#2445](https://github.com/getsentry/relay/pull/2445))
 
 ## 23.8.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 - Fixes the `TraceContext.status` not being defaulted to `unknown` before the new metrics extraction pipeline. ([#2436](https://github.com/getsentry/relay/pull/2436))
 - Support on-demand metrics for alerts and widgets in external Relays. ([#2440](https://github.com/getsentry/relay/pull/2440))
 - Prevent sporadic data loss in `EnvelopeProcessorService`. ([#2454](https://github.com/getsentry/relay/pull/2454))
-- Shut the global config service down when Relay is shutting down. ([#2445](https://github.com/getsentry/relay/pull/2445))
 
 **Internal**:
 
--- a/relay-server/Cargo.toml
+++ b/relay-server/Cargo.toml
@@ -98,7 +98,12 @@ symbolic-unreal = { version = "12.1.2", optional = true, default-features = fals
     "serde",
 ] }
 thiserror = "1.0.38"
-tokio = { version = "1.28.0", features = ["rt-multi-thread", "sync", "macros"] }
+tokio = { version = "1.28.0", features = [
+    "rt-multi-thread",
+    "sync",
+    "macros",
+    "test-util",
+] }
 tower = { version = "0.4.13", default-features = false }
 tower-http = { version = "0.4.0", default-features = false, features = [
     "catch-panic",
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -85,9 +85,6 @@ impl UpstreamQuery for GetGlobalConfig {
 pub struct Get;
 
 /// The message for receiving a watch that subscribes to the [`GlobalConfigService`].
-///
-/// On Relay shutdown, the sender of the channel is dropped and receiving new
-/// updates errors.
 pub struct Subscribe;
 
 /// An interface to get [`GlobalConfig`]s through [`GlobalConfigService`].
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -289,7 +289,7 @@ mod tests {
 
     /// Tests whether the service shuts down gracefully when the signal is
     /// received, without requesting more global configs.
-    #[tokio::test]
+    #[tokio::test(start_paused = true)]
     async fn test_service_shutdown() {
         for mode in &[ShutdownMode::Graceful, ShutdownMode::Immediate] {
             shutdown_service_with_mode(*mode).await;
@@ -306,7 +306,7 @@ mod tests {
         assert!(service.send(Get).await.is_ok());
 
         Controller::shutdown(mode);
-        tokio::time::sleep(Duration::from_secs(1)).await;
+        tokio::time::advance(Duration::from_millis(200)).await;
 
         assert!(service.send(Get).await.is_err());
     }
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -246,7 +246,7 @@ impl Service for GlobalConfigService {
 
     fn spawn_handler(mut self, mut rx: relay_system::Receiver<Self::Interface>) {
         tokio::spawn(async move {
-            let mut shutdown = Controller::shutdown_handle();
+            let mut shutdown_handle = Controller::shutdown_handle();
 
             relay_log::info!("global config service starting");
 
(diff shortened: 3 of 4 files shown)


relay-server/src/actors/global_config.rs
```diff
@@ -271,3 +274,37 @@ impl Service for GlobalConfigService {
...
+        let (upstream, _) = mock_service("upstream", (), |&mut (), _| {});
+
+        Controller::start(Duration::from_secs(1));
+
+        let service = GlobalConfigService::new(Arc::default(), upstream).start();
+
+        assert!(service.send(Get).await.is_ok());
+
+        Controller::shutdown(mode);
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
iker-barriocanal: I did not manage to get around tokio to avoid an actual sleep and make this test pass. Open to suggestions!

relay-server/src/actors/processor.rs
```diff
@@ -2795,14 +2795,20 @@ impl Service for EnvelopeProcessorService {
...
+
             loop {
                 let next_msg = async { tokio::join!(rx.recv(), semaphore.clone().acquire_owned()) };
 
                 tokio::select! {
                    biased;
 
-                    // TODO(iker): deal with the error when the sender of the channel is dropped.
-                    _ = subscription.changed() => self.global_config = subscription.borrow().clone(),
+                    update = subscription.changed(), if !shutdown => {
```
iker-barriocanal: From the [docs](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.changed):

> This method returns an error if and only if the [`Sender`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html) is dropped.

relay-server/src/actors/processor.rs
```diff
@@ -2795,14 +2795,20 @@ impl Service for EnvelopeProcessorService {
...
+                        match update {
+                            Ok(()) => self.global_config = subscription.borrow().clone(),
+                            Err(_) => shutdown = true,
+                        };
```
jjbayer: Are the changes to this file actually worth it? What's the damage without this change? One additional clone of the global config at the end of the lifecycle?

relay-server/src/actors/global_config.rs
jjbayer: Have you tried `tokio::pause` and `tokio::advance`?

relay-server/src/actors/processor.rs
iker-barriocanal: When the global config service gets the shutdown signal it immediately shuts down and drops all the memory it owns, including the sender of the channel. When the sender is dropped, reading a new subscription always errors, and that results in the processor not handling new messages and thus dropping data (not enough time to process in-flight envelopes). Shutting the processor down on the shutdown signal before consuming the remaining messages has the same effect.

This is reproducible with `tests/integration/test_basic.py::test_graceful_shutdown` -- waiting for 2 seconds after sending the shutdown signal is enough to process the data, but not waiting for it results in data loss.

I thought about adding a comment about this, but I didn't want to write too much and assumed the clarification on the `Subscription` message in the global config service was good enough. Let me know if you feel otherwise and I'll add it.

jjbayer: Interesting! I had to try it out myself to understand it, so a code comment would definitely help!

iker-barriocanal: Added in https://github.com/getsentry/relay/pull/2445/commits/6a1d346a079573594a99f8d16e2b3989e09d1c19.

TBS1996 (approved):
seems pretty clever, nice job thinking through this!

relay-server/src/actors/processor.rs
iker-barriocanal: Simplified in [d2cf59a](https://github.com/getsentry/relay/pull/2445/commits/d2cf59a763797fa0d63a38bd68afe8c57c6f3aea).

relay-server/Cargo.toml
```diff
@@ -98,7 +98,12 @@ symbolic-unreal = { version = "12.1.2", optional = true, default-features = fals
     "serde",
 ] }
 thiserror = "1.0.38"
-tokio = { version = "1.28.0", features = ["rt-multi-thread", "sync", "macros"] }
+tokio = { version = "1.28.0", features = [
+    "rt-multi-thread",
+    "sync",
+    "macros",
+    "test-util",
```
jan-auer: Please move this to dev dependencies; the `test-util` flag should not be part of dependencies. You can add it to dev dependencies like so:

https://github.com/getsentry/relay/blob/5bccc34926dc5e8af11aeaf6e2004a4af18c5dae/relay-system/Cargo.toml#L21

Also, please rebase or merge master, since dependencies have changed.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Fixes the `TraceContext.status` not being defaulted to `unknown` before the new metrics extraction pipeline. ([#2436](https://github.com/getsentry/relay/pull/2436))
 - Support on-demand metrics for alerts and widgets in external Relays. ([#2440](https://github.com/getsentry/relay/pull/2440))
 - Prevent sporadic data loss in `EnvelopeProcessorService`. ([#2454](https://github.com/getsentry/relay/pull/2454))
+- Shut the global config service down when Relay is shutting down. ([#2445](https://github.com/getsentry/relay/pull/2445))
```
jan-auer: What is the user-observable bug that's being fixed?

relay-server/src/actors/global_config.rs
```diff
@@ -263,6 +268,7 @@ impl Service for GlobalConfigService {
                     () = &mut self.fetch_handle => self.update_global_config(),
                     Some(result) = self.internal_rx.recv() => self.handle_result(result),
                     Some(message) = rx.recv() => self.handle_message(message),
+                    _ = shutdown.notified() => break,
```
jan-auer: Putting this here means two things:
1. The service will continue to process queued message even after the shutdown signal has been dispatched
2. New messages querying for the global config during the shutdown sequence will fail as soon as the global config service has processed the signal.

I believe both of these are wrong behavior. Shutdown should stop the refresh cycle, which means it should reset the `fetch_handle` and if necessary ensure no new fetch is scheduled. However, the service needs to continue accepting new messages so that other services can wrap up their work.

This also means the test below asserts the wrong behavior.

relay-server/src/actors/global_config.rs
```diff
@@ -271,3 +277,37 @@ impl Service for GlobalConfigService {
...
+        let (upstream, _) = mock_service("upstream", (), |&mut (), _| {});
+
+        Controller::start(Duration::from_secs(1));
+
+        let service = GlobalConfigService::new(Arc::default(), upstream).start();
+
+        assert!(service.send(Get).await.is_ok());
+
+        Controller::shutdown(mode);
+        tokio::time::advance(Duration::from_millis(200)).await;
```
jan-auer: Call `tokio::time::sleep` here to simulate more accurate behavior. As [per docs](https://docs.rs/tokio/latest/tokio/time/fn.advance.html):

> this method will not wait for the sleep calls it advanced past to complete. If you want to do that, you should instead call [sleep](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html) and rely on the runtime’s auto-advance feature.

jan-auer (changes requested):

relay-server/src/actors/global_config.rs
```diff
@@ -271,3 +279,37 @@ impl Service for GlobalConfigService {
...
+    use relay_test::mock_service;
+    use std::sync::Arc;
+    use std::time::Duration;
+
+    use crate::actors::global_config::{Get, GlobalConfigService};
+
+    /// Tests that the service can still handle requests after sending a
+    /// shutdown signal.
+    #[tokio::test]
+    async fn test_service_shutdown() {
```
iker-barriocanal: This test does not ensure the service doesn't make upstream requests.
I'm not sure if it's worth the time and effort to test this -- the global
config service doesn't support generics for `upstream` and the service
mocking is rather naive.

relay-server/src/actors/global_config.rs
```diff
@@ -256,13 +261,16 @@ impl Service for GlobalConfigService {
                 relay_log::info!("fetching global configs disabled: no credentials configured");
             }
 
+            let mut shutdown = false;
```
iker-barriocanal: I don't really like this variable and I'd rather change the internal state (e.g. resetting the fetch handle) and let branches not resolve. However, this approach would still continue processing internal messages*, which I've tried to avoid with this solution.

*I've not found any way to drop all the remaining messages in the channel in constant time, and blocking on that seems a waste of service availability (closing the channel doesn't drop the messages).

relay-server/src/actors/processor.rs
```diff
@@ -2806,8 +2806,8 @@ impl Service for EnvelopeProcessorService {
 
                 tokio::select! {
                    biased;
-                    // TODO(iker): deal with the error when the sender of the channel is dropped.
-                    _ = subscription.changed() => self.global_config = subscription.borrow().clone(),
+
+                    Ok(()) = subscription.changed() => self.global_config = subscription.borrow().clone(),
```
iker-barriocanal, line 2810: I'm still leaving this `Ok(())` to ensure future regressions to the global config service don't result in envelope loss.

jan-auer (approved):

Issue #2334, opened 2023-07-19
[Crons API] Support check-ins with upsert (using POST)


evanpurkhiser:
This will be a follow up to #2333 

For compatibility with the existing API endpoints, we'll also need to support the `monitor_config` field of a check in.

Hopefully this is as simple as passing through the JSON from the endpoint.

gaprl:
Let's make sure we update docs to reflect this new method.

Issue #2333, opened 2023-07-19
[Crons API] Support check-ins using post with a JSON body


evanpurkhiser:
Currently the monitor check-ins endpoints allow you to specify the status, environment, etc via the query parameters.

This is good, but for compatibility with our existing endpoints, we would also like to support `POST` with a JSON body that mimics the check-in

We can probably have a second envelope extractor for `GET` vs `POST`

PR #2448, opened 2023-08-31
feat(monitors): Support sending monitor check-ins as JSON payload

evanpurkhiser:
Implements https://github.com/getsentry/relay/issues/2334 + https://github.com/getsentry/relay/issues/2333

---

Thank you very much @jan-auer for pairing with me on this!

### DIFF
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -129,6 +129,7 @@ pub struct CheckIn {
     pub check_in_id: Uuid,
 
     /// Identifier of the monitor for this check-in.
+    #[serde(default)]
     pub monitor_slug: String,
 
     /// Status of this check-in. Defaults to `"unknown"`.
--- a/tests/integration/test_monitors.py
+++ b/tests/integration/test_monitors.py
@@ -1,4 +1,5 @@
 import base64
+import json
 
 
 def generate_check_in(slug):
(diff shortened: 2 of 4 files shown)


relay-monitors/src/lib.rs
```diff
@@ -129,6 +129,7 @@ pub struct CheckIn {
     pub check_in_id: Uuid,
 
     /// Identifier of the monitor for this check-in.
+    #[serde(default)]
     pub monitor_slug: String,
```
evanpurkhiser, line 133: Slug is now no longer required since it is not required when using the `POST` API endpoint.

Theoretically you could already send the empty string as @jan-auer pointed out, so we probably don't need to handle an error condition for this missing

tests/integration/test_monitors.py
```diff
@@ -10,6 +11,49 @@ def generate_check_in(slug):
     }
 
 
+def test_monitor_query_params(mini_sentry, relay):
```
evanpurkhiser: This isn't actually using a query string, this is ingesting the monitor as an envelope, this can probably be split into a separate PR

jan-auer (approved):
Good to merge once a changelog is added.

tests/integration/test_monitors.py
evanpurkhiser: #2479

PR #2452, opened 2023-09-01
feat(protocol): Improve docs for get_value! and get_path!

### LINKED ISSUE #2365: Improve facilities to read data from Annotated
For tests, we have introduced a `get_value!` macro that travereses a path of `Annotated` structs and returns the (optional) typed value. This is not used consistently throughout the code base, but the macro can probably also be improved.

We should look into streamlining this and employ one way for this throughout the codebase.

More recently, efforts like #2296 also tried to add higher-level accessors for commonly used data. Many of those convenience methods are still useful, though some could be removed.

jan-auer:
Adds doc comments to the `get_path!` macro, which provides typed recursive
access to annotated data structures, and the `get_value!` macro, which provides
typed recursive access to values within annotated data structures.

Alongside this change, this also adds missing feature flagging for test files,
which did not compile with default features.

#skip-changelog
Closes https://github.com/getsentry/relay/issues/2365

---

get_value in relay_protocol - Rust

### DIFF
--- a/relay-protocol/src/macros.rs
+++ b/relay-protocol/src/macros.rs
@@ -1,4 +1,55 @@
-/// Returns `&Annotated<T>` for the annotated value at the given path.
+/// Returns a reference to the typed `Annotated` value at a given path.
+///
+/// The return type depends on the path expression used. By default, this macro will resolve to an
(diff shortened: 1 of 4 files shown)


jjbayer (approved):
Nice!

PR #2453, opened 2023-09-01
feat(schema): Increase the maximum length of event tag keys to 200

jan-auer:
Transaction event tags are being used to store open telemetry attribute names,
which can easily exceed the current limit of 32 characters. There is no
technical reason for that limit, so we can accommodate for longer attribute
names by increasing the tag value length.

See https://github.com/getsentry/sentry/pull/55608

### DIFF
--- a/relay-event-schema/src/processor/attrs.rs
+++ b/relay-event-schema/src/processor/attrs.rs
@@ -123,14 +123,17 @@ impl MaxChars {
             MaxChars::Symbol => 256,
             MaxChars::Path => 256,
             MaxChars::ShortPath => 128,
-            // these are from constants.py or limits imposed by the database
-            MaxChars::Logger => 64,
-            MaxChars::Email => 75,
-            MaxChars::Culprit => 200,
-            MaxChars::TagKey => 32,
-            MaxChars::TagValue => 200,
-            MaxChars::Environment => 64,
-            MaxChars::Distribution => 64,
+            // These limits below were initially derived from Sentry's original implementation in
+            // Python. Before changing any of these limits, check if the values are inserted into
+            // the database by the Sentry processing pipeline, and if there are limits on the
+            // database column.
+            MaxChars::Logger => 64,       // src/sentry/constants.py:???
+            MaxChars::Email => 75,        // src/sentry/constants.py:MAX_EMAIL_FIELD_LENGTH
+            MaxChars::Culprit => 200,     // src/sentry/constants.py:MAX_CULPRIT_LENGTH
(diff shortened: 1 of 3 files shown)


relay-event-schema/src/processor/attrs.rs
```diff
@@ -123,14 +123,17 @@ impl MaxChars {
...
-            // these are from constants.py or limits imposed by the database
-            MaxChars::Logger => 64,
-            MaxChars::Email => 75,
-            MaxChars::Culprit => 200,
-            MaxChars::TagKey => 32,
-            MaxChars::TagValue => 200,
-            MaxChars::Environment => 64,
-            MaxChars::Distribution => 64,
+            // These limits below were initially derived from Sentry's original implementation in
+            // Python. Before changing any of these limits, check if the values are inserted into
```
iker-barriocanal, line 127: > These limits below were initially derived from Sentry's original implementation in Python.

With the increase in tag keys, this no longer holds. I'd remove this sentence and leave the rest, but no strong opinions.

relay-event-schema/src/processor/attrs.rs
```diff
@@ -123,14 +123,17 @@ impl MaxChars {
...
+            // the database by the Sentry processing pipeline, and if there are limits on the
+            // database column.
+            MaxChars::Logger => 64,       // src/sentry/constants.py:???
+            MaxChars::Email => 75,        // src/sentry/constants.py:MAX_EMAIL_FIELD_LENGTH
+            MaxChars::Culprit => 200,     // src/sentry/constants.py:MAX_CULPRIT_LENGTH
+            MaxChars::TagKey => 200,      // src/sentry/constants.py:MAX_TAG_KEY_LENGTH
```
iker-barriocanal, line 133: ~Hmmm I'd expect to have a failing test with this change?~ Wrote too soon 😄

iker-barriocanal (approved):

relay-event-schema/src/processor/attrs.rs
jan-auer: There was, fixed with [ea62f02](https://github.com/getsentry/relay/pull/2453/commits/ea62f02e96b5557f2fbb827da253af7d08120ede)

relay-event-schema/src/processor/attrs.rs
jan-auer: I can try to clarify the wording here, but the sentence should hold. These limits _were originally_ derived from constants, and they are evolving. The intent here is to clarify on:
- The origins of the limit are outside of Relay's codebase and history
- There are technical reasons for some of the limits
- Any change needs to be carefully reviewed.

PR #2457, opened 2023-09-03
fix(profiles): Bump android-trace-log version to prevent panic

### LINKED ISSUE #2451: panic in android_trace_log
> invalid or out-of-range datetime

Panic occurs here: https://github.com/getsentry/relay/blob/63390b6da1e03bf437e4da0e7d903bef98e2d268/relay-profiling/src/android.rs#L115

Presumably because of call to panicking function here: https://github.com/chayleaf/android-trace-log/blob/master/src/lib.rs#L849

See [RELAY-2NNP](https://sentry.my.sentry.io/organizations/sentry/issues/399461)

jjbayer:
Fixes https://github.com/getsentry/relay/issues/2451.

See also https://github.com/chayleaf/android-trace-log/pull/2.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ debug = true
 
 [workspace.dependencies]
 anyhow = "1.0.66"
-chrono = { version = "0.4.24", default-features = false, features = [
+chrono = { version = "0.4.28", default-features = false, features = [
     "std",
     "serde",
 ] }
--- a/relay-common/src/time.rs
+++ b/relay-common/src/time.rs
@@ -130,7 +130,8 @@ impl UnixTimestamp {
 
     /// Returns the timestamp as chrono datetime.
     pub fn as_datetime(self) -> Option<DateTime<Utc>> {
-        NaiveDateTime::from_timestamp_opt(self.0 as i64, 0).map(|n| DateTime::from_utc(n, Utc))
+        NaiveDateTime::from_timestamp_opt(self.0 as i64, 0)
+            .map(|n| DateTime::from_naive_utc_and_offset(n, Utc))
     }
 
     /// Converts the UNIX timestamp into an `Instant` based on the current system timestamp.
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -15,7 +15,7 @@ debug = true
 
 [workspace.dependencies]
 anyhow = "1.0.66"
-chrono = { version = "0.4.24", default-features = false, features = [
+chrono = { version = "0.4.28", default-features = false, features = [
```
jjbayer, line 18: `android_trace_log` uses a chrono function that is not available in 0.4.24.

relay-common/src/time.rs
```diff
@@ -130,7 +130,8 @@ impl UnixTimestamp {
 
     /// Returns the timestamp as chrono datetime.
     pub fn as_datetime(self) -> Option<DateTime<Utc>> {
-        NaiveDateTime::from_timestamp_opt(self.0 as i64, 0).map(|n| DateTime::from_utc(n, Utc))
+        NaiveDateTime::from_timestamp_opt(self.0 as i64, 0)
+            .map(|n| DateTime::from_naive_utc_and_offset(n, Utc))
```
jjbayer, line 134: `from_utc` is [deprecated](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_utc) since chrono 0.4.27. The function does the same [under the hood](https://docs.rs/chrono/0.4.28/src/chrono/datetime/mod.rs.html#123-136).

iker-barriocanal (approved):
Should we add a test?

TBS1996 (approved):
@iker-barriocanal  he wrote a test in `android_trace_log` if that counts 😛 https://github.com/chayleaf/android-trace-log/pull/2/files

jjbayer:
> he wrote a test in `android_trace_log` if that counts 😛 https://github.com/chayleaf/android-trace-log/pull/2/files

@iker-barriocanal the test would replicate the one that @TBS1996 linked, so it would not add value IMO.

PR #2458, opened 2023-09-04
feat(server): Use global config from file if provided

TBS1996:
Provides a way to load global config statically from a file.

* If the relay mode is in managed mode, this file will be ignored as we want in that case to fetch from upstream.
* If it's not in managed mode, it will attempt the load the file, if it exists it will serve it, if it doesn't exist, it will serve the default config. If it exists but we fail to open/parse it, an error will be logged and we will serve the default. 

The location of the static config will be in the same folder as `config.yml` and `credentials.json`.


relay-config/src/config.rs
```diff
@@ -1250,6 +1252,7 @@ impl ConfigObject for ConfigValues {
 pub struct Config {
     values: ConfigValues,
     credentials: Option<Credentials>,
+    global_config: Option<Arc<GlobalConfig>>,
```
TBS1996: wasn't 100% sure if i should put it here or not. My reasoning was that if we fail to parse the file, it's better to fail early than later, and it just seems idiomatic within the codebase that we don't do these kind of risky IO-stuff in the services.

jjbayer: We have a precedent where we defer the deserialization of a static config file to a service: https://github.com/getsentry/relay/blob/498401a4795a7bca5974efd0c7053b345cf0d029/relay-config/src/config.rs#L1916-L1919

That service reloads the config every 10 seconds though, which I think is overkill. In short, I think your way of doing it is better because it fails early. I'll leave final review to @jan-auer.

relay-server/src/actors/global_config.rs
```diff
@@ -146,8 +147,39 @@ pub struct GlobalConfigService {
...
+                relay_log::info!("global config service starting with fetching disabled: using static global config");
+                watch::channel(global_config.clone())
+            }
+            (RelayMode::Capture, _, None) => {
+                relay_log::info!(
+                    "global config service starting with fetching disabled: using default config"
+                );
+                watch::channel(Arc::new(GlobalConfig::default()))
+            }
+        };
```
TBS1996: looks a bit unwieldy but im not sure theres a better way given all the variables that affect the outcome. 

To any non-ikers reading this: the logic is from this specification: https://github.com/getsentry/team-ingest/issues/183#issuecomment-1705405809  
 
I moved this logic from spawn_handler because I didn't want to initialize the global_config_watch with default if we are supposed to use static file, although I suppose it might not matter.

relay-server/src/actors/global_config.rs
```diff
@@ -182,12 +214,10 @@ impl GlobalConfigService {
...
-    fn update_global_config(&mut self) {
-        self.fetch_handle.reset();
-
-        let upstream_relay = self.upstream.clone();
-        let internal_tx = self.internal_tx.clone();
-
+    fn request_global_config(
+        upstream_relay: Addr<UpstreamRelay>,
+        internal_tx: mpsc::Sender<UpstreamQueryResult>,
+    ) {
```
TBS1996: imo, this name makes more sense, as the previous name implies that this function is the one that updates the global config.  
 
I changed it to an associated function so that I could call it from `fn new` easily.  The fetch_handle.reset() is done in spawn_handler now.

relay-server/src/actors/global_config.rs
```diff
@@ -148,8 +149,39 @@ pub struct GlobalConfigService {
...
+        let (global_config_watch, _) = match (
+            config.relay_mode(),
+            config.has_credentials(),
+            config.global_config(),
+        ) {
+            (RelayMode::Proxy | RelayMode::Static, true, None) | (RelayMode::Managed, true, _) => {
+                relay_log::info!("global config service starting");
+                // This request will trigger the request intervals when internal_rx receives the
+                // result from upstream.
+                Self::request_global_config(upstream.clone(), internal_tx.clone());
```
iker-barriocanal: We should not make requests during initialization, as this can block relay startup. We should only do requests once the service is actually started, see my comment below in the `spawn_handler`.

relay-server/src/actors/global_config.rs
```diff
@@ -253,25 +283,15 @@ impl Service for GlobalConfigService {
         tokio::spawn(async move {
             let mut shutdown_handle = Controller::shutdown_handle();
 
-            relay_log::info!("global config service starting");
-
-            if self.config.has_credentials() {
```
iker-barriocanal, line 261: I'd suggest we deal with the relay mode here, and maybe simplify a bit:

- If the mode is managed, run the code below. This requests from upstream and ignores the file.
- If the mode is proxy or static, we send one update to the watch channel with the config we read from the file and never request from upstream (see `handle_shutdown` function above).

iker-barriocanal (changes requested):

relay-server/src/actors/global_config.rs
TBS1996: Even though the request is non-blocking ? 

also, is it fine to then initialize the watch a default global config even if we don't want to use that?  
 
or we should initialize with an option<arc<config>>, where it will be none in the time between initialization and the calling of the spawn_handler

relay-server/src/actors/global_config.rs
TBS1996: I simplified the match statement (with the assumption that we never fetch from upstream in statix/proxy mode), and moved it to spawn handler. 
 
I kept the match statement to ensure we are explicit on all combinations of relay-mode, credentials, and the static file.

relay-server/src/actors/global_config.rs
```diff
@@ -253,25 +251,55 @@ impl Service for GlobalConfigService {
...
-                    Some(result) = self.internal_rx.recv() => self.handle_result(result),
+                    () = &mut self.fetch_handle => {
+                        Self::request_global_config(self.upstream.clone(), self.internal_tx.clone());
+                        // Disable new requests interval until we receive the result from upstream.
+                        self.fetch_handle.reset();
+                    }
+                    Some(result) = self.internal_rx.recv() => {
+                        self.handle_upstream_result(result);
+                        // Start timer for making new global config request.
+                        self.schedule_fetch();
```
TBS1996: I moved the timer reset and scheduling away from the functions to make it more explicit, I think it seems a lot more clear here, lmk if any disagreements

relay-server/src/actors/global_config.rs
```diff
@@ -253,25 +251,55 @@ impl Service for GlobalConfigService {
...
+                    RelayMode::Proxy | RelayMode::Static | RelayMode::Capture,
+                    _,
+                    Some(global_config),
+                ) => {
+                    relay_log::info!("serving static global config loaded from file");
+                    self.global_config_watch.send(global_config.clone()).ok();
+                }
+                (RelayMode::Proxy | RelayMode::Static | RelayMode::Capture, _, None) => {
+                    relay_log::info!(
+                        "serving default global configs due to lacking static global config file"
```
TBS1996: not a big fan of the implicit nature here, Id prefer we create the watch here, but then we'd have to pass it into `handle_message` and `handle_upstream_result`.

relay-server/src/actors/global_config.rs
```diff
@@ -148,8 +149,9 @@ pub struct GlobalConfigService {
 impl GlobalConfigService {
     /// Creates a new [`GlobalConfigService`].
     pub fn new(config: Arc<Config>, upstream: Addr<UpstreamRelay>) -> Self {
-        let (global_config_watch, _) = watch::channel(Arc::default());
         let (internal_tx, internal_rx) = mpsc::channel(1);
+        let (global_config_watch, _) = watch::channel(Arc::new(GlobalConfig::default()));
```
iker-barriocanal: `Arc::default` already handles the default, see [docs](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Default-for-Arc%3CT%3E):

> Creates a new Arc<T>, with the Default value for T.

relay-server/src/actors/global_config.rs
```diff
@@ -253,25 +251,57 @@ impl Service for GlobalConfigService {
...
                     biased;
+                    () = &mut self.fetch_handle => {
+                        // Disable upstream requests timer until we receive result of query.
+                        self.fetch_handle.reset();
+
+                        self.request_global_config();
+                    }
+                    Some(result) = self.internal_rx.recv() => {
+                        // Enable upstream requests timer for global configs.
+                        self.schedule_fetch();
```
iker-barriocanal: What's the motivation to move the `reset` and `schedule_fetch` from the functions to this `select!`? I see no added benefit and find it harder to read. I'd suggest not modifying this code unless it's really needed.

Also, the new schedule should happen after handling the result from upstream.

relay-server/src/actors/global_config.rs
iker-barriocanal: The number of combinations of these 3 is high and I think having a `match` statement cover all of them is hard to understand. What do you think about something like the following?

```rust
match mode {
    Proxy | Static | Capture => {
        if let Some(from_file) = config.global_config() {
            watch.send(from_file);
        }
        // else, the default is already there.
        // Alternatively, we can use `unwrap_or_default` and always send it to the channel.
    }
    Managed => {
        if config.has_credentials() {
            update_global_config();
        } else {
            relay_log();
        }
    }
}
```

Note that the sleep handle is initialized to `idle` (indefinitely pending) and it's not scheduled until a response from upstream is received. This means that not making an initial fetch from upstream results in never doing another fetch.

iker-barriocanal (changes requested):

relay-server/src/actors/global_config.rs
TBS1996: Personally I find it easier to read, as it seems more like a side-effect that you don't expect when reading from the caller. However, if it's indeed harder to read for other people then I'll revert it

relay-server/src/actors/global_config.rs
TBS1996: ok it seems good actually, I changed it to that. 

> Note that the sleep handle is initialized to idle (indefinitely pending) and it's not scheduled until a response from upstream is received. This means that not making an initial fetch from upstream results in never doing another fetch. 
  
I did make an intial fetch when i call the `request_global_config` method

relay-config/src/config.rs
jan-auer: I concur with this: We can load the global configs directly from within the actual service like we do it for project configs. That keeps the responsibility of the `relay-config` crate lower. The overall implementation in this PR does look good, so it would just have to be moved.

jan-auer (commented):
_nit_: This is more of an internal change, in case you'd like to move the changelog entry.

relay-config/src/config.rs
TBS1996: alright, I moved it now to the spawn handler, where I'll log an error if the file exists yet it fails to load it.

relay-server/src/actors/global_config.rs
TBS1996: I've reverted it now

relay-server/src/actors/global_config.rs
```diff
@@ -253,25 +257,47 @@ impl Service for GlobalConfigService {
         tokio::spawn(async move {
             let mut shutdown_handle = Controller::shutdown_handle();
 
-            relay_log::info!("global config service starting");
+            let global_config = match GlobalConfig::load(self.config.path()) {
```
iker-barriocanal: We only need this config in non-managed modes, so let's move this block of code to the section below in line 284.

iker-barriocanal (changes requested):
The PR overall looks good! Can we add a test to verify the following?
- A proxy relay forwards the global config in the config. To load the config, we can probably assume the file is read correctly and do something [like this](https://github.com/getsentry/relay/blob/56a0ba7462242efa03182263c35bebeba0f7b87c/relay-server/src/actors/spooler/mod.rs#L983).
- A proxy relay does not make upstream requests (for guidance, see [this commit](https://github.com/getsentry/relay/commit/631989a6061c8f78e6ac246eba469f0e8662de14#diff-9e66add7db47c3f18146ebef0427032761cc7c185870ff3e603e73441e585b2fR304-R310)).

relay-server/src/actors/global_config.rs
TBS1996: hm I did use it to notify the user that if they had the file in their config folder that it'd be ignored, but I suppose that's not really important, i moved the code

TBS1996:
>The PR overall looks good! Can we add a test to verify the following?

>A proxy relay forwards the global config in the config. To load the config, we can probably assume the file is read correctly and do something [like this](https://github.com/getsentry/relay/blob/56a0ba7462242efa03182263c35bebeba0f7b87c/relay-server/src/actors/spooler/mod.rs#L983).
>A proxy relay does not make upstream requests (for guidance, see [this commit](https://github.com/getsentry/relay/commit/631989a6061c8f78e6ac246eba469f0e8662de14#diff-9e66add7db47c3f18146ebef0427032761cc7c185870ff3e603e73441e585b2fR304-R310)).  
 
Trying to make tests checking if upstream is requested, current problem is that the upstream response fails and logs an error which automatically fails my test, Im not interested in the result of the upstream, do you know the best way to go around that?

https://gist.github.com/TBS1996/7a6712840eed4b592a46c52ac93a1e25

relay-server/src/actors/global_config.rs
```diff
@@ -310,14 +334,69 @@ mod tests {
...
         Controller::shutdown(ShutdownMode::Immediate);
-        tokio::time::sleep(config.global_config_fetch_interval() * 2).await;
+        tokio::time::sleep(fetch_interval * 2).await;
 
         assert!(service.send(Get).await.is_ok());
     }
+
+    #[tokio::test]
+    #[should_panic]
+    async fn managed_relay_makes_upstream_request() {
```
TBS1996, line 347: nice job fixing this! was it the timer that you added which fixed it ?

relay-server/src/actors/global_config.rs
```diff
@@ -254,23 +258,43 @@ impl Service for GlobalConfigService {
...
-            } else {
-                // NOTE(iker): not making a request results in the sleep handler
-                // not being reset, so no new requests are made.
-                relay_log::info!("fetching global configs disabled: no credentials configured");
-            }
+            match self.config.relay_mode() {
+                RelayMode::Managed => {
+                    if self.config.has_credentials() {
+                        relay_log::info!("serving global configs fetched from upstream");
+                        self.request_global_config();
```
TBS1996: btw, would you think its bad that before the first response comes, that the default global config will be served? are there cases where we would rather spool envelopes until it arrives?

relay-server/src/actors/global_config.rs
iker-barriocanal: I don't remember what the error was, but the missing piece initially was the `#[should_panic]` if that's what you mean.

relay-server/src/actors/global_config.rs
iker-barriocanal: That's an interesting edge case. Technically it's possible that it happens, but it should be unlikely so we accepted it when we discussed the state machine in the processor. I documented this limitation [here (internal)](https://www.notion.so/sentry/Limitations-of-the-global-config-service-611c9f3700794c5a8e0255512f74be81?pvs=4#0637008d6aa248f9849e85bfb6ddb850).

iker-barriocanal (commented):
I'd merge this as-is, but I'm biased as I've also worked on this so I'll leave the approval to the team.

relay-server/src/actors/global_config.rs
TBS1996: I think I see the problem, when I wanted to return a bool from the mock, itd wait until the response from upstream and thereby always fail as an error would be logged. Pity because it'd make it easier to write these tests and would make them more explicit.

relay-server/src/actors/global_config.rs
```diff
@@ -254,23 +258,43 @@ impl Service for GlobalConfigService {
...
+                    } else {
+                        relay_log::info!(
+                            "serving default global configs due to lacking credentials"
+                        );
```
TBS1996: doing nothing means serving default config, because we initiate the watch with default in new();

relay-config/src/config.rs
```diff
@@ -1465,6 +1465,12 @@ impl Config {
         Ok(())
     }
 
+    /// Updates the credentials without creating the file, for testing purposes.
+    pub fn update_credentials(&mut self) {
+        let creds = Credentials::generate();
+        self.credentials = Some(creds);
+    }
```
TBS1996: is it possible to conditionally compile such functions only on testing? I tried the cfg(test) feature, but when i run `make test-rust-all` it fails to find the function, despite that it should compile in testing mode. Strangely, running the tests directly from vscode works.

iker-barriocanal: I also made some attempts playing around with features and didn't manage. I'm open to updating this if somebody knows a better way.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 **Features**:
 
+- Use static global configuration if file is provided and not in managed mode. ([#2458](https://github.com/getsentry/relay/pull/2458))
```
jan-auer: This is either "Internal" or a "Bug Fix", but it's not a user-visible feature.

By the way, neither is the next entry.

relay-dynamic-config/src/global.rs
```diff
@@ -13,6 +15,28 @@ pub struct GlobalConfig {
     pub measurements: Option<MeasurementsConfig>,
 }
 
+impl GlobalConfig {
+    /// Returns the full filepath of the global config.
+    fn path(base: &Path) -> PathBuf {
+        base.join("global_config.json")
```
jan-auer: _nit_: Since this function is so simple and used a single time, feel free to inline it.

relay-dynamic-config/src/global.rs
```diff
@@ -13,6 +15,28 @@ pub struct GlobalConfig {
...
+    }
+
+    /// Loads the [`GlobalConfig`] from a file if it's provided.
```
jan-auer, line 21: In this doc, please describe what should be passed as `base`.

relay-dynamic-config/src/global.rs
```diff
@@ -13,6 +15,28 @@ pub struct GlobalConfig {
...
+    pub fn load(base: &Path) -> anyhow::Result<Option<Self>> {
+        let path = Self::path(base);
+
+        match path.exists() {
+            true => {
```
jan-auer: _nit_: use `if ... else` here.

relay-dynamic-config/src/global.rs
```diff
@@ -13,6 +15,28 @@ pub struct GlobalConfig {
...
+                let file_contents = std::fs::read_to_string(path)?;
+                let global_config = serde_json::from_str::<GlobalConfig>(file_contents.as_str())?;
```
jan-auer: _nit_: There's a way to avoid going through an intermediate string. See the example of [`serde_json::from_reader`](https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html).

```suggestion
    let file = BufReader::new(File::open(path)?);
    let global_config = serde_json::from_reader(file)?;
```

relay-server/src/actors/global_config.rs
```diff
@@ -257,23 +261,43 @@ impl Service for GlobalConfigService {
...
-                // for now.
-                self.update_global_config();
-            } else {
-                // NOTE(iker): not making a request results in the sleep handler
-                // not being reset, so no new requests are made.
-                relay_log::info!("fetching global configs disabled: no credentials configured");
-            }
+            match self.config.relay_mode() {
+                RelayMode::Managed => {
+                    if self.config.has_credentials() {
```
jan-auer: As a thought for simplifying this, what if we simply match for managed mode and catch the rest collectively?

```rust
if config.relay_mode() == RelayMode::Managed && config.has_credentials() {
    // ...
} else {
    // ...
}
```

The case that managed mode is enabled but there are no credentials is caught during startup, so that would be unreachable.

relay-server/src/actors/global_config.rs
```diff
@@ -313,14 +337,69 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    #[should_panic]
+    async fn managed_relay_makes_upstream_request() {
+        relay_test::setup();
+        tokio::time::pause();
+
+        let (upstream, handle) = mock_service("upstream", (), |(), _| {
+            panic!();
```
jan-auer, line 352: Instead of `#[should_panic]`, which is intended to be used for expected panics, we should rather make sure the upstream service returns a default config.

iker-barriocanal: IIUC, you suggest returning a response as if it were the actual upstream service. To actually return a global config we need to do an async operation in the closure, currently unstable in Rust. I think the `panic` makes the test easy to understand, and removes the complexity of dealing with `mock_service`. Maybe I'm missing something?

relay-server/src/actors/global_config.rs
iker-barriocanal: In this case, I think being explicit provides more benefits, mostly ease of understanding and being more defensive to changes.

- It's harder to understand the case where relay mode is managed and there aren't credentials, besides that it requires prior knowledge of how relay startup works. We could add a comment, but I don't think that's a good alternative to avoid an `if` branch.
- Changes to relay startup don't impact the global config service negatively.
- The exhaustive `match` on relay modes ensures that we must consider the global config service if adding new modes to relay.

relay-config/src/config.rs
olksdr: The`#[cfg(test)]` does not propagate to the external crates, and only work within the current one.

It's possible to achieve this, if really really necessary with `features`, by e.g. adding `testing` feature to the crate and hiding all the functionality behind that feature. Then in the crate which has to use this `testing` functionality you will need to add maybe `dev` dependency and enable the feature.

relay-dynamic-config/src/global.rs
olksdr: I think what @jan-auer  meant is just remove the function and use the join in the `load` function.  Since it's used in only one place, from what I can see.

relay-dynamic-config/src/global.rs
olksdr: This was resolved but the docs are not updated. Please, do not resolve the comments, which are not addressed.

relay-server/src/actors/global_config.rs
olksdr: I'm here with @jan-auer . The thing is with unit tests, it can also panic for completely different reason when something changes in the code, and therefore the tests will be testing completely wrong thing. 

It's better to make sure the behaviour is correct, but checking the status of some operation, like in this case that resulting config is `default`

relay-server/src/actors/global_config.rs
```diff
@@ -257,23 +261,43 @@ impl Service for GlobalConfigService {
...
+                        relay_log::info!(
+                            "serving default global configs due to lacking credentials"
+                        );
+                    }
+                }
+                RelayMode::Static | RelayMode::Proxy | RelayMode::Capture => {
+                    match GlobalConfig::load(self.config.path()) {
+                        Ok(Some(from_file)) => {
+                            relay_log::info!("serving static global config loaded from file");
+                            self.global_config_watch.send(Arc::new(from_file)).ok();
```
olksdr: From the docs on [subscribe](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.subscribe) 
> All messages sent before this call to subscribe are initially marked as seen by the new Receiver.

Can it happen, that e.g.  `EnvelopeProcessorService` will never gets update over the watch, since it will subscribe to the `watch` after this static config is sent? And then we actually never send any other update? 

Or maybe I'm missing something?

relay-server/src/actors/global_config.rs
```diff
@@ -257,23 +261,43 @@ impl Service for GlobalConfigService {
...
+                    }
+                }
+            };
 
             loop {
                 tokio::select! {
                     biased;
 
-                    () = &mut self.fetch_handle => self.update_global_config(),
+                    () = &mut self.fetch_handle => self.request_global_config(),
```
olksdr, line 291: In the static mode we actually never get this `fetch_handle` trigger? Is it right?

relay-server/src/actors/global_config.rs
jan-auer: It's not a concern of the global config service to understand whether or not credentials are needed. It only needs to know that if it's in managed mode, it should make requests. Whether or not all preconditions for upstream are met to make requests is outside of its scope. 

The issue with adding this here is that it duplicates validation logic or invariants across the codebase, which makes them hard to maintain or change. It creates tighter-than-necessary coupling.

relay-server/src/actors/global_config.rs
jan-auer: @TBS1996 and I spent some time on this today. In order to move forward with this PR, let's keep the `should_panic` for now. He'll link a follow-up issue here to clean this up. 

We'll implement some utilities to make mocking upstream requests easier. The boilerplate needed to achieve this now for a single test is too complex.

relay-config/src/config.rs
jan-auer: This is now addressed by two subsequent changes:
- At first, we introduced a `test` feature like @olksdr described above
- Later, we updated the existing `regenerate_credentials` method and added a flag to opt out of saving to a file.

relay-server/src/actors/global_config.rs
jan-auer: You're right. Every subscriber must use the _initial_ value that the subscription has before awaiting `changed()`. Essentially, this would be the flow:

```rust
let subscription = global_config.send(Subscribe)?;
self.config = subscription.borrow().clone();

while let Ok(()) = subscription.changed().await {
  self.config = subscription.borrow().clone();
}
```

or

```rust
let subscription = global_config.send(Subscribe)?;

loop {
  self.config = subscription.borrow().clone();
  subscription.changed().await?;
}
```

jan-auer (approved):

relay-server/src/actors/global_config.rs
iker-barriocanal: Yes, that's right. The fetch handle is only scheduled when handling a request, and a request is only made in managed modes; thus, in static mode, this fetch handle is never resolved.

iker-barriocanal (approved):

PR #2460, opened 2023-09-04
test(metrics): Add tests for protocol docs

jan-auer:
To ensure that the documented payloads are valid and roundtrip properly, this PR
moves all payloads from doc comments into test fixture files and adds tests to
assert their content. As an added benefit, the module-level docs and struct docs
now contain identical examples.

#skip-changelog

### DIFF
--- a/relay-system/Cargo.toml
+++ b/relay-system/Cargo.toml
@@ -14,7 +14,7 @@ futures = { workspace = true }
 once_cell = { workspace = true }
 relay-log = { path = "../relay-log" }
 relay-statsd = { path = "../relay-statsd" }
-tokio = { workspace = true, features = ["signal"] }
+tokio = { workspace = true, features = ["rt", "signal"] }
 
 [dev-dependencies]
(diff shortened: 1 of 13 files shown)


relay-system/Cargo.toml
```diff
@@ -14,7 +14,7 @@ futures = { workspace = true }
 once_cell = { workspace = true }
 relay-log = { path = "../relay-log" }
 relay-statsd = { path = "../relay-statsd" }
-tokio = { workspace = true, features = ["signal"] }
+tokio = { workspace = true, features = ["rt", "signal"] }
```
iker-barriocanal, line 17: Should this be a dev dependency?

iker-barriocanal (approved):

relay-system/Cargo.toml
jan-auer: It's actually required by the implementation in `relay_system::service`, so it's needed to compile the non-dev version of the crate. The reason this didn't show up in CI is because the `rt` feature is implicitly enabled by another dependency in the workspace.

Issue #2449, opened 2023-08-31
Expose relay metrics protocol to clients


iambriccardo:
The StatsD protocol Relay has implemented so far is a DataDog flavor, called the "[DogStatsD protocol](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics#the-dogstatsd-protocol)". On top of vanilla statsd, this protocol offers these additional options:
1. Tags. This has been implemented from the start
2. Sample rates. Not implemented, TBD.
3. Value packing. Not implemented and should be added.
4. Container IDs. Not implemented, but also not needed.
5. Timestamps. Implemented with a different syntax, can be aligned.

There's no reason that our submission protocol diverges from DogStatsD, so we can implement full compatibility with this protocol. Along with this change, Relay can parse all metric payloads directly into buckets, rather than a separate metric type.

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/relay/pull/2460
- [ ] https://github.com/getsentry/relay/pull/2466
- [ ] https://github.com/getsentry/relay/pull/2468
- [ ] https://github.com/getsentry/relay/pull/2470
- [ ] https://github.com/getsentry/relay/pull/2472
- [ ] https://github.com/getsentry/relay/pull/2474
```

jan-auer:
I'll compare with protocols from existing collectors first and then bring forward a proposal for improvements.

jan-auer:
Completed.

PR #2468, opened 2023-09-04
feat(metrics): Implement statsd parsing for buckets

jan-auer:
Implements a superset of the existing statsd protocol that allows to send entire buckets with value aggregations. Metric items containing statsd payloads are now parsed directly into buckets and then merged into the aggregator without a detour through the `Metric` type. This supports payloads with both a single value and aggregate values.

Note that `Bucket::parse_all` now assumes statsd rather than JSON. All existing use of `Bucket::parse_all` has been replaced with a direct call to `serde_json::from_slice` in 61770fa. In turn, parsing statsd payloads into the `Metric` type has been removed in this PR. However, there is remaining use of `Metric` in metric extraction, which will be removed in follow-up PRs.

Existing caveats in this PR:
 - Set hashing is not implemented at this point and will be re-introduced in a followup. 
 - Sets are not statsd compatible. In statsd, sets behave more like gauges, where the submitted number is the number of unique values seen. In our case, the submitted number is a unique identifier. To fix this, we will introduce a new metric type for sets.
 - Distribution and set bucket values are potentially inefficient when a low number of values is being inserted as they always require a heap allocation. Since the `GaugeValue` leaves room for at least 5 values, we can explore the use of `SmallVec` and similar types.

Follow-up tasks have been recorded in https://github.com/getsentry/relay/issues/2449


relay-metrics/src/bucket.rs
```diff
@@ -614,20 +700,85 @@ pub struct Bucket {
...
+        let value = match ty {
+            MetricType::Counter => BucketValue::Counter(parse_counter(values_str)?),
+            MetricType::Distribution => BucketValue::Distribution(parse_distribution(values_str)?),
+            MetricType::Set => BucketValue::Set(parse_set(values_str)?),
+            MetricType::Gauge => BucketValue::Gauge(parse_gauge(values_str)?),
+        };
+
+        let mut bucket = Bucket {
+            timestamp,
+            width: 0,
```
iker-barriocanal, line 773: Are we using `0` as a placeholder for now?

relay-metrics/src/bucket.rs
```diff
@@ -614,20 +700,85 @@ pub struct Bucket {
...
+            name: mri.to_string(),
+            value,
+            tags: Default::default(),
+        };
 
-    // /// Parses a set of metric bucket from the JSON protocol.
-    // pub fn parse_all(slice: &[u8]) -> Result<Vec<Bucket>, ParseBucketError> {
-    //     serde_json::from_slice(slice).map_err(ParseBucketError)
-    // }
+        for component in components {
```
iker-barriocanal, line 779: According to the docstrings' format `[<ns>/]<name>[@<unit>]:<value>|<type>[|#<tags>]`, there should only be one component left. Why are we doing a loop here?

relay-metrics/src/bucket.rs
```diff
@@ -614,20 +700,85 @@ pub struct Bucket {
...
+    /// See the [`Bucket`] for more information on the protocol.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use relay_metrics::{Bucket, UnixTimestamp};
+    ///
+    /// let bucket = Bucket::parse(b"response_time@millisecond:57|d", UnixTimestamp::now())
+    ///     .expect("metric should parse");
+    /// ```
```
iker-barriocanal, line 799: nit: should this example be a doctest?

relay-metrics/src/bucket.rs
```diff
@@ -648,6 +799,50 @@ impl MetricsContainer for Bucket {
...
+            if self.slice.is_empty() {
+                return None;
+            }
+
+            let mut split = self.slice.splitn(2, |&b| b == b'\n');
+            let current = split.next()?;
+            self.slice = split.next().unwrap_or_default();
+
+            let string = match std::str::from_utf8(current) {
+                Ok(string) => string.strip_suffix('\r').unwrap_or(string),
```
iker-barriocanal, line 883: If this is because of Windows' end lines, I think a comment could be helpful.

relay-metrics/src/bucket.rs
```diff
@@ -648,6 +799,50 @@ impl MetricsContainer for Bucket {
...
+                Err(_) => return Some(Err(ParseBucketError(()))),
+            };
+
+            if !string.is_empty() {
```
iker-barriocanal, line 887: Q: if this branch is not matched, does the method return `None` by default?

relay-server/src/actors/processor.rs
```diff
@@ -2632,13 +2632,21 @@ impl EnvelopeProcessorService {
...
 
-                let metrics =
-                    Metric::parse_all(&payload, timestamp).filter_map(|result| result.ok());
+                let mut buckets = Vec::new();
+                for bucket_result in Bucket::parse_all(&payload, timestamp) {
+                    match bucket_result {
+                        Ok(bucket) => buckets.push(bucket),
+                        Err(error) => relay_log::debug!(
+                            error = &error as &dyn Error,
+                            "failed to parse metric bucket from statsd",
```
iker-barriocanal: IIUC, the metric parsing can fail for both statsd and sentry metrics. Am I missing something?

relay-metrics/src/bucket.rs
```diff
@@ -556,10 +565,105 @@ impl From<MetricValue> for BucketValue {
...
 
-/// Error returned when parsing or serializing a [`Bucket`].
-#[derive(Debug, Error)]
-#[error("failed to parse metric bucket")]
-pub struct ParseBucketError(#[source] serde_json::Error);
+/// Parses a list of counter values separated by colons and sums them up.
+fn parse_counter(string: &str) -> Option<CounterType> {
+    let mut sum = CounterType::default();
+    for component in string.split(':') {
+        sum += component.parse::<CounterType>().ok()?;
```
iker-barriocanal, line 574: Not sure if a `sum.insert` or `sum.add` helper keeps the API cleaner and consistent with other distribution types.

relay-metrics/src/bucket.rs
```diff
@@ -614,19 +747,84 @@ pub struct Bucket {
...
+    /// ```
+    pub fn parse(slice: &[u8], timestamp: UnixTimestamp) -> Result<Self, ParseMetricError> {
+        let string = std::str::from_utf8(slice).or(Err(ParseMetricError(())))?;
+        Self::parse_str(string, timestamp).ok_or(ParseMetricError(()))
     }
 
-    /// Serializes the given buckets to the JSON protocol.
-    pub fn serialize_all(buckets: &[Self]) -> Result<String, ParseBucketError> {
-        serde_json::to_string(&buckets).map_err(ParseBucketError)
+    /// Parses a set of metric aggregates from the raw protocol.
```
iker-barriocanal, line 805: nit: technically, this function does not parse the metrics (as that's done lazily in the iterator). Should we update the docstrings? (This was not added in this PR, but I thought it's worth pointing it out.)

iker-barriocanal (approved):
Nicee. Left some questions, PR looks good.

relay-metrics/src/bucket.rs
```diff
@@ -556,10 +565,105 @@ impl From<MetricValue> for BucketValue {
...
+    for component in string.split(':') {
+        sum += component.parse::<CounterType>().ok()?;
+    }
+    Some(sum)
+}
+
+/// Parses a distribution from a list of floating point values separated by colons.
+fn parse_distribution(string: &str) -> Option<DistributionValue> {
+    let mut dist = DistributionValue::default();
+    for component in string.split(':') {
```
jjbayer: nit: Should we define a constant for `:`?

```suggestion
    for component in string.split(VALUE_SEPARATOR) {
```

relay-metrics/src/bucket.rs
```diff
@@ -556,10 +565,105 @@ impl From<MetricValue> for BucketValue {
...
+        dist.insert(component.parse().ok()?);
+    }
+    Some(dist)
+}
+
+/// Parses a set of hashed numeric values.
+fn parse_set(string: &str) -> Option<SetValue> {
+    let mut set = SetValue::default();
+    for component in string.split(':') {
+        set.insert(component.parse().ok()?);
```
jjbayer, line 593: You mentioned it in the PR description, but could we add a code comment about the missing hashing step here?

 > Set hashing is not implemented at this point and will be re-introduced in a followup.

relay-metrics/src/bucket.rs
```diff
@@ -556,10 +565,105 @@ impl From<MetricValue> for BucketValue {
...
+/// The gauge can either be given as a single floating point value, or as a list of exactly five
+/// values in the order of [`GaugeValue`] fields.
+fn parse_gauge(string: &str) -> Option<GaugeValue> {
+    let mut components = string.split(':');
+
+    let first = components.next()?.parse().ok()?;
+    Some(if let Some(min) = components.next() {
+        GaugeValue {
+            max: first,
+            min: min.parse().ok()?,
```
jjbayer: Kinda weird to have `max` before `min`. Maybe we should flip them in both the protocol and the `GaugeValue` type?

relay-metrics/src/bucket.rs
jan-auer: StatsD cannot convey a bucket width, so we mark the bucket with a width zero to indicate that it carries metrics for a specific point in time. This is in line with how `Metric` is converted to `Bucket`.

relay-metrics/src/bucket.rs
jan-auer: We'll soon add new components for time stamps and sample rates.

relay-metrics/src/bucket.rs
jan-auer: It is a doc test, that's how it looks.

relay-metrics/src/bucket.rs
jan-auer: It's publicly documented in the doc comment of `Bucket::parse_all` and also tested.

relay-metrics/src/bucket.rs
jjbayer: @iker-barriocanal it keeps looping.

jan-auer: If this branch isn't matched, you go to the end of the loop, and then it goes into the next iteration. In this specific case, it will continue to consume lines from the slice until the slice is empty and then it goes to the explicit `return None`.

relay-server/src/actors/processor.rs
jan-auer: what is a "sentry metric" in this case?

relay-metrics/src/bucket.rs
jan-auer: This is not part of the API, so we can keep this simple.

relay-metrics/src/bucket.rs
jan-auer: It's fine, it returns an iterator over parsed buckets. The signature and doc comment should make it clear that this is a lazy iterator.

relay-server/src/actors/processor.rs
jjbayer: Maybe this helps:

```suggestion
                            "failed to parse metric bucket from statsd format",
```

jjbayer (approved):
See comments, especially the one about the order of gauge fields.

relay-metrics/src/bucket.rs
jan-auer: Great point, updated.

jjbayer (approved):

PR #2471, opened 2023-09-05
feat(spans): Simplify SQL CASE expressions

jjbayer:
Dynamically generated `CASE WHEN .. THEN .. ` expressions can generate high cardinality, so replace them with a simple placeholder:

```sql
UPDATE tbl SET foo = CASE WHEN 1 THEN 10 WHEN 2 THEN 20 ELSE 30 END
-- becomes
UPDATE tbl SET foo = CASE WHEN .. THEN .. END
```

ref: [Internal issue](https://www.notion.so/sentry/Parameterize-CASE-WHEN-clauses-63d746fd7d8545f580251998341b40fe?pvs=4)

#skip-changelog


relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -540,6 +540,23 @@ mod tests {
         "SELECT %s; SELECT %s"
     );
 
+    scrub_sql_test!(
+        case_when,
+        "UPDATE tbl SET foo = CASE WHEN 1 THEN 10 WHEN 2 THEN 20 ELSE 30 END",
+        "UPDATE tbl SET foo = CASE WHEN .. THEN .. END"
+    );
+    scrub_sql_test!(
```
iker-barriocanal:
```suggestion
    );

    scrub_sql_test!(
```

iker-barriocanal (approved):

gggritso (approved):
🎉 👍🏻

PR #2473, opened 2023-09-05
feat(spans): Collapse simple SQL update statements

jjbayer:
Collapse lists of column-to-value assignments in SQL `UPDATE` statements. For example,

```sql
UPDATE foo SET a = 1, b = 2 WHERE some_condition
-- becomes
UPDATE foo SET .. WHERE some_condition
```

#skip-changelog


relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -181,10 +186,21 @@ impl VisitorMut for NormalizeVisitor {
...
+                    && assignments
+                        .iter()
+                        .all(|a| matches!(a.value, Expr::Value(_)))
+                {
+                    *assignments = vec![Assignment {
+                        id: vec![Ident::new("___UPDATE_LHS___")],
+                        value: Expr::Value(Value::Null),
+                    }]
+                } else {
+                    for assignment in assignments.iter_mut() {
```
iker-barriocanal, line 209: When there's either 1 or 0 assignments, I'd just take it (instead of iterating).

iker-barriocanal (approved):

PR #2476, opened 2023-09-05
ref(server): Create metric buckets in place of metrics

jan-auer:
The `Metrics` type is being phased out in favor of `Bucket`. Since all metrics
can be represented as a bucket with a single entry, this PR updates metric
extraction to emit buckets directly.

Note that we currently hard-code the bucket width of `0` everywhere. This is a
workaround for the current public API and is pending revision.

#skip-changelog

### DIFF
--- a/relay-metrics/src/bucket.rs
+++ b/relay-metrics/src/bucket.rs
@@ -10,7 +10,7 @@ use crate::protocol::{
     self, CounterType, DistributionType, GaugeType, MetricResourceIdentifier, MetricType,
     MetricValue, MetricsContainer, SetType,
 };
-use crate::ParseMetricError;
+use crate::{hash_set_value, ParseMetricError};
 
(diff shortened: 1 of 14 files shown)


relay-metrics/src/bucket.rs
```diff
@@ -513,6 +513,36 @@ pub enum BucketValue {
...
+    }
+
+    /// Returns a bucket value representing a distribution with a single given value.
+    pub fn distribution(value: DistributionType) -> Self {
+        Self::Distribution(dist![value])
+    }
+
+    /// Returns a bucket value representing a set with a single given hash value.
+    pub fn set(value: SetType) -> Self {
+        Self::Set(std::iter::once(value).collect())
```
jjbayer, line 528: TIL `std::iter::once`.

jjbayer (approved):
Looks great!

PR #2479, opened 2023-09-05
feat(crons): Add test for basic monitor ingestion

evanpurkhiser:
This is extracted from https://github.com/getsentry/relay/pull/2448

It was actually incorrectly named in that PR and pretty unrealted to the changes

#skip-changelog

### DIFF
--- a/tests/integration/test_monitors.py
+++ b/tests/integration/test_monitors.py
@@ -11,6 +11,27 @@ def generate_check_in(slug):
     }
 
 
+def test_monitor_ingest(mini_sentry, relay):
+    relay = relay(mini_sentry)
+    mini_sentry.add_basic_project_config(42)
+
+    check_in = generate_check_in("my-monitor")
+    relay.send_check_in(42, check_in)
+
+    envelope = mini_sentry.captured_events.get(timeout=1)
+    assert len(envelope.items) == 1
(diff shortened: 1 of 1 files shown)


tests/integration/test_monitors.py
```diff
@@ -10,6 +11,26 @@ def generate_check_in(slug):
     }
 
 
+def test_monitor_ingest(mini_sentry, relay):
```
iker-barriocanal, line 14: Not sure if I understand. What's the behavior we're trying to test? Looking at the test below the only difference I see is the `monitors_consumer` fixture, but not relay behavior. Maybe I'm missing something?

evanpurkhiser: This test doesn’t run the full processing pipeline. I believe it’s faster cc @jan-auer

jan-auer (approved):

tests/integration/test_monitors.py
jan-auer: That's right. This test runs without processing mode and doesn't need Kafka, so it's a fair bit faster to run. Also, it ensures that PoPs work correctly in forwarding their data.

jan-auer:
@evanpurkhiser please run `make format-python`; then this is good to merge!

PR #2480, opened 2023-09-05
feat(monitors): Adds MonitorSeat data category

mbauer404:
Creates the MonitorSeat data category that represents a monitor job. The seats are not ingested (the monitor checkins use the Monitor data category), but we define it here to prevent clashing values since this data category enumeration is also used outside of Relay via the Python package.


jan-auer (approved):

jan-auer:
Pre-approving for when this passes CI.

dashed (approved):

vbro (approved):

PR #2481, opened 2023-09-05
fix(crons): Allow empty check_in_id via POST

evanpurkhiser:
You are not required to send a check_in_id when POSTing a check-in.

#skip-changelog

### DIFF
--- a/tests/integration/test_monitors.py
+++ b/tests/integration/test_monitors.py
@@ -122,21 +122,20 @@ def test_monitor_post_json_body(mini_sentry, relay):
     relay = relay(mini_sentry)
     mini_sentry.add_basic_project_config(42)
 
-    check_in = generate_check_in("my-monitor")
-    # NB: Slug Should not be part of the payload!
-    monitor_slug = check_in.pop("monitor_slug")
-
-    # Include monitor_config since this is a major use-case for POST w/ json
-    check_in["monitor_config"] = {
-        "schedule": {"type": "crontab", "value": "0 * * * *"},
-        "checkin_margin": 5,
-        "max_runtime": 10,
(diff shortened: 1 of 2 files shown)


tests/integration/test_monitors.py
```diff
@@ -122,21 +122,20 @@ def test_monitor_post_json_body(mini_sentry, relay):
     relay = relay(mini_sentry)
     mini_sentry.add_basic_project_config(42)
 
-    check_in = generate_check_in("my-monitor")
```
evanpurkhiser, line 125: `generate_check_in` wasn't giving us much, I just dropped it in favor of inlining the check-in body.

tests/integration/test_monitors.py
```diff
@@ -145,7 +144,7 @@ def test_monitor_post_json_body(mini_sentry, relay):
 
     check_in = json.loads(item.get_bytes().decode())
     assert check_in == {
-        "check_in_id": "a460c25ff2554577b920fcfacae4e5eb",
+        "check_in_id": "00000000000000000000000000000000",
```
evanpurkhiser, line 147: This is the main change. It should be possible to NOT specify an ID.

jan-auer (approved):
LGTM. Keep in mind that now you will always have to check if the ID is NIL rather than `is not None` in python.

evanpurkhiser:
> LGTM. Keep in mind that now you will always have to check if the ID is NIL rather than is not None in python.

Not a problem, we already were handling this: https://github.com/getsentry/sentry/blob/49f7d58744444a414aa5c78fffc748e7fe2667cb/src/sentry/monitors/consumers/monitor_consumer.py#L296-L304

PR #2483, opened 2023-09-06
ref(metrics): Improve distribution value performance

jan-auer:
Represents distribution metrics as plain unsorted vectors instead of
deduplicated btree maps. This offers greatly improved performance
characteristics during insertion, but comes at the expense of
potentially more memory usage if there are a lot of duplicate values.

During benchmarks, this change reveals approximately 500x speedup for
metrics with 10k to 1M data points per bucket. In practice, large
buckets are much more prevalent than duplicate values.

The new implementation uses `SmallVec`. Since distribution values are
always stored within a `BucketValue` that also contains gauges. Since
`GaugeValue` is a larger struct, that leaves space to inline three float
values before allocating a vector for distributions. Usual production
traffic indicates that this covers the majority of distribution buckets,
so this is a worthwhile optimization.


olksdr (approved):
lgtm :shipit:

PR #2485, opened 2023-09-06
fix(spans): Strip more quotes

jjbayer:
This PR fixes two regressions that were caused by switching to a proper SQL parser:

* Simplify identifiers when casts are involved, e.g. `SELECT UPPER("b"."c"::text)`
* Strip quotes from table names in JOINs, e.g. `SELECT * FROM a join "b"`

#skip-changelog


gggritso (approved):
🧹

olksdr (approved):

PR #2495, opened 2023-09-07
ref(metrics): Remove InsertMetrics and related traits

jan-auer:
This change removes `InsertMetrics` in favor of `MergeBuckets`. The
former message was no longer used by production code, as the `Metric`
type is being phased out in favor of directly using `Bucket` for all
metrics. Most notably, all methods that were generic over both metrics
and buckets are now specific to buckets.

Continuation of #2476. 
#skip-changelog

### DIFF
--- a/relay-metrics/src/bucket.rs
+++ b/relay-metrics/src/bucket.rs
@@ -167,7 +167,7 @@ pub enum BucketValue {
     #[serde(rename = "s")]
     Set(SetValue),
     /// Aggregates [`MetricValue::Gauge`](crate::MetricValue::Gauge) values always retaining the
-    /// maximum, minimum, and last value, as well as the sum and count of all values.
+    /// latest, minimum, and maximum value, as well as the sum and count of all values.
     ///
     /// **Note**: The "last" component of this aggregation is not commutative.
     ///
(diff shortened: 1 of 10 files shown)


relay-metrics/src/bucket.rs
```diff
@@ -163,8 +166,8 @@ pub enum BucketValue {
     /// This variant serializes to a list of 32-bit integers.
     #[serde(rename = "s")]
     Set(SetValue),
-    /// Aggregates [`MetricValue::Gauge`] values always retaining the maximum, minimum, and last
-    /// value, as well as the sum and count of all values.
+    /// Aggregates [`MetricValue::Gauge`](crate::MetricValue::Gauge) values always retaining the
+    /// maximum, minimum, and last value, as well as the sum and count of all values.
```
jjbayer:
```suggestion
    /// Aggregates [`MetricValue::Gauge`](crate::MetricValue::Gauge) values always retaining the
    /// latest, minimum, and maximum value, as well as the sum and count of all values.
```

jjbayer (approved):

olksdr (approved):
🚢  it

PR #2496, opened 2023-09-07
ref(crons): Partition monitor check-ins by monitor slug

evanpurkhiser:
This change will group check-ins beloning to the same monitor into the same partition.

This is valuable since it will help to better grantee the ordering of check-ins during processing. Though by no means does this solve all ordering issues

#skip-changelog

### DIFF
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -199,13 +199,14 @@ pub fn process_check_in(
 
     // Use the project_id + monitor_slug as the routing key hint. This helps ensure monitor
     // check-ins are processed in order by consistently routing check-ins from the same monitor.
-    let project_id_slug_key: Vec<u8> = [
-        &project_id.value().to_be_bytes()[..],
+
+    let project_id_slug_key = [
+        &project_id.value().to_be_bytes(),
         check_in.monitor_slug.as_bytes(),
     ]
     .concat();
 
-    let routing_hint = Uuid::new_v5(namespace, project_id_slug_key.as_slice());
+    let routing_hint = Uuid::new_v5(namespace, &project_id_slug_key);
 
     Ok(ProcessedCheckInResult {
         routing_hint,
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -16,6 +16,8 @@
 )]
 #![warn(missing_docs)]
 
+use once_cell::sync::OnceCell;
+use relay_base_schema::project::ProjectId;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
@@ -153,8 +155,23 @@ pub struct CheckIn {
     pub contexts: Option<CheckInContexts>,
 }
 
+/// The result from calling process_check_in
+pub struct ProcessedCheckInResult {
+    /// The routing key to be used for the check-in payload.
+    ///
+    /// Important to help ensure monitor check-ins are processed in order by routing check-ins from
+    /// the same monitor to the same place.
+    pub routing_hint: Uuid,
+
+    /// The JSON payload of the processed check-in.
+    pub payload: Vec<u8>,
+}
+
 /// Normalizes a monitor check-in payload.
-pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError> {
+pub fn process_check_in(
+    payload: &[u8],
+    project_id: ProjectId,
+) -> Result<ProcessedCheckInResult, ProcessCheckInError> {
     let mut check_in = serde_json::from_slice::<CheckIn>(payload)?;
 
     // Missed status cannot be ingested, this is computed on the server.
@@ -176,7 +193,24 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError> 
         return Err(ProcessCheckInError::InvalidEnvironment);
     }
 
-    Ok(serde_json::to_vec(&check_in)?)
+    static NAMESPACE: OnceCell<Uuid> = OnceCell::new();
+    let namespace = NAMESPACE
+        .get_or_init(|| Uuid::new_v5(&Uuid::NAMESPACE_URL, b"https://sentry.io/crons/#did"));
+
+    // Use the project_id + monitor_slug as the routing key hint. This helps ensure monitor
+    // check-ins are processed in order by consistently routing check-ins from the same monitor.
+    let project_id_slug_key: Vec<u8> = [
+        &project_id.value().to_be_bytes()[..],
+        check_in.monitor_slug.as_bytes(),
+    ]
+    .concat();
+
+    let routing_hint = Uuid::new_v5(namespace, project_id_slug_key.as_slice());
+
+    Ok(ProcessedCheckInResult {
+        routing_hint,
+        payload: serde_json::to_vec(&check_in)?,
+    })
 }
 
 fn trim_slug(slug: &mut String) {
@@ -291,6 +325,23 @@ mod tests {
     }
 
     #[test]
+    fn process_simple() {
+        let json = r#"{"check_in_id":"a460c25ff2554577b920fcfacae4e5eb","monitor_slug":"my-monitor","status":"ok"}"#;
+
+        let result = process_check_in(json.as_bytes(), ProjectId::new(1));
+
+        // The routing_hint should be consistent for the (project_id, monitor_slug)
+        let expected_uuid = Uuid::parse_str("3612580a-5d37-594b-9a90-d8142792f9c8").unwrap();
+
+        if let Ok(processed_result) = result {
+            assert_eq!(String::from_utf8(processed_result.payload).unwrap(), json);
+            assert_eq!(processed_result.routing_hint, expected_uuid);
+        } else {
+            panic!("Failed to process check-in")
+        }
+    }
+
+    #[test]
     fn process_empty_slug() {
         let json = r#"{
           "check_in_id": "a460c25ff2554577b920fcfacae4e5eb",
@@ -298,7 +349,7 @@ mod tests {
           "status": "in_progress"
         }"#;
 
-        let result = process_check_in(json.as_bytes());
+        let result = process_check_in(json.as_bytes(), ProjectId::new(1));
         assert!(matches!(result, Err(ProcessCheckInError::EmptySlug)));
     }
 
@@ -311,7 +362,7 @@ mod tests {
           "environment": "1234567890123456789012345678901234567890123456789012345678901234567890"
         }"#;
 
-        let result = process_check_in(json.as_bytes());
+        let result = process_check_in(json.as_bytes(), ProjectId::new(1));
         assert!(matches!(
             result,
             Err(ProcessCheckInError::InvalidEnvironment)
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -200,13 +200,10 @@ pub fn process_check_in(
     // Use the project_id + monitor_slug as the routing key hint. This helps ensure monitor
     // check-ins are processed in order by consistently routing check-ins from the same monitor.
 
-    let project_id_slug_key = [
-        &project_id.value().to_be_bytes(),
-        check_in.monitor_slug.as_bytes(),
-    ]
-    .concat();
+    let slug = &check_in.monitor_slug;
+    let project_id_slug_key = format!("{project_id}:{slug}");
 
-    let routing_hint = Uuid::new_v5(namespace, &project_id_slug_key);
+    let routing_hint = Uuid::new_v5(namespace, project_id_slug_key.as_bytes());
 
     Ok(ProcessedCheckInResult {
         routing_hint,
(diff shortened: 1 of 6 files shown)


evanpurkhiser:
This is more of a proposal right now since it's not building

relay-server/src/actors/store.rs
```diff
@@ -1092,7 +1092,8 @@ impl Message for KafkaMessage {
             Self::Profile(_message) => Uuid::nil(),
             Self::ReplayEvent(message) => message.replay_id.0,
             Self::ReplayRecordingNotChunked(_message) => Uuid::nil(), // Ensure random partitioning.
-            Self::CheckIn(_message) => Uuid::nil(),
+            // Group associated check-ins into the same partition for better ordering.
+            Self::CheckIn(message) => message.check_in_id.0,
```
jjbayer: If I understand this correctly, the `check_in_id` is unique for every check-in, while the monitor slug identifies the monitor? If that's the case, this strategy would not group anything together.

https://github.com/getsentry/relay/blob/8aa5ecde5261c2a522a42f37ba83d493a1827d00/relay-monitors/src/lib.rs#L127-L133

evanpurkhiser: > the check_in_id is unique for every check-in,

This is not actually entirely true, it is possible, and many SDKs do this, to send the SAME check-in for 'updating' check-ins.

What I am talking about is when you have check ins that look like this

```
- 01:00:23 -- 69054ffa-16df-459a-8b3d-6282fb75f973: in_progress
- 01:00:28 -- 69054ffa-16df-459a-8b3d-6282fb75f973: ok (5 second run)
- 01:01:20 -- f0d4caca-354e-49b4-894b-41a842bd9ff0: in_progress
- 01:00:30 -- f0d4caca-354e-49b4-894b-41a842bd9ff0: ok (10 second run)
```

It is not strictly required that you send the check-in ID, in fact, for HTTP style check-ins using the endpoints on relay, check-ins may have a nil ID

```
- 01:00:23 -- 00000000-0000-0000-0000-000000000000: in_progress
- 01:00:28 -- 00000000-0000-0000-0000-000000000000: ok (5 second run)
- 01:01:20 -- 00000000-0000-0000-0000-000000000000: in_progress
- 01:00:30 -- 00000000-0000-0000-0000-000000000000: ok (10 second run)
```

What happens in this scenario is that in our consumer we will identify "the most recent" check-in that is not in a terminal state (`ok`, `error`, `timed_out` are terminal state. `in_progress` is not)

---

What this change is fixing is **the first scenario**. We want those `in_progress` and `ok` check-ins to be in the same partition.

In the future ideally what we'll do is handle ordering issues across partitions at the consumer level. But for now this quick change can hopefully improve ordering reliability.

jan-auer (approved):

evanpurkhiser:
Alright coming back to this. I could use some opinions from the relay team :-)

So I would like for it to partiton by check-in ID, which is a UUID. The problem is the UUID is buried inside the item payload.

Here's where we pull out the UUID from the payload sent to relay

https://github.com/getsentry/relay/blob/086399e58b8cb4468865ade2d0e18010bc5de890/relay-monitors/src/lib.rs#L128-L129

However, the `process_check_in` function just serializes it back to a json byte vector. We call that here

https://github.com/getsentry/relay/blob/086399e58b8cb4468865ade2d0e18010bc5de890/relay-server/src/actors/processor.rs#L1106-L1125

When we finally get to producing the message to kafka here

https://github.com/getsentry/relay/blob/086399e58b8cb4468865ade2d0e18010bc5de890/relay-server/src/actors/store.rs#L717-L723

The payload is opaque.

I considered if the ID of the check-in should also be the event ID, but I think that doesn't make sense since we reuse the check-in ID across check-ins, we probably don't want duplicate event IDs.

What would be the best way of pulling the `check_in_id` out of the envelope here?

relay-server/src/actors/store.rs
jjbayer: > it is possible, and many SDKs do this, to send the SAME check-in for 'updating' check-ins

Thanks, good to know! Might be worth adding this information to the doc comment above `check_in_id`.

jjbayer:
> What would be the best way of pulling the `check_in_id` out of the envelope here?

In `process_check_in`, you could set an internal item header on the item, similar to this:

https://github.com/getsentry/relay/blob/6c3622489324e00bf9ef817f4eab3582a27b65ac/relay-server/src/envelope.rs#L461-L463

That is, an item header that does not get (de)serialized when sending the envelope over the wire, but can be read by the kafka producer. Maybe @jan-auer has better ideas.

relay-server/src/actors/store.rs
```diff
@@ -1100,13 +1104,26 @@ impl Message for KafkaMessage<'_> {
...
-            Self::CheckIn(_message) => Uuid::nil(),
-            Self::Span(_) => Uuid::nil(), // random partitioning
+
+            // Random partitioning
+            Self::Session
+            | Self::Session
+            | Self::Metric
+            | Self::Profile
+            | Self::ReplayRecordingNotChunked
+            | Self::Span => Uuid::nil(),
```
evanpurkhiser: Organized this a bit better I think, clearer separation between message types that are being randomly partitioned and not.

evanpurkhiser:
I've updated this PR to now correctly partition by **monitor slug** (not ID, and not Check-in ID as I originally intended).

This is a partial requirement for https://github.com/getsentry/sentry/issues/55821

---

The changes I've made are as follows

- The envelope has learned a new `routing_hint`. This is similar to what @jjbayer suggested

  > That is, an item header that does not get (de)serialized when sending the envelope over the wire, but can be read by the kafka producer

  However my implementation has it so it **does** get serialized over the wire. We need this because if relay forwards an envelope it MUST remember what the partition key hint is for the monitor check-in.

  Originally I was planning to just store the monitor slug directly as a header on the envelope, but it just felt wrong to add a new header type. So instead with support from @mitsuhiko I've generalized this as `routing_hint`, which _will_ be serialized and passed through the envelope.

  The `routing_hint` is a UUID, which matches what our other routing keys already are. Really this just means it's a 16 byte value though.

- The kafka producer (`store.rs`) has learned to use the `routing_hint` as the kafka partition key for the check-in messages.

  > [!NOTE]
  > In the future I think it would be a good idea for this to work for ALL messages. I left a `TODO` for this.

- The monitors crate `process_check_in` has learned to produce a deterministic `routing_hint` for each check-in, using the monitors SLUG as a factor. This is done using a UUIDv5.

  I had to add the `v5` feature to the UUID crate we pull in.

- The processor has learned to set the `routing_hint` on the envelope item's headers from the monitors `routing_hint`.

relay-monitors/src/lib.rs
```diff
@@ -176,7 +189,19 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
         return Err(ProcessCheckInError::InvalidEnvironment);
     }
 
-    Ok(serde_json::to_vec(&check_in)?)
+    static NAMESPACE: OnceCell<Uuid> = OnceCell::new();
+    let namespace = NAMESPACE
+        .get_or_init(|| Uuid::new_v5(&Uuid::NAMESPACE_URL, b"https://sentry.io/crons/#did"));
```
evanpurkhiser, line 198: Wasn't sure what the best namespace URL would be to use here. cc @mitsuhiko any thoughts?

relay-server/src/actors/store.rs
```diff
@@ -1100,13 +1104,20 @@ impl Message for KafkaMessage<'_> {
...
+
+            // Monitor check-ins use the hinted UUID passed through from the Envelope.
+            //
+            // XXX(epurkhiser): In the future it would be better if all KafkaMessage's would
+            // recieve the routing_key_hint form their envelopes.
+            Self::CheckIn(message) => message.routing_key_hint.unwrap_or_else(Uuid::nil),
+
+            // Random partitioning
+            Self::Session(_)
+            | Self::Metric { .. }
```
jan-auer: Please keep the TODO on metrics, and make it a separate match arm, I still want to address this eventually.

evanpurkhiser: Ah sorry I meant to bring it back, my fault

relay-monitors/src/lib.rs
```diff
@@ -176,7 +189,19 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
...
+
+    // Use the monitor slug as the routing key hint. This helps ensure monitor
+    // check-ins are processed in order by routing check-ins from the same monitor
+    // to the same place.
+    let routing_hint = Uuid::new_v5(namespace, check_in.monitor_slug.as_bytes());
```
jjbayer: This UUID is only unique within a single project, that is, monitors with the same name but from different orgs will get the same routing_hint. That might be good enough for consistency, but should we throw the project ID in here to spread the load more evenly? For example, the name `celerybackend_cleanup` [seems to be a built-in](https://docs.celeryq.dev/en/stable/userguide/configuration.html#result-expires) that's being used a lot, see

```sql
select slug, count(*) from sentry_monitor group by slug order by count desc limit 100;
```

jjbayer (changes requested):
LGTM in general, pressing "Request changes" because of the open question around routing hint uniqueness (see comments).

> However my implementation has it so it does get serialized over the wire. We need this because if relay forwards an envelope it MUST remember what the partition key hint is for the monitor check-in.

@evanpurkhiser For now, `process_check_ins` only runs on processing relays so the item header will only be set in the innermost relay instance IIUC. So strictly speaking it does not need to get serialized into the envelope, but I agree that we should still do it, because if we ever move `process_check_ins` to non-processing relays, an unserialized routing hint would be easy to miss.

relay-monitors/src/lib.rs
evanpurkhiser: Ah yeah, that does sound good to me, I'll update it to handle this. Good call!

relay-monitors/src/lib.rs
```diff
@@ -176,7 +193,20 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
...
-    Ok(serde_json::to_vec(&check_in)?)
+    static NAMESPACE: OnceCell<Uuid> = OnceCell::new();
+    let namespace = NAMESPACE
+        .get_or_init(|| Uuid::new_v5(&Uuid::NAMESPACE_URL, b"https://sentry.io/crons/#did"));
+
+    // Use the monitor slug as the routing key hint. This helps ensure monitor
+    // check-ins are processed in order by routing check-ins from the same monitor
+    // to the same place.
+    let project_id_slug_key = format!("{}-{}", project_id, check_in.monitor_slug);
+    let routing_hint = Uuid::new_v5(namespace, project_id_slug_key.as_bytes());
```
evanpurkhiser: There's probably a more efficient way to slap these together, let me see if I can figure it out

relay-monitors/src/lib.rs
```diff
@@ -176,7 +193,24 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
...
+    let namespace = NAMESPACE
+        .get_or_init(|| Uuid::new_v5(&Uuid::NAMESPACE_URL, b"https://sentry.io/crons/#did"));
+
+    // Use the project_id + monitor_slug as the routing key hint. This helps ensure monitor
+    // check-ins are processed in order by consistently routing check-ins from the same monitor.
+    let project_id_slug_key: Vec<u8> = [
+        &project_id.value().to_be_bytes()[..],
+        check_in.monitor_slug.as_bytes(),
+    ]
+    .concat();
```
evanpurkhiser: Was struggling a bit to get this to work, not 100% sure if this is the most efficient implementation or not.

Feel free to update the PR if there is a better approach

relay-monitors/src/lib.rs
```diff
@@ -176,7 +193,22 @@ pub fn process_check_in(payload: &[u8]) -> Result<Vec<u8>, ProcessCheckInError>
...
+    let namespace = NAMESPACE
+        .get_or_init(|| Uuid::new_v5(&Uuid::NAMESPACE_URL, b"https://sentry.io/crons/#did"));
+
+    // Use the project_id + monitor_slug as the routing key hint. This helps ensure monitor
+    // check-ins are processed in order by consistently routing check-ins from the same monitor.
+
+    let slug = &check_in.monitor_slug;
+    let project_id_slug_key = format!("{project_id}:{slug}");
+
+    let routing_hint = Uuid::new_v5(namespace, project_id_slug_key.as_bytes());
```
jjbayer, line 206: The most efficient would be to instantiate a custom SHA1 hasher here and hash the three values in sequence without allocating. The `uuid` crate provides a `Builder` helper but it does not provide this functionality unfortunately. TL;DR I think allocating here is fine. I simplified a bit to and added a separator, such that for example `(1, '1pass')` and `(11, 'pass')` map to different UUIDs.

jjbayer (approved):

relay-monitors/src/lib.rs
evanpurkhiser: cool. Yeah this is actually what I originally had haha

PR #2497, opened 2023-09-07
chore(self-hosted): Publish relay image to dockerhub again

hubertdeng123:
was removed during the [removal of e2e tests](https://github.com/getsentry/relay/pull/2446), this adds it back

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -440,3 +440,49 @@ jobs:
         run: |
           echo "Testing against ${RELAY_TEST_IMAGE}"
           make test-relay-integration
+
+  publish-to-dockerhub:
+    name: Publish Relay to DockerHub
+    needs: build
+    runs-on: ubuntu-20.04
+    if: ${{ (github.ref_name == 'master') }}
+    steps:
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+      - name: Pull the test image
+        id: image_pull
+        env:
+          IMAGE_URL: ghcr.io/getsentry/relay:${{ github.sha }}
+        shell: bash
+        run: |
+          echo "We poll for the Docker image that the GCB/GHA build produces until it succeeds or this job times out."
+          echo "Polling for $IMAGE_URL"
+          timeout 20m bash -c 'until docker pull "$IMAGE_URL" 2>/dev/null; do sleep 10; done'
+      - name: Get short SHA for docker tag
+        id: short_sha
+        shell: bash
+        run: |
+          SHORT_SHA=$(git rev-parse --short "$GITHUB_SHA")
+          if [[ -z "$SHORT_SHA" ]]; then
+            echo "Short SHA empty? Re-running rev-parse."
+            git rev-parse --short "$GITHUB_SHA"
+          else
+            echo "sha=$SHORT_SHA" >> $GITHUB_OUTPUT
+          fi
+      - name: Push built docker image
+        shell: bash
+        env:
+          SHORT_SHA: ${{ steps.short_sha.outputs.sha }}
+          IMAGE_URL: ghcr.io/getsentry/relay:${{ github.sha }}
+        run: |
+          # only login if the password is set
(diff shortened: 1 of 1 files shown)


azaslavsky (approved):

jan-auer:
@hubertdeng123 the link in the description goes to the same PR, could you link the actual source? I'd like to double-check the original change; then we can merge this back in.

.github/workflows/ci.yml
```diff
@@ -440,3 +440,49 @@ jobs:
...
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+      - name: Pull the test image
+        id: image_pull
+        env:
+          IMAGE_URL: ghcr.io/getsentry/relay:${{ github.sha }}
+        shell: bash
+        run: |
+          echo "We poll for the Docker image that the GCB/GHA build produces until it succeeds or this job times out."
+          echo "Polling for $IMAGE_URL"
+          timeout 20m bash -c 'until docker pull "$IMAGE_URL" 2>/dev/null; do sleep 10; done'
```
jan-auer: - Could you configure a job or step timeout rather than using the `timeout` command here? 
- The `echo` seems redundant since we're not gonna look at that; let's optimize for less verbose actions please.
- Could we inline `IMAGE_URL` instead of declaring it as an env, to make the action smaller?
- Remove the step `id` since it's not needed. I'd even go as far as remove the step name, since it's clear what this does.
- The shell is not needed. 

To put this together, we could write this step as:

```yaml
steps:
   - run: docker pull "ghcr.io/getsentry/relay:${{ github.sha }}"
```

I'm making these suggestions because our CI has become complex and hard to comprehend. We should take great care to write it in a concise way.

.github/workflows/ci.yml
```diff
@@ -440,3 +440,49 @@ jobs:
...
+      - name: Get short SHA for docker tag
+        id: short_sha
+        shell: bash
+        run: |
+          SHORT_SHA=$(git rev-parse --short "$GITHUB_SHA")
+          if [[ -z "$SHORT_SHA" ]]; then
+            echo "Short SHA empty? Re-running rev-parse."
+            git rev-parse --short "$GITHUB_SHA"
+          else
+            echo "sha=$SHORT_SHA" >> $GITHUB_OUTPUT
```
jan-auer: Could you explain this block? 
 - It seems nothing changes between the two invocations of `rev-parse`; is it flakey?
 - If this enters the `if` branch, it skips the `echo`, which means the output for the next step is missing. 
 
 I'd prefer if we just inline `SHORT_SHA=$(git rev-parse --short "$GITHUB_SHA")` into the next step.

.github/workflows/ci.yml
```diff
@@ -440,3 +440,49 @@ jobs:
...
+          fi
+      - name: Push built docker image
+        shell: bash
+        env:
+          SHORT_SHA: ${{ steps.short_sha.outputs.sha }}
+          IMAGE_URL: ghcr.io/getsentry/relay:${{ github.sha }}
+        run: |
+          # only login if the password is set
+          if [[ "${{ secrets.DOCKER_HUB_RW_TOKEN }}" ]]; then echo "${{ secrets.DOCKER_HUB_RW_TOKEN }}" | docker login --username=sentrybuilder --password-stdin; fi
```
jan-auer: Question: What does this step do if the password is _not_ set?

jan-auer (changes requested):

hubertdeng123:
@jan-auer I've updated the link in the description. Most of this code is coming from existing code of a composite GitHub action here: https://github.com/getsentry/action-self-hosted-e2e-tests/blob/main/action.yml

.github/workflows/ci.yml
hubertdeng123: Good suggestion. This block is not useful. Changing.

.github/workflows/ci.yml
hubertdeng123: The tagging and pushing of docker images will then fail. It's probably better to throw an error here if we're not able to login so I'll change that.

jan-auer (approved):

PR #2501, opened 2023-09-08
doc(services): Document global config service must start first

iker-barriocanal:
Services wanting to get timely global config updates can easily get them by subscribing to the global config service. The requirement is for the global config service to start first, to handle subscription messages. This PR documents this requirement.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -85,6 +85,9 @@ impl UpstreamQuery for GetGlobalConfig {
 pub struct Get;
 
 /// The message for receiving a watch that subscribes to the [`GlobalConfigService`].
+///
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/global_config.rs
```diff
@@ -85,6 +85,9 @@ impl UpstreamQuery for GetGlobalConfig {
 pub struct Get;
 
 /// The message for receiving a watch that subscribes to the [`GlobalConfigService`].
+///
+/// The global config service must be up and running, else the subscription
+/// fails.
```
jjbayer, line 90: What happens exactly when you try to subscribe earlier? Might be worth elaborating.

jjbayer (approved):

olksdr (approved):

PR #2503, opened 2023-09-11
ref(metrics): Remove deprecated `Metric` type and improve docs

jan-auer:
Removes the unused `Metric` and `MetricValue` types in favor of `Bucket` and
`BucketValue`. Documentation has been moved to these types and expanded with
more examples.

### DIFF
--- a/relay-metrics/src/bucket.rs
+++ b/relay-metrics/src/bucket.rs
@@ -614,7 +614,7 @@ pub struct Bucket {
 }
 
 impl Bucket {
-    /// Parse a statsd-compatible payload.
+    /// Parses a statsd-compatible payload.
     ///
     /// ```text
     /// [<ns>/]<name>[@<unit>]:<value>|<type>[|#<tags>]`
--- a/relay-metrics/src/bucket.rs
+++ b/relay-metrics/src/bucket.rs
@@ -136,52 +136,124 @@ pub type SetValue = BTreeSet<SetType>;
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(tag = "type", content = "value")]
 pub enum BucketValue {
-    /// Aggregates [`MetricValue::Counter`](crate::MetricValue::Counter) values by adding them into
-    /// a single value.
+    /// Counts instances of an event ([`MetricType::Counter`]).
+    ///
+    /// Counters can be incremented and decremented. The default operation is to increment a counter
+    /// by `1`, although increments by larger values are equally possible.
+    ///
+    /// # Statsd Format
+    ///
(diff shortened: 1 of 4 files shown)


relay-metrics/src/protocol.rs
```diff
@@ -213,26 +173,70 @@ impl fmt::Display for MetricNamespace {
...
-/// (excl. timestamp and tags).
+/// A unique identifier for metrics including typing and namespacing.
+///
+/// MRIs have the format `<type>:<namespace>/<name>[@<unit>]`. The unit is optional and defaults to
+/// [`MetricUnit::None`].
+///
+/// # Statsd Format
+///
+/// In the statsd submission payload, MRIs are sent in a more relaxed format:
+/// `[<namespace/]name[@unit]`. The differences to the internal MRI format are:
```
iker-barriocanal: to match the format above:
```suggestion
/// `[<namespace>/]<name>[@<unit>]`. The differences to the internal MRI format are:
```

relay-metrics/src/bucket.rs
```diff
@@ -452,33 +517,105 @@ fn parse_timestamp(string: &str) -> Option<UnixTimestamp> {
...
+    /// # Example
+    ///
+    /// ```text
+    /// endpoint.hits:1|c|T1615889440
+    /// ```
     pub timestamp: UnixTimestamp,
+
     /// The length of the time window in seconds.
+    ///
+    /// To initialize a new bucket, choose `0` as width. Once the bucket is tracked by Relay's
```
iker-barriocanal, line 540: What's relay behavior if I don't set `0`?

iker-barriocanal (approved):

relay-metrics/src/bucket.rs
```diff
@@ -452,33 +517,105 @@ fn parse_timestamp(string: &str) -> Option<UnixTimestamp> {
...
+    /// ```text
+    /// endpoint.hits:1|c|#route:user_index,environment:production,release:1.4.0
+    /// ```
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub tags: BTreeMap<String, String>,
 }
 
 impl Bucket {
-    /// Parse statsd-compatible payload of format
+    /// Parse a statsd-compatible payload.
```
olksdr:
```suggestion
    /// Parses a statsd-compatible payload.
```

olksdr (approved):
🚢  it!

relay-metrics/src/bucket.rs
jan-auer: It will merge the bucket based on the timestamp at the middle of the bucket's time range.

PR #2505, opened 2023-09-12
fix(server): Immediatly set global config to watch contents.

### LINKED ISSUE #2504: Watch might miss out on static global config file if subscribed after watch was updated
https://github.com/getsentry/relay/pull/2458#discussion_r1322529076

when you load the static global config and update the watch, if someone subsequently subscribes to the watch, they won't be notified that it got updated with the static config file. the solution is to immediatly clone the contents instead of just waiting for it to be updated.

TBS1996:
Envelopeprocessorservice updated the global config only when the watch changed, however, if we use static config then the watch won't change after we first set its contents to the static config. If we subscribe after it was first set, we wouldn't ever get to use the static config in the envelope processor service.

to fix this, we immediatly update the  global config in envelope processor service to the contents of the watch. 

#skip-changelog

### DIFF
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -87,7 +87,9 @@ pub struct Get;
 /// The message for receiving a watch that subscribes to the [`GlobalConfigService`].
 ///
 /// The global config service must be up and running, else the subscription
-/// fails.
+/// fails. Subscribers should use the initial value when they get the watch
+/// rather than only waiting for the watch to update, in case a global config
+///  is only updated once, such as is the case with the static config file.
 pub struct Subscribe;
 
 /// An interface to get [`GlobalConfig`]s through [`GlobalConfigService`].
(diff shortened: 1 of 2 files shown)


iker-barriocanal (approved):

jan-auer (approved):

jan-auer:
Please update the docs on `Subscribe` to mention this and add an example that shows correct usage. See this comment for inspiration on the examples:

https://github.com/getsentry/relay/pull/2458#discussion_r1322683506

olksdr (approved):

relay-server/src/actors/global_config.rs
```diff
@@ -88,7 +88,9 @@ pub struct Get;
 /// The message for receiving a watch that subscribes to the [`GlobalConfigService`].
 ///
 /// The global config service must be up and running, else the subscription
-/// fails.
+/// fails. Subscribers should use the initial value when they get the watch
+/// rather than only waiting for the watch to update, in case a global config
+///  is only updated once, such as is the case with the static config file.
```
iker-barriocanal, line 93: Should we add a code example?

PR #2506, opened 2023-09-12
feat(build): Separate images for processing and PoPs Relays

olksdr:
Updating CI to build the 2 different images for Relay processing and Relay PoP with different features specified. 

I've opted for matrix job, to build images in parallel to spend as little time as possible on them. 

related to: https://github.com/getsentry/team-ingest/issues/196

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,18 +185,20 @@ jobs:
       matrix:
         # the arm64 build takes too long, so disable for now
         arch: [amd64]
+        image_name: [relay, relay-pop]
 
-    name: Build Docker Image (${{ matrix.arch }})
+
+    name: Build Docker Image (${{ matrix.arch }}) for ${{ matrix.image_name }}
     runs-on: ubuntu-latest
 
     # Skip redundant checks for library releases
     if: "!startsWith(github.ref, 'refs/heads/release-library/')"
 
     env:
-      IMG_BASE: ghcr.io/getsentry/relay
+      IMG_BASE: ghcr.io/getsentry/${{ matrix.image_name }}
       IMG_DEPS: ghcr.io/getsentry/relay-deps:${{ matrix.arch }}
       # GITHUB_SHA in pull requests points to the merge commit
-      IMG_VERSIONED: ghcr.io/getsentry/relay:${{ github.event.pull_request.head.sha || github.sha }}
+      IMG_VERSIONED: ghcr.io/getsentry/${{ matrix.image_name }}:${{ github.event.pull_request.head.sha || github.sha }}
       ARCH: ${{ matrix.arch }}
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -294,17 +300,19 @@ jobs:
...
         run: |
           set -euxo pipefail
-          docker tag "$IMG_VERSIONED" "us.gcr.io/sentryio/relay:nightly"
-          docker push "us.gcr.io/sentryio/relay:nightly"
+          docker tag "$IMG_VERSIONED" "us.gcr.io/sentryio/${{ matrix.image_name }}:nightly"
+          docker push "us.gcr.io/sentryio/${{ matrix.image_name }}:nightly"
 
       - name: Upload gocd deployment assets
+        # Collect assets only for processing Relay, since it contains more information.
+        if: matrix.image_name == 'relay'
```
iker-barriocanal, line 315: In the future, not only images will be different but also assets, including debug symbols. Should we upload both, even if relay-pop is a duplicate for now?

olksdr: I would say it doesn't matter now and we can revisit this later. 
Since it will also require more changes, which I would better address in the followups.

iker-barriocanal (approved):

.github/workflows/ci.yml
```diff
@@ -185,18 +185,20 @@ jobs:
       matrix:
         # the arm64 build takes too long, so disable for now
         arch: [amd64]
+        image_name: [relay, relay-pop]
 
-    name: Build Docker Image (${{ matrix.arch }})
+
+    name: Build Docker Image (${{ matrix.arch }}) for ${{ matrix.image_name }}
```
jan-auer:
```suggestion
    name: Build Docker Image
```

.github/workflows/ci.yml
```diff
@@ -247,6 +249,10 @@ jobs:
     timeout-minutes: 5
     needs: build
```
jan-auer, line 250: Can we make these depend on the specific build job (relay/relay-pop) as opposed to the entire `build` matrix? Currently, it looks like this:

image

This would be especially useful for the "Sentry-Relay Integration Tests", which only need to depend on the "relay" build step, but not on "relay-pop".

jan-auer: On some quick research, it seems "[Reusable Workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-a-matrix-strategy-with-a-reusable-workflow)" would be the solution. That's too big of a change for now, so I'm fine if we stick with this.

.github/workflows/ci.yml
iker-barriocanal: If building an image fails, I think it's helpful to know what image was trying to build. We do this for [tests](https://github.com/getsentry/relay/blob/ded63aca6aa76b42df544e238bb71d57dc02712b/.github/workflows/ci.yml#L81) on macos/windows.

olksdr: now they will look like  this
image

iker-barriocanal: Oh nice. How does it work, is it documented somewhere? I couldn't find anything relevant beyond [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#name).

PR #2509, opened 2023-09-13
chore(dashboard): Move dashboard functionality to separate submodule

olksdr:
Better code organization: 
* move the dashboard-related code into separate sub-module in the logs crate
* include the submodule only when the `dashboard` feature is enabled

#skip-changelog

### DIFF
--- a/relay-log/src/dashboard.rs
+++ b/relay-log/src/dashboard.rs
@@ -1,16 +1,24 @@
+//! This module provides the functionality for subscribing to the logs stream and delivering it to
+//! the external consumers through the [`tokio::sync::broadcast`] channel.
+
 use once_cell::sync::Lazy;
 use tokio::sync::broadcast;
 use tracing::Subscriber;
 use tracing_subscriber::Layer;
 
 /// Channel to deliver logs.
-pub static LOGS: Lazy<broadcast::Sender<Vec<u8>>> = Lazy::new(|| {
+static LOGS: Lazy<broadcast::Sender<Vec<u8>>> = Lazy::new(|| {
     let (tx, _) = broadcast::channel(2000);
     tx
 });
 
+/// Returns the [`tokio::sync::broadcast::Receiver`] part of logs subscriber.
+pub fn receiver() -> broadcast::Receiver<Vec<u8>> {
+    LOGS.subscribe()
+}
+
 /// Writer for the logs out to the [`tokio::sync::broadcast`] channel.
-pub struct LogsWriter;
+struct LogsWriter;
 
 impl std::io::Write for LogsWriter {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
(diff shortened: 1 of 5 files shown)


relay-log/src/lib.rs
```diff
@@ -117,6 +117,11 @@ mod setup;
 #[cfg(feature = "init")]
 pub use setup::*;
 
+#[cfg(feature = "dashboard")]
+mod dashboard;
+#[cfg(feature = "dashboard")]
+pub use dashboard::*;
```
jan-auer: What if we make this a `pub mod`, instead?

relay-log/src/dashboard.rs
```diff
@@ -0,0 +1,43 @@
+use once_cell::sync::Lazy;
+use tokio::sync::broadcast;
+use tracing::Subscriber;
+use tracing_subscriber::Layer;
+
+/// Channel to deliver logs.
+pub static LOGS: Lazy<broadcast::Sender<Vec<u8>>> = Lazy::new(|| {
```
jan-auer: _nit_: Could we make this private and instead expose an async fn / stream / function that returns a broadcast receiver?

relay-log/src/dashboard.rs
```diff
@@ -0,0 +1,43 @@
...
+    LogsWriter
+}
+
+/// Returns the dashbaord subscriber.
+pub fn dashboard_subscriber<S>() -> impl Layer<S> + Send + Sync
+where
+    S: Subscriber + for<'a> tracing_subscriber::registry::LookupSpan<'a>,
+{
+    tracing_subscriber::fmt::layer()
+        .with_writer(make_logs_writer)
```
jan-auer:
```suggestion
        .with_writer(|| LogsWriter)
```

relay-log/src/dashboard.rs
```diff
@@ -0,0 +1,43 @@
...
+use tracing_subscriber::Layer;
+
+/// Channel to deliver logs.
+pub static LOGS: Lazy<broadcast::Sender<Vec<u8>>> = Lazy::new(|| {
+    let (tx, _) = broadcast::channel(2000);
+    tx
+});
+
+/// Writer for the logs out to the [`tokio::sync::broadcast`] channel.
+pub struct LogsWriter;
```
jan-auer:
```suggestion
struct LogsWriter;
```

relay-log/src/dashboard.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+/// Returns the log writer.
+fn make_logs_writer() -> impl std::io::Write {
+    LogsWriter
+}
+
+/// Returns the dashbaord subscriber.
+pub fn dashboard_subscriber<S>() -> impl Layer<S> + Send + Sync
```
jan-auer:
```suggestion
pub(crate) fn dashboard_subscriber<S>() -> impl Layer<S> + Send + Sync
```

jan-auer (approved):

relay-log/src/dashboard.rs
olksdr: yeah, makes sense

relay-log/src/lib.rs
olksdr: yeah, this is even better! Thanks!

PR #2510, opened 2023-09-13
chore(deploy): Deploy separate images to processing and PoPs Relays

olksdr:
Built in https://github.com/getsentry/relay/pull/2506 images must be used in the deployments. To do this following is done:
* configure debug information upload per different types of services - separately for `relay` and `relay-pop`
* configure GOCD deployment templates to use new images
* update bash scripts to follow new naming conventions 

blocked by: https://github.com/getsentry/relay/pull/2506
depends on: https://github.com/getsentry/relay/pull/2529
fix: https://github.com/getsentry/team-ingest/issues/197
fix: https://github.com/getsentry/team-ingest/issues/196

#skip-changelog


.github/workflows/ci.yml
```diff
@@ -331,7 +329,7 @@ jobs:
             /home/runner/.gsutil/tracker-files \
             /home/runner/.gsutil/tracker-files/upload_TRACKER_*.rc.zip__JSON.url \
           || true
-          gsutil -m cp -L gsutil.log ./libs.tar ./relay-debug.zip ./relay.src.zip ./release-name \
+          gsutil -m cp -L gsutil.log ./${{ matrix.image_name }}-libs.tar ./${{ matrix.image_name }}-debug.zip ./${{ matrix.image_name }}.src.zip ./${{ matrix.image_name }}-release-name \
```
jan-auer: Idea: In devinfra assets, what if instead of renaming every file, we just put them in separate folders, such as: `.../$REVISION/relay/` and `.../$REVISION/relay-internal/`?

iker-barriocanal (approved):
Have we verified this doesn't cause issues with the self-hosted release process, happening tomorrow?

nit: personally, I prefer `relay-processing` over `relay-internal` as it's easier to associate with the processing feature/mode, but no strong opinions.

gocd/templates/libs/relay.libsonnet
```diff
@@ -53,14 +53,14 @@ function(region) {
             timeout: 1200,
             elastic_profile_id: 'relay',
             tasks: [
-              gocdtasks.script(importstr '../bash/create-sentry-release.sh'),
+              gocdtasks.script(importstr '../bash/create-sentry-relay-internal-release.sh'),
```
jan-auer: To double check: Is there no way to parameterize these scripts?

jan-auer (approved):
LGTM

iker-barriocanal (approved):

jan-auer (approved):

mattgauntseo-sentry (approved):

gocd/templates/libs/relay.libsonnet
mattgauntseo-sentry: I've found using environment variables are the easiest way to pass in parameters, but there may be other options using Jsonnet (i.e. string replace https://stackoverflow.com/questions/49011412/how-to-replace-string-for-each-item-in-a-list-in-jsonnet)

PR #2511, opened 2023-09-13
feat(spans): Filter spans based on module

phacops:
We're trying to filter spans based on the `op` field as we prepare to ingest spans for GA. Due to the high cardinality of groupings, we would like to be selective on what we ingest.

At first, we'd want to ingest spans for our DB and Browser modules. If the feature is enabled for the project, we'd add the necessary conditions to match spans corresponding to `op` we're interesting in.

If no module is enabled, we ingest all spans.


relay-dynamic-config/src/defaults.rs
```diff
@@ -25,12 +28,71 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
+        span_op_conditions.push(RuleCondition::And(AndCondition {
+            inner: vec![
+                RuleCondition::Glob(GlobCondition {
+                    name: span_op_field_name.into(),
+                    value: GlobPatterns::new(vec!["db*".into()]),
+                }),
+                RuleCondition::Not(NotCondition {
+                    inner: Box::new(RuleCondition::Eq(EqCondition {
+                        name: span_op_field_name.into(),
+                        value: Value::String("db.redis".into()),
```
jan-auer: You can make this a `Value::Array`, in which case the rule will match (or NOT match since you're negating it) if any of the strings in the array match:

```rust
                RuleCondition::Not(NotCondition {
                    inner: Box::new(RuleCondition::Eq(EqCondition {
                        name: span_op_field_name.into(),
                        value: Value::Array(vec!["db.redis".into(), "db.clickhouse".into(), "db.sql.query".into()]),
                        options: Default::default(),
                    })),
                }),
```

relay-dynamic-config/src/defaults.rs
```diff
@@ -70,7 +132,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 key: key.into(),
                 field: Some(format!("span.data.{}", key.replace('.', "\\."))),
                 value: None,
-                condition: None,
+                condition: Some(additional_conditions.clone()),
```
jan-auer: You do not need to add conditions to the tag mappings. The metrics themselves are already extracted conditionally, and tag mapping only applies on the extracted mapping.

relay-dynamic-config/src/defaults.rs
```diff
@@ -25,12 +28,61 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
+        .features
+        .has(Feature::SpanMetricsExtractionBrowserModule)
+    {
+        span_op_conditions.push(RuleCondition::Glob(GlobCondition {
+            name: span_op_field_name.into(),
+            value: GlobPatterns::new(vec![
+                "ui.*".into(),
+                "browser.*".into(),
+                "http.*".into(),
+                "grouping".into(),
```
phacops: @AbhiPrasad Can you validate these are the right values for the browser module? Maybe it should be without dots?

relay-server/src/actors/processor.rs
```diff
@@ -2278,6 +2279,24 @@ impl EnvelopeProcessorService {
...
+            "db.clickhouse".into(),
+            "db.redis".into(),
+            "db.sql.query".into(),
+        ]);
+        let span_op_browser_module_prefix_allowlist: Vec<String> = vec![
+            "browser".into(),
+            "http".into(),
+            "resource".into(),
+            "ui".into(),
+        ];
```
iker-barriocanal: Since these don't change during execution, what do you think about making them `static` and wrapping them in [`OnceCell`](https://doc.rust-lang.org/stable/std/cell/struct.OnceCell.html)? See another [example in Relay](https://github.com/getsentry/relay/blob/b2186c9da45a44d5d39b48e50b8d65b7e592d347/relay-event-normalization/src/normalize/user_agent.rs#L377).

iker-barriocanal (approved):

relay-dynamic-config/src/defaults.rs
```diff
@@ -25,12 +28,77 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
+        if project_config
+            .features
+            .has(Feature::SpanMetricsExtractionBrowserModule)
+        {
+            span_op_conditions.push(RuleCondition::Glob(GlobCondition {
+                name: span_op_field_name.into(),
+                value: GlobPatterns::new(vec![
+                    "browser*".into(),
+                    "http*".into(),
+                    "resource".into(),
```
AbhiPrasad: `resource` can also be `resource.script`, and `resource.css` so I think we have to glob here as well.

AbhiPrasad (approved):

relay-dynamic-config/src/defaults.rs
AbhiPrasad: The values should as per https://develop.sentry.dev/sdk/performance/span-operations/, there are some example span ops in there as well.

relay-dynamic-config/src/defaults.rs
```diff
@@ -25,12 +28,71 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
+                        name: span_op_field_name.into(),
+                        value: GlobPatterns::new(vec!["db*".into()]),
+                    }),
+                    RuleCondition::Not(NotCondition {
+                        inner: Box::new(RuleCondition::Eq(EqCondition {
+                            name: span_op_field_name.into(),
+                            value: Value::Array(vec![
+                                "db.clickhouse".into(),
+                                "db.redis".into(),
+                                "db.sql.query".into(),
```
gggritso: @AbhiPrasad what do you think about `"db.sql.query"`? I feel like we should allow this by default, and opt out the one problematic project that came up so far

gggritso (commented):
👀 nice!

relay-dynamic-config/src/defaults.rs
AbhiPrasad: ah yeah I think we should allow it - good catch!

phacops:
I ended up simplifying into 2 groups: GA and all modules.

We'll keep the filtering in Relay for the GA case and accepts all spans for projects in the allowlist.

relay-dynamic-config/src/defaults.rs
```diff
@@ -42,7 +82,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
             category: DataCategory::Span,
             mri: "d:spans/exclusive_time_light@millisecond".into(),
             field: Some("span.exclusive_time".into()),
-            condition: None,
+            condition: span_op_conditions.clone(),
```
jan-auer: I think this clone should be redundant, since it's the last reference of this struct and can be moved.

```suggestion
            condition: span_op_conditions,
```

relay-dynamic-config/src/feature.rs
```diff
@@ -22,6 +22,12 @@ pub enum Feature {
     /// Allow ingestion of metrics in the "custom" namespace.
     #[serde(rename = "organizations:custom-metrics")]
     CustomMetrics,
+    /// Enable extracting spans for modules in GA.
+    #[serde(rename = "projects:span-metrics-extraction-ga-modules")]
+    SpanMetricsExtractionGAModules,
```
jan-auer: Keep in mind that we will have to keep these feature flags around, even if we make them the default at some point. I'd suggest some more functional naming for that, such as "db" instead of "GA".

relay-server/src/actors/processor.rs
```diff
@@ -2287,14 +2294,21 @@ impl EnvelopeProcessorService {
                 let Some(inner_span) = span.value_mut() else {
                     continue;
                 };
-                inner_span.segment_id = transaction_span.segment_id.clone();
-                inner_span.is_segment = Annotated::new(false);
-                add_span(span);
+                // HACK: filter spans based on module until we figure out grouping.
+                let Some(span_op) = inner_span.op.value() else {
+                    continue;
```
jan-auer: The span is cloned even though you continue a couple of times. Could you move the call to `span.clone()` into the `if` branch below? For example:

1. grab `let Some(inner_span) = span.value() else ...`
2. do all checks on the value and the op
3. in the branch call `inner_span.clone()`
4. make your mutations
5. call `add_span(Annotated::new(inner_span))`

jan-auer (approved):

relay-dynamic-config/src/feature.rs
phacops: I had DB at first but we'll have more modules than this as time go by. I'm not sure DB would be a better name if we end up filtering on more than those modules.

PR #2512, opened 2023-09-13
fix(spans): Add a description to the segment as well

phacops:
We were not getting any description in case the span was the segment.


shruthilayaj (approved):

jan-auer (approved):

jan-auer:
cc @jjbayer

PR #2513, opened 2023-09-13
feat(stats-detectors): Extract new metric

udameli:
The perf team wants to run a cron job every hour where we'll be fetching last 14d timeseries of p95(transaction.duration) for the short list of transaction names that we suspect are regressing, for all projects. This is a part of continuous regression detection for Statistical Detector issues. Since percentiles are expensive to calculate, the query takes some time to run (example for p50 [here](https://sentry.sentry.io/performance/sentry:5e034004a32c4945bf3e465598cdfe00/?project=1&query=query.period%3A1209600+performance.trendsv2.stats_data%3ATrue&showTransactions=p95&sort=-transaction.duration&statsPeriod=14d&transaction=%2Fapi%2F0%2Forganizations%2F%7Borganization_slug%7D%2Fevents-trends-statsv2%2F#span-81b097bf8268d706) with duration being not too far from `p95`). I’m hoping to speed up the query by materializing a new metric without extra tags that we don’t care about. All we want is to be able to fetch transaction duration for a given project for a given transaction name. SnS has helped me to run a cardinality analysis to see if dropping other tags will help reduce cardinality and it does ([JIRA ticket](https://getsentry.atlassian.net/browse/REQ3S-54)).

Backend PR: https://github.com/getsentry/sentry/pull/56202


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -324,6 +335,19 @@ impl TransactionExtractor<'_> {
             .into_metric(timestamp),
         );
 
+        // Lower cardinality duration
+        metrics.project_metrics.push(
+            TransactionMetric::Duration {
```
udameli: Should this be `TransactionMetric`?

udameli: @jan-auer I added a new metric type DurationCoarse to match the string [here](https://github.com/getsentry/sentry/pull/56202/files#diff-c795730a1d305a53dcd95ed823c1167a473790e14dc73c160375d0e8ccf6c56fR90)

relay-server/src/metrics_extraction/transactions/types.rs
```diff
@@ -62,6 +68,12 @@ impl IntoMetric for TransactionMetric {
                 MetricUnit::Duration(unit),
                 tags.into(),
             ),
+            TransactionMetric::DurationCoarse { unit, value, tags } => (
+                Cow::Borrowed("duration_coarse"),
```
udameli: open to better naming suggestions https://github.com/getsentry/sentry/pull/56202#discussion_r1330333443

jan-auer: We sadly do not have a naming scheme for such metrics yet, and likely over time they will rather become a materialization in storage and would be fully hidden in the namespace.

`duration_coarse` sounds good to me. It would allow us to add other uncontroversial tags without introducing a new metric every time. 

@jjbayer any thoughts on this?

jan-auer (approved):
Thanks for providing this PR and the thorough reasoning! 
This is good to merge once you've settled on final naming and tests have been updated.

relay-server/src/metrics_extraction/transactions/types.rs
jjbayer: No strong preference, but we do already have a similar metric for spans that we suffixed with `_light` instead of `_coarse`, so I would use the same suffix for consistency: https://github.com/getsentry/relay/blob/4d0650ee8a2cfa70adaef2a51774f266857ab367/relay-dynamic-config/src/defaults.rs#L100

jjbayer (approved):
Please make sure to notify SnS before this goes live. `transaction` is still a high-cardinality tag so the increase in the number of buckets we send will be noticeable.

udameli:
not sure if there's something wrong with my local set up but when I try running `cargo insta test` I'm seeing this error:
```
error: couldn't read relay-ua/src/../uap-core/regexes.yaml: No such file or directory (os error 2)
  --> relay-ua/src/lib.rs:19:22
   |
19 |     let ua_regexes = include_bytes!("../uap-core/regexes.yaml");
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)

   Compiling erased-serde v0.3.26
   ```
  
Could I get some help updating the snapshots please?

jjbayer:
@udameli `uap-core` is a submodule, could you run `git submodule update --init --recursive` and then try `cargo insta test` again?

PR #2514, opened 2023-09-14
ref(sampling): Refactor dynamic sampling

TBS1996:
* Got rid of the old style of SamplingResult, now we save the entire sampling match along with a flag of whether it's kept or not. 
* SamplingResult now has a `Pending` variant which suggests that dynamic sampling has yet to be run. The alternative, Option<SamplingResult>, was too vague for my taste. 
* Introduced `SamplingEvaluator`, a state machine for matching sampling rules.
* Removing a lot of responsibilities of the sampling crate, in order to make it more generic and reusable.
* More general refactoring, including a lot of simplified API for functions.
* Moved dynamic sampling tests from `lib.rs` to wherever the functionality they test resides.
* Deleted many tests that relied on the old APIs, or tested responsibilities that the sampling crate no longer has.
* Created many new tests.

#skip-changelog


relay-sampling/src/config.rs
```diff
@@ -73,6 +73,42 @@ impl SamplingRule {
     fn supported(&self) -> bool {
         self.condition.supported() && self.ty != RuleType::Unsupported
     }
+
+    /// Returns the sample rate if the rule is active.
+    pub fn sample_rate(&self, now: DateTime<Utc>) -> Option<f64> {
```
TBS1996: This replaced the `SamplingValueEvaluator`.

relay-sampling/src/config.rs
```diff
@@ -73,6 +73,41 @@ impl SamplingRule {
...
+        let sampling_base_value = self.sampling_value.value();
+
+        match self.decaying_fn {
```
TBS1996: inlined from: https://github.com/getsentry/relay/blob/cb88f1227eefdad162430f232557b2f98db58f9d/relay-sampling/src/evaluation.rs#L319-L341

relay-sampling/src/dsc.rs
```diff
@@ -111,8 +111,7 @@ impl DynamicSamplingContext {
 
     /// Compute the effective sampling rate based on the random "diceroll" and the sample rate from
     /// the matching rule.
-    pub fn adjusted_sample_rate(&self, rule_sample_rate: f64) -> f64 {
-        let client_sample_rate = self.sample_rate.unwrap_or(1.0);
+    pub fn adjusted_sample_rate(client_sample_rate: f64, rule_sample_rate: f64) -> f64 {
```
TBS1996: i found it ugly to pass in the entire sampling context just to get the `client_sample_rate` here: https://github.com/getsentry/relay/pull/2514/files#diff-c36d16143aa181bfcd1fca3ca1fba4f23ee13c549c08083ba67895bbb815436fR286

relay-sampling/src/evaluation.rs
```diff
@@ -70,22 +67,76 @@ fn check_unsupported_rules(
     Ok(())
 }
 
-/// Gets the sampling match result by creating the merged configuration and matching it against
-/// the sampling configuration.
-pub fn merge_configs_and_match(
```
TBS1996, line 75: this functionality is now in `fn match_rules`

relay-sampling/src/evaluation.rs
```diff
@@ -95,139 +114,193 @@ pub fn merge_configs_and_match(
     // The determination of the sampling mode occurs with the following priority:
     // 1. Non-root project sampling mode
     // 2. Root project sampling mode
-    let Some(primary_config) = sampling_config.or(root_sampling_config) else {
-        relay_log::error!("cannot sample without at least one sampling config");
```
TBS1996, line 99: previously this was unreachable as a lack of either config would have resulted in no rules, thereby no matches, which would have returned early here: https://github.com/getsentry/relay/blob/cb88f1227eefdad162430f232557b2f98db58f9d/relay-sampling/src/evaluation.rs#L88 

I changed it from error to info so not to fail all the CI tests that dont pass in any configs.

relay-sampling/src/evaluation.rs
```diff
@@ -95,139 +114,193 @@ pub fn merge_configs_and_match(
...
-            None => match_result.sample_rate,
-        },
-        SamplingMode::Unsupported => {
+    let adjust_sample_rate = match sampling_config
+        .or(root_sampling_config)
+        .map(|config| config.mode)
+    {
+        Some(SamplingMode::Total) => true,
+        Some(SamplingMode::Received) => false,
+        Some(SamplingMode::Unsupported) => {
```
TBS1996: dealing with unsupported here simplifies the `get_sampling_match` function

relay-sampling/src/evaluation.rs
```diff
@@ -95,139 +114,193 @@ pub fn merge_configs_and_match(
...
 
 /// Represents the specification for sampling an incoming event.
-#[derive(Clone, Debug, PartialEq, Serialize)]
-pub struct SamplingMatch {
-    /// The sample rate to use for the incoming event.
-    pub sample_rate: f64,
+#[derive(Default, Clone, Debug, PartialEq)]
+pub enum SamplingResult {
+    /// The event matched a sampling condition.
+    Match {
```
TBS1996: @jan-auer  had 2nd thoughts on creating a named struct here, where it's used the anonymous one seems just as good

relay-sampling/src/evaluation.rs
```diff
@@ -95,139 +114,193 @@ pub fn merge_configs_and_match(
...
+/// it finds a sample rate rule. Once a sample rate rule is found, the final sample rate is
+/// computed by multiplying it with the previously accumulated factors.
+///
+/// The default accumulated factors equal to 1 because it is the identity of the multiplication
+/// operation, thus in case no factor rules are matched, the final result will just be the
+/// sample rate of the matching rule.
+///
+/// In case no sample rate rule is matched, we are going to return a None, signaling that no
+/// match has been found.
+pub(crate) fn get_sampling_match<'a>(
```
TBS1996: pub(crate) only due to test, i can look into moving stuff around to get rid of it

relay-sampling/src/evaluation.rs
```diff
@@ -95,139 +114,193 @@ pub fn merge_configs_and_match(
...
+                        };
                     }
                 }
             }
         }
-
-        // In case no match is available, we won't return any specification.
-        None
     }
+
```
TBS1996: question: if there are no `SamplingValue::SampleRate` in the rules, should we log that?

relay-sampling/src/evaluation.rs
```diff
@@ -95,139 +114,193 @@ pub fn merge_configs_and_match(
...
+                        let Some(seed) = seed else {
+                            return SamplingResult::NoMatch;
+                        };
+
+                        let is_kept = sampling_match(sample_rate, seed);
+
+                        return SamplingResult::Match {
+                            sample_rate,
+                            seed,
+                            matched_rules: MatchedRuleIds(matched_rule_ids),
```
TBS1996: thoughts on making a constructor where you pass in the sample_rate, seed, and matched rules, and it will generate the is_kept value? 

its what i did in `sampling_match_from_parts` that i used for testing: https://github.com/getsentry/relay/pull/2514/files#diff-266da88d9f1e67f709b8b4286a60a6bb63ac51cf982ba7c3c46af957a9c6a123R117

relay-sampling/src/lib.rs
```diff
@@ -108,40 +108,48 @@ mod tests {
...
     };
     use crate::dsc::TraceUserContext;
-    use crate::evaluation::{merge_configs_and_match, MatchedRuleIds, SamplingMatch};
+    use crate::evaluation::{
+        get_sampling_match, match_rules, sampling_match, MatchedRuleIds, SamplingResult,
+    };
 
     use super::*;
 
-    macro_rules! assert_transaction_match {
```
TBS1996, line 115: I found the use of macros a bit unnecessary and harder to maintain

relay-sampling/src/lib.rs
```diff
@@ -108,40 +108,48 @@ mod tests {
...
-            assert_eq!(
-                $res,
-                Some(SamplingMatch {
-                    sample_rate: $sr,
-                    seed: $sd.trace_id,
-                    matched_rule_ids: MatchedRuleIds(vec![$(RuleId($id),)*])
-                })
-            )
-        }
+    fn transaction_match(
```
TBS1996: These are now just helpers to construct transaction and trace SamplingResults, i found it clearer to do it like this and then use the normal assert_eq in the tests

relay-sampling/src/lib.rs
```diff
@@ -205,26 +213,19 @@ mod tests {
...
 
-    fn mocked_dynamic_sampling_context(
-        transaction: &str,
-        release: &str,
-        environment: &str,
-        user_segment: &str,
-        user_id: &str,
-        replay_id: Option<Uuid>,
-    ) -> DynamicSamplingContext {
+    fn mocked_dsc(environment: &str, user_segment: &str, release: &str) -> DynamicSamplingContext {
```
TBS1996: removed the fields that weren't changing in any of the tests. 

shorter names to ensure calling them in one line for easier readability

relay-sampling/src/lib.rs
```diff
@@ -355,8 +356,10 @@ mod tests {
         event: &Event,
         dsc: &DynamicSamplingContext,
         now: DateTime<Utc>,
-    ) -> Option<SamplingMatch> {
-        SamplingMatch::match_against_rules(config.rules_v2.iter(), Some(event), Some(dsc), now)
```
TBS1996, line 359: so here the tests was (and still is) bypassing the rule verification. I'm not sure if this is intended behaviour or an oversight. You could say its more unit-y but, it also is less relevant.

relay-server/src/actors/processor.rs
```diff
@@ -1363,7 +1359,7 @@ impl EnvelopeProcessorService {
             event_metrics_extracted: false,
             metrics: Metrics::default(),
             sample_rates: None,
-            sampling_result: SamplingResult::Keep,
+            sampling_result: SamplingResult::NoMatch,
```
TBS1996: equivalent to Keep, although im not a big fan of how implicit this seems here

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -1,116 +1,20 @@
...
-        // requires it.
-        Utc::now(),
-    );
-    SamplingResult::determine_from_sampling_match(sampling_result)
-}
-
 /// Runs dynamic sampling if the dsc and root project state are not None and returns whether the
 /// transactions received with such dsc and project state would be kept or dropped by dynamic
 /// sampling.
 pub fn is_trace_fully_sampled(
```
TBS1996, line 64: not a big fan of this function, but we agreed to keep it, maybe we can revisit it later

relay-sampling/src/config.rs
jan-auer: This is a very reasonable change, thanks! This functionality makes more sense on the types rather than inside the evaluator. Would it make sense to leave a few more words that this also interpolates sample rates based on the decaying function?

Reading through the implementation, however, it's a bit unintuitive why this returns a float and not the typed value. The float could be either a factor or a sample rate, which the caller now always must check separately. I think it would make sense to return `Option<SamplingValue>` here, instead.

relay-sampling/src/config.rs
```diff
@@ -73,6 +73,41 @@ impl SamplingRule {
...
+                        let now_timestamp = now.timestamp() as f64;
+                        let start_timestamp = start.timestamp() as f64;
+                        let end_timestamp = end.timestamp() as f64;
+                        let progress_ratio = ((now_timestamp - start_timestamp)
+                            / (end_timestamp - start_timestamp))
+                            .clamp(0.0, 1.0);
+
+                        // This interval will always be < 0.
+                        let interval = decayed_value - sampling_base_value;
+                        return Some(sampling_base_value + (interval * progress_ratio));
```
jan-auer: While we're at it, what if we make this a bit more readable?

```suggestion
                        let now = now.timestamp() as f64;
                        let start = start.timestamp() as f64;
                        let end = end.timestamp() as f64;
                        let ratio = ((now - start) / (end - start)).clamp(0.0, 1.0)

                        // This interval will always be < 0.
                        let interval = decayed_value - sampling_base_value;
                        return Some(sampling_base_value + (interval * ratio));
```

relay-sampling/src/dsc.rs
```diff
@@ -111,8 +111,7 @@ impl DynamicSamplingContext {
 
     /// Compute the effective sampling rate based on the random "diceroll" and the sample rate from
     /// the matching rule.
-    pub fn adjusted_sample_rate(&self, rule_sample_rate: f64) -> f64 {
-        let client_sample_rate = self.sample_rate.unwrap_or(1.0);
+    pub fn adjust_sample_rate(client_sample_rate: f64, rule_sample_rate: f64) -> f64 {
```
jan-auer: This was actually an oversight of mine. Let's move this function off of the DSC type entirely.  This can be a private function inside `evaluation.rs`.

relay-server/src/actors/processor.rs
jan-auer: You mentioned that you didn't want `Option<SamplingMatch>` here because that would indicate dynamic sampling hasn't run yet. Yet, we have a default value anyway. Would you like to follow through, and either make this an `Option` or remove the `SamplingResult` type? Right now, this is an in-between state.

relay-server/src/actors/processor.rs
```diff
@@ -42,7 +42,7 @@ use relay_protocol::{Annotated, Array, Empty, FromValue, Object, Value};
 use relay_quotas::{DataCategory, ReasonCode};
 use relay_redis::RedisPool;
 use relay_replays::recording::RecordingScrubber;
-use relay_sampling::evaluation::MatchedRuleIds;
+use relay_sampling::evaluation::{match_rules, MatchedRuleIds, SamplingResult};
```
jan-auer: Instead of importing the function, import the module or fully qualify the function call at call site as `relay_sampling::evaluation::match_rules(...)`.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -119,20 +23,17 @@ pub fn is_trace_fully_sampled(
...
+        match_rules(
+            processing_enabled,
+            None,
+            root_project_state.config.dynamic_sampling.as_ref(),
+            None,
+            Some(dsc),
+            Utc::now(),
+        )
+        .should_keep(),
+    )
```
jan-auer:
```suggestion
    let sampling_match = relay_sampling::evaluation::match_rules(
        processing_enabled,
        None,
        root_project_state.config.dynamic_sampling.as_ref(),
        None,
        Some(dsc),
        Utc::now(),
    );
        
    Some(sampling_match.should_keep())
```

relay-sampling/src/evaluation.rs
jan-auer: It's less convenient this way. You can see the effects in usage sites even in this PR already. Named structs can have their own methods, can be passed to functions, can have private fields (which this one should) and can be matched with a name. 

For instance, instead of writing the following:

```rust
match result {
  SamplingResult::Match {
    is_kept,
    matched_rules,
    ..
  } if is_kept => { ... }
}
```

... one can write this:

```rust
match result {
  SamplingResult::Match(m) if m.should_keep() => handle_match(m),
  ...
}
```

Note how in the first example it would be impossible to write a function that takes the match. Either you'd have to pass all fields, or you'd have to match again in the function and unwrap.

Additionally, I'd ask to make all fields private and add convenience methods.

relay-sampling/src/evaluation.rs
```diff
@@ -70,164 +67,229 @@ fn check_unsupported_rules(
...
+            return Some(*sample_rate);
+        }
+        None
+    }
 
-    /// The list of rule ids that have matched the incoming event and/or dynamic sampling context.
-    pub matched_rule_ids: MatchedRuleIds,
-}
+    /// Returns true if the event matched on any rules.
+    pub fn no_match(&self) -> bool {
```
jan-auer:
```suggestion
    pub fn is_no_match(&self) -> bool {
```

And consider to introduce parity via `is_match`.

jan-auer (commented):
Preliminary review.

relay-sampling/src/evaluation.rs
TBS1996: update: got rid of it by moving the tests

relay-sampling/src/evaluation.rs
TBS1996: done :)

relay-sampling/src/config.rs
TBS1996: thanks!

relay-sampling/src/condition.rs
TBS1996, line 1: Moved here all the tests concerning `RuleCondition`

relay-sampling/src/config.rs
```diff
@@ -36,6 +36,13 @@ impl SamplingConfig {
     pub fn unsupported(&self) -> bool {
         !self.rules_v2.iter().all(SamplingRule::supported)
     }
+
+    /// Filters the sampling rules by the given [`RuleType`].
+    pub fn filter_rules(&self, rule_type: RuleType) -> impl Iterator<Item = &SamplingRule> {
```
TBS1996, line 41: @jan: moved from `ProjectState` because it was simply calling the SamplingConfig anyway, and it makes more sense to have it here based on where im calling it

relay-sampling/src/condition.rs
TBS1996, line 1: moved all the tests that are purely about the `RuleCondition` logic to here

relay-sampling/src/dsc.rs
```diff
@@ -108,33 +108,6 @@ impl DynamicSamplingContext {
             other: Default::default(),
         })
     }
-
-    /// Compute the effective sampling rate based on the random "diceroll" and the sample rate from
-    /// the matching rule.
-    pub fn adjusted_sample_rate(&self, rule_sample_rate: f64) -> f64 {
```
TBS1996, line 114: Moved: it's now a method of `DynamicSamplingEvaluator`.

relay-sampling/src/dsc.rs
```diff
@@ -295,37 +268,6 @@ where
 mod tests {
     use super::*;
 
-    #[test]
-    fn test_adjust_sample_rate() {
```
TBS1996, line 299: moved to `evaluation.rs`

relay-sampling/src/evaluation.rs
```diff
@@ -7,226 +7,205 @@ use chrono::{DateTime, Utc};
...
     let dist = Uniform::new(0f64, 1f64);
     generator.sample(dist)
 }
 
-/// Returns an iterator of references that chains together and merges rules.
+/// State of the [`SamplingEvaluator`]'s rule matching.
 ///
-/// The chaining logic will take all the non-trace rules from the project and all the trace/unsupported
-/// rules from the root project and concatenate them.
-pub fn merge_rules_from_configs<'a>(
```
TBS1996, line 35: the selection of rules is no longer a concern of `dynamic-sampling` crate.

relay-sampling/src/evaluation.rs
```diff
@@ -7,226 +7,205 @@ use chrono::{DateTime, Utc};
...
-
-    let trace_rules = root_sampling_config
-        .into_iter()
-        .flat_map(|config| config.rules_v2.iter())
-        .filter(|&rule| rule.ty == RuleType::Trace);
-
-    transaction_rules.chain(trace_rules)
+/// Enforces a pattern to avoid matching on more rules if a match has already been found.
+#[derive(Debug)]
+pub enum RuleMatchingState {
```
TBS1996: jan: we discussed something like this recently. 

my thinking: 

1. It's now impossible to call `match_rules` if a match has already been found. 
2. We need the `rule_ids` in `SamplingMatch`, previous iteration we used std::mem::take which probably avoids allocation, however it leaves `SamplingEvaluator` in an invalid state. If we cloned it instead, we would have an unnecessary allocation. this implementaiton solves both issues, as we can do a partial move from `SamplingEvaluator` since it gets destroyed when the enum switches. 

inspect the `match_rules` and its callers to see what i mean

relay-sampling/src/evaluation.rs
```diff
@@ -7,226 +7,205 @@ use chrono::{DateTime, Utc};
...
+    /// No match has been found.
+    Evaluator(SamplingEvaluator),
+    /// Rule(s) have been matched.
+    SamplingMatch(SamplingMatch),
 }
 
-/// Checks whether unsupported rules result in a direct keep of the event or depending on the
-/// type of Relay an ignore of unsupported rules.
-fn check_unsupported_rules(
```
TBS1996, line 54: Ensuring the rules are supported is no longer a concern of this crate.

relay-sampling/src/evaluation.rs
```diff
@@ -7,226 +7,205 @@ use chrono::{DateTime, Utc};
...
+
+                    return RuleMatchingState::SamplingMatch(SamplingMatch::new(
+                        self.adjusted_sample_rate(sample_rate),
+                        seed,
+                        self.rule_ids,
+                    ));
+                }
             }
+        }
+        RuleMatchingState::Evaluator(self)
```
TBS1996: if i understand it right, this kinda pattern doesnt make new allocations, but please let me know if thats not the case

relay-sampling/src/evaluation.rs
```diff
@@ -7,226 +7,205 @@ use chrono::{DateTime, Utc};
...
-    /// it finds a sample rate rule. Once a sample rate rule is found, the final sample rate is
-    /// computed by multiplying it with the previously accumulated factors.
-    ///
-    /// The default accumulated factors equal to 1 because it is the identity of the multiplication
-    /// operation, thus in case no factor rules are matched, the final result will just be the
-    /// sample rate of the matching rule.
-    ///
-    /// In case no sample rate rule is matched, we are going to return a None, signaling that no
-    /// match has been found.
-    pub fn match_against_rules<'a, I>(
```
TBS1996, line 158: This functionality is now in `DynamicSamplingEvaluator::match_rules`.

relay-sampling/src/evaluation.rs
```diff
@@ -7,226 +7,205 @@ use chrono::{DateTime, Utc};
...
-                    Some(EventType::Transaction) => rule.condition.matches(event),
-                    _ => false,
-                }),
-                _ => false,
-            };
+    /// Returns the matched rules for the sampling match.
+    ///
+    /// Takes ownership, useful if you don't need the [`SamplingMatch`] anymore
+    /// and you want to avoid allocations.
+    pub fn take_matched_rules(self) -> MatchedRuleIds {
```
TBS1996: using it for partial move in `sample_envelope`.

relay-sampling/src/evaluation.rs
```diff
@@ -265,86 +244,255 @@ impl fmt::Display for MatchedRuleIds {
         Ok(())
     }
 }
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
 
-/// A struct representing the evaluation context of a sample rate.
-#[derive(Debug, Clone, Copy)]
-enum SamplingValueEvaluator {
```
TBS1996, line 271: this type and all its methods are now simply replaced by `fn sample_rate()` method on `SamplingRule`

relay-sampling/src/evaluation.rs
```diff
@@ -391,107 +539,13 @@ mod tests {
...
-                mode: SamplingMode::Received,
-            }),
-        )
-        .cloned()
-        .collect()
-    }
-
     #[test]
-    /// Tests the merged config of the two configs with rules.
-    fn test_get_merged_config_with_rules_in_both_project_config_and_root_project_config() {
```
TBS1996, line 439: removed: merging configs is no longer a feature of this crate

relay-sampling/src/lib.rs
TBS1996, line 1: I moved all the tests to the modules where the thing theyre testing is being used. 

I kept the rulecondition helpers, since they were used in multiple modules.

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-    }
-
-    fn add_sampling_rule_to_config(
-        sampling_config: &mut SamplingConfig,
-        sampling_rule: SamplingRule,
-    ) {
-        sampling_config.rules_v2.push(sampling_rule);
-    }
-
-    fn match_against_rules(
```
TBS1996, line 353: moved to `condition.rs`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-
-        for (rule_test_name, condition) in conditions.iter() {
-            let failure_name = format!("Failed on test: '{rule_test_name}'!!!");
-            assert!(condition.matches(&dsc), "{failure_name}");
-        }
-    }
-
-    #[test]
-    /// test matching for various rules
-    fn test_matches_events() {
```
TBS1996, line 476: replaced by `test_condition_matching` in `evaluation.rs`. 
 
I no longer test both dsc and event here as this crate is supposed to be agnostic to different types that implement `Getter`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-        };
-
-        for (rule_test_name, condition) in conditions.iter() {
-            let failure_name = format!("Failed on test: '{rule_test_name}'!!!");
-            assert!(condition.matches(&evt), "{failure_name}");
-        }
-    }
-
-    #[test]
-    fn test_or_combinator() {
```
TBS1996, line 528: moved to `condition.rs`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-        };
-
-        for (rule_test_name, expected, condition) in conditions.iter() {
-            let failure_name = format!("Failed on test: '{rule_test_name}'!!!");
-            assert!(condition.matches(&dsc) == *expected, "{failure_name}");
-        }
-    }
-
-    #[test]
-    fn test_and_combinator() {
```
TBS1996, line 588: moved to `condition.rs`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-        };
-
-        for (rule_test_name, expected, condition) in conditions.iter() {
-            let failure_name = format!("Failed on test: '{rule_test_name}'!!!");
-            assert!(condition.matches(&dsc) == *expected, "{failure_name}");
-        }
-    }
-
-    #[test]
-    fn test_not_combinator() {
```
TBS1996, line 648: moved to `condition.rs`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-
-        for (rule_test_name, expected, condition) in conditions.iter() {
-            let failure_name = format!("Failed on test: '{rule_test_name}'!!!");
-            assert!(condition.matches(&dsc) == *expected, "{failure_name}");
-        }
-    }
-
-    #[test]
-    /// test various rules that do not match
-    fn test_does_not_match() {
```
TBS1996, line 686: moved to `condition.rs`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-        };
-
-        for (rule_test_name, condition) in conditions.iter() {
-            let failure_name = format!("Failed on test: '{rule_test_name}'!!!");
-            assert!(!condition.matches(&dsc), "{failure_name}");
-        }
-    }
-
-    #[test]
-    fn test_partial_trace_matches() {
```
TBS1996, line 745: removed because I don't want to test specific Getters in this crate, maybe i should keep it if we don't want to move `DynamicSamplingContext` ?

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-
-        assert!(
-            condition.matches(&dsc),
-            "did not match with missing release, user segment, environment and transaction"
-        );
-    }
-
-    #[test]
-    /// test that the multi-matching returns none in case there is no match.
-    fn test_multi_matching_with_transaction_event_non_decaying_rules_and_no_match() {
```
TBS1996, line 864: removed because its already tested in     `fn test_get_sampling_match_result_with_no_match()`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-                None,
-            ),
-            Utc::now(),
-        );
-        assert_eq!(result, None, "did not return none for no match");
-    }
-
-    #[test]
-    /// Tests that the multi-matching works for a mixture of trace and transaction rules with interleaved strategies.
-    fn test_match_against_rules_with_multiple_event_types_non_decaying_rules_and_matches() {
```
TBS1996, line 903: Removed because this crate no longer cares about event types

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-            "user-id",
-            None,
-        );
-        let result = match_against_rules(&config, &event, &dsc, Utc::now());
-        assert_trace_match!(result, 0.03, dsc, 5, 6);
-    }
-
-    #[test]
-    /// Test that the multi-matching works for a mixture of decaying and non-decaying rules.
-    fn test_match_against_rules_with_trace_event_type_decaying_rules_and_matches() {
```
TBS1996, line 1042: removed because this crate no longer cares about event types

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-            &event,
-            &dsc,
-            Utc.with_ymd_and_hms(1970, 10, 11, 0, 0, 0).unwrap(),
-        );
-        assert_trace_match!(result, 0.02, dsc, 3);
-    }
-
-    #[test]
-    /// test that the correct match is performed when replay id is present in the dsc.
-    fn test_sampling_match_with_trace_replay_id() {
```
TBS1996, line 1148: Removed because it's testing behaviour to a specific thing that implements `Getter` which is not the domain of this crate.

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-            &event,
-            &dsc,
-            Utc::now(),
-        );
-        assert_trace_match!(result, 1.0, dsc, 1)
-    }
-
-    #[test]
-    /// Tests that no match is done when there are no matching rules.
-    fn test_get_sampling_match_result_with_no_match() {
```
TBS1996, line 1177: removed because its already tested in     `fn test_get_sampling_match_result_with_no_match()`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-            None,
-            Some(&event),
-            Utc::now(),
-        );
-        assert_no_match!(result);
-    }
-
-    #[test]
-    /// Tests that matching is still done on the transaction rules in case trace params are invalid.
-    fn test_get_sampling_match_result_with_invalid_trace_params() {
```
TBS1996, line 1194: Removed because this crate no longer cares about specific rule types

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-            Some(&dsc),
-            Some(&event),
-            Utc::now(),
-        );
-        assert_transaction_match!(result, 0.1, event, 1);
-    }
-
-    #[test]
-    /// Tests that a match with early return is done in the project sampling config.
-    fn test_get_sampling_match_result_with_project_config_match() {
```
TBS1996, line 1225: removed tests using configs as input, as we now serve rules directly and I have tests for that.

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-            Some(&event),
-            Utc::now(),
-        );
-        assert_transaction_match!(result, 0.625, event, 3);
-    }
-
-    #[test]
-    /// Tests that the correct match is raised in case we have unsupported rules with processing both
-    /// enabled and disabled.
-    fn test_get_sampling_match_result_with_unsupported_rules() {
```
TBS1996, line 1323: unsupported rules is no longer a concern of this crate

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-            None,
-            Some(&event),
-            Utc::now(),
-        );
-        assert_transaction_match!(result, 0.1, event, 1);
-    }
-
-    #[test]
-    /// Tests that match is returned with sample rate value interpolated with linear decaying function.
-    fn test_get_sampling_match_result_with_linear_decaying_function() {
```
TBS1996, line 1415: replaced by `fn test_linear_decaying_rule()`

relay-sampling/src/lib.rs
```diff
@@ -84,1522 +84,39 @@ pub use dsc::DynamicSamplingContext;
...
-            mode: SamplingMode::Received,
-        };
-        let result =
-            merge_configs_and_match(true, Some(&sampling_config), None, None, Some(&event), now);
-        assert_no_match!(result);
-    }
-
-    #[test]
-    /// Tests that no match is returned when the linear decaying function has invalid time range.
-    fn test_get_sampling_match_result_with_linear_decaying_function_and_invalid_time_range() {
```
TBS1996, line 1467: replaced by `fn test_linear_decaying_rule()`

relay-server/src/actors/processor.rs
```diff
@@ -3116,73 +3186,53 @@ mod tests {
...
-        let (outcome_aggregator, test_store) = services();
-
-        // an empty json still produces a valid config
-        let json_config = serde_json::json!({});
-
-        let config = Config::from_json_value(json_config).unwrap();
-
-        let service = create_test_processor(config);
-
+    #[test]
```
TBS1996, line 3212: changing the signature of `compute_sampling_decision` made this test a lot simpler

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -1,116 +1,73 @@
...
     #[default]
-    Keep,
-
-    /// Drop the event, due to a list of rules with provided identifiers.
-    Drop(MatchedRuleIds),
+    Pending,
 }
 
 impl SamplingResult {
-    fn determine_from_sampling_match(sampling_match: Option<SamplingMatch>) -> Self {
```
TBS1996, line 26: moved to `fn sampling_match` which is called in the constructor of `SamplingMatch`

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -119,20 +76,26 @@ pub fn is_trace_fully_sampled(
...
-    let sampled = match sampling_result {
-        SamplingResult::Keep => true,
-        SamplingResult::Drop(_) => false,
+    if root_project_config.unsupported() {
+        if processing_enabled {
+            relay_log::error!("found unsupported rules even as processing relay");
+        } else {
+            return Some(true);
+        }
+    }
```
TBS1996, line 83: since we no longer check unsupported rules in sampling config, we have to do it here instead, this code replaces: https://github.com/getsentry/relay/blob/4d0650ee8a2cfa70adaef2a51774f266857ab367/relay-sampling/src/evaluation.rs#L54

here it returns `Some(true)` because before this PR, in `fn check_unsupported_rules`, it would return an `Err`, which would return `Option<SamplingMatch>::None` in `fn merge_configs_and_match`, which would turn into a `SamplingResult::Keep`, which would turn into a `true` in this function.

relay-server/src/lib.rs
```diff
@@ -310,3 +310,20 @@ pub fn run(config: Config) -> anyhow::Result<()> {
...
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use relay_base_schema::events::EventType;
+    use relay_event_schema::protocol::{Event, EventId, LenientString};
+    use relay_protocol::Annotated;
+
+    pub fn mocked_event(event_type: EventType, transaction: &str, release: &str) -> Event {
```
TBS1996: Same function was used multiple test modules so i did this to reduce duplication

relay-server/src/actors/processor.rs
```diff
@@ -2358,14 +2364,68 @@ impl EnvelopeProcessorService {
...
+        sampling_config: Option<&SamplingConfig>,
+        event: Option<&Event>,
+        root_sampling_config: Option<&SamplingConfig>,
+        dsc: Option<&DynamicSamplingContext>,
+    ) -> SamplingResult {
+        if (sampling_config.is_none() || event.is_none())
+            && (root_sampling_config.is_none() || dsc.is_none())
+        {
+            return SamplingResult::NoMatch;
+        }
```
TBS1996, line 2378: Early return.

I don't like that the type system doesn't represent how `sampling_config` and `event` is related. and how `root_sampling_config` and `dsc` is connected.

Perhaps some new struct which encapsulates both the configuration and rules, and the instance to match on. Would be a separate PR of course.

relay-server/src/actors/processor.rs
```diff
@@ -2358,14 +2364,68 @@ impl EnvelopeProcessorService {
...
     /// Computes the sampling decision on the incoming transaction.
-    fn compute_sampling_decision(&self, state: &mut ProcessEnvelopeState) {
-        state.sampling_result = utils::get_sampling_result(
-            self.inner.config.processing_enabled(),
-            Some(&state.project_state),
-            state.sampling_project_state.as_deref(),
-            state.envelope().dsc(),
-            state.event.value(),
-        );
+    fn compute_sampling_decision(
```
TBS1996, line 2367: removed the `&self` because it was only used to check if processing is enabled, which also made it way easier to unit test. 

Not sure if it should still be under `EnvelopeProcessorService` though, how about we move it to `utils` like the `utils::get_sampling_result`?

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -216,149 +167,146 @@ mod tests {
...
 
-        let result = get_sampling_result(true, Some(&project_state), None, None, Some(&event));
-        assert_eq!(result, SamplingResult::Keep)
+        assert!(result.is_no_match());
+        assert!(result.should_keep());
     }
 
     #[test]
-    /// Tests that an event is kept when there are unsupported rules with no processing and vice versa.
-    fn test_get_sampling_result_return_keep_with_unsupported_rule() {
```
TBS1996, line 271: replaced by `fn test_is_trace_fully_sampled_return_true_with_unsupported_rules()` and `fn test_matching_with_unsupported_rule`

relay-sampling/src/condition.rs
```diff
@@ -411,4 +438,263 @@ mod tests {
...
+            ),
+            (
+                "undefined environments",
+                and(vec![
+                    glob("trace.release", &["1.1.1"]),
+                    eq("trace.user.segment", &["vip"], true),
+                ]),
+            ),
+            ("match no conditions", RuleCondition::all()),
+        ];
```
jjbayer, line 530: Should we split this into individual test cases? See [internal guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#a93e1c6cb4424f64bef2485e543b2ac8).

relay-sampling/src/evaluation.rs
```diff
@@ -7,226 +7,205 @@ use chrono::{DateTime, Utc};
...
+    /// Returns true if a rule has matched.
+    pub fn is_match(&self) -> bool {
+        matches!(self, &Self::SamplingMatch(_))
     }
 
-    Ok(())
+    /// Returns true if no rule have matched.
+    pub fn is_no_match(&self) -> bool {
+        !self.is_match()
+    }
```
jjbayer, line 54: IMO there's no benefit in writing `if foo.is_no_match()` over `! foo.is_match()`.
```suggestion
```

jjbayer (commented):
LGTM, though a refactor like this makes it hard to catch logical errors. I'll leave final review to @jan-auer.

relay-sampling/src/condition.rs
TBS1996: I agree yeah, this test i just moved so I didn't create it, but I might as well change it since i changed so many tests already

relay-sampling/src/evaluation.rs
TBS1996: hm, I took inspiration from `is_some()`/`is_none()` from std, id say it makes it a little more readable but i see your point too

relay-sampling/src/config.rs
jan-auer: Looks good. Do you no longer need to access rules directly from the project state in the processor?

relay-sampling/src/condition.rs
```diff
@@ -282,8 +282,35 @@ impl RuleCondition {
...
 #[cfg(test)]
 mod tests {
+    use std::collections::BTreeMap;
+
+    use relay_base_schema::project::ProjectKey;
+    use uuid::Uuid;
+
+    use crate::dsc::TraceUserContext;
+    use crate::tests::{and, eq, glob, not, or};
+    use crate::DynamicSamplingContext;
```
jan-auer, line 292: _NB:_ it's great you're moving these tests now. I hadn't moved them previously since they depend on the DSC, and these rules should be independent of implementations of the `Getter` trait. 

In a follow-up, we'll move this module into another crate and then also update the tests.

relay-sampling/src/condition.rs
jan-auer: Let's keep it for this PR, we can come back to these tests in a follow-up. See my previous comment, too.

relay-sampling/src/evaluation.rs
jan-auer: _nit_: We can shorten the naming of this enum and its variants and introduce more semantics:

```rust
pub enum Evaluation {
    /// Matching has not completed and more rules can be evaluated.
    ///
    /// This can happen either if no active rules match the provided data, or if the matched rules
    /// are factors that require a final sampling value. In this case, the returned evaluator stores
    /// the state of the matched rules as well as the accumulated sampling factor and can be used to
    /// evaluate more rules.
    ///
    /// If evaluation returns `Continue` and there are no more rules to match, this should be
    /// considered as "no match".
    Continue(SamplingEvaluator),
    /// One or more rules have matched.
    Matched(SamplingMatch),
}
```

In a follow-up, we can try to use `std::ops::ControlFlow` for this. This implements `std::ops::Try`, which could allow us to write functions like this:

```rust
evaluator = evaluator.match_rules(seed, event, rules)?;
```

However, if I'm not mistaken that would require the containing function to also return a `ControlFlow`, which isn't exactly great.

relay-sampling/src/evaluation.rs
jan-auer: As per signature, this should be called `into_matched_rules` or even shorter `into_rules`.

The naming `take` indicates a signature with `&mut self`.

relay-server/src/actors/processor.rs
jan-auer: I'd suggest to revert this change for three reasons:
 1. The parameter list of this function is long. That was also the reason why `ProcessEnvelopeState` was introduced in the first place, to pass processing information together. 
 2. We will soon need more fields from the state, such as access to Redis as well as the `received` timestamp (it is currently a bug that this function uses `Utc::now()`).
 3. Associated functions like this one that neither work with the type that they are declared on, nor return it, are unidiomatic.

relay-server/src/lib.rs
jan-auer: By convention, `mod tests` is used exclusively for unit tests of the containing module. If you really want to deduplicate this helper, consider to place it in a module under `utils`. However, in this case it's probably not worth it. Tests should be as self-contained and explicit as possible, so it's better not to import such functions from other modules.

I would therefore suggest to revert this.

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -1,116 +1,71 @@
...
-                    relay_log::trace!("keeping event that matched the configuration");
-                    SamplingResult::Keep
-                }
-            }
-            None => {
-                relay_log::trace!("keeping event that didn't match the configuration");
-                SamplingResult::Keep
-            }
-        }
+    /// Returns true if the event matched on any rules.
```
jan-auer:
```suggestion
    /// Returns `true` if the event matched on any rules.
```

x 3

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -119,20 +74,26 @@ pub fn is_trace_fully_sampled(
...
     };
 
-    Some(sampled)
+    // TODO(tor): pass correct now timestamp
+    let evaluator = SamplingEvaluator::new(Utc::now()).adjust_client_sample_rate(adjustment_rate);
+
+    let rules = root_project_config.filter_rules(RuleType::Trace);
+
+    let sampling_result: SamplingResult = evaluator.match_rules(dsc.trace_id, dsc, rules).into();
+    Some(sampling_result.should_keep())
```
jan-auer: Instead of converting to `SamplingResult`, consider to match the `RuleMatchingState` directly.

```suggestion
    Some(match evaluator.match_rules(dsc.trace_id, dsc, rules) {
        RuleMatchingState::SamplingMatch(m) => m.should_keep(),
        _ => false,
    })
```

relay-sampling/src/condition.rs
TBS1996: I was debating the idea of making a dummy Getter struct, just to make it more clear that this crate shouldn't care about implementations, but since we want to keep `DynamicSamplingContext`  in this crate then it's simpler just to use that. All the tests that used `Event` I've replaced with dsc.

relay-sampling/src/config.rs
TBS1996: indeed, it was used by `fn is_trace_fully_sampled`, which took in a state as an argument. I changed the signature to take in a `SamplingConfig` because the function only used state to extract the config

relay-sampling/src/evaluation.rs
TBS1996: That looks a lot more clear, thx! 

and as mentioned I love the idea of using ControlFlow, will implement later!

relay-server/src/actors/processor.rs
TBS1996: alright, i'll revert it, although personally I prefer it like this. My thinking was that we would clearly delineate the `process_state` function which runs entirely on side effects, and it would in turn call pure functions, meaning we would avoid passing in `state` any further from `processing_state`. 

I'd say the long parameter list might tell us something about this function that simply passing in state would mask. Passing in `state` here feels a lot like using global variables. 

in the function (as expressed by the first early return) you can see how 'event' and 'sampling_config' need each other, same with dsc and root_sampling_config. Perhaps we could explore a newtype that encapsulates the "Getter" and a sampling config in a later PR? 

3:  I agree its not idiomatic, so i think it would be good in `utils` of the same crate

TBS1996: ok so, reverting it leads to complicating many tests I wrote that relied on its functional approach, where you have to start the service, create a mock `processenvelopestate` and `projectstate` to check if a field on processenvelopestate has changed. 
 
could you verify again if you think I should revert this? my suggestion would be to keep it functional, move it to utils, and let the long param list stand as a sign for us to think about refactoring in the future (rather than hiding the params in state).

relay-server/src/lib.rs
TBS1996: alright, that makes sense, I've reverted it!

relay-server/src/utils/dynamic_sampling.rs
TBS1996: hmm, im not sure if it's obvious that a lack of match should result in `true` here, case in point: in your example you return `false` whereas it should be `true`. Maybe it's better to use the implementation in `SamplingResult` for that?

TBS1996: if we return the bool manually we're essentially re-implementing this behaviour:

```rust
impl From<Evaluation> for SamplingResult {
    fn from(value: Evaluation) -> Self {
        match value {
            Evaluation::Matched(sampling_match) => Self::Match(sampling_match),
            Evaluation::Continue(_) => Self::NoMatch,
        }
    }
}
    /// Returns `true` if the event should be kept.
    pub fn should_keep(&self) -> bool {
        match self {
            SamplingResult::Match(sampling_match) => sampling_match.should_keep(),
            // If no rules matched on an event, we want to keep it.
            SamplingResult::NoMatch => true,
            SamplingResult::Pending => true,
        }
    }
```

TBS1996: in the original implementation, it turns into `Keep` if there's no match in `fn determine_from_sampling_match`, which becomes 'true' here:

```rust
    let sampled = match sampling_result {
        SamplingResult::Keep => true,
        SamplingResult::Drop(_) => false,
```

iker-barriocanal (commented):
Have we considered splitting this refactor into smaller ones so that it's easier to review and identify regressions?

jan-auer:
@iker-barriocanal next time we should try to make some of these changes in smaller increments. Now that review has almost concluded, we can keep the PR.

TBS1996:
> Have we considered splitting this refactor into smaller ones so that it's easier to review and identify regressions?

I regret not having a separate PR for moving the tests, but other than that, the reason it's so huge is because the new implementation changed the API used for tests, and there was a lottt of tests here.   
 
at this point it's easier to just go forward with it.

relay-server/src/utils/dynamic_sampling.rs
jan-auer: You're absolutely right, the `false` was an outright typo in my suggestion :) I was slightly concerned with the additional conversion into a sampling result, however my own typo is a good example of why this sort of deduplication should exist. 

Let's keep it. A suggestion on how to format it, though:

```rust
let evaluation = evaluator.match_rules(dsc.trace_id, dsc, rules);
Some(SamplingResult::from(evaluation).should_keep())
```

relay-server/src/actors/processor.rs
jan-auer: Makes sense. Given this is an internal function of the processor, we can easily revisit this at a later point. There's limited exposure of this API. Let's move forward.

jan-auer (approved):

relay-server/src/utils/dynamic_sampling.rs
TBS1996: nice, thanks! that looks a lot nicer than what I wrote

PR #2518, opened 2023-09-15
fix(upstream): Fix hot-loop on unavailable upstream

iker-barriocanal:
Fixes https://github.com/getsentry/team-ingest/issues/218

## Problem

When the upstream service is unavailable, the upstream actor has the following behavior:

1. Get an incoming message.
2. Enqueue the message.
3. Dequeue the message and send the request. Yet upstream is unavailable, the request fails.
4. Attempt to retry. This enqueues the message as high priority, effectively step 2.

As a result, there's a hot loop between steps 2 and 4, causing high CPU usage until Relay goes into network outage mode.

## Solution

This PR introduces a new internal queue in `UpstreamQueue`, besides the current existing high and low priority queues. When a request fails, the upstream broker adds the message to retry to the retry queue and starts an exponential backoff. Messages from the retry queue are only dequeued when the backoff has elapsed, and the backoff is reset on successful upstream requests.


relay-server/src/actors/upstream.rs
```diff
@@ -1449,7 +1512,7 @@ impl Service for UpstreamRelayService {
         // and authentication state.
         let mut broker = UpstreamBroker {
             client: client.clone(),
-            queue: UpstreamQueue::new(),
+            queue: UpstreamQueue::new(config.http_outage_grace_period()),
```
iker-barriocanal: This value should never be reached as Relay gets into network outage mode.

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +890,115 @@ impl Entry {
...
     }
 
-    /// Puts an entry into the queue at the given position.
-    fn put(&mut self, entry: Entry, position: EnqueuePosition) {
+    /// Puts an entry into the queue.
+    ///
+    /// High priority entries go in the front of the queue, and low priority
+    /// items in the back. [Retries][RequestAttempt::Retry] are tracked
+    /// independently.
+    fn put(&mut self, entry: Entry, attempt: RequestAttempt) {
```
jan-auer: This function no longer needs to exist. `enqueue` and `enqueue_retry` do entirely separate things, so inline their respective implementations and make them look as similar as possible.

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +890,115 @@ impl Entry {
 /// is synchronous and managed by the [`UpstreamBroker`].
 #[derive(Debug)]
 struct UpstreamQueue {
+    /// High priority queue.
     high: VecDeque<Entry>,
+    /// Low priority queue.
     low: VecDeque<Entry>,
+    /// Retry queue, sorted by [priority][`RequestPriority`].
+    retries: VecDeque<Entry>,
```
jan-auer: Instead of a single retry queue, this needs to have one for high and one for low priority. Otherwise, you break queue order on retries, since you cannot simply push_front into the retry queue.

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +890,115 @@ impl Entry {
...
+    /// Backoff handler for retry entries.
+    retry_backoff: RetryBackoff,
+    /// Retries should not be dequeued before this instant.
+    next_retry: Instant,
```
jan-auer: Simplify here. Simply have a single retry delay without backoff, since those retries only run for a _very_ short amount of time before network outage handling kicks in.

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +890,115 @@ impl Entry {
...
-    /// `enqueue_immediate` follows.
-    pub fn enqueue_immediate(&mut self, entry: Entry) {
-        self.put(entry, EnqueuePosition::Front)
+    /// If the entry is high priority it's placed at the front of the queue, or
+    /// at the back if it's low priority. Entries in the retry queue are
+    /// dequeued from front to back.
+    ///
+    /// It also schedules the next retry time, based on the retry back off. The
+    /// retry queue is not dequeued until the next retry has elapsed.
+    pub fn enqueue_retry(&mut self, entry: Entry) {
```
jan-auer:
```suggestion
    pub fn retry(&mut self, entry: Entry) {
```

relay-server/src/actors/upstream.rs
```diff
@@ -1010,6 +1065,10 @@ enum Action {
     ///
     /// The entry is placed on the front of the [`UpstreamQueue`].
     Retry(Entry),
+    /// Notifies a request has been completed successfully to upstream.
+    ///
+    /// Successful requests reset the retry backoff.
+    SuccessfulRequest,
```
jan-auer: This action is redundant. There's already a `Complete` with a request outcome that could reset the backoff.

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +890,115 @@ impl Entry {
...
+    /// 4. [Low priority][RequestPriority::Low] requests.
     pub fn dequeue(&mut self) -> Option<Entry> {
-        self.high.pop_front().or_else(|| self.low.pop_front())
+        if !self.high.is_empty() {
+            return self.high.pop_front();
+        }
+        if !self.retries.is_empty() && self.next_retry <= Instant::now() {
+            return self.retries.pop_front();
+        }
+        self.low.pop_front()
```
jan-auer: This should _first_ check the retries (but with separate queues for high and low), and then the new queues, otherwise incoming requests can starve retries. Essentially your logic should be:

```rust
let mut entry = None;

if self.next_retry <= Instant::now() {
  entry = self.retry_high.pop_front().or_else(|| self.retry_low.pop_front());
}

entry.or_else(|| self.high.pop_front()).or_else(|| self.low.pop_front())
```

jan-auer (commented):
The overall approach looks good. I think there's a possible starvation issue with the retry queue, please see the below comments.

relay-server/src/actors/upstream.rs
iker-barriocanal: I wrongly assumed the `VecDeque` was smart and worked like a linked list.

relay-server/src/actors/upstream.rs
iker-barriocanal: Completed = failed + succeeded. Failed requests are retried, and we need a way to know if a request succeeded.

relay-server/src/actors/upstream.rs
iker-barriocanal: This is a trade-off and really depends on what we want to prioritize. Let's consider the following scenarios, assuming Relay produces more low-priority requests than high-priority (e.g. more outgoing envelopes than project config requests):

1. Under regular operation, everything works as expected.
2. When upstream is completely down, the order doesn't matter.
3. When upstream is flaky but downstream works fine, Relay is still processing envelopes but needs to decide what is prioritizing. If high-priority requests are prioritized, there's a chance that envelopes starve; if low-priority requests are prioritized, there's a chance we don't get project config updates (or even new ones).
4. When upstream recovers after being down, all requests will eventually go out and we need to decide what to request first. Requesting high-priority first means prioritizing config fetching and maybe more accurate envelope processing; requesting low-priority first means flushing envelopes but possibly processing incoming envelopes with stale configs.

Another possibility is the initial assumption to be incorrect and Relay produces more high-priority requests than low-priority ones. That is a different issue that should be handled independently.

In my opinion, keeping envelopes in memory and ensuring up-to-date envelope processing is more important. I'm optimistic and think envelopes will be forwarded in the end, but if this can really be an issue we may want to consider spooling them at this stage. I've followed this approach in [36c065a](https://github.com/getsentry/relay/pull/2518/commits/36c065a7a3b171723981c8421bad7fcf0ce1410e). I'm not strongly opinionated and happy to change the prioritization if somebody thinks otherwise or I'm missing something.

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +872,108 @@ impl Entry {
...
-            EnqueuePosition::Back => queue.push_back(entry),
+        match priority {
+            RequestPriority::High => self.high.push_back(entry),
+            RequestPriority::Low => self.low.push_back(entry),
         }
-
         relay_statsd::metric!(
             histogram(RelayHistograms::UpstreamMessageQueueSize) = self.len() as u64,
             priority = priority.name(),
+            attempt = "first"
```
iker-barriocanal, line 925: I think observing how many attempts are coming from retries can be useful in certain scenarios.

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +872,108 @@ impl Entry {
 /// is synchronous and managed by the [`UpstreamBroker`].
 #[derive(Debug)]
 struct UpstreamQueue {
+    /// High priority queue.
     high: VecDeque<Entry>,
+    /// Low priority queue.
     low: VecDeque<Entry>,
+    /// Low priority retry queue.
```
TBS1996:
```suggestion
    /// High priority retry queue.
```

relay-server/src/actors/upstream.rs
TBS1996: the complete message does contain whether it was succeeded or not though, so I think you could do it in `fn complete`

relay-server/src/actors/upstream.rs
```diff
@@ -1378,7 +1412,10 @@ impl UpstreamBroker {
                     entry.retries += 1;
                     action_tx.send(Action::Retry(entry)).ok();
                 }
-                _ => entry.request.respond(result).await,
+                _ => {
+                    entry.request.respond(result).await;
+                    action_tx.send(Action::SuccessfulRequest).ok();
```
TBS1996: shouldn't this only send if the `RequestOutcome` is of variant `Received` ? here it will be sent if it's `Dropped` and the entry is not marked as `retry`

theres essentially 3 arms here

`RequestOutcome::Dropped && entry.request.retry() == true`,
`RequestOutcome::Dropped && entry.request.retry() == false`,
`RequestOutcome::Received`,

I think we should only reset the retry-backoff on the final arm

relay-server/src/actors/upstream.rs
```diff
@@ -1394,14 +1424,17 @@ impl UpstreamBroker {
 
         match status {
             RequestOutcome::Dropped => self.conn.notify_error(&self.action_tx),
-            RequestOutcome::Received => self.conn.reset_error(),
+            RequestOutcome::Received => {
+                self.conn.reset_error();
+                self.queue.retry_backoff_reset();
```
TBS1996:
```suggestion
                self.queue.reset_retry_backoff();
```

Personally I find it more clear to lead with 'reset' if it's a function that resets something, like in the `reset_error` function, what do you think ?

TBS1996 (approved):
to me this looks good, seems like an nice solution!

relay-server/src/actors/upstream.rs
iker-barriocanal: I agree! I prefer (starting with) verbs in function names.

relay-config/src/config.rs
```diff
@@ -706,6 +706,10 @@ struct Http {
     /// During a network outage relay will try to reconnect and will buffer all upstream messages
     /// until it manages to reconnect.
     outage_grace_period: u64,
+    /// Time Relay waits before retrying an upstream request, in milliseconds.
+    ///
+    /// This time is only used before going into a network outage mode.
+    unavailable_upstream_retry_period: u64,
```
olksdr: Maybe it's better to call this interval? 
```suggestion
    unavailable_upstream_retry_interval: u64,
```
Since this is the time Relay will wait to retry again and again, won't it ?

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +872,108 @@ impl Entry {
...
-    /// Since entries are dequeued in FIFO order, this entry will be dequeued last within its
-    /// priority class.
-    pub fn enqueue(&mut self, entry: Entry) {
-        self.put(entry, EnqueuePosition::Back)
-    }
-
-    /// Place an entry at the front of the queue.
+    /// If the entry is high priority it's placed at the front of the queue, or
+    /// at the back if it's low priority. Entries in the retry queue are
+    /// dequeued from front to back.
```
olksdr: This description is wrong here. There are 2 queues now, and entry places at the back of each queue.

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +872,108 @@ impl Entry {
...
+    /// Low priority retry queue.
+    retry_low: VecDeque<Entry>,
+    /// Retries should not be dequeued before this instant.
+    ///
+    /// This retry increments by the constant `retry_after` instead of backoff,
+    /// since it only kicks in for a short period of time before Relay gets into
+    /// network outage mode (see [`IsNetworkOutage`]).
+    next_retry: Instant,
+    /// Time to wait before retrying another request from the retry queue.
+    retry_after: Duration,
```
olksdr: To make it clearer I would have it `interval`

```suggestion
    retry_interval: Duration,
```

relay-server/src/actors/upstream.rs
```diff
@@ -881,69 +872,108 @@ impl Entry {
...
     ///
-    /// This always returns entries with [high](RequestPriority::High) first before dequeueing
-    /// entries with low priority.
+    /// Highest priority entry is determined by (1) request priority and (2)
+    /// retries first.
     pub fn dequeue(&mut self) -> Option<Entry> {
-        self.high.pop_front().or_else(|| self.low.pop_front())
+        let now = Instant::now();
+
+        if !self.retry_high.is_empty() && self.next_retry <= now {
```
olksdr: you don't need this `now` here, you could use `Instant::elapsed()` function.

olksdr (commented):
Some suggestions to the code.

jan-auer (approved):
Good to go, thanks!

PR #2522, opened 2023-09-20
feat(spans): Ignore more spans for the DB module

phacops:
We've found more problematic spans we'd like to ignore as our grouping gets better.

https://www.notion.so/sentry/Problematic-Projects-During-Database-View-V0-Rollout-463d6f3f1f3d47baa127c97cf0e56716


relay-server/src/actors/processor.rs
```diff
@@ -2283,6 +2283,8 @@ impl EnvelopeProcessorService {
             .project_state
             .has_feature(Feature::SpanMetricsExtractionAllModules);
 
+        let op_denylist = vec!["clickhouse", "mongodb", "redis", "activerecord"];
```
AbhiPrasad: can this be a set? I think we can avoid the clone into iter boilerplate then below.

relay-server/src/actors/processor.rs
```diff
@@ -2294,8 +2296,21 @@ impl EnvelopeProcessorService {
...
                     if all_modules_enabled
-                        || span_op.starts_with("db") && !span_description.contains(r#""$"#)
+                        || (span_op.starts_with("db")
+                            && op_denylist
+                                .clone()
+                                .into_iter()
+                                .any(|op| span_op.contains(op))
+                            && !(span_op == "db.sql.query"
+                                && !(span_description.contains(r#""$"#)
+                                    || span_system == "mongodb")))
```
AbhiPrasad: we should probably string enum all of these sometime soon given there will be more coming.

AbhiPrasad (approved):

relay-server/src/actors/processor.rs
phacops: I'll refactor when we start having more. Maybe we'll fix them fast and we won't need that.

relay-server/src/actors/processor.rs
```diff
@@ -2283,6 +2284,8 @@ impl EnvelopeProcessorService {
             .project_state
             .has_feature(Feature::SpanMetricsExtractionAllModules);
 
+        let op_denylist = HashSet::from(["clickhouse", "mongodb", "redis", "activerecord"]);
```
jan-auer: This constructs a data structure for static data in every run. Instead of this, you could make this a function that uses `match` to check for the four allowed values.

jan-auer (approved):

relay-server/src/actors/processor.rs
phacops: I ended up refactoring the conditions so it's just code. Let me know if that's better.

PR #2523, opened 2023-09-20
fix(profiling): Remove profile ID from transaction context if no profile is found in the envelope

phacops:
Some SDKs might send us a profile context in the transaction, containing a profile ID, when they're not supposed to.


relay-server/src/actors/processor.rs
```diff
@@ -1129,29 +1129,16 @@ impl EnvelopeProcessorService {
     /// Process profiles and set the profile ID in the profile context on the transaction if successful
     #[cfg(feature = "processing")]
     fn process_profiles(&self, state: &mut ProcessEnvelopeState) {
+        let mut found_profile_id: Option<EventId> = Default::default();
```
jjbayer: The compiler should be able to deduce the type here AFAIK:

```suggestion
        let mut found_profile_id = None;
```

jjbayer (approved):

PR #2527, opened 2023-09-22
feat(spans): Scrub SQL aliases

jjbayer:
Aliases in SQL queries are sometimes generated and can cause high cardinality.

ref: [internal issue](https://www.notion.so/sentry/Scrub-Postgres-AS-aliases-6c70248883174805a6e1f009046243d9?pvs=4)

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/sql/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/mod.rs
@@ -443,13 +443,13 @@ mod tests {
     scrub_sql_test!(
         collapse_partial_column_lists,
         r#"SELECT myfield1, "a"."b", count(*) AS c, another_field, another_field2 FROM table1 WHERE %s"#,
-        "SELECT .., count(*) AS c, .. FROM table1 WHERE %s"
+        "SELECT .., count(*), .. FROM table1 WHERE %s"
     );
 
     scrub_sql_test!(
         collapse_partial_column_lists_2,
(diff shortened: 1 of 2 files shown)


relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -443,13 +443,13 @@ mod tests {
...
         r#"SELECT myfield1, "a"."b", count(*) AS c, another_field, another_field2 FROM table1 WHERE %s"#,
-        "SELECT .., count(*) AS c, .. FROM table1 WHERE %s"
+        "SELECT .., count(*), .. FROM table1 WHERE %s"
     );
 
     scrub_sql_test!(
         collapse_partial_column_lists_2,
         r#"SELECT DISTINCT a, b,c ,d , e, f, g, h, COALESCE(foo, %s) AS "id" FROM x"#,
-        "SELECT DISTINCT .., COALESCE(foo, %s) AS id FROM x"
+        "SELECT DISTINCT .., COALESCE(foo, %s) FROM x"
```
jjbayer, line 452: @gggritso this will again change a lot of group hashes and thereby mess with the "time spent" column, are we OK with that?

olksdr (approved):

relay-event-normalization/src/normalize/span/description/sql/mod.rs
gggritso: I asked in the team channel just in case, but IMO this is OK and we should do it sooner rather than later!

PR #2532, opened 2023-09-25
fix(crash): Fix the crash-handler feature

jan-auer:
Repairs the crash handler feature, which was effectively disabled in recent builds.

- Internal feature flagging was broken, and for this reason the functionality was never active when the feature flag was enabled.
- Handling of envelopes from the native crash reporter was broken so that they could not be uploaded.
- The configured environment was not used for native crashes.

### DIFF
--- a/relay-log/Cargo.toml
+++ b/relay-log/Cargo.toml
@@ -34,8 +34,8 @@ tracing-subscriber = { version = "0.3.17", features = [
 ], optional = true }
 
 [features]
-dashboard = ["dep:once_cell", "dep:tokio"]
 default = []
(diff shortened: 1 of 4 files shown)


relay-log/Cargo.toml
```diff
@@ -32,10 +34,7 @@ tracing-subscriber = { version = "0.3.17", features = [
 ], optional = true }
 
 [features]
-dashboard = [
-    "dep:once_cell",
-    "dep:tokio",
-]
+dashboard = ["dep:once_cell", "dep:tokio"]
```
olksdr: Do you want to put `default = []` on the first place?

olksdr (approved):

PR #2536, opened 2023-09-26
fix(spans): Use db.system for SQL parsing

jjbayer:
The SQL parser used for span description normalization and grouping accessed the wrong field to pick the SQL dialect. This bug did not durface in unit tests, because I passed the system string into those tests directly.

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -34,7 +34,7 @@ pub(crate) fn scrub_span_description(span: &mut Span, rules: &Vec<SpanDescriptio
             ("db", _) => sql::scrub_queries(
                 span.data
                     .value()
-                    .and_then(|d| d.get("system"))
+                    .and_then(|d| d.get("db.system"))
                     .and_then(|v| v.as_str()),
                 description,
             ),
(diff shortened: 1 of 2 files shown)


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -34,7 +34,7 @@ pub(crate) fn scrub_span_description(span: &mut Span, rules: &Vec<SpanDescriptio
             ("db", _) => sql::scrub_queries(
                 span.data
                     .value()
-                    .and_then(|d| d.get("system"))
+                    .and_then(|d| d.get("db.system"))
```
jjbayer, line 37: See https://develop.sentry.dev/sdk/performance/span-data-conventions/#database.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -428,4 +427,30 @@ mod tests {
...
+        let mut span = Annotated::<Span>::from_json(json).unwrap();
+        let span = span.value_mut().as_mut().unwrap();
+        scrub_span_description(span, &vec![]);
+        let scrubbed = span
+            .data
+            .value()
+            .and_then(|d| d.get("description.scrubbed"))
+            .and_then(|v| v.value())
+            .and_then(|v| v.as_str());
+        assert_eq!(scrubbed, Some("SELECT %s"));
```
jjbayer, line 454: Without the fix, this would be `SELECT not an identifier`, because the default parser would assume that double quotes delimit an identifier.

olksdr (approved):
🚢  it!

PR #2538, opened 2023-09-26
fix(spans): Prevent stack overflow on SQL serialization

jjbayer:
We see stack overflows when serializing parsed SQL statements back to strings. Interestingly, the parser and the visitor that run before serialization do not run into the same issue (so far).

This PR attempts to measure the depth of expressions in the AST while visiting and returns an error for heavily nested expressions, thereby hopefully preventing stack overflows.

ref: https://github.com/getsentry/team-ingest/issues/223


relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -344,3 +381,62 @@ impl Dialect for DialectWithParameters {
...
+
+    use super::*;
+
+    #[test]
+    fn depth() {
+        let query = "SELECT (a + b + (c1 + (c2 + c3)) + c)";
+        let mut parsed = Parser::parse_sql(&GenericDialect {}, query).unwrap();
+        let mut visitor = NormalizeVisitor::new();
+        parsed.visit(&mut visitor);
+        assert_eq!(visitor.max_depth, 8);
```
jjbayer: Both the parentheses and the expression within account for one level.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -344,3 +381,62 @@ impl Dialect for DialectWithParameters {
...
+        assert_eq!(visitor.max_depth, 1);
+    }
+
+    #[test]
+    fn depth_nested_from_2() {
+        let query = "SELECT * FROM (SELECT * FROM (SELECT 1))";
+        let mut parsed = Parser::parse_sql(&GenericDialect {}, query).unwrap();
+        let mut visitor = NormalizeVisitor::new();
+        parsed.visit(&mut visitor);
+        assert_eq!(visitor.max_depth, 2);
```
jjbayer: This seems off-by-one, but it should give us an approximation that is good enough to deny heavily nested queries.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -8,6 +8,8 @@ use sqlparser::ast::{
 };
 use sqlparser::dialect::{Dialect, GenericDialect};
 
+const MAX_DEPTH: usize = 64;
```
iker-barriocanal: Is this arbitrary (for now)?

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -142,6 +168,14 @@ impl NormalizeVisitor {
         }
     }
 
+    /// Subtracts from depth and records maximum found depth.
+    ///
+    /// Use when leaving an expression.
+    fn record_depth(&mut self) {
+        self.max_depth = self.max_depth.max(self.current_depth);
+        self.current_depth = self.current_depth.saturating_sub(1);
```
iker-barriocanal: TIL `saturating_sub`. Should we `saturate_add` on line 196?

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -116,21 +140,23 @@ impl NormalizeVisitor {
...
 
-    fn transform_table_factor(table_factor: &mut TableFactor) {
+    fn transform_table_factor(&mut self, table_factor: &mut TableFactor) {
         match table_factor {
             // Recurse into subqueries.
             TableFactor::Derived { subquery, .. } => {
-                Self::transform_query(subquery);
+                self.current_depth += 1;
+                self.transform_query(subquery);
+                self.record_depth();
```
iker-barriocanal: We update the depth on pre/post expression visiting, is repeating it here redundant?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
olksdr: Max depth of what is it? Nested queries (sub-queris)? Maybe `MAX_NESTED_QUERIES` sounds better here?
Or each element of the statement we currently parse (e.g. Declre, Fetch, Savepoint etc?). 
It would be great to  some docs here, or/and change the name to better reflect the use case.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -344,3 +381,62 @@ impl Dialect for DialectWithParameters {
...
+        let query = "SELECT (SELECT (SELECT 1)) AS one FROM a";
+        let mut parsed = Parser::parse_sql(&GenericDialect {}, query).unwrap();
+        let mut visitor = NormalizeVisitor::new();
+        parsed.visit(&mut visitor);
+        assert_eq!(visitor.max_depth, 3);
+    }
+
+    #[test]
+    fn parse_deep_expression() {
+        let query = "SELECT 1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1";
```
olksdr, line 417: From this tests, it seems to me that `MAX_DEPTH` is the number of element in the query we parse? like `SELECT` is one and than each `1` and `+` and  two more and so forth ?

relay-event-normalization/src/normalize/span/description/sql/parser.rs
olksdr: This should be `3` here? If we count the number overall queries?  But it says `2` - maybe because only 2 of them are nested?  
Asking because want to understand how this works.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
olksdr: I kind of lost why `8` here.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: It's supposed to reflect the depth of the abstract syntax tree parsed by `sqlparser`. In this case, the arithmetic expression gets parsed as

```
   +
 /   \
1     +
    /   \
   1     etc.
```

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: Yes.

jjbayer: It's supposed to limit the level of nesting within a query. I simplified the way it's measured now, + added some docs.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: The parser creates a new `Expression` object for both the parentheses _and_ every binary operation contained within the parentheses.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: I was trying to measure the level of nesting of not just expressions, but also subqueries. I reverted that part now though because it made the measured depth kind of voodoo.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: The chance that we exhaust `u64` is very low, so I would argue that `saturating_add` is unnecessary in practice.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
olksdr: cool, thanks! 
That's where my confusion came from.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: I simplified the PR to measure only the level of nesting of expressions instead of trying to partially cover other compositions as well.

PR #2541, opened 2023-09-26
fix(beta): Address beta lint warnings

olksdr:
New rust version will introduce new  lint warnings and suggestions. 
This PR fixes those warning for `beta`, so once the new version is released everything still continues to function. 

#skip-changelog


iker-barriocanal (approved):
Thanks for doing this!

PR #2544, opened 2023-09-27
ref(sampling): Replace Evaluator with ControlFlow.

TBS1996:
The `Evaluator` enum functioned exactly as `ControlFlow`, so jan had the idea of simply using ControlFlow instead. This is the implementation of that. 

pros:
 * Idiomatic
 * One less type you have to look up when reading the code
 
cons:
 * Conversion to `SamplingResult` looks more messy now.
 * Had to forego some helpful methods I used in testing. 
 
I moved the documentation from the type to the `match_rules` function. Not sure if that's a benefit or disadvantage. 

#skip-changelog

### DIFF
--- a/relay-sampling/src/evaluation.rs
+++ b/relay-sampling/src/evaluation.rs
@@ -238,7 +238,6 @@ impl fmt::Display for MatchedRuleIds {
     }
 }
 
-/*
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


iker-barriocanal (approved):
Nice! And thanks for writing the docs!

PR #2545, opened 2023-09-27
feat(spans): Allow activerecord with db.system

jjbayer:
We've seen SQL query normalization fail when the queries are prepended with an active record comment, for example

```SQL
/*some comment `my_function'*/ SELECT 1
```

Normalization works fine when `sqlparser` is able to parse the query though. Having a `db.system` in `span.data` is a strong indicator that we will be able to parse the query. So we re-allow activerecord spans if they have a `db.system` set.

In addition, this PR also improves the fallback regex behavior for this kind of comments.

ref: [internal issue](https://www.notion.so/sentry/Scrub-ActiveRecord-query-comments-236925a9eaa24423880ae285939f8588?pvs=4)

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -47,7 +47,6 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                     inner: Box::new(RuleCondition::Glob(GlobCondition {
                         name: span_op_field_name.into(),
                         value: GlobPatterns::new(vec![
-                            "*active*record*".into(),
                             "*clickhouse*".into(),
                             "*mongodb*".into(),
                             "*redis*".into(),
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -488,7 +488,7 @@ mod tests {
         scrub_span_description(span.value_mut().as_mut().unwrap(), &vec![]);
         let scrubbed = get_value!(span.data["description.scrubbed"]);
 
-        // Do not scrub active record.
+        // When db.system is missing, no scrubbed description (i.e. no group) is set.
         assert!(scrubbed.is_none());
     }
 
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -24,6 +24,13 @@ pub(crate) fn scrub_span_description(span: &mut Span, rules: &Vec<SpanDescriptio
         return;
     };
 
+    let db_system: &str = span
+        .data
+        .value()
+        .and_then(|v| v.get("span.system"))
+        .and_then(|system| system.as_str())
+        .unwrap_or_default();
+
     let scrubbed = span
         .op
         .as_str()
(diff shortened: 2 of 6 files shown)


relay-dynamic-config/src/defaults.rs
```diff
@@ -47,7 +47,6 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                     inner: Box::new(RuleCondition::Glob(GlobCondition {
                         name: span_op_field_name.into(),
                         value: GlobPatterns::new(vec![
-                            "*active*record*".into(),
```
jjbayer, line 50: We now allow metrics from active record spans.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -24,20 +24,31 @@ pub(crate) fn scrub_span_description(span: &mut Span, rules: &Vec<SpanDescriptio
...
-                    .value()
-                    .and_then(|d| d.get("db.system"))
-                    .and_then(|v| v.as_str()),
-                description,
-            ),
+            ("db", sub) => {
+                if sub.contains("clickhouse")
+                    || sub.contains("mongodb")
+                    || sub.contains("redis")
+                    || is_legacy_activerecord(sub, db_system)
```
jjbayer, line 44: When active record does not provide a `db.system`, we cannot parse it, so provide no scrubbed description at all.

relay-server/src/actors/processor.rs
```diff
@@ -2244,12 +2244,8 @@ impl EnvelopeProcessorService {
...
             .unwrap_or_default();
         op.starts_with("db")
-            && !(op.contains("active_record")
-                || op.contains("activerecord")
-                || op.contains("clickhouse")
-                || op.contains("mongodb")
-                || op.contains("redis"))
-            && !(op == "db.sql.query" && !(description.contains(r#""$"#) || system == "mongodb"))
+            && !(op.contains("clickhouse") || op.contains("mongodb") || op.contains("redis"))
+            && !(op == "db.sql.query" && (description.contains(r#""$"#) || system == "mongodb"))
```
jjbayer, line 2248: Drive by fix: There was a double negation here that included `db.sql.query` _only_ for mongodb.

relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -130,9 +130,9 @@ fn scrub_queries_inner(db_system: Option<&str>, string: &str) -> (Option<String>
     let mut string = Cow::from(string.trim());
 
     for (regex, replacement) in [
-        (&NORMALIZER_REGEX, "$pre%s"),
         (&COMMENTS, "\n"),
         (&INLINE_COMMENTS, ""),
+        (&NORMALIZER_REGEX, "$pre%s"),
```
jjbayer, line 135: Comments may contain string delimiters (`'`), and strings may contain comment delimiters. So no matter how we order the regexes, one can construct a case where scrubbing does not work. Prefer this order for now because we've seen that the previous order destroyed active record descriptions.

olksdr (approved):
lgtm!

iker-barriocanal (approved):

PR #2546, opened 2023-09-27
fix(spans): Scrub literals inside SQL casts

jjbayer:
Due to a bug in the SQL normalization visitor we did not normalize the contents of SQL `CAST`s and cast expressions.

This PR makes sure that the children of a cast are visited before it gets replaced by its contents.

#skip-changelog


relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -235,9 +231,14 @@ impl VisitorMut for NormalizeVisitor {
...
 
     fn post_visit_expr(&mut self, expr: &mut Expr) -> ControlFlow<Self::Break> {
-        if let Expr::CompoundIdentifier(parts) = expr {
-            Self::simplify_compound_identifier(parts);
+        // Casts are omitted for simplification. Because we replace the entire expression,
+        // the replacement has to occur *after* visiting its children.
+        if let Expr::Cast { expr: inner, .. } = expr {
+            let mut swapped = Expr::Value(Value::Null);
+            std::mem::swap(&mut swapped, inner);
+            *expr = swapped;
```
jjbayer, line 232: This is clumsy, but I wanted to prevent cloning `inner`. Open to better ideas.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -235,9 +231,14 @@ impl VisitorMut for NormalizeVisitor {
     }
 
     fn post_visit_expr(&mut self, expr: &mut Expr) -> ControlFlow<Self::Break> {
-        if let Expr::CompoundIdentifier(parts) = expr {
-            Self::simplify_compound_identifier(parts);
```
jjbayer, line 232: This was an earlier attempt at fixing normalization inside casts, but it only worked for a single use case.

gggritso (approved):

olksdr (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
iker-barriocanal: It seems expressions are generally updating in the pre_visit. Is there any reason to do this one in the post_visit?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: Yep, the reasoning is in lines 234 - 235! When I replace `expr` in the pre-visit, it's children are not being visited anymore.

PR #2549, opened 2023-09-27
feat(spans): Set description for ui.load metrics

jjbayer:
Start setting a `description.scrubbed` for `ui.load` spans.

We expect these descriptions to be low-cardinality, and metrics for non-db spans are only extracted for a handful of projects at the moment, so it should be low-risk to use the given descriptions as-is for now, though scrubbing might become necessary in the future.

#skip-changelog


olksdr (approved):

shruthilayaj (approved):
thanks!

iker-barriocanal (approved):

nikhars (approved):

PR #2550, opened 2023-09-27
feat(sampling): Reservoir sampling

TBS1996:
Relay implementation of the reservoir project: https://github.com/getsentry/sentry/issues/54449

Reservoir bias uses a type of `SamplingRule` which will sample all matches until a certain limit has been reached. This limit is tracked both locally on each relay, and with a global synchronized one in redis that procesisng relays can have access to. The redis counter will update the local counter if it's available. 

The counters are saved on the `Project` struct, with a Mutex<BTreeMap<Ruleid, i64>> . 
When we send an envelope for processing we send its corresponding project counters in the `ProcessEnvelopeState` to the `EnvelopeProcessorService`. 

There, in the `dynamic-sampling` crate, we introduce a `ReservoirEvaluator`, which will, when a reservoir rule is matching, check if the rule has been reached or not by using the local counters we sent or if applicable the redis global count. The `ReservoirEvaluator` also takes care of updating both redis and the local counter. 

After the limit is reached, the rule is no longer valid and will be ignored, so that the normal `SampleRate` and `Factor` variant of `SamplingValue` will apply. 

Sentry is responsible for removing the reservoir rule from the `SamplingConfig` when it has reached its limit. 

Whenever we receive a new `ProjectConfig`, we remove all the reservoir counters from its project that are no longer in the `DynamicSamplingConfig`. 

regarding the use of mutex:
We use try_lock to avoid getting blocked in case the mutex is already in use. There's two reasons it might be blocked. 

1. Another thread is handling a reservoir rule from the same project at the same time.
2. We are in the process of removing counters from that same project.


relay-kafka/src/lib.rs
```diff
@@ -20,7 +20,7 @@
 //!
 //!     // build the client
 //!     let kafka_client = builder.build();
-//!
```
TBS1996, line 23: I didn't do this change on purpose, linter is forcing me. But It should be like this anyway, so might as well leave it like this.

relay-sampling/src/config.rs
```diff
@@ -81,14 +83,33 @@ impl SamplingRule {
...
         if !self.time_range.contains(now) {
             // Return None if rule is inactive.
             return None;
         }
 
-        let sampling_base_value = self.sampling_value.value();
+        let sampling_base_value = match self.sampling_value {
+            SamplingValue::SampleRate { value } => value,
+            SamplingValue::Factor { value } => value,
+            SamplingValue::Reservoir { limit } => {
```
TBS1996, line 99: ugly: we return SamplingValue::Reservoir{limit}, even though the limit no longer has any meaning passed this point. It was the easiest solution, will refactor in the future.

relay-sampling/src/config.rs
```diff
@@ -140,18 +164,16 @@ pub enum SamplingValue {
...
     /// accumulated factors before moving onto the next possible match.
     Factor {
-        /// The fator to apply on another matched sample rate.
+        /// The factor to apply on another matched sample rate.
         value: f64,
     },
-}
-
-impl SamplingValue {
-    pub(crate) fn value(&self) -> f64 {
```
TBS1996, line 149: function no longer makes sense because reservoir limit doesn't have an analogous value to these

relay-sampling/src/config.rs
```diff
@@ -81,14 +83,33 @@ impl SamplingRule {
...
 
-    /// Returns the sample rate if the rule is active.
-    pub fn sample_rate(&self, now: DateTime<Utc>) -> Option<SamplingValue> {
+    /// Returns `true` if rule is a reservoir rule.
+    pub fn is_reservoir(&self) -> bool {
+        matches!(&self.sampling_value, &SamplingValue::Reservoir { .. })
+    }
+
+    /// Returns the updated [`SamplingValue`] if it's valid.
+    pub fn evaluate(
```
TBS1996, line 86: renamed the `sample_rate` function.

relay-sampling/src/evaluation.rs
```diff
@@ -24,13 +28,116 @@ fn pseudo_random_from_uuid(id: Uuid) -> f64 {
...
+/// - Redis Counter: For processing relays, a Redis-based counter provides synchronization
+///   across multiple relay-instances. When incremented, the Redis counter returns the current global
+///   count for the given rule, which is then used to update the local counter.
+#[derive(Debug)]
+pub struct ReservoirEvaluator {
+    counters: ReservoirCounters,
+    #[cfg(feature = "redis")]
+    redis_pool: Option<Arc<RedisPool>>,
+    #[cfg(feature = "redis")]
+    org_id: Option<u64>,
```
TBS1996: Ideally this would be one option, since we only care about having both or none. But a new type is overkill and a tuple might be messy. 

We always have access to the org id but I opted for putting it behind an option rather than initializing with an invalid value, even though it wouldn't matter for the code execution

relay-sampling/src/evaluation.rs
```diff
@@ -24,13 +28,116 @@ fn pseudo_random_from_uuid(id: Uuid) -> f64 {
...
+        self.org_id = Some(org_id);
+        self.redis_pool = Some(redis_pool);
+        self
+    }
+
+    /// Evaluates a reservoir rule, returning `true` if it should be sampled.
+    pub fn evaluate(&self, rule: RuleId, limit: i64) -> bool {
+        // If the mutex is already locked, we abort the match, for performance reasons.
+
+        let Ok(mut map_guard) = self.counters.try_lock() else {
```
TBS1996: thoughts on creating a metric here? We assume that we will almost always get the match, but we can't be sure.

perhaps it could crash with `fn remove_expired_reservoir_rules` when a new project config is received.

relay-sampling/src/evaluation.rs
```diff
@@ -93,6 +207,9 @@ impl SamplingEvaluator {
                         self.rule_ids,
                     ));
                 }
+                SamplingValue::Reservoir { .. } => {
+                    return ControlFlow::Break(SamplingMatch::new(1.0, seed, vec![rule.id]));
```
TBS1996, line 220: we only take the last rule id in the reservoir case because the reservoir matching overrides all previously matched rules.

relay-server/src/actors/processor.rs
```diff
@@ -565,6 +573,8 @@ impl EnvelopeProcessorService {
         });
 
         let inner = InnerProcessor {
+            #[cfg(feature = "processing")]
+            redis_pool: _redis.clone().map(Arc::new),
             #[cfg(feature = "processing")]
             rate_limiter: _redis
                 .map(|pool| RedisRateLimiter::new(pool).max_limit(config.max_rate_limit())),
```
TBS1996, line 580: thoughts on sending the RedisRateLimiter the same arc as in the redis_pool field in a follow-up pr?

relay-server/src/actors/project.rs
```diff
@@ -421,6 +424,25 @@ impl Project {
...
+        let Some(config) = self
+            .state
+            .as_ref()
+            .and_then(|state| state.config.dynamic_sampling.as_ref())
+        else {
+            return;
+        };
+
+        if let Ok(mut guard) = self.reservoir_counters.try_lock() {
+            guard.retain(|key, _| config.rules_v2.iter().any(|rule| rule.id == *key));
```
TBS1996, line 444: @jan-auer  avoiding allocations by simply passing in all the rules, regardless of if they're reservoir or not. Shouldn't matter at all since non-reservoir ones wont match anyway.

relay-server/src/actors/project_cache.rs
```diff
@@ -570,11 +570,9 @@ impl ProjectCacheBroker {
...
 
         let project_cache = self.services.project_cache.clone();
-        self.get_or_create_project(project_key).update_state(
-            project_cache,
-            state.clone(),
-            no_cache,
-        );
+        let project = self.get_or_create_project(project_key);
+        project.update_state(project_cache, state.clone(), no_cache);
+        project.remove_expired_reservoir_rules();
```
TBS1996: we check for expired reservoir rules every time we get a new project config. It's simply about cleaning up memory, having expired rules in the counters isn't an invalid state or anything.

relay-server/src/actors/project_cache.rs
```diff
@@ -680,6 +678,8 @@ impl ProjectCacheBroker {
             ..
         }) = project.check_envelope(managed_envelope, self.services.outcome_aggregator.clone())
         {
+            let reservoir_counters = project.reservoir_counters();
```
TBS1996, line 683: placement is a bit odd due to borrowing rules

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Introduce reservoir sampling rule. ([#2550](https://github.com/getsentry/relay/pull/2550))
```
jan-auer: Let's move this to "Internal" since the user-visible feature merely indirectly makes use of this.

relay-sampling/Cargo.toml
```diff
@@ -9,14 +9,20 @@ edition = "2021"
 license-file = "../LICENSE"
 publish = false
 
+[features]
+default = []
+redis = ["dep:relay-redis"]
+
 [dependencies]
+anyhow = { workspace = true }
```
jan-auer: It looks like `anyhow` is only needed for the redis feature, so you can update this to:
```suggestion
redis = ["dep:anyhow", "dep:relay-redis"]

[dependencies]
anyhow = { workspace = true, optional = true }
```

relay-sampling/src/config.rs
```diff
@@ -81,14 +83,33 @@ impl SamplingRule {
         self.condition.supported() && self.ty != RuleType::Unsupported
     }
 
-    /// Returns the sample rate if the rule is active.
-    pub fn sample_rate(&self, now: DateTime<Utc>) -> Option<SamplingValue> {
+    /// Returns `true` if rule is a reservoir rule.
+    pub fn is_reservoir(&self) -> bool {
+        matches!(&self.sampling_value, &SamplingValue::Reservoir { .. })
+    }
```
jan-auer: This function is not referenced anywhere, and it's a rather specific function -- looks like it can be removed.

relay-sampling/src/evaluation.rs
```diff
@@ -24,13 +28,115 @@ fn pseudo_random_from_uuid(id: Uuid) -> f64 {
...
+#[cfg(feature = "redis")]
+fn increment_redis_reservoir_count(
+    redis: &Arc<RedisPool>,
+    org_id: u64,
+    rule_id: RuleId,
+) -> anyhow::Result<i64> {
+    let mut command = relay_redis::redis::cmd("INCR");
+
+    let key = format!("bias:{}:{}", org_id, rule_id);
+    command.arg(key);
```
jan-auer: I would've expected clippy to alert on this.

```suggestion
    command.arg(format!("bias:{org_id}:{rule_id}"));
```

Separately, should we rather call this something like `reservoir:`, to make it more specific to what this counter is for? It's not unlikely that there will be other biases in the future that track stuff in Redis.

relay-server/src/actors/processor.rs
```diff
@@ -1352,6 +1363,19 @@ impl EnvelopeProcessorService {
...
+            #[allow(unused_mut)]
+            let mut reservoir = ReservoirEvaluator::new(reservoir_counters);
+
+            #[cfg(feature = "processing")]
+            if let Some(redis_pool) = self.inner.redis_pool.as_ref() {
+                let org_id = managed_envelope.scoping().organization_id;
+                reservoir = reservoir.set_redis(org_id, redis_pool.clone());
+            }
+
+            Arc::new(reservoir)
```
jan-auer: This `Arc` is not strictly necessary. The `SamplingEvaluator` can hold a mutable reference to the `ReservoirEvaluator`, which means you can pass `&mut state.reservoir` into `compute_sampling_decision`.

relay-sampling/src/evaluation.rs
```diff
@@ -41,9 +147,16 @@ impl SamplingEvaluator {
             rule_ids: vec![],
             factor: 1.0,
             client_sample_rate: None,
+            reservoir: None,
         }
     }
 
+    /// Sets a [`ReservoirEvaluator`].
+    pub fn set_reservoir(mut self, reservoir: Option<Arc<ReservoirEvaluator>>) -> Self {
```
jan-auer: _nit_: What if we change the signature:

```suggestion
    pub fn set_reservoir(mut self, reservoir: Arc<ReservoirEvaluator>) -> Self {
```

Or in combination with my other comment:

```rust
impl <'a> SamplingEvaluator<'a> {
    // ...
    pub fn set_reservoir(mut self, reservoir: &'a mut ReservoirEvaluator) -> Self {
```

relay-server/src/actors/project_cache.rs
jan-auer: I would prefer if this was done internally in the `Project` rather than from the outside. It's already a bad pattern that the project cache needs to handle the dequeueing, but let's try to keep this here untouched and instead issue cleanup internally inside `update_state`.

Consequently the function `remove_expired_reservoir_rules` should become private.

relay-server/src/actors/project.rs
```diff
@@ -421,6 +424,26 @@ impl Project {
...
+    pub fn remove_expired_reservoir_rules(&self) {
+        let Some(config) = self
+            .state
+            .as_ref()
+            .and_then(|state| state.config.dynamic_sampling.as_ref())
+        else {
+            return;
+        };
+
+        if let Ok(mut guard) = self.reservoir_counters.try_lock() {
```
jan-auer, line 443: Using `try_lock` comes with the downside of skipping cleanup in case the rules are being used right now. It is an unlikely case, so it's a fair assumption, but please leave a code comment of why this is so crucial to prevent that someone changes this into a full `lock()` in the future. 

The only way to lock this here is using a tokio Mutex, but before we do that we should explore other approaches.

relay-sampling/src/evaluation.rs
jan-auer: You can run a proper `lock` here. Since we're in sync context, we're allowed to block now. However, I'd split this method into three logical sections: 

 1. (locked) check the cached entry and return early if we've exceeded our budget
 2. (unlocked) run the redis increment and obtain a new count.
 3. (locked) write the value returned from redis back into the cache
 
 there can be a fast-path if redis is not configured, in which case you just lock once, check, and increment.

relay-sampling/src/config.rs
TBS1996: nice catch yeah, stopped being used when I changed to no allocation removing counters

relay-sampling/src/evaluation.rs
```diff
@@ -24,26 +30,146 @@ fn pseudo_random_from_uuid(id: Uuid) -> f64 {
...
+        #[cfg(feature = "redis")]
+        if let (Some(pool), Some(org_id)) = (self.redis_pool.as_ref(), self.org_id) {
+            let key = ReservoirRuleKey::new(org_id, rule);
+
+            let mut redis_client = pool.client().ok()?;
+            let mut redis_connection = redis_client.connection().ok()?;
+
+            if redis_sampling::set_redis_expiry(&mut redis_connection, &key, rule_expiry).is_err() {
+                relay_log::error!("failed to set redis reservoir rule expiry");
+            }
```
TBS1996: @jan-auer  I wasn't sure what we would do if this returned an error, so I opted for just setting the expiry every time as we mentioned, then we could think about optimizing later.

relay-sampling/src/evaluation.rs
```diff
@@ -24,26 +30,148 @@ fn pseudo_random_from_uuid(id: Uuid) -> f64 {
...
+
+        match map_guard.get_mut(&rule) {
+            Some(value) => *value = new_value,
+            // Logging an error because at this point the value should definitively be here.
+            None => relay_log::error!("failed to retrieve counter entry"),
+        }
+    }
+
+    /// Evaluates a reservoir rule, returning `true` if it should be sampled.
+    pub fn evaluate(&self, rule: RuleId, limit: i64, rule_expiry: Option<&DateTime<Utc>>) -> bool {
```
TBS1996: This function should satisfy all the performance optimizations mentioned. 

1. it gets the local count and increments it if limit hasn't been reached.
2. fast return if its above the limit.
3. if redis isnt configured, we return early
4. if redis is configured but the received value is less than local value, we avoid locking again
5. only if we receive from redis, and it is higher, do we update the count by locking the mutex again

relay-sampling/src/evaluation.rs
```diff
@@ -24,26 +30,148 @@ fn pseudo_random_from_uuid(id: Uuid) -> f64 {
...
+                Ok(redis_val) => return Some(redis_val),
+                Err(e) => relay_log::error!("failed to increment redis reservoir count: {}", e),
+            }
+        }
+
+        None
+    }
+
+    /// Gets the local count of a reservoir rule. Increments the count if limit isnt reached.
+    fn local_count(&self, rule: RuleId, limit: i64) -> Option<i64> {
```
TBS1996: not sure whats worse, returning an option when it should be a result, or having`LockResult<MutexGuard<'_, BTreeMap<RuleId, i64>>>` in the signature

relay-sampling/src/evaluation.rs
```diff
@@ -24,26 +30,148 @@ fn pseudo_random_from_uuid(id: Uuid) -> f64 {
...
+    #[allow(unused_variables)]
+    fn redis_count(&self, rule: RuleId, rule_expiry: Option<&DateTime<Utc>>) -> Option<i64> {
+        #[cfg(feature = "redis")]
+        if let (Some(pool), Some(org_id)) = (self.redis_pool.as_ref(), self.org_id) {
+            let key = ReservoirRuleKey::new(org_id, rule);
+
+            let mut redis_client = pool.client().ok()?;
+            let mut redis_connection = redis_client.connection().ok()?;
+
+            if redis_sampling::set_redis_expiry(&mut redis_connection, &key, rule_expiry).is_err() {
```
jan-auer: This runs `EXPIRE` on a key that potentially does not exist. I could not figure out in EXPIRE docs what happens in this case. To be safe -- even though it's a theoretical race condition -- let's please increment first and only then run expire.

relay-sampling/src/evaluation.rs
jan-auer: We could log and ignore the error on EXPIRE like you do here, and handle the increment error gracefully. The more correct version, however, would be: Encapsulate the two calls into a single function that throws an error, and if the error occurs simply do _not_ match the rule. 

That would mean if Redis is unavailable we skip all reservoir rules and apply other matching rules. A graceful and predictable fallback behavior that does not cause excess indexing.

relay-sampling/src/evaluation.rs
```diff
@@ -24,26 +30,148 @@ fn pseudo_random_from_uuid(id: Uuid) -> f64 {
...
+    }
+
+    /// Evaluates a reservoir rule, returning `true` if it should be sampled.
+    pub fn evaluate(&self, rule: RuleId, limit: i64, rule_expiry: Option<&DateTime<Utc>>) -> bool {
+        let Some(incremented_local_count) = self.local_count(rule, limit) else {
+            relay_log::error!("failed to read local reservoir count");
+            return false;
+        };
+
+        if incremented_local_count >= limit {
```
jan-auer: There's an edge case you could handle better: Assume a limit _N_. Counter is _N - 1_. After running `local_count`, the counter is _N_, and you do not increment Redis, which you should, however.

A fixed version of this would:
- Check the local count without incrementing, bail if _count >= N_.
- Increment redis
- Get the new count
- Return _count <= N_

relay-server/src/actors/processor.rs
```diff
@@ -533,6 +539,8 @@ pub struct EnvelopeProcessorService {
 
 struct InnerProcessor {
     config: Arc<Config>,
+    #[cfg(feature = "processing")]
+    redis_pool: Option<Arc<RedisPool>>,
```
jan-auer: Just noticed this now, the `RedisPool` can be cloned and is internally arc'ed. You can safely remove the `Arc`.

Otherwise, the clone below in the constructor would be invalid.

relay-server/src/actors/processor.rs
```diff
@@ -1352,6 +1363,14 @@ impl EnvelopeProcessorService {
         //  2. The DSN was moved and the envelope sent to the old project ID.
         envelope.meta_mut().set_project_id(project_id);
 
+        #[allow(unused_mut)]
+        let mut reservoir = ReservoirEvaluator::new(reservoir_counters);
+        #[cfg(feature = "processing")]
+        if let Some(redis_pool) = self.inner.redis_pool.as_ref() {
+            let org_id = managed_envelope.scoping().organization_id;
+            reservoir = reservoir.set_redis(org_id, redis_pool.clone());
```
jan-auer: _nit_: Since we need this in a branch, what if we change the signature to `set_redis(&mut self, ...)` and no return value?

relay-sampling/src/evaluation.rs
TBS1996: ok, theres an edge case if incrementing works but setting redis expiry doesn't in that case, but that should be very unlikely. If that happens htough, we'll count up without sampling.

relay-sampling/src/evaluation.rs
TBS1996: Thank you! I should have been more attentive. 

about your solution, if we check local count without incrementing, then we would have to access the lock once more in the end of the function to update it in case of no redis update.

I changed it now to address the edge case without any additional locking.

relay-server/src/actors/processor.rs
TBS1996: like putting reference to redis pool in the evaluator ? It seems I have to use `PhantomData` for that since the lifetimes are only used under redis feature flag

relay-sampling/src/config.rs
```diff
@@ -174,7 +192,7 @@ pub enum RuleType {
 ///
 /// This number must be unique within a Sentry organization, as it is recorded in outcomes and used
 /// to infer which sampling rule caused data to be dropped.
-#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
TBS1996, line 195: needed for btreemap

jan-auer (approved):

PR #2552, opened 2023-09-28
feat(spans): Scrub IDs in SQL table names

jjbayer:
Automatically generated table names can cause high cardinality for span metrics & deteriorate the user experience by creating a separate group for every generated table.

Use a regex to scrub obviously generated names:

1. Numbers with more than one digit.
2. 32-digit hexadecimals a.k.a. UUIDs.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/sql/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/mod.rs
@@ -665,7 +665,7 @@ mod tests {
         "/*some comment `my_function'*/ SELECT * FROM foo WHERE ...",
         "SELECT * FROM foo WHERE ..."
     );
- 
+
     scrub_sql_test!(
         jsonb,
         r#"SELECT * FROM "a" WHERE "a"."b" = $1 AND (obj->'id' <@ '[123]'::jsonb) AND "a"."b" != $2"#,
@@ -694,6 +694,7 @@ mod tests {
         uuid_in_table_name,
         "SELECT * FROM prefix_0123456789abcdef0123456789ABCDEF_006_suffix",
         "SELECT * FROM prefix_00_00_suffix"
+    );
 
     scrub_sql_test!(
         clickhouse,
--- a/relay-event-normalization/src/normalize/span/description/sql/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/mod.rs
@@ -667,6 +667,30 @@ mod tests {
     );
 
     scrub_sql_test!(
+        digits_in_table_name,
+        "SELECT * FROM temp_12c",
+        "SELECT * FROM temp_00c"
+    );
+
+    scrub_sql_test!(
+        single_digit_in_table_name,
+        "SELECT * FROM temp_1c",
+        "SELECT * FROM temp_1c"
+    );
+
+    scrub_sql_test!(
+        digits_in_table_name,
+        r#"SELECT * FROM "foo".temp_12c"#,
+        "SELECT * FROM temp_00c"
+    );
+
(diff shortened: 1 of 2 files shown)


relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -672,6 +672,30 @@ mod tests {
...
+    scrub_sql_test!(
+        digits_in_compound_table_name,
+        r#"SELECT * FROM "foo"."temp_12c""#,
+        "SELECT * FROM temp_00c"
+    );
+
+    scrub_sql_test!(
+        uuid_in_table_name,
+        "SELECT * FROM prefix_0123456789abcdef0123456789ABCDEF_006_suffix",
+        "SELECT * FROM prefix_00_00_suffix"
```
jjbayer: @gggritso let me know what you think of the placeholder here. `00` is not very explicit, but I like that it's short and still vaguely implies that some generated number was here. I'm open to using a different placeholder though.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -150,23 +155,26 @@ impl NormalizeVisitor {
...
         if let Some(mut last) = parts.pop() {
-            last.quote_style = None;
+            Self::scrub_name(&mut last);
             *parts = vec![last];
         }
     }
 
     fn scrub_name(name: &mut Ident) {
+        name.quote_style = None;
+        if let Cow::Owned(s) = TABLE_NAME_REGEX.replace_all(&name.value, "00") {
```
olksdr: what if the name would contain something like `XXX` instead of `00` ?

olksdr (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -15,6 +18,8 @@ use sqlparser::dialect::{Dialect, GenericDialect};
 /// because not everything in SQL is an expression.
 const MAX_EXPRESSION_DEPTH: usize = 64;
 
+static TABLE_NAME_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i)[0-9a-f]{32}|\d\d+").unwrap());
```
iker-barriocanal, line 22: Should we add some docs?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/description/sql/mod.rs
gggritso: IMO any of the known placeholder strings would be great here! `"%s` is my favourite, since it's so familiar, but even `"?"` or others

jjbayer: Went for `{%s}` now, because `foo{%s}bar` is clearer than `foo%sbar` IMO. Let me know what you think!

https://github.com/getsentry/relay/blob/54dededf7c18c2fc49fa23f84ef0659bfc99e163/relay-event-normalization/src/normalize/span/description/sql/mod.rs#L406

https://github.com/getsentry/relay/blob/54dededf7c18c2fc49fa23f84ef0659bfc99e163/relay-event-normalization/src/normalize/span/description/sql/mod.rs#L678

https://github.com/getsentry/relay/blob/54dededf7c18c2fc49fa23f84ef0659bfc99e163/relay-event-normalization/src/normalize/span/description/sql/mod.rs#L696

gggritso: Makes sense!

PR #2553, opened 2023-09-28
docs(outcomes): Fix typo

iker-barriocanal:
#skip-changelog

### DIFF
--- a/relay-server/src/actors/outcome.rs
+++ b/relay-server/src/actors/outcome.rs
@@ -707,7 +707,7 @@ impl KafkaOutcomesProducer {
     /// Creates and connects the Kafka producers.
     ///
     /// If the given Kafka configuration parameters are invalid, or an error happens during
-    /// connecting during the broker, an error is returned.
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/outcome.rs
```diff
@@ -707,7 +707,7 @@ impl KafkaOutcomesProducer {
     /// Creates and connects the Kafka producers.
     ///
     /// If the given Kafka configuration parameters are invalid, or an error happens during
-    /// connecting during the broker, an error is returned.
+    /// connecting the broker, an error is returned.
```
olksdr:
```suggestion
    /// connecting to the broker, an error is returned.
```

olksdr (approved):

relay-server/src/actors/outcome.rs
jjbayer: Let's make it "while connecting the broker" then:

> We use during before nouns and noun phrases but not before verbs

https://dictionary.cambridge.org/grammar/british-grammar/during

Issue #2494, opened 2023-09-07
Make transaction spans great


jjbayer:
Since https://github.com/getsentry/relay/pull/2417, we extract a span not only for every span object contained in a transaction, but also for the transaction itself. This span is also known as a "segment", and has the flag `is_segment` = 1.

However, these transaction spans lack two features:

1. They do not have a `description`. Unclear what `description` value to set for segments, but the transaction name is an obvious candidate.
2. We do not extract span metrics for them, and also do not materialize any of the metrics tags into their payloads.

The difference in data quality between span-from-span and span-from-transaction can be seen clearly in the following test case: https://github.com/getsentry/relay/blob/b06f04e4d1a7df23a9a1aafb23f7e98944d6f2a2/tests/integration/test_store.py#L1213-L1262

jjbayer:
cc @phacops

jjbayer:
Upgrading this to P1 because the `usage` metric is also not extracted for the transaction span, and it's used for cost estimations / future billing decisions.

PR #2555, opened 2023-09-29
ref(spans): Write sentry tags to dedicated field

### LINKED ISSUE #2412: Materialize span metric tags into dedicated field
Currently, we write all tags that are used on span metrics into the `span.data` data bag. Consumers of this data bag cannot know which keys in `span.data` are metric tags and which are user fields. There might even be conflicts if the user uses the same key.

Another disadvantage of this approach is that these tags show up in the Sentry UI.

Instead of writing to `span.data`, we should write to a separate top-level field in the span, e.g. `metric_tags`.

* This requires a double write until all consumers have been updated to use the new format.
* Update the sentry codebase to make sure that the new field does not lead to error messages in the UI.

jjbayer:
Up until now, we've materialized span tags into `span.data`. This creates several problems:

* Potential clashes between keys set by the user and keys set by sentry.
* Pollution of the UI because `span.data` entries show up in the event details view.
* Most importantly: `span.data` is [subject to PII scrubbing](https://github.com/getsentry/relay/pull/1953), so tags computed by Relay are sometimes overwritten by PII scrubbing in the next Relay instance.

Instead of writing into `span.data`, create a new top-level object in the span called `sentry_tags`. The same naming has been used [here](https://github.com/getsentry/sentry-kafka-schemas/blob/79fb0900e3a9a4da6f0db15eab1b5d27f42ffeb7/schemas/snuba-spans.v1.schema.json#L81-L83).

For now, double-write span tags into both `span.data` and `span.sentry_tags` until all users of these tags have switched to `sentry_tags`.

Fixes #2412


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -55,34 +55,63 @@ pub enum SpanTagKey {
 }
 
 impl SpanTagKey {
-    /// Whether or not this tag should be added to metrics extracted from the span.
-    pub fn is_metric_tag(&self) -> bool {
-        !matches!(self, SpanTagKey::Release | SpanTagKey::User)
+    /// The key used to write this tag into `span.data`.
+    ///
+    /// This key corresponds to the tag key on span metrics.
+    // NOTE: This method can be removed once we stop double-writing span tags.
```
olksdr:
```suggestion
    /// NOTE: This method can be removed once we stop double-writing span tags.
```

olksdr (approved):
@jjbayer  you mentioned that current tags are subject to PII scrubbing. 
Don't we want to PII scrub these tags as well, even though they are generated by Relay?

relay-event-schema/src/protocol/span.rs
```diff
@@ -67,6 +67,9 @@ pub struct Span {
     #[metastructure(pii = "true")]
     pub data: Annotated<Object<Value>>,
 
+    /// Tags generated by Relay. These tags are a superset of the tags set on span metrics.
+    pub sentry_tags: Annotated<Object<String>>,
```
iker-barriocanal, line 71: Should we run PII here?

relay-server/src/actors/processor.rs
```diff
@@ -2337,6 +2337,15 @@ impl EnvelopeProcessorService {
         if all_modules_enabled {
             // Extract tags to add to this span as well
             let shared_tags = span::tag_extraction::extract_shared_tags(event);
+            transaction_span.sentry_tags = Annotated::new(
+                shared_tags
+                    .clone()
+                    .into_iter()
+                    .map(|(k, v)| (k.sentry_tag_key().to_owned(), Annotated::new(v)))
+                    .collect(),
+            );
```
iker-barriocanal, line 2346: Not related to this PR -- it seems the top-level transaction, converted to a span, only has the shared tags and not span-specific tags. Is this expected?

iker-barriocanal (approved):

relay-event-schema/src/protocol/span.rs
jjbayer, line 71: Note: We would normally have to upgrade the relay library in sentry when adding properties to prevent errors like [this](https://github.com/getsentry/sentry-react-native/issues/1174). But because `Span` still has `retain = true` on the `other` field, upgrading is not necessary in this case.

jjbayer: We have seen cases where the `group` tag got replaced by `[Filtered]` because of PII scrubbing. So no: We want these tags to contain the same values as the span metrics, and exclude them from PII scrubbing deliberately.

relay-server/src/actors/processor.rs
jjbayer: Good observation! It's unclear which span-specific tags a transaction should have. We will come back to this in https://github.com/getsentry/relay/issues/2494.

jjbayer:
> Don't we want to PII scrub these tags as well, even though they are generated by Relay?

@olek we want these tags to contain the same values as the span _metric_ tags, so I would say no. This does raise a follow-up question though: should we scrub metrics data that are sourced from PII-scrubbable fields? Should PII scrubbing run before metrics extraction?

PR #2556, opened 2023-09-29
ref(spans): JSON Kafka message with metadata

jjbayer:
This PR makes the following changes to the relay -> sentry kafka schema for spans:

* [breaking change] Encode the payload as JSON, not msgpack. Msgpack should only be used for payloads that contain (partial) binary data.
* Add the fields `organization_id` and `retention_days` so the consumer does not have to look them up.
* [through #2555] New field `span.sentry_tags` can be used instead of `span.data`. It now contains the same keys as expected by snuba. The values are guaranteed to be strings.
* Some extra validation to make sure we do not send corrupted data.

Sentry counterpart: https://github.com/getsentry/sentry/pull/57284


relay-server/src/actors/store.rs
```diff
@@ -1025,19 +1099,50 @@ struct CheckInKafkaMessage {
...
+    segment_id: Option<String>,
+    /// Whether or not the current span is the root of the segment.
+    is_segment: bool,
+    /// Timestamp when the span started in milliseconds.
+    start_timestamp_ms: u64,
+    /// Total duration of the span in milliseconds.
+    duration_ms: u64,
+    /// The amount of time spent in this span, excluding its immediate child spans.
+    exclusive_time_ms: u64,
+    retention_days: u16,
```
jjbayer, line 1048: @phacops the `retention_days` field in combination with `organization_id` should make the [db lookup in the sentry consumer](https://github.com/getsentry/sentry/blob/808f610f97f78a5ad5df2381efecad642c368f63/src/sentry/spans/consumers/process/factory.py#L58-L60) unnecessary, which will hopefully save some resources.

relay-server/src/actors/store.rs
```diff
@@ -732,31 +738,99 @@ impl StoreService {
...
+                .ok_or(anyhow!("missing is_segment"))?,
+            start_timestamp_ms,
+            duration_ms: timestamp_ms.saturating_sub(start_timestamp_ms),
+            exclusive_time_ms: exclusive_time.0.ok_or(anyhow!("missing exclusive_time"))? as u64,
+            retention_days,
+            user_tags: tags
+                .into_value()
+                .unwrap_or_default()
+                .into_iter()
+                .filter_map(|(k, v)| v.into_value().map(|v| (k, v.to_string())))
```
phacops: Will it remove `null` or empty values?

relay-server/src/actors/store.rs
```diff
@@ -1025,19 +1099,50 @@ struct CheckInKafkaMessage {
...
+    /// transaction.
     #[serde(skip_serializing_if = "Option::is_none")]
-    event_id: Option<EventId>,
+    segment_id: Option<String>,
+    /// Whether or not the current span is the root of the segment.
+    is_segment: bool,
+    /// Timestamp when the span started in milliseconds.
+    start_timestamp_ms: u64,
+    /// Total duration of the span in milliseconds.
+    duration_ms: u64,
```
phacops: This ends up being a `u32` in ClickHouse.

relay-server/src/actors/store.rs
```diff
@@ -1025,19 +1099,50 @@ struct CheckInKafkaMessage {
...
+    /// The amount of time spent in this span, excluding its immediate child spans.
+    exclusive_time_ms: u64,
```
phacops: This ends up as a `f64` in ClickHouse.

phacops (commented):
Can you add `profile_id` as well? Only filled if the `profile` context contains a `profile_id` key. Should be filled for every spans generated by the transaction (and the segment as well).

phacops (commented):
I noticed some values missing from before or which could be added to help:
- `description`
- `start_timestamp_ms`

Is `group` in the `sentry_tags` guaranteed to be a good value (convertible to an hex value)?

And regarding `tags` and `sentry_tags`, can we make sure we don't add a `None` value?

relay-server/src/actors/processor.rs
```diff
@@ -2360,6 +2367,53 @@ impl EnvelopeProcessorService {
...
+                        true
+                    }
+                }
+                // Drop empty string values.
+                None => false,
+            });
+        }
+        if let Some(tags) = tags.value_mut() {
+            tags.retain(|_, value| !value.is_empty())
+        }
```
jjbayer: @phacops this should give some additional guarantees about the data.

CHANGELOG.md
```diff
@@ -13,11 +13,12 @@
 
 - Exclude more spans fron metrics extraction. ([#2522](https://github.com/getsentry/relay/pull/2522))
 - Remove filtering for Android events with missing close events. ([#2524](https://github.com/getsentry/relay/pull/2524))
-- Exclude more spans fron metrics extraction. ([#2522](https://github.com/getsentry/relay/pull/2522), [#2525](https://github.com/getsentry/relay/pull/2525), [#2545](https://github.com/getsentry/relay/pull/2545))
+pull/2522), [#2525](https://github.com/getsentry/relay/pull/2525), [#2545](https://github.com/getsentry/relay/pull/2545))
```
olksdr: Something strange here, looks like it got broken, maybe on `master` merge?

olksdr (approved):
Please, fix the changelog. Otherwise lgtm

relay-server/src/actors/processor.rs
```diff
@@ -2360,6 +2367,53 @@ impl EnvelopeProcessorService {
...
+        exclusive_time
+            .value()
+            .ok_or(anyhow::anyhow!("missing exclusive_time"))?;
+
+        if let Some(sentry_tags) = sentry_tags.value_mut() {
+            sentry_tags.retain(|key, value| match value.value() {
+                Some(s) => {
+                    if key == "group" {
+                        // Only allow 16-char hex strings in group.
+                        s.len() == 16 && s.chars().all(|c| c.is_ascii_hexdigit())
```
iker-barriocanal: Q: why not longer/shorter strings?

relay-server/src/actors/store.rs
```diff
@@ -1026,25 +1034,30 @@ struct CheckInKafkaMessage {
...
+    /// The numeric ID of the organization.
+    organization_id: u64,
+    /// The numeric ID of the project.
+    project_id: ProjectId,
+    /// Number of days until these data should be deleted.
+    retention_days: u16,
+    /// Fields from the original span payload.
+    /// See [`relay-event-schema::protocol::span::Span`] for schema.
+    ///
+    /// By using a [`RawValue`] here, we can embed the span's JSON without additional parsing.
```
iker-barriocanal, line 1052: 👍

tests/integration/fixtures/processing.py
```diff
@@ -437,3 +426,12 @@ def get_check_in(self):
         wrapper = msgpack.unpackb(message.value(), raw=False, use_list=False)
         assert wrapper["type"] == "check_in"
         return json.loads(wrapper["payload"].decode("utf8")), wrapper
+
+
+class SpansConsumer(ConsumerBase):
+    def get_span(self):
+        message = self.poll()
```
iker-barriocanal, line 433: Should we use the timeout here? Calling the `spans_consumer` with a timeout and it not being used may generate some confusion.

iker-barriocanal (approved):
Is a new release of the sentry kafka schemas library required before deploying this?

relay-server/src/actors/processor.rs
iker-barriocanal: If tags is not `None` but it's empty, is it forwarded as an empty object?

relay-server/src/actors/processor.rs
phacops: Yes, I think this is a mistake. We are converting from an hex string into a `u64` but the length of the hex string can be 8 I believe.

phacops (commented):
Is there still a possibility for `status` and `op` to be in `sentry_tags` and at the span top-level and have a different value? Would be good to clean this up and have only 1 value, wherever it's easier to read from.

phacops:
> Is a new release of the sentry kafka schemas library required before deploying this?

No, we haven't written one yet for `ingest-spans`. When this schema is stabilized (after this PR), I'll write one and start using it in the Sentry consumer.

phacops:
And to confirm, we're not removing the `span.` prefix from `sentry_tags` key names, correct?

phacops:
Can you also validate `span.status_code` is a valid integer even if we're going to send it as a string?

relay-server/src/actors/processor.rs
jjbayer: The group hash is a 64-bit hash, i.e. 8 bytes, but it takes two hex characters to encode a byte. So the group hash must always be 16 characters long. I'm fine with relaxing the condition to `<= 16` though.

https://github.com/getsentry/relay/blob/e27eff09c32f2604e5b9e4533d891dabdd52978a/relay-event-normalization/src/normalize/span/tag_extraction.rs#L339-L341

relay-server/src/actors/processor.rs
jjbayer: Probably. The consumer should be able to handle that though.

tests/integration/fixtures/processing.py
jjbayer: See https://github.com/getsentry/relay/blob/415e5cb803338486ac312f4dff698666177d3ec3/tests/integration/fixtures/processing.py#L160-L162

CHANGELOG.md
jjbayer: 👍 fixed now

relay-server/src/actors/processor.rs
iker-barriocanal: [`String::len` returns bytes](https://doc.rust-lang.org/std/string/struct.String.html#method.len), so 16 bytes * 2 chars / byte = 32 chars?

jjbayer:
@phacops 

> Can you also validate span.status_code is a valid integer even if we're going to send it as a string?

Done.

> Is there still a possibility for status and op to be in sentry_tags and at the span top-level and have a different value? Would be good to clean this up and have only 1 value, wherever it's easier to read from.

Removed `status` from `sentry_tags` now because it's an exact copy of the top-level status. Keeping `op` for now because it does a lowercase transform: https://github.com/getsentry/relay/blob/b45dce843967b44d2c174f561e13e2ff4a970ca3/relay-event-normalization/src/normalize/span/tag_extraction.rs#L235-L238

relay-server/src/actors/processor.rs
jjbayer: 16 is the length of the hexidecimal representation, e.g. `0123456789abcdef`. But it still only encodes 8 bytes of data: `01 23 45 67 89 ab cd ef`.

phacops (approved):

PR #2557, opened 2023-09-30
ref(sampling): Refactor sampling evaluator

TBS1996:
A follow-up to #2550. 

The main motivation for this PR is that the `SamplingRule::Evaluate` method became pretty ugly because it relied on that the two variants of `SamplingValue` both had a sample rate attached, which it no longer does with the introduction of the `Reservoir` variant. 

The function has mainly been replaced by `SamplingEvaluator::try_compute_sample_rate`. Whose job it is to validate the rule, and if it's valid, return a sample rate. 

Validating for the `SamplingRate` and `Factor` variant means that it's not out of bounds with the given time range (depending on the decaying function), and for the `Reservoir` variant it means that the limit has not been exceeded. 

It will return an optional `ControlFlow`, where `None` means the rule is invalid and should be skipped, `Break` is analogous to `SamplingValue::SampleRate` (but includes reservoir), and `Continue` is analogous to `SamplingValue::Factor`. 

Some other adjustments have also been made, this should be all the changes:

* Created the `try_compute_sample_rate` method as described.
* Moved the decaying function logic to its own method.
* Moved the checking of time range constraints to the beginning of for loop, so we don't do the more expensive condition-matching first. 
* Updated tests

#skip-changelog


relay-sampling/src/evaluation.rs
```diff
@@ -195,19 +195,21 @@ impl<'a> SamplingEvaluator<'a> {
...
+            if !rule.time_range.contains(self.now) || !rule.condition.matches(instance) {
                 continue;
             };
 
-            let Some(sampling_value) = rule.evaluate(self.now, self.reservoir) else {
+            let Some(sampling_value) = rule.adjusted_sampling_value(self.now) else {
                 continue;
             };
 
-            self.rule_ids.push(rule.id);
```
TBS1996, line 206: I had to move this into the arms of `Factor` and `SampleRate` because previously if a reservoir rule didn't match the `rule.evaluate` would have returned None and continued in the lines above. 

If I left it here, the reservoir rules would have been part of the matched_ids for a `SampleRate` match

relay-sampling/src/config.rs
```diff
@@ -82,57 +81,19 @@ impl SamplingRule {
...
-                };
-
-                (sampling_base_value > decayed_value).then_some(())?;
-
-                let now = now.timestamp() as f64;
-                let start = start.timestamp() as f64;
-                let end = end.timestamp() as f64;
+    /// Returns an adjusted [`SamplingValue`] based on the [`DecayingFunction`].
+    pub fn adjusted_sampling_value(&self, now: DateTime<Utc>) -> Option<SamplingValue> {
+        let mut sampling_value = self.sampling_value;
```
TBS1996, line 86: its a `copy` type

relay-sampling/src/evaluation.rs
```diff
@@ -195,19 +195,21 @@ impl<'a> SamplingEvaluator<'a> {
         I: Iterator<Item = &'b SamplingRule>,
     {
         for rule in rules {
-            if !rule.condition.matches(instance) {
+            if !rule.time_range.contains(self.now) || !rule.condition.matches(instance) {
                 continue;
             };
 
-            let Some(sampling_value) = rule.evaluate(self.now, self.reservoir) else {
+            let Some(sampling_value) = rule.adjusted_sampling_value(self.now) else {
```
TBS1996, line 202: at this point, it is no longer determined if there is a match or not, as it is only determined in the following match statement in the case of the reservoir variant.

Perhaps that's considered bad design, it's hard to make it ideal within the current schema.

relay-sampling/src/config.rs
```diff
@@ -512,19 +499,28 @@ mod tests {
...
-            condition: RuleCondition::all(),
-            sampling_value: SamplingValue::SampleRate { value: 1.0 },
-            ty: RuleType::Trace,
-            id: RuleId(0),
-            time_range: TimeRange {
-                start: Some(Utc.with_ymd_and_hms(1970, 10, 10, 0, 0, 0).unwrap()),
-                end: Some(Utc.with_ymd_and_hms(1970, 10, 12, 0, 0, 0).unwrap()),
-            },
-            decaying_fn: DecayingFunction::Linear { decayed_value: 0.5 },
+    fn test_decay_fn_linear() {
```
TBS1996, line 517: Now that decaying function logic is a method on the decayingfunction type, we can unit test it more directly.

relay-sampling/src/config.rs
```diff
@@ -533,150 +529,36 @@ mod tests {
...
-            rule.time_range.end = None;
-            rule
-        };
-
-        assert!(rule_without_end.evaluate(halfway, None).is_none());
-    }
-
-    /// If the decayingfunction is set to `Constant` then it shouldn't adjust the sample rate.
-    #[test]
-    fn test_sample_rate_with_constant_decayingfn() {
```
TBS1996, line 575: replaced with `test_decay_fn_constant`

relay-sampling/src/config.rs
```diff
@@ -533,150 +529,36 @@ mod tests {
...
-        let halfway = Utc.with_ymd_and_hms(1970, 10, 11, 0, 0, 0).unwrap();
-
-        assert_eq!(rule.evaluate(halfway, None), Some(sampling_value));
-    }
-
-    /// Validates the `sample_rate` method for different time range configurations.
-    /// The method should return `None` for times outside of the valid range and `Some` for times within it.
-    /// When the `start` or `end` of the range is missing, it defaults to always include times before the `end` or after the `start`, respectively.
-    #[test]
-    fn test_sample_rate_valid_time_range() {
```
TBS1996, line 599: moved to `evaluation.rs` because we now check if the time range is valid in the `match_rules` function for performance reasons.

relay-sampling/src/evaluation.rs
```diff
@@ -195,32 +229,30 @@ impl<'a> SamplingEvaluator<'a> {
         I: Iterator<Item = &'b SamplingRule>,
     {
         for rule in rules {
-            if !rule.condition.matches(instance) {
+            if !rule.time_range.contains(self.now) || !rule.condition.matches(instance) {
```
TBS1996, line 198: moving the time range constraint here makes no difference in the logic, it only improves performance. Previously it was done right after in SamplingRule::evaluate

relay-sampling/src/config.rs
```diff
@@ -255,6 +207,39 @@ pub enum DecayingFunction {
     Constant,
 }
 
+impl DecayingFunction {
```
TBS1996, line 210: by the way, the time_range field on `SamplingRule` says the following:

>If a time range is closed on at least one end, the rule is considered a decaying rule.  
 
Yet, decaying function is only run if both ends are closed, is there a contradiction here?

jan-auer (approved):

PR #2558, opened 2023-10-02
chore(config): Remove TLS from generated config

olksdr:
Relay does not support TLS and the suggestion is to use the proxy in front of it for this use case. 

This means we should not include any mentions about TLS into the generated config. 

#skip-changelog

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -441,10 +441,13 @@ pub struct Relay {
     /// The port to bind for the unencrypted relay HTTP server.
     pub port: u16,
     /// Optional port to bind for the encrypted relay HTTPS server.
+    #[serde(skip_serializing)]
     pub tls_port: Option<u16>,
     /// The path to the identity (DER-encoded PKCS12) to use for TLS.
+    #[serde(skip_serializing)]
     pub tls_identity_path: Option<PathBuf>,
     /// Password for the PKCS12 archive.
+    #[serde(skip_serializing)]
     pub tls_identity_password: Option<String>,
     /// Always override project IDs from the URL and DSN with the identifier used at the upstream.
     ///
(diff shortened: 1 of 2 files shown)


relay-config/src/config.rs
```diff
@@ -441,10 +441,13 @@ pub struct Relay {
...
     pub port: u16,
     /// Optional port to bind for the encrypted relay HTTPS server.
+    #[serde(skip_serializing)]
     pub tls_port: Option<u16>,
     /// The path to the identity (DER-encoded PKCS12) to use for TLS.
+    #[serde(skip_serializing)]
     pub tls_identity_path: Option<PathBuf>,
     /// Password for the PKCS12 archive.
+    #[serde(skip_serializing)]
     pub tls_identity_password: Option<String>,
```
jjbayer, line 451: If we don't support them, can we remove these options?

jjbayer (approved):

relay-config/src/config.rs
olksdr: So, the one reason I still kept this, because of the old Relays  - in-case they will get updated and we want to ensure the message is delivered that TLS is not supported anymore, see https://github.com/getsentry/relay/blob/4eecaba5db4d076bfc66552fdde0fd2aff27f3c7/relay-server/src/actors/server.rs#L50-L55

But I'm also fine if we ok with removing all this altogether. I can include the cleanup in my PR as well.

iker-barriocanal: If we keep the config to indicate they are no longer supported, I'd log a message to indicate it to users. If not, I'd remove the options too.

iker-barriocanal (approved):

relay-config/src/config.rs
jjbayer: @olksdr I was missing that piece of context, I'm fine with keeping the options then!

PR #2559, opened 2023-10-03
instr(metrics): Tag aggregator

jjbayer:
We've been seeing more invalid timestamps in the metrics aggregator over the past weeks. The timing suggests these are span metrics, so tag metrics and errors with the aggregator name (`"spans"` vs `"default"`) to verify.

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregation.rs
+++ b/relay-metrics/src/aggregation.rs
@@ -340,10 +340,6 @@ impl AggregatorConfig {
         let output_timestamp = UnixTimestamp::from_secs(ts);
 
         if !self.timestamp_range().contains(&output_timestamp) {
-            let delta = (ts as i64) - (UnixTimestamp::now().as_secs() as i64);
-            relay_statsd::metric!(
-                histogram(MetricHistograms::InvalidBucketTimestamp) = delta as f64,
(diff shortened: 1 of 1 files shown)


relay-metrics/src/aggregation.rs
```diff
@@ -920,6 +916,28 @@ impl AggregatorService {
...
+    ) -> Result<UnixTimestamp, AggregateMetricsError> {
+        let res = self.config.get_bucket_timestamp(timestamp, bucket_width);
+        if let Err(AggregateMetricsError {
+            kind: AggregateMetricsErrorKind::InvalidTimestamp(ts),
+        }) = res
+        {
+            let delta = (ts.as_secs() as i64) - (UnixTimestamp::now().as_secs() as i64);
+            relay_statsd::metric!(
+                histogram(MetricHistograms::InvalidBucketTimestamp) = delta as f64,
+                aggregator = &self.name,
```
jjbayer, line 934: This is the most relevant change: With this tag, we should be able to verify that the rise in `InvalidTimestamp` errors can be explained by span metrics.

iker-barriocanal (approved):

PR #2560, opened 2023-10-03
ref(spans): Scrub data image descriptions

iker-barriocanal:
In span descriptions fetching encoded data images, the actual encoded image provides no value to the span. However, the number of `/`s is forcing the span description clusterer to misbehave.

Resolves https://github.com/getsentry/team-ingest/issues/221.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Scrub span descriptions with encoded data images. ([#2560](https://github.com/getsentry/relay/pull/2560))
+
 **Bug Fixes**:
 
 - Remove profile_id from context when no profile is in the envelope. ([#2523](https://github.com/getsentry/relay/pull/2523))
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Scrub span descriptions with encoded data images. ([#2560](https://github.com/getsentry/relay/pull/2560))
+
 ## 0.8.30
 
 - Filter out exceptions originating in Safari extensions. ([#2408](https://github.com/getsentry/relay/pull/2408))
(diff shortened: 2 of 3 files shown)


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -367,6 +371,13 @@ mod tests {
         "GET *"
     );
 
+    span_description_test!(
+        span_description_scrub_data_images,
+        "GET data:image/png;base64,drtfghaksjfdhaeh/blah/blah/blah",
+        "http.client",
+        "GET data:image/*"
```
iker-barriocanal, line 378: We can relax the scrubbing to include `png` or `base64`, but I don't think it provides any value and complicates a bit the logic.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Scrub span descriptions with encoded data images. ([#2560](https://github.com/getsentry/relay/pull/2560))
```
olksdr, line 7: I would omit this changelog records to be honest. And it's not really a feature in my opinion, maybe something like internal fits better here.

py/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Scrub span descriptions with encoded data images. ([#2560](https://github.com/getsentry/relay/pull/2560))
```
olksdr, line 5: Same here.

olksdr (approved):
Looks good to me

CHANGELOG.md
iker-barriocanal: Initially, I was planning to omit it. I saw the [changelog section in the README](https://github.com/getsentry/relay?tab=readme-ov-file#instructions-for-changelogs) and some scrubbing related entries in both features and internal. Since this is user-visible I'll leave it under features, unless someone strongly disagrees.

PR #2562, opened 2023-10-03
feat(server): Delay metrics extraction for indexed transactions

TBS1996:
ref #2083 

We want to delay extracting metrics in pop relays if sampling decision is 'keep' in order to synchronize rate limits on indexed and processed transactions. 

#skip-changelog


tests/integration/test_metrics.py
```diff
@@ -1241,35 +1240,36 @@ def test_generic_metric_extraction(mini_sentry, relay):
...
+            }
+        ],
+    }
     transaction = generate_transaction_item()
     timestamp = datetime.now(tz=timezone.utc)
     transaction["timestamp"] = timestamp.isoformat()
     transaction["start_timestamp"] = (timestamp - timedelta(seconds=2)).isoformat()
-
-    # Explicitly test a chain of Relays
-    relay = relay(relay(mini_sentry, options=TEST_CONFIG), options=TEST_CONFIG)
```
jjbayer: We should keep this.

TBS1996: im just messing around because the testing stopped working locally a bit but i fixed it now. 

anyway the tests are a bit tricky now, we can discuss it tomorrow

tests/integration/test_metrics.py
```diff
@@ -539,20 +539,20 @@ def test_session_metrics_processing(
...
         (False, "transaction", False),
         (False, False, False),
+        (False, False, True),
         ("corrupted", "transaction", False),
     ],
     ids=[
         "extract from transaction-sampled, external relay",
         "extract from transaction-sampled",
         "extract from trace-sampled",
-        "extract from unsampled, external relay",
```
TBS1996, line 552: Now we want the exact opposite behaviour so I simply changed the expectation

tests/integration/test_metrics.py
```diff
@@ -777,23 +777,10 @@ def test_transaction_metrics_count_per_root_project(
     }
 
 
-@pytest.mark.parametrize(
-    "send_extracted_header,expect_extracted_header,expect_metrics_extraction",
-    [(False, True, True), (True, True, False)],
-    ids=["must extract metrics", "mustn't extract metrics"],
-)
 def test_transaction_metrics_extraction_external_relays(
```
TBS1996, line 793: Not sure what to do with this test, we already test metrics extraction in the `test_generic_metric_extraction`.

tests/integration/test_metrics.py
```diff
@@ -1241,35 +1200,40 @@ def test_generic_metric_extraction(mini_sentry, relay):
             }
         ],
     }
+    config["transactionMetrics"] = {"version": 1}
```
TBS1996: in order to emit the metrics in pop relays, we have to sample the envelope

tests/integration/test_metrics.py
```diff
@@ -1241,35 +1200,40 @@ def test_generic_metric_extraction(mini_sentry, relay):
...
-        {
-            "timestamp": int(timestamp.timestamp()),
-            "width": 1,
-            "name": "c:transactions/on_demand@none",
-            "type": "c",
-            "value": 1.0,
-            "tags": {"query_hash": "c91c2e4d"},
-        }
-    ]
+    assert {
```
TBS1996, line 1277: Because we drop the envelope, we get more metrics, so I just check if the expected metrics is among the extracted metrics

tests/integration/test_metrics.py
```diff
@@ -1241,35 +1200,40 @@ def test_generic_metric_extraction(mini_sentry, relay):
...
     transaction["start_timestamp"] = (timestamp - timedelta(seconds=2)).isoformat()
 
-    # Explicitly test a chain of Relays
-    relay = relay(relay(mini_sentry, options=TEST_CONFIG), options=TEST_CONFIG)
+    relay = relay(mini_sentry, options=TEST_CONFIG)
     relay.send_transaction(PROJECT_ID, transaction)
 
     envelope = mini_sentry.captured_events.get(timeout=3)
-    assert envelope.items[0].headers.get("type") == "transaction"
-    assert envelope.items[0].headers.get("metrics_extracted") is True
```
TBS1996, line 1260: We can't do this anymore as we drop the envelope in order to extract metrics

tests/integration/test_metrics.py
```diff
@@ -813,47 +797,19 @@ def test_transaction_metrics_extraction_external_relays(
...
-            m_item_body_sorted[1]["tags"]["transaction"]
-            == "/organizations/:orgId/performance/:eventSlug/"
-        )
-        assert (
-            m_item_body_sorted[0]["name"]
-            == "c:transactions/count_per_root_project@none"
-        )
-        assert m_item_body_sorted[0]["tags"]["transaction"] == "root_transaction"
-        assert m_item_body_sorted[0]["value"] == 1.0
-
```
jjbayer: Can we restore this test case and all its variants by adding a "sample-all" dynamic sampling rule at the beginning?

tests/integration/test_metrics.py
jjbayer: Would be good to negate the assertion here, i.e. none of the items have type "transaction".

tests/integration/test_metrics.py
```diff
@@ -1241,35 +1197,40 @@ def test_generic_metric_extraction(mini_sentry, relay):
...
+    relay = relay(mini_sentry, options=TEST_CONFIG)
     relay.send_transaction(PROJECT_ID, transaction)
 
     envelope = mini_sentry.captured_events.get(timeout=3)
-    assert envelope.items[0].headers.get("type") == "transaction"
-    assert envelope.items[0].headers.get("metrics_extracted") is True
-
     envelope = mini_sentry.captured_events.get(timeout=3)
     item = envelope.items[0]
-    assert item.headers.get("type") == "metric_buckets"
```
jjbayer, line 1264: Is this not still true?

relay-dynamic-config/src/metrics.rs
```diff
@@ -79,7 +79,7 @@ pub struct CustomMeasurementConfig {
 /// Maximum supported version of metrics extraction from transactions.
 ///
 /// The version is an integer scalar, incremented by one on each new version.
-const TRANSACTION_EXTRACT_VERSION: u16 = 2;
+const TRANSACTION_EXTRACT_VERSION: u16 = 3;
```
jjbayer, line 87: Please sync with @jan-auer on this version bump.

jjbayer (changes requested):
Couple of comments on the tests, besides from that LGTM!

relay-dynamic-config/src/metrics.rs
jan-auer: Note that I'm also listing changes to the versions in the doc comment now. Your PR will likely merge first, so I'll take care of listing this change, too.

tests/integration/test_metrics.py
TBS1996: yeah, my bad

relay-dynamic-config/src/metrics.rs
```diff
@@ -81,7 +81,10 @@ pub struct CustomMeasurementConfig {
 /// The version is an integer scalar, incremented by one on each new version:
 ///  - 1: Initial version.
 ///  - 2: Moves `acceptTransactionNames` to global config.
-///  - 3: Emit a `usage` metric and use it for rate limiting.
+///  - 3:
+///      - Emit a `usage` metric and use it for rate limiting.
+///      - Delay metrics extraction for indexed transactions.
```
iker-barriocanal, line 86: Could we provide some more details about what "delay" means?

tests/integration/test_metrics.py
```diff
@@ -786,15 +786,14 @@ def test_transaction_metrics_count_per_root_project(
 
 
 @pytest.mark.parametrize(
-    "send_extracted_header,expect_extracted_header,expect_metrics_extraction",
-    [(False, True, True), (True, True, False)],
```
iker-barriocanal, line 790: Why are we no longer verifying the extracted header? If external relays no longer extract metrics, we should expect the next relay not to get that header.

tests/integration/test_metrics.py
```diff
@@ -808,6 +807,18 @@ def test_transaction_metrics_extraction_external_relays(
     config["transactionMetrics"] = {
         "version": 1,
     }
+    config["transactionMetrics"] = {"version": 1}
```
iker-barriocanal:
```suggestion
```
(duplicated from above line)

tests/integration/test_metrics.py
```diff
@@ -808,6 +807,18 @@ def test_transaction_metrics_extraction_external_relays(
...
+        "rules": [],
+        "rulesV2": [
+            {
+                "id": 1,
+                "samplingValue": {"type": "sampleRate", "value": 0.0},
+                "type": "transaction",
+                "condition": {"op": "and", "inner": []},
+            }
+        ],
+    }
```
iker-barriocanal: A sampling rate of 0.0 means metrics won't be extracted, independently of the changes in this PR. Is this correct? Should we remove the dynamic sampling config? Same with the test below.

Related -- setting the transaction metric extraction version to anything below 3 makes Relay to not extract metrics IIRC.

relay-server/src/actors/processor.rs
```diff
@@ -2689,7 +2689,16 @@ impl EnvelopeProcessorService {
             self.normalize_dsc(state);
             self.filter_event(state)?;
             self.run_dynamic_sampling(state);
-            self.extract_metrics(state)?;
+
+            let is_processing =
+                cfg!(feature = "processing") && self.inner.config.processing_enabled();
```
jan-auer: You could simplify here. Enabling processing without the feature flag will throw errors during startup. Put another way, you can assume that `processing_enabled()` should never return `true` when not compiled with the relevant feature flags.

```suggestion
            let is_processing = self.inner.config.processing_enabled();
```

relay-server/src/actors/processor.rs
```diff
@@ -2689,7 +2689,16 @@ impl EnvelopeProcessorService {
...
+
+            // We avoid extracting metrics if not sampling on pops, in order to synchronize
```
jan-auer: _nit_: Could you refer to a code concept such as the processing mode or some configuration **instead of** our current deployment names ("PoP") for extra clarity?

tests/integration/test_metrics.py
jan-auer: Shouldn't we set this to `"version": 3` then?

relay-server/src/actors/processor.rs
```diff
@@ -2689,7 +2689,16 @@ impl EnvelopeProcessorService {
...
+            // rate limits on indexed and processed transactions.
+            if is_processing || state.sampling_result.should_drop() {
```
jan-auer: If we were strict, you should only skip this on version `3`. However, this also performs correctly if you unconditionally skip here, so I'm +1 on the simplicity.

tests/integration/test_metrics.py
jan-auer: This is redundant, it's the same as the statement exactly above.

```suggestion
```

tests/integration/test_metrics.py
jan-auer: Please keep this open until it's resolved. We need to ensure that the relay chain persists in this test.

tests/integration/test_metrics.py
TBS1996: makes sense yeah! I just copied from other tests

relay-server/src/actors/processor.rs
TBS1996: makes sense, I just inlined it then

iker-barriocanal (commented):
Since we're not bucketing metrics in PoP relays, the number of messages produced to Kafka may increase and consumers may have a hard time with the increased load. Should we worry about this?

The approach of the PR looks good!

relay-server/src/actors/processor.rs
TBS1996: alright, i wrote non-processing

tests/integration/test_metrics.py
jan-auer: Sampling rates are independent of metric extraction, so you can have a rate of 0% and still metrics extracted. Is that documented differently somewhere?

tests/integration/test_metrics.py
TBS1996: because we have to drop the entire event in order to extract metrics, so we can't check the header of the non-existing event

relay-dynamic-config/src/metrics.rs
TBS1996: if im gonna write out the full conditions then I'm not sure how to do that succinctly 🤔 the above comments also do not go into much details, so im not sure how much to write here

jjbayer (approved):

tests/integration/test_metrics.py
iker-barriocanal: Ooops sorry, missed stuff hahah that's the purpose of metrics :D

PR #2563, opened 2023-10-03
ref(server): Refactor handle_process_metrics.

TBS1996:
me and joris came across this function while investigating #2083 and thought it needed a refactor so I took a stab at it. 

The function was pretty hard to read, it also repeated a lot of logic for both types
#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -2758,19 +2758,17 @@ impl EnvelopeProcessorService {
                     }
                 })
                 .collect(),
-            ItemType::MetricBuckets => {
-                match serde_json::from_slice::<Vec<Bucket>>(&item.payload()) {
-                    Ok(buckets) => buckets,
-                    Err(error) => {
-                        relay_log::debug!(
-                            error = &error as &dyn Error,
-                            "failed to parse metric bucket",
-                        );
-                        metric!(counter(RelayCounters::MetricBucketsParsingFailed) += 1);
-                        vec![]
-                    }
+            ItemType::MetricBuckets => match serde_json::from_slice(&item.payload()) {
+                Ok(buckets) => buckets,
+                Err(error) => {
+                    relay_log::debug!(
+                        error = &error as &dyn Error,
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -2741,70 +2741,80 @@ impl EnvelopeProcessorService {
...
+            ClockDriftProcessor::new(sent_at, received).at_least(MINIMUM_CLOCK_DRIFT);
+
+        for bucket in &mut buckets {
+            clock_drift_processor.process_timestamp(&mut bucket.timestamp);
+        }
+
+        relay_log::trace!("inserting metric buckets into project cache");
+        self.inner
+            .project_cache
+            .send(MergeBuckets::new(public_key, buckets));
```
TBS1996, line 2813: The Mergebuckets message is now sent only once, but with all the buckets, is there any implications of this that i am not aware of?

relay-server/src/actors/processor.rs
```diff
@@ -2741,70 +2741,80 @@ impl EnvelopeProcessorService {
...
-            if item.ty() == &ItemType::Statsd {
-                let mut buckets = Vec::new();
-                for bucket_result in Bucket::parse_all(&payload, received_timestamp) {
-                    match bucket_result {
-                        Ok(mut bucket) => {
-                            clock_drift_processor.process_timestamp(&mut bucket.timestamp);
-                            buckets.push(bucket);
-                        }
-                        Err(error) => relay_log::debug!(
+    fn extract_metric_buckets_from_item(
```
TBS1996, line 2744: Looking for recommendation of where to put fn extract_metric_buckets_from_item. I know it's not good to leave it as an associated function. Would perhaps a method on item itself be fitting? I thought maybe methods on our fundamental types shouldn't have a lot of logic, and no logs. But may be wrong.

TBS1996: I wish I could use IntoMetric trait but as the name suggests, it's for a singular Metric (now bucket).

PR #2564, opened 2023-10-04
feat(metrics): Document the supported character set

### LINKED ISSUE #2531: Specify allowed metrics character set
There are scattered implementations of validation and sanitation for metric names, tag keys, and tag values. We need to establish a specification that covers:

- The legal character sets for the above three
- The legal character sets in statsd payloads
- Escape sequences for statsd if needed
- Sanitation and substitution rules for clients (e.g. replace dashes with underscores)
- Whether or not sanitation should happen in ingest or just on clients

Tasks

- [x] Write the specification in Relay (as part of protocol docs)
- [x] Implement sanitation in Minimetrics
- [ ] ~Update validation in Relay~
- [ ] ~Update validation in the indexer~
- [ ] ~Update validation in Snuba~

jan-auer:
Adds documentation on the allowed character set of metric names, tag keys, and
tag values. The actual parser implementation is much more lenient than what is
documented.

Closes https://github.com/getsentry/relay/issues/2531
#skip-changelog

### DIFF
--- a/relay-metrics/src/bucket.rs
+++ b/relay-metrics/src/bucket.rs
@@ -475,7 +475,7 @@ fn parse_timestamp(string: &str) -> Option<UnixTimestamp> {
 /// <name>[@unit]:<value>[:<value>...]|<type>|#<tag_key>:<tag_value>,<tag>|T<timestamp>
 /// ```
 ///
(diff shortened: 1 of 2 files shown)


iambriccardo (approved):

olksdr (approved):
🚢  it!

PR #2565, opened 2023-10-04
build: Do not block deploys on macOS and Windows tests

jjbayer:
We don't run macOS or Windows in production, so we do not have to block deploys on these platform-specific tests.

#skip-changelog


TBS1996 (approved):
🤔 makes sense

olksdr (approved):
I think this makes sense for our deployment.

jan-auer (approved):

PR #2570, opened 2023-10-05
feat(spans): Enable spans needed for the mobile Starfish module

phacops:
- `app.*` span descriptions are low cardinality.
- `ui.load*` are also scrubbed but not high cardinality (it's a function of the number of methods in an application).
- `file.*` are scrubbed to just get the extension.


relay-server/src/metrics_extraction/event.rs
```diff
@@ -896,6 +896,51 @@ mod tests {
                     "data": {
                         "db.system": "redis"
                     }
+                },
+                {
+                    "data":{
+                        "device.class":"2",
+                        "environment":"production",
+                        "http.status_code":"200",
```
jjbayer: If you want to see this as a tag on the metric, use `http.response.status_code` instead: https://github.com/getsentry/relay/blob/bce54057b84cb1d34bacc8ab91c95d3c7d6be1a2/relay-event-normalization/src/normalize/utils.rs#L15

https://develop.sentry.dev/sdk/performance/span-data-conventions/#http

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics.snap
```diff
@@ -1075,4 +1075,82 @@ expression: metrics
...
+            ],
+        ),
+        tags: {
+            "environment": "fake_environment",
+            "http.status_code": "500",
+            "span.category": "app",
+            "span.op": "app.start.cold",
+            "transaction": "gEt /api/:version/users/",
+            "transaction.method": "POST",
+            "transaction.op": "myop",
```
jjbayer: There is no `description` and no `span.group` here. For that, we need to explicitly allowlist the allowed descriptions in `scrub_span_description`.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -476,6 +489,20 @@ mod tests {
...
+        "data.data (42 KB)",
+        "file.write",
+        "data"
+    );
+
+    span_description_test!(
+        span_description_file_read_keep_extension_only,
+        "Info.plist",
+        "file.read",
+        "plist"
```
jjbayer: @shruthilayaj is this the desired format, or do we need something like `*.plist`?

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -111,6 +113,17 @@ fn scrub_http(string: &str) -> Option<String> {
...
 
+fn scrub_file(description: &str) -> Option<String> {
+    let filename = match description.split_once(' ') {
+        Some((filename, _)) => filename,
+        _ => description,
+    };
+    match Path::new(filename).extension() {
+        Some(extension) => extension.to_os_string().into_string().ok(),
+        _ => None,
+    }
```
jjbayer: I think this saves one allocation

```suggestion
    Some(Path::new(filename).extension()?.to_str()?.into_owned())
```

jjbayer (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
shruthilayaj: yes please! can we do `*.plist`, `*.data` etc

phacops: `*` in case there's no extension?

shruthilayaj: I think we don't extract description for these - so null description, some of the examples i've seen without extensions have had pretty bad descriptions

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics.snap
phacops: I'm confused, I think it's already enabled. Maybe the example comes from before it was enabled though.

relay-event-normalization/src/normalize/span/description/mod.rs
shruthilayaj: sorry or `*` whatever works actually, i didn't read your question properly

PR #2571, opened 2023-10-05
ref(server): Add a dedicated metric for transaction usage

jan-auer:
Adds a dedicated and cheap counter metric for usage tracking. It contains the `has_profile` tag, which will later be removed from the durations metric to lower its cardinality.

This metric should improve the performance of all usage queries. In the long term, it allows to decouple product metrics from bookkeeping.


jan-auer:
This is ready for review while I fix integration tests.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -322,13 +319,24 @@ impl TransactionExtractor<'_> {
...
         // Duration
+        let duration = relay_common::time::chrono_to_positive_millis(end - start);
         metrics.project_metrics.push(
             TransactionMetric::Duration {
                 unit: DurationUnit::MilliSecond,
-                value: relay_common::time::chrono_to_positive_millis(end - start),
+                value: duration,
                 tags: TransactionDurationTags {
-                    has_profile: self.has_profile,
+                    has_profile: self.has_profile, // TODO: Stop producing this in v3
```
iker-barriocanal, line 339: Are we still producing this tag until we update the consumers?

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -38,7 +38,12 @@ impl<Q: AsRef<Vec<Quota>>> MetricsLimiter<Q> {
     /// Create a new limiter instance.
     ///
     /// Returns Ok if `metrics` contain transaction metrics, `metrics` otherwise.
-    pub fn create(buckets: Vec<Bucket>, quotas: Q, scoping: Scoping) -> Result<Self, Vec<Bucket>> {
+    pub fn create(
+        buckets: Vec<Bucket>,
+        quotas: Q,
+        scoping: Scoping,
+        usage: bool,
```
iker-barriocanal, line 45: `usage` may be hard to understand outside this scope. What do you think about making this an enum of two variants (usage and transaction duration)?

iker-barriocanal (approved):

relay-server/src/metrics_extraction/transactions/mod.rs
jjbayer: We have to.

jjbayer (approved):

relay-server/src/metrics_extraction/transactions/mod.rs
jan-auer: There's no functional need to remove the tag, everything works correctly with and without the tag, but as @jjbayer pointed out we need to keep it until the consumers are fully updated.

relay-server/src/utils/metrics_rate_limits.rs
jan-auer: Agreed. The metrics rate limiter overall has a convoluted signature, so I chose the simplest change to follow up with a proper refactoring. I'm not yet sure about whether this should be an enum, the config, or something else, so I'd prefer to stick with the simplest choice. Would that work?

Issue #2450, opened 2023-08-31
`X-Forwarded-For` not working in SaaS


jjbayer:
Steps to reproduce:

Sending the following in my local dev environment:

```shell
curl -v http://localhost:8000/api/2/envelope/ \
     -H X-forwarded-for:4.4.4.4 \
    --data-binary '{"dsn":"$SENTRY_DSN"}\n{"type": "event"}\n{"user": {"ip_address": "{{auto}}"}}'
```

Results in 
![image](https://github.com/getsentry/relay/assets/1565449/f236a18d-58b1-43f1-afd0-e71e3cac1c45)

However, the same does not work for SaaS (ignores the `X-Forwarded-For` header):

```shell
curl -v https://o123.ingest.sentry.io/api/2/envelope/ \
     -H X-forwarded-for:4.4.4.4 \
    --data-binary '{"dsn":"$SENTRY_DSN"}\n{"type": "event"}\n{"user": {"ip_address": "{{auto}}"}}'
```

![image](https://github.com/getsentry/relay/assets/1565449/e35ef25c-ef91-413a-b5ef-0fc9d024f2fd)

jjbayer:
This was an attempt to emulate a [tunnel](https://docs.sentry.io/platforms/javascript/troubleshooting/#using-the-tunnel-option) setup, as discussed here: https://github.com/getsentry/sentry-javascript/discussions/5798#discussioncomment-6867907

jan-auer:
@beezz could we get some help on this one, please? I vaguely remember that we do not pass this header through for a reason, but I cannot find documentation on that.

Note: We need to apply the same policy to [`x-vercel-forwarded-for`](https://vercel.com/docs/edge-network/headers#x-vercel-forwarded-for)

AbhiPrasad:
Synced offline - created an OPs ticket here for the SRE folks to investigate further. https://getsentry.atlassian.net/browse/OPS-3823

iker-barriocanal:
We'll track this in the OPS ticket above.

AaronFriel:
@iker-barriocanal sorry, can you keep this open until your internal tracker resolves the issue? 

I am unable to subscribe to notifications on your internal tracker, and it appears this is closed as resolved.

IanVS:
Does that mean this hasn't been fixed yet?  I was just about to test it out.

AbhiPrasad:
Let's re-open - I'll take care of updating with the OPs tickets.

[...]

PR #2572, opened 2023-10-05
feat: Add support for X-Sentry-forwarded-for header

AbhiPrasad:
ref https://getsentry.atlassian.net/browse/OPS-3823
ref https://github.com/getsentry/relay/issues/2450

For users who use a [SDK tunnel](https://docs.sentry.io/platforms/javascript/troubleshooting/#using-the-tunnel-option) or proxy to route SDK events to Sentry, they often use the x-forwarded-for or [x-vercel-forwarded-for](https://vercel.com/docs/edge-network/headers#x-vercel-forwarded-for) to pass along the real client IP address to Sentry (so it can be stored with the Sentry event).

Current this doesn't work for SaaS Sentry because of restrictions we've placed on the infrastructure side.

~This PR adds support for reading from `X-Sentry-Forwarded-For` header. It is slightly worse that this is a proprietary header, but it unblocks this for our customers without having to suffer the same challenges x-forwarded-for does.~

With https://github.com/getsentry/ops/pull/8175 we're making changes to the infra to map the value in `X-Forwarded-For` to `X-Sentry-Forwarded-For` when the request gets forwarded from nginix to relay. This helps reduce the security concerns from blindly forwarding `X-Forwarded-For`, and preserves existing usage of `X-Forwarded-For`. In addition, self-hosted and local relay usage can still use `X-Forwarded-For` as usual.


jjbayer (approved):
> It is slightly worse that this is a proprietary header

Agree with this statement, but the code change LGTM.

jan-auer:
https://getsentry.atlassian.net/browse/OPS-3823?focusedCommentId=89761

jan-auer (changes requested):
The code change looks good. I would just like [this question](https://getsentry.atlassian.net/browse/OPS-3823?focusedCommentId=89761) addressed before moving ahead.

AbhiPrasad:
Alright I think we're going to move forward as was outlined by Anton in the [JIRA ticket](https://getsentry.atlassian.net/browse/OPS-3823?focusedCommentId=89802), implemented here https://github.com/getsentry/ops/pull/8175.

@jan-auer sounds good to you?

jan-auer (approved):

relay-server/src/extractors/forwarded_for.rs
```diff
@@ -11,12 +11,14 @@ pub struct ForwardedFor(String);
...
     const FORWARDED_HEADER: &str = "X-Forwarded-For";
     const VERCEL_FORWARDED_HEADER: &str = "X-Vercel-Forwarded-For";
+    const SENTRY_FORWARDED_HEADER: &str = "X-Sentry-Forwarded-For";
 
     /// We prefer the Vercel header because the normal one could get overwritten as explained here.
     /// `https://vercel.com/docs/concepts/edge-network/headers#x-vercel-forwarded-for`
     fn get_forwarded_for_ip(header_map: &HeaderMap) -> &str {
         header_map
             .get(Self::VERCEL_FORWARDED_HEADER)
+            .or_else(|| header_map.get(Self::SENTRY_FORWARDED_HEADER))
```
Dav1dde, line 28: The question came up if we should value our own header over the vercel header, the conclusion was:

> 1. our infrastructure requires the "standard" x-forwarded-for to be trusted, so it puts the contents of just that header into the sentry version.
> 2. at the same time, x-vercel-forwarded-for is the more accurate header overall if it is there. so it should be preferred over the other two.

Can you add an explanation  to comment, so somebody in the future knows why the sentry header is not the highest priority (like there is already an explanation why the vercel header is preferred over the standard X-Forwarded-For header)?

AbhiPrasad: Done with 3a97475d1e6d11ad6724c935a39e62fa61489582

PR #2573, opened 2023-10-05
feat(profiling): Restore profiling feature killswitch

phacops:
In order to smooth out the rollout of profiling for ST, we'd like to restore the killswitch, now enabled by default (https://github.com/getsentry/getsentry/blob/master/getsentry/features.py#L1210).


Zylphrex (approved):
Confirmed the feature flag is still being sent from sentry. So we should be good

iker-barriocanal (approved):

jjbayer (approved):

PR #2574, opened 2023-10-06
ref(metrics): Split aggregator service and data

TBS1996:
Splits the `AggregetorService` by extracting the aggregation logic from the service, which will be useful for re-using the aggregator for other purposes. Specifically #2083 for the time being. 

This change should have zero functional changes. 
 
open questions:

- Do we want the cost tracker to be in the new aggregator struct, or have it in the service?
- Do we want to have the aggregator and its service in separate modules? 

#skip-changelog


relay-metrics/src/aggregator.rs
```diff
@@ -693,7 +666,7 @@ impl MergeBuckets {
 
 /// Aggregator service interface.
 #[derive(Debug)]
-pub enum Aggregator {
+pub enum AggregatorManager {
```
TBS1996: rename because now `Aggregator` refers to the data-part of the service that we extracted.

relay-metrics/src/aggregator.rs
```diff
@@ -748,52 +721,107 @@ impl FromMessage<BucketCountInquiry> for Aggregator {
...
-            receiver,
-            state: AggregatorState::Running,
-            cost_tracker: CostTracker::default(),
+    /// Returns the config of the aggregator.
+    pub fn config(&self) -> &AggregatorConfig {
+        &self.config
+    }
+
+    /// Returns the number of buckets in the aggregator.
+    pub fn bucket_qty(&self) -> usize {
```
TBS1996: helper to avoid exposing the internal types in the bucket map

relay-metrics/src/aggregator.rs
```diff
@@ -748,52 +721,107 @@ impl FromMessage<BucketCountInquiry> for Aggregator {
...
+
+    /// Returns `true` if the cost trackers value is larger than the given max cost.
+    pub fn totals_cost_exceeded(&self, max_total_cost: Option<usize>) -> bool {
+        self.cost_tracker.totals_cost_exceeded(max_total_cost)
+    }
+
+    /// Pop and return the buckets that are eligible for flushing out according to bucket interval.
+    ///
+    /// Note that this function is primarily intended for tests.
+    pub fn pop_flush_buckets(&mut self, force: bool) -> HashMap<ProjectKey, Vec<HashedBucket>> {
```
TBS1996, line 493: alternatively, could have a `force_pop_flush_bucket` and `pop_flush_bucket` that each call an inner function with/without the force flag

relay-metrics/src/aggregator.rs
jjbayer: We can also accept the naming clash and qualify the internal aggregator as `aggregator::Aggregator`.

relay-metrics/src/aggregator.rs
```diff
@@ -208,7 +202,7 @@ pub enum ShiftKey {
     Bucket,
 }
 
-/// Parameters used by the [`AggregatorService`].
+/// Parameters used by the [`Aggregator`].
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(default)]
 pub struct AggregatorConfig {
```
jjbayer, line 122: We could split this into `AggregatorConfig` and `AggregatorServiceConfig`, and then `#[serde(flatten)]` them into a single object for the serialized config.

relay-metrics/src/aggregator.rs
jjbayer: nit 
```suggestion
    pub fn bucket_count(&self) -> usize {
```

relay-metrics/src/aggregator.rs
jjbayer, line 493: I would find it more clean if the `Aggregator` had a method `.retain(predicate)` and `pop_flush_buckets()` was a method on the service. But then we would probably also abolish the `QueuedBucket` type and let the service decide when to flush a bucket.

That might be too big a change for this PR. But maybe we can leave a comment about it in the documentation of this method.

relay-metrics/src/aggregator.rs
```diff
@@ -1163,7 +1114,7 @@ impl AggregatorService {
     /// Split the provided buckets into batches and process each batch with the given function.
     ///
     /// For each batch, log a histogram metric.
-    fn process_batches<F>(&self, buckets: impl IntoIterator<Item = Bucket>, mut process: F)
+    pub fn process_batches<F>(&self, buckets: impl IntoIterator<Item = Bucket>, mut process: F)
```
jjbayer: This is a helper function of `try_flush`, so it should be a private method of `AggregatorService`.

relay-metrics/src/lib.rs
```diff
@@ -68,13 +68,15 @@
...
-mod aggregation;
+mod aggregator;
+mod aggregatorservice;
 mod bucket;
 mod protocol;
 mod router;
 mod statsd;
 
-pub use aggregation::*;
+pub use aggregator::*;
```
jjbayer:
```suggestion
```

relay-metrics/src/aggregator.rs
TBS1996: I was thinking of doing something like that, but yeah  I also came to the conclusion that it'd be better for a future PR.

relay-metrics/src/lib.rs
TBS1996: hmm, this would require some refactor of the benchmarking, either moving the benchmarks to the aggregator module, or benchmark the process from you send a message to merge to the service

jjbayer: Can you add a comment that this is only public for benchmarks?

relay-metrics/src/aggregator.rs
```diff
@@ -565,13 +565,14 @@ impl fmt::Debug for CostTracker {
 /// Since this uses an approximate function to estimate the size of buckets, the actual serialized
 /// payload may exceed the size. The estimation function is built in a way to guarantee the same
 /// order of magnitude.
-struct CappedBucketIter<T: Iterator<Item = Bucket>> {
+pub struct CappedBucketIter<T: Iterator<Item = Bucket>> {
```
jjbayer: Please apply recursively: this is an exclusive helper of process_batches AFAICT, so it should also move to the service module.

relay-metrics/src/lib.rs
TBS1996: yeah, did that now. 

Maybe we should do benchmarks of private functions within the file itself like unit tests in the future?

relay-metrics/src/aggregator.rs
TBS1996: id have to make a third config to split and merge from for that right?

jjbayer: Something like this should work:

```rust
struct AggregatorServiceConfig {
    #[serde(flatten)]
    aggregator: AggregatorConfig,
    some_service_related_field: u64,
}

File: relay-metrics/src/lib.rs [hunk 8]
```diff
@@ -68,13 +68,16 @@
...
+pub mod aggregator;
+
+mod aggregatorservice;
 mod bucket;
 mod protocol;
 mod router;
 mod statsd;
 
-pub use aggregation::*;
+pub use aggregatorservice::*;
```
jjbayer, line 80: We can still re-export the parts of the `aggregator` module that are not supposed to be hidden:

```suggestion
pub use aggregator::AggregatorConfig;
pub use aggregatorservice::*;
```

jjbayer (approved):
I would still do the config split (see comment), after that this is good to go!

relay-metrics/src/aggregator.rs
TBS1996: would this serviceconfig actually be used directly or would it pass the service related fields to the service and the aaggregator config to the aggregator?

jjbayer: The `AggregatorService` would get the `AggregatorServiceConfig`, and delegate the `AggregatorConfig` to the `Aggregator`. That way, every component gets only the config that is relevant to that component and nothing else.

PR #2576, opened 2023-10-06
instr(metrics): Another tag for aggregation error

jjbayer:
Might help us understand https://github.com/getsentry/team-ingest/issues/227.

#skip-changelog

### DIFF
--- a/relay-metrics/src/statsd.rs
+++ b/relay-metrics/src/statsd.rs
@@ -184,3 +184,39 @@ impl GaugeMetric for MetricGauges {
         }
     }
 }
+
+/// Returns a low-cardinality metric name for use as a tag key on statsd metrics.
+///
+/// In order to keep this low-cardinality, we only enumerate a handful of well-known, high volume
+/// names. The rest gets mapped to "other".
+pub(crate) fn metric_name_tag(value: &str) -> &'static str {
+    if let Some(value) = [
+        "c:sessions/session@none",
+        "s:sessions/user@none",
+        "s:sessions/error@none",
+        "d:transactions/duration@millisecond",
+        "s:transactions/user@none",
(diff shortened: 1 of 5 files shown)


relay-metrics/src/statsd.rs
```diff
@@ -184,3 +184,39 @@ impl GaugeMetric for MetricGauges {
...
+
+    if value.starts_with("s:spans/") {
+        return "s:spans/";
+    }
+    if value.starts_with("d:spans/") {
+        return "d:spans/";
+    }
+
+    "other"
+}
```
jjbayer, line 222: I brought this back temporarily. We can delete it again once we're done investigating.

relay-server/src/actors/processor.rs
```diff
@@ -2604,17 +2609,9 @@ impl EnvelopeProcessorService {
...
-                        .aggregator_config_for(MetricNamespace::Transactions)
-                        .timestamp_range(),
-                ),
+                transaction_range: Some(agg.timestamp_range()),
                 max_name_and_unit_len: Some(
-                    self.inner
-                        .config
-                        .aggregator_config()
-                        .max_name_length
+                    agg.max_name_length
```
jjbayer: Drive-by fix: this was reading from the wrong aggregator config (without consequences because we rely on defaults anyway).

relay-server/src/actors/processor.rs
```diff
@@ -2594,6 +2594,11 @@ impl EnvelopeProcessorService {
             .project_state
             .has_feature(Feature::SpanMetricsExtraction);
 
+        let agg = self
```
```suggestion
        let aggregator_config = self
```

I would really give it a full name

olksdr (approved):

PR #2578, opened 2023-10-06
feat(spans): Extract more tags for resource spans

phacops:
This PR extracts the following new metrics and tags for `resource` spans:

* `d:spans/http.response_content_length@byte`
* `http.decoded_response_body_length@byte`
* `d:spans/http.response_transfer_size@byte`

These are tagged by environment, description, domain, group, span op, transaction and the new `resource.render_blocking_status` tag. The new tag is also added to the `exclusive_time` and `exclusive_time` light metrics.

Static strings added here: https://github.com/getsentry/sentry/pull/57810

ref: [internal issue](https://www.notion.so/sentry/Ensure-resource-spans-contain-correct-data-1c5c2cbe58bc408b84e763a29843384a)

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -160,15 +160,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 value: None,
                 condition: None,
             })
-            .into_iter()
-            // Tags taken directly from the span payload:
-            .chain(std::iter::once(TagSpec {
-                key: "span.status".into(),
-                field: Some("span.status".into()),
-                value: None,
-                condition: None,
-            }))
-            .collect(),
+            .into(),
         },
     ]);
 
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -110,6 +110,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 ("span.", "system"),
                 ("", "transaction.method"),
                 ("", "transaction.op"),
+                ("", "resource.render_blocking_status"), // only set for resource spans.
             ]
             .map(|(prefix, key)| TagSpec {
                 key: format!("{prefix}{key}"),
@@ -152,7 +153,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 ("span.", "group"),
                 ("span.", "op"),
                 ("", "transaction"),
-                ("", "resource.render_blocking_status"), // only set for resource spans.
+                ("", "resource.render_blocking_status"),
             ]
             .map(|(prefix, key)| TagSpec {
                 key: format!("{prefix}{key}"),
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -36,26 +36,26 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
         .features
         .has(Feature::SpanMetricsExtractionAllModules)
     {
-        None
+        RuleCondition::all()
     } else {
         let is_mongo = RuleCondition::eq("span.system", "mongodb")
             | RuleCondition::glob("span.description", "*\"$*");
 
-        let condition = RuleCondition::eq("span.op", "http.client")
+        RuleCondition::eq("span.op", "http.client")
             | (RuleCondition::glob("span.op", "db*")
                 & !RuleCondition::glob("span.op", DISABLED_DATABASES)
                 & !(RuleCondition::eq("span.op", "db.sql.query") & is_mongo))
-            | RuleCondition::glob("span.op", MOBILE_OPS);
-
-        Some(condition)
+            | RuleCondition::glob("span.op", MOBILE_OPS)
     };
+
+    let resource_condition = RuleCondition::glob("span.op", "resource.*");
 
     config.metrics.extend([
         MetricSpec {
             category: DataCategory::Span,
             mri: "d:spans/exclusive_time@millisecond".into(),
             field: Some("span.exclusive_time".into()),
-            condition: span_op_conditions.clone(),
+            condition: Some(span_op_conditions.clone()),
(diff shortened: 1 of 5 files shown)


DominikB2014:
> Following this list (https://www.notion.so/sentry/Ensure-resource-spans-contain-correct-data-1c5c2cbe58bc408b84e763a29843384a), I added tags needed for `resource` spans.
> 
> I'm not sure what to do for:
> 
> * `spm`
> * `duration`: wouldn't that already be the value we tag or available via the `exclusive_time` column?
> * `url`: this value is already extracted as `domain`, I think we don't need to tag it differently?
> * `type`: isn't it something the frontend can extract from the description?

@phacops We shouldn't need to action on these,
We already have spm, duration, url and type would be under `span.op` or from the description like you mentioned, so no need to worry about that right now.

DominikB2014 (approved):

DominikB2014:
@phacops out of curiosity, once this PR deploys, will he have access to this data right away?

phacops:
> @phacops out of curiosity, once this PR deploys, will he have access to this data right away?

Yes, as long as `resource` spans are enabled for the organization.

> > Following this list (https://www.notion.so/sentry/Ensure-resource-spans-contain-correct-data-1c5c2cbe58bc408b84e763a29843384a), I added tags needed for `resource` spans.
> > I'm not sure what to do for:
> > 
> > * `spm`
> > * `duration`: wouldn't that already be the value we tag or available via the `exclusive_time` column?
> > * `url`: this value is already extracted as `domain`, I think we don't need to tag it differently?
> > * `type`: isn't it something the frontend can extract from the description?
> 
> @phacops We shouldn't need to action on these, We already have spm, duration, url and type would be under `span.op` or from the description like you mentioned, so no need to worry about that right now.

So we're good? The tags I needed to add are added and you already have the rest?

relay-dynamic-config/src/defaults.rs
```diff
@@ -117,6 +117,24 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 })
                 .into(),
         },
+        TagMapping {
+            metrics: vec![LazyGlob::new("d:spans/exclusive_time*@millisecond".into())],
+            tags: [
+                ("", "http.decoded_response_body_length"),
+                ("", "http.response_content_length"),
+                ("", "http.response_transfer_size"),
```
jjbayer: These will have to be their own distribution metrics instead of tags. Beside the cardinality impact, having numeric values as discrete tags does not help UX, since tags are something you group by or filter on (on specific values). I will look into how we can make these separate metrics.

jjbayer (changes requested):

relay-dynamic-config/src/defaults.rs
```diff
@@ -67,7 +67,28 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
             mri: "d:spans/exclusive_time_light@millisecond".into(),
             field: Some("span.exclusive_time".into()),
-            condition: span_op_conditions,
+            condition: Some(span_op_conditions.clone()),
+            tags: Default::default(),
+        },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "d:spans/http.response_content_length@none".into(),
+            field: Some("span.data.http\\.response_content_length".into()),
```
jjbayer, line 76: We have to take this value directly from `data`, because `sentry_tags` is strictly typed to strings.

relay-dynamic-config/src/defaults.rs
```diff
@@ -117,6 +138,38 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
+        // Resource-specific tags:
+        TagMapping {
+            metrics: vec![
+                LazyGlob::new("d:spans/http.response_content_length@none".into()),
+                LazyGlob::new("d:spans/http.decoded_response_body_length@none".into()),
+                LazyGlob::new("d:spans/http.response_transfer_size@none".into()),
+            ],
+            tags: [
+                ("", "environment"),
+                ("span.", "description"),
```
jjbayer, line 151: @DominikB2014 this was not mentioned in the Notion doc, but I assume you need the span description and group hash on these metrics as well?

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -340,6 +382,57 @@ pub(crate) fn extract_tags(span: &Span, config: &Config) -> BTreeMap<SpanTagKey,
...
+        if span_op.starts_with("resource.") {
+            if let Some(http_response_content_length) = span
+                .data
+                .value()
+                .and_then(|data| data.get("http.response_content_length"))
+                .and_then(|value| value.as_str())
+            {
+                span_tags.insert(
+                    SpanTagKey::HttpResponseContentLength,
+                    http_response_content_length.to_owned(),
```
jjbayer, line 395: @phacops this is something we did not have to deal with before: We want to propagate numeric attributes to snuba, but `sentry_tags` can only contain string values. Would you prefer we leave these values in `span.data` instead (and letting the sentry-side consumer transfer them to snuba as numbers somehow)?

relay-dynamic-config/src/defaults.rs
DominikB2014: Yes we will, good call on that!

iker-barriocanal (commented):
The code overall looks good to me! Couple of questions:
- The new metrics are lengths and sizes. Should these have units (i.e. something else than `none`)?
- Are the new metrics and tags added to the string indexer? Could we link to that PR in the description?

jjbayer:
> * The new metrics are lengths and sizes. Should these have units (i.e. something else than `none`)?

@iker-barriocanal good point, I did not think about units at all. Waiting for clarification from @DominikB2014 / @AbhiPrasad.

> * Are the new metrics and tags added to the string indexer? Could we link to that PR in the description?

Yes, will link the updated PR after adding the units.

jjbayer (changes requested):
Need to double-check metric units before proceeding.

jjbayer (approved):

PR #2581, opened 2023-10-09
feat(spool): Avoid evicting old project for incoming envelopes

olksdr:
In case of the incident , when we keep getting envelopes for the specific project, we should not evict it, and keep it around till the incident resolved , project updated, and buffered envelopes processed. 

related to https://github.com/getsentry/team-ingest/issues/222
followup for https://github.com/getsentry/relay/pull/2294

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project.rs
+++ b/relay-server/src/actors/project.rs
@@ -490,7 +490,7 @@ impl Project {
         self.last_updated_at
     }
 
-    /// The last time when this project was used for the incoming envelope.
+    /// The last time that this project was used for an incoming envelope.
     pub fn last_envelope_seen_at(&self) -> Instant {
         self.last_envelope_seen
     }
@@ -820,8 +820,7 @@ impl Project {
         let state = self.valid_state().filter(|state| !state.invalid());
         let mut scoping = envelope.scoping();
 
-        // On every incoming envelopes, which belongs to this project, update when the last it was
-        // seen.
+        // On every incoming envelope, which belongs to this project, update when it was last seen.
         self.last_envelope_seen = envelope.start_time().into();
 
         if let Some(ref state) = state {
--- a/relay-server/src/actors/project.rs
+++ b/relay-server/src/actors/project.rs
@@ -388,6 +388,7 @@ pub struct Project {
     backoff: RetryBackoff,
     next_fetch_attempt: Option<Instant>,
     last_updated_at: Instant,
+    last_envelope_seen: Instant,
     project_key: ProjectKey,
     config: Arc<Config>,
     state: Option<Arc<ProjectState>>,
@@ -404,6 +405,7 @@ impl Project {
             backoff: RetryBackoff::new(config.http_max_retry_interval()),
             next_fetch_attempt: None,
             last_updated_at: Instant::now(),
+            last_envelope_seen: Instant::now(),
             project_key: key,
             config,
             state: None,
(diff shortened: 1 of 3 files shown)


relay-server/src/actors/project.rs
```diff
@@ -813,6 +820,10 @@ impl Project {
         let state = self.valid_state().filter(|state| !state.invalid());
         let mut scoping = envelope.scoping();
 
+        // On every incoming envelopes, which belongs to this project, update when the last it was
+        // seen.
+        self.last_envelope_seen = envelope.start_time().into();
```
iker-barriocanal: I think we should only update the last seen if the envelope is more recent. If we are checking an older envelope, we probably don't want to update it.

relay-server/src/actors/project_cache.rs
```diff
@@ -519,9 +519,16 @@ impl ProjectCacheBroker {
         let eviction_start = Instant::now();
         let delta = 2 * self.config.project_cache_expiry() + self.config.project_grace_period();
 
-        let expired = self
-            .projects
-            .drain_filter(|_, entry| entry.last_updated_at() + delta <= eviction_start);
+        // Get all the projects which have not updated in configured time and also have not gotten
+        // any new incoming envelopes in meantime.
```
```suggestion
        // Get all the projects which have not updated in the configured time and also have not gotten
        // any new incoming envelopes in the meantime.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -519,9 +519,16 @@ impl ProjectCacheBroker {
...
+        //
+        // In case of incident it can happen, that the ProjectState cannot get updated, and we
```
```suggestion
        // In case of an incident it can happen, that the ProjectState cannot get updated, and we
```

relay-server/src/actors/project.rs
```diff
@@ -488,6 +490,11 @@ impl Project {
         self.last_updated_at
     }
 
+    /// The last time when this project was used for the incoming envelope.
```
```suggestion
    /// The last time that this project was used for an incoming envelope.
```

relay-server/src/actors/project.rs
```diff
@@ -813,6 +820,10 @@ impl Project {
         let state = self.valid_state().filter(|state| !state.invalid());
         let mut scoping = envelope.scoping();
 
+        // On every incoming envelopes, which belongs to this project, update when the last it was
+        // seen.
```
```suggestion
        // On every incoming envelope which belongs to this project, update when it was last
        // seen.
```

TBS1996 (approved):
seems good as far as i can tell 🤔

relay-server/src/actors/project_cache.rs
```diff
@@ -1091,4 +1105,60 @@ mod tests {
...
+            }
+
+            tokio::time::sleep(Duration::from_millis(100)).await;
+        }
+
+        // One of the project will be removed.
+        broker.evict_stale_project_caches();
+        // Project 1 did not receive any envelopes, and should be removed now.
+        assert_eq!(broker.projects.len(), 1);
+        assert!(broker.projects.get(&key2).is_some());
```
jjbayer, line 1171: To make the difference to the old implementation more obvious, can we assert here that `project2` has an outdated project state i.e. `last_updated_at < X`. And maybe check that `project2.expiry_state() == ExpiryState::Expired`, to make sure that these envelopes will not be processed but rather enqueued.

relay-server/src/actors/project_cache.rs
```diff
@@ -1091,4 +1105,60 @@ mod tests {
...
+                services.outcome_aggregator.clone(),
+                services.test_store.clone(),
+            );
+            if let hashbrown::hash_map::Entry::Occupied(mut e) = broker.projects.entry(key2) {
+                e.get_mut()
+                    .check_envelope(envelope, services.outcome_aggregator.clone())
+                    .unwrap();
+            }
+
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
jjbayer: Can we call `tokio::pause` at the beginning of this test to make it run faster?

jjbayer (approved):
Have a request for the test case, other than that LGTM!

iker-barriocanal (approved):

PR #2582, opened 2023-10-10
fix(server): Bind exclusively to the port

### LINKED ISSUE #2423: Not properly binding the port since 23.4.0 (therefore unable to deploy to AWS app runner)
### The Problem
Starting with version `23.4.0` I cannot deploy the image to AWS app runner anymore. It seems the new version does not properly / "eagerly" bind the port. I cannot understand the issue, as I am not into Rust. But at least I can prove there is a problem, so thank you for your help:

### Environment

Official docker image from docker hub at versions `23.3.1`, `23.4.0` and `latest`.

### Steps to Reproduce

Run the image with this command:

```sh
docker run --rm -it -p 3000:3000 getsentry/relay:23.4.0 bash -c "apt update && apt install -y python3 && python3 -m http.server --bind 0.0.0.0 3000 & sleep 10 && relay run --mode proxy --host 0.0.0.0 --port 3000"
```

This looks weird at first, but is to prove that the new HTTP server is not properly binding the port. What I am doing here is installing `python3` to run an HTTP server on port 3000 in the background, then sleeping a bit and launching `relay` in foreground on the same port. Until `23.3.1` this is not possible as the service crashes with

```
ERROR relay_log::utils                         > bind to interface failed
caused by: Address already in use (os error 98)
```
But starting with `23.4.0` the port is not properly bound and the service keeps running. 

Of course, in AWS app runner I am not launching any python3 HTTP stuff, but still the port cannot be bound and the health check keeps failing.

### Expected Result

As with `23.3.1` a port failure from

```sh
docker run --rm -it -p 3000:3000 getsentry/relay:23.3.1 bash -c "apt update && apt install -y python3 && python3 -m http.server --bind 0.0.0.0 3000 & sleep 10 && relay run --mode proxy --host 0.0.0.0 --port 3000"
```

image

### Actual Result

Also with `latest`, no complaint about port 3000 already in use:

```sh
docker run --rm -it -p 3000:3000 getsentry/relay:latest bash -c "apt update && apt install -y python3 && python3 -m http.server --bind 0.0.0.0 3000 & sleep 10 && relay run --mode proxy --host 0.0.0.0 --port 3000"
```

image

olksdr:
Bind eagerly (and exclusively) to the port for the HTTP server, and fail to start the Relay otherwise.

The proper error should be also provided, e.g.

```
ERROR relay_server::actors::server: Failed to start the HTTP server: Address already in use (os error 48)
```

fix https://github.com/getsentry/relay/issues/2423

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fix reporting of Relay's crashes to Sentry. The `crash-handler` feature did not enable the crash reporter and uploads of crashes were broken. ([#2532](https://github.com/getsentry/relay/pull/2532))
 - Use correct field to pick SQL parser for span normalization. ([#2536](https://github.com/getsentry/relay/pull/2536))
 - Prevent stack overflow on SQL serialization. ([#2538](https://github.com/getsentry/relay/pull/2538))
+- Bind exclusively to the port for the HTTP server. ([#2582](https://github.com/getsentry/relay/pull/2582))
 
 **Internal**:
 
--- a/relay-server/src/actors/server.rs
+++ b/relay-server/src/actors/server.rs
@@ -1,4 +1,4 @@
-use std::net::SocketAddr;
+use std::net::{SocketAddr, TcpListener};
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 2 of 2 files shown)


iker-barriocanal (approved):
Would be great if we also had a test for this!

tuky:
thx a lot! will try it soon in our setup using AWS AppRunner!

tuky:
Yes, it works again :tada: I could successfully launch a nightly image from 5 days ago to AWS AppRunner.

olksdr:
@tuky, great! Thanks for reporting back!

PR #2583, opened 2023-10-10
instr(metrics): Fix tag on error and add full name

jjbayer:
Fix previous instrumentation of `AggregateMetricsError`, and add the full metric name as an additional (non-indexed) data field for verification.

ref: https://github.com/getsentry/team-ingest/issues/227

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregation.rs
+++ b/relay-metrics/src/aggregation.rs
@@ -19,9 +19,7 @@ use tokio::time::Instant;
 
 use crate::bucket::{Bucket, BucketValue, DistributionValue};
 use crate::protocol::{self, MetricNamespace, MetricResourceIdentifier};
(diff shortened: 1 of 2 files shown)


relay-metrics/src/aggregation.rs
```diff
@@ -1041,12 +1039,17 @@ impl AggregatorService {
         I: IntoIterator<Item = Bucket>,
     {
         for bucket in buckets.into_iter() {
-            let tag = metric_name_tag(&self.name);
```
jjbayer, line 1044: I accidentally passed the aggregator name instead of the bucket name here, so the previous tag was useless (defaulted to "other").

olksdr (approved):

iker-barriocanal (approved):

PR #2585, opened 2023-10-10
fix(replays): Cap replay event size by setting bag-size for unsized fields

cmanallen:
Ref: https://sentry.my.sentry.io/organizations/sentry/issues/396171/?referrer=slack

#skip-changelog

### DIFF
--- a/relay-event-schema/src/protocol/replay.rs
+++ b/relay-event-schema/src/protocol/replay.rs
@@ -34,6 +34,14 @@ use crate::protocol::{
     ClientSdkInfo, Contexts, EventId, LenientString, Request, Tags, Timestamp, User,
 };
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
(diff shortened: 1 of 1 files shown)


relay-event-schema/src/protocol/replay.rs
```diff
@@ -82,6 +82,7 @@ pub struct Replay {
     ///   "replay_type": "session"
     /// }
     /// ```
+    #[metastructure(bag_size = "small")]
     pub replay_type: Annotated<String>,
```
jjbayer, line 86: For string fields, we have an attribute called `max_chars` which might be better suited here, see for example https://github.com/getsentry/relay/blob/5de78d9b5e89a42e27819b601867c9026dc57a93/relay-event-schema/src/protocol/event.rs#L297-L298.

Limits for each category are listed here:

https://github.com/getsentry/relay/blob/5de78d9b5e89a42e27819b601867c9026dc57a93/relay-event-schema/src/processor/attrs.rs#L116-L139

jjbayer (approved):

PR #2586, opened 2023-10-10
feat(spans): Add a feature for the resource spans rollout

phacops:


relay-server/src/actors/processor.rs
```diff
@@ -2266,7 +2266,8 @@ impl EnvelopeProcessorService {
             .and_then(|v| v.get("span.system"))
             .and_then(|system| system.as_str())
             .unwrap_or_default();
-        op == "http.client"
+        resource_span_extraction_enabled && op.contains("resource.")
```
jjbayer: I'm never sure which operator takes precedence, so I would prefer parentheses here.
```suggestion
        (resource_span_extraction_enabled && op.contains("resource."))
```

jjbayer (approved):

PR #2587, opened 2023-10-10
feat: Allow max_chars to accept an integer literal

cmanallen:
### What

Instead of `#[metastructure(max_chars="environment")]` you can specify the exact maximum you want `#[metastructure(max_chars=61)]`. The `TrimmingProcessor` will go through the replay and process the strings, trimming those that are larger.

### Why

Specifying `max_chars` equal to some string representation (i.e. `environment`) is confusing and obscures the intent of the operation.  It's not immediately clear what a `max_chars` of `environment` means without introspecting the `metastructure` macro.

Defining `max_chars` limits to be associated to a particular field's limits leads to brittle code. Presumably the `environment` `max_chars` value could change if the internal representation of the `environment` column changes.  But my environment field (or any field which relies on the `environment` `max_chars` value) may not change and so we would expect things to break in unexpected ways.

Additionally, adding more enumerations which are static and represent common length values (i.e. `limit32`, `limit64`, `limit128`) is unnecessarily constraining and still presents a risk of breakage if un-careful programmers change these values.

By allowing the programmer to specify an integer literal you ensure that that programmer is in complete control of their implementation and is at minimal risk of breakage from other programmers changing their implementations.  New integer literals can be added without the programmer needing to make changes to the `MaxChars` enumeration.  This improves the velocity of development and removes the barrier of introducing new validation limits.

---

I'm unwrapping the integer parsing because the error is expected at the compile step.  What's a good method and location to test this?

#skip-changelog


iker-barriocanal (commented):
Hi! Could you provide some context for the need for this feature? I wonder if we could extend the mapping of items in [maxchars](https://github.com/getsentry/relay/blob/4edcb4a6a5020bbb57b671f2cd3bd98a1bb0aa9a/relay-event-schema/src/processor/attrs.rs#L96) instead.

cmanallen:
@iker-barriocanal @jjbayer Updated the PR description.

iker-barriocanal:
Thanks for elaborating, @cmanallen!

I think `MaxChars` provides consistency without impacting development velocity. This consistency guarantees all variations of an item have the same threshold across the codebase, all of them are updated at the same time (if e.g. internal representation changes), provides clarity on what the actual limits of items are, and aligns them with [sentry constants](https://github.com/getsentry/sentry/blob/master/src/sentry/constants.py). To introduce new limits, the enum can be extended with new items and without additional friction.

Your concern about reusing different limits is valid. If sentry handles an item differently, adding new variants to the enum is perfectly fine. Note that updating `MaxChars` limits is an impactful change, and updates should synchronize with infra and product folks. Not sure if I addressed all your concerns, let me know if I'm missing something!

cmanallen:
Thanks for the thoughtful reply!  It gave me an appreciation for `max_chars`'s goals and priorities.  Apparently I feel very strongly about this because I ended up writing a lot lol.  Or maybe I just needed a lot of words to get my point across.  Hopefully its received well, I think I've said all I wanted to with this!

If you agree -- awesome -- if not then we can close this and move on.

---

Regarding friction.  When adding a character limit you first have to understand what `max_chars = "hash"` means.  It's not immediately obvious if `max_chars` is a poorly named parameter or if `hash` has special meaning.  Originally, I chose to use `bag_size` because the usage of `max_chars` did not make it immediately obvious what it was trying to accomplish. Once you've determined you need to use `max_chars` you then have to grep for `metastructure`'s definition.  Then you read Rust's macro syntax while you grep for the implementation of `max_chars`.  Once you find it, you are redirected to the `MaxChars` enum where you can add your entry, add its hard limit, add its soft limit, update the implementation match, grep around to see if you missed anything, and finally update your `metastructure` macro to use the new word.

That's friction.

In an alternative world I could see:

```rust
#[metastructure(max_chars = HASH)] // defined elsewhere as: pub const HASH: u32 = 128;
```

Which, in my opinion, is more clear.

---

Regarding consistency.  I understand the desire to maintain consistency but its not clear to me that we're enforcing it strongly. 

For example, `environment` is more than its length.  It has a type, its not required, it must be non-empty, and its white-space must be trimmed.  None of that information is encapsulated within `MaxChars` and yet this information defines the nature of the field.  If, for example, we decide to remove white-space trimming from `environment` what will be the method of informing product owners?  grep `pub environment: Annotated<String>`?  If that's the case I feel we could do the same for `max_chars` changes.

We enforce consistency very well with `protocol::User`.  I think these attribute fields are hurt because we have difficulty defining them outside of a struct.  Or at least I assume we do because we haven't defined them in the same way as we have other well-typed fields (`User`, `Request`, etc.).

iker-barriocanal:
Those are indeed good points! Let's go ahead with this PR 😄 

About the test, I'd add a test at the bottom of that file. I think creating a dummy struct with a dummy property and testing if the metastructure works on it should do the job.

cmanallen:
@iker-barriocanal I think I'll need help with the test coverage.  I wanted to experiment with implementing this change for replay but I'm not getting the output I expect.

If I run this on the master branch:

```rust
let input = format!(r#"{{"environment":"{}"}}"#, str::repeat("a", 65));
let event = Annotated::<Replay>::from_json(&input);
assert!(event.is_err());
```

I would expect this to fail but it does not.  Is there a special deserialization/validation method I'm supposed to call?

relay-event-derive/src/lib.rs
iker-barriocanal, line 1: Could we also add a test on this behavior in this crate, so that we don't have to rely on other crates to validate it?

relay-event-schema/src/protocol/replay.rs
```diff
@@ -299,4 +299,35 @@ mod tests {
...
         assert_eq!(output, event.to_json().unwrap());
     }
+
+    #[test]
+    fn test_length_validation() {
+        // String values do not pass validation due to length.
+
+        let input = format!(r#"{{"dist":"{}"}}"#, str::repeat("a", 65));
+        let event = Annotated::<Replay>::from_json(&input);
+        assert!(event.is_err());
```
iker-barriocanal: This validates the replay deserialization, and content validation is done in normalization steps. Light normalization should catch the lengths, and the test would look something like the following:

```rs
let json = r#"..."#;
let event = Annotated::<Replay>::from_json(json).unwrap();
let config = LightNormalizationConfig {
  enable_trimming: true,
  ..Default::default()
};
light_normalize_event(&event, config).unwrap();
// TODO: ensure the field has an error
```
You may want to move these tests to `relay-event-normalization` where normalization functionality is defined.

iker-barriocanal (dismissed):

relay-event-derive/src/lib.rs
cmanallen: I have no idea how to test this.

iker-barriocanal: I've spent some (probably way too much) time trying to do this and didn't manage, so I'd suggest adding a few tests on a specific Replay field like you initially did and moving on. Sorry for the back and forth 😅

iker-barriocanal:
Hi @cmanallen, did you have a chance to get back to this? The PR looks pretty good, I think adding a test like in https://github.com/getsentry/relay/pull/2587#discussion_r1360473698 is good to go ahead with it.

cmanallen:
@iker-barriocanal Hey sorry I've been busy with other things.  I'm in SF right now but I will be able to do this later in the week!

relay-event-normalization/src/replay.rs
```diff
@@ -60,15 +62,22 @@ pub fn validate(replay: &Replay) -> Result<(), ReplayError> {
 
 /// Adds default fields and normalizes all values in to their standard representation.
 pub fn normalize(
-    replay: &mut Replay,
+    replay: &mut Annotated<Replay>,
```
iker-barriocanal, line 65: We may want to follow the same structure for replay normalization as events, but that's out of scope for now.

iker-barriocanal:
Requesting review from @getsentry/owners-ingest as I've worked on it and thus I'm biased.

relay-event-schema/src/protocol/replay.rs
iker-barriocanal: Added a test on https://github.com/getsentry/relay/pull/2587/files#diff-c3b1b3a23c63306f4043665940e399a231d91ffc3795a95a45371ca940893a58R381

relay-event-derive/src/lib.rs
iker-barriocanal: Added a test on https://github.com/getsentry/relay/pull/2587/files#diff-c3b1b3a23c63306f4043665940e399a231d91ffc3795a95a45371ca940893a58R381

iker-barriocanal:
I'll break this PR down into two different steps and align them with the rest of the codebase.

PR #2588, opened 2023-10-10
feat(server): Add aggregator to `Project` to buffer metrics for uninitialized projects

TBS1996:
#2083 

Puts an aggregator in the `Project` struct where we buffer metrics if the project has yet to receive a `ProjectState`. 

Open question:
* If a project expires without having received a projectstate, this pr would drop the metrics. Should we also send those to the aggregator?


relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,40 @@ enum GetOrFetch<'a> {
     Scheduled(&'a mut StateChannel),
 }
 
+/// Enum representing either the project state or an aggregation of metrics.
+///
+/// We want to wait with enforcing rate limits on metrics until we have a project state,
+/// So when we don't have one yet, we hold them in this aggregator until the project state arrives.
+#[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
+enum AggregationState {
```
TBS1996: Feel free to suggest a better name

relay-metrics/src/aggregator.rs
```diff
@@ -487,6 +487,23 @@ impl Aggregator {
         self.cost_tracker.totals_cost_exceeded(max_total_cost)
     }
 
+    /// Takes out the buckets from the aggregator, leaving an empty map.
+    pub fn take_buckets(&mut self) -> Vec<Bucket> {
```
TBS1996: would have preffered if it took self by value, but couldn't please the borrow checker when doing the state change in `Project`

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,40 @@ enum GetOrFetch<'a> {
...
+    State(Arc<ProjectState>),
+    Aggregator(aggregator::Aggregator),
+}
+
+impl AggregationState {
+    fn state_value(&self) -> Option<Arc<ProjectState>> {
```
TBS1996, line 398: naming convention from: https://doc.rust-lang.org/std/ops/enum.ControlFlow.html#method.continue_value

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - Introduce a dedicated usage metric for transactions that replaces the duration metric. ([#2571](https://github.com/getsentry/relay/pull/2571), [#2589](https://github.com/getsentry/relay/pull/2589))
 - Restore the profiling killswitch. ([#2573](https://github.com/getsentry/relay/pull/2573))
 - Add `scraping_attempts` field to the event schema. ([#2575](https://github.com/getsentry/relay/pull/2575))
+- Don't aggregate metrics until we received its' project state. ([#2588](https://github.com/getsentry/relay/pull/2588))
```
```suggestion
- Don't aggregate metrics until we received its project state. ([#2588](https://github.com/getsentry/relay/pull/2588))
```

relay-metrics/src/aggregator.rs
```diff
@@ -487,6 +487,23 @@ impl Aggregator {
...
+
+        buckets
+            .into_iter()
+            .map(|(key, entry)| Bucket {
+                timestamp: key.timestamp,
+                width: bucket_interval,
+                name: key.metric_name,
+                value: entry.value,
+                tags: key.tags,
+            })
```
```suggestion
        let buckets = self.buckets.drain()
            .map(|(key, entry)| Bucket {
                timestamp: key.timestamp,
                width: bucket_interval,
                name: key.metric_name,
                value: entry.value,
                tags: key.tags,
            })
```

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,40 @@ enum GetOrFetch<'a> {
     Scheduled(&'a mut StateChannel),
 }
 
+/// Enum representing either the project state or an aggregation of metrics.
+///
+/// We want to wait with enforcing rate limits on metrics until we have a project state,
```
```suggestion
/// We have to delay rate limiting on metrics until we have a valid project state,
```

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,40 @@ enum GetOrFetch<'a> {
...
+/// So when we don't have one yet, we hold them in this aggregator until the project state arrives.
```
jjbayer, line 388:
```suggestion
/// So when we don't have one yet, we hold them in this aggregator until the project state arrives.
///
/// TODO: spool queued metrics to disk when the in-memory aggregator becomes too full.
```

relay-server/src/actors/project.rs
jjbayer: Maybe just `State`?

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,40 @@ enum GetOrFetch<'a> {
...
+#[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
+enum AggregationState {
+    State(Arc<ProjectState>),
+    Aggregator(aggregator::Aggregator),
```
jjbayer: IMO it's better to describe the state itself instead of the data that are contained in each variant.

```suggestion
    Cached(Arc<ProjectState>),
    Pending(aggregator::Aggregator),
```

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,40 @@ enum GetOrFetch<'a> {
     Scheduled(&'a mut StateChannel),
 }
 
+/// Enum representing either the project state or an aggregation of metrics.
+///
+/// We want to wait with enforcing rate limits on metrics until we have a project state,
+/// So when we don't have one yet, we hold them in this aggregator until the project state arrives.
+#[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
```
jjbayer: Should we just `Box` the aggregator instead of allowing the warning? That should save some memory for the many project states that are already loaded.

relay-server/src/actors/project.rs
```diff
@@ -407,12 +444,12 @@ impl Project {
             last_updated_at: Instant::now(),
             last_envelope_seen: Instant::now(),
             project_key: key,
-            config,
-            state: None,
+            state: AggregationState::new(config.default_aggregator_config().aggregator.clone()),
```
jjbayer: This is where stuff gets tricky: We actually need to apply secondary aggregator configs to metrics that will later be routed into the secondary aggregator services.

So analogous to the `Aggregator` vs. `AggregatorService` split, we need a separate PR that does a `MetricsRouter` vs. `MetricsRouterService` split.

relay-server/src/actors/project.rs
```diff
@@ -573,8 +614,16 @@ impl Project {
     ) {
         if self.metrics_allowed() {
             let buckets = self.rate_limit_metrics(buckets, outcome_aggregator);
+
             if !buckets.is_empty() {
-                aggregator.send(MergeBuckets::new(self.project_key, buckets));
+                match &mut self.state {
+                    AggregationState::State(_) => {
+                        aggregator.send(MergeBuckets::new(self.project_key, buckets));
```
jjbayer, line 626:
```suggestion
                        // We can send metrics straight to the aggregator.
                        aggregator.send(MergeBuckets::new(self.project_key, buckets));
```

relay-server/src/actors/project.rs
```diff
@@ -573,8 +614,16 @@ impl Project {
...
+                    }
+                    AggregationState::Aggregator(inner_agg) => {
+                        inner_agg.merge_all(self.project_key, buckets, None);
```
jjbayer, line 631:
```suggestion
                        // We need to queue the metrics in a temporary aggregator until the project state becomes available.
                        inner_agg.merge_all(self.project_key, buckets, None);
```

relay-server/src/actors/project.rs
```diff
@@ -573,8 +614,16 @@ impl Project {
     ) {
         if self.metrics_allowed() {
             let buckets = self.rate_limit_metrics(buckets, outcome_aggregator);
+
             if !buckets.is_empty() {
-                aggregator.send(MergeBuckets::new(self.project_key, buckets));
+                match &mut self.state {
```
jjbayer, line 622: Note to self: We need to verify that `Project::state` is always up-to-date, and there is no hidden state in `ProjectCache` that invalidates project states without updating the `Project`.

jjbayer (changes requested):
I like the looks of this very much! Pressing "Request changes" because we need to figure out how to deal with `MetricsRouter` (see comments).

relay-config/src/config.rs
```diff
@@ -2003,6 +2004,75 @@ impl Config {
...
+
+        AggregatorConfig {
+            bucket_interval,
+            max_secs_in_past,
+            max_secs_in_future,
+            max_name_length,
+            max_tag_key_length,
+            max_tag_value_length,
+            max_project_key_bucket_bytes,
+            ..Default::default()
```
jjbayer: Let's list every field explicitly here. When new fields are added to the config, we have to make a deliberate decision about what value to pick for the "permissive" config.

relay-config/src/config.rs
```diff
@@ -2003,6 +2004,75 @@ impl Config {
         self.values.processing.max_rate_limit.map(u32::into)
     }
 
+    /// Returns the most permissible [`AggregatorConfig`] based on the values of all the configured aggregators.
+    pub fn permissive_aggregator_config(&self) -> AggregatorConfig {
```
```suggestion
    /// Creates an [`AggregatorConfig`] that is compatible with every other aggregator.
    ///
    /// A lossless aggregator can be put in front of any of the configured aggregators without losing data that the configured aggregator would keep. This is useful for pre-aggregating metrics together in a single aggregator instance.
    pub fn lossless_aggregator_config(&self) -> AggregatorConfig {
```

relay-config/src/config.rs
```diff
@@ -2003,6 +2004,75 @@ impl Config {
...
+            .bucket_interval as usize;
+
+        let max_tag_value_length = aggregators
+            .iter()
+            .max_by_key(|agg| agg.max_tag_value_length)
+            .unwrap()
+            .bucket_interval as usize;
+
+        // None means 'no limit'.
+        let max_project_key_bucket_bytes = None;
```
jjbayer: I think there's a simpler way to code this:

1. Make a private `aggregators()` helper method. Can return an `impl Iterator`, does not have to be a vector.
2. Write a for loop that collects the min/max values
3. Write a second for loop that checks that all bucket intervals align with the min bucket interval

relay-metrics/src/aggregator.rs
```diff
@@ -488,11 +488,10 @@ impl Aggregator {
     }
 
     /// Takes out the buckets from the aggregator, leaving an empty map.
-    pub fn take_buckets(&mut self) -> Vec<Bucket> {
+    pub fn into_buckets(self) -> Vec<Bucket> {
```
jjbayer, line 491: nice!

relay-config/src/config.rs
TBS1996: is the concern performance or readability ? I think your suggestion would be faster but im not sure if the code would be any simpler, or maybe im not visualizing it correctly

relay-server/src/actors/project.rs
TBS1996: note to others: we solved this by creating the permissible aggregatorconfig, to ensure we don't lose any data in the pre-aggregator

relay-config/src/config.rs
```diff
@@ -2003,6 +2004,80 @@ impl Config {
...
+        let aggregators: Vec<AggregatorConfig> = self
+            .values
+            .secondary_aggregators
+            .iter()
+            .map(|agg| &agg.config.aggregator)
+            .chain(std::iter::once(&self.values.aggregator.aggregator))
+            .cloned()
+            .collect();
+
+        // The unwraps are safe, as it's only 'None' if the vector is empty.
```
olksdr: Please, do not use unwraps, even though they are safe now.
It's better to be defensive in the beginning and make sure the future code, if it's about to change, won't fail because of these unwraps. 

I would suggest to fallback to defaults for now.

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,39 @@ enum GetOrFetch<'a> {
     Scheduled(&'a mut StateChannel),
 }
 
+/// Enum representing either the project state or an aggregation of metrics.
```
olksdr:
```suggestion
/// Represents either the project state or an aggregation of metrics.
```

relay-config/src/config.rs
```diff
@@ -2003,6 +2004,80 @@ impl Config {
...
+            .iter()
+            .max_by_key(|agg| agg.max_tag_key_length)
+            .unwrap()
+            .max_tag_key_length;
+
+        let max_tag_value_length = aggregators
+            .iter()
+            .max_by_key(|agg| agg.max_tag_value_length)
+            .unwrap()
+            .max_tag_value_length;
```
jjbayer: I find this slightly easier to digest. It's also shorter + gets rid of the `unwrap()`s.

```suggestion
let AggregatorConfig {
            mut bucket_interval,
            mut max_secs_in_past,
            mut max_secs_in_future,
            mut max_name_length,
            mut max_tag_key_length,
            mut max_tag_value_length,
            ..
        } = &self.default_aggregator_config().aggregator;

        for secondary_config in self.secondary_aggregator_configs() {
            let agg = &secondary_config.config.aggregator;
            bucket_interval = bucket_interval.max(agg.bucket_interval);
            max_secs_in_past = max_secs_in_past.max(agg.max_secs_in_past);
            max_secs_in_future = max_secs_in_future.max(agg.max_secs_in_future);
            max_name_length = max_name_length.max(agg.max_name_length);
            max_tag_key_length = max_tag_key_length.max(agg.max_tag_key_length);
            max_tag_value_length = max_tag_value_length.max(agg.max_tag_value_length);
        }

        for agg in self
            .secondary_aggregator_configs()
            .iter()
            .map(|sc| &sc.config)
            .chain(std::iter::once(self.default_aggregator_config()))
        {
            if agg.aggregator.bucket_interval % bucket_interval != 0 {
                relay_log::error!("buckets don't align");
            }
        }
```

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,39 @@ enum GetOrFetch<'a> {
...
+///
+/// We have to delay rate limiting on metrics until we have a valid project state,
+/// So when we don't have one yet, we hold them in this aggregator until the project state arrives.
+///
+/// TODO: spool queued metrics to disk when the in-memory aggregator becomes too full.
```
olksdr, line 390: It this TODO for followup PRs?

relay-config/src/config.rs
TBS1996: ok i like that, added it 👍

relay-config/src/config.rs
TBS1996: I followed joris approach now, which is without unwraps

relay-server/src/actors/project.rs
TBS1996: Yes, not for this PR.

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,39 @@ enum GetOrFetch<'a> {
...
+    /// Sets the new `ProjectState`. If the variant was pending, the buckets will be returned.
+    fn set_state(&mut self, state: Arc<ProjectState>) -> Option<Vec<Bucket>> {
+        match std::mem::replace(self, Self::Cached(state)) {
+            State::Pending(agg) => Some(agg.into_buckets()),
+            State::Cached(_) => None,
+        }
+    }
+
+    fn new(config: AggregatorConfig) -> Self {
+        Self::Pending(Box::new(aggregator::Aggregator::new(config)))
```
jjbayer: The aggregator created here should get a name, e.g. `"pre"`. Creating multiple aggregators with name `default` (through `Aggregator::new`) is undesirable because the name is used to tag statsd metrics that we collect, and we don't want to mix those for multiple aggregators.

Let's pass the name as an argument to `Aggregator::new`.

relay-server/src/actors/project.rs
```diff
@@ -706,6 +758,17 @@ impl Project {
         }
     }
 
+    fn set_state(&mut self, state: Arc<ProjectState>, aggregator: Addr<Aggregator>) {
+        let buckets = self.state.set_state(state);
+
```
jjbayer, line 768: We should drop buckets here if the project has been disabled, see

https://github.com/getsentry/relay/blob/a00778f110e818c6ce3a5c8e381a05466f04a26d/relay-server/src/actors/project.rs#L899-L902

jjbayer (changes requested):
I think this could work. I would feel more comfortable if we had an integration test verifying the behavior though.

It's unfortunate that a part of a project's enqueued data is now handled by the `Buffer` service (through `ProjectCacheBroker::enqueue` / `dequeue`) and another part (i.e. metrics is handled by the `Project`), but that's not a blocker IMO.

@olksdr could I ask you for a review of this PR to check whether there is something in the `ProjectCache` queueing / dequeueing logic that could interfere with the `update_state` logic in this PR?

relay-server/src/actors/project.rs
TBS1996: oh yeah that makes sense!

TBS1996:
@jjbayer  added some unit tests for the pre-aggregator, lmk if it's good enough or you want python integration tests instead.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 - Introduce a dedicated usage metric for transactions that replaces the duration metric. ([#2571](https://github.com/getsentry/relay/pull/2571), [#2589](https://github.com/getsentry/relay/pull/2589))
 - Restore the profiling killswitch. ([#2573](https://github.com/getsentry/relay/pull/2573))
 - Add `scraping_attempts` field to the event schema. ([#2575](https://github.com/getsentry/relay/pull/2575))
+- Don't aggregate metrics until we received its project state. ([#2588](https://github.com/getsentry/relay/pull/2588))
```
```suggestion
- Postpone metrics aggregation until we received the project state. ([#2588](https://github.com/getsentry/relay/pull/2588))
```

relay-server/src/actors/project.rs
```diff
@@ -379,6 +382,42 @@ enum GetOrFetch<'a> {
...
+impl State {
+    fn state_value(&self) -> Option<Arc<ProjectState>> {
+        match self {
+            State::Cached(state) => Some(Arc::clone(state)),
+            State::Pending(_) => None,
+        }
+    }
+
+    /// Sets the new `ProjectState`. If the variant was pending, the buckets will be returned.
```
olksdr: This comment a bit misleading here. 
```suggestion
    /// Sets the cached state using provided `ProjectState`. 
    /// If the variant was pending, the buckets will be returned.
```

relay-server/src/actors/project.rs
```diff
@@ -1148,6 +1152,61 @@ mod tests {
...
+    /// until it has received a project state.
+    #[tokio::test]
+    async fn test_pre_aggregator_no_flush_without_state() {
+        // Project without project state.
+        let mut project = Project {
+            state: State::new(Config::default().permissive_aggregator_config()),
+            ..create_project(None)
+        };
+
+        let bucket_state = Arc::new(Mutex::new(false));
```
jjbayer, line 1166: nit: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html

jjbayer (approved):

relay-server/src/actors/project.rs
TBS1996: tried and wasnt sensible, we need to share state to see what happened inside the mock_service, so for that a mutex is good, and then we also need it to be sync so we wrap it in arc, then there's no point in the atomicbool which has a harder API

PR #2590, opened 2023-10-11
instr(metrics): Ignore invalid timestamp errors

jjbayer:
Ignore `InvalidTimestamp` error on metrics aggregation.

The metrics aggregator only accept timestamps from the last 5 days. But incoming metrics buckets can always contain invalid timestamps, even when we validated the timestamps on the payload before, because the timestamp may have become invalid while the metric was aggregated in a downstream relay.

With custom metrics, there is no pre-validation at all, so we will see even more `InvalidTimestamp` errors as their usage increases.

ref: https://github.com/getsentry/team-ingest/issues/227

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregatorservice.rs
+++ b/relay-metrics/src/aggregatorservice.rs
@@ -1,4 +1,3 @@
-use std::error::Error;
 use std::iter::FusedIterator;
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


relay-metrics/src/aggregatorservice.rs
```diff
@@ -265,16 +264,8 @@ impl AggregatorService {
...
-                .merge_all(project_key, buckets, self.max_total_bucket_bytes)
-        {
-            relay_log::error!(
-                tags.aggregator = &self.aggregator.name(),
-                error = &err as &dyn Error,
-                "failed to merge buckets"
-            );
-        }
+        self.aggregator
+            .merge_all(project_key, buckets, self.max_total_bucket_bytes);
```
jjbayer, line 268: Drive-by refactor: `merge_all` always returned `Ok(())`, so I made it return `()` instead.

olksdr (approved):

PR #2591, opened 2023-10-11
fix(spans): Scrub resource span description even when we don't detect an extension

phacops:
- Fix fallback case for resource span scrubbing when the regex doesn't work
- Expand the patterns we look for to identify a MongoDB query
- Remove scrubbing for CoreData transactions

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -602,6 +602,15 @@ mod tests {
         "data:image/svg+xml"
     );
 
+    span_description_test!(
+        db_category_with_mongodb_query,
+        "find({some_id:1234567890},{limit:100})",
+        "db",
+        ""
+    );
+
+    span_description_test!(db_category_with_not_sql, "{someField:someValue}", "db", "");
+
     #[test]
     fn informed_sql_parser() {
         let json = r#"
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -43,7 +43,7 @@ pub(crate) fn scrub_span_description(span: &mut Span, rules: &Vec<SpanDescriptio
                     || sub.contains("mongodb")
                     || sub.contains("redis")
                     || is_legacy_activerecord(sub, db_system)
-                    || is_sql_mongodb(sub, description, db_system)
+                    || is_sql_mongodb(description, db_system)
                 {
                     None
                 } else {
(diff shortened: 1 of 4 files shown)


nikhars (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -84,15 +90,26 @@ pub(crate) fn scrub_span_description(span: &mut Span, rules: &Vec<SpanDescriptio
...
 
 /// We are unable to parse active record when we do not know which database is being used.
 fn is_legacy_activerecord(sub_op: &str, db_system: Option<&str>) -> bool {
     db_system.is_none() && (sub_op.contains("active_record") || sub_op.contains("activerecord"))
 }
 
+fn scrub_core_data(string: &str) -> Option<String> {
+    match DB_SQL_TRANSACTION_CORE_DATA_REGEX.replace_all(string, "*") {
+        Cow::Owned(scrubbed) => Some(scrubbed),
+        Cow::Borrowed(_) => Some("".into()),
```
jjbayer: Might as well omit the tag

```suggestion
        Cow::Borrowed(_) => None,
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,10 +215,16 @@ fn scrub_resource_identifiers(mut string: &str) -> Option<String> {
...
-            let domain = normalize_domain(&url.host()?.to_string(), url.port())?;
-            Some(format!("{domain}/*.{extension}"))
+            // If we managed to parse the URL, we'll try to get an extension.
+            if let Ok(url) = Url::parse(string) {
+                let domain = normalize_domain(&url.host()?.to_string(), url.port())?;
+                // If there is an extension, we add it to the domain.
+                if let Some(extension) = url.path().rsplit_once('.') {
+                    return Some(format!("{domain}/*.{}", extension.1));
+                }
+                return Some(domain);
```
```
domain.com/foo.gif -> domain.com/*.gif
domain.com/foo -> domain.com/*
```
```suggestion
                return Some(format!("{domain}/*"));
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,10 +215,16 @@ fn scrub_resource_identifiers(mut string: &str) -> Option<String> {
...
+            }
+            Some(string.into())
```
jjbayer, line 227: If we cannot parse a domain, I don't think we want to accept the unfiltered string here, right?
```suggestion
            None
```

jjbayer (changes requested):

relay-event-normalization/src/normalize/span/description/mod.rs
jjbayer: Since span metrics for resource spans are still protected by a feature flag, I'm actually fine with experimenting here, i.e. leaving `Some(string.into())` and then analyze what data we get.

jjbayer (approved):

PR #2595, opened 2023-10-12
feat(inbound-filters): Implement generic inbound filters

### LINKED ISSUE #57962: Implement new rules support in Relay

iambriccardo:
This PR implements a new kind of inbound filter type, known as `genericFilters`. Such filters enable the definition of inbound filters using the DSL of `RuleCondition`.

For backwards compatibility the new matching logic will first check the `genericFilters` and then the normal static filters that are still hardcoded.

Closes https://github.com/getsentry/sentry/issues/57962


relay-filter/src/config.rs
```diff
@@ -192,6 +196,83 @@ impl LegacyBrowsersFilterConfig {
...
+impl GenericFilterConfig {
+    /// Returns true if the filter is not enabled or the condition is not supported.
+    pub fn is_empty(&self) -> bool {
+        !self.is_enabled || matches!(self.condition, RuleCondition::Unsupported)
+    }
+}
+
+/// Configuration for generic filters.
+#[derive(Clone, Debug, Default)]
+pub struct GenericFiltersConfig(pub Vec<(String, GenericFilterConfig)>);
```
jjbayer: Why `Vec` and not `BTreeMap`? Do you want to guarantee the order of filters?

relay-filter/src/common.rs
```diff
@@ -62,13 +65,17 @@ impl FilterStatKey {
             FilterStatKey::WebCrawlers => "web-crawlers",
             FilterStatKey::InvalidCsp => "invalid-csp",
             FilterStatKey::FilteredTransactions => "filtered-transaction",
+            FilterStatKey::GenericFilters(filter_name) => {
+                return format!("generic-filters@{filter_name}")
```
jjbayer: I thought about this again and I think we can get away with omitting the `generic-filters@` prefix, i.e. implement `From<&str>` instead of `TryFrom<&str>` for `FilterStatKey`, and map any string to `GenericFilter`.

relay-filter/src/common.rs
```diff
@@ -35,6 +35,9 @@ pub enum FilterStatKey {
 
     /// Filtered due to the fact that it was a call to a filtered transaction
     FilteredTransactions,
+
+    /// Filtered due to a generic filter.
+    GenericFilters(String),
```
jjbayer:
```suggestion
    GenericFilter(String),
```

relay-filter/src/common.rs
```diff
@@ -51,7 +54,7 @@ pub enum FilterStatKey {
 
 impl FilterStatKey {
     /// Returns the string identifier of the filter stat key.
-    pub fn name(self) -> &'static str {
+    pub fn name(self) -> String {
```
jjbayer: Save one allocation for non-generic filters:

```suggestion
    pub fn name(self) -> Cow<'static, str> {
```

relay-sampling/src/condition.rs
```diff
@@ -441,6 +441,7 @@ pub enum RuleCondition {
 
     /// An unsupported condition for future compatibility.
     #[serde(other)]
+    #[default]
```
jjbayer: Does `GenericFilterConfig` need to be default (I don't think so). For future compatibility with global config, it might make sense to make the `condition` field for `GenericFilterConfig` optional, though.

relay-filter/Cargo.toml
```diff
@@ -17,6 +17,7 @@ relay-common = { path = "../relay-common" }
 relay-event-schema = { path = "../relay-event-schema" }
 relay-protocol = { path = "../relay-protocol" }
 relay-ua = { path = "../relay-ua" }
+relay-sampling = { path = "../relay-sampling" }
```
jjbayer: This is unfortunate. Could you try if anything breaks when we move `condition.rs` to `relay-protocol` instead?

jjbayer (changes requested):
Few comments, but overall this looks good to me!

relay-filter/src/config.rs
iambriccardo: Yep, exactly. It's for matching purposes, since this list is now dynamic and i want deterministic behaviors with respect to the order defined in the JSON.

relay-filter/src/common.rs
iambriccardo: Sounds good!

relay-sampling/src/condition.rs
iambriccardo: Good points! Will fix

relay-filter/Cargo.toml
iambriccardo: Should we make the move in the same PR?

relay-filter/src/config.rs
```diff
@@ -192,6 +196,86 @@ impl LegacyBrowsersFilterConfig {
...
+        !self.is_enabled || matches!(self.condition, Some(RuleCondition::Unsupported))
+    }
+}
+
+/// Configuration for generic filters.
+///
+/// We use a vector in order to guarantee consistent total ordering of filters that is required
+/// by the matching algorithm.
+#[derive(Clone, Debug, Default)]
+pub struct GenericFiltersConfig(pub Vec<(String, GenericFilterConfig)>);
```
iker-barriocanal: In the interest of future extensibility, we should make this field an object and add the array inside it. This will allow adding new properties to `GenericFiltersConfig` easier.

relay-filter/Cargo.toml
jjbayer: Can also be a separate one. Would definitely make the diff more manageable.

relay-filter/src/generic_filters.rs
```diff
@@ -0,0 +1,98 @@
...
+    use relay_protocol::Annotated;
+    use relay_sampling::condition::RuleCondition;
+
+    fn mock_filters() -> Vec<(String, GenericFilterConfig)> {
+        vec![
+            (
+                "hydrationError".to_string(),
+                GenericFilterConfig {
+                    is_enabled: true,
+                    condition: Some(RuleCondition::eq("event.release", "1.0")),
```
iker-barriocanal: Could we use a filter name that better reflects the condition and avoid some confusion? Same for the filter below.

relay-filter/src/common.rs
```diff
@@ -62,13 +66,16 @@ impl FilterStatKey {
             FilterStatKey::WebCrawlers => "web-crawlers",
             FilterStatKey::InvalidCsp => "invalid-csp",
             FilterStatKey::FilteredTransactions => "filtered-transaction",
-        }
+            FilterStatKey::GenericFilter(filter_name) => {
```
iker-barriocanal: nit: instead of looking at it as a "name", we could look at it as an identifier. Today this is a name, but tomorrow could be a UUID.

relay-filter/src/config.rs
```diff
@@ -192,6 +196,86 @@ impl LegacyBrowsersFilterConfig {
...
+    /// Specifies whether this filter is enabled.
+    pub is_enabled: bool,
+    /// The condition for the filter.
+    pub condition: Option<RuleCondition>,
+}
+
+impl GenericFilterConfig {
+    /// Returns true if the filter is not enabled or the condition is not supported.
+    pub fn is_empty(&self) -> bool {
+        !self.is_enabled || matches!(self.condition, Some(RuleCondition::Unsupported))
```
iker-barriocanal: What should the behavior be if there's no condition, always filter?

iker-barriocanal (changes requested):
The approach looks good to me! Please see the comment below on the protocol object/list.

relay-filter/src/config.rs
iambriccardo: Yeah, that is a good idea, I can definitely do it. If the new object is also a struct without named fields, it will be equivalent to the current `JSON` representation, I suppose. Will try it out!

relay-filter/src/config.rs
iambriccardo: Here the semantics need to be extended. I would classify as filter as disabled iff (the condition is OR):
* is_enabled = False
* condition is None
* condition is Unsupported

iambriccardo: But since the `Unsupported` handling is done directly in the `matches` of `RuleCondition` and now we switched to an optional, I will just keep `is_enabled = False or condition is None`.

iambriccardo:
This PR is blocked on https://github.com/getsentry/relay/pull/2608

relay-filter/src/config.rs
jjbayer: @iker-barriocanal I get the idea, but in this case it is overkill IMO. Every generic filter will be configured independently of other generic filters, so the chance that they will ever have shared properties is low. If they do, we can always add additional config one level higher (`{"filters": {"generic": ..., "genericSharedProps": ...}`).

jjbayer:
@iker-barriocanal I thought about your idea of adding a `version` field to the generic filters config, and I think we can avoid it in the following way: If we need to ensure that outdated relays stop applying generic filters, we can move the config to a new top-level field (e.g. `generic_v2`). This is arguably less pretty, but it helps avoid the complexity of adding & checking a version right now.

relay-filter/src/generic_filters.rs
```diff
@@ -0,0 +1,106 @@
...
+//!
+//! Multiple generic filters can be defined and they are going to be checked in FIFO order. The
+//! first one that matches, will result in the event being discarded with a [`FilterStatKey`]
+//! identifying the matching filter.
+
+use crate::{FilterStatKey, GenericFiltersConfig};
+use relay_event_schema::protocol::Event;
+use relay_protocol::RuleCondition;
+
+/// Version of the generic filters schema.
```
iker-barriocanal:
```suggestion
/// Maximum version of the generic filters schema.
```

relay-filter/src/generic_filters.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// Checks events by patterns in their error messages.
+pub fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
+    // TODO: the condition DSL needs to be extended to support more complex semantics, such as
+    //  collections operations.
+    condition.map_or(false, |condition| condition.matches(event))
+}
+
+/// Filters events by patterns in their error messages.
+pub fn should_filter(event: &Event, config: &GenericFiltersConfig) -> Result<(), FilterStatKey> {
```
iker-barriocanal: I'd move the maximum version check here. It'd also be great if we could test it.

relay-filter/src/generic_filters.rs
```diff
@@ -0,0 +1,106 @@
...
+                GenericFilterConfig {
+                    is_enabled: true,
+                    condition: Some(RuleCondition::eq("event.transaction", "/hello")),
+                },
+            ),
+        ]
+    }
+
+    #[test]
+    fn test_should_filter_event() {
```
iker-barriocanal: This test tries to validate multiple cases, could we separate them into different tests? See [relay coding guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#fd14645580f848d9969d124a59cc3f7d).

relay-filter/src/config.rs
```diff
@@ -192,6 +196,102 @@ impl LegacyBrowsersFilterConfig {
...
+impl GenericFilterConfig {
+    /// Returns true if the filter is not enabled or no condition was supplied.
+    pub fn is_empty(&self) -> bool {
+        !self.is_enabled || self.condition.is_none()
+    }
+}
+
+/// Configuration for a set of ordered filters.
+#[derive(Clone, Debug, Default)]
+pub struct OrderedFilters(pub Vec<(String, GenericFilterConfig)>);
```
iker-barriocanal: Do we need these structs outside `relay-filter`? Same with `GenericFilterConfig` above.
```suggestion
pub(crate) struct OrderedFilters(pub Vec<(String, GenericFilterConfig)>);
```

relay-filter/src/config.rs
```diff
@@ -192,6 +196,102 @@ impl LegacyBrowsersFilterConfig {
...
+    {
+        let mut map = serializer.serialize_map(Some(self.0.len()))?;
+        for (filter_name, filter_config) in self.0.iter() {
+            map.serialize_entry(filter_name, filter_config)?
+        }
+        map.end()
+    }
+}
+
+struct OrderedFiltersVisitor();
```
iker-barriocanal: I think you don't need the `()` here:
```suggestion
struct OrderedFiltersVisitor;
```

relay-filter/src/config.rs
```diff
@@ -192,6 +196,102 @@ impl LegacyBrowsersFilterConfig {
...
+
+    fn visit_map<M>(self, mut access: M) -> Result<Self::Value, M::Error>
+    where
+        M: MapAccess<'de>,
+    {
+        let mut filters = OrderedFilters(vec![]);
+
+        // We don't perform any kind of de-duplication in case of duplicate keys, since this might
+        // lead to an opaque behavior from the outside. The resolution of multiple filters with
+        // the same name, will be a responsibility of the matching algorithm.
```
iker-barriocanal: > The resolution of multiple filters with the same name, will be a responsibility of the matching algorithm.

Sending multiple filters with the same name should be sentry's responsibility, and not Relays to know if two filters are the same. How should Relay deduplicate two different filters with the same name? Relay should just apply them.

relay-filter/src/config.rs
```diff
@@ -192,6 +196,102 @@ impl LegacyBrowsersFilterConfig {
...
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_map(OrderedFiltersVisitor())
+    }
+}
+
+/// Configuration for generic filters.
+///
+/// We use a vector in order to guarantee consistent total ordering of filters that is required
+/// by the matching algorithm.
```
iker-barriocanal: Stale comment? The vector is used in the struct above.

relay-filter/src/config.rs
```diff
@@ -386,4 +520,58 @@ mod tests {
         }
         "###);
     }
+
+    #[test]
+    fn test_deserialize_generic_filters() {
+        let json = r#"{
+            "version": 1,
+            "filters": {
+                "hydrationError": {
```
iker-barriocanal: I think moving the filter name to the filter configuration itself should have an easier to understand protocol, for example:

```json
{
  "version": 1,
  "filters": [
    {
        "id": "hydrationError",
        "enabled": true,
        "condition": { ... }
    }
  ]
}
```
It's also easier to deal with, as you can get a single object instead of a tuple. What do you think?

iker-barriocanal (changes requested):
The approach LGTM, see comment below on the protocol.

relay-filter/src/config.rs
iambriccardo: This comment entails that Relay will naively match first in first out independently of duplicates, exactly because it's not Relay's responsibility.

relay-filter/src/lib.rs
```diff
@@ -20,6 +20,7 @@ pub mod browser_extensions;
 pub mod client_ips;
 pub mod csp;
 pub mod error_messages;
+pub mod generic_filters;
```
jjbayer: nit: I would call this module, `generic` as we are already in the `relay_filter` crate. See https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-%5Brfc-356%5D.

relay-filter/src/lib.rs
```diff
@@ -45,9 +46,16 @@ pub fn should_filter(
     client_ip: Option<IpAddr>,
     config: &FiltersConfig,
 ) -> Result<(), FilterStatKey> {
-    // NB: The order of applying filters should not matter as they are additive. Still, be careful
-    // when making changes to this order.
+    // We support the filtering with the new filters, only for the same or older versions.
+    if config.generic.version <= generic_filters::VERSION {
```
jjbayer: nit: I would put this comparison in a helper method `is_enabled`, and check it inside `should_filter`. That way, `VERSION` can even become a private constant if I'm not mistaken.

See for example

https://github.com/getsentry/relay/blob/7392b5d502bb760a0baa6800f2876737a06aedf5/relay-dynamic-config/src/metrics.rs#L51-L53

relay-filter/src/generic_filters.rs
jjbayer:
```suggestion
/// Maximum supported version of the generic filters schema. If the version in the project config is higher, no generic filters are applied.
```

relay-filter/Cargo.toml
jjbayer:
```suggestion
```

relay-filter/src/generic_filters.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use crate::{FilterStatKey, GenericFiltersConfig};
+use relay_event_schema::protocol::Event;
+use relay_protocol::RuleCondition;
+
+/// Version of the generic filters schema.
+pub const VERSION: u16 = 1;
+
+/// Checks events by patterns in their error messages.
+pub fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
```
jjbayer:
```suggestion
fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
```

jjbayer (commented):
Almost there! See comments.

relay-filter/src/config.rs
iambriccardo: The idea for the protocol, was to initially implement it very similarly to existing filters. For me it's not a problem to use the standard object implementation. @jjbayer what do you think?

jjbayer: What I like about @iker-barriocanal's proposal is that it does not require custom (de)serializers, and that it preserves the order of filters regardless of the JSON implementation. I just read that the [JSON standard](https://www.json.org/) defines an object as an

> unordered set of name/value pairs.

In that light I would say let's implement @iker-barriocanal's proposal.

relay-filter/src/config.rs
```diff
@@ -192,6 +196,102 @@ impl LegacyBrowsersFilterConfig {
     }
 }
 
+/// Configuration for a generic filter.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct GenericFilterConfig {
```
jjbayer, line 199: The fact that we need so many `pub(crate)` declarations in this crate tells me that the code structure is suboptimal. Ideally this crate would have a slim public interface with private submodules and a few public re-exports. Out of scope for this PR though, because it applies to the entire crate.

See [internal guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#0505679040be41b78bb13086dec4319f).

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - Add a setting to rollout ingesting all resource spans. ([#2586](https://github.com/getsentry/relay/pull/2586))
 - Drop events starting or ending before January 1, 1970 UTC. ([#2613](https://github.com/getsentry/relay/pull/2613))
 - Add support for X-Sentry-Forwarded-For header. ([#2572](https://github.com/getsentry/relay/pull/2572))
+- Add new generic inbound filters. ([#2595](https://github.com/getsentry/relay/pull/2595))
```
```suggestion
- Add a generic way of configuring inbound filters via project configs. ([#2595](https://github.com/getsentry/relay/pull/2595))
```

jjbayer (approved):
LGTM, I would actually implement @iker-barriocanal's suggestion though (see comments).

relay-filter/src/config.rs
iambriccardo: Yep, this is why it was good, since we would remove the custom ser/deser code.

relay-filter/src/config.rs
iambriccardo: Exactly, I wanted to propose this refactor.

jjbayer (approved):

relay-filter/src/generic.rs
```diff
@@ -0,0 +1,148 @@
...
+//! Multiple generic filters can be defined and they are going to be checked in FIFO order. The
+//! first one that matches, will result in the event being discarded with a [`FilterStatKey`]
+//! identifying the matching filter.
+
+use crate::{FilterStatKey, GenericFiltersConfig};
+use relay_event_schema::protocol::Event;
+use relay_protocol::RuleCondition;
+
+/// Maximum supported version of the generic filters schema. If the version in the project config
+/// is higher, no generic filters are applied.
```
iker-barriocanal:
```suggestion
/// Maximum supported version of the generic filters schema.
///
/// If the version in the project config is higher, no generic filters are applied.
```
From [internal guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#987c0416a9f54674999eed59eecbf846).

iker-barriocanal (approved):

PR #2596, opened 2023-10-12
chore(dev): Check in .python-version

jjbayer:
Multiple developers have encountered issues when running `make` commands in their local environment.

Our makefile sets `RELAY_PYTHON_VERSION` to `python3`, but that might be `python3.11`, which we do not support yet.

The sentry repo has this file under version control as well: https://github.com/getsentry/sentry/blob/942c87a918bca11d5ac1c028a08bcc375d43e550/.python-version

#skip-changelog


olksdr (commented):
What the problem with python `3.11`? 

Currently I have `Python 3.11.4` and there is no problem with relay or tests (at least I haven't noticed any). 

Does it really that critical for Relay to pin the python version?

jjbayer:
> What the problem with python `3.11`?

When trying to install the relay python package, `make setup-venv` fails with

```
      File "/Users/joris/code/relay/.venv/lib/python3.11/site-packages/milksnake/ffi.py", line 8, in <module>
        _directive_re = re.compile(r'^\s*#.*?$(?m)')
 ```

This was actually fixed in https://github.com/getsentry/milksnake/pull/32. Will try to get that package released instead of bumping the version here.

PR #2597, opened 2023-10-12
ref(metrics): Remove unused result.

TBS1996:
`merge_all` never returns anything else than Ok, so it's annoying for the caller to have to deal with an error case that can never happen. 

#skip-changelog


iker-barriocanal (commented):
Seems the same as https://github.com/getsentry/relay/commit/3e0d1f8d4edb82ee453eee791fb127fc32564964? Not sure why I can see the diff again?

TBS1996:
oh quite a coincident lol. I hadn't pulled master in the last 2 hours and I wasnt aware that he fixed it too.

PR #2599, opened 2023-10-12
build(py): Bump milksnake dependency

jjbayer:
Bump milksnake version to be compatible with Python 3.11.

See https://github.com/getsentry/milksnake/pull/32.

#skip-changelog


olksdr (approved):
🚢  it!

PR #2600, opened 2023-10-12
feat(metrics): Add device.class to transaction metrics

jjbayer:
Transfer the `device.class` tag that we already set on transaction payloads to transaction metrics.

`device.class` is a low-cardinality tag with only three possible values ("1", "2", "3"), and it is only set for mobile SDKs.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -182,6 +182,10 @@ fn extract_universal_tags(event: &Event, config: &TransactionMetricsConfig) -> C
         tags.insert(CommonTag::HttpStatusCode, status_code);
     }
 
+    if let Some(device_class) = event.tag_value("device.class") {
+        tags.insert(CommonTag::DeviceClass, device_class.to_owned());
+    }
+
     let custom_tags = &config.extract_custom_tags;
     if !custom_tags.is_empty() {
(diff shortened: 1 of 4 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -457,7 +463,8 @@ mod tests {
             },
             "tags": {
                 "fOO": "bar",
-                "bogus": "absolutely"
+                "bogus": "absolutely",
+                "device.class": "1"
```
jjbayer, line 467: Even though the device.class is tagged here, it is not extracted as a metric.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -1317,6 +1324,59 @@ mod tests {
...
+            "measurements": {
+                "frames_frozen": {
+                    "value": 3
+                }
+            },
+            "tags": {
+                "device.class": "2"
+            },
+            "sdk": {
+                "name": "sentry.cocoa"
```
jjbayer, line 1349: With the appropriate SDK name, we add the new tag to metrics.

shruthilayaj (approved):
thanks!

iker-barriocanal (approved):

olksdr (approved):

PR #2603, opened 2023-10-12
fix(spans): Don't split the description to get the domain for resource spans

phacops:
#skip-changelog


olksdr (approved):
lgtm

Maybe a small tests with different `description` will also be good. If we do not have any now.

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_all_modules.snap
```diff
@@ -1262,6 +1264,7 @@ expression: metrics
             "resource.render_blocking_status": "blocking",
             "span.category": "resource",
             "span.description": "https://cdn.domain.com/path/to/file-hk2YHeW7Eo2XLCi*z*KuljsgCAD6hyWCyOYZM.css",
+            "span.domain": "cdn.domain.com",
```
jjbayer, line 1267: Ideally this should be `*.domain.com` to be consistent with HTTP spans. I will clean this up in a follow-up PR, I think resource and HTTP spans can share the same code path.

jjbayer:
> Maybe a small tests with different `description` will also be good.

The [updated snapshots](https://github.com/getsentry/relay/pull/2603/commits/6fa1c9f229327d85726a69a9c150c602b7a7ee82) provide some test coverage.

iker-barriocanal (approved):

PR #2604, opened 2023-10-13
feat(feedback): add User Feedbacks as Events

JoshFerge:
Relay changes for the revamped User Feedback feature. User Feedbacks (Known as `UserFeedbackV2`), will be events that flow through our system instead of attachments. See [here](https://www.notion.so/sentry/User-Feedback-Beta-11d68db039bc436bae017ad56c0c20f9?pvs=4) for more information on the tech spec (and the link at the bottom for what we do once they're ingested).

Note that within this code, we're calling it `UserReportV2`. This is to avoid confusion with the existing `UserReport`. A PR to shim the previous `UserReports` as new `UserReportV2`'s will follow, at which point we'll rename this in the code to be `UserFeedback`. The string `feedback` is used for the context and ItemType / EventType parsing.

- [x] Adds a feature flag for ingest `UserReportV2Ingest`. This is backed by options https://github.com/getsentry/getsentry/issues/11573
- [x] Creates a new EventType `UserReportV2`
- [x] Creates a new ItemType `UserReportV2`
- [x] Creates a new DataCategory `UserReportV2`
- [x] Adds a new context `UserReportV2Context` that for now contains two properties, `contact_email` (no pii stripping), and the `message` which contains the feedback text.


relay-base-schema/src/data_category.rs
```diff
@@ -108,6 +114,7 @@ impl DataCategory {
             Self::Monitor => "monitor",
             Self::Span => "span",
             Self::MonitorSeat => "monitor_seat",
+            Self::UserReportV2 => "user_report_v2",
```
iker-barriocanal: Please also update the `from_name` function above.

relay-base-schema/src/events.rs
```diff
@@ -41,6 +41,9 @@ pub enum EventType {
     ExpectStaple,
     /// Performance monitoring transactions carrying spans.
     Transaction,
+    /// User feedback payload. TODO(Jferg): Change this to UserFeedback once old UserReport
+    /// logic is deprecated.
+    UserReportV2,
```
iker-barriocanal: As per Relay's coding guidelines, documentation should have one line followed by additional paragraphs:
```suggestion
    /// User feedback payload.
    ///
    /// TODO(Jferg): Change this to UserFeedback once old UserReport logic is deprecated.
    UserReportV2,
```

relay-base-schema/src/data_category.rs
```diff
@@ -57,6 +57,12 @@ pub enum DataCategory {
     /// but we define it here to prevent clashing values since this data category enumeration
     /// is also used outside of Relay via the Python package.
     MonitorSeat = 13,
+    /// User Feedback
+    ///
+    /// Represents a User Feedback processed.
+    /// Currently standardized on name UserReportV2 to avoid clashing with the old UserReport.
+    /// TODO(jferg): Rename this to UserFeedback once old UserReport is deprecated.
+    UserReportV2 = 14,
```
iker-barriocanal, line 65: Depending on when UserReport -> UserFeedback transition happens, we may need to support both alternatives to avoid future breaking changes. Have we considered this? Is there any ETA on the deprecation?

relay-dynamic-config/src/feature.rs
```diff
@@ -11,6 +11,10 @@ pub enum Feature {
     /// Enables data scrubbing of replay recording payloads.
     #[serde(rename = "organizations:session-replay-recording-scrubbing")]
     SessionReplayRecordingScrubbing,
+    /// Enables new User Feedback ingest. TODO(jferg): rename to UserFeedbackIngest
+    /// once old UserReport logic is deprecated.
```
iker-barriocanal:
```suggestion
    /// Enables new User Feedback ingest.
    ///
    /// TODO(jferg): rename to UserFeedbackIngest once old UserReport logic is deprecated.
```

relay-event-schema/src/protocol/contexts/mod.rs
```diff
@@ -27,6 +27,7 @@ pub use reprocessing::*;
 pub use response::*;
 pub use runtime::*;
 pub use trace::*;
+pub use user_report_v2::*;
```
iker-barriocanal, line 30: `*` exports above aren't ideal, could we only export the types we need?

relay-server/src/actors/processor.rs
```diff
@@ -1712,6 +1716,14 @@ impl EnvelopeProcessorService {
...
                 self.event_from_json_payload(item, Some(EventType::Transaction))?
             })
+        } else if let Some(item) = user_report_v2_item {
+            relay_log::trace!("processing user_report_v2");
+            let project_state = &state.project_state;
+            let user_report_v2_ingest = project_state.has_feature(Feature::UserReportV2Ingest);
+            if !user_report_v2_ingest {
+                return Err(ProcessingError::NoEventPayload);
+            }
+            self.event_from_json_payload(item, Some(EventType::UserReportV2))?
```
iker-barriocanal, line 1729: Returning an error here means the envelope will not be processed, even if it also contains metrics. Should we only call `event_from_json_payload` if the flag is enabled?

relay-server/src/envelope.rs
```diff
@@ -546,6 +551,7 @@ impl Item {
             ItemType::Statsd | ItemType::MetricBuckets => None,
             ItemType::FormData => None,
             ItemType::UserReport => None,
+            ItemType::UserReportV2 => None,
```
iker-barriocanal, line 564: This means no outcomes will be generated. Is this intended?

iker-barriocanal (changes requested):
Leaving some questions/points from a preliminary review:
- Since we're adding a new data category, we should update the C ABI headers -- `make header` in `relay-cabi` should work. We should also update the py/Changelog.
- What's the plan for current feedback items that live as attachments?
- Should these new feedback items go into a new Kafka topic?
- See my comment below on UserReport -> UserFeedback transition.

relay-base-schema/src/data_category.rs
jjbayer: Why not use `UserFeedback` right away? I agree with Iker that renaming later could bring extra complexity and possible compatibility issues.

relay-base-schema/src/events.rs
```diff
@@ -71,6 +74,7 @@ impl FromStr for EventType {
             "expectct" => EventType::ExpectCt,
             "expectstaple" => EventType::ExpectStaple,
             "transaction" => EventType::Transaction,
+            "feedback" => EventType::UserReportV2,
```
jjbayer, line 78: Similar question: If the string representation is `"feedback"`, why not call it `UserFeedback` in code?

relay-server/src/actors/processor.rs
jjbayer: @iker-barriocanal we split event-like items into a separate envelope, so that should be fine.

Raising an error here will log an `invalid` outcome, but only if `outcome_category()` returns `Some` (see discussion below).

relay-base-schema/src/data_category.rs
jjbayer, line 65: Separate question: Why do we need a new data category, if user feedback items are converted to events (i.e. `Error` category)?

JoshFerge: they are not going to be Error category events. They will be billed separately.

relay-base-schema/src/events.rs
JoshFerge: when discussing with @jan-auer over initial call, he called out to not have it so there are "synonyms" in the code. I can change to UserFeedback if that is preferred.

relay-server/src/envelope.rs
JoshFerge: for now, yes. will change this once we have data flowing through system and we're closer to alpha etc.

relay-base-schema/src/data_category.rs
JoshFerge: when discussing with @jan-auer over initial call, he called out to not have it so there are "synonyms" in the code. I can change to UserFeedback if that is preferred.

relay-event-schema/src/protocol/contexts/mod.rs
JoshFerge: should i include that refactor in this PR, or could I follow up with that? not sure how you all like to split changes like that into separate PRs or not.

JoshFerge:
> * Since we're adding a new data category, we should update the C ABI headers -- `make header` in `relay-cabi` should work. We should also update the py/Changelog.

done.

> * What's the plan for current feedback items that live as attachments?

see PR description:
> A PR to shim the previous UserReports as new UserReportV2's will follow, at which point we'll rename this in the code to be UserFeedback.

> * Should these new feedback items go into a new Kafka topic?

No -- we'll process them in our ingest consumer conditioned on the event type. this was chosen over a separate Kafka topic in conversation w/ @jan-auer last week.

> * See my comment below on UserReport -> UserFeedback transition.

Replied, when Jan and I spoke, there was a decision to avoid synonyms in the code. can change if ya'll would like -- I don't have a strong opinion here. also see PR description:

> at which point we'll rename this in the code to be UserFeedback

relay-base-schema/src/data_category.rs
jjbayer: I think more important than the internal naming is making sure that the serialized string representations in the public interface do not have to be changed, e.g `"user_report_v2"` for the data category.

The internal names of the types are less important, but I did a case insensitive search for "feedback" and found only a few references in comments, so I would personally go for the new naming straight away, but I'll leave it up to you. cc @jan-auer

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - Accept spans needed for the mobile Starfish module. ([#2570](https://github.com/getsentry/relay/pull/2570))
 - Extract size metrics and blocking status tag for resource spans. ([#2578](https://github.com/getsentry/relay/pull/2578))
 - Add a setting to rollout ingesting all resource spans. ([#2586](https://github.com/getsentry/relay/pull/2586))
+- Add User Feedback Ingestion. ([#2604](https://github.com/getsentry/relay/pull/2604))
```
iker-barriocanal: Please also update py/changelog

iker-barriocanal (commented):
Pending the `relay-cabi` update, LGTM. Thanks a lot for working on this PR!

CHANGELOG.md
jjbayer: Also, please move this to the `## Unreleased` section at the top of this file.

py/CHANGELOG.md
```diff
@@ -2,44 +2,64 @@
 
 ## Unreleased
 
-- Add `scraping_attempts` field to the event schema. ([#2575](https://github.com/getsentry/relay/pull/2575))
-- Drop events starting or ending before January 1, 1970 UTC. ([#2613](https://github.com/getsentry/relay/pull/2613))
-- Remove event spans starting or ending before January 1, 1970 UTC. ([#2627](https://github.com/getsentry/relay/pull/2627))
-- Remove event breadcrumbs dating before January 1, 1970 UTC. ([#2635](https://github.com/getsentry/relay/pull/2635))
-- Add `locale` ,`screen_width_pixels`, `screen_height_pixels`, and `uuid` to the device context. ([#2640](https://github.com/getsentry/relay/pull/2640))
+- Add `scraping_attempts` field to the event schema.
+  ([#2575](https://github.com/getsentry/relay/pull/2575))
```
jjbayer: Did the autoformatter run wild here?

relay-event-schema/src/protocol/contexts/user_report_v2.rs
```diff
@@ -0,0 +1,78 @@
...
+mod tests {
+    use super::*;
+    use crate::protocol::Context;
+
+    #[test]
+    fn test_feedback_context() {
+        let json = r#"{
+  "message": "test message",
+  "contact_email": "test@test.com",
+  "type": "userreportv2"
```
jjbayer: I think in the serialized format we should use the future-proof name, i.e. `"feedback"` or `"userfeedback"`. You can accomplish that by adding a `#[metastructure(tag = "feedback")]` attribute to the variant.

py/CHANGELOG.md
JoshFerge: fixed! not sure why my markdown formatter does this.

relay-event-schema/src/protocol/contexts/user_report_v2.rs
JoshFerge: good call. just made that change.

jjbayer (approved):

JoshFerge:
@iker-barriocanal could you take a look again? need your approval since you requested changes.

relay-base-schema/src/data_category.rs
```diff
@@ -108,6 +114,7 @@ impl DataCategory {
             Self::Monitor => "monitor",
             Self::Span => "span",
             Self::MonitorSeat => "monitor_seat",
+            Self::UserReportV2 => "feedback",
```
iker-barriocanal, line 118: Please add this string to the `from_name` above before merging, otherwise Relay won't recognize the data category when deserializing.

iker-barriocanal (approved):

PR #2607, opened 2023-10-13
feat(event-schema): Add getters for most contexts in Discover

jan-auer:
Add field implementations in the `Getter` on `Event` for most contexts declared by Discover. This is needed for ondemand metric extraction in widgets, so that all context fields can be used in widget queries. 

For the full list of Discover fields, see:

https://github.com/getsentry/sentry/blob/master/src/sentry/snuba/events.py

After this has merged, the metrics extraction allow list in Sentry can be updated. It is located here:

https://github.com/getsentry/sentry/blob/8c61340c93045bc1c9e4d1fa1a77b062c8dfb44e/src/sentry/snuba/metrics/extraction.py#L51

### DIFF
--- a/relay-event-schema/src/protocol/event.rs
+++ b/relay-event-schema/src/protocol/event.rs
@@ -12,7 +12,7 @@ use uuid::Uuid;
 
 use crate::processor::ProcessValue;
 use crate::protocol::{
-    Breadcrumb, Breakdowns, BrowserContext, ClientSdkInfo, Contexts, Csp, DebugMeta,
+    AppContext, Breadcrumb, Breakdowns, BrowserContext, ClientSdkInfo, Contexts, Csp, DebugMeta,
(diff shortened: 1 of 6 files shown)


olksdr (approved):

iker-barriocanal (approved):

relay-event-schema/src/protocol/event.rs
```diff
@@ -652,10 +652,58 @@ impl Getter for Event {
...
+                .other
+                .get("screen_height_pixels")?
+                .value()?
+                .into(),
+            "contexts.device.screen_width_pixels" => self
+                .context::<DeviceContext>()?
+                .other
+                .get("screen_width_pixels")?
+                .value()?
+                .into(),
```
jjbayer: Should we add these fields to the schema?

jjbayer (approved):

iambriccardo (approved):

PR #2608, opened 2023-10-16
ref(rules): Move rules DSL to relay-protocol

iambriccardo:
This PR moves the `condition` module from `relay-sampling` to `relay-protocol`. In order to perform the refactor, an adaptation of tests had to be done, in which dependencies from `relay-sampling` where removed by adding mock structs that implement `Getter`.


relay-cabi/src/processing.rs
```diff
@@ -20,8 +20,8 @@ use relay_event_schema::protocol::{Event, VALID_PLATFORMS};
 use relay_pii::{
     selector_suggestions_from_value, DataScrubbingConfig, PiiConfig, PiiConfigError, PiiProcessor,
 };
+use relay_protocol::RuleCondition;
 use relay_protocol::{Annotated, Remark};
```
jjbayer: Let's merge these imports.

jjbayer (approved):
LGTM! @jan-auer there might be a better place to put the rule-matching logic, but I think putting it in `relay-protocol` is an improvement over importing `relay-samping` everywhere.

iker-barriocanal (approved):

relay-protocol/src/lib.rs
```diff
@@ -17,19 +17,23 @@
...
 mod impls;
 mod macros;
 mod meta;
 mod size;
 mod traits;
+mod utils;
 mod value;
 
 pub use self::annotated::*;
+pub use self::condition::*;
```
jan-auer, line 30: This module contains a lot of types that usually do not need to be imported. By glob-exporting them, we're littering the global crate namespace as can be seen in [the docs](https://getsentry.github.io/relay/relay_protocol/index.html) now.

When you compare the imports with how `relay-sampling` was structured, there was an explicit re-export of the `RuleCondition` type, as this is the most commonly used one, but all the other types were left in a `pub mod condition` instead. I'd prefer if we do this here, too.

That is independent of whether other modules are glob-exported or not.

relay-protocol/src/lib.rs
```diff
@@ -17,19 +17,23 @@
...
 pub use self::impls::*;
 pub use self::macros::*;
 pub use self::meta::*;
 pub use self::size::*;
 pub use self::traits::*;
+pub use self::utils::*;
```
jan-auer, line 36: Definitely do not glob-export internal utils to a public namespace. This file contains the function `is_default`, which shouldn't be part of the public API of `relay-protocol`. Instead, we should remove that module and just declare the function locally where needed.

relay-sampling/src/evaluation.rs
```diff
@@ -399,13 +399,13 @@ mod tests {
...
     use uuid::Uuid;
 
-    use crate::condition::RuleCondition;
     use crate::config::{
         DecayingFunction, RuleId, RuleType, SamplingRule, SamplingValue, TimeRange,
     };
     use crate::dsc::TraceUserContext;
     use crate::evaluation::MatchedRuleIds;
     use crate::DynamicSamplingContext;
+    use relay_protocol::RuleCondition;
```
jan-auer, line 408: Please move this import into the import group above.

jan-auer (commented):
Please excuse the late review. See the comments below for a follow-up PR.

Eventually, we can look for an even better place for these types; `relay-protocol` is not the intended crate for this, as it is meant to contain foundational types for building our data schemas, which rule conditions are not part of. That said, I also haven't found the right place yet.

relay-protocol/src/lib.rs
iambriccardo: I thought about this, you are right that it's not good. Definitely needs to be re-worked.

PR #2610, opened 2023-10-16
feat(pii): Allow pii scrubbing expressions for safe fields

jjbayer:
This doesn't simply make matching on arrays work, it enables full expression support for safe fields in the datascrubbing config.

### DIFF
--- a/relay-pii/src/convert.rs
+++ b/relay-pii/src/convert.rs
@@ -164,9 +164,10 @@ pub fn to_pii_config(
 
 #[cfg(test)]
 mod tests {
+    use insta::assert_debug_snapshot;
     use relay_event_schema::processor::{process_value, ProcessingState};
     use relay_event_schema::protocol::Event;
-    use relay_protocol::{assert_annotated_snapshot, FromValue};
+    use relay_protocol::{assert_annotated_snapshot, get_value, FromValue};
     use similar_asserts::assert_eq;
 
     use crate::PiiProcessor;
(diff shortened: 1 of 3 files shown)


relay-pii/src/convert.rs
```diff
@@ -1504,4 +1516,36 @@ THd+9FBxiHLGXNKhG/FRSyREXEt+NyYIf/0cyByc9tNksat794ddUqnLOg0vwSkv
...
+                "extra.'do_not_scrub.dot'.**".to_owned(),
+                "$user.id".to_owned(),
+            ],
+            ..simple_enabled_config()
+        })
+        .unwrap();
+
+        let mut pii_processor = PiiProcessor::new(pii_config.compiled());
+        process_value(&mut data, &mut pii_processor, ProcessingState::root()).unwrap();
+        assert_annotated_snapshot!(data);
```
jjbayer: nit: IMO this snapshot is small enough to inline: 
```suggestion
        assert_annotated_snapshot!(data, @"");
```

relay-pii/src/convert.rs
```diff
@@ -143,9 +143,21 @@ pub fn to_pii_config(
             continue;
         }
 
-        conjunctions.push(SelectorSpec::Not(Box::new(SelectorSpec::Path(vec![
-            SelectorPathItem::Key(field.to_owned()),
-        ]))));
+        let spec = match field.parse() {
+            Ok(spec) => spec,
+            Err(error) => {
+                relay_log::error!(
```
jjbayer, line 151:
```suggestion
                // Invalid safe fields should be caught by sentry-side validation. Log an error if they are not.
                relay_log::error!(
```

jjbayer (commented):
Looks good! As discussed, let's discuss briefly with the team before merging.

iker-barriocanal (approved):

Dav1dde:
Re-opened in #2670

PR #2611, opened 2023-10-16
ref(metrics): Remove aggregator service

TBS1996:
Previously we had a `RouterService` which routed metrics to different aggregator services based on their namespace, those services each had an aggregator. 

This PR removes the middle man, now the RouterService is renamed to AggregatorService, which has direct access to a `Router` object which has takes care of routing to different aggregators. 

The goal of the PR is both performance (function calls vs. message sending),  easier to understand code, and allowing us to re-use the `Router` when we want aggregation with routing outside of the `RouterService`. 

before: 
```mermaid
classDiagram
    RouterService "1" *-- "n" AggregatorService
    AggregatorService "1" *-- "n" Aggregator

``` 

after:
```mermaid
classDiagram
    AggregatorService "1" *-- "n" AggregatorRouter
    AggregatorRouter "1" *-- "1" Aggregator
```


relay-metrics/src/aggregator.rs
```diff
@@ -2,20 +2,177 @@ use std::collections::hash_map::Entry;
...
+///  - `(Some, None)` if the bucket fits entirely into the size budget. There is no split.
+///  - `(None, Some)` if the size budget cannot even hold the bucket name and tags. There is no
+///    split, the entire bucket is moved.
+///  - `(Some, Some)` if the bucket fits partially. Remaining values are moved into a new bucket
+///    with all other information cloned.
+///
+/// This is an approximate function. The bucket is not actually serialized, but rather its
+/// footprint is estimated through the number of data points contained. See
+/// `estimate_size` for more information.
+fn split_at(mut bucket: Bucket, size: usize) -> (Option<Bucket>, Option<Bucket>) {
```
TBS1996: not related to this pr, but isn't it cleaner if this is a method on bucket?

relay-metrics/src/aggregator.rs
```diff
@@ -487,6 +644,84 @@ impl Aggregator {
         self.cost_tracker.totals_cost_exceeded(max_total_cost)
     }
 
+    /// Split the provided buckets into batches and process each batch with the given function.
+    ///
+    /// For each batch, log a histogram metric.
+    pub fn process_batches(
+        &self,
+        buckets: impl IntoIterator<Item = Bucket>,
+        process: impl Fn(Vec<Bucket>),
```
TBS1996, line 653: also not related, but i find it odd to have a generic `process` when we only call it one. It seems a bit like premature generalization to me

relay-metrics/src/aggregator.rs
```diff
@@ -878,6 +993,22 @@ impl fmt::Debug for Aggregator {
     }
 }
 
+impl Drop for Aggregator {
```
TBS1996: this is changed from `impl Drop for AggregatorService`.

relay-metrics/src/aggregatorservice.rs
```diff
@@ -128,144 +94,250 @@ pub struct FlushBuckets {
...
+        let max_flush_bytes = aggregator_config.max_flush_bytes;
+        let flush_partitions = aggregator_config.flush_partitions;
+
+        if !secondary_aggregators.iter().all(|agg| {
+            agg.config.flush_partitions == flush_partitions
+                && agg.config.max_total_bucket_bytes == max_total_bucket_bytes
+                && agg.config.max_flush_bytes == max_flush_bytes
+        }) {
+            relay_log::error!("aggregatorserviceconfigs are inconsistent");
+        }
```
TBS1996: These three settings used to be on each `AggregatorService`, now we've decided we only need them once for the router, but they are still available on each `AggregatorServiceConfig`, so we log an error if they're inconsistent, and simply use the ones from the default config. 

Perhaps we should have a separate `AggregatorServiceConfig` that doesn't include `AggregatorConfig`?

relay-metrics/src/aggregatorservice.rs
```diff
@@ -128,144 +94,250 @@ pub struct FlushBuckets {
...
+            .aggregators_ref()
+            .map(|agg| agg.name())
+            .zip(flush_buckets.into_iter())
+        {
+            if buckets.is_empty() {
+                continue;
+            };
+
+            self.try_flush(name, buckets);
+        }
```
TBS1996: We need mutable access to the aggregators to take out the flush buckets, so I couldn't take out the flush buckets and the aggregator names (&str) at the same time, so this is the best i came up with it. If anyone know a better way then feel free to give it a go.  
 
It sucks we need to take out the names in the first place, it's only used for metrics and has no functional purpose

relay-metrics/src/aggregatorservice.rs
```diff
@@ -274,11 +346,47 @@ impl AggregatorService {
             Aggregator::MergeBuckets(msg) => self.handle_merge_buckets(msg),
             #[cfg(test)]
             Aggregator::BucketCountInquiry(_, sender) => {
```
TBS1996, line 308: `BucketCountInquiry` was used to test a single aggregator service, now that we got rid of it, i changed it to test the default aggregator, it's only used for testing.

relay-metrics/src/aggregatorservice.rs
```diff
@@ -128,144 +94,250 @@ pub struct FlushBuckets {
...
-/// Receivers must implement a handler for the [`FlushBuckets`] message.
+/// Each aggregator gets its own configuration.
+/// Metrics are routed to the first aggregator which matches the configuration's [`Condition`].
+/// If no condition matches, the metric/bucket is routed to the `default_aggregator`.
 pub struct AggregatorService {
-    aggregator: aggregator::Aggregator,
-    state: AggregatorState,
-    receiver: Option<Recipient<FlushBuckets, NoResponse>>,
+    default_aggregator: aggregator::Aggregator,
+    secondary_aggregators: BTreeMap<MetricNamespace, aggregator::Aggregator>,
```
jjbayer: Part of the motivation for this refactor was to have a "metrics router light" in the pre-aggregator that wraps multiple aggregators, instead of having the suboptimal "lossless" pre-aggregator. This PR should already introduce that wrapper type IMO.

TBS1996: alright, i planned to do it in the next pr but i can just do it now yeah

relay-metrics/src/aggregator.rs
iker-barriocanal: I agree (same with the functions below), we can approach that in a follow-up PR to not overload this one.

relay-metrics/src/aggregatorservice.rs
```diff
@@ -73,14 +40,17 @@ pub enum Aggregator {
...
     #[cfg(test)]
     BucketCountInquiry(BucketCountInquiry, Sender<usize>),
 }
 
 impl Interface for Aggregator {}
 
+/// Check whether the aggregator has not (yet) exceeded its total limits. Used for health checks.
+#[derive(Debug)]
+pub struct AcceptsMetrics;
+
```
iker-barriocanal: Could you move the definition of this struct and the one below after all implementations of `Aggregator`, to keep them together?

relay-metrics/src/aggregatorservice.rs
```diff
@@ -128,144 +94,250 @@ pub struct FlushBuckets {
...
     flush_partitions: Option<u64>,
+    state: AggregatorState,
+    receiver: Option<Recipient<FlushBuckets, NoResponse>>,
 }
 
 impl AggregatorService {
-    /// Create a new aggregator service and connect it to `receiver`.
-    ///
-    /// The aggregator will flush a list of buckets to the receiver in regular intervals based on
-    /// the given `config`.
```
iker-barriocanal, line 159: How does flushing work in this PR, is it documented?

relay-metrics/src/aggregatorservice.rs
iker-barriocanal: Q: why not have a specific config for `AggregatorService`? Not sure if this requires two steps to keep compatibility with the config file.

relay-metrics/src/aggregatorservice.rs
```diff
@@ -128,144 +94,250 @@ pub struct FlushBuckets {
...
-            project_key,
-            buckets,
-        } = msg;
-        self.aggregator
-            .merge_all(project_key, buckets, self.max_total_bucket_bytes);
+    /// Flushes all the aggregators and sends the [`FlushBuckets`] message to the receiver
+    /// in the fire and forget fashion. It is up to the receiver to send the [`MergeBuckets`]
+    /// message back if buckets could not be flushed and we require another re-try.
+    ///
+    /// If `force` is true, flush all buckets unconditionally and do not attempt to merge back.
```
iker-barriocanal: Please keep the docs to the team's coding standards: one short line for the description and then you can elaborate more if appropriate.

iker-barriocanal (commented):
Looking forward to the changes with the pre-aggregator!

For the next time, I'd suggest splitting the PR into smaller ones to make reviews easier (it's hard to review changes in 1,200+k LOC and know what was moved and what's new). Stacking PRs is also a possibility.

relay-metrics/src/aggregatorservice.rs
TBS1996: do you mean what i wrote in the last sentence?

>Perhaps we should have a separate AggregatorServiceConfig that doesn't include AggregatorConfig?

Mainly trying not to make this PR even bigger than it already is, and was unsure if there's some compatibility issues

relay-metrics/src/aggregatorservice.rs
TBS1996: hey i just moved it 😅  
 
But yeah i can change it, I assumed it wasn't to the standard because its a private function

relay-metrics/src/aggregatorservice.rs
TBS1996: good point, i'll update the docs to reflect its new responsibilities

relay-metrics/src/aggregatorservice.rs
TBS1996: maybe we can reconsider this? I agree with @iker-barriocanal  that this PR is already pretty big, and I'd say the current refactor stands well as a PR on its own, where splitting the 'router' and 'routerservice' would be a good PR by itself

relay-metrics/src/aggregator.rs
```diff
@@ -16,6 +17,275 @@ use tokio::time::Instant;
...
+        self.aggregators_mut().map(move |agg| {
+            let flush_buckets = agg.pop_flush_buckets(force);
+            let name = agg.name();
+
+            (name, flush_buckets)
+        })
+    }
+}
+
+impl Drop for AggregatorRouter {
```
TBS1996, line 108: would have made more sense to implement on `Aggregator`, but then the `into_buckets` won't work, due to stricter borrowing rules for types that implement `drop`.

relay-metrics/src/router.rs
```diff
@@ -1,189 +0,0 @@
...
-                    else => break,
-                }
-            }
-            relay_log::info!("metrics router stopped");
-        });
-    }
-}
-
-/// Helper struct that holds the [`Addr`]s of started aggregators.
-struct StartedRouter {
```
TBS1996, line 105: this type isn't needed anymore and thus removed

relay-metrics/src/aggregatorservice.rs
```diff
@@ -300,185 +378,23 @@ impl Service for AggregatorService {
...
-/// An iterator returning batches of buckets fitting into a size budget.
-///
-/// The size budget is given through `max_flush_bytes`, though this is an approximate number. On
-/// every iteration, this iterator returns a `Vec<Bucket>` which serializes into a buffer of the
-/// specified size. Buckets at the end of each batch may be split to fit into the batch.
-///
-/// Since this uses an approximate function to estimate the size of buckets, the actual serialized
-/// payload may exceed the size. The estimation function is built in a way to guarantee the same
-/// order of magnitude.
-struct CappedBucketIter<T: Iterator<Item = Bucket>> {
```
TBS1996: moved to `aggregator` module, along with its' tests

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 - Add `scraping_attempts` field to the event schema. ([#2575](https://github.com/getsentry/relay/pull/2575))
 - Postpone metrics aggregation until we received the project state. ([#2588](https://github.com/getsentry/relay/pull/2588))
 - Move `condition.rs` from `relay-sampling` to `relay-protocol`. ([#2608](https://github.com/getsentry/relay/pull/2608))
+- Remove `AggregationService` and rename `RouterService` to `AggregationService`. ([#2611](https://github.com/getsentry/relay/pull/2611))
```
olksdr: This must go ton unreleased section, together with 
```
 Postpone metrics aggregation until we received the project state. ([#2588](https://github.com/getsentry/relay/pull/2588))`
```

Which was overlooked

relay-metrics/src/aggregator.rs
```diff
@@ -16,6 +17,275 @@ use tokio::time::Instant;
...
+        BucketValue::Gauge(_) => (None, Some(bucket)),
+    }
+}
+
+/// Estimates the number of bytes needed to serialize the bucket without value.
+///
+/// Note that this does not match the exact size of the serialized payload. Instead, the size is
+/// approximated through tags and a static overhead.
+fn estimate_base_size(bucket: &Bucket) -> usize {
+    50 + bucket.name.len() + aggregator::tags_cost(&bucket.tags)
```
olksdr: Can we move this `50` in the constant with the name instead?

relay-metrics/src/aggregatorservice.rs
```diff
@@ -8,61 +11,24 @@ use relay_system::{
...
-    pub max_total_bucket_bytes: Option<usize>,
-
-    /// The number of logical partitions that can receive flushed buckets.
-    ///
-    /// If set, buckets are partitioned by (bucket key % flush_partitions), and routed
-    /// by setting the header `X-Sentry-Relay-Shard`.
-    pub flush_partitions: Option<u64>,
+/// A Bucket and its hashed key.
+/// This is cheaper to pass around than a (BucketKey, Bucket) pair.
+pub struct HashedBucket {
```
olksdr:
```suggestion
/// A Bucket and its hashed key.
///
/// This is cheaper to pass around than a (BucketKey, Bucket) pair.
pub struct HashedBucket {
```

relay-metrics/src/aggregatorservice.rs
jjbayer: 1. I don't believe the diff would be larger if we introduced some modularity here.
2. Think about why we're doing this: The only reason we are touching the router service is that we want to use the routing mechanism in the metrics buffer a.k.a. pre-aggregator. So we have to move it from the service level to the utility level:

Before:

```mermaid
classDiagram
    RouterService "1" *-- "n" AggregatorService
    AggregatorService "1" *-- "1" Aggregator
    Project "1" *-- "1" Aggregator
```

After:

```mermaid
classDiagram
    AggregatorService "1" *-- "1" Router
    Router "1" *-- "n" Aggregator
    Project "1" *-- "1" Router
```

jjbayer (changes requested):

relay-metrics/src/aggregatorservice.rs
TBS1996: should have updated the comment, the PR includes the router already!

CHANGELOG.md
TBS1996: good point, my bad!

relay-metrics/src/aggregator.rs
TBS1996: you mean adding a new constant? I did that now. 

note: I simply moved this function, I did not create it.

relay-metrics/src/aggregatorservice.rs
jjbayer: Right, sorry! I was looking at the PR comment without checking the latest changes.

relay-metrics/src/aggregator.rs
```diff
@@ -16,6 +17,278 @@ use tokio::time::Instant;
...
+        }
+    }
+
+    #[cfg(test)]
+    pub fn default_bucket_count(&self) -> usize {
+        self.default_aggregator.bucket_count()
+    }
+
+    /// Returns an iterator of references to all the aggregators in the [`AggregatorRouter`].
+    fn aggregators_ref(&self) -> impl Iterator<Item = &aggregator::Aggregator> {
```
jjbayer: nit: I would just call this `aggregators()`. See for example `values()`, `values_mut()` on `HashMap`: https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.values

```suggestion
    fn aggregators(&self) -> impl Iterator<Item = &aggregator::Aggregator> {
```

relay-metrics/src/aggregator.rs
```diff
@@ -16,6 +17,278 @@ use tokio::time::Instant;
...
+/// An iterator returning batches of buckets fitting into a size budget.
+///
+/// The size budget is given through `max_flush_bytes`, though this is an approximate number. On
+/// every iteration, this iterator returns a `Vec<Bucket>` which serializes into a buffer of the
+/// specified size. Buckets at the end of each batch may be split to fit into the batch.
+///
+/// Since this uses an approximate function to estimate the size of buckets, the actual serialized
+/// payload may exceed the size. The estimation function is built in a way to guarantee the same
+/// order of magnitude.
+pub struct CappedBucketIter<T: Iterator<Item = Bucket>> {
```
jjbayer: This type is only used in `aggregatorservice` AFAICT. Should live there (that might also help make the diff more readable).

relay-metrics/src/aggregator.rs
```diff
@@ -346,14 +619,6 @@ impl Ord for QueuedBucket {
     }
 }
 
-/// A Bucket and its hashed key.
-/// This is cheaper to pass around than a (BucketKey, Bucket) pair.
-pub struct HashedBucket {
-    // This is only public because pop_flush_buckets is used in benchmark.
-    hashed_key: u64,
-    bucket: Bucket,
-}
```
jjbayer: Why did this type move to `AggregatorService`? It's still used in this file on line 840. IMO the `aggregator` module should import nothing from `aggregatorservice`.

relay-metrics/src/aggregator.rs
```diff
@@ -472,6 +737,21 @@ pub struct Aggregator {
...
+    /// Like [`Self::new`], but with a provided name.
+    pub fn named(name: String, config: AggregatorConfig) -> Self {
+        Self {
+            name,
+            config,
+            buckets: HashMap::new(),
+            cost_tracker: CostTracker::default(),
+        }
+    }
+
```
jjbayer: It makes sense to move these methods to the top of the `impl` block. However, it makes the diff even harder to read, so for the sake of PR reviewers, please create a separate PR next time (IMO it's fine to piggyback changes like this onto small PRs, but not large ones like this one).

relay-metrics/src/aggregator.rs
```diff
@@ -472,6 +737,21 @@ pub struct Aggregator {
 }
 
 impl Aggregator {
+    /// Create a new aggregator.
+    pub fn new(config: AggregatorConfig) -> Self {
```
jjbayer: This is only used in tests, right?

```suggestion
    #[cfg(test)]
    pub fn new(config: AggregatorConfig) -> Self {
```

relay-metrics/src/aggregator.rs
```diff
@@ -16,6 +17,278 @@ use tokio::time::Instant;
...
+        // TODO: Parse MRI only once, move validation from Aggregator here.
+        for (namespace, group) in metrics_by_namespace.into_iter() {
+            let aggregator = namespace
+                .and_then(|ns| self.secondary_aggregators.get_mut(&ns))
+                .unwrap_or(&mut self.default_aggregator);
+
+            aggregator.merge_all(
+                project_key,
+                group.collect::<Vec<Bucket>>(),
+                max_total_bucket_bytes,
```
jjbayer: nit: Now that this is a direct call to the aggregator, we could just call `.merge()` instead of constructing a vector that is then immediately deconstructed by the implementation of `merge_all()`.

relay-metrics/src/aggregatorservice.rs
```diff
@@ -128,157 +94,268 @@ pub struct FlushBuckets {
...
+/// For now, the only way to scope an aggregator is by [`MetricNamespace`].
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ScopedAggregatorConfig {
+    /// Name of the aggregator, used to tag statsd metrics.
+    pub name: String,
+    /// Condition that needs to be met for a metric or bucket to be routed to a
+    /// secondary aggregator.
+    pub condition: Condition,
+    /// The configuration of the secondary aggregator.
+    pub config: AggregatorServiceConfig,
```
jjbayer: Shouldn't this be `AggregatorConfig` (and this type be moved to the `aggregator` module)?
```suggestion
    /// The configuration of the secondary aggregator.
    pub config: AggregatorConfig,
```

jjbayer (commented):
Leaving a preliminary review for now. I think this looks good in general, my main concerns are about code structure.

relay-metrics/src/aggregator.rs
TBS1996: good point, moving it made sense in a previous diff but not anymore, changing it back

relay-metrics/src/aggregator.rs
TBS1996: It had access to the fields directly and i thought i'd avoid making the fields public but yeah probably better just to make them public than changing the location.

relay-metrics/src/aggregator.rs
TBS1996: no, it's used for the default aggregator in the router

relay-metrics/src/aggregator.rs
TBS1996: calling merge directly would just copy the function which isnt very neat, since we use it elsewhere, but i found its possible to directly put the group iterator as an argument so then i guess we avoid the allocations

relay-metrics/src/aggregatorservice.rs
TBS1996: I thought there would be some compatiblity questions, but since not then Ill just change it to aggregatorconfig

relay-metrics/src/aggregator.rs
TBS1996: makes sense!

PR #2613, opened 2023-10-17
feat(normalization): Reject stale events

iker-barriocanal:
There are some cases where events with stale timestamps reach Relay. Relay doesn't perform full normalization on these and forwards them, causing issues in later parts of the pipeline. This PR introduces a new processor to deal with these timestamps, rejecting events with old timestamps as a first step.


relay-event-normalization/src/normalize/mod.rs
```diff
@@ -1021,6 +1022,8 @@ pub fn light_normalize_event(
         // Check for required and non-empty values
         schema::SchemaProcessor.process_event(event, meta, ProcessingState::root())?;
 
+        TimestampProcessor.process_event(event, meta, ProcessingState::root())?;
```
iker-barriocanal, line 1024: The timestamp processor runs before the clock drift processor. With this PR, Relay will drop events created before and coming from systems with a date before January 1, 1970. I believe it's acceptable to assume no systems will have their date set to a value prior to 1970.

relay-event-normalization/src/timestamp.rs
```diff
@@ -0,0 +1,145 @@
...
+    ProcessingAction, ProcessingResult, ProcessingState, Processor,
+};
+use relay_event_schema::protocol::Event;
+use relay_protocol::Meta;
+
+/// Ensures an event's timestamps are not stale.
+///
+/// Stale timestamps are those that happened before January 1, 1970 UTC. The
+/// processor validates the start and end timestamps of an event and returns an
+/// error if any of these is stale.
```
iker-barriocanal: This will evolve as this processor checks for more values.

TBS1996:
```suggestion
/// error if any of these are stale.
```

(i feel so petty doing this)

TBS1996 (approved):
looks good to me

PR #2615, opened 2023-10-17
test(inbound-filters):Add some tests for the real health-check globs

### LINKED ISSUE #58219: Tighten relase health rules to filter more specific paths
Currently the health-check filter uses the following rules to filter transactions:

```python
HEALTH_CHECK_GLOBS = [
    "*healthcheck*",
    "*healthy*",
    "*live*",
    "*ready*",
    "*heartbeat*",
    "*/health",
    "*/healthz",
    "*/ping",
]
```

This is covers too many strings and transactions for paths like `service-delivery` are captured as health-check paths.

The more problematic patterns seem to be `*live*` and `*ready*` since there are may words that may contain them .

All other patterns will be kept as they are.
`*live*` and `*ready*` are replaced by:
`live`,  `live[z/-]*`, `*[/-]live` , `*[/-]live[z/-]*`
`ready`, `ready[z/-]*`, `*[/-]ready` , `*[/-]ready[z/-]*`

**Note:** these are blobs  there is no way I found to write the above in a more concise way ( there is no `or` and there is no optional character ).... `?` means any character one (and only one) time  and `*` means any char any number of times.

```python
HEALTH_CHECK_GLOBS = [
    "*healthcheck*",
    "*healthy*",
    "live",
    "live[z/-]*",
    "*[/-]live" ,
    "*[/-]live[z/-]*"
    "ready",
    "ready[z/-]*",
    "*[/-]ready" ,
    "*[/-]ready[z/-]*",
    "*heartbeat*",
    "*/health",
    "*/healthz",
    "*/ping",
]
```

RaduW:
This PR does not change any functionality !

It just adds a few tests that use the actual blobs sent by Sentry as the health-check patterns to 
verify that they work as expected.

#skip-changelog

Resolves https://github.com/getsentry/sentry/issues/58219

### DIFF
--- a/relay-filter/src/transaction_name.rs
+++ b/relay-filter/src/transaction_name.rs
@@ -41,7 +41,25 @@ mod tests {
     use super::*;
 
     fn _get_patterns() -> GlobPatterns {
-        let patterns_raw = vec!["*healthcheck*".into(), "*/health".into()];
+        let patterns_raw = [
+            "*healthcheck*",
+            "*healthy*",
+            "live",
+            "live[z/-]*",
+            "*[/-]live",
+            "*[/-]live[z/-]*",
+            "ready",
+            "ready[z/-]*",
+            "*[/-]ready",
+            "*[/-]ready[z/-]*",
+            "*heartbeat*",
+            "*/health",
+            "*/healthz",
+            "*/ping",
+        ]
+        .map(|val| val.to_string())
+        .to_vec();
+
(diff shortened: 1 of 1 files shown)


iker-barriocanal (approved):
LGTM, thanks for adding these!

As a side note, could we rename the PR to use the `test` prefix as per the [commit message types](https://develop.sentry.dev/commit-messages/#type)?

olksdr (approved):

PR #2616, opened 2023-10-17
ref(spans): Stop applying clustering rules

jjbayer:
We currently still apply scrubbing rules discovered by [the clusterer](https://develop.sentry.dev/transaction-clustering/) to spans, but this is effectively dead code, because we decided to scrub HTTP spans more aggressively a while ago (see details below).

By removing the clusterer-related code, we can simplify a lot.

Note: I'm leaving the span clustering rules in the project config for now, because there might still be outdated configs that do apply it.

### Details

`apply_span_rename_rules` still gets called, but only if a span has `op:http*` and it does not have a `description.scrubbed` yet:

https://github.com/getsentry/relay/blob/e188c586f81a641a00596a7b39ff87802e835e77/relay-event-normalization/src/normalize/span/description/mod.rs#L239-L240

https://github.com/getsentry/relay/blob/e188c586f81a641a00596a7b39ff87802e835e77/relay-event-normalization/src/normalize/span/description/mod.rs#L261

However, spans with `op:http*` always have a `description.scrubbed` before calling `span_rename_rules`, because we fall back to a method-only description:

https://github.com/getsentry/relay/blob/e188c586f81a641a00596a7b39ff87802e835e77/relay-event-normalization/src/normalize/span/description/mod.rs#L131-L133

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -7,22 +7,20 @@ use std::collections::BTreeMap;
 use std::path::Path;
 
 use itertools::Itertools;
-use relay_event_schema::processor::{self, ProcessingResult};
 use relay_event_schema::protocol::Span;
-use relay_protocol::{Annotated, Remark, RemarkType, Value};
+use relay_protocol::{Annotated, Value};
 use url::Url;
 
 use crate::regexes::{
     DB_SQL_TRANSACTION_CORE_DATA_REGEX, REDIS_COMMAND_REGEX, RESOURCE_NORMALIZER_REGEX,
 };
 use crate::span::tag_extraction::HTTP_METHOD_EXTRACTOR_REGEX;
-use crate::transactions::SpanDescriptionRule;
 
 /// Attempts to replace identifiers in the span description with placeholders.
 ///
 /// The resulting scrubbed description is stored in `data.description.scrubbed`, and serves as input
 /// for the span group hash.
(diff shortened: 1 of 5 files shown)


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -3,29 +3,22 @@ mod sql;
...
 
 /// Attempts to replace identifiers in the span description with placeholders.
 ///
-/// The resulting scrubbed description is stored in `data.description.scrubbed`, and serves as input
-/// for the span group hash.
-pub(crate) fn scrub_span_description(span: &mut Span, rules: &Vec<SpanDescriptionRule>) {
-    let Some(description) = span.description.as_str() else {
-        return;
-    };
+/// Returns `None` if no Scrubbing can be performed.
```
iker-barriocanal, line 19:
```suggestion
/// Returns `None` if no scrubbing can be performed.
```

iker-barriocanal (approved):
Negative LOC and removing tech debt? Love it! 🚀

Should we add a note on the span description clustering rules in the project configs that they only exist there for old relays?

olksdr (approved):
:shipit:

relay-event-normalization/src/normalize/span/description/mod.rs
jjbayer: Whoops, forgot that I still wanted to address these comments before merging, sorry.

PR #2618, opened 2023-10-17
fix(metrics): Fix the measurement unit for ttfd

shruthilayaj:
I think we meant to say "time_to_full_display" here
since that's the measurement the mobile sdks send!

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/mod.rs
+++ b/relay-event-normalization/src/normalize/mod.rs
@@ -412,7 +412,7 @@ fn get_metric_measurement_unit(measurement_name: &str) -> Option<MetricUnit> {
         "frames_frozen" => Some(MetricUnit::None),
         "frames_frozen_rate" => Some(MetricUnit::Fraction(FractionUnit::Ratio)),
         "time_to_initial_display" => Some(MetricUnit::Duration(DurationUnit::MilliSecond)),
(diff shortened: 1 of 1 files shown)


relay-event-normalization/src/normalize/mod.rs
```diff
@@ -412,7 +412,7 @@ fn get_metric_measurement_unit(measurement_name: &str) -> Option<MetricUnit> {
         "frames_frozen" => Some(MetricUnit::None),
         "frames_frozen_rate" => Some(MetricUnit::Fraction(FractionUnit::Ratio)),
         "time_to_initial_display" => Some(MetricUnit::Duration(DurationUnit::MilliSecond)),
-        "time_to_first_display" => Some(MetricUnit::Duration(DurationUnit::MilliSecond)),
+        "time_to_full_display" => Some(MetricUnit::Duration(DurationUnit::MilliSecond)),
```
iker-barriocanal, line 415: Are we ingesting this measurement already? Should we support both to keep compatibility with old SDKs?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/mod.rs
shruthilayaj: We're ingesting `time_to_full_display` already! We never had a `time_to_first_display` measurement, I think it was just accidentally added as `time_to_first_display` here (cc: @wmak can confirm!)

PR #2620, opened 2023-10-17
feat(spans): Ingest standalone spans

phacops:
This PR aims to add support for ingesting standalone spans. This will support 3 ways to ingest spans:
- an HTTP endpoint accepting OpenTelemetry spans, encoded as JSON
- an `Item` of type `OtelSpan` accepting an OpenTelemetry span, encoded as JSON
- an `Item` of type `Span` accepting a span matching the span protocol, encoded as JSON

After we accept those spans, they are converted to a span as the protocol describes it and they each go through the span processor which will: 
- normalize
- validate
- extract metrics

They will then be sent to Kafka for storage.

In the end, we're aiming for this except our common format is the span protocol:
image (4)


mydea:
Here some OTLP spans I generated from a node-experimental project:

```json
{
   "traceId": "89143b0763095bd9c9955e8175d1fb23",
   "spanId": "e342abb1214ca181",
   "parentSpanId": "0c7a7dea069bf5a6",
   "name": "middleware - fastify -> @fastify/multipart",
   "kind": 1,
   "startTimeUnixNano": 1697620454980000000,
   "endTimeUnixNano": 1697620454980078800,
   "attributes": [
     {
       "key": "fastify.type",
       "value": {
         "stringValue": "middleware"
       }
     },
     {
       "key": "plugin.name",
       "value": {
         "stringValue": "fastify -> @fastify/multipart"
       }
     },
     {
       "key": "hook.name",
       "value": {
         "stringValue": "onResponse"
       }
     },
     {
       "key": "sentry.sample_rate",
       "value": {
         "intValue": 1
       }
     },
     {
       "key": "sentry.parentSampled",
       "value": {
         "boolValue": true
       }
     }
   ],
   "droppedAttributesCount": 0,
   "events": [],
   "droppedEventsCount": 0,
   "status": {
     "code": 0
   },
   "links": [],
   "droppedLinksCount": 0
 }
```

```json
{
  "traceId": "89143b0763095bd9c9955e8175d1fb23",
  "spanId": "0c7a7dea069bf5a6",
  "name": "GET /bottles",
  "kind": 2,
  "startTimeUnixNano": 1697620454925000000,
  "endTimeUnixNano": 1697620454979637200,
  "attributes": [
    {
      "key": "http.url",
      "value": {
        "stringValue": "http://localhost:4000/bottles"
      }
    },
    {
      "key": "http.host",
      "value": {
        "stringValue": "localhost:4000"
      }
    },
    {
      "key": "net.host.name",
      "value": {
        "stringValue": "localhost"
      }
    },
    {
      "key": "http.method",
      "value": {
        "stringValue": "GET"
      }
    },
    {
      "key": "http.scheme",
      "value": {
        "stringValue": "http"
      }
    },
    {
      "key": "http.target",
      "value": {
        "stringValue": "/bottles"
      }
    },
    {
      "key": "http.user_agent",
      "value": {
        "stringValue": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36"
      }
    },
    {
      "key": "http.flavor",
      "value": {
        "stringValue": "1.1"
      }
    },
    {
      "key": "net.transport",
      "value": {
        "stringValue": "ip_tcp"
      }
    },
    {
      "key": "sentry.sample_rate",
      "value": {
        "intValue": 1
      }
    },
    {
      "key": "sentry.origin",
      "value": {
        "stringValue": "auto.http.otel.http"
      }
    },
    {
      "key": "net.host.ip",
      "value": {
        "stringValue": "::1"
      }
    },
    {
      "key": "net.host.port",
      "value": {
        "intValue": 4000
      }
    },
    {
      "key": "net.peer.ip",
      "value": {
        "stringValue": "::1"
      }
    },
    {
      "key": "net.peer.port",
      "value": {
        "intValue": 60502
      }
    },
    {
      "key": "http.status_code",
      "value": {
        "intValue": 500
      }
    },
    {
      "key": "http.status_text",
      "value": {
        "stringValue": "INTERNAL SERVER ERROR"
      }
    },
    {
      "key": "http.route",
      "value": {
        "stringValue": "/bottles"
      }
    }
  ],
  "droppedAttributesCount": 0,
  "events": [],
  "droppedEventsCount": 0,
  "status": {
    "code": 2
  },
  "links": [],
  "droppedLinksCount": 0
 }
```

relay-server/src/endpoints/spans.rs
```diff
@@ -0,0 +1,45 @@
...
+    meta: RequestMeta,
+    extract::Json(span): extract::Json<Span>,
+) -> Result<impl IntoResponse, BadStoreRequest> {
+    println!("{:#?}", span);
+
+    // Convert to an Event span.
+    let event_span: Annotated<EventSpan> = Annotated::new(span.into());
+
+    // Pack into an envelope for further processing.
+    let span_json = event_span.to_json().map_err(BadStoreRequest::InvalidJson)?;
```
jjbayer: We probably don't want to deserialize or convert anything between request and response, to keep the response time as short as possible (see for example the security report endpoint). Instead, we can add a new item type `OtelSpan` that contains the raw JSON, and then unpack it in `process_` function (as we do for profiles).

relay-spans/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub start_time_unix_nano: i64,
+    pub end_time_unix_nano: i64,
+    pub attributes: Vec<KeyValue>,
+    pub dropped_attributes_count: u32,
+    pub events: Vec<Event>,
+    pub dropped_events_count: u32,
+    pub links: Vec<Link>,
+    pub dropped_links_count: u32,
+    pub status: Option<Status>,
+}
```
jjbayer: For fields that are not required, we should add `#[serde(default)]` and `#[serde(skip_serializing_if)]` attributes.

relay-server/src/actors/processor.rs
```diff
@@ -2427,6 +2427,29 @@ impl EnvelopeProcessorService {
         Ok(span)
     }
 
+    #[cfg(feature = "processing")]
+    fn process_spans(&self, state: &mut ProcessEnvelopeState) {
```
jjbayer: I think we also want a `filter_spans` method that drops spans in customer relays and/or PoPs when the feature flag is enabled.

relay-spans/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct InstrumentationScope {
+    pub name: String,
+    pub version: String,
+    pub attributes: Vec<KeyValue>,
+    pub dropped_attributes_count: u32,
+}
+
+impl From<Span> for EventSpan {
```
jjbayer: @phacops we briefly discussed this offline: I'm wondering if instead of converting to `EventSpan`, we should convert event spans to `OtelSpan` instead when we extract them from transactions?

I realize that this would slow down development because we'd have to make sure that the sentry consumer supports the OTEL format first, so I am also fine with converting to event spans for now.

jjbayer (commented):
Good stuff! Long-term we'll need a rate limiting and dynamic sampling strategy for these spans, but as long as the ingestion's feature-flagged we can ship without it IMO.

relay-spans/src/lib.rs
phacops: Yes, that's precisely why I started like this. I realize in the long term, we want the reverse but we're long from being ready for this.

I think in the interest of figuring out how an OTel SDK would interact with this, it's still worth doing this and then reversing when we're ready.

jjbayer: Makes sense!

relay-dynamic-config/src/feature.rs
```diff
@@ -31,6 +31,9 @@ pub enum Feature {
     /// Enable extracting resource spans.
     #[serde(rename = "projects:span-metrics-extraction-resource")]
     SpanMetricsExtractionResource,
+    /// Enable OTel span ingestion.
+    #[serde(rename = "organizations:otel-span-ingestion")]
+    OTelSpanIngestion,
```
jjbayer: nit

https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case

```suggestion
    OtelSpanIngestion,
```

relay-server/src/actors/processor.rs
```diff
@@ -2427,6 +2428,59 @@ impl EnvelopeProcessorService {
...
+
+    #[cfg(feature = "processing")]
+    fn process_spans(&self, state: &mut ProcessEnvelopeState) {
+        let mut items: Vec<Item> = Vec::new();
+        state.managed_envelope.retain_items(|item| match item.ty() {
+            ItemType::OtelSpan => {
+                let span: relay_spans::Span = match serde_json::from_slice(&item.payload()) {
+                    Ok(span) => span,
+                    Err(e) => {
+                        relay_log::error!("Invalid span: {e}");
```
```suggestion
                        relay_log::error!(error = &e as &dyn Error, "failed to parse OTel span");
```

relay-server/src/actors/processor.rs
```diff
@@ -2427,6 +2428,59 @@ impl EnvelopeProcessorService {
...
+                        return ItemAction::DropSilently;
+                    }
+                };
+                let event_span: Annotated<Span> = Annotated::new(span.into());
+                match self.validate_span(event_span) {
+                    Ok(span) => {
+                        let payload: Bytes = match span.to_json() {
+                            Ok(payload) => payload.into(),
+                            Err(e) => {
+                                relay_log::error!("Invalid span: {e}");
```
```suggestion
                                relay_log::error!(
                                    error = &e as &dyn Error,
                                    "Failed to serialize OTel span"
                                );
```

relay-spans/src/span.rs
```diff
@@ -0,0 +1,350 @@
...
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Span {
+    pub trace_id: String,
+    pub span_id: String,
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub trace_state: String,
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub parent_span_id: Option<String>,
```
jjbayer: Could not find a formal OTel schema definition, but it seems that `parent_span_id` is an empty string for root spans: https://opentelemetry.io/docs/concepts/signals/traces/#spans

relay-spans/src/span.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub name: String,
+    pub kind: SpanKind,
+    pub start_time_unix_nano: i64,
+    pub end_time_unix_nano: i64,
```
jjbayer: What formal schema are we implementing? Is it this one? https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto.  cc @mydea 

Asking because on [this page](https://opentelemetry.io/docs/concepts/signals/traces/#spans) I only see `start_time` and `end_time` as strings, no `_unix_nano` suffix.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub name: String,
+    pub attributes: Vec<KeyValue>,
+    pub dropped_attributes_count: u32,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Link {
+    pub trace_id: String,
+    pub span_id: String,
+    pub trace_state: String,
```
jjbayer: Same question for all the fields on all the sub-structs: Is this field optional? If so, `#[serde(default)]`.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,350 @@
...
+        match value {
+            "SPAN_KIND_UNSPECIFIED" => Some(Self::Unspecified),
+            "SPAN_KIND_INTERNAL" => Some(Self::Internal),
+            "SPAN_KIND_SERVER" => Some(Self::Server),
+            "SPAN_KIND_CLIENT" => Some(Self::Client),
+            "SPAN_KIND_PRODUCER" => Some(Self::Producer),
+            "SPAN_KIND_CONSUMER" => Some(Self::Consumer),
+            _ => None,
+        }
+    }
```
jjbayer: Can we remove these methods? Doesn't seem like we need them anywhere.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub fn from_str_name(value: &str) -> ::core::option::Option<Self> {
+        match value {
+            "STATUS_CODE_UNSET" => Some(Self::Unset),
+            "STATUS_CODE_OK" => Some(Self::Ok),
+            "STATUS_CODE_ERROR" => Some(Self::Error),
+            _ => None,
+        }
+    }
+}
```
jjbayer: Same question about removing these.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,350 @@
...
+            span.parent_span_id = SpanId(parent_span_id).into();
+        } else {
+            span.is_segment = true.into();
+            span.segment_id = span.span_id.clone();
+        }
+        span
+    }
+}
+
+impl Span {
```
jjbayer: nit: let's put this block right after `Span`.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,350 @@
...
+        assert_eq!(
+            event_span
+                .value()
+                .unwrap()
+                .data
+                .value()
+                .unwrap()
+                .get("sentry.environment")
+                .unwrap()
+                .as_str(),
```
jjbayer: nit: You can probably use `get_path!` here.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,350 @@
...
+use serde_repr::Deserialize_repr;
+
+use relay_event_schema::protocol::{Span as EventSpan, SpanId, SpanStatus, Timestamp, TraceId};
+use relay_protocol::{Annotated, Object, Value};
+
+use crate::status_codes;
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Span {
```
jjbayer: Took me a while to realize that we're implementing [this schema](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto), might be worth linking in the doc comments.

relay-server/src/endpoints/spans.rs
```diff
@@ -0,0 +1,45 @@
...
+        return Ok(StatusCode::UNSUPPORTED_MEDIA_TYPE);
+    }
+
+    if body.is_empty() {
+        return Err(BadStoreRequest::EmptyBody);
+    }
+
+    let mut item = Item::new(ItemType::OtelSpan);
+    item.set_payload(ContentType::Json, body);
+    let mut envelope = Envelope::from_request(Some(EventId::new()), meta);
```
jjbayer:
```suggestion
    let mut envelope = Envelope::from_request(None, meta);
```

relay-spans/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+pub use crate::span::Span;
```
jjbayer: Let's be pedantic about docs.

```suggestion
#![warn(missing_docs)]
#![doc(
    html_logo_url = "https://raw.githubusercontent.com/getsentry/relay/master/artwork/relay-icon.png",
    html_favicon_url = "https://raw.githubusercontent.com/getsentry/relay/master/artwork/relay-icon.png"
)]

pub use crate::span::Span;
```

relay-spans/src/span.rs
phacops: It's indeed this one we're implementing. All SDKs auto-generated those structures and would send us a JSON structure based on this protobuf file.

The other page you link is the more user friendly implementation to build a span in the SDK.

Ideally, I would want us to use the auto-generated version eventually but they seemed to have only files meant to be used with their GPRC/tonic implementations and I wanted serde + JSON. Eventually it'd be good to migrate to something auto-generated from the protobuf though.

relay-spans/src/span.rs
phacops: I intended to remove the `Option` as you suggested before, so I'll fix this. Are you also suggesting to remove the `skip_serializing_if` in case the string is empty so we propagate the empty value in case of a root span?

jjbayer: > Are you also suggesting to remove the skip_serializing_if in case the string is empty so we propagate the empty value in case of a root span?

No, `skip_serializing_if` is fine here IMO.

relay-spans/src/span.rs
mydea: hmm good point, I generated the example based on what opentelemetry JS exports for OTLP 🤔

relay-server/src/actors/processor.rs
```diff
@@ -2489,6 +2490,62 @@ impl EnvelopeProcessorService {
...
+            _ => ItemAction::Keep,
+        });
+    }
+
+    #[cfg(feature = "processing")]
+    fn process_spans(&self, state: &mut ProcessEnvelopeState) {
+        let mut items: Vec<Item> = Vec::new();
+        state.managed_envelope.retain_items(|item| match item.ty() {
+            ItemType::OtelSpan => {
+                let span: relay_spans::Span = match serde_json::from_slice(&item.payload()) {
```
jjbayer: nit: Could use [`let .. else`](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) here.

relay-server/src/actors/processor.rs
```diff
@@ -2489,6 +2490,62 @@ impl EnvelopeProcessorService {
...
+                    Ok(span) => span,
+                    Err(e) => {
+                        relay_log::error!(error = &e as &dyn Error, "failed to parse OTel span");
+                        return ItemAction::DropSilently;
+                    }
+                };
+                let event_span: Annotated<Span> = Annotated::new(span.into());
+                match self.validate_span(event_span) {
```
jjbayer, line 2518: nit: Could use `let .. else` here.

relay-server/src/actors/processor.rs
```diff
@@ -2489,6 +2490,62 @@ impl EnvelopeProcessorService {
...
+                        let mut new_item = Item::new(ItemType::Span);
+                        new_item.set_payload(ContentType::Json, payload);
+                        items.push(new_item);
+                        ItemAction::DropSilently
+                    }
+                    Err(e) => {
+                        relay_log::error!("Invalid span: {e}");
+                        ItemAction::DropSilently
+                    }
+                }
```
jjbayer: nit: We have a mutable reference to `item` here, so instead of pushing to a new vector, we could just write straight into the item.

relay-server/src/envelope.rs
```diff
@@ -162,6 +164,7 @@ impl fmt::Display for ItemType {
             Self::ReplayRecording => write!(f, "replay_recording"),
             Self::CheckIn => write!(f, "check_in"),
             Self::Span => write!(f, "span"),
+            Self::OtelSpan => write!(f, "otel_span"),
```
jjbayer, line 170: Please also add an entry to `from_str` below. Unfortunately the compiler does not catch this.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,447 @@
...
+use relay_protocol::{Annotated, Object, Value};
+
+use crate::otel_to_sentry_tags::OTEL_TO_SENTRY_TAGS;
+use crate::status_codes;
+
+/// This is a serde implementation of <https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto>.
+/// A Span represents a single operation performed by a single component of the system.
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Span {
```
jjbayer: Should we call this type `OtelSpan`? Just to be super clear what it is. Then at a later point, when we invert the pipeline and convert event spans into otel spans, we can rename this type back to `Span`.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,447 @@
...
+        let mut attributes: Object<Value> = Object::new();
+        let start_timestamp = Utc.timestamp_nanos(from.start_time_unix_nano);
+        for attribute in from.attributes.clone() {
+            let key: String = if let Some(key) = OTEL_TO_SENTRY_TAGS.get(attribute.key.as_str()) {
+                key.to_string()
+            } else {
+                attribute.key
+            };
+            match attribute.value {
+                AnyValue::Array(_) => todo!(),
```
jjbayer: Should we fix the `todo!()`s here? We should be able to convert to `Object` or `Array`.

relay-server/src/actors/processor.rs
jjbayer: `validate_span` currently relies on certain validations by the `TransactionProcessor`, see https://github.com/getsentry/relay/blob/dda78339e126f9bf6800861efd5ea871aecb80f6/relay-server/src/actors/processor.rs#L2412-L2416. We should replicate this validation somehow (happy to contribute).

relay-server/src/actors/processor.rs
phacops: Does it mean I need to drop the error logging or is there a way to do it with error logging as well?

jjbayer: Oh right, you can't get the error info that way. Ignore me.

relay-server/src/actors/processor.rs
phacops: I was doing this to avoid looping on new items.

relay-spans/src/span.rs
phacops: I removed the `todo!()`. I don't know yet how we need to handle the `Kvlist` or `Array` as we don't support that in our Sentry tags.

We should stringify them one way or another and show their values or flatten them. It's not blocking now because we already don't support them so we can make a decision on this later.

relay-server/src/actors/processor.rs
```diff
@@ -2504,6 +2524,110 @@ impl EnvelopeProcessorService {
...
+                crate::metrics_extraction::generic::extract_metrics(
+                    span,
+                    span_metrics_extraction_config,
+                );
+            }
+        };
+        for item in state.managed_envelope.envelope().items() {
+            if item.ty() == &ItemType::OtelSpan {
+                if let Ok(otel_span) =
+                    serde_json::from_slice::<relay_spans::OtelSpan>(&item.payload())
```
jjbayer: Instead of deserializing again here (we also deserialize in `process_spans`), let's have a more linear pipeline inside `process_spans`, i.e.

```
FOR EACH span item
    1. Convert otel to sentry span if needed
    2. validate / normalize
    3. extract metrics
    4. write back into envelope
```

relay-server/src/actors/processor.rs
```diff
@@ -2504,6 +2524,110 @@ impl EnvelopeProcessorService {
...
+            .has_feature(Feature::StandaloneSpanIngestion)
+        {
+            return;
+        }
+        let span_metrics_extraction_config = match state.project_state.config.metric_extraction {
+            ErrorBoundary::Ok(ref config) if config.is_enabled() => config,
+            _ => return,
+        };
+        let extract_span_metrics = |span: Annotated<Span>| {
+            let span = match self.validate_span(span) {
```
jjbayer: I will check event normalization to see if any more validation and/or normalization is needed.

relay-server/src/envelope.rs
```diff
@@ -114,6 +114,10 @@ pub enum ItemType {
     CheckIn,
     /// A standalone span.
     Span,
+    /// A standalone OpenTelemetry span.
+    OtelSpan,
+    /// A standalone span.
+    EventSpan,
```
jjbayer: I think we can make this work without an additional item type, and use `Span` for ingesting sentry-style spans directly.

```suggestion
```

relay-server/src/actors/processor.rs
```diff
@@ -2799,10 +2924,16 @@ impl EnvelopeProcessorService {
             });
         }
 
+        // Extract span metrics from standalone spans only
+        if self.inner.config.processing_enabled() || state.sampling_result.should_drop() {
+            self.extract_span_metrics(state);
+        }
+
```
jjbayer: Let's call `extract_span_metrics` from within `process_spans` (see other comments).

```suggestion
```

relay-server/src/actors/processor.rs
```diff
@@ -2504,6 +2524,110 @@ impl EnvelopeProcessorService {
...
+            };
+            if let Ok(payload) = span.to_json() {
+                let mut item = Item::new(ItemType::Span);
+                item.set_payload(ContentType::Json, payload);
+                items.push(item);
+            }
+        };
+        let standalone_span_ingestion_enabled = state
+            .project_state
+            .has_feature(Feature::StandaloneSpanIngestion);
```
jjbayer: nit: In theory we don't have to check this feature flag here because it is already done in `filter_spans`.

jjbayer (changes requested):
I will contribute an integration test to see if spans & metrics make it through to kafka.

relay-server/src/actors/processor.rs
phacops: So, we'll only extract metrics on processing nodes?

Also, this works while we're not dynamically sampling spans, but when we do, we'll need to split this again.

relay-server/src/actors/processor.rs
phacops: Could someone send spans through an old Relay version (not having this filtering logic) and somehow have spans be ingested?

I added this logic to potentially protect against cases like that.

relay-server/src/actors/processor.rs
phacops: So, we'll only extract metrics on processing nodes?

Also, this works while we're not dynamically sampling spans, but when we do, we'll need to split this again.

jjbayer: Yes, I would keep everything in processing nodes for now, so we can iterate fast without having to think about outdated external relays or duplicate metrics extraction.

relay-server/src/actors/processor.rs
jjbayer: > Could someone send spans through an old Relay version (not having this filtering logic) and somehow have spans be ingested?

Those would still get caught by the `filter_spans` function of the newer Relay, so I think it's safe to remove here.

relay-spans/src/span.rs
```diff
@@ -0,0 +1,451 @@
...
+        }
+
+        "unknown_error"
+    }
+}
+
+impl From<OtelSpan> for EventSpan {
+    fn from(from: OtelSpan) -> Self {
+        let end_timestamp = Utc.timestamp_nanos(from.end_time_unix_nano);
+        let exclusive_time = (from.end_time_unix_nano - from.start_time_unix_nano) as f64 / 1e6f64;
```
jjbayer: Are we running into a fundamental problem with span streaming here? Exclusive time cannot be computed without knowing how much time the child spans took up.

phacops: I don't think it's a fundamental problem, SDKs will have to track this themselves and make it an attribute instead.

It's handled like that here because I needed to make a valid span.

relay-server/src/actors/processor/span.rs
```diff
@@ -0,0 +1,418 @@
...
+#[cfg(feature = "processing")]
+#[derive(Clone, Debug)]
+pub struct NormalizeSpanConfig {
+    /// The time at which the event was received in this Relay.
+    pub received_at: DateTime<Utc>,
+    /// Allowed time range for transactions.
+    pub transaction_range: std::ops::Range<UnixTimestamp>,
+    /// The maximum allowed size of tag values in bytes. Longer values will be cropped.
+    pub max_tag_value_size: usize,
+}
```
jjbayer: Does not need to be `pub` anymore.
```suggestion
/// Config needed to normalize a standalone span.
#[cfg(feature = "processing")]
#[derive(Clone, Debug)]
struct NormalizeSpanConfig {
    /// The time at which the event was received in this Relay.
    pub received_at: DateTime<Utc>,
    /// Allowed time range for transactions.
    pub transaction_range: std::ops::Range<UnixTimestamp>,
    /// The maximum allowed size of tag values in bytes. Longer values will be cropped.
    pub max_tag_value_size: usize,
}
```

relay-server/src/actors/processor/span.rs
```diff
@@ -0,0 +1,418 @@
...
+
+        *item = new_item;
+
+        ItemAction::Keep
+    });
+}
+
+/// We do not extract spans with missing fields if those fields are required on the Kafka topic.
+#[cfg(feature = "processing")]
+pub fn validate(mut span: Annotated<Span>) -> Result<Annotated<Span>, anyhow::Error> {
```
jjbayer: Can be private now.

```suggestion
fn validate(mut span: Annotated<Span>) -> Result<Annotated<Span>, anyhow::Error> {
```

relay-server/src/actors/processor/span.rs
```diff
@@ -0,0 +1,418 @@
...
+        || op.starts_with("db")
+            && !(op.contains("clickhouse")
+                || op.contains("mongodb")
+                || op.contains("redis")
+                || op.contains("compiler"))
+            && !(op == "db.sql.query" && (description.contains("\"$") || system == "mongodb"))
+}
+
+#[cfg(feature = "processing")]
+pub fn extract_from_event(state: &mut ProcessEnvelopeState) {
```
jjbayer, line 145: Let's move all public functions to the top of the file.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 **Features**:
 
+- Ingest OpenTelemetry spans via HTTP or an envelope. ([#2620](https://github.com/getsentry/relay/pull/2620))
```
```suggestion
- Ingest OpenTelemetry and standalone Sentry spans via HTTP or an envelope. ([#2620](https://github.com/getsentry/relay/pull/2620))
```

relay-server/src/actors/processor/span.rs
```diff
@@ -0,0 +1,418 @@
...
+        }
+        (None, _) => {
+            return Err(anyhow::anyhow!("start_timestamp hard-required for spans"));
+        }
+    }
+
+    // `is_segment` is set by `extract_span`.
+    exclusive_time
+        .value()
+        .ok_or(anyhow::anyhow!("missing exclusive_time"))?;
```
jjbayer: This check means that we only accept spans for which the SDK sets `exclusive_time`. So only from a Sentry SDK, not any other client.

relay-spans/src/otel_to_sentry_tags.rs
```diff
@@ -0,0 +1,36 @@
...
+        ("enduser.id", "user.id"),
+        ("http.request.cookies", "request.cookies"),
+        ("http.request.env", "request.env"),
+        (
+            "http.request.headers.content-type",
+            "request.headers.content-type",
+        ),
+        ("http.request.method", "request.method"),
+        ("sentry.environment", "environment"),
+        ("sentry.exclusive_time_ns", "exclusive_time_ns"),
```
jjbayer: TODO: Why is there a mapping for this?

relay-spans/src/span.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// A developer-facing human readable error message.
+    #[serde(default)]
+    pub message: String,
+    /// The status code.
+    #[serde(default)]
+    pub code: StatusCode,
+}
+
+#[derive(Clone, Debug, Deserialize_repr, Default, PartialEq)]
+#[repr(u32)]
```
jjbayer:
```suggestion
#[repr(u8)]
```

Any reason for `u32` here?

relay-spans/src/span.rs
```diff
@@ -0,0 +1,459 @@
...
+pub struct Status {
+    /// A developer-facing human readable error message.
+    #[serde(default)]
+    pub message: String,
+    /// The status code.
+    #[serde(default)]
+    pub code: StatusCode,
+}
+
+#[derive(Clone, Debug, Deserialize_repr, Default, PartialEq)]
```
jjbayer:
```suggestion
#[derive(Clone, Copy, Debug, Deserialize_repr, Default, PartialEq)]
```

relay-spans/src/span.rs
```diff
@@ -0,0 +1,459 @@
...
+}
+
+impl OtelSpan {
+    /// sentry_status() returns a status as defined by Sentry based on the span status.
+    pub fn sentry_status(&self) -> &'static str {
+        let status_code = self.status.code.clone();
+
+        if status_code == StatusCode::Unset || status_code == StatusCode::Ok {
+            return "ok";
+        }
```
jjbayer: Might be a matter or personal taste, but why not

```suggestion
        if self.status.code == StatusCode::Unset || self.status.code == StatusCode::Ok {
            return "ok";
        }
```

relay-spans/src/span.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        if let Some(code) = self
+            .attributes
+            .clone()
+            .into_iter()
+            .find(|a| a.key == "http.status_code")
```
jjbayer:
```suggestion
        if let Some(code) = self.attributes.iter().find(|a| a.key == "http.status_code") {
```

relay-spans/src/span.rs
```diff
@@ -0,0 +1,459 @@
...
+                    return sentry_status;
+                }
+            }
+        }
+
+        if let Some(code) = self
+            .attributes
+            .clone()
+            .into_iter()
+            .find(|a| a.key == "rpc.grpc.status_code")
```
jjbayer:
```suggestion
        if let Some(code) = self
            .attributes
            .iter()
            .find(|a| a.key == "rpc.grpc.status_code")
```

relay-spans/src/span.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+}
+
+impl From<OtelSpan> for EventSpan {
+    fn from(from: OtelSpan) -> Self {
+        let mut exclusive_time_ms = 0f64;
+        let mut attributes: Object<Value> = Object::new();
+        let start_timestamp = Utc.timestamp_nanos(from.start_time_unix_nano);
+        let end_timestamp = Utc.timestamp_nanos(from.end_time_unix_nano);
+        for attribute in from.attributes.clone() {
```
jjbayer: To prevent unnecessary clones, deconstruct and then use the fields directly:

```suggestion
        let OtelSpan {
            trace_id,
            span_id,
            parent_span_id,
            name,
            attributes: from_attributes,
            ..
        } = from;

        for attribute in from_attributes.clone() {
```

jjbayer (commented):
Almost there!

relay-spans/src/span.rs
phacops: I wanted to avoid the `clone` twice but since you added `Copy` above, I guess it's handled for me so sure.

relay-spans/src/span.rs
phacops: I was about to say I wanted to stick with what was [generated](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-proto/src/proto/grpcio/opentelemetry.proto.trace.v1.rs#L305C1-L305C1) but turns out, I'm even changing that to unsigned integer, so might as well use a `u8`.

relay-spans/src/span.rs
phacops: I tried to push that as much as I could, I still needed some `clone()` for attributes, let me know if I actually made the situation completely better or just changed it.

relay-spans/src/otel_to_sentry_tags.rs
phacops: I thought it might be a good idea to have it but thinking about it more, it's useless, so I removed it.

relay-server/src/actors/processor/span.rs
phacops: I changed it so the fallback is to calculate the duration of the span using start and end timestamps. Like this, the validation still works as intended.

relay-spans/src/span.rs
```diff
@@ -112,59 +112,67 @@ pub struct OtelSpan {
...
+        for attribute in attributes.iter() {
             let key: String = if let Some(key) = OTEL_TO_SENTRY_TAGS.get(attribute.key.as_str()) {
                 key.to_string()
             } else {
-                attribute.key
+                attribute.key.clone()
             };
-            if key == "exclusive_time_ns" {
-                let value = match attribute.value {
+            if key.contains("exclusive_time_ns") {
```
jjbayer, line 173: Is `.contains` really what we want here? Don't we want to check for a specific key (or a set of keys)?

relay-spans/src/span.rs
```diff
@@ -173,52 +180,62 @@ impl From<OtelSpan> for EventSpan {
...
                 }
             };
         }
-        let mut span = EventSpan {
-            data: attributes.into(),
-            description: from.name.clone().into(),
+        if exclusive_time_ms == 0f64 {
+            exclusive_time_ms =
+                (from.end_time_unix_nano - from.start_time_unix_nano) as f64 / 1e6f64;
+        }
```
jjbayer, line 213: This is ultimately a product decision, but I don't think we should assume that a span has zero child spans just because the sentry-specific "exclusive time" attribute is missing. I'd rather drop these spans for now than ingest incorrect data for them. 

```suggestion
```

relay-spans/src/span.rs
jjbayer: 👍 I pushed a commit to reduce clones even further.

relay-server/src/actors/processor/span.rs
jjbayer: See other comment:

> This is ultimately a product decision, but I don't think we should assume that a span has zero child spans just because the sentry-specific "exclusive time" attribute is missing. I'd rather drop these spans for now than ingest incorrect data for them.

jjbayer (approved):
I still have some reservations around the handling of "exclusive time", but other than that, this looks good to me!

relay-spans/src/span.rs
phacops: We can check for 2 specific keys indeed, I thought that'd be fine to just use `contains` instead of repeating nearly exact key names.

relay-spans/src/span.rs
phacops: I think I'd rather keep them for now. Obviously, this is something that will change as we understand how to use this better from an OTLP point of view but I wouldn't reject them right now.

This is not meant to be used right away by customers and what we do with this field has to change before we release it but it's important to let us ingest some data, maybe a bit wrong at first, to continue testing.

phacops:
I understand your reservations, I don't think we'll stick with this but it's important to get us started by not rejecting those spans until we have a solution. We'll likely have to talk to a few people to see what we'll need to do here.

PR #2627, opened 2023-10-18
feat(normalization): Remove stale spans from transactions

iker-barriocanal:
There are some cases where spans with stale timestamps reach Relay. Relay doesn't perform full normalization on these and forwards them, causing issues in later parts of the pipeline. This PR extends the `TimestampProcessor` to remove stale spans from events.


relay-event-normalization/src/timestamp.rs
```diff
@@ -39,15 +39,45 @@ impl Processor for TimestampProcessor {
             }
         }
 
+        event.process_child_values(self, state)?;
+
+        Ok(())
+    }
+
+    fn process_span(
```
iker-barriocanal, line 47: Docs for the processor are up-to-date, I accidentally updated them in the previous PR.

relay-event-normalization/src/timestamp.rs
```diff
@@ -39,15 +39,45 @@ impl Processor for TimestampProcessor {
...
+        }
+        if let Some(end_timestamp) = span.timestamp.value() {
+            if end_timestamp.into_inner().timestamp_millis() < 0 {
+                meta.add_error(Error::invalid(format!(
+                    "timestamp is too stale: {}",
+                    end_timestamp
+                )));
+                return Err(ProcessingAction::DeleteValueHard);
+            }
+        }
```
jjbayer: What if we put this logic in `process_timestamp` instead of `process_span`? That should automatically cover all timestamps, plus it would attach the meta error to the correct field.

iker-barriocanal: That will deal with only the timestamp itself, but we want to take action on the container of the timestamp. For spans, this means dropping the span; for transactions, this means dropping the entire event. We can't have this level of control in `process_timestamp` unfortunately.

iker-barriocanal: I explored introducing a new type of error to remove the parent container. This approach can work for calls on top-level containers like events, but it requires major refactors and decreases development experience a lot for child components like distributions. For that reason, I'm going to stick to the current approach.

relay-event-normalization/src/timestamp.rs
```diff
@@ -61,6 +91,17 @@ mod tests {
...
+            error
+                .value()
+                .unwrap()
+                .timestamp
+                .value()
+                .unwrap()
+                .into_inner()
+                .timestamp(),
+            1
+        );
```
jjbayer: Wanna use `get_value!` here?

relay-event-normalization/src/timestamp.rs
```diff
@@ -39,15 +39,45 @@ impl Processor for TimestampProcessor {
...
+    fn process_span(
+        &mut self,
+        span: &mut Span,
+        meta: &mut Meta,
+        _: &ProcessingState<'_>,
+    ) -> ProcessingResult {
+        if let Some(start_timestamp) = span.start_timestamp.value() {
+            if start_timestamp.into_inner().timestamp_millis() < 0 {
+                meta.add_error(Error::invalid(format!(
+                    "timestamp is too stale: {}",
```
jjbayer:
```suggestion
                    "start_timestamp is too stale: {}",
```

jjbayer (approved):

PR #2628, opened 2023-10-18
feat(globalconfig): Only report upstream errors after elapsed interval

### LINKED ISSUE #2609: Improve global config fetching error logging
Currently, the global config service emits an error on every failed attempt to request global configs. It's common that relays get 503s when upstream is being deployed, and they can also happen when the network is flaky. See [example error](https://sentry.my.sentry.io/organizations/sentry/issues/399977/). This noise doesn't allow to identify when the error should get attention.

We should not log errors for every request attempt. Instead, log an error when a number of attempts have failed, or until some time has elapsed without fetching the config (e.g. 30s, 60s).

iker-barriocanal:
Especially during deployments, Relay sometimes faces some 503s or timeouts from upstream, and reports an error. Having a handful of such errors is noisy and doesn't help to identify when problems really exist.

This change introduces an interval the global config service has to wait before reporting such errors, while consistently failing fetches.

Note that the interval does not impact other types of errors, like failure to send requests to upstream (e.g. network is down) or global config missing from the response. These errors are consistent and should be reported as early as possible.

Resolves https://github.com/getsentry/relay/issues/2609.

### DIFF
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -11,6 +11,7 @@
 
 use std::borrow::Cow;
 use std::sync::Arc;
+use std::time::Duration;
 
 use relay_config::Config;
 use relay_config::RelayMode;
@@ -20,6 +21,7 @@ use relay_system::{Addr, AsyncResponse, Controller, FromMessage, Interface, Serv
 use reqwest::Method;
 use serde::{Deserialize, Serialize};
 use tokio::sync::{mpsc, watch};
+use tokio::time::Instant;
 
 use crate::actors::upstream::{
     RequestPriority, SendQuery, UpstreamQuery, UpstreamRelay, UpstreamRequestError,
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/global_config.rs
```diff
@@ -164,6 +170,8 @@ impl GlobalConfigService {
             internal_rx,
             upstream,
             fetch_handle: SleepHandle::idle(),
+            last_fetched: Instant::now(),
+            upstream_failure_interval: Duration::from_secs(35),
```
iker-barriocanal, line 174: 35 is arbitrary, but should have time for 3 requests (after 10s each).

jjbayer: Should this be in `Config`?

jjbayer (approved):
I would make the interval configurable, apart from that LGTM!

relay-server/src/actors/global_config.rs
iker-barriocanal: We may not need it for now, but I'll keep this in mind to address it in the future.

PR #2629, opened 2023-10-18
feat(profiling): Add `platform` field to profile frames

krystofwoldrich:
- Related to RN Mixed Profiles (includes Hermes JS Frames and Native iOS frames, JVM frames in the future)
- https://github.com/getsentry/sentry-react-native/pull/3349

- At the moment we don't have logic to automatically recognize the frame type.

- We can add a platform field based on which we can correctly symbolicate the frame

#skip-changelog

### DIFF
--- a/relay-profiling/src/sample.rs
+++ b/relay-profiling/src/sample.rs
@@ -32,6 +32,8 @@ struct Frame {
     lineno: Option<u32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     module: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    platform: Option<String>,
 }
 
 impl Frame {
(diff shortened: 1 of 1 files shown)


phacops (approved):

viglia (approved):

relay-profiling/src/sample.rs
```diff
@@ -14,7 +14,7 @@ use crate::MAX_PROFILE_DURATION;
 
 const MAX_PROFILE_DURATION_NS: u64 = MAX_PROFILE_DURATION.as_nanos() as u64;
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
```
olksdr:
```suggestion
#[derive(Clone, Debug, Default, Serialize, Deserialize)]
```

olksdr (approved):

PR #2632, opened 2023-10-18
feat(perf): Add dynamic calculation of perf score

k-fish:
### Summary
This adds a performance score computation into 'measurements' using a project config. 

The score will be used internally to evaluate performance of frontend pages (matching the current [UI only calculation ](https://github.com/getsentry/sentry/blob/03d1596bc5b4e274f6e65d860a63cc7cf967e79b/static/app/views/performance/browser/webVitals/utils/calculatePerformanceScore.tsx#L97) we're doing), it uses a log-normal cdf function to normalize score for each individual web vital into it's own measurement as well as an aggregate score. Weights are stored in case multiple profiles from different devices (Desktop vs. Mobile-web etc.) have differing weight values.

### Other
- Using `RuleCondition` similar to extracted metrics in lieu of re-creating conditions all over again.
- Uses project config, related PR [here](https://github.com/getsentry/sentry/pull/58418)
- More information about the utility and purpose of this score [here](https://www.notion.so/Browser-Starfish-Performance-Score-30bbed2e0aa540a9a755e2bc23a124fb)

### Screenshots
- Confirmed w/ Sentry PR this is working on local 
Screenshot 2023-10-18 at 6 22 46 PM


relay-event-normalization/src/normalize/mod.rs
```diff
@@ -479,6 +480,71 @@ fn normalize_measurements(
...
+/// Computes performance score measurements
+///
+/// This computes score from vital measurements, using config options to define how it is
+/// calculated.
+fn normalize_performance_score(
+    event: &mut Event,
+    performance_score_config: Option<&PerformanceScoreConfig>,
+) {
+    if let Some(performance_score_config) = performance_score_config {
+        for profile in &performance_score_config.profiles {
```
TBS1996: perhaps this could be a method on `PerformanceScoreProfile`? 

example: https://gist.github.com/TBS1996/f3aa3d1125e68e6cc8d51744eba489c7

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -990,6 +1060,49 @@ impl<'a> DynamicMeasurementsConfig<'a> {
     }
 }
 
+/// Defines a weighted component for a performance score
+///
+/// Weight is the % of score it can take up (eg. LCP is a max of 35% weight for desktops)
+/// Currently also contains (p10, p50) which are used for log CDF normalization of the weight score
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct PerformanceScoreWeightedComponent {
```
TBS1996, line 1097: be careful about using 'pub' too much, it seems here most of your data can be internal but it's marked as pub

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -479,6 +480,71 @@ fn normalize_measurements(
...
+
+                    measurements.insert(
+                        "measurements.score.total".to_owned(),
+                        Measurement {
+                            value: score_total.into(),
+                            unit: (MetricUnit::Fraction(FractionUnit::Ratio)).into(),
+                        }
+                        .into(),
+                    );
+                }
```
TBS1996: ~~this could also be a method, ex: https://gist.github.com/TBS1996/0ce7892ed6b3d14cb566d2a2bc382cf7~~

on second thought, maybe not a good idea since the primary function would be a side effect

relay-dynamic-config/src/project.rs
```diff
@@ -54,6 +55,9 @@ pub struct ProjectConfig {
     /// Configuration for operation breakdown. Will be emitted only if present.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub breakdowns_v2: Option<BreakdownsConfig>,
+    /// Configuration for performance score calculations. Will be emitted only if present.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub performance_score_config: Option<PerformanceScoreConfig>,
```
jjbayer: It seems that the performance score computation only requires data from `event.measurements`. So should we move this key to `MeasurementsConfig`?

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -1078,6 +1191,7 @@ pub fn light_normalize_event(
         normalize_exceptions(event)?; // Browser extension filters look at the stacktrace
         normalize_user_agent(event, config.normalize_user_agent); // Legacy browsers filter
         normalize_measurements(event, config.measurements, config.max_name_and_unit_len); // Measurements are part of the metric extraction
+        normalize_performance_score(event, config.performance_score_config);
```
jjbayer: If we move `performance_score_config` into `MeasurementsConfig`, this function could be a subroutine of `normalize_measurements`, where we already have logic for computing derived measurements (e.g. `frames_frozen_rate`).

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -2853,6 +2967,220 @@ mod tests {
...
+                            "p10": 100,
+                            "p50": 300
+                        },
+                        {
+                            "measurement": "cls",
+                            "weight": 0.25,
+                            "p10": 0.1,
+                            "p50": 0.25
+                        },
+                    ],
```
jjbayer: How likely is it that the formula for the score will every involve other measurements or even input from somewhere else in the event, or that it will have other parameters than weight, p10, and p50?

If likely, we could use this as an opportunity to build a fully generic computation config, that can take and combine values from anywhere in the event (similar to our [metrics extraction config](https://github.com/getsentry/relay/blob/b9e9f9377d7e9a4e477a51335e030a46613c2844/relay-dynamic-config/src/metrics.rs#L216)), so that even the formula for the `erf` function would be provided by config.

If unlikely, we can probably simplify this to 

```json
{"fcp": {"weight": 0.15, "p10": 900, "p50": 1600}, "lcp": ...}
```

relay-event-normalization/src/normalize/mod.rs
jjbayer: @TBS1996 all our normalization logic lives in this crate, so I would stick to that structure here.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -479,6 +480,71 @@ fn normalize_measurements(
...
 
+/// Computes performance score measurements
+///
+/// This computes score from vital measurements, using config options to define how it is
+/// calculated.
+fn normalize_performance_score(
+    event: &mut Event,
+    performance_score_config: Option<&PerformanceScoreConfig>,
+) {
+    if let Some(performance_score_config) = performance_score_config {
```
jjbayer: nit: Might make sense to check this outside of the function, or refactor to `if let Some(performance_score_config) = performance_score_config else { return }`

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -479,6 +480,71 @@ fn normalize_measurements(
     }
 }
 
+/// Computes performance score measurements
```
jjbayer:
```suggestion
/// Computes performance score measurements.
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -479,6 +480,71 @@ fn normalize_measurements(
...
+fn normalize_performance_score(
+    event: &mut Event,
+    performance_score_config: Option<&PerformanceScoreConfig>,
+) {
+    if let Some(performance_score_config) = performance_score_config {
+        for profile in &performance_score_config.profiles {
+            if let Some(condition) = &profile.condition {
+                if !condition.matches(event) {
+                    continue;
+                }
```
jjbayer, line 532: Should we `break` at the end of this loop iteration so we don't accidentally apply two profiles (first matching profile wins)?

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -479,6 +480,71 @@ fn normalize_measurements(
...
+                if let Annotated(Some(ref mut measurements), _) = event.measurements {
```
jjbayer:
```suggestion
                if let Some(measurements) = event.measurements.value_mut() {
```

relay-dynamic-config/src/project.rs
k-fish: We _may_ want to try scores relating to event.duration in the future, not entirely certain but nesting it inside measurements config therefore didn't make a ton of sense.

relay-event-normalization/src/normalize/mod.rs
k-fish: True, see above, I'm open to it if you feel strongly about moving the config inside the MeasurementsConfig

relay-event-normalization/src/normalize/mod.rs
k-fish: fwiw, this is still semi experimental since we aren't using it in any GA'd product (yet).

> involve other measurements or even input from somewhere else in the event, or that it will have other parameters than weight, p10, and p50?

Very possible, hence why I didn't hardcode it. We've been debating including TTFB for example, that may change and this allows us that flexibility. Input from elsewhere in the event _may_ take into account duration etc. 

>  build a fully generic computation config

Yeah I think at the point in time we know what we want for the next iteration we can make it generic then? I was thinking to avoid anything generic until we've confirm a couple cases.

relay-event-normalization/src/normalize/mod.rs
k-fish: oh yeah we should, missed that, thanks 👍

relay-event-normalization/src/normalize/utils.rs
```diff
@@ -153,3 +155,29 @@ pub fn extract_transaction_op(trace_context: &TraceContext) -> Option<String> {
     }
     Some(op.to_string())
 }
+
+/// Error function
+/// <https://en.wikipedia.org/wiki/Error_function>
```
jjbayer:
```suggestion
/// The Gauss error function.
/// 
/// See <https://en.wikipedia.org/wiki/Error_function>.
```

relay-event-normalization/src/normalize/utils.rs
```diff
@@ -153,3 +155,29 @@ pub fn extract_transaction_op(trace_context: &TraceContext) -> Option<String> {
...
+    let x = x.abs();
+    // A&S formula 7.1.26
+    let t = 1.0 / (1.0 + p * x);
+    let y = 1.0 - ((((a5 * t + a4) * t + a3) * t + a2) * t + a1) * t * (-x * x).exp();
+    sign * y
+}
+
+/// Calculates a log-normal CDF score based on a log-normal with a specific p10 and p50
+pub fn calculate_cdf_score(value: f64, p10: f64, p50: f64) -> f64 {
+    let sigma = |p10: f64, p50: f64| (p10.ln() - p50.ln()).abs() / (SQRT_2 * 0.9061938024368232);
```
jjbayer: Let's make this a `fn` instead of a closure.

jjbayer: nit: `SQRT_2 * 0.9061938024368232` could be a precomputed constant.

relay-event-normalization/src/normalize/utils.rs
```diff
@@ -153,3 +155,29 @@ pub fn extract_transaction_op(trace_context: &TraceContext) -> Option<String> {
...
+
+    0.5 * (1.0 - erf((f64::ln(value) - f64::ln(p50)) / (f64::sqrt(2.0) * sigma(p50, p10))))
```
jjbayer:
```suggestion
    0.5 * (1.0 - erf((f64::ln(value) - f64::ln(p50)) / SQRT_2 * sigma(p50, p10))))
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -990,6 +1061,49 @@ impl<'a> DynamicMeasurementsConfig<'a> {
...
+    /// Median used to define the log-normal for calculation
+    pub p50: f64,
+}
+
+/// Defines a profile for performance score
+///
+/// A profile contains weights for a score of 100% and match against an event using a condition.
+/// eg. Desktop vs. Mobile(web) profiles for better web vital score calculation.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct PerformanceScoreProfile {
```
jjbayer, line 1115: `score_components` -> `scoreComponents`.

```suggestion
#[serde(rename_all = "camelCase")]
pub struct PerformanceScoreProfile {
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -479,6 +480,72 @@ fn normalize_measurements(
...
+                let mut score_total = 0.0;
+                for component in &profile.score_components {
+                    if let Some(value) = measurements.get_value(component.measurement.as_str()) {
+                        let subscore =
+                            utils::calculate_cdf_score(value, component.p10, component.p50)
+                                * component.weight;
+                        score_total += subscore;
+
+                        measurements.insert(
+                            format!("measurements.score.{}", component.measurement),
```
jjbayer: Do we need the `measurements.` prefix?

Also, we have a precedent where we group similar measurements by prefix:

https://github.com/getsentry/relay/blob/7e44fb03b58723cac120e1de1c7c6cecbea4532b/tests/integration/fixtures/mini_sentry.py#L435-L436

Should we do the same here? That is, `fid.score` instead of `measurements.score.fid`?

relay-dynamic-config/src/project.rs
jjbayer: OK, let's leave as is then.

relay-event-normalization/src/normalize/mod.rs
jjbayer: Makes sense, let's leave the format as-is 👍 .

relay-dynamic-config/src/project.rs
jjbayer: We also need to add this field to `LimitedProjectConfig`, otherwise edge relays won't extract metrics for the new measurements.

jjbayer (changes requested):
Thanks for contributing! This already looks very good to me, pressing "Request changes" because of `LimitedProjectConfig` and some open questions.

relay-event-normalization/src/normalize/mod.rs
k-fish: Yeah `measurements` should be removed 👍 

We _should_ keep `measurements.score` though, I re-arranged it so it's simpler to hide it in UI etc. In the case where scores aren't against measurements in the future too I think this naming makes a bit more sense.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -990,6 +1065,50 @@ impl<'a> DynamicMeasurementsConfig<'a> {
     }
 }
 
+/// Defines a weighted component for a performance score
```
jjbayer: nit
```suggestion
/// Defines a weighted component for a performance score.
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -990,6 +1065,50 @@ impl<'a> DynamicMeasurementsConfig<'a> {
...
+    pub measurement: String,
+    /// Weight [0,1.0] of this component in the performance score
+    pub weight: f64,
+    /// p10 used to define the log-normal for calculation
+    pub p10: f64,
+    /// Median used to define the log-normal for calculation
+    pub p50: f64,
+}
+
+/// Defines a profile for performance score
```
jjbayer:
```suggestion
/// Defines a profile for performance score.
```

relay-event-normalization/src/normalize/utils.rs
```diff
@@ -153,3 +155,33 @@ pub fn extract_transaction_op(trace_context: &TraceContext) -> Option<String> {
...
+    // Save the sign of x
+    let sign = if x < 0.0 { -1.0 } else { 1.0 };
+    let x = x.abs();
+    // A&S formula 7.1.26
+    let t = 1.0 / (1.0 + p * x);
+    let y = 1.0 - ((((a5 * t + a4) * t + a3) * t + a2) * t + a1) * t * (-x * x).exp();
+    sign * y
+}
+
+/// Sigma function for CDF score calculation
```
jjbayer:
```suggestion
/// Sigma function for CDF score calculation.
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -2853,6 +2973,220 @@ mod tests {
...
+                        "op":"eq",
+                        "name": "event.contexts.browser.name",
+                        "value": "Chrome"
+                    }
+                }
+            ]
+        }))
+        .unwrap();
+
+        normalize_measurements(&mut event, None, None);
```
jjbayer: nit: Could we use a test fixture with measurements already normalized instead? We have a new guideline on this:

> Do not call into other production code to fulfill preconditions. For example, if your subject assumes normalized data, do not call the normalization function, but provide hard-coded normalized data.

([internal doc](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#d8c37053239e46fbb319a06428fac18d))

relay-dynamic-config/src/project.rs
jjbayer: Just noticed that most keys in this config do not have a `_config` suffix, (e.g. `measurements: Option<MeasurementsConfig>`), should we remove it here as well?

jjbayer (approved):
Knowing that the new config is feature-flagged on sentry side, this PR looks good to go!

See comments for nitpicks.

relay-event-normalization/src/normalize/mod.rs
k-fish: Sure that's fine, they're already pretty close to normalized anyway, at least for the vitals that come in as-is.

relay-dynamic-config/src/project.rs
k-fish: Yep, sure

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -479,6 +480,76 @@ fn normalize_measurements(
...
+                            .into(),
+                        );
+                        measurements.insert(
+                            format!("score.weight.{}", component.measurement),
+                            Measurement {
+                                value: component.weight.into(),
+                                unit: (MetricUnit::Fraction(FractionUnit::Ratio)).into(),
+                            }
+                            .into(),
+                        );
```
iker-barriocanal, line 557: Do weights depend on the event? It seems these measurements are the same values as in the project config:
https://github.com/getsentry/sentry/pull/58418/files#diff-5e30755537fcfd5687d4399272065e49affcc12a4cf7a0bc377cff5234b8a08aR418-R421

iker-barriocanal (commented):
Fantastic PR, thanks for contributing!

relay-event-normalization/src/normalize/mod.rs
k-fish: The weights can depend on the profile used to calculate the score, which depend on device type, and potentially browser etc. as well. To be able to graph the 'missed performance' each score needs to be compared to the aggregate weight. It's fine if we only have a single calculation, but additional calculations begin to make the weight logic complicated.

k-fish: eg. one event may have an FCP score of say 10 of 20, and another could use a different profile and have a score of 20 of 40. If you were to average scores it'd be 15, if you naively use weights it'd seem your score is 15 of 20 = 75%, but in reality it's 50% since those event have different evaluations.

jjbayer: @k-fish I think @iker-barriocanal is on to something here: It might make sense to materialize all the components of the score computation back into the event, but putting the weights in `measurements` means that we will also extract expensive distribution metrics for them (see [code here](https://github.com/getsentry/relay/blob/42357e36358c1158b4d16664a1cc2c18de6b63ff/relay-server/src/metrics_extraction/transactions/mod.rs#L265-L293)), which I cannot imagine to be useful. Should we write the intermediate components of the computation into `event.data` instead of `event.measurements`?

jjbayer: cc @edwardgou-sentry

iker-barriocanal: Depending how useful these metrics are for users to look directly at, we could also add them to the event's context.

edwardgou-sentry: Hey @jjbayer and @iker-barriocanal, I don't think we'll need distributions for the weights. If we move the weights to `event.data`, will it be possible to query scores by weight tag value? I think that's mostly what we will be using these weights for, @k-fish please let me know if im wrong.

iker-barriocanal: Not sure I understand the question, it doesn't seem we're using tags here.

Am I correct to assume that the relevant data is the ratio? In the examples above, we're not interested in the 10 or 20, or 20 or 40; only that they have a 50% score. In that case, Relay could send a `0.5` representing the 50% instead of the raw values and weights. I also assume the weights won't change often.

jjbayer: > will it be possible to query scores by weight tag value?

Not claiming I understand the use case 100%, but if we want to group by / filter to a specific value of the weight, would it make sense to add a metrics tag `"profile"` so we can query for example the avg. score for a specific profile?

edwardgou-sentry: > Not claiming I understand the use case 100%, but if we want to group by / filter to a specific value of the weight, would it make sense to add a metrics tag "profile" so we can query for example the avg. score for a specific profile?

Yeah I think this would work. Sorry I wasn't really clear about this 😅 . I think we just need a way for us to know what "weight profile" a performance score distribution belongs to so that this can be reflected in the product UI or adjust in some way in the product.

> In that case, Relay could send a 0.5 representing the 50% instead of the raw values and weights.

I haven't thought too much about this, but I _think_ this would work as well, but we would need to do a lot of the calculation back to the score values and weights in post for display in the frontend. I don't expect these weights to change often.

I'm not sure which solution is more efficient here storage wise? But I think I would prefer having some tag(s) on the weights/profiles of the distributions since it might be more flexible and easier to work with.

k-fish: Sorry, back now. Catching up, I think you realized that you _at the minimum_ have to differentiate the profile with a tag to be able to calculate the overall room for improvement properly, since weights become blended with >1 profile, no matter what. This has two downsides though:
- No historical data upon changes of profile % if we do, though this will be rare
- More complex calculation tracking the distribution of scores.

Also remember that adding the tag increases the storage regardless since it's another row _for each measurement_ with the same distributions, they are just different pieces of data. As your number of profiles grow the actual stored data exceeds the weights I proposed (I believe). The simplest solution is to just store the used weight at that point in time. Since this is still in the experimental phase, imo it's fine to try whichever method since mostly we'll find out quickly if something doesn't make sense once org_id=1 starts using these.

jjbayer: > Since this is still in the experimental phase, imo it's fine to try whichever method since mostly we'll find out quickly if something doesn't make sense once org_id=1 starts using these.

I'm fine with shipping this as-is, but let's have another discussion after the dogfooding phase please.

jjbayer:
@edwardgou-sentry before enabling this, could you add the new measurement metrics to the static indexer string lookup? It will prevent creating new db entries for every organization. See for example https://github.com/getsentry/sentry/blob/42000491f5f28a2a7bae2027f918872f74a33109/src/sentry/sentry_metrics/indexer/strings.py#L71

edwardgou-sentry:
@jjbayer Done! https://github.com/getsentry/sentry/pull/58789

PR #2634, opened 2023-10-19
ref(protocol): Clean up imports

jjbayer:
Follow up to https://github.com/getsentry/relay/pull/2608:

* Remove `utils` crate from `relay-protocol`.
* Only re-export `RuleCondition` from the `condition` module.

I was not able to remove the `relay-sampling` dependency anywhere yet, because `relay-cabi` and `relay-dynamic-config` import `SamplingConfig`. But we will be able to change the dependency in https://github.com/getsentry/relay/pull/2595.

#skip-changelog

### DIFF
--- a/relay-protocol/src/lib.rs
+++ b/relay-protocol/src/lib.rs
@@ -23,17 +23,15 @@ mod macros;
 mod meta;
 mod size;
 mod traits;
-mod utils;
 mod value;
 
 pub use self::annotated::*;
-pub use self::condition::*;
+pub use self::condition::RuleCondition;
 pub use self::impls::*;
 pub use self::macros::*;
 pub use self::meta::*;
 pub use self::size::*;
 pub use self::traits::*;
-pub use self::utils::*;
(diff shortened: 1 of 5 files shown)


olksdr (approved):

iambriccardo (commented):
LGTM, but I will leave it to @jan-auer

iker-barriocanal (approved):
Thanks for doing this!

relay-protocol/src/lib.rs
```diff
@@ -23,17 +23,15 @@ mod macros;
 mod meta;
 mod size;
 mod traits;
-mod utils;
 mod value;
 
 pub use self::annotated::*;
-pub use self::condition::*;
+pub use self::condition::RuleCondition;
```
jan-auer, line 30: In addition to this, we will need to update to `pub mod condition`, otherwise the variants of `RuleCondition` reference private types.

jan-auer (approved):

PR #2636, opened 2023-10-19
ref(spool): Clearer error message on unsupported paths

iker-barriocanal:
By default, the spool file is created if it doesn't exist, but it's assumed the directory to exist, failing if that assumption is not valid. This PR attempts to create the parent directories recursively if they don't exist so that the spool file can successfully be created. This is documented in https://github.com/getsentry/sentry-docs/pull/8306.

#skip-changelog


relay-server/src/actors/spooler/mod.rs
```diff
@@ -662,6 +666,26 @@ pub struct BufferService {
...
+                    .await
+                    .map_err(|e| BufferError::BufferSetupFailed(e.to_string()))?;
+            }
+        } else {
+            // sqlx doesn't support relative paths and Sqlite initialization
+            // fails. Maybe: https://github.com/launchbadge/sqlx/issues/1260
+            return Err(BufferError::BufferSetupFailed(
+                "relative paths for spooling are not supported".to_string(),
+            ));
+        }
```
olksdr: how was this tested? 
I've ran relay locally and the relative paths working just fine. 

Could you, please, post your test or config, so I could run it as well?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -79,7 +80,10 @@ pub enum BufferError {
     FileSizeReadFailed(sqlx::Error),
 
     #[error("failed to setup the database: {0}")]
-    SetupFailed(sqlx::Error),
+    SqlxSetupFailed(sqlx::Error),
+
+    #[error("failed to setup the spool buffer: {0}")]
+    BufferSetupFailed(String),
```
olksdr: These 2 errors seems like the same. 

The `BufferSetupFailed` is basically about the file creation?  So maybe we call `BufferFileError` ?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -662,6 +666,26 @@ pub struct BufferService {
...
+            let mut dir = path.clone();
+            dir.pop();
+
+            if !dir.exists() {
+                relay_log::debug!("creating directory for spooling file: {}", dir.display());
+                DirBuilder::new()
+                    .recursive(true)
+                    .create(dir)
+                    .await
+                    .map_err(|e| BufferError::BufferSetupFailed(e.to_string()))?;
```
olksdr: If we just document this behaviour better , would this be also ok? 

But also if that really simplifies the usage , I'm also ok if the relay will create the folders automatically.

olksdr (changes requested):

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: I updated the description to include the spool path. Happy to add a test if it provides more clarity.

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: `BufferSetupFailed` only happens during the setup process of the buffer service, and `SqlxSetupFailed` can happen at any time sqlx is doing initialization work (including in the public function `create`). Since their purpose is different I think having two different errors makes sense, but I'm not strongly opinionated and can unify them if you think that's an overkill.

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: Documenting it would work, but I think creating the directories directly provides some user experience improvements that outweigh that additional complexity.

relay-server/src/actors/spooler/mod.rs
olksdr: It's fine to have another error type in needed but I would do here few things:
* create the error with more descriptive name, exactly what happens
* and also the error message should be clearer with the proper error , the reason why it happens

```suggestion
    SqlxSetupFailed(sqlx::Error),

    #[error("failed to create the spool buffer file : {0}")]
    FileError(std::io::Error),
```

relay-server/src/actors/spooler/mod.rs
olksdr: We will still have to document this behaviour as well, to make sure it's clear for user that all the folders and sub-folders will be created automatically.

relay-server/src/actors/spooler/mod.rs
olksdr: `~` is a shell thingy and won't be expanded to user's home directory, and will be treated as the literal. 

As I posted before the relative paths working just fine. And we do not need any extra check for those.

iker-barriocanal: You're right, removed this piece of code. I clarified the behavior of `~` in the docs PR in the description.

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: Documented in the docs PR in the description.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -79,7 +80,10 @@ pub enum BufferError {
     FileSizeReadFailed(sqlx::Error),
 
     #[error("failed to setup the database: {0}")]
-    SetupFailed(sqlx::Error),
+    SqlxSetupFailed(sqlx::Error),
+
+    #[error("failed to create the spool buffer: {0}")]
```
olksdr:
```suggestion
    #[error("failed to create the spool file: {0}")]
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -661,7 +665,12 @@ pub struct BufferService {
 
 impl BufferService {
     /// Set up the database and return the current number of envelopes.
+    ///
+    /// The directories and spool file will be created if they don't already
+    /// exist.
     async fn setup(path: &PathBuf) -> Result<(), BufferError> {
+        BufferService::create_spool_directory(path.clone()).await?;
```
olksdr: We do not need to clone the path here, just use the reference to it as the incoming `path` parameter. 

```suggestion
        BufferService::create_spool_directory(path).await?;
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -670,12 +679,26 @@ impl BufferService {
...
+            .map_err(BufferError::SqlxSetupFailed)?;
 
         sqlx::migrate!("../migrations").run(&db).await?;
         Ok(())
     }
 
+    /// Creates the directories for the spool file.
+    async fn create_spool_directory(mut path: PathBuf) -> Result<(), BufferError> {
+        path.pop();
+        if !path.exists() {
```
olksdr: We could accept the reference to the path, and just take the [`parent`](https://doc.rust-lang.org/std/path/struct.Path.html#method.parent), since that's what you're doing with `pop` here

Maybe something like this: 

```suggestion
    async fn create_spool_directory(path: &PathBuf) -> Result<(), BufferError> {
        let Some(path) = path.parent() else { Ok(()) };
        if !path.as_os_str().is_empty() && !path.exists() {
    
```

relay-server/src/actors/spooler/mod.rs
```diff
@@ -977,6 +1000,28 @@ mod tests {
...
+                "envelopes": {
+                    "path": target_dir,
+                }
+            }
+        }))
+        .unwrap()
+        .into();
+        BufferService::create(buffer_guard, services(), config)
+            .await
+            .unwrap();
```
olksdr, line 1025: Could we actually also add assert there to check that the folders and the path to the file exist?

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: Also updated `setup`'s arg to use `Path` as it's a [slice of `PathBuf`](https://doc.rust-lang.org/stable/std/path/struct.Path.html).

relay-server/src/actors/spooler/mod.rs
```diff
@@ -670,12 +679,28 @@ impl BufferService {
...
         sqlx::migrate!("../migrations").run(&db).await?;
         Ok(())
     }
 
+    /// Creates the directories for the spool file.
+    async fn create_spool_directory(path: &Path) -> Result<(), BufferError> {
+        let Some(parent) = path.parent() else {
+            return Ok(());
+        };
+        if !parent.exists() {
```
olksdr: it can happen, that the path is just a file `spool.db`, which means that this parent will be `Some("")` (empty path) (this also would happen with `.pop` you used in previous version). Meaning that the file is relative to the binary root, which we do not have to create. In the suggestion I also put the check if the path is just an empty string, just skip the creation.

olksdr (approved):
Left one comment which I think should be addressed before merging this. Otherwise this looks good to me! 👍

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: I applied the change and added a test, but it shouldn't mater as a `""` directory doesn't exist.

PR #2638, opened 2023-10-19
fix(starfish): check for `decoded_content_length`

DominikB2014:
There was an error in our span data convention doc, our SDK actually collects `http.decoded_response_content_length`, see  [this line of code](https://github.com/getsentry/sentry-javascript/blob/80ee8f8f64b2d76a5cf1c6f7e98b6a07d8ab5af6/packages/tracing-internal/src/browser/metrics/index.ts#L411) and not `http.decoded_response_body_length`, this error was corrected in the following PR https://github.com/getsentry/develop/pull/1059.

This PR ensures we collect the former, ~~but still collect the later for backwards compatibility~~. I have not worked with relay before, so please lmk if there's any other changes necessary!


relay-dynamic-config/src/defaults.rs
```diff
@@ -106,6 +106,17 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
             ),
             tags: Default::default(),
```
jjbayer: If no SDK is sending data with the wrong key, we can safely remove the old `MetricSpec`.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -360,6 +360,18 @@ pub(crate) fn extract_tags(span: &Span, config: &Config) -> BTreeMap<SpanTagKey,
                 );
             }
```
jjbayer: Same here: If `span.data` never actually contains the key `http.decoded_response_body_length`, you can remove the old if block.

relay-dynamic-config/src/defaults.rs
```diff
@@ -106,6 +106,17 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "d:spans/http.decoded_response_body_length@byte".into(),
```
jjbayer: @DominikB2014 do you still want to use the old MRI here, or rather make a clean cut and update the MRI as well? If I understand correctly, we did not collect any data for the old MRI yet.

jjbayer (commented):
Given that we have not collected any data under the old key, should we just fix the typo everywhere (MRI name, `sentry_tags`, [strings.py](https://github.com/getsentry/sentry/blob/37ed8788a1790ff5ec44a12243bfcf8390202f94/src/sentry/sentry_metrics/indexer/strings.py#L186), queries)?

If we don't, we'll have to live with the typo in our internal naming forever, right?

DominikB2014:
> Given that we have not collected any data under the old key, should we just fix the typo everywhere (MRI name, `sentry_tags`, [strings.py](https://github.com/getsentry/sentry/blob/37ed8788a1790ff5ec44a12243bfcf8390202f94/src/sentry/sentry_metrics/indexer/strings.py#L186), queries)?
> 
> 
> 
> If we don't, we'll have to live with the typo in our internal naming forever, right?

@jjbayer if that's possible that's what I'd prefer. Is there any issues with that?

relay-dynamic-config/src/defaults.rs
DominikB2014: We shouldn't have collected data unless someone manually instrumented spans using the old span conventions

DominikB2014: I'd prefer updating the MRI in that case

jjbayer (approved):
Let's do it!

Please merge & deploy [the sentry PR](https://github.com/getsentry/sentry/pull/58522) before merging this one.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Restrict resource spans to script and css only. ([#2623](https://github.com/getsentry/relay/pull/2623))
 - Postpone metrics aggregation until we received the project state. ([#2588](https://github.com/getsentry/relay/pull/2588))
 - Scrub random strings in resource span descriptions. ([#2614](https://github.com/getsentry/relay/pull/2614))
+- Collect `http.decoded_response_content_length`. ([#2638](https://github.com/getsentry/relay/pull/2638))
```
olksdr: Please, merge `master` in and update the CHANGELOG moving this record in `Unreleased` section again.

DominikB2014: @olksdr made the change!

PR #2639, opened 2023-10-19
feat(gocd): Deploy Relay automatically

phacops:
This activates auto deploy with GoCD.

#skip-changelog

### DIFF
--- a/gocd/templates/relay.jsonnet
+++ b/gocd/templates/relay.jsonnet
@@ -6,7 +6,6 @@ local pipedream = import 'github.com/getsentry/gocd-jsonnet/libs/pipedream.libso
 
 local pipedream_config = {
   name: 'relay',
-  auto_deploy: true,
   materials: {
     relay_repo: {
       git: 'git@github.com:getsentry/relay.git',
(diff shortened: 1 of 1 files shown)


gocd/templates/relay.jsonnet
```diff
@@ -15,6 +14,11 @@ local pipedream_config = {
       destination: 'relay',
     },
   },
+  rollback: {
+    material_name: 'relay_repo',
+    stage: 'deploy-production',
```
phacops, line 19: @mattgauntseo-sentry with targeting this stage, it will also rollback `deploy-pops` which is the next stage, right?

iker-barriocanal:
The team had a short discussion today and we agreed to not deploy Relay automatically for now.

- For risky PRs, we'd like someone from Ingest to be around in case something goes wrong.
- Non-risky PRs could be deployed automatically, but that introduces more complexity and edge cases (e.g. ensure previous non-deployed PRs aren't risky).
- This doesn't improve the situation of deploying without monitoring either. Relay is a critical system and should always be monitored while deploying.

For these reasons, I'm going to close the PR but we can get back to this in the future.

PR #2640, opened 2023-10-20
feat(on-demand): Extend support fields in the event getter

### LINKED ISSUE #58504: Adding fields support in Relay

iambriccardo:
This PR adds support for additional fields in the `Event` `Getter` implementation in order to satisfy a wider range of conditions for on-demand metrics.

In addition, it adds the following new fields to the `DeviceContext`:
* locale
* screen_height_pixels
* screen_width_pixels
* uuid

Closes https://github.com/getsentry/sentry/issues/58504


relay-event-schema/src/protocol/contexts/device.rs
```diff
@@ -161,6 +162,22 @@ pub struct DeviceContext {
     /// Whether location support is available on the device.
     pub supports_location_service: Annotated<bool>,
 
+    /// Width of the screen in pixels.
+    #[metastructure(pii = "maybe")]
+    pub screen_width_pixels: Annotated<u64>,
```
iker-barriocanal, line 167: Does this property having pii make sense? Same for the fields below.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Restrict resource spans to script and css only. ([#2623](https://github.com/getsentry/relay/pull/2623))
 - Postpone metrics aggregation until we received the project state. ([#2588](https://github.com/getsentry/relay/pull/2588))
 - Scrub random strings in resource span descriptions. ([#2614](https://github.com/getsentry/relay/pull/2614))
+- Extend the number of supported fields for the `Event` `Getter`. ([#2640](https://github.com/getsentry/relay/pull/2640))
```
iker-barriocanal: This PR extends the event schema. Could we move this to the features section, and also update the py/changelog?

iker-barriocanal (approved):

CHANGELOG.md
iambriccardo: Sure

relay-event-schema/src/protocol/contexts/device.rs
iambriccardo: I thought about this, and I was like, maybe some people want to scrape those fields, since they might not want to share that information with sentry (e.g., a new device with a new resolution is being secretly tested), maybe the UUID is the only thing but also that can be delicate for some users. We can definitely omit it, idk if there is a performance overhead to this.

jan-auer: ~I'd rather not mark them PII. They are not, and none of the other fields are marked as such.~
The screen size is not PII, however since the other screen attributes are marked as PII, we should stay consistent here.

relay-event-schema/src/protocol/event.rs
```diff
@@ -1028,6 +1077,12 @@ mod tests {
                     name: Annotated::new("iphone".to_string()),
                     family: Annotated::new("iphone-fam".to_string()),
                     model: Annotated::new("iphone7,3".to_string()),
+                    screen_dpi: Annotated::new(560),
+                    screen_width_pixels: Annotated::new(1920),
+                    screen_height_pixels: Annotated::new(1080),
+                    locale: Annotated::new("US".into()),
```
jan-auer, line 1083: A correct value for this field would be `en-US` or `de-AT`.

relay-event-schema/src/protocol/contexts/device.rs
```diff
@@ -161,6 +162,22 @@ pub struct DeviceContext {
...
+
+    /// Height of the screen in pixels.
+    #[metastructure(pii = "maybe")]
+    pub screen_height_pixels: Annotated<u64>,
+
+    /// Locale of the device.
```
jan-auer, line 173: Let's try to add more information to the doc comments, such as what is the format. I'll pick this up in my follow-up.

jan-auer (commented):
The following fields are missing in this PR:
- `contexts.app.in_foreground`
- `contexts.unreal.crash_type`

I'll rebase #2607 and address these.

PR #2641, opened 2023-10-20
fix(pii): Disable scrubbing for the User-Agent header

### LINKED ISSUE #1826: IP scrubbing removes chrome version
Michael Chai reported that a customer got a false positive where ip scrubbing removed the browser version. Customer issue [here](https://getsentry.atlassian.net/browse/ISSUE-1652).

https://files.slack.com/files-pri/T024ZCV9U-F04NU8URXGR/image.png

Perhaps we should add user agent to the default safe fields?

Dav1dde:
Some User-Agents are being scrubbed by the IP rule, this disables scrubbing for the entire Header as suggested here https://github.com/getsentry/relay/issues/1826#issuecomment-1427631617

Fixes: #1826

### DIFF
--- a/relay-pii/src/processor.rs
+++ b/relay-pii/src/processor.rs
@@ -754,6 +754,35 @@ mod tests {
     }
 
     #[test]
+    fn test_ignore_user_agent_ip_scrubbing() {
+        let mut data = Event::from_value(
+            serde_json::json!({
+                "request": {
+                    "headers": [
+                        ["User-Agent", "127.0.0.1"],
+                        ["X-Client-Ip", "10.0.0.1"]
+                    ]
+                },
+            })
+            .into(),
+        );
+
+        let scrubbing_config = DataScrubbingConfig {
+            scrub_data: true,
+            scrub_ip_addresses: true,
+            scrub_defaults: true,
+            ..Default::default()
+        };
+
+        let pii_config = to_pii_config(&scrubbing_config).unwrap();
+        let mut pii_processor = PiiProcessor::new(pii_config.compiled());
+
+        process_value(&mut data, &mut pii_processor, ProcessingState::root()).unwrap();
+
+        assert_debug_snapshot!(&data);
+    }
+
+    #[test]
     fn test_remove_debugmeta_path() {
         let config = serde_json::from_str::<PiiConfig>(
             r#"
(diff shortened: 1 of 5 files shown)


relay-pii/src/processor.rs
```diff
@@ -753,6 +753,35 @@ mod tests {
         assert_annotated_snapshot!(event);
     }
 
+    #[test]
+    fn test_ignore_user_agent_ip_scrubbing() {
```
olksdr, line 757: Another test I would like to see here, if the user would like to scrub the `User-Agent` explicitly now, if that's still possible?

relay-pii/src/processor.rs
```diff
@@ -753,6 +753,35 @@ mod tests {
...
+            scrub_defaults: true,
+            ..Default::default()
+        };
+
+        let pii_config = to_pii_config(&scrubbing_config).unwrap();
+        let mut pii_processor = PiiProcessor::new(pii_config.compiled());
+
+        process_value(&mut data, &mut pii_processor, ProcessingState::root()).unwrap();
+
+        assert_debug_snapshot!(&data);
```
iker-barriocanal: nit: `assert_annotated_snapshot!` on `event.request` may produce a smaller snapshot.

iker-barriocanal (commented):
Same comment as Olek, other than that LGTM!

relay-pii/src/processor.rs
Dav1dde: It's a much nicer snapshot now just using `assert_annotated_snapshot`.

Using `event.request` is tricky because the event itself is also wrapped in an `Annotated` which means I'd have to do something like `data.0.unwrap().request`.

relay-pii/src/processor.rs
Dav1dde: So I've looked into this ...

The rule which gets generated from the `DataScrubbingConfig` is this:


```json
"($string || $number || $array || $object) && !(<other ignored fields> || $http.headers.user-agent)": [
  "@common:filter",
  "@ip:replace",
   "strip-fields"
],
```

Where `strip-fields` contains the information gathered from `sensitive_fields` (what the customer can enter in the UI). So the `User-Agent` will never match on the sensitive fields rule.

**But:** with advanced data scrubbing rules the User-Agent can still be scrubbed, since this is an entirely separate process from the `DataScrubbingConfig`.

Should I make a test with `sensitive_fields` set to the user agent and asserting it won't be scrubbed? I feel like this is just a side effect and not something we specifically want to enforce.

iker-barriocanal (approved):

PR #2642, opened 2023-10-20
chore(beta): Address lint errors for next stable

olksdr:
Make sure the `beta` compiles so that we have smooth switch once it becomes stable.

Fixes following warning: 
```
warning: `&` without an explicit lifetime name cannot be used here
```

#skip-changelog


jjbayer (approved):

iker-barriocanal (approved):
🙌

PR #2644, opened 2023-10-20
ref(profiling): Log better errors

phacops:
We want to move logging to inside the profiling crate in order to log more helpful information like the platform and the actual key causing the JSON error.

#skip-changelog


jjbayer:
@phacops any chance we could meet your requirements without the additional dependency? We have to review the dependency for both legal reasons and to check if they might cause accidental panics.

phacops:
Unfortunately we can't, the crate reimplements a deserializer keeping track of fields deserialized and we'd need to do the same thing for what we need.

I could be scrappier and get the line + column and just try to get a few chars before to get a sense of what the field name is and remove the crate. Otherwise, I'd have to re-implement a deserializer in the same way.

relay-profiling/src/lib.rs
```diff
@@ -126,26 +128,48 @@ struct MinimalProfile {
...
+    let d = &mut Deserializer::from_slice(payload);
+    serde_path_to_error::deserialize(d).map_err(ProfileError::InvalidJson)
 }
 
 pub fn parse_metadata(payload: &[u8]) -> Result<(), ProfileError> {
     let profile = match minimal_profile_from_json(payload) {
         Ok(profile) => profile,
-        Err(err) => return Err(err),
+        Err(err) => {
+            relay_log::warn!(error = &err as &dyn Error, "invalid minimal profile");
```
jjbayer: nit: Why did you move the warning to `relay-profiling`? Is it to know the difference between an error in minimum profile parsing and an error in sample parsing? For that, we could also use separate `ProfileError::InvalidMinimalJson` and `ProfileError::InvalidSampleJson`. That way, we could remove the `relay-log` dependency again.

relay-profiling/src/lib.rs
```diff
@@ -126,26 +128,48 @@ struct MinimalProfile {
...
+            let _: sample::ProfileMetadata = match serde_path_to_error::deserialize(d) {
                 Ok(profile) => profile,
-                Err(err) => return Err(ProfileError::InvalidJson(err)),
+                Err(err) => {
+                    relay_log::warn!(
+                        error = &err as &dyn Error,
+                        "invalid sample (v{:?}) profile for {}. invalid key: {}",
+                        profile.version,
+                        profile.platform,
+                        err.path(),
```
jjbayer: Have you tried if you actually need to add `err.path()` explicitly here? the `error` passed as first argument should already give you all the error information in the logs.

relay-profiling/Cargo.toml
```diff
@@ -14,9 +14,11 @@ android_trace_log = { version = "0.3.0", features = ["serde"] }
 chrono = { workspace = true }
 data-encoding = "2.3.3"
 relay-event-schema = { path = "../relay-event-schema" }
+relay-log = { path = "../relay-log" }
 relay-protocol = { path = "../relay-protocol" }
 serde = { workspace = true }
 serde_json = { workspace = true }
+serde_path_to_error = "0.1.14"
```
jjbayer, line 21: Code of this library looks OK to me. Given it has the same author & license as serde, I think we can include it.

jjbayer (approved):
No blockers, but please take a look at the comments before merging.

relay-profiling/src/lib.rs
phacops: The reason I did this is to be able to expose the platform and possible version of the profiling schema in the error in the actual parsing of the profile. I don't necessarily care about differentiating minimal vs full parsing error, it's just something I have to do if I want to show the platform.

relay-profiling/src/lib.rs
phacops: It was indeed redundant information as it's already displayed.

PR #2646, opened 2023-10-23
docs(project): Split first line of docstrings

iker-barriocanal:
#skip-changelog

### DIFF
--- a/relay-server/src/actors/project.rs
+++ b/relay-server/src/actors/project.rs
@@ -153,10 +153,10 @@ impl ProjectState {
         self.disabled
     }
 
-    /// Returns whether the project state obtained from upstream could not be parsed.
+    /// Returns `true` if the project state obtained from upstream could not be parsed.
     ///
-    /// This results in events being dropped similar to disabled states, but can
-    /// provide separate metrics.
+    /// This results in events being dropped similar to disabled states, but can provide separate
+    /// metrics.
     pub fn invalid(&self) -> bool {
         self.invalid
     }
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/project.rs
```diff
@@ -739,12 +740,14 @@ impl Project {
...
+    /// This first checks if the state needs to be updated. This is the case if
+    /// the project state has passed its cache timeout. The `no_cache` flag
+    /// forces an update. This does nothing if an update is already running in
+    /// the background.
     ///
-    /// Independent of updating, _stale_ states are passed to the sender immediately as long as they
-    /// are in the [grace period](Config::project_grace_period).
+    /// Independent of updating, _stale_ states are passed to the sender
+    /// immediately as long as they are in the [grace
+    /// period](Config::project_grace_period).
```
olksdr: Why do we do this, what the reasoning behind making the lines shorter?  It's much harder to read now.

iker-barriocanal: Ooops didn't notice. Put the link back together in [31372cf](https://github.com/getsentry/relay/pull/2646/commits/31372cfe23bb8398b6ad99b548ccc80ad08864ea).

relay-server/src/actors/project.rs
```diff
@@ -153,9 +153,10 @@ impl ProjectState {
         self.disabled
     }
 
-    /// Returns `true` if the project state obtained from the upstream could not be parsed. This
-    /// results in events being dropped similar to disabled states, but can provide separate
-    /// metrics.
+    /// Returns whether the project state obtained from upstream could not be parsed.
```
olksdr: I find the old sentence much easier to understand.  And describes exactly what to expect. 

```suggestion
    /// Returns `true` if the project state obtained from upstream could not be parsed.
```

olksdr (commented):
My questions still stands: 
What is the reasoning behind making the doc comments shorter? 

The length they have now is good enough. As for me could be even longer.

iker-barriocanal:
@olksdr I did not notice we had a preference for [line width](https://github.com/getsentry/relay/blob/ecfa6727ed3d76249388367e9a98e2efdf047ce7/.vscode/settings.json#L7) and found 80 cols easier to read. As this is an opinion, I reverted that change. The PR now only includes the line split, as per our [internal guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#987c0416a9f54674999eed59eecbf846).

olksdr (approved):

PR #2647, opened 2023-10-23
chore(spans): Disable resource link span indestion

phacops:
`resource.link` spans are collected when a resource (`script`, `css` or `img`) is preloaded which means they will likely appear as its own `resource.(img|css|script)` span soon after, producing 2 spans of the same data which might not be useful.

As we're chasing cardinality, we'll disable ingesting those spans until we have a better handle on cardinality and we feel they add value to the product.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unrelease
+
+**Internal**:
+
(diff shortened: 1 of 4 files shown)


jjbayer (approved):
Could you add a short motivation to the PR description please? Thanks!

AbhiPrasad (approved):

PR #2652, opened 2023-10-24
feat(server): Consistent rate limiting before aggregation

TBS1996:
part of: #2083

We want to do consistent rate-limiting prior to aggregation. 

currently we send metrics from pre-aggregator to aggregator, and then we rate limit afterwards. Now we will perform rate limiting before aggregation instead. 

previously:

after:


relay-server/src/actors/processor.rs
```diff
@@ -2898,8 +2901,7 @@ impl EnvelopeProcessorService {
         use relay_quotas::ItemScoping;
 
         let RateLimitFlushBuckets {
-            mut bucket_limiter,
-            partition_key,
```
TBS1996, line 2902: we need the partition_key to send the SendMetrics msg, but now handle_rate_limiting isn't sending metrics to the EnvelopeManager any longer, but rather to the aggregator

relay-server/src/actors/project.rs
```diff
@@ -606,24 +607,92 @@ impl Project {
...
 
+    fn ratelimit_and_merge_buckets(
+        &self,
+        project_state: Arc<ProjectState>,
+        mut buckets: Vec<Bucket>,
+        aggregator: Addr<Aggregator>,
+        #[allow(unused_variables)] envelope_processor: Addr<EnvelopeProcessor>,
+        outcome_aggregator: Addr<TrackOutcome>,
+    ) {
+        let Some(scoping) = self.scoping() else {
```
TBS1996, line 618: I find it a bit ugly that it returns an option, essentially because it might lack a project_id. Maybe we should explore ways of not making the project_id field be optional

relay-server/src/actors/project.rs
```diff
@@ -606,24 +607,92 @@ impl Project {
...
+            let buckets = self.rate_limit_metrics(buckets, outcome_aggregator.clone());
 
             if !buckets.is_empty() {
                 match &mut self.state {
-                    State::Cached(_) => {
+                    State::Cached(state) => {
                         // We can send metrics straight to the aggregator.
                         relay_log::debug!("sending metrics straight to aggregator");
-                        aggregator.send(MergeBuckets::new(self.project_key, buckets));
+                        let state = state.clone();
```
TBS1996, line 692: borrow-checker shenanigans

relay-server/src/actors/project.rs
```diff
@@ -826,9 +909,18 @@ impl Project {
 
         match self.expiry_state() {
             // If the new state is invalid but the old one still usable, keep the old one.
-            ExpiryState::Updated(old) | ExpiryState::Stale(old) if state.invalid() => state = old,
+            ExpiryState::Updated(old) | ExpiryState::Stale(old)
+                if state.invalid() || state.project_id.is_none() =>
```
TBS1996: need some critical eyes on this `state.project_id.is_none()` line.

currently we check after the aggregator if the project_id is there before ratelimiting, and if not, we send it back to the aggregator so that it may be tried again later.  
 
Now that the rate-limiting is entirely before the aggregator, we can no longer do that, so to avoid losing these buckets (or not rate-limiting them), I thought we could rather keep them in the pre-aggregator longer until we have a projectstate with a project_id. 

if anyone can think of an unintended bad consequence of this then let me know. I've looked through it but will check it more later.

relay-server/src/actors/processor.rs
```diff
@@ -547,6 +545,8 @@ struct InnerProcessor {
     project_cache: Addr<ProjectCache>,
     global_config: Addr<GlobalConfigManager>,
     outcome_aggregator: Addr<TrackOutcome>,
+    #[allow(unused)]
```
TBS1996, line 548: because it's feature flagged

relay-server/src/actors/processor.rs
```diff
@@ -555,6 +555,7 @@ struct InnerProcessor {
 
 impl EnvelopeProcessorService {
     /// Creates a multi-threaded envelope processor.
+    #[allow(clippy::too_many_arguments)]
```
jjbayer, line 559: Should we create a helper struct like we did here?

https://github.com/getsentry/relay/blob/75397010d18044a17759161e74475297fe9b0493/relay-server/src/actors/project_cache.rs#L405-L407

relay-server/src/actors/processor.rs
```diff
@@ -563,6 +564,7 @@ impl EnvelopeProcessorService {
         project_cache: Addr<ProjectCache>,
         global_config: Addr<GlobalConfigManager>,
         upstream_relay: Addr<UpstreamRelay>,
+        _aggregator: Addr<Aggregator>,
```
```suggestion
        #[cfg(feature = "processing")]
        aggregator: Addr<Aggregator>,
```

relay-server/src/actors/project.rs
```diff
@@ -606,24 +607,92 @@ impl Project {
...
+            Ok(mut bucket_limiter) => {
+                let cached_rate_limits = self.rate_limits().clone();
+                #[allow(unused_variables)]
+                let was_rate_limited =
+                    bucket_limiter.enforce_limits(Ok(&cached_rate_limits), outcome_aggregator);
+
+                #[cfg(feature = "processing")]
+                if !was_rate_limited && self.config.processing_enabled() {
+                    // If there were no cached rate limits active, let the processor check redis:
+                    envelope_processor.send(RateLimitFlushBuckets { bucket_limiter });
```
jjbayer: Let's rename to `RateLimitBuckets` -- "flush" seems to imply they were already flushed by the aggregator.

relay-server/src/actors/project.rs
```diff
@@ -606,24 +607,92 @@ impl Project {
         });
     }
 
+    fn ratelimit_and_merge_buckets(
```
```suggestion
    fn rate_limit_and_merge_buckets(
```

relay-server/src/actors/project.rs
```diff
@@ -606,24 +607,92 @@ impl Project {
...
+
+        // Only send if the project state is valid, otherwise drop the buckets.
+        if project_state.check_disabled(self.config.as_ref()).is_err() {
+            relay_log::trace!("project state invalid: dropping {} buckets", buckets.len());
+            return;
+        }
+
+        // Re-run feature flag checks since the project might not have been loaded when the buckets
+        // were initially ingested, or feature flags have changed in the meanwhile.
+        self.filter_metrics(&mut buckets);
```
jjbayer, line 634: Check if buckets is empty here? See https://github.com/getsentry/relay/blob/75397010d18044a17759161e74475297fe9b0493/relay-server/src/actors/project.rs#L559-L563

relay-server/src/actors/project.rs
```diff
@@ -606,24 +607,92 @@ impl Project {
...
+    fn ratelimit_and_merge_buckets(
+        &self,
+        project_state: Arc<ProjectState>,
+        mut buckets: Vec<Bucket>,
+        aggregator: Addr<Aggregator>,
+        #[allow(unused_variables)] envelope_processor: Addr<EnvelopeProcessor>,
+        outcome_aggregator: Addr<TrackOutcome>,
+    ) {
+        let Some(scoping) = self.scoping() else {
+            // This shouldn't be possible since we check if project id is present before setting the new project state.
```
jjbayer: If it's supposed to be impossible, let's log an `error!` instead of a `trace!`.

relay-server/src/actors/project.rs
jjbayer: Hmm, not sure if this check is even necessary, will have to think about it a bit more.

jjbayer (commented):
Looks good! Not sure about the `project_id.is_none()` check yet.

relay-server/src/actors/project.rs
TBS1996: nice catch!

relay-server/src/actors/project.rs
jjbayer: Sentry always sends a `project_id`, so I would rather not alter the logic of this check, and log an error on unexpected `None` in `rate_limit_and_merge_buckets`, as you are already doing.

https://github.com/getsentry/sentry/blob/ea1023245fa1962fc3989cb5bb016f6c70b3ea57/src/sentry/relay/config/__init__.py#L354

TBS1996: right, so  I guess it's only lacking in static/proxy mode? it's a bit annoying that the type system doesnt take care of this, but yes then it makes sense to rather log an error

relay-server/src/actors/processor.rs
TBS1996: I followed the convention with _redis there, but if i'll change mine to cfg then i might as well do the same with redis?

relay-server/src/actors/project.rs
jjbayer: Even in static mode we guarantee a project_id: https://github.com/getsentry/relay/blob/2a86991dfbdd2ae6b1dc2a188906947ba89f0f86/relay-server/src/actors/project_local.rs#L101-L108

relay-server/src/actors/processor.rs
TBS1996: So we already have the `InnerProcessor`, if we wanna do something like this we should probably just expose it. I started it but I reverted it because it doesn't really seem to be any point in it, we just end up exposing more than we need, and in the end it still looks the same whether we put a long list of variables directly into a struct rather than having many args in `new`

relay-server/src/actors/project.rs
TBS1996: so given that it's an optional, when is it ever None?

jjbayer (approved):
Let's revert test to limit=2, then this should be good to go!

PR #2654, opened 2023-10-25
feat(spans): Group by domain and last segment of the path

phacops:
Instead of trying to group by the full path, since it's containing a whole lot of cardinality, we'll isolate the last part of the path (which is usually the filename of the resource but can also be a part of the path or empty if the URL finishes with a `/`), scrub this part only and construct a group based on the scrubbed domain and this scrubbed last part.

@jjbayer I don't think the logic meaning to add a dummy base URL to relative or absolute paths is working as intended. In the case of domains without a scheme, the dummy domain is added and is now part of the path, which means it's scrapped with this PR. I haven't figured out a proper solution to this issue yet. Ideally, we'd identify the scheme is empty and use a dummy scheme or something.


jjbayer (approved):
> Ideally, we'd identify the scheme is empty and use a dummy scheme or something.

@phacops we could use a dummy domain for paths that start with `/`, and for paths that don't, use a dummy scheme instead.

phacops:
Let's not handle that case as I'm not sure it's a real situation yet. It'll go to the relative path logic and ends up with a wildcard and the filename.

Issue #2653, opened 2023-10-24
Wrong return value of `Meta::has_remarks` ?


Uinelj:
### Environment

What version are you running? Etc.
relay 23.10.1, Rust 1.73.0

### Steps to Reproduce

I'm checking if a `String` has been altered to remove sensitive information.
This code is based on the unit tests here: https://github.com/getsentry/relay/blob/91719634a1348c15f5493df17f553426820a44ee/relay-pii/src/builtin.rs#L440

```rs
let mut config = PiiConfig::default();
        config.applications = {
            let mut map = BTreeMap::new();
            map.insert(
                ValueType::String.into(),
                vec![
                    "@ip".into(),
                ],
            );
            map
        };
        let compiled = config.compiled();
        let mut processor = PiiProcessor::new(&compiled);
        let state = processing_state();

        let input = "before ::1 after".to_string();
        let input2 = "nothing to filter here".to_string();

        let mut root = Annotated::new(input);
        process_value(&mut root, &mut processor, &state).unwrap();
        println!("{root:?}"); // ip has been masked
        println!("{:?}", root.meta().has_remarks()); // returns false 
        println!("{:?}", root.meta().iter_remarks().count()); // returns 1

        let mut root = Annotated::new(input2);
        process_value(&mut root, &mut processor, &state).unwrap();
        println!("{root:?}"); // nothing has been done
        println!("{:?}", root.meta().has_remarks()); // returns false
        println!("{:?}", root.meta().iter_remarks().count()); // returns 0
```

### Expected Result

I'd expect `has_remarks()` to return `true` on the first value, since ip is masked and using `root.meta().iter_remarks().count()` returns `1`.

### Actual Result

`has_remarks()` returns `false` even if a remark is present. 

I think the issue is here:

```rs
  /// Indicates whether this field has remarks.
    pub fn has_remarks(&self) -> bool {
        self.0.as_ref().map_or(false, |x| x.remarks.is_empty())
    }
```

I don't precisely know how `Meta` is laid out, but would it make sense to rely solely on `x.remarks.is_empty`?

Let me know if you need more details :)

TBS1996:
Hi, great catch! You are right, this is a bug. It's a simple case of a flipped bool. I've made a fix now:  https://github.com/getsentry/relay/pull/2655

PR #2655, opened 2023-10-25
fix(protocol): Remove meta::has_remarks method.

TBS1996:
`has_remarks` should return true if there are any remarks. However, it returns true if the remarks vector is *empty*. 
it seems the method was not in use at all, however. 

credits to @Uinelj for discovering this: #2653 

#skip-changelog


iker-barriocanal (commented):
Having a test in cases like this could be beneficial I think

TBS1996:
> Having a test in cases like this could be beneficial I think

well that would certainly have stopped this bug in the first place. I added a test now!

iker-barriocanal (changes requested):
Is this the method that's not being used? We should get rid of it in that case.

iker-barriocanal (approved):

PR #2656, opened 2023-10-25
chore(project_state): Add config option for always requesting full project state

olksdr:
Introduce the configuration option, which allows to configure Relay to always request the full project state from the upstream. 

related: https://github.com/getsentry/team-ingest/issues/231

#skip-changelog

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -1746,7 +1746,7 @@ impl Config {
     }
 
     /// Returns `true` if the full project state should be requested from upstream.
-    pub fn project_state_full(&self) -> bool {
+    pub fn request_full_project_config(&self) -> bool {
         self.values.cache.project_request_full_config
     }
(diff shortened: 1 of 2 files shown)


iker-barriocanal (approved):

relay-config/src/config.rs
```diff
@@ -821,6 +821,8 @@ pub struct Spool {
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(default)]
 struct Cache {
+    /// The full project state will be requested by this Relay if set to `true`.
+    project_state_full: bool,
```
jjbayer: nit: Should we call this `project_request_full_config` or something similar?

olksdr: @jjbayer  I like it. Renamed.

jjbayer (approved):

PR #2657, opened 2023-10-25
feat(normalization): Run store processor in light normalization

iker-barriocanal:
Run `StoreProcessor` as part of light normalization in all relays. Since `is_renormalize` is enabled, effectively this change is a NOOP.

## Context

As we plan on getting rid of light normalization and unifying all normalization, we'll move all normalization steps from light normalization to `StoreNormalization`.


iker-barriocanal:
I'll create a new processor instead in a new PR.

PR #2659, opened 2023-10-25
feat(spans): Index segments for all organizations

phacops:
We need segment information in the spans dataset to power the web services module for Starfish. This PR will start indexing segments for all organizations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add inbound filters option to filter legacy Edge browsers (i.e. versions 12-18 ) ([#2650](https://github.com/getsentry/relay/pull/2650))
 - Group resource spans by scrubbed domain and filename. ([#2654](https://github.com/getsentry/relay/pull/2654))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add inbound filters option to filter legacy Edge browsers (i.e. versions 12-18 ) ([#2650](https://github.com/getsentry/relay/pull/2650))
 - Group resource spans by scrubbed domain and filename. ([#2654](https://github.com/getsentry/relay/pull/2654))
+- Index segments for all organizations. ([#2659](https://github.com/getsentry/relay/pull/2659))
```
```suggestion
- Convert transactions to spans for all organizations. ([#2659](https://github.com/getsentry/relay/pull/2659))
```

jjbayer (approved):
@phacops any idea how much extra load this will cause? IIRC the avg. number of spans per transactions is ~20, so I guess a 5% increase in load should be manageable.

Please add a PR description so we can later figure out why this change was made at this point in time.

PR #2662, opened 2023-10-27
feat(spans): Add TTID and TTFD tags

jjbayer:
If `ui.load.initial_display` and/or `ui.load.full_display` spans are present, mark spans happening before those with `ttid` and `ttfd` tags.

### DIFF
--- a/relay-event-normalization/src/normalize/span/tag_extraction.rs
+++ b/relay-event-normalization/src/normalize/span/tag_extraction.rs
@@ -133,7 +133,8 @@ pub(crate) fn extract_span_tags(event: &mut Event, config: &Config) {
         return;
     };
 
-    let (ttid, ttfd) = display_times(spans.as_ref());
+    let ttid = timestamp_by_op(spans, "ui.load.initial_display");
+    let ttfd = timestamp_by_op(spans, "ui.load.full_display");
 
     for span in spans {
         let Some(span) = span.value_mut().as_mut() else {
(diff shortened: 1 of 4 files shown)


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -772,4 +819,93 @@ LIMIT 1
...
+            },
+        );
+
+        let spans = event.spans.value().unwrap();
+
+        // First two spans contribute to initial display & full display:
+        for span in &spans[..2] {
+            let tags = span.value().unwrap().sentry_tags.value().unwrap();
+            assert_eq!(tags.get("ttid").unwrap().as_str(), Some("true"));
+            assert_eq!(tags.get("ttfd").unwrap().as_str(), Some("true"));
```
jjbayer: @shruthilayaj spans that contribute to initial display also contribute to full display. Does that make sense?

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -376,9 +389,43 @@ pub(crate) fn extract_tags(span: &Span, config: &Config) -> BTreeMap<SpanTagKey,
...
+                span_tags.insert(SpanTagKey::TimeToFullDisplay, "ttfd".to_owned());
+            }
+        }
+    }
+
     span_tags
 }
 
+/// Finds time-to-initial/full-display spans and returns its end times.
+fn display_times(spans: &[Annotated<Span>]) -> (Option<Timestamp>, Option<Timestamp>) {
```
olksdr: Maybe instead of the tuple here it would be clearer to create a struct and set the fields with proper names in it. 
It can be easy to mistaken the order of the returned values just looking at the function signature.

olksdr (approved):
Left one small comment, otherwise LGTM.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: Good point. I refactored this to use `.find()` instead of writing a homemade loop now, let me know what you think.

olksdr: yeah, nice!

relay-event-normalization/src/normalize/span/tag_extraction.rs
shruthilayaj: yes, TTID contributing spans are a subset of TTFD contributing spans!

shruthilayaj (approved):
🚢

PR #2663, opened 2023-10-27
test(spool): Remove test with relative path

jjbayer:
The test `create_spool_directory_root_path` creates a couple of local files: `spool.db`, `spool.db-wal`, `spool.db-shm`.

Assuming that this test was introduced to test if relative paths work at all, we can consider it proven and remove the test.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/spooler/mod.rs
+++ b/relay-server/src/actors/spooler/mod.rs
@@ -970,7 +970,6 @@ impl Drop for BufferService {
 
 #[cfg(test)]
 mod tests {
-    use std::path::PathBuf;
     use std::time::{Duration, Instant};
 
     use insta::assert_debug_snapshot;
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/spooler/mod.rs
```diff
@@ -1026,8 +1026,35 @@ mod tests {
...
+            }
+        }
+    }
+
+    impl Drop for TempDirGuard {
+        fn drop(&mut self) {
+            std::env::set_current_dir(&self.previous_path).unwrap();
+            std::fs::remove_dir_all(&self.path).unwrap();
+        }
+    }
```
olksdr: Do we want to move this into a separate `testutils` module and then reuse it for other tests as well?

olksdr (approved):

relay-server/src/actors/spooler/mod.rs
jjbayer: Done. I also changed the helper to *not* automatically change the working directory, because it caused problems in parallel test runs: https://github.com/getsentry/relay/actions/runs/6666168668/job/18117173821?pr=2663#step:5:1661

TBS1996:
seems like std::env::tempdir always make the same directory path and that's why youre not able to clean up, since its used by someone else? 

```Rust
pub fn temp_dir() -> PathBuf {
    crate::env::var_os("TMPDIR").map(PathBuf::from).unwrap_or_else(|| {
        if cfg!(target_os = "android") {
            PathBuf::from("/data/local/tmp")
        } else {
            PathBuf::from("/tmp")
        }
    })
}
```


maybe we can use the tempfile crate instead, it generates a random name: https://github.com/Stebalien/tempfile/blob/21c639e089ae503ec5865801caba5435e4709d9b/src/dir.rs#L123

so that shouldn't cause problems when running tests in parallel

iker-barriocanal (approved):
Thanks for the cleanup!

olksdr (approved):

PR #2668, opened 2023-10-30
ref(normalization): Introduce NormalizeProcessor

iker-barriocanal:
We're working on having a single processor to run all normalization-related steps in Relay, and this PR introduces it: `NormalizeProcessor`. This is only the first step towards that goal, and more PRs will follow up to merge other processors. The scope of this PR is the following:

- Create a new `NormalizeProcessor`.
- Since there was already a processor with such a name in `StoreProcessor`, rename this one to `StoreNormalizeProcessor`. Note that this will eventually be merged into `NormalizeProcessor` too.
- `StoreNormalization` has no further modifications. It'll be addressed in future PRs.
- Move light normalization steps into `NormalizeProcessor`. These are moved directly, copy-pasting, without additional refactors.
- Keep the old `light_normalize_event` function. This function is called from plenty of places in the codebase including tests, so we'll keep it until `NormalizeProcessor` is completed and only then remove the function indirection, in order to minimize the risk of bugs.

There are no functional changes in this PR.

#skip-changelog


relay-event-normalization/src/normalize/processor.rs
```diff
@@ -0,0 +1,1495 @@
+use std::collections::hash_map::DefaultHasher;
```
jjbayer, line 9: Even though it's a private module, should we add a module doc comment here?

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -0,0 +1,1495 @@
...
+        if user.ip_address.value().is_none() {
+            // In an ideal world all SDKs would set {{auto}} explicitly.
+            if let Some("javascript") | Some("cocoa") | Some("objc") = platform {
+                user.ip_address = Annotated::new(client_ip.to_owned());
+            }
+        }
+    }
+}
+
+// Sets the user's GeoIp info based on user's IP address.
```
jjbayer:
```suggestion
/// Sets the user's GeoIp info based on user's IP address.
```

jjbayer (approved):
Looks good to me, but I'll leave final review to @olksdr as the driver of this initiative.

Please run `make doc-rust` to check if the crate documentation still makes sense after this restructuring.

relay-event-normalization/src/normalize/processor.rs
iker-barriocanal: I'd defer documenting the undocumented modules of the crate until we've moved all the processors, as the structure may change significantly.

olksdr: I would still add docs here, to at least clarify the general idea behind this normalize processor.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -0,0 +1,1495 @@
...
+};
+
+use crate::LightNormalizationConfig;
+
+/// Configuration for [`NormalizeProcessor`].
+#[derive(Debug, Default)]
+pub struct NormalizeProcessorConfig<'a> {
+    /// Light normalization config.
+    // XXX(iker): we should split this config appropriately.
+    light_config: LightNormalizationConfig<'a>,
```
olksdr, line 44: To have this indirection with few layers of configs is really petty.  
Can this be already flattened into the `NormalizeProcessorConfig` in this PR?

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -0,0 +1,1495 @@
...
+    fn from(config: LightNormalizationConfig<'a>) -> Self {
+        Self {
+            light_config: config,
+        }
+    }
+}
+
+/// Normalizes an event, rejecting it if necessary.
+///
+/// The normalization consists on applying a series of transformations on the
```
olksdr:
```suggestion
/// The normalization consists of applying a series of transformations on the
```

olksdr (approved):
I think it's a good beginning. 
But definitely have to followup with cleanup to avoid confusion and make sure the further contributions are done in connection to the new normalize processor.

relay-event-normalization/src/normalize/processor.rs
iker-barriocanal: I'll keep the light normalization config around for now, and move keys out of it as we get rid of it. Eventually, it'll have no keys and we'll remove it.

relay-event-normalization/src/normalize/processor.rs
iker-barriocanal: Added some vague comments on [4a26e79](https://github.com/getsentry/relay/pull/2668/commits/4a26e79eb0155cebb3a895b14fcb8979f0d2b30f), I'll elaborate more once we've purged light normalization and moved all processing steps.

olksdr (changes requested):
When the `master` was merged I think there are few functions got lost, like `normalize_nel_report`. Have that been moved somewhere?

olksdr (approved):

PR #2670, opened 2023-10-30
feat(pii): Allow pii scrubbing expressions for safe fields

Dav1dde:
This doesn't simply make matching on arrays work, it enables full expression support for safe fields in the datascrubbing config.


olksdr (approved):
lgtm

And next I think we will need to update the docs on our site , once this is deployed.

PR #2673, opened 2023-10-30
feat(profiling): Log SDK name and version in errors

phacops:
Follow-up to my previous PR to make logs even more useful by adding the previous error containing the value with the issue and the SDK name and version.

#skip-changelog

### DIFF
--- a/relay-profiling/src/lib.rs
+++ b/relay-profiling/src/lib.rs
@@ -134,19 +134,26 @@ fn minimal_profile_from_json(
 }
 
 pub fn parse_metadata(payload: &[u8], event: Option<&Event>) -> Result<(), ProfileError> {
-    let (sdk_name, sdk_version) = match event {
-        Some(event) => (event.sdk_name(), event.sdk_version()),
-        _ => ("unknown", "unknown"),
+    let (sdk_name, sdk_version, platform, project_id) = match event {
+        Some(event) => (
+            event.sdk_name(),
+            event.sdk_version(),
+            event.platform.as_str().unwrap_or("unknown"),
+            event.project.value().unwrap_or(&0),
+        ),
+        _ => ("unknown", "unknown", "unknown", &0),
     };
     let profile = match minimal_profile_from_json(payload) {
         Ok(profile) => profile,
         Err(err) => {
             relay_log::warn!(
                 error = &err as &dyn Error,
-                "invalid profile (minimal, sdk: {} v{}). original: {}.",
-                sdk_name,
-                sdk_version,
-                err.inner(),
+                from = "minimal",
+                original = err.inner().to_string(),
+                platform = platform,
+                project_id = project_id,
+                sdk_name = sdk_name,
+                sdk_version = sdk_version,
             );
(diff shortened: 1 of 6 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -1152,7 +1152,13 @@ impl EnvelopeProcessorService {
                 if !profiling_enabled {
                     return ItemAction::DropSilently;
                 }
-                match relay_profiling::expand_profile(&item.payload(), state.event.value()) {
+                // If we don't have an event at this stage, we need to drop the profile.
+                let event = if let Some(event) = state.event.value() {
+                    event
+                } else {
+                    return ItemAction::DropSilently;
+                };
```
iker-barriocanal:
```suggestion
                let Some(event) = state.event.value() else {
                    return ItemAction::DropSilently;
                };
```
I think you can do a `let..else` here

relay-profiling/src/lib.rs
```diff
@@ -127,17 +126,29 @@ struct MinimalProfile {
...
         Err(err) => {
-            relay_log::warn!(error = &err as &dyn Error, "invalid profile (minimal)");
-            return Err(err);
+            relay_log::warn!(
+                error = &err as &dyn Error,
+                "invalid profile (minimal, sdk: {} v{}). original: {}.",
+                sdk_name,
+                sdk_version,
+                err.inner(),
+            );
```
iker-barriocanal: Moving the SDK name and version to the message's tags means Sentry will group these errors under the same issue, and provide the ability to have insights into how many events are coming from what SDKs. Your call depending what you prefer hehe

iker-barriocanal (approved):

relay-profiling/src/lib.rs
phacops: My understanding is those won't be creating issues because they're warnings. We intentionally did this as to not overwhelm s4s with a ton of potentially unactionable profiling issues.

I would love to use a more structured logging way of doing things, like being able to pass a map of tags I'd be able to filter in GCP logging but my understanding is it's not exactly possible and that's why we're doing it this way.

iker-barriocanal: These messages will get to s4s in the form of warning messages. These are the same as regular messages captured with SDK's `capture_message` but with a different level -- `level:warning` in the issues page should allow you to filter these, for example. Just to be clear, no strong opinions here, just providing a different way you may like more.

jjbayer: > My understanding is those won't be creating issues because they're warnings.

That's right, only `error!` gets converted to sentry error events, so I'd argue it does not really matter here. But I agree that the following would be slightly more semantic, while still printing all info to stdout/stderr:

```rust
relay_log::warn!(
    error = &err as &dyn Error,
    sdk_name = sdk_name,
    sdk_version = sdk_version,
    original = err.inner(),
    "invalid profile (minimal)",
);
```

phacops: Ah, I didn't know I could do that. Perfect then.

PR #2676, opened 2023-10-31
feat(spans): Group all browser resources together

jjbayer:
We want to filter out resource spans from browser extensions, so scrub their description to `browser-extension://*` such that they all end up in the same group.

ref: https://github.com/getsentry/sentry/issues/58723

#skip-changelog


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -545,7 +540,7 @@ mod tests {
         chrome_extension,
         "chrome-extension://begnopegbbhjeeiganiajffnalhlkkjb/img/assets/icon-10k.svg",
         "resource.other",
-        "chrome-extension://*/img/assets/icon-*k.svg"
+        "browser-extension://*"
```
jjbayer, line 543: @DominikB2014 my idea here was to group all browser extension resources together so they can be easily filtered out on the querying side. Is this an acceptable solution, or do we need to actually remove those spans from our data?

iker-barriocanal (approved):

olksdr (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
DominikB2014: Yeah i think this would be good, if can can filter out all `browser-extensions` when querying.

One question, In this case what would the domain be set to? We've been having very random domains set for chrome extensions
image

jjbayer: Good question, let me add a test for that.

jjbayer: With the current change, the domain will be `"*"`.

DominikB2014: So to filter out extensions, we would add `span.domain != '*'`?

jjbayer: I guess the safer way would be to filter on the `group` or `description != 'browser-extension://*`, because other resources could theoretically also have `domain == '*'`.

DominikB2014: Yeah that makes sense, this should work for the product then, if there's any issues (I don't suspect any), we can revist.
Thanks for making the update @jjbayer!

DominikB2014 (approved):

PR #2679, opened 2023-10-31
ref(normalization): Move span validation to NormalizeProcessor

iker-barriocanal:
`TransactionsProcessor` performs some span attribute validation logic. This PR moves that logic to `NormalizeProcessor` in an attempt to move a step closer to having a single processor.

Notes:
- This new logic runs before light normalization's `is_renormalize` flag check. This means that this new code will run even if the flag is disabled, which previously wasn't the case. The purpose of `is_renormalize` is to make a decision on non-idempotent steps, and these checks are idempotent so this action should be safe.
- I've started moving stuff from light normalization to `NormalizeProcessor` from top to bottom, but the end result will be the same. The span-related functionality at the bottom of the function is being updated often, so I'd rather to approach that at a later time to avoid conflicts.


relay-event-normalization/src/normalize/span/attributes.rs
```diff
@@ -608,4 +660,174 @@ mod tests {
...
+
+        process_value(
+            &mut event,
+            &mut NormalizeProcessor::default(),
+            ProcessingState::root(),
+        )
+        .unwrap();
+
+        let span = get_value!(event.spans).unwrap().first().unwrap();
+        assert_eq!(get_value!(span.op).unwrap(), &"default".to_owned());
```
iker-barriocanal: This is modified from the original test. There used to be a snapshot over the whole span that now fails because `transaction.received` and `transaction.contexts.trace.status` exist. Instead of updating the snapshot, I've updated the assertion of what the test is validating.

relay-event-normalization/src/normalize/span/attributes.rs
```diff
@@ -142,12 +144,62 @@ fn compute_span_exclusive_time(event: &mut Event) {
     }
 }
 
+/// Returns a [`ProcessingAction`] error if a span's timestamps are invalid.
+///
+/// For a span's timestamps to be valid, the following should happen:
+/// - Start and end timestamps must exist.
+/// - Start timestamp must be no greater than end timestamp.
+pub(crate) fn validate_span_timestamps(span: &Span) -> Result<(), ProcessingAction> {
```
iker-barriocanal: I decided to put these functions in this module as it seems the most suitable place. The structure of the `relay-event-normalization` crate is based on having multiple processors, so once we have a single one the file and module structure will be much simpler.

relay-event-normalization/src/normalize/span/attributes.rs
```diff
@@ -608,4 +660,174 @@ mod tests {
...
+                trace_id: Annotated::new(TraceId("4c79f60c11214eb38604f4ae0781bfb2".into())),
+                ..Default::default()
+            })]),
+            ..Default::default()
+        });
+
+        assert_eq!(
+            process_value(
+                &mut event,
+                &mut NormalizeProcessor::default(),
```
iker-barriocanal: These tests also use `NormalizeProcessor` instead of `TransactionProcessor`.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -188,6 +197,22 @@ impl<'a> Processor for NormalizeProcessor<'a> {
 
         Ok(())
     }
+
+    fn process_span(
```
jjbayer, line 205: Does this now also run for error events? I believe the `TransactionProcessor` prevented that.

relay-event-normalization/src/normalize/span/attributes.rs
jjbayer: This module was previously specifically about `SpanAttribute`. Should we move this function to either `span/mod.rs` or a separate `span/timestamps.rs` instead?

relay-event-normalization/src/normalize/span/attributes.rs
jjbayer: Could we simplify this test and just call `validate_span_timestamps` on a `Span` instead of creating an event?

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -68,11 +69,13 @@ impl<'a> Processor for NormalizeProcessor<'a> {
         &mut self,
         event: &mut Event,
         meta: &mut Meta,
-        _: &ProcessingState<'_>,
+        state: &ProcessingState<'_>,
     ) -> ProcessingResult {
-        let config = &self.config.light_config;
+        let light_normalization_config = self.config.light_config.clone();
```
jjbayer: Does this need cloning if we declare it after calling `process_child_values`?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Disable resource link span ingestion. ([#2647](https://github.com/getsentry/relay/pull/2647))
 - Collect `http.decoded_response_content_length`. ([#2638](https://github.com/getsentry/relay/pull/2638))
 - Add TTID and TTFD tags to mobile spans. ([#2662](https://github.com/getsentry/relay/pull/2662))
+- Validate span timestamps and IDs in light normalization. ([#2679](https://github.com/getsentry/relay/pull/2679))
```
jjbayer: Do we actually need a changelog for this? Span validation was already part of light normalization (through the `TransactionProcessor`), so that does not change from an outsider's point of view. Let me know if I'm missing something.

relay-event-normalization/src/normalize/span/attributes.rs
iker-barriocanal: I moved the validation functions to `span/validate.rs`, let me know if this doesn't sound right.

relay-event-normalization/src/normalize/span/attributes.rs
iker-barriocanal: I'd like to minimize the changes on this PR, but happy to update it in a follow-up.

relay-event-normalization/src/normalize/processor.rs
iker-barriocanal: Yes, it does for errors. Errors aren't supposed to have spans so I thought it'd be convenient to skip that check over being a bit more defensive.

CHANGELOG.md
iker-barriocanal: Good catch! These checks didn't run when `renormalize` was enabled, which now do (first note on the PR description). I updated the changelog to be a bit more explicit.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -89,8 +91,8 @@ impl<'a> Processor for NormalizeProcessor<'a> {
         // TODO: Parts of this processor should probably be a filter so we
         // can revert some changes to ProcessingAction)
         let mut transactions_processor = transactions::TransactionsProcessor::new(
-            config.transaction_name_config.clone(),
-            config.transaction_range.clone(),
+            light_normalization_config.transaction_name_config.clone(),
```
olksdr: What the reason to change the `config` to `light_normalization_config` in this PR? I'm failing to see the benefit but it also introduces a lot of unreleased diffs.

relay-event-normalization/src/normalize/span/validation.rs
```diff
@@ -0,0 +1,234 @@
...
+use relay_event_schema::protocol::Span;
+
+use relay_event_schema::processor::ProcessingAction;
+
+/// Returns a [`ProcessingAction`] error if a span's timestamps are invalid.
+///
+/// For a span's timestamps to be valid, the following should happen:
+/// - Start and end timestamps must exist.
+/// - Start timestamp must be no greater than end timestamp.
+pub(crate) fn validate_span_timestamps(span: &Span) -> Result<(), ProcessingAction> {
```
olksdr: Instead of introducing this module, can these function be moved into the `utils`? Which seems to be a good place to have the helper functions.

And the tests in this file can go to normalize processor module.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -199,6 +207,22 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
     }
+
+    fn process_span(
+        &mut self,
+        span: &mut Span,
+        _: &mut Meta,
+        state: &ProcessingState<'_>,
+    ) -> ProcessingResult {
+        validate_span_timestamps(span)?;
+        validate_span_ids(span)?;
```
olksdr: I'm not convince that splitting this function into helper functions is the right thing to do for this PR. It's small enough that I would keep as it was.

olksdr (commented):
> This new logic runs before light normalization's is_renormalize flag check. This means that this new code will run even if the flag is disabled, which previously wasn't the case. The purpose of is_renormalize is to make a decision on non-idempotent steps, and these checks are idempotent so this action should be safe.

Do we have enough test coverage to prove this?

relay-event-normalization/src/normalize/span/attributes.rs
iker-barriocanal: Ended up moving it to `utils` to not create yet another module.

olksdr (approved):

PR #2681, opened 2023-10-31
ref(metrics): Flatten AggregatorServiceConfig

TBS1996:
Prior to splitting the aggregatorservice with the aggregator we had just one big config. After we split it, we also split the config to reflect the service and hte aggregator, where the service config was just the aggregatorconfig flattened together with the values related to the service.

This nested structure, while it makes more sense in the implementation, made it harder to configure. This PR therefore 'unflattens' the service config so it looks as it did in the past. We keep the config for the inner aggregator and simply implement a `From<&aggregatorserviceconfig> for aggregatorconfig` in order to keep some of the implementation benefits of separation.  
 
The only downside is that it's less 'dry'. The values for inner aggregator is now present in both the serviecconfig and the inner config, so thats something we have to keep in mind in the future if we add new values. 

#skip-changelog


jjbayer (commented):
I ran `make doc-rust` and the resulting docs look good to me! @TBS1996 could you also remove the doc comment that says "Only public for benchmarks", and add a doc comment to the module file itself instead, something like 

```rust
//! Core functionality of metrics aggregation.
```

TBS1996:
yup, done!

jjbayer (approved):
LGTM, please ask @jan-auer if he wants to have another look.

jan-auer (approved):
LGTM. 

The docs for `Aggregator` still reside on `AggregatorService`. It would be good to move them, so that the module-level links point at a documented interface.

PR #2682, opened 2023-10-31
ref(metrics): Partition and split metrics buckets just before sending

Dav1dde:
Moves metric partitioning and splitting into the envelope processor. Metrics are now sliced without any allocations.

Also track metric outcomes now instead of merging buckets back into the aggregator.

Tracking outcomes for metrics has introduced a side effect that all relay generated envelops for metrics are now properly tracked which impacts the accepted metrics (`relay.event.accepted`) of envelopes. This will cause a considerable bump in the accepted metric, but does not actually indicate a change in the overall amount of envelopes accepted by relay.


.gitignore
```diff
@@ -10,6 +10,7 @@ target
 
 # Runtime files
 /.relay/
+spool.db
```
Dav1dde: @Dav1dde fix this

relay-metrics/src/aggregator.rs
```diff
@@ -92,6 +96,24 @@ impl BucketKey {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
```
TBS1996, line 104: docs explaining the purpose of bucketkeyref would be nice

relay-metrics/src/aggregator.rs
```diff
@@ -526,7 +540,7 @@ impl Aggregator {
             aggregator = &self.name,
         );
 
-        let mut buckets = HashMap::<ProjectKey, Vec<HashedBucket>>::new();
+        let mut buckets = HashMap::<ProjectKey, Vec<_>>::new();
```
TBS1996:
```suggestion
        let mut buckets = HashMap::<ProjectKey, Vec<Bucket>>::new();
```

i dont really see why we should have fewer annotations when it doesn't add verbosity, maybe im missing something

iker-barriocanal:
Any chance we can split this PR? 2,000+k LOC is complicated to follow in a review.

TBS1996:
tried to make a visualization

Before:

```mermaid
graph TD
    Project -->|RateLimitBuckets| EnvelopeProcessorService
    Project -->|MergeBuckets| AggregatorService
    EnvelopeProcessorService -->|MergeBuckets| AggregatorService
    AggregatorService -->|FlushBuckets| ProjectCacheService
    ProjectCacheService -->|SendMetrics| EnvelopeManagerService
    EnvelopeManagerService -->|SendRequest| UpstreamRelay

```



After:

```mermaid
graph TD
    Project-->|RateLimitBuckets|EnvelopeProcessorService
    Project-->|MergeBuckets|AggregatorService
    EnvelopeProcessorService-->|MergeBuckets|AggregatorService
    AggregatorService -->|FlushBuckets| ProjectCacheService
    ProjectCacheService-->|SendMetrics|EnvelopeManagerService
    EnvelopeManagerService-->|EncodeMetrics|EnvelopeProcessor
    EnvelopeProcessor-->|SubmitEnvelope|EnvelopeManagerService


```


edit: your newest one:


```mermaid
graph TD
    Project-->|RateLimitBuckets|EnvelopeProcessorService
    Project-->|MergeBuckets|AggregatorService
    EnvelopeProcessorService-->|MergeBuckets|AggregatorService
    AggregatorService -->|FlushBuckets| ProjectCacheService
    ProjectCacheService-->|EncodeMetrics|EnvelopeProcessor
    EnvelopeProcessor-->|SubmitEnvelope|EnvelopeManagerService


```

relay-server/src/actors/processor.rs
```diff
@@ -482,6 +486,21 @@ impl EncodeEnvelope {
     }
 }
 
+/// Encodes metrics into an envelop ready to be sent upstream.
```
TBS1996:
```suggestion
/// Encodes metrics into an envelope ready to be sent upstream.
```

TBS1996 (commented):
as iker said, if you're able to split this up then that would be greatly appreciated, It's hard to confidently approve a PR with so many changes

relay-metrics/src/aggregator.rs
Dav1dde: It was annoying me, because if it wasn't specified in the first place I wouldn't have had to change it because the compiler would have correctly inferred it.

But I can actually completely get rid of the annotations, will do that instead.

Dav1dde:
@TBS1996 @iker-barriocanal the PR is now 2 commits:

1) Everything related to slicing buckets into multiple smaller buckets without re-allocating and cloning
2) Wrapping Metrics in a `ManagedEnvelope` and producing outcomes

relay-config/src/config.rs
```diff
@@ -968,6 +972,26 @@ pub struct Processing {
...
+    /// adds some additional overhead, this number is approxmate and some safety margin should be
+    /// left to hard limits.
+    #[serde(default = "default_metrics_max_batch_size")]
+    pub metrics_max_batch_size: ByteSize,
+    /// The approximate maximum number of bytes submitted in one metrics batch for processing
+    /// relays.
+    ///
+    /// Overrides [`Self::metrics_max_batch_size`] when specified on processing relays.
+    #[serde(default)]
+    pub metrics_max_batch_size_processing: Option<ByteSize>,
```
Dav1dde: Maybe should set the default to 100KiB what it currently is (configured in ops configs)

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -34,6 +36,65 @@ pub struct MetricsLimiter<Q: AsRef<Vec<Quota>> = Vec<Quota>> {
...
+    };
+
+    let has_profile =
+        matches!(mri.name, "usage" | "duration") && metric.tag(PROFILE_TAG) == Some("true");
+
+    Some(TransactionCount { count, has_profile })
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ExtractionMode {
```
TBS1996, line 78: maybe docs here ?

relay-metrics/src/aggregator.rs
```diff
@@ -900,6 +886,36 @@ impl fmt::Debug for Aggregator {
     }
 }
 
+/// Split buckets into N logical partitions, determined by the bucket key.
+pub fn partition_buckets(
```
TBS1996, line 892: could you move this over the 'new' function to make the diff smaller?

relay-server/src/actors/processor.rs
```diff
@@ -3119,6 +3198,26 @@ impl Service for EnvelopeProcessorService {
     }
 }
 
+fn create_metrics_item(buckets: &BucketsView<'_>, extraction_mode: ExtractionMode) -> Item {
```
TBS1996, line 2842: thoughts on this as an associated function?

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -34,6 +36,65 @@ pub struct MetricsLimiter<Q: AsRef<Vec<Quota>> = Vec<Quota>> {
...
+        if usage {
+            Self::Usage
+        } else {
+            Self::Duration
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct TransactionCount {
```
TBS1996, line 102: also docs here

relay-server/src/actors/envelopes.rs
```diff
@@ -331,34 +334,29 @@ impl EnvelopeManagerService {
...
-        };
-
-        let mut item = Item::new(ItemType::MetricBuckets);
-        item.set_payload(ContentType::Json, serde_json::to_vec(&buckets).unwrap());
-        let mut envelope = Envelope::from_request(None, RequestMeta::outbound(dsn));
-        envelope.add_item(item);
+        #[allow(unused_mut)]
+        let mut partitions = self.config.metrics_partitions();
+        #[allow(unused_mut)]
+        let mut max_batch_size_bytes = self.config.metrics_max_batch_size_bytes();
```
TBS1996: im wondering if we really need this function, couldn't we send directly to the envelopeprocessor from the projectcache?

relay-server/src/actors/processor.rs
Dav1dde: What would be your reasoning? The only relationship with the processor I see is that the processor calls it (e.g. this function could also be near the `Item` itself or utils package).

relay-server/src/actors/envelopes.rs
Dav1dde: Kind of, it currently exists to specify different processing configs depending on the upstream sink (Kafka, Http etc.) and this is the only service that should know what the upstream is (because it is also the service which decides where to send envelopes to).

E.g. we want smaller batches for Kafka and bigger batches for Http.

TBS1996: couldn't we just do that with the processing_enabled() flag in config? that sets the store to `Some` [here](https://github.com/getsentry/relay/blob/2757887550a59a9e6f64c00a55932a6cfe505a77/relay-server/src/service.rs#L310-L312) which deterministically sets the store_forwarded to `Some` on `EnvelopeManager` [here](https://github.com/getsentry/relay/blob/2757887550a59a9e6f64c00a55932a6cfe505a77/relay-server/src/service.rs#L162-L166). Although if you think it might be too implicit then i could agree with that,

Dav1dde: Yes we could in practice, but I specifically didn't want this implicit behaviour.

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+///
+/// See [`estimate_size`] for how the size of a bucket is calculated.
+struct BucketsViewBySizeIter<'a> {
+    /// Source slice.
+    inner: &'a [Bucket],
+    /// Current position in the slice.
+    current: Index,
+    /// Terminal position.
+    end: Index,
+    /// Maximum size of in bytes of each slice.
```
TBS1996:
```suggestion
    /// Maximum size in bytes of each slice.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+/// There are three possible return values:
+///  - `BucketFits` if the bucket fits entirely into the size budget.
+///  - `MoveToNextBatch` if the size budget cannot even hold the bucket name and tags. There is no
+///    split, the entire bucket is moved.
+///  - `Split(at)` if the bucket fits partially, the bucket should be split `at`.
+///
+/// This is an approximate function. The bucket is not actually serialized, but rather its
+/// footprint is estimated through the number of data points contained. See
+/// `estimate_size` for more information.
+fn split_at(bucket: &BucketView<'_>, size: usize, min_split_size: usize) -> SplitDecision {
```
TBS1996: maybe max_size rather than just size?

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+use crate::{aggregator, CounterType, DistributionType, GaugeValue, SetType, SetValue};
+use std::collections::BTreeMap;
+use std::fmt;
+use std::ops::Range;
+
+use crate::bucket::Bucket;
+use crate::BucketValue;
+
+/// The fraction of size passed to [`BucketsView::by_size()`] at which buckets will be split. A value of
+/// `2` means that all buckets smaller than half of max_flush_bytes will be moved in their entirety,
```
TBS1996:
```suggestion
/// `2` means that all buckets smaller than half of metrics_max_batch_size will be moved in their entirety,
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+        self.len() == 0
+    }
+
+    /// Selects a sub-view of the current view.
+    ///
+    /// # Panics
+    ///
+    /// Function panics when:
+    /// * the passed range is not contained in the current view.
+    /// * trying to split a counter or gauage bucket
```
TBS1996:
```suggestion
    /// * trying to split a counter or gauge bucket
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+    pub fn select(mut self, range: Range<usize>) -> Self {
+        assert!(
+            range.start >= self.range.start,
+            "range not contained in view"
+        );
+        assert!(range.end <= self.range.end, "range not contained in view");
+        assert!(
```
TBS1996: so we should try to avoid writing code that can panic, even if we've verified the input. Maybe use a normal result and log an error, and return None in the next() method?

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+        );
+    }
+
+    #[test]
+    fn test_bucket_view_select_invalid_counter() {
+        let bucket = Bucket::parse(b"b0:1|c", UnixTimestamp::from_secs(5000)).unwrap();
+
+        let select_fail = |range| {
+            let view = BucketView::new(&bucket);
+            std::panic::catch_unwind(|| view.select(range)).expect_err("expected to panic")
```
TBS1996: catch_unwind is pretty clever, I didn't see that elsewhere here. Although is it possible to suppress the output if the test is passing?

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+}
+
+/// A view into a slice of metric buckets.
+///
+/// The view can be used to iterate over a large slice
+/// of metric data slicing even into the buckets themselves.
+///
+/// ```txt
+///                    Full View
+///  /---------------------------------------------\
```
TBS1996, line 39: thanks for taking the time to draw this!

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+        }
+        // We are way past, including sub-bucket offset.
+        if self.current.slice > self.end.slice {
+            return None;
+        }
+
+        // This doesn't overflow because the last bucket in the inner slice will always have a 0 bucket index.
+        let next = &self.inner[self.current.slice];
+
+        // Choose the bucket end, this will always the full bucket except if it is the last.
```
TBS1996:
```suggestion
        // Choose the bucket end, this will always be the full bucket except if it is the last.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+        let end = match self.current.slice == self.end.slice {
+            false => next.value.len(),
+            true => self.end.bucket,
+        };
+
+        let next = BucketView::new(next).select(self.current.bucket..end);
+
+        // Even if the current Bucket was partial, the next one will be full,
+        // except it is the last one.
```
TBS1996:
```suggestion
        // except if it is the last one.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+            if (self.current.slice > self.end.slice)
+                || (self.current.slice == self.end.slice && self.end.bucket == 0)
+            {
+                break;
+            }
+
+            // Select next potential bucket,
+            // this won't overflow because `end` will never go past the slice and
+            // we just validated that current is constrained by end.
+            let bucket = &self.inner[self.current.slice];
```
TBS1996: could try to handle overflow here as well even though it's supposed to never happen.

relay-server/src/actors/processor.rs
TBS1996: yes I meant the Item struct, but nevermind, it was just a thought

TBS1996:
I think the PR is looking very good, there's a lot of code that looks a lot cleaner now. It's difficult to have everything in my working memory though, which makes it hard to approve it. Btw, it's helpful if you leave some comments on the github diff making some explanations, such as which functions you've changed and which you've just moved,  or if you've both moved a variable and renamed it, it's nice to have a small note about that. It's not really necessary in smaller PRs but here it would have been very helpful. 

I'll probably approve soon, in the meantime you could ping someone else to take a look too, there should hopefully be more than 1 reviewer

relay-server/src/actors/envelopes.rs
TBS1996: understandable. Although maybe in the future we can look at it again, I think there's a bit too much 'ping-pong' between services in relay in general.

relay-server/src/actors/processor.rs
```diff
@@ -3056,14 +3085,64 @@ impl EnvelopeProcessorService {
...
     fn handle_message(&self, message: EnvelopeProcessor) {
         match message {
             EnvelopeProcessor::ProcessEnvelope(message) => self.handle_process_envelope(*message),
             EnvelopeProcessor::ProcessMetrics(message) => self.handle_process_metrics(*message),
             EnvelopeProcessor::EncodeEnvelope(message) => self.handle_encode_envelope(*message),
+            EnvelopeProcessor::EncodeMetrics(message) => self.handle_encode_metrics(*message),
             #[cfg(feature = "processing")]
-            EnvelopeProcessor::RateLimitFlushBuckets(message) => {
-                self.handle_rate_limit_flush_buckets(message);
+            EnvelopeProcessor::RateLimitBuckets(message) => {
```
iker-barriocanal, line 2798: Especially in such big PRs, I'd suggest avoiding unrelated changes. If the change is relevant, we can review it in another PR.

iker-barriocanal (changes requested):
This PR introduces some interesting logic, but I think it's trying to do too many things at once.

Slicing metrics without additional allocations is a good performance improvement. However:
- It introduces a case that may result in panics. We should not introduce panics and do our best effort to identify and mitigate them.
- It has a significant amount of additional complexity to fix a problem I'm not sure we have -- do we have a performance issue on metric bucket splitting?
- I believe the previous implementation produces full metric buckets (but the last one), which seems to not be the case now. I'm not sure what the impact of this change is, but more buckets to send through the wire may have higher performance implications than splitting buckets in memory.
- The complexity (not just size) makes providing a meaningful review difficult.

Let me know if I'm missing or misunderstanding something.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 - Scrub all DB Core Data spans differently. ([#2686](https://github.com/getsentry/relay/pull/2686))
 - Support generic metrics extraction version 2. ([#2692](https://github.com/getsentry/relay/pull/2692))
 - Emit error on continued project config fetch failures after a time interval. ([#2700](https://github.com/getsentry/relay/pull/2700))
+- Partition and split metric buckets just before sending. ([#2682](https://github.com/getsentry/relay/pull/2682))
```
olksdr: Please, merge the `master` in and move this record to `Unreleased`.

relay-metrics/src/aggregator.rs
```diff
@@ -92,6 +96,28 @@ impl BucketKey {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct BucketKeyRef<'a> {
+    project_key: ProjectKey,
+    timestamp: UnixTimestamp,
+    metric_name: &'a str,
+    tags: &'a BTreeMap<String, String>,
+}
+
+impl<'a> BucketKeyRef<'a> {
+    // Create a 64-bit hash of the bucket key using FnvHasher.
```
olksdr:
```suggestion
    // Creates a 64-bit hash of the bucket key using FnvHasher.
```

olksdr: And I also would use the `///` for docs comments, even though this fn is private.

relay-metrics/src/aggregator.rs
```diff
@@ -900,6 +886,36 @@ impl fmt::Debug for Aggregator {
     }
 }
 
+/// Split buckets into N logical partitions, determined by the bucket key.
```
olksdr:
```suggestion
/// Splits buckets into N logical partitions, determined by the bucket key.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+
+/// The fraction of size passed to [`BucketsView::by_size()`] at which buckets will be split. A value of
+/// `2` means that all buckets smaller than half of max_flush_bytes will be moved in their entirety,
+/// and buckets larger will be split up.
+const BUCKET_SPLIT_FACTOR: usize = 32;
+
+/// The average size of values when serialized.
+const AVG_VALUE_SIZE: usize = 8;
+
+/// Just an internal type representing an index into a slice of buckets.
```
olksdr:
```suggestion
/// An internal type representing an index into a slice of buckets.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+/// Iterating over a [`BucketsView`] yields [`BucketView`] items,
+/// only the first and last elements may be partial buckets.
+///
+/// Using the above example, iterating over `View 1` yields the buckets:
+/// `[C:1], [C:12], [D:0, 1, 2, 3]`.
+pub struct BucketsView<'a> {
+    /// Source slice of buckets.
+    inner: &'a [Bucket],
+    /// Start index, slice index indicates bucket,
+    /// bucket index indicates offset in the selected bucket.
```
olksdr:
```suggestion
    /// Start index.
    /// - slice index indicates bucket,
    /// - bucket index indicates offset in the selected bucket.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+    start: Index,
+    /// End index, slice index indicates exclusive end,
+    /// bucket index, indicates offset into the *next* bucket past the end.
```
olksdr:
```suggestion
    /// End index.
    /// - slice index indicates exclusive end,
    /// - bucket index, indicates offset into the *next* bucket past the end.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+        }
+        len
+    }
+
+    /// Returns whether the view contains any buckets.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Iterater over all buckets in the view.
```
olksdr:
```suggestion
    /// Iterator over all buckets in the view.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+    /// Current index.
+    current: Index,
+    /// End index.
+    end: Index,
+}
+
+impl<'a> BucketsViewIter<'a> {
+    /// Creates a new iterator.
+    ///
+    /// **Notes:** if `start` and `end` are not valid indices the iterator may panic.
```
olksdr:
```suggestion
    /// # Panics
    /// 
    /// If `start` and `end` are not valid indices the iterator may panic.
```

Maybe we can add the proper `Panic` note on here, something like in the official rust docs used

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+/// ```txt
+///                    Full View
+///  /---------------------------------------------\
+/// [[C:1], [C:12], [D:0, 1, 2, 3, 5, 5], [S:42, 75]]
+///  \--------------------------/  \---------------/
+///            View 1                    View 2
+/// ```
+///
+/// Iterating over a [`BucketsView`] yields [`BucketView`] items,
+/// only the first and last elements may be partial buckets.
```
olksdr, line 46: Do not understand why the first element can be partial?

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+    /// Terminal position.
+    end: Index,
+    /// Maximum size of in bytes of each slice.
+    max_size_bytes: usize,
+}
+
+impl<'a> BucketsViewBySizeIter<'a> {
+    /// Creates a new iterator.
+    ///
+    /// **Notes:** if `start` and `end` are not valid indices the iterator may panic.
```
olksdr: Same here

```suggestion
    /// # Panics 
    ///
    /// If `start` and `end` are not valid indices the iterator may panic.
```

relay-metrics/src/view.rs
```diff
@@ -0,0 +1,1061 @@
...
+    /// The source bucket.
+    inner: &'a Bucket,
+    /// Non-empty and valid range into the bucket.
+    /// The full range is constrained by `0..bucket.value.len()`
+    range: Range<usize>,
+}
+
+impl<'a> BucketView<'a> {
+    /// Creates a new bucket view of a bucket.
+    /// The resulting view contains the entire bucket.
```
olksdr:
```suggestion
    /// Creates a new bucket view of a bucket.
    /// 
    /// The resulting view contains the entire bucket.
```

relay-metrics/src/view.rs
olksdr: It's a good idea to avoid panics altogether. Maybe return Option<Self> to indicate that provided range is bad or cannot be split , and issue the error.

relay-server/src/utils/rate_limits.rs
```diff
@@ -139,6 +139,16 @@ pub struct EnvelopeSummary {
     /// The number of monitor check-ins.
     pub checkin_quantity: usize,
 
+    /// Secondary number of of transactions.
```
```suggestion
    /// Secondary number of transactions.
```

olksdr (approved):
Overall looks good to me. 

I would suggest to avoid the code which could panic and I think we can good to go and test it in production.

relay-metrics/src/view.rs
Dav1dde: Discussed offline and I added another sentence to the docs explaining it

relay-server/src/actors/envelopes.rs
Dav1dde: I actually changed this now, thanks for bringing this up!

relay-metrics/src/aggregator.rs
Dav1dde: This is no longer an associated function/method, so it produces the same diff unfortunately

Dav1dde:
@iker-barriocanal sorry for the late response, some points adressed:

> This PR introduces some interesting logic, but I think it's trying to do too many things at once.

It's 2 things basically:

* Make the splits not clone data anymore, for this to work the next step must be Serialization to JSON. Serialization and processing should have always been in the envelope processor from the start for 2 reason:
  * Serialization is CPU heavy
  * Dropping the entire list of buckets is actually also quite CPU intensive
* Collecting outcomes for Metrics now. This is actually a late addition, because we realized merging back the buckets is not as easy anymore because we're sending full envelopes now and in order to merge back we would have to either keep the original metric buckets around or Deserialize the envelope again, either case is not desirable. The previous implementation had a future open until the envelope was sent for each batch (!). So this was unfortunately a necessity.

There is a way to keep the splitting in the aggregator by reference counting the split buckets, but I feel like this introduces a bigger mental overhead and also it does not logically belong to the aggregator.

> * It introduces a case that may result in panics. We should not introduce panics and do our best effort to identify and mitigate them.

Panics are gone now.

> * It has a significant amount of additional complexity to fix a problem I'm not sure we have -- do we have a performance issue on metric bucket splitting?

I do not have actual numbers on this, this all depends on the volume of data. So this means with less relay instances the problem actually becomes worse.

> * I believe the previous implementation produces full metric buckets (but the last one), which seems to not be the case now. I'm not sure what the impact of this change is, but more buckets to send through the wire may have higher performance implications than splitting buckets in memory.

I am not sure what you mean by this. The logic should have stayed the same. If you split on the last one, there will be also a split bucket at the start of the next batch.
If there is difference in logic I consider this a bug.

iker-barriocanal (approved):

relay-metrics/src/aggregator.rs
```diff
@@ -67,12 +67,16 @@ struct BucketKey {
 }
 
 impl BucketKey {
-    // Create a 64-bit hash of the bucket key using FnvHasher.
-    // This is used for partition key computation and statsd logging.
+    /// Creates a 64-bit hash of the bucket key using FnvHasher.
```
TBS1996, line 70:
```suggestion
    /// Creates a 64-bit hash of the bucket key using FnvHasher.
    ///
```

First line in docs should stand alone

relay-metrics/src/aggregator.rs
```diff
@@ -92,6 +96,28 @@ impl BucketKey {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct BucketKeyRef<'a> {
+    project_key: ProjectKey,
+    timestamp: UnixTimestamp,
+    metric_name: &'a str,
+    tags: &'a BTreeMap<String, String>,
+}
+
+impl<'a> BucketKeyRef<'a> {
+    /// Creates a 64-bit hash of the bucket key using FnvHasher.
```
TBS1996, line 113:
```suggestion
    /// Creates a 64-bit hash of the bucket key using FnvHasher.
    ///
```

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -34,6 +36,77 @@ pub struct MetricsLimiter<Q: AsRef<Vec<Quota>> = Vec<Quota>> {
...
+    };
+
+    let has_profile =
+        matches!(mri.name, "usage" | "duration") && metric.tag(PROFILE_TAG) == Some("true");
+
+    Some(TransactionCount { count, has_profile })
+}
+
+/// Wether to extract transaction and profile count based on the usage
+/// or duration metric.
```
TBS1996:
```suggestion
/// Wether to extract transaction and profile count based on the usage or duration metric.
```

might as well be one line

relay-server/src/actors/project.rs
```diff
@@ -1015,20 +1020,34 @@ impl Project {
...
+        };
+        let extraction_mode = ExtractionMode::from_usage(usage);
+
         if !buckets.is_empty() {
-            envelope_manager.send(SendMetrics {
+            envelope_processor.send(EncodeMetrics {
                 buckets,
                 scoping,
-                partition_key,
+                extraction_mode,
```
TBS1996, line 1140: nice job here cutting down a call to envelopemanager!

Dav1dde:
The roundtrip through the envelope manager is gone now. Metrics are sent directly to the processor instead.

TBS1996 (approved):

PR #2683, opened 2023-11-01
fix(profiling): Fix issues detected after logging better errors

phacops:
- We'll start accepting `null` values for the `release` field as we accept no field/empty string already and some SDKs are sending `null` values already.
- `measurements` parsing was done when validating the profile metadata and it's not necessary to do this at this stage so we moved it when we process the profile.
- We wanted to log the invalid value when trying to decode a number or float to a string and we ended up not taking `null` values into account. With this change, we'll log we tried to decode a `null` value, making it more explicit on what the error is.

#skip-changelog

### DIFF
--- a/relay-profiling/src/lib.rs
+++ b/relay-profiling/src/lib.rs
@@ -197,6 +197,7 @@ pub fn expand_profile(payload: &[u8], event: &Event) -> Result<(EventId, Vec<u8>
                 project_id = event.project.value().unwrap_or(&0),
                 sdk_name = event.sdk_name(),
                 sdk_version = event.sdk_version(),
+                transaction_id = format!("{:?}", event.id.value()),
                 "invalid profile",
             );
             return Err(ProfileError::InvalidJson(err));
@@ -226,6 +227,7 @@ pub fn expand_profile(payload: &[u8], event: &Event) -> Result<(EventId, Vec<u8>
                     project_id = event.project.value().unwrap_or(&0),
                     sdk_name = event.sdk_name(),
                     sdk_version = event.sdk_version(),
+                    transaction_id = format!("{:?}", event.id.value()),
                     "invalid profile",
                 );
                 Err(ProfileError::InvalidJson(err))
@@ -238,6 +240,7 @@ pub fn expand_profile(payload: &[u8], event: &Event) -> Result<(EventId, Vec<u8>
                     project_id = event.project.value().unwrap_or(&0),
                     sdk_name = event.sdk_name(),
                     sdk_version = event.sdk_version(),
+                    transaction_id = format!("{:?}", event.id.value()),
                     "invalid profile",
                 );
                 Err(err)
--- a/relay-profiling/src/measurements.rs
+++ b/relay-profiling/src/measurements.rs
@@ -73,6 +73,7 @@ mod tests {
         let measurement_json = r#"{"elapsed_since_start_ns":1234567890,"value":1e3}"#;
         let measurement = serde_json::from_str::<MeasurementValue>(measurement_json);
         assert!(measurement.is_ok());
+        assert_eq!(measurement.unwrap().value, 1e3f64);
     }
 
     #[test]
--- a/relay-profiling/src/measurements.rs
+++ b/relay-profiling/src/measurements.rs
@@ -47,9 +47,38 @@ mod tests {
     #[test]
     fn test_value_as_string() {
         let measurement_json = r#"{"elapsed_since_start_ns":1234567890,"value":"1234.56789"}"#;
-        assert!(serde_json::from_str::<MeasurementValue>(measurement_json).is_ok());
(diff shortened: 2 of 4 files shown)


relay-profiling/src/sample.rs
```diff
@@ -158,8 +158,8 @@ pub struct ProfileMetadata {
     platform: String,
     timestamp: DateTime<Utc>,
 
-    #[serde(default, skip_serializing_if = "String::is_empty")]
-    release: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    release: Option<String>,
```
iker-barriocanal: Does this mean sending an empty `""` release is now possible? If yes, we should set the release to `None`. We should also trim the release if we're not already doing it.

relay-profiling/src/measurements.rs
```diff
@@ -47,9 +47,38 @@ mod tests {
...
+        let measurement = serde_json::from_str::<MeasurementValue>(measurement_json);
+        assert!(measurement.is_ok());
+        assert_eq!(measurement.unwrap().value, f64::INFINITY);
+    }
+
+    #[test]
+    fn test_value_as_float_scientific_notation() {
+        let measurement_json = r#"{"elapsed_since_start_ns":1234567890,"value":1e3}"#;
+        let measurement = serde_json::from_str::<MeasurementValue>(measurement_json);
+        assert!(measurement.is_ok());
```
iker-barriocanal, line 75: Should we verify that the value is correct in this test too?

relay-profiling/src/sample.rs
```diff
@@ -168,9 +168,6 @@ pub struct ProfileMetadata {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     transaction: Option<TransactionMetadata>,
 
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    measurements: Option<HashMap<String, Measurement>>,
```
iker-barriocanal, line 172: Does this moves where `measurements` are in the protocol? Is this a breaking change?

relay-profiling/src/sample.rs
phacops: It does mean that it is now possible. In practice, it'd be dropped later anyway. It seemed easier than to roll out a custom function to check the option not being `None` or the value. I'll change it.

relay-profiling/src/sample.rs
phacops: It doesn't, it just changes when we parse them. This function is meant to parse metadata, so light parsing of most of the fields to check we should go further instead of dropping the profile.

I consider measurements to be data, so I'd want to parse them later, on a processing relay. It might not appear to make a ton of differences but it'll allow us to be able to show the SDK name/version in case there's malformed measurements data and will lower the load on pops, however slightly.

The protocol doesn't change, it stays where it is in the structure: before, it was part of the `ProfileMetadata` struct which was flatten, now it's part of the overall struct right away (which is only parsed when we expand the profile, not on pops).

relay-profiling/src/utils.rs
```diff
@@ -16,28 +16,34 @@ where
...
             "unsupported value: {:?}",
             v
         ))),
+        AnyType::Null => Err(serde::de::Error::custom("unsupported null value")),
     }
 }
+
+pub fn string_is_null_or_empty(s: &Option<String>) -> bool {
+    s.as_ref().unwrap_or(&"".to_string()).is_empty()
+}
```
iker-barriocanal: This avoids allocating an extra string:
```suggestion
pub fn string_is_null_or_empty(s: &Option<String>) -> bool {
    s.as_ref().map(|s| s.is_empty()).unwrap_or(true)
}
```
Generally, it's more convenient to work with `&Option<>` vs `Option<&>`, so updating the arg would be great!

iker-barriocanal (approved):

relay-profiling/src/utils.rs
Dav1dde: This can be nicely expressed with `map_or`:

```rs
pub fn string_is_null_or_empty(s: &Option<String>) -> bool {
    s.as_deref().map_or(true, |s| s.is_empty())
}
```

relay-profiling/src/lib.rs
```diff
@@ -197,6 +197,7 @@ pub fn expand_profile(payload: &[u8], event: &Event) -> Result<(EventId, Vec<u8>
                 project_id = event.project.value().unwrap_or(&0),
                 sdk_name = event.sdk_name(),
                 sdk_version = event.sdk_version(),
+                transaction_id = format!("{:?}", event.id.value()),
```
Dav1dde, line 200: I think this can be expressed as:

```
transaction_id = ?event.id.value(),
```

Dav1dde: Fixed in a separate PR

PR #2685, opened 2023-11-02
feat(profiling): Add mixed profiles (profile event V2)

krystofwoldrich:
- context https://github.com/getsentry/sentry-react-native/issues/2668

This PR drafts a structure of mixed profiles which enable us to ingest for example Mixed React Native and Android profiles.

### DIFF
--- a/relay-profiling/src/profile_version.rs
+++ b/relay-profiling/src/profile_version.rs
@@ -1,0 +1,11 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default, PartialEq, Eq)]
+pub enum Version {
+    #[default]
+    Unknown,
+    #[serde(rename = "1")]
+    V1,
+    #[serde(rename = "2")]
+    V2,
+}
--- a/relay-profiling/src/android.rs
+++ b/relay-profiling/src/android.rs
@@ -81,7 +81,7 @@ pub struct ProfileMetadata {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-struct AndroidProfile {
+pub struct AndroidProfile {
     #[serde(flatten)]
     metadata: ProfileMetadata,
 
--- a/relay-profiling/src/lib.rs
+++ b/relay-profiling/src/lib.rs
@@ -110,8 +110,11 @@ mod android;
 mod error;
 mod extract_from_transaction;
 mod measurements;
+mod mixed;
 mod native_debug_image;
 mod outcomes;
+mod profile_metadata;
+mod profile_version;
 mod sample;
 mod transaction_metadata;
 mod utils;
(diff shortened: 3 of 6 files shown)


relay-profiling/src/profile_version.rs
```diff
@@ -0,0 +1,11 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default, PartialEq, Eq)]
+pub enum Version {
```
phacops, line 4: Let's keep a specific `Version` enum for each type of profile, inside their respective module.

phacops (changes requested):
I don't think creating a mixed format like this is the right approach. We could add a field to the current Android format to make it backwards compatible (we wouldn't need any change to the pipeline if the field is empty).

You can still reuse some elements of the sample format (like the `Profile` struct) but besides that, let's keep things separate.

krystofwoldrich:
> we wouldn't need any change to the pipeline if the field is empty

I agree adding an extra field on the Android profile is a faster way to go, but it would work only for Android.

The mixed format would allow a general approach for Hybrid SDKs to pack multiple profiles. Compared to that adding a field to the Android profile works only on Android. (Mixed profiles could be used for iOS, Hermes...)

Adding a field on the Android profile moves part of the complexity to SDKs, where the assembled event will change metadata based on the content. (Hermes -> Profiles V1, Hermes + iOS -> Profiles V1, Hermes + Android -> Android Profiles "V0")

The SDK should send one format of metadata no matter the actual profiling data. If needed to reduce the complexity we can change the format in Relay.

I'll adjust the PR, the mentioned points can be addressed in the future.

phacops:
Do we actually have to create another format for sample profilers? Why wouldn't we add the samples/stacks/frames the way we have the schema already, mixed in with the rest with adding the proper platform field to frames? I'm not convinced we need a new mixed schema.

Android has always been treated differently because of the binary format it sends us not fitting anything else. For the rest, it seems like the sample format we created fits our needs, even with RN frames and samples mixed in with native frames and samples.

krystofwoldrich:
The current format works for mixed profiles, but we mixed them on a device at the moment. We also convert RN Hermes profiles on a device. It's still a time-consuming task, but faster than parsing the Android binary trace.

I'm not sure about adding `rnProfile` property to the Android profile is:
- The RN SDK controls the Transactions and Profiles.
- The top-level platform still should be `javascript` not `android`.
- The Android Metadata is different from the V1 profiles. (The RN SDK could create the Android format of metadata, but it seems more like a workaround)

But as we would like to have our own Android profiler in the future which will produce our format, which will be simpler to process and merge on a device, maybe the workaround is acceptable in this case.

krystofwoldrich:
Closing this in favor of simpler RN/Android specific format -> https://github.com/getsentry/relay/pull/2706

PR #2687, opened 2023-11-03
feat(py): Add validate_pii_selector

Dav1dde:
Expose a function to validate PII selectors directly, so we can use it to validate safe fields in Sentry.

### DIFF
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add `validate_pii_selector` to validate safe fields. ([#2687](https://github.com/getsentry/relay/pull/2687))
+
 ## 0.8.34
 
 - Add context for NEL (Network Error Logging) reports to the event schema. ([#2421](https://github.com/getsentry/relay/pull/2421))
--- a/py/sentry_relay/processing.py
+++ b/py/sentry_relay/processing.py
@@ -21,6 +21,7 @@ __all__ = [
     "is_glob_match",
     "is_codeowners_path_match",
     "parse_release",
+    "validate_pii_selector",
     "validate_pii_config",
     "convert_datascrubbing_config",
     "pii_strip_event",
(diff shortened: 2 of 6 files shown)


py/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add `validate_pii_selector_spec` to validate safe fields. ([#2687](https://github.com/getsentry/relay/pull/2687))
```
iker-barriocanal: This PR exposes a function in `relay-cabi`. Could we also update Relay's changelog?

py/tests/test_processing.py
```diff
@@ -128,6 +128,18 @@ def test_normalize_user_agent(must_normalize):
         assert "contexts" not in event
 
 
+def test_validate_pii_selector_spec():
+    sentry_relay.validate_pii_selector_spec("test")
+    sentry_relay.validate_pii_selector_spec("$user.id")
+    sentry_relay.validate_pii_selector_spec("extra.'sys.argv'.**")
+
+    with pytest.raises(ValueError):
```
iker-barriocanal: Does the error come with an error message? If yes: is it actionable, or does it help identifying why the selector is not valid?

Dav1dde: The error messages are good, I've added some assertions to the python tests

py/tests/test_processing.py
```diff
@@ -128,6 +128,24 @@ def test_normalize_user_agent(must_normalize):
...
 
 
+def test_validate_pii_selector_spec():
+    sentry_relay.validate_pii_selector_spec("test")
+    sentry_relay.validate_pii_selector_spec("$user.id")
+    sentry_relay.validate_pii_selector_spec("extra.'sys.argv'.**")
+
+    with pytest.raises(ValueError) as e:
+        sentry_relay.validate_pii_selector_spec("no_spaces allowed")
+    assert "1 | no_spaces allowed" in str(e.value)
```
iker-barriocanal: Q: any reason to not assert the whole error message?

iker-barriocanal (approved):

py/sentry_relay/processing.py
```diff
@@ -166,6 +167,17 @@ def is_codeowners_path_match(value, pattern):
     )
 
 
+def validate_pii_selector_spec(selector):
```
jjbayer: nit: Should we call this `validate_pii_selector`? As a user of this function I would not know what the `_spec` suffix refers to.

jjbayer (approved):

py/sentry_relay/processing.py
olksdr: Agreed with @jjbayer

olksdr (approved):

py/tests/test_processing.py
Dav1dde: I changed the error message to something more concise now and I am also asserting the full message now

py/sentry_relay/processing.py
Dav1dde: Renamed

PR #2688, opened 2023-11-03
feat(spans): Track when each span was received

phacops:
In order to track the end to end latency of span ingestion, we need the time when the span was received by Relay.

Related PRs, in order they should merged in:
- https://github.com/getsentry/sentry-kafka-schemas/pull/189
- https://github.com/getsentry/sentry/pull/59360
- https://github.com/getsentry/snuba/pull/4979


nikhars (approved):

relay-event-schema/src/protocol/span.rs
```diff
@@ -73,6 +73,9 @@ pub struct Span {
     /// Tags generated by Relay. These tags are a superset of the tags set on span metrics.
     pub sentry_tags: Annotated<Object<String>>,
 
+    /// Timestamp when the event has been received by Sentry.
```
jjbayer:
```suggestion
    /// Timestamp when the span has been received by Sentry.
```

jjbayer (approved):

PR #2689, opened 2023-11-06
feat(server): Disable graphql scrubbing when datascrubbing is disabled

### LINKED ISSUE #2666: Graphql Scrubbing can't be disabled via the Data Scrubber toggle option
### Steps to Reproduce
1. Disable the Data Scrubber in the project settings.
2. Send an event with a Graphql request body.
3.  See variables still scrubbed like the image in this [PR](https://github.com/getsentry/sentry-laravel/pull/720)
Specifically, I believe it is this[ rule in the repo](https://github.com/getsentry/relay/blob/75397010d18044a17759161e74475297fe9b0493/relay-pii/src/processor.rs#L218).

### Expected Result

Disabling the Data Scrubber should disable the scrubbing.

### Actual Result

Disabling the Data Scrubber does not disable the scrubbing.

TBS1996:
issue: #2666 

Previously the graphql scrubbing would run any time there is a pii config, which would always be the case even if datascrubbing is disabled as it would simply be a config without rules. 

The expected behaviour would be that if you disable datascrubbing, graphql scrubbing will also be disabled. 

this PR moves the graphql scrubbing outside of the processor, and simply checks the datascrubbing flag for whether to scrub graphql or not. Moving it out because we don't have access to the datascrubbing config in the processor. 

in the future, it would be better to have a separate graphql option for the user to configure in sentry and the graphql rules could be part of the normal datascrubbing rules.


relay-pii/src/processor.rs
```diff
@@ -200,23 +200,10 @@ impl<'a> Processor for PiiProcessor<'a> {
         replay.process_child_values(self, state)?;
         Ok(())
     }
-
-    fn process_event(
```
TBS1996, line 204: Since removing scrub_graphl makes it match the default implementation we can simply remove this

relay-pii/src/processor.rs
```diff
@@ -200,23 +200,10 @@ impl<'a> Processor for PiiProcessor<'a> {
...
-
-        event.process_child_values(self, state)?;
-
-        Ok(())
-    }
 }
 
 /// Scrubs GraphQL variables from the event.
-fn scrub_graphql(event: &mut Event) {
+pub fn scrub_graphql(event: &mut Event) {
```
TBS1996, line 206: Since it's not used in the processor directly we might consider moving it elsewhere, thoughts?

relay-server/src/actors/processor.rs
```diff
@@ -2176,6 +2176,12 @@ impl EnvelopeProcessorService {
         let event = &mut state.event;
         let config = &state.project_state.config;
 
+        if config.datascrubbing_settings.scrub_data {
+            if let Some(event) = event.value_mut() {
+                scrub_graphql(event);
+            }
+        }
```
jjbayer, line 2228: So the reason for moving this here is that inside the processor, we do not have access to `datascrubbing_settings`? Might be worth pointing that out in the PR description.

jjbayer (approved):

PR #2693, opened 2023-11-07
test(normalization): Reduce transaction processor test scope

iker-barriocanal:
The transaction processor tests rely on asserting entire event payloads to validate their goal. This does not only go against [relay coding guidelines (internal)](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58), but they also may fail when surrounding logic is updated, even if the logic the test tries to validate remains correct. As part of the removal of the transaction processor, more logic will run for the processing of an event, generating a different output. This PR updates the impacted tests to check specific values instead of complete snapshots.

#skip-changelog


relay-event-normalization/src/transactions/processor.rs
```diff
@@ -2181,7 +1860,10 @@ mod tests {
             scrub_identifiers(&mut s).unwrap();
             s.0.unwrap()
         });
-        assert_debug_snapshot!(replaced);
+        assert_eq!(
+            replaced,
+            ["/*", "/*", "/*", "/test/*/url",].map(str::to_owned)
+        )
```
iker-barriocanal, line 1866: This test is not really required to be updated. I decided to include it in this PR anyway, as it's the only last test remaining with a snapshot and it's very simple.

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -1644,132 +1497,47 @@ mod tests {
...
-                    "/foo/*/**",
-                    "s"
-                  ]
-                ],
-                "val": "/foo/rule-target/user/123/0/"
-              }
-            }
-          }
-        }
-        "#);
```
jjbayer, line 1772: This test and the one above are a bit hard to follow. The names of the tests changed, did the test objective also change?

jjbayer (commented):
Nice! One question, see below.

relay-event-normalization/src/transactions/processor.rs
iker-barriocanal: Good question! Forgot to add a comment on this. There are three things happening here:

1. I deleted `test_transaction_name_normalize_mark_as_sanitized_when_ready`. It used to test a behavior that was removed in https://github.com/getsentry/relay/commit/c057e370834049843932c3d11ac4d74850aaeb1d: there's no `ready` anymore so the test is redundant. This leaves `test_transaction_name_rename_with_rules`.
2. `test_transaction_name_rename_with_rules` does two things: run regular rules, and then run again with an expired rule. I split it into two tests, `test_transaction_name_rename_with_rules` and `test_transaction_name_rules_skip_expired`, each running a different use case.
3. Replace the snapshots with explicit assertion checks, just like the rest of the PR.

There are some tests in this module that are very similar. I think we could remove some tests by picking the right use cases to test, but that's out of the scope of the PR. I may get back to it in the future if this is still the case.

jjbayer (approved):

olksdr (approved):

PR #2695, opened 2023-11-07
chore(metric): Tag metric with the project state type

olksdr:
Add the `full` tag to `ProjectStateRequest` metric, to measure what types of the project states are  requested. 
This can be just temporary for the rollout procedure and once we sure that PoPs are fetching full project config, these changes can be reverted.  

related: https://github.com/getsentry/team-ingest/issues/231

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_upstream.rs
+++ b/relay-server/src/actors/project_upstream.rs
@@ -274,14 +274,18 @@ impl UpstreamProjectSourceService {
                     channels_batch.len() as u64
             );
 
+            let full_config = config.processing_enabled() || config.request_full_project_config();
             let query = GetProjectStates {
                 public_keys: channels_batch.keys().copied().collect(),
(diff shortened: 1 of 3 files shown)


iker-barriocanal (commented):
The code is correct for the PR description. However, I wonder if we should measure the responses instead, to discard scenarios where the full config is requested and a partial config is received. What do you think?

jjbayer (approved):

olksdr:
> The code is correct for the PR description. However, I wonder if we should measure the responses instead, to discard scenarios where the full config is requested and a partial config is received. What do you think?

 Also good idea, added additional counter to make sure we know what kind of project state will be returned by upstream, see [4034fbf](https://github.com/getsentry/relay/pull/2695/commits/4034fbf604963c0a6385df853d84d3eaf04135c5)

PR #2696, opened 2023-11-08
fix(spans): Add resource size tags

jjbayer:
Add the following data items to `sentry_tags`, so they can be queried from indexed spans:

* `http.response_content_length`
* `http.decoded_response_content_length`
* `http.response_transfer_size`

Downside: Because `sentry_tags` have to be strings, these numbers will be encoded as strings.

#skip-changelog


olksdr (approved):

DominikB2014 (approved):
👍

iker-barriocanal (approved):

iker-barriocanal (approved):
🚀

PR #2697, opened 2023-11-08
feat(server): Delay serving global config in managed mode before upstream response

### LINKED ISSUE #2584: Make global config required for processing
Before a global config is loaded / if a global config cannot be loaded during an incident, we currently fall back to using a default global config everywhere.

There are several use cases where if we actually want to rely on dynamic configuration, we have to be able to load it, e.g.

* When we move span metrics extraction config to sentry,
* when we move config for generic inbound filters to sentry.

## Solution

Just like envelopes get buffered until a valid project state is available, we should queue envelopes until a valid global config is available.

TBS1996:
## Problem
currently, before we request global config from upstream, we will simply serve the default global config. This kind of implicit behaviour is not ideal. 

## Solution
This PR will  affect the behaviour in managed mode, where we won't start processing envelopes until we have received a valid global config. To do this we hook into the existing buffer logic in projectcache where we buffer until we receive a projectstate for an envelope, and simply add global config as a second requirement. 

## Implementation

* we move the logic for getting global config from the envelopeprocessor to projectcache, and we include it in the message we send to envelopeprocessor for processing, which guarantees there's a valid global config at time of processing.
* When we receive an envelope, we check if we have a global config `handle_validate_envelope`, if not we buffer the envelope.  
* When we receive projectstates, this used to always trigger a dequeue from the buffer, but now if we don't have global config we stop dequeing and push the projectkey that would have been dequeued into a BTreeSet. 
* When we receive a global config, we request new projectstates for all of the projectkeys in the BTreeSet, which will trigger a new dequeue when their projectstates arrive, this time it should dequeue successfully as we have a global config.

The reason we don't dequeue directly when we receive the global config is because the projectstates might have expired by the time the global config arrives. This might lead to some redundant fetching of projectstates but i figured it would be rather insignificant and it will only happen once, the added complexity of checking was not worth it imo. If you disagree, let me know!


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - Add TTID and TTFD tags to mobile spans. ([#2662](https://github.com/getsentry/relay/pull/2662))
 - Scrub all DB Core Data spans differently. ([#2686](https://github.com/getsentry/relay/pull/2686))
 - Support generic metrics extraction version 2. ([#2692](https://github.com/getsentry/relay/pull/2692))
+- Don't serve default global config if waiting for global config from upstream. ([#2697](https://github.com/getsentry/relay/pull/2697))
```
```suggestion
- Postpone processing till the global config is available. ([#2697](https://github.com/getsentry/relay/pull/2697))
```

relay-server/src/actors/global_config.rs
```diff
@@ -105,32 +105,61 @@ pub struct Subscribe;
...
-        sender: relay_system::Sender<watch::Receiver<Arc<GlobalConfig>>>,
-    ) -> Self {
+    fn from_message(_: Subscribe, sender: relay_system::Sender<watch::Receiver<Status>>) -> Self {
         Self::Subscribe(sender)
     }
 }
 
+/// The status of global config serving of the global config service.
+///
+/// This is the message that will be sent to the services requesting a global config.
```
olksdr: But this is not a message, it's just a wrapper over the global config, isn't it? 

```suggestion
/// Describes the current fetching status of the [`GlobalConfig`] from the upstream.
```

relay-server/src/actors/global_config.rs
```diff
@@ -105,32 +105,61 @@ pub struct Subscribe;
...
+#[derive(Debug, Clone, Default)]
+pub enum Status {
+    /// Global config ready to be used by othe services.
```
olksdr:
```suggestion
    /// Global config ready to be used by other services.
```

relay-server/src/actors/global_config.rs
```diff
@@ -105,32 +105,61 @@ pub struct Subscribe;
...
+    ///
+    /// This variant implies different things in different circumstances. In managed mode, it means
+    /// that we have received a config from upstream. If not managed mode then this could either
+    /// be from file or the default global config.
```
olksdr:
```suggestion
    /// This variant implies different things in different circumstances. In managed mode, it means
    /// that we have received a config from upstream. In other modes the config is either
    /// from a file or the default global config.
```

relay-server/src/actors/global_config.rs
```diff
@@ -105,32 +105,61 @@ pub struct Subscribe;
...
+    Ready(Arc<GlobalConfig>),
+    /// We requested a global config from upstream but it has not yet arrived.
```
olksdr:
```suggestion
    /// The global config is requested from the upstream but it has not arrived yet.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -455,9 +461,57 @@ struct ProjectCacheBroker {
...
+#[derive(Debug)]
+enum GlobalConfigStatus {
+    /// Global config needed for envelope processing.
+    Ready(Arc<GlobalConfig>),
+    /// Project keys waiting for global config to dequeue from buffer.
+    ///
+    /// These project keys were used to try to dequeue the buffer, but were blocked because we
+    /// lacked a global config. When global config arrives, we will request project states for these keys again
+    /// which will trigger another dequeue that should be succesful. The reason we don't dequeue
+    /// directly when we receive globalconfig is that the projectstates might have expired in the meantime,
```
olksdr:
```suggestion
    /// directly when we receive [`GlobalConfig`] is that the [`ProjectState`]s might have expired in the meantime,
```

relay-server/src/actors/project_cache.rs
```diff
@@ -455,9 +461,57 @@ struct ProjectCacheBroker {
     /// Index of the buffered project keys.
     index: BTreeMap<ProjectKey, BTreeSet<QueueKey>>,
     buffer: Addr<Buffer>,
+    global_config: GlobalConfigStatus,
+}
+
+#[derive(Debug)]
+enum GlobalConfigStatus {
```
olksdr, line 472: Please, add more docs for this enum.

relay-server/src/actors/project_cache.rs
```diff
@@ -665,12 +725,20 @@ impl ProjectCacheBroker {
     /// The following pre-conditions must be met before calling this function:
     /// - Envelope's project state must be cached and valid.
     /// - If dynamic sampling key exists, the sampling project state must be cached and valid.
+    /// - Global config from the global config service must be available.
```
olksdr:
```suggestion
    /// - [`GlobalConfig`] from the [`GlobalConfigManager`] must be available.
```

relay-server/src/actors/project_cache.rs
```diff
@@ -696,23 +764,19 @@ impl ProjectCacheBroker {
         {
             let reservoir_counters = project.reservoir_counters();
 
-            let sampling_state = utils::get_sampling_key(managed_envelope.envelope())
+            let sampling_project_state = utils::get_sampling_key(managed_envelope.envelope())
                 .and_then(|key| self.projects.get(&key))
-                .and_then(|p| p.valid_state());
+                .and_then(|p| p.valid_state())
+                .filter(|state| state.organization_id == own_project_state.organization_id);
```
olksdr, line 769: Why was this changed?

relay-server/src/actors/project_cache.rs
```diff
@@ -899,20 +983,29 @@ impl Service for ProjectCacheService {
...
-                    Some(message) = rx.recv() => broker.handle_message(message),
-                    else => break,
+                                biased;
+
+                                    Ok(()) = subscription.changed() => {
+                                        match subscription.borrow().clone() {
+                                            global_config::Status::Ready(global_config) => broker.set_global_config(global_config),
+                                            // The watch should only be updated if it gets a new value.
+                                            // This would imply a logical bug.
+                                            global_config::Status::Waiting => relay_log::error!("global config updated with None value"),
```
olksdr:
```suggestion
                                            global_config::Status::Waiting => relay_log::error!("Still waiting for the global config"),
```

relay-server/src/endpoints/project_configs.rs
```diff
@@ -130,7 +130,13 @@ async fn inner(
     let mut configs = HashMap::with_capacity(keys_len);
     let mut pending = Vec::with_capacity(keys_len);
     let global_config = match inner.global {
-        true => Some(state.global_config().send(global_config::Get).await?),
+        true => Some(
+            state
+                .global_config()
+                .send(global_config::Get)
+                .await?
+                .get_ready_or_default(),
```
olksdr, line 138: So here we always actually return the default if the config currently unavailable?

relay-server/src/actors/project_cache.rs
```diff
@@ -899,20 +983,29 @@ impl Service for ProjectCacheService {
...
-                    Some(message) = state_rx.recv() => broker.merge_state(message),
-                    // Buffer will not dequeue the envelopes from the spool if there is not enough
-                    // permits in `BufferGuard` available. Currently this is 50%.
-                    Some(managed_envelope) = buffer_rx.recv() => broker.handle_processing(managed_envelope),
-                    _ = ticker.tick() => broker.evict_stale_project_caches(),
-                    Some(message) = rx.recv() => broker.handle_message(message),
-                    else => break,
+                                biased;
+
+                                    Ok(()) = subscription.changed() => {
```
olksdr: Why the formatting looks a bit off? 

And also wondering if we still want to be biased for processing the incoming messages first.

relay-server/src/actors/project_cache.rs
```diff
@@ -467,9 +521,15 @@ impl ProjectCacheBroker {
...
     ///
-    /// All the found envelopes will be send back through the `buffer_tx` channel and dirrectly
+    /// All the found envelopes will be send back through the `buffer_tx` channel and directly
     /// forwarded to `handle_processing`.
     pub fn dequeue(&mut self, partial_key: ProjectKey) {
+        // If we don't yet have the global config, we will defer dequeuing until we do.
+        if let GlobalConfigStatus::Pending(keys) = &mut self.global_config {
+            keys.insert(partial_key);
+            return;
+        }
```
olksdr, line 537: As I mentioned before, this potentially can be harmful , we can potentially accumulate a lot of data, in case we never can get the proper global config from the upstream.

olksdr (dismissed):
Overall approach looks ok to me so far. And it would be great to add actually the integration test, to check the exact behaviour of the service we expect to see - like we should not process anything till the global config is available and fetched by the Relay.

relay-server/src/actors/project_cache.rs
TBS1996: I think the code is way more clear this way, not related to PR but just following the principle of cleaning up things in the same module I'm working in as i go along.

relay-server/src/actors/project_cache.rs
TBS1996: i thought my rustformatter was broken so i'd wait a bit later until it does, but maybe it's not supposed to work within a macro ? I'll fix it manually now

relay-server/src/actors/project_cache.rs
TBS1996: I had that in mind when making the recent change as a set of projectkeys should take a lot less space than a vector of messages 

i dont think the memory should be a problem here though, we already have many growing vectors in the projectcachebroker, although maybe we can use the indexed keys to fetch new states when the global config arrives. 

we could sync on this in case we end up going back-and-forth a lot here

relay-server/src/endpoints/project_configs.rs
TBS1996: yes, not because i think it's good but because It's what we currently do and I wasn't sure to add another behaviour in the scope of this PR

TBS1996:
>Overall approach looks ok to me so far. And it would be great to add actually the integration test, to check the exact behaviour of the service we expect to see - like we should not process anything till the global config is available and fetched by the Relay.


I definitively agree we need integration test for this, I thought i'd wait until later but I can start right away since the purpose of the integration test is the behaviour not the implementation

relay-server/src/actors/global_config.rs
```diff
@@ -105,32 +105,59 @@ pub struct Subscribe;
...
-        sender: relay_system::Sender<watch::Receiver<Arc<GlobalConfig>>>,
-    ) -> Self {
+    fn from_message(_: Subscribe, sender: relay_system::Sender<watch::Receiver<Status>>) -> Self {
         Self::Subscribe(sender)
     }
 }
 
+/// Describes the current fetching status of the [`GlobalConfig`] from the upstream.
+#[derive(Debug, Clone, Default)]
+pub enum Status {
```
iker-barriocanal, line 133: Is an `Option` a good replacement for `Status`? If `None`, no global config is available; if `Some(Arc<GlobalConfig>)`, there's a global config available and ready to be used. I think an `Option` is simpler to understand than introducing a new enum.

relay-server/src/actors/project_cache.rs
```diff
@@ -899,19 +984,27 @@ impl Service for ProjectCacheService {
...
-                    else => break,
+                biased;
+                Ok(()) = subscription.changed() => {
+                    match subscription.borrow().clone() {
+                        global_config::Status::Ready(global_config) => broker.set_global_config(global_config),
+                        // The watch should only be updated if it gets a new value.
+                        // This would imply a logical bug.
+                        global_config::Status::Waiting => relay_log::error!("still waiting for the global config"),
+                        }
+                    },
```
iker-barriocanal, line 999: nit: could we align these `}` with their opening brackets?

relay-server/src/endpoints/project_configs.rs
iker-barriocanal: Does this mean that PoP relays can get a global config and process envelopes when they should not? A processing relay that hasn't fetched a global config from Sentry will send the default config to PoP relays, and PoPs will start processing. This is also the case for the following relays in the chain.

iker-barriocanal (commented):
Approach looks good!

relay-server/src/actors/global_config.rs
TBS1996: Personally I find custom enum like this more easy to understand, i dont think it's obvious that a None should eventually become a Some. 

But if most of team agree with you then i'll change it to option

relay-server/src/endpoints/project_configs.rs
TBS1996: For others: me and iker synced and concluded that it's not a problem in practice because we deploy processing before pops with a significant time delay, and by that time processing should have a global config ready.

please others come with feedback about this

relay-server/src/actors/project_cache.rs
TBS1996: yeah fixed, im not a very good rustformatter X)

relay-server/src/actors/global_config.rs
```diff
@@ -231,9 +258,14 @@ impl GlobalConfigService {
...
                     Some(global_config) => {
+                        // Log the first time we receive a global config from upstream.
+                        if !self.global_config_watch.borrow().is_ready() {
+                            relay_log::info!("received global config from upstream");
+                        }
                         // Notifying subscribers only fails when there are no
                         // subscribers.
-                        self.global_config_watch.send(Arc::new(global_config)).ok();
+                        self.global_config_watch
+                            .send_replace(Status::Ready(Arc::new(global_config)));
```
TBS1996, line 266: we replace `send` with `send_replace` because in case theres not yet a receiver `send_replace` will populate the receiver when it gets constructed later.

relay-server/src/actors/processor.rs
```diff
@@ -319,6 +318,9 @@ struct ProcessEnvelopeState<'a> {
 
     /// Reservoir evaluator that we use for dynamic sampling.
     reservoir: ReservoirEvaluator<'a>,
+
+    /// Global config used for envelope processing.
+    global_config: Arc<GlobalConfig>,
```
TBS1996, line 323: moved global config from the envelopeprocessor to the processorstate

relay-server/src/actors/processor.rs
```diff
@@ -539,7 +542,6 @@ impl FromMessage<RateLimitBuckets> for EnvelopeProcessor {
 /// This service handles messages in a worker pool with configurable concurrency.
 #[derive(Clone)]
 pub struct EnvelopeProcessorService {
```
TBS1996, line 544: now that 'inner' is irrelevant, i'll remove it in following pr

tests/integration/test_query.py
```diff
@@ -140,9 +140,9 @@ def get_project_config():
     try:
         relay.send_event(42)
 
-        event = mini_sentry.captured_events.get(timeout=8).get_event()
+        event = mini_sentry.captured_events.get(timeout=12).get_event()
         assert event["logentry"] == {"formatted": "Hello, World!"}
-        assert retry_count == 2
+        assert retry_count == 4
```
TBS1996, line 145: I can look deeper into it but I believe it simply needs more retries because after the first 2 retries it have to wait for a bit before it gets the global config and can proceed

relay-server/src/actors/global_config.rs
jjbayer: I agree that explicit & documented enums make the code easier to understand.

relay-server/src/actors/global_config.rs
```diff
@@ -105,32 +105,59 @@ pub struct Subscribe;
...
+    ///
+    /// This variant implies different things in different circumstances. In managed mode, it means
+    /// that we have received a config from upstream. In other modes the config is either
+    /// from a file or the default global config.
+    Ready(Arc<GlobalConfig>),
+    /// The global config is requested from the upstream but it has not arrived yet.
+    ///
+    /// This variant should never be sent after the first `Ready` has occured.
+    #[default]
+    Waiting,
```
jjbayer: nit: Should we call this `Pending` to align with naming elsewhere in the code?

```suggestion
    Pending,
```

TBS1996: Agree about the pending name!

relay-server/src/actors/global_config.rs
```diff
@@ -231,9 +258,12 @@ impl GlobalConfigService {
                 let mut success = false;
                 match config.global {
                     Some(global_config) => {
-                        // Notifying subscribers only fails when there are no
-                        // subscribers.
```
iker-barriocanal, line 235: I find this comment useful and I'd suggest restoring it.

relay-server/src/actors/project_cache.rs
```diff
@@ -502,17 +508,13 @@ impl ProjectCacheBroker {
         }
     }
 
-    fn get_global_config(&self) -> Option<Arc<GlobalConfig>> {
+    fn global_config_value(&self) -> Option<Arc<GlobalConfig>> {
```
```suggestion
    fn global_config(&self) -> Option<Arc<GlobalConfig>> {
```
?

tests/integration/test_query.py
iker-barriocanal: Why do project configs need more retries with this PR?

tests/integration/test_envelope.py
```diff
@@ -579,3 +579,51 @@ def test_sample_rates_metrics(mini_sentry, relay_with_processing, events_consume
 
     event, _ = events_consumer.get_event()
     assert event["_metrics"]["sample_rates"] == sample_rates
+
+
+# Checks that we buffer envelopes until we have a global config in processing mode.
```
iker-barriocanal: Could we move this comment as a docstring of the test?

tests/integration/test_envelope.py
```diff
@@ -579,3 +579,51 @@ def test_sample_rates_metrics(mini_sentry, relay_with_processing, events_consume
...
+    include_global = False
+    original_endpoint = mini_sentry.app.view_functions["get_project_config"]
+
+    @mini_sentry.app.endpoint("get_project_config")
+    def get_project_config():
+        nonlocal include_global
+
+        res = original_endpoint().get_json()
+        if not include_global:
+            res.pop("global", None)
```
iker-barriocanal: What is the second argument, `None`, for?

tests/integration/test_envelope.py
```diff
@@ -579,3 +579,51 @@ def test_sample_rates_metrics(mini_sentry, relay_with_processing, events_consume
...
+        return res
+
+    mini_sentry.add_basic_project_config(42)
+    events_consumer = events_consumer()
+    options = {"cache": {"global_config_fetch_interval": 1}}
```
iker-barriocanal, line 605: If we set the value to `0`, does it speed the test up? Does this enable reducing the wait time at the bottom of the test?

tests/integration/test_envelope.py
```diff
@@ -579,3 +579,51 @@ def test_sample_rates_metrics(mini_sentry, relay_with_processing, events_consume
...
+    envelope.add_event({"message": "hello, world!"})
+    relay.send_envelope(42, envelope)
+
+    error_raised = False
+    try:
+        # Event is still in buffer because we have not provided a global config
+        events_consumer.get_event(timeout=1)
+    except AssertionError:
+        error_raised = True
+    assert error_raised
```
iker-barriocanal: See [`pytest.raises`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.raises).

It seems a bit fishy that we only wait 1s to check if there are events when there aren't supposed to be, and wait 10s when they are supposed to be. Could we be running into false negatives?

relay-server/src/actors/global_config.rs
TBS1996: we now use send_replace, which never fails, so the old doc is outdated

relay-server/src/actors/project_cache.rs
TBS1996: yeahh makes sense,  i went by [this](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html#method.continue_value) precedent but i think it doesnt apply in the same manner here actually.

tests/integration/test_envelope.py
TBS1996: copied from an example, seems i dont need, thx!

tests/integration/test_envelope.py
TBS1996: yeah thanks! 

i had forgotten to reduce the 10 sec interval after i reduced the fetch_interval, it would work with 2 sec timeout but now it works with 1 sec after putting the fetch_interval to 0 ^^

tests/integration/test_envelope.py
TBS1996: hmm, trying but it doesn't seem to work, using it with assertionerror because thats what gets logged in mini_sentry.test_failures:

>[('/api/666/envelope/', AssertionError('Relay sent us event: global config missing in upstream response')), ('/api/666/envelope/', AssertionError('Relay sent us event: global config missing in upstream response')), ('/api/666/envelope/', AssertionError('Relay sent us event: error fetching project states: upstream request returned error 500 Internal Server Error: no error details')), ('/api/666/envelope/', AssertionError('Relay sent us event: global config missing in upstream response'))]

tests/integration/test_query.py
TBS1996: we trigger another dequeue by asking for fresh projectstates

https://github.com/getsentry/relay/pull/2697/files#diff-ed683d4a7b32e65843545cca6a8d6acd8d9e6779fb4df8cd6153bf9b8b844eb9R476

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 - Validate span timestamps and IDs in light normalization on renormalization. ([#2679](https://github.com/getsentry/relay/pull/2679))
 - Scrub all DB Core Data spans differently. ([#2686](https://github.com/getsentry/relay/pull/2686))
 - Support generic metrics extraction version 2. ([#2692](https://github.com/getsentry/relay/pull/2692))
+- Postpone processing till the global config is available. ([#2697](https://github.com/getsentry/relay/pull/2697))
```
olksdr: Please, merge the `master` in and move this record to `unreleased` section.

relay-server/src/actors/project_cache.rs
```diff
@@ -455,9 +461,60 @@ struct ProjectCacheBroker {
     /// Index of the buffered project keys.
     index: BTreeMap<ProjectKey, BTreeSet<QueueKey>>,
     buffer: Addr<Buffer>,
+    global_config: GlobalConfigStatus,
+}
+
+/// Status of global config needed for envelope processing.
```
olksdr: Maybe something like this will be better. 

```suggestion
/// Describes the current status of the [`GlobalConfig`], either it's ready to be used or contains the list of in-flight project keys, to be processed once the config arrives. 
```

relay-server/src/endpoints/project_configs.rs
olksdr: The time delay does not matter, if the processing Relays cannot get the global config from upstream. 
When the PoPs are deployed while there is an incident with fetching global config, PoPs will get default one.

tests/integration/test_envelope.py
olksdr: Why this resolved and not addressed?

tests/integration/test_envelope.py
```diff
@@ -579,3 +579,51 @@ def test_sample_rates_metrics(mini_sentry, relay_with_processing, events_consume
...
+            res.pop("global", None)
+        return res
+
+    mini_sentry.add_basic_project_config(42)
+    events_consumer = events_consumer()
+    options = {"cache": {"global_config_fetch_interval": 1}}
+    relay = relay_with_processing(options=options)
+
+    envelope = Envelope()
+    envelope.add_event({"message": "hello, world!"})
```
olksdr: Please, send more than 1 envelope, 3 - 5 would do.

tests/integration/test_envelope.py
olksdr: Was this answered or addressed ?

tests/integration/test_envelope.py
```diff
@@ -579,3 +579,51 @@ def test_sample_rates_metrics(mini_sentry, relay_with_processing, events_consume
...
+    for err in mini_sentry.test_failures:
+        assert (
+            str(err[1])
+            == "Relay sent us event: global config missing in upstream response"
+        )
+    mini_sentry.test_failures.clear()
+
+    # Global configs are fetched in 10 second intervals, so the event should have come
+    # through after a 10 sec timeout.
+    events_consumer.get_event(timeout=10)
```
olksdr: Please, assert that after the global config is fetched we get all sent envelopes here.

relay-server/src/endpoints/project_configs.rs
TBS1996: youre right, its not ideal. I discussed this with @Dav1dde  who also pointed this out. We would have to communicate to the downstream relays that it's a global config they're receiving, or not send back at all. Backwards compatibility is an issue here.

relay-server/src/actors/project_cache.rs
TBS1996: i agree, thanks!

tests/integration/test_envelope.py
TBS1996: do you mean to do it in a more explicit manner? because afaict, calling get_event does assert that we got the envelope, since it will panic if not present

olksdr (commented):
Looks fine to me

You mentioned  you might want to wait on @jan-auer  to give a final review.  Otherwise I would carefully test it in prod.

One important issue here to handle is upstream Relays, to make sure they receive a correct global config. It can be done in the immediate followup PR. But  to be honest, I would make sure it's finished in this PR, so the feature is fully complete.

tests/integration/test_envelope.py
```diff
@@ -579,3 +579,51 @@ def test_sample_rates_metrics(mini_sentry, relay_with_processing, events_consume
...
+    options = {"cache": {"global_config_fetch_interval": 1}}
+    relay = relay_with_processing(options=options)
+
+    envelope_qty = 5
+    for _ in range(envelope_qty):
+        envelope = Envelope()
+        envelope.add_event({"message": "hello, world!"})
+        relay.send_envelope(42, envelope)
+
+    events_consumer.assert_empty()
```
iker-barriocanal: IIUC, this runs without any time delay. I have the same concern as in https://github.com/getsentry/relay/pull/2697#discussion_r1394280536.

iker-barriocanal (commented):
Code LGTM, one open question before approving

tests/integration/test_envelope.py
Dav1dde: iirc it has an internal 1 second timeout

relay-server/src/endpoints/project_configs.rs
Dav1dde: I think this still mirrors current behaviour, we can tackle this in a separate issue imo. The possible solutions are actually quite tricky and depends on wether we want to change the API or not.

tests/integration/test_envelope.py
Dav1dde: Actually I am wrong, it does **not** have an internal timeout of 1 second. 🚨

TBS1996: yeah good catch! I added 2 sec sleep now

tests/integration/test_envelope.py
TBS1996: @olksdr yeah we now wait when asserting empty here

TBS1996:
> Looks fine to me
> 
> You mentioned you might want to wait on @jan-auer to give a final review. Otherwise I would carefully test it in prod.
> 
> One important issue here to handle is upstream Relays, to make sure they receive a correct global config. It can be done in the immediate followup PR. But to be honest, I would make sure it's finished in this PR, so the feature is fully complete.

do you mean the issue of them potentially getting a default global config and thinking its from sentry? how to handle that case isnt obvious, so we should probably discuss this together properly before doing anything here

tests/integration/test_envelope.py
```diff
@@ -579,3 +580,52 @@ def test_sample_rates_metrics(mini_sentry, relay_with_processing, events_consume
...
+        assert (
+            str(err[1])
+            == "Relay sent us event: global config missing in upstream response"
+        )
+    mini_sentry.test_failures.clear()
+
+    # Check that we received exactly {envelope_qty} envelopes.
+    for _ in range(envelope_qty):
+        events_consumer.get_event(timeout=2)
+    events_consumer.assert_empty()
```
TBS1996, line 629: this one doesn't require sleeping first, since we assert that it's not empty in the previous lines, and we just wanna check that we take out exactly `envelope_qty` envelopes

tests/integration/test_envelope.py
Dav1dde: @TBS1996 `assert_empty` takes a timeout parameter, just use that

TBS1996: that makes sense, thx! changed it

olksdr (approved):
Good to test.

The proper propagation of the global config to PoPs can be handled in the followups.

tests/integration/test_envelope.py
```diff
@@ -597,7 +597,7 @@ def get_project_config():
 
         res = original_endpoint().get_json()
         if not include_global:
-            res.pop("global")
+            res["global"] = None
```
iker-barriocanal, line 600: What's the reason for this change?

This means `global` will be a key with a `None` value. The previous implementation, and Sentry's behavior, is not to include `global`. As a result, this test emulates a different behavior from what Sentry does.

PR #2700, opened 2023-11-08
ref(project-upstream): Emit error on multiple project fetch failures

iker-barriocanal:
This PR makes Relay emit an error after continued project config fetch failures for an interval. Currently, Relay emits some errors on some specific failure cases, but there's a metric to tell whether it's failing all project config fetches or just a few of them. The interval can be configured with `http.project_failure_interval`.

## Approach

Relay tracks the time of a failed fetch, and resets it on a successful one. If some time elapses and Relay hasn't reset the time, it emits an error for each failed request. This should create a big spike of errors after a continued time of failed fetches, demanding attention to the issue.

The fetch time is represented as `Option<Instant>`. Without the option, resetting the instant means setting the most recent time. In that case, Relay would emit an error if a fetch fails after some time of not emitting any fetches at all. This scenario is more likely to happen in low-volume environments, like some Self-Hosted instances.

### Default interval

I've set an arbitrary default value of 90 seconds. This should be short enough to determine Relay is having issues without false positives, and long enough to retry the same request twice with default values (`max_retry_interval = 60s`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Add TTID and TTFD tags to mobile spans. ([#2662](https://github.com/getsentry/relay/pull/2662))
 - Scrub all DB Core Data spans differently. ([#2686](https://github.com/getsentry/relay/pull/2686))
 - Support generic metrics extraction version 2. ([#2692](https://github.com/getsentry/relay/pull/2692))
+- Emit error on continued project config fetch failures for an interval. ([#2700](https://github.com/getsentry/relay/pull/2700))
 
 ## 23.10.1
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 - Add TTID and TTFD tags to mobile spans. ([#2662](https://github.com/getsentry/relay/pull/2662))
 - Scrub all DB Core Data spans differently. ([#2686](https://github.com/getsentry/relay/pull/2686))
 - Support generic metrics extraction version 2. ([#2692](https://github.com/getsentry/relay/pull/2692))
+- Emit error on continued project config fetch failures for an interval. ([#2700](https://github.com/getsentry/relay/pull/2700))
```
TBS1996:
```suggestion
- Emit error on continued project config fetch failures after a time interval. ([#2700](https://github.com/getsentry/relay/pull/2700))
```

I think this makes more sense ? 🤔

TBS1996 (approved):
looks good to me, seems very sensible

PR #2701, opened 2023-11-08
fix(pii): Removes legacy special behaviour from safe fields parsing

Dav1dde:
safe fields parsing accidentally triggered legacy behaviour which in some cases could disable data scrubbing for all strings

### DIFF
--- a/relay-pii/src/selector.rs
+++ b/relay-pii/src/selector.rs
@@ -165,6 +165,19 @@ pub enum SelectorSpec {
 }
 
 impl SelectorSpec {
+    /// Parses a selector from a string without legacy special handling.
+    pub fn parse_non_legacy(s: &str) -> Result<SelectorSpec, InvalidSelectorError> {
+        handle_selector(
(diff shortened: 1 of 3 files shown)


relay-pii/src/selector.rs
```diff
@@ -165,6 +165,19 @@ pub enum SelectorSpec {
...
+    /// Parses a selector from a string without legacy special handling.
+    pub fn parse_non_legacy(s: &str) -> Result<SelectorSpec, InvalidSelectorError> {
+        handle_selector(
+            SelectorParser::parse(Rule::RootSelector, s)
+                .map_err(|e| InvalidSelectorError::ParseError(Box::new(e)))?
+                .next()
+                .unwrap()
+                .into_inner()
+                .next()
+                .unwrap(),
```
jjbayer, line 177: In a next step we should try to get rid of these unwraps.

jjbayer (approved):

PR #2702, opened 2023-11-08
ref(project-upstream): Extract IDs from sentry errors

iker-barriocanal:
Very similar to https://github.com/getsentry/relay/pull/2698. This PR moves the project key out of the error message in the project upstream, in order to reduce the scope of the problem in the current observability.

The diff aims to solve the following issues:
1. Sentry groups these issues together, having a different name for the issue description and error message. For example, you could look at a project key that didn't have issues for 1w, while the issue gets new errors from different project keys. To identify these keys, you need to go through each new event and inspect that error message.
2. It's not viable to measure the impacted project keys. With project keys as tags, it's possible to see the breakdown of project keys and see e.g. if there are just a few impacted projects.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_upstream.rs
+++ b/relay-server/src/actors/project_upstream.rs
@@ -218,7 +218,7 @@ impl UpstreamProjectSourceService {
                         result = "timeout",
                     );
                     relay_log::error!(
-                        project_key = id.as_str(),
+                        tags.project_key = id.as_str(),
                         "error fetching project state: deadline exceeded"
                     );
                 }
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/project_upstream.rs
```diff
@@ -358,7 +361,11 @@ impl UpstreamProjectSourceService {
                             .remove(&key)
                             .unwrap_or(ErrorBoundary::Ok(None))
                             .unwrap_or_else(|error| {
-                                relay_log::error!(error, "error fetching project state {key}");
+                                relay_log::error!(
+                                    tags.project_key = key.as_str(),
+                                    error,
+                                    "error fetching project state: deserialization error"
```
iker-barriocanal, line 367: I took the deserialization part from the docs of the `ProjectState::err` error created below:
https://github.com/getsentry/relay/blob/b33f76e4c9926b90cc9b56e0a6e645601e8c690d/relay-server/src/actors/project.rs#L139-L140

iker-barriocanal:
Main goal of the PR was addressed in https://github.com/getsentry/relay/commit/e296698581754ef25062e787764ce6beecefa3ab.

PR #2706, opened 2023-11-09
feat(profiling): Add optional js profile to Android profiles events

krystofwoldrich:
- context https://github.com/getsentry/sentry-react-native/issues/2668
- supersedes https://github.com/getsentry/relay/pull/2685

This PR drafts a structure of mixed profiles which enable us to ingest for example Mixed React Native and Android profiles.


krystofwoldrich:
The PR is ready for review. Haven't tested it locally.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - `normalize_performance_score` stores 0 to 1 cdf score instead of weighted score for each performance score component. ([#2734](https://github.com/getsentry/relay/pull/2734))
 - Add Bytespider (Bytedance) to web crawler filter. ([#2747](https://github.com/getsentry/relay/pull/2747))
+- Add Mixed JS/Android Profiles events processing. ([#2706](https://github.com/getsentry/relay/pull/2706))
```
olksdr: Please, merge `master` in and if needed move this record to `Unreleased` section.

relay-profiling/src/sample.rs
```diff
@@ -85,6 +85,34 @@ struct Profile {
 }
 
 impl Profile {
+    pub fn parse(&mut self, platform: &str, architecture: &str) -> Result<(), ProfileError> {
```
olksdr: The name of this function is maybe not really `parse` but more like `validate` or `normalize`.  

Also, please add docs.

relay-profiling/src/sample.rs
```diff
@@ -95,6 +123,107 @@ impl Profile {
             frame.strip_pointer_authentication_code(addr);
         }
     }
+
+    fn remove_idle_samples_at_the_edge(&mut self) {
+        let mut active_ranges: HashMap<u64, Range<usize>> = HashMap::new();
```
olksdr, line 136:
```suggestion
        let mut active_ranges = HashMap::new();
```
I think this should also work.

relay-profiling/src/sample.rs
```diff
@@ -95,6 +123,107 @@ impl Profile {
...
+                    false
+                }
+            })
+            .map(|(_, sample)| sample)
+            .collect();
+    }
+
+    /// Removes a sample when it's the only sample on its thread
+    fn remove_single_samples_per_thread(&mut self) {
+        let mut sample_count_by_thread_id: HashMap<u64, u32> = HashMap::new();
```
olksdr:
```suggestion
        let mut sample_count_by_thread_id = HashMap::new();
```

I think this should also work.

relay-profiling/src/sample.rs
```diff
@@ -95,6 +123,107 @@ impl Profile {
...
+        }
+
+        // Only keep data from threads with more than 1 sample so we can calculate a duration
+        sample_count_by_thread_id.retain(|_, count| *count > 1);
+
+        self.samples
+            .retain(|sample| sample_count_by_thread_id.contains_key(&sample.thread_id));
+    }
+
+    fn check_samples(&self) -> bool {
```
olksdr: Maybe other name will be better here
```suggestion
    fn has_valid_samples(&self) -> bool {
```

And, please, add docs on this function.

relay-profiling/src/sample.rs
```diff
@@ -95,6 +123,107 @@ impl Profile {
...
+        for sample in &self.samples {
+            if self.stacks.get(sample.stack_id).is_none() {
+                return false;
+            }
+        }
+        true
+    }
+
+    fn check_stacks(&self) -> bool {
```
olksdr:
```suggestion
    fn has_valid_stacks(&self) -> bool {
```

Please, add docs for this function

relay-profiling/src/sample.rs
```diff
@@ -95,6 +123,107 @@ impl Profile {
...
+        }
+        false
+    }
+
+    fn cleanup_thread_metadata(&mut self) {
+        if let Some(thread_metadata) = &mut self.thread_metadata {
+            let mut thread_ids: HashSet<String> = HashSet::new();
+            for sample in &self.samples {
+                thread_ids.insert(sample.thread_id.to_string());
+            }
```
olksdr:
```suggestion
        if let Some(thread_metadata) = &mut self.thread_metadata {
            let thread_ids = self
                .samples
                .iter()
                .map(|sample| sample.thread_id.to_string())
                .collect::<HashSet<_>>();

```

Maybe something like this will be a little bit better to read

relay-profiling/src/sample.rs
```diff
@@ -95,6 +123,107 @@ impl Profile {
...
+    }
+
+    fn cleanup_queue_metadata(&mut self) {
+        if let Some(queue_metadata) = &mut self.queue_metadata {
+            let mut queue_addresses: HashSet<&String> = HashSet::new();
+            for sample in &self.samples {
+                if let Some(queue_address) = &sample.queue_address {
+                    queue_addresses.insert(queue_address);
+                }
+            }
```
olksdr: Also here,

```suggestion
            let queue_addresses = self
                .samples
                .iter()
                .filter_map(|sample| sample.queue_address.as_ref())
                .collect::<HashSet<_>>();

```
What do you think?

olksdr (commented):
Thanks for the PR. Overall looks ok. I left few comments which I think should be addressed before we continue. 

It would be also a good idea to add more unit tests to see if the things you expect end up in the correct places.

relay-profiling/src/sample.rs
krystofwoldrich: Rust complains that the type must be known before hand.

relay-profiling/src/sample.rs
krystofwoldrich: It works but than value is typed as unknown. I would say having the type is a bit better.

relay-profiling/src/sample.rs
krystofwoldrich: I like that thanks.

krystofwoldrich:
@olksdr 
> It would be also a good idea to add more unit tests to see if the things you expect end up in the correct places.

This code is not new, I just moved it to the Profile. There are tests for the functions in https://github.com/getsentry/relay/blob/a0e342db495a419bf8993e6e8cd522ed36f0d225/relay-profiling/src/sample.rs#L411, or would you like to see more here?

relay-profiling/src/sample.rs
```diff
@@ -95,6 +126,110 @@ impl Profile {
...
+            })
+            .map(|(_, sample)| sample)
+            .collect();
+    }
+
+    /// Removes a sample when it's the only sample on its thread
+    fn remove_single_samples_per_thread(&mut self) {
+        let mut sample_count_by_thread_id: HashMap<u64, u32> = HashMap::new();
+
+        for sample in &self.samples {
```
Dav1dde: Looks like [`Itertools::count_by`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.count_by) (or

relay-profiling/src/sample.rs
```diff
@@ -95,6 +126,110 @@ impl Profile {
...
+            let is_active = match self.stacks.get(sample.stack_id) {
+                Some(stack) => !stack.is_empty(),
+                None => true,
+            };
+
+            if !is_active {
+                continue;
+            }
+
+            if let Some(range) = active_ranges.get_mut(&sample.thread_id) {
```
Dav1dde: Untested, but something like this should work:

```rs
active_ranges.entry(sample.thread_id).and_modify(|range| range.end += 1).or_insert(i..i+1)
```

relay-profiling/src/sample.rs
```diff
@@ -95,6 +126,110 @@ impl Profile {
...
+                active_ranges.insert(sample.thread_id, i..i + 1);
+            }
+        }
+
+        self.samples = self
+            .samples
+            .drain(..)
+            .enumerate()
+            .filter(|(i, sample)| {
+                if let Some(range) = active_ranges.get(&sample.thread_id) {
```
Dav1dde:
```rs
active_ranges.get(&sample.thread_id).map_or(false, |range| range.contains(i))
```

Actually thinking about it, this entire `filter` could be a `filter_map`, something like this (undecided if this is actually better, I feel like it is harder to understand for someone who isn't as familar with Rust):

```rs
.filter_map(|(i, sample)|
    active_ranges.get(&sample.thread_id)
        .map_or(false, |range| range.contains(i))
        .then_some(sample)
)
```

relay-profiling/src/sample.rs
Dav1dde: I agree with olek here, we should not do type annotations except when necessary. But see my other comment, this should be obsolete with `Itertools::count_by` anyways.

relay-profiling/src/sample.rs
```diff
@@ -95,6 +126,110 @@ impl Profile {
...
+        // Only keep data from threads with more than 1 sample so we can calculate a duration
+        sample_count_by_thread_id.retain(|_, count| *count > 1);
+
+        self.samples
+            .retain(|sample| sample_count_by_thread_id.contains_key(&sample.thread_id));
+    }
+
+    /// Checks that all stacks referenced by the samples exist in the stacks.
+    fn all_stacks_referenced_by_samples_exist(&self) -> bool {
+        for sample in &self.samples {
```
Dav1dde: When expressed as an iterator we can use the same logic as in the method name (`all`):

```rs
self.samples.iter().all(|sample| self.stacks.get(sample.stack_id).is_some())
```

relay-profiling/src/sample.rs
```diff
@@ -95,6 +126,110 @@ impl Profile {
...
+            if self.stacks.get(sample.stack_id).is_none() {
+                return false;
+            }
+        }
+        true
+    }
+
+    /// Checks that all frames referenced by the stacks exist in the frames.
+    fn all_frames_referenced_by_stacks_exist(&self) -> bool {
+        for stack in &self.stacks {
```
Dav1dde: Can also be expressed as an iterator as above

relay-profiling/src/android.rs
```diff
@@ -7,6 +7,8 @@ use relay_event_schema::protocol::EventId;
 use serde::{Deserialize, Serialize};
 
 use crate::measurements::Measurement;
+use crate::native_debug_image::NativeDebugImage;
+use crate::sample::Profile as SampleProfile;
```
Dav1dde: This import is really confusing, with there being a closely related type called `SampleProfile`

Dav1dde (commented):
I am commenting on a lot of code that was moved (only saw that after), but maybe now is a good time to clean this up a bit.

relay-profiling/src/android.rs
```diff
@@ -221,6 +238,17 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_roundtrip_react_native() {
```
iker-barriocanal, line 242: This test validates the parsing doesn't fail, but we don't know if the contents are what we expect. What do you think about validating that too?

relay-profiling/src/sample.rs
```diff
@@ -85,6 +85,37 @@ struct Profile {
 }
 
 impl Profile {
+    /// Mutates the profile. Removes invalid samples and threads.
```
iker-barriocanal, line 91: Could we add a first short line describing the function and then the rest of the explanation, to match [relay coding guidelines (internal)](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#987c0416a9f54674999eed59eecbf846)?

relay-profiling/src/sample.rs
```diff
@@ -95,6 +126,110 @@ impl Profile {
...
+
+        for (i, sample) in self.samples.iter().enumerate() {
+            let is_active = match self.stacks.get(sample.stack_id) {
+                Some(stack) => !stack.is_empty(),
+                None => true,
+            };
+
+            if !is_active {
+                continue;
+            }
```
iker-barriocanal: IIUC this skips the sample when there's an empty stack. Could we simplify it to something like the following?

```rs
if self.stacks.get(sample.stack_id).unwrap_or(Vec::new()).is_empty() {
  continue;
}
```
(haven't tested if it compiles)

relay-profiling/tests/fixtures/profiles/android/roundtrip.rn.json
iker-barriocanal, line 1: The sample profile fixture seems a bit too excessive with 969KB, is this the minimal profile we can use?

relay-profiling/src/sample.rs
krystofwoldrich: Yes, can be `.counts_by(|sample| sample.thread_id)`

relay-profiling/src/sample.rs
krystofwoldrich: with small edit works :) thanks

relay-profiling/src/sample.rs
krystofwoldrich: What about 
```rust
|(i, sample)| active_ranges
                .get(&sample.thread_id)
                .is_some_and(|range| range.contains(&i)))```

Dav1dde: Oh that's stable now, nice!

relay-profiling/src/android.rs
krystofwoldrich: I've adjusted the name a bit. Let me know if that is more clear.

relay-profiling/src/android.rs
krystofwoldrich: Tests ensuring values are in current places are in `sample.rs`.

relay-profiling/src/sample.rs
krystofwoldrich: Thanks, I didn't know we had this.

relay-profiling/src/sample.rs
krystofwoldrich: I changed it to something similar

relay-profiling/tests/fixtures/profiles/android/roundtrip.rn.json
krystofwoldrich: Yes, it's not the minimum, but it's an improvement compared to the previous roundtrip test (1,5 MB).

iker-barriocanal (approved):

Dav1dde (approved):
Thanks for cleaning this up!

krystofwoldrich:
Thanks for the ✅ I'm waiting for one from the profiling team and then this can be merged.

phacops (approved):

viglia (approved):

PR #2709, opened 2023-11-09
ref(pii): Remove unwraps when parsing SelectorSpec

Dav1dde:
#skip-changelog

### DIFF
--- a/relay-pii/src/selector.rs
+++ b/relay-pii/src/selector.rs
@@ -35,6 +35,10 @@ pub enum InvalidSelectorError {
     /// Internal parser bug: An unexpected item was consumed.
     #[error("parser bug: consumed {0} (expected {1})")]
     UnexpectedToken(String, &'static str),
+
+    /// Internal parsing error, this should never happen and is a bug that needs to be fixed.
+    #[error("internal parser error")]
+    InternalError,
 }
 
 mod parser {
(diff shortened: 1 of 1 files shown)


relay-pii/src/selector.rs
```diff
@@ -170,6 +170,11 @@ impl SelectorSpec {
         handle_selector(
             SelectorParser::parse(Rule::RootSelector, s)
                 .map_err(|e| InvalidSelectorError::ParseError(Box::new(e)))?
+                // Unwrapping here is guaranteed to work based on the grammar that was just
+                // successfully parsed.
+                //
+                // Extracts the first `OrSelector` spanning the entire selector from the `RootSelector`.
+                // The `RootSelector` is guaranteed to have exactly one `OrSelector`.
```
olksdr: The problem with this `unwraps` that is something changes in the future and we lose these guarantees, we can forget revisit this piece of code and behave differently.  What do you think? 

We prefer not to have the unwraps in the first place. 

Will it be possible to actually map them to error at the end (even if that can never occur now) and just use `?` to propagate it?

Dav1dde: I was thinking, these unwraps will never actually throw and our tests would verify this. 

But thinking about it, might as well just "yeet" some internal error and create a sentry issue, this will still be caught by tests (hopefully) and if it doesn't we at least get the sentry error.

olksdr (approved):
LGTM!
Thanks for cleaning this up

PR #2711, opened 2023-11-09
feat(spans): Do not scrub Prisma spans

phacops:
Until now, since we assumed they were SQL queries, we tried to scrub them and failed so we'd return `None` and were discarding them.


jjbayer (approved):
Should be fine as long as these descriptions are low-cardinality, which https://github.com/getsentry/sentry-javascript/commit/fa8eb4841d740180f8792a8c871d5644f347402d indicates.

If it turns out there are auto generated entity names etc., we'll have to revert.

PR #2712, opened 2023-11-09
feat(spans): Add a metric to count spans per op

phacops:
In order to guide our focus, we'd like to count spans to understand better which op is the most represented and useful to our users.

https://github.com/getsentry/sentry/pull/59722


jjbayer (approved):
cc @nikhars

olksdr (approved):

PR #2714, opened 2023-11-10
ref(normalization): Remove TransactionsProcessor

iker-barriocanal:
This PR removes the transaction processor and moves its normalization to `NormalizeProcessor`, bringing us a step closer to having a single processor. There are no new normalization steps in this PR.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/mod.rs
+++ b/relay-event-normalization/src/normalize/mod.rs
@@ -30,6 +30,7 @@ use crate::{
 
 pub mod breakdowns;
 pub mod nel;
+pub(crate) mod processor;
 pub mod span;
 pub mod user_agent;
 pub mod utils;
@@ -37,7 +38,6 @@ pub mod utils;
 mod contexts;
 mod logentry;
 mod mechanism;
-mod processor;
 mod request;
 mod stacktrace;
 
--- a/relay-event-normalization/src/normalize/processor.rs
+++ b/relay-event-normalization/src/normalize/processor.rs
@@ -11,9 +11,11 @@ use std::collections::BTreeSet;
 
 use std::hash::{Hash, Hasher};
 use std::mem;
+use std::ops::Range;
 
 use chrono::{DateTime, Duration, Utc};
 use relay_base_schema::metrics::{is_valid_metric_name, DurationUnit, FractionUnit, MetricUnit};
+use relay_common::time::UnixTimestamp;
 use relay_event_schema::processor::{
     self, MaxChars, ProcessValue, ProcessingAction, ProcessingResult, ProcessingState, Processor,
 };
@@ -31,24 +33,43 @@ use crate::span::tag_extraction::{self, extract_span_tags};
 use crate::timestamp::TimestampProcessor;
 use crate::utils::MAX_DURATION_MOBILE_MS;
 use crate::{
-    breakdowns, schema, span, transactions, trimming, user_agent, BreakdownsConfig,
-    ClockDriftProcessor, DynamicMeasurementsConfig, GeoIpLookup, RawUserAgentInfo,
+    breakdowns, end_all_spans, normalize_transaction_name, schema, set_default_transaction_source,
+    span, trimming, user_agent, validate_transaction, BreakdownsConfig, ClockDriftProcessor,
+    DynamicMeasurementsConfig, GeoIpLookup, RawUserAgentInfo, TransactionNameConfig,
 };
 
 use crate::LightNormalizationConfig;
 
 /// Configuration for [`NormalizeProcessor`].
 #[derive(Clone, Debug, Default)]
-pub struct NormalizeProcessorConfig<'a> {
+pub(crate) struct NormalizeProcessorConfig<'a> {
     /// Light normalization config.
     // XXX(iker): we should split this config appropriately.
-    light_config: LightNormalizationConfig<'a>,
+    pub light_config: LightNormalizationConfig<'a>,
+
+    /// Configuration to apply to transaction names, especially around sanitizing.
+    pub transaction_name_config: TransactionNameConfig<'a>,
+
+    /// Timestamp range in which a transaction must end.
+    ///
+    /// Transactions that finish outside of this range are considered invalid.
+    /// This check is skipped if no range is provided.
+    pub transaction_range: Option<Range<UnixTimestamp>>,
 }
 
 impl<'a> From<LightNormalizationConfig<'a>> for NormalizeProcessorConfig<'a> {
-    fn from(config: LightNormalizationConfig<'a>) -> Self {
+    fn from(mut config: LightNormalizationConfig<'a>) -> Self {
+        // HACK(iker): workaround to avoid cloning of config items. We'll get
+        // rid of this when we remove light normalization in the next step.
+        let transaction_name_config = config.transaction_name_config;
+        config.transaction_name_config = TransactionNameConfig::default();
+        let transaction_range = config.transaction_range;
+        config.transaction_range = None;
+
         Self {
             light_config: config,
+            transaction_name_config,
+            transaction_range,
         }
     }
 }
@@ -61,7 +82,7 @@ impl<'a> From<LightNormalizationConfig<'a>> for NormalizeProcessorConfig<'a> {
 /// The returned [`ProcessingResult`] indicates whether the passed event should
 /// be ingested or dropped.
 #[derive(Debug, Default)]
-pub struct NormalizeProcessor<'a> {
+pub(crate) struct NormalizeProcessor<'a> {
     /// Configuration for the normalization steps.
     config: NormalizeProcessorConfig<'a>,
 }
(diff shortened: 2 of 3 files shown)


relay-event-normalization/src/normalize/processor.rs
```diff
@@ -79,23 +100,40 @@ impl<'a> Processor for NormalizeProcessor<'a> {
...
     ) -> ProcessingResult {
+        if event.ty.value() == Some(&EventType::Transaction) {
+            // TODO: Parts of this processor should probably be a filter so we
+            // can revert some changes to ProcessingAction)
+
+            validate_transaction(event, self.config.transaction_range.as_ref())?;
+
+            if let Some(trace_context) = event.context_mut::<TraceContext>() {
+                trace_context.op.get_or_insert_with(|| "default".to_owned());
+            }
```
iker-barriocanal, line 110: This was part of `validate_transaction`, but moving it here allows the validation function to accept a `&event` instead of `&mut event`.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -79,23 +100,40 @@ impl<'a> Processor for NormalizeProcessor<'a> {
         meta: &mut Meta,
         state: &ProcessingState<'_>,
     ) -> ProcessingResult {
+        if event.ty.value() == Some(&EventType::Transaction) {
+            // TODO: Parts of this processor should probably be a filter so we
+            // can revert some changes to ProcessingAction)
+
+            validate_transaction(event, self.config.transaction_range.as_ref())?;
```
iker-barriocanal, line 106: The validation happened after normalizing the transaction name, a few lines below. Since validation does not depend on the transaction name, moving it to the top here allows Relay to reject invalid transactions a bit earlier and save some computing. Note the outcome does not change. Happy to move this to another PR if this feels out of scope.

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -1362,6 +1326,11 @@ mod tests {
     },
 ]"#);
 
+        assert_eq!(
```
iker-barriocanal, line 1329: These assertions on the transaction source were lost in a previous PR, adding them back in this one.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -1732,4 +1770,51 @@ mod tests {
...
+      "span_id": "bd429c44b67a3eb4"
+    }
+  }
+}"#;
+
+        let mut processed = Annotated::<Event>::from_json(json).unwrap();
+        let processor_config = NormalizeProcessorConfig::default();
+        let mut processor = NormalizeProcessor::new(processor_config.clone());
+        process_value(&mut processed, &mut processor, ProcessingState::root()).unwrap();
+        remove_event_received(&mut processed);
```
iker-barriocanal: `event.received` is always updated with the latest timestamp, so removing it from the events makes us compare them easily. This is not new functionality and is already the case, see:
https://github.com/getsentry/relay/blob/6c294c96a83fd4c2bc1967aaf3ed9a890febded9/relay-event-normalization/src/normalize/mod.rs#L2173
We can probably get rid of some tests when we unify more functionality.

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -510,7 +478,7 @@ mod tests {
     fn test_replace_missing_timestamp() {
         let span = Span {
             start_timestamp: Annotated::new(
-                Utc.with_ymd_and_hms(1968, 1, 1, 0, 0, 1).unwrap().into(),
+                Utc.with_ymd_and_hms(1970, 1, 1, 0, 0, 1).unwrap().into(),
```
iker-barriocanal, line 481: Since the normalize processor now runs more logic, it'll drop spans happening before 1970/1/1. This fixes the test while keeping it valid.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -31,24 +33,43 @@ use crate::span::tag_extraction::{self, extract_span_tags};
...
 
 impl<'a> From<LightNormalizationConfig<'a>> for NormalizeProcessorConfig<'a> {
-    fn from(config: LightNormalizationConfig<'a>) -> Self {
+    fn from(mut config: LightNormalizationConfig<'a>) -> Self {
+        // HACK(iker): workaround to avoid cloning of config items. We'll get
+        // rid of this when we remove light normalization in the next step.
+        let transaction_name_config = config.transaction_name_config;
+        config.transaction_name_config = TransactionNameConfig::default();
+        let transaction_range = config.transaction_range;
+        config.transaction_range = None;
```
TBS1996:
```suggestion
        let transaction_name_config = mem::take(&mut config.transaction_name_config);
        let transaction_range = mem::take(&mut config.transaction_range);
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -30,14 +30,14 @@ use crate::{
 
 pub mod breakdowns;
 pub mod nel;
+pub(crate) mod processor;
```
olksdr, line 33: Why does it have to be `pub(crate)` ?

relay-event-normalization/src/normalize/processor.rs
olksdr: Something like this would work as well, much shorter and easier to understand I think. 

```suggestion
        let transaction_name_config = std::mem::take(&mut config.transaction_name_config);
        let transaction_range = config.transaction_range.take();
```

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -1732,4 +1770,51 @@ mod tests {
...
+        remove_event_received(&mut reprocessed);
+        assert_eq!(processed, reprocessed);
+
+        let mut reprocessed2 = reprocessed.clone();
+        process_value(&mut reprocessed2, &mut processor, ProcessingState::root()).unwrap();
+        remove_event_received(&mut reprocessed2);
+        assert_eq!(reprocessed, reprocessed2);
+    }
+
+    fn remove_event_received(event: &mut Annotated<Event>) {
```
olksdr:
```suggestion
    fn remove_received_from_event(event: &mut Annotated<Event>) {
```

relay-event-normalization/src/transactions/processor.rs
olksdr, line 1: Can this file actually be remove/renamed now? It doesn't contain processor anymore and it's more like utils for validation and looks like only 1 function for normalization.

TBS1996 (approved):
would be nice if we could avoid pub(crate) i think, but other than that looks good to me.

I assume you didn't change any of the functions you moved where you didn't specify it while i was reviewing

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: This crate is used in plenty of tests in `transactions/processor.rs`, which is a different module than `normalize/`. I don't want to bleed scope in this PR, so I'll get to this cleanup once the module structure is simplified.

relay-event-normalization/src/transactions/processor.rs
iker-barriocanal: Good suggestion! I'll get to this in a follow-up PR, this one is already big enough.

relay-event-normalization/src/normalize/mod.rs
olksdr: I'm ok with this if this will be revisited in the followup.  If the module made `pub(crate)` it means the structure of the modules is not really optimal and some code should be moved around. Like tests in this case.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -62,7 +81,7 @@ impl<'a> From<LightNormalizationConfig<'a>> for NormalizeProcessorConfig<'a> {
 /// The returned [`ProcessingResult`] indicates whether the passed event should
 /// be ingested or dropped.
 #[derive(Debug, Default)]
-pub struct NormalizeProcessor<'a> {
+pub(crate) struct NormalizeProcessor<'a> {
```
olksdr, line 84: same comment about `pub(crate)`

olksdr (approved):

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: I agree with the suboptimal module structure. I'm going to merge this as-is and address it in a follow-up PR.

relay-event-normalization/src/normalize/processor.rs
iker-barriocanal: Resolving as this is addressed in a follow-up PR: https://github.com/getsentry/relay/pull/2729.

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: Resolving as this is addressed in a follow-up PR: https://github.com/getsentry/relay/pull/2729.

PR #2715, opened 2023-11-10
feat(sampling): Add profile context early and expose getter for it

### LINKED ISSUE #2710: Make it possible to create Investigation Bias with has:profile.id
Users would like to create Investigation Biases (reservoir rules that boost Dynamic Sampling to 100%) with the condition containing `has:profile.id`. This ticket is about the Relay part that's needed in order for this to happen.
Jan mentioned that we don't have `profile.id` exposed as a queryable field, actually, it's not a field at all.

jjbayer:
In order to create dynamic sampling rules depending on the profile ID, add the profile ID to the transaction event's `profile` context and expose it via the event's `Getter` implementation.

Note that most transaction processing still occurs only in processing relays, and it might happen that events that have a profile ID at the time of sampling will not have a profile ID anymore after processing.

Fixes https://github.com/getsentry/relay/issues/2710

### DIFF
--- a/relay-profiling/src/lib.rs
+++ b/relay-profiling/src/lib.rs
@@ -118,10 +118,12 @@ mod utils;
 
 const MAX_PROFILE_DURATION: Duration = Duration::from_secs(30);
 
+pub type ProfileId = EventId;
+
 #[derive(Debug, Deserialize)]
 struct MinimalProfile {
     #[serde(alias = "profile_id")]
-    event_id: EventId,
+    event_id: ProfileId,
     platform: String,
     #[serde(default)]
     version: sample::Version,
@@ -134,7 +136,7 @@ fn minimal_profile_from_json(
     serde_path_to_error::deserialize(d)
 }
 
-pub fn parse_metadata(payload: &[u8], project_id: ProjectId) -> Result<(), ProfileError> {
+pub fn parse_metadata(payload: &[u8], project_id: ProjectId) -> Result<ProfileId, ProfileError> {
     let profile = match minimal_profile_from_json(payload) {
         Ok(profile) => profile,
         Err(err) => {
@@ -183,7 +185,7 @@ pub fn parse_metadata(payload: &[u8], project_id: ProjectId) -> Result<(), Profi
             _ => return Err(ProfileError::PlatformNotSupported),
         },
     };
-    Ok(())
+    Ok(profile.event_id)
 }
 
 pub fn expand_profile(payload: &[u8], event: &Event) -> Result<(EventId, Vec<u8>), ProfileError> {
--- a/relay-profiling/src/lib.rs
+++ b/relay-profiling/src/lib.rs
@@ -118,6 +118,9 @@ mod utils;
 
 const MAX_PROFILE_DURATION: Duration = Duration::from_secs(30);
 
+/// Unique identifier for a profile.
+///
+/// Same format as event IDs.
 pub type ProfileId = EventId;
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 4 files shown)


relay-profiling/src/lib.rs
```diff
@@ -118,10 +118,15 @@ mod utils;
 
 const MAX_PROFILE_DURATION: Duration = Duration::from_secs(30);
 
+/// Unique identifier for a profile.
+///
+/// Same format as event IDs.
+pub type ProfileId = EventId;
```
jjbayer, line 124: I added this type alias to make the return type more obvious in function signatures.

relay-server/src/actors/processor.rs
```diff
@@ -1118,7 +1118,22 @@ impl EnvelopeProcessorService {
...
+                if let Some(profile_id) = profile_id {
+                    contexts.add(ProfileContext {
+                        profile_id: Annotated::new(profile_id),
+                    });
+                // If not, we delete the profile context.
+                } else {
+                    contexts.remove::<ProfileContext>();
+                }
+            }
+        }
```
jjbayer, line 1136: This block was copied and adapted from `process_profiles`.

relay-server/src/actors/processor.rs
```diff
@@ -1182,17 +1197,13 @@ impl EnvelopeProcessorService {
...
-                // If not, we delete the profile context.
-                } else {
-                    contexts.remove::<ProfileContext>();
+        if found_profile_id.is_none() {
+            // Remove profile context from event.
+            if let Some(event) = state.event.value_mut() {
+                if event.ty.value() == Some(&EventType::Transaction) {
+                    if let Some(contexts) = event.contexts.value_mut() {
+                        contexts.remove::<ProfileContext>();
+                    }
```
jjbayer, line 1206: When profile processing fails, the profile ID is removed here. This means that some (invalid) profiles that were opted in by a dynamic sampling rule based on `contexts.profile.profile_id` might end up _without_ a profile ID in the final payload.

iambriccardo: So this implies that we are performing sampling on profile ids that we don't know whether they are valid of not, am I right?

From the code flow, I see that we do filter the profile with context injection, we then move onto sampling (assuming the profile id refers to a valid profile), and then we process the profile and remove the context in case it's invalid?

Is this an expected behavior? Just curious

olksdr (approved):
lgtm

It might be a good idea to add a test for this, to make sure that profile id will be added where it's needed.

relay-profiling/src/lib.rs
Dav1dde: Shouldn't we just new type it here instead?

Then it's actually a separate type and cant get mixed

relay-server/src/actors/processor.rs
```diff
@@ -1118,7 +1118,22 @@ impl EnvelopeProcessorService {
...
+                let contexts = event.contexts.get_or_insert_with(Contexts::new);
+                // If we found a profile, add its ID to the profile context on the transaction.
+                if let Some(profile_id) = profile_id {
+                    contexts.add(ProfileContext {
+                        profile_id: Annotated::new(profile_id),
+                    });
+                // If not, we delete the profile context.
+                } else {
+                    contexts.remove::<ProfileContext>();
+                }
```
iker-barriocanal, line 1134: If a `ProfileContext` already exists, this code will either create a "duplicated" profile context or remove it -- is this desired? I think these two shouldn't happen because the profile id is taken from the item header, but not sure.

relay-server/src/actors/processor.rs
jjbayer: > So this implies that we are performing sampling on profile ids that we don't know whether they are valid of not, am I right?

That is correct and a bit unfortunate. To fix this we would have to move the entirety of `process_profiles` to non-processing relays (and before dynamic sampling), and I'm not sure if that's currently possible. cc @jan-auer

relay-server/src/actors/processor.rs
jjbayer: `add` actually replaces the existing profile context.

relay-server/src/actors/processor.rs
iambriccardo: Sounds good, thanks for the explanation!

jan-auer (approved):
The caveat of over-sampling is fine, at least it will still catch events with a profile. 

In the longer run, a sampling rule like this might not have the ideal semantics. We would ideally want to sample events where there's a profile in the same trace. However, we're currently not able to guarantee such a rule.

PR #2720, opened 2023-11-14
ref(py): Rename sampling to rule condition

jan-auer:
Renames `validate_sampling_condition` to `validate_rule_condition` in the C-ABI
and python bindings to make it clear that this function validates all kinds of
rule conditions.


olksdr (approved):
:shipit:

PR #2721, opened 2023-11-14
feat(protocol): Support comparisons in rule conditions on strings

jan-auer:
Adds the ability to run all comparison operators (`>`, `>=`, `<`, `<=`)
on strings. This applies to dynamic sampling and metric extraction.

Since this change breaks backwards compatibility, the versioning for
metrics extraction and dynamic sampling were bumped.

Dynamic sampling configs did not have a versioning schema before. The
last breaking change introduced a `rules_v2` array. To fix this, there
is now versioning built into the new config. Additionally, an
`ErrorBoundary` is added around sampling configuration to prevent
accidental deserialization failures.

Backwards compatibility is as follows:

- New Relay <> Old Sentry: Rules are automatically upgraded from
  `rules_v2`.
- Old Relay <> New Sentry: Cannot see the new rules and forwards events.

Relates to https://github.com/getsentry/sentry/issues/59896
Resolves https://github.com/getsentry/team-ingest/issues/244

### DIFF
--- a/py/tests/test_processing.py
+++ b/py/tests/test_processing.py
@@ -248,7 +248,7 @@ def test_invalid_sampling_condition():
         sentry_relay.validate_rule_condition(condition)
 
 
-def test_validate_sampling_configuration():
+def test_validate_legacy_sampling_configuration():
     """
     Tests that a valid sampling rule configuration passes
     """
@@ -281,6 +281,32 @@ def test_validate_sampling_configuration():
                     "options": {"ignoreCase": true}
                 },
                 "id":2
+            }
+        ]
+    }"""
+    # Should NOT throw
+    sentry_relay.validate_sampling_configuration(config)
+
+
+def test_validate_legacy_sampling_configuration():
+    """
+    Tests that a valid sampling rule configuration passes
+    """
+    config = """{
+        "version": 2,
+        "rules": [
+            {
+                "type": "trace",
+                "samplingValue": {
+                    "type": "sampleRate",
+                    "value": 0.7
+                },
+                "condition": {
+                    "op": "custom",
+                    "name": "event.legacy_browser",
+                    "value":["ie10"]
+                },
+                "id":1
             }
         ]
     }"""
--- a/py/tests/test_processing.py
+++ b/py/tests/test_processing.py
@@ -264,9 +264,9 @@ def test_validate_legacy_sampling_configuration():
                 "condition": {
                     "op": "custom",
                     "name": "event.legacy_browser",
-                    "value":["ie10"]
+                    "value": ["ie10"]
                 },
-                "id":1
+                "id": 1
             },
             {
                 "type": "trace",
@@ -277,10 +277,10 @@ def test_validate_legacy_sampling_configuration():
                 "condition": {
                     "op": "eq",
                     "name": "event.release",
-                    "value":["1.1.*"],
+                    "value": ["1.1.*"],
                     "options": {"ignoreCase": true}
                 },
-                "id":2
+                "id": 2
             }
         ]
     }"""
@@ -304,10 +304,10 @@ def test_validate_sampling_configuration():
                 "condition": {
                     "op": "eq",
                     "name": "event.release",
-                    "value":["1.1.*"],
+                    "value": ["1.1.*"],
                     "options": {"ignoreCase": true}
                 },
-                "id":2
+                "id": 2
             }
         ]
     }"""
--- a/py/tests/test_processing.py
+++ b/py/tests/test_processing.py
@@ -288,7 +288,7 @@ def test_validate_legacy_sampling_configuration():
     sentry_relay.validate_sampling_configuration(config)
 
 
-def test_validate_legacy_sampling_configuration():
+def test_validate_sampling_configuration():
     """
     Tests that a valid sampling rule configuration passes
     """
--- a/py/tests/test_processing.py
+++ b/py/tests/test_processing.py
@@ -299,14 +299,15 @@ def test_validate_sampling_configuration():
                 "type": "trace",
                 "samplingValue": {
                     "type": "sampleRate",
-                    "value": 0.7
+                    "value": 0.9
                 },
                 "condition": {
-                    "op": "custom",
-                    "name": "event.legacy_browser",
-                    "value":["ie10"]
+                    "op": "eq",
+                    "name": "event.release",
+                    "value":["1.1.*"],
+                    "options": {"ignoreCase": true}
                 },
-                "id":1
+                "id":2
             }
         ]
     }"""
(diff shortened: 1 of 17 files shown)


py/tests/test_processing.py
```diff
@@ -288,6 +288,33 @@ def test_validate_sampling_configuration():
...
+            {
+                "type": "trace",
+                "samplingValue": {
+                    "type": "sampleRate",
+                    "value": 0.9
+                },
+                "condition": {
+                    "op": "eq",
+                    "name": "event.release",
+                    "value":["1.1.*"],
```
Dav1dde:
```suggestion
                    "value": ["1.1.*"],
```

relay-protocol/src/condition.rs
```diff
@@ -106,12 +106,12 @@ macro_rules! impl_cmp_condition {
             /// Path of the field that should match the value.
             pub name: String,
             /// The numeric value to check against.
-            pub value: Number,
+            pub value: Value,
```
Dav1dde, line 109: Now since this is a value, should we also add array support like `Eq` has?

relay-sampling/src/config.rs
```diff
@@ -7,22 +7,36 @@ use serde::{Deserialize, Serialize};
...
 pub struct SamplingConfig {
+    /// The required version to run dynamic sampling.
+    ///
+    /// Defaults to legacy version (`1`) when missing.
+    #[serde(default = "SamplingConfig::legacy_version")]
+    pub version: u16,
+
     /// The ordered sampling rules for the project.
     ///
     /// This field will remain here to serve only for old customer Relays to which we will
```
Dav1dde: This is now outdated

relay-sampling/src/config.rs
```diff
@@ -7,22 +7,36 @@ use serde::{Deserialize, Serialize};
...
     /// forward the sampling config. The idea is that those Relays will get the old rules as
     /// empty array, which will result in them not sampling and forwarding sampling decisions to
     /// upstream Relays.
-    #[serde(default, skip_deserializing)]
+    #[serde(default)]
     pub rules: Vec<SamplingRule>,
 
     /// The ordered sampling rules v2 for the project.
```
Dav1dde: Should note that is only used for version 1, even though it has v2 in its name

```suggestion
    /// Ordered sampling rules, used in version 1. Removed in favour of `Self::rules` in version 2.
```

relay-sampling/src/config.rs
```diff
@@ -31,16 +45,42 @@ pub struct SamplingConfig {
...
 
 impl SamplingConfig {
+    /// Creates an enabled configuration with empty defaults and the latest version.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
     /// Returns `true` if any of the rules in this configuration is unsupported.
     pub fn unsupported(&self) -> bool {
-        !self.rules_v2.iter().all(SamplingRule::supported)
```
Dav1dde, line 36: Maybe a `debug_assert!(self.version > 1)` to catch when this is called without normalization first.

relay-sampling/src/config.rs
```diff
@@ -31,16 +45,42 @@ pub struct SamplingConfig {
...
+        self.version > SAMPLING_CONFIG_VERSION || !self.rules.iter().all(SamplingRule::supported)
     }
 
     /// Filters the sampling rules by the given [`RuleType`].
     pub fn filter_rules(&self, rule_type: RuleType) -> impl Iterator<Item = &SamplingRule> {
-        self.rules_v2
-            .iter()
-            .filter(move |rule| rule.ty == rule_type)
```
Dav1dde, line 43: dito `debug_assert!(self.version > 1)`

relay-sampling/src/config.rs
```diff
@@ -422,20 +462,9 @@ mod tests {
...
-                     "type": "trace",
-                     "active": true,
-                     "condition": {
-                        "op": "and",
-                        "inner": []
-                     },
-                     "id": 1000
-                 }
-               ],
+               "rules": [],
```
Dav1dde, line 468: Legacy does/may not contain a rules element, it was annotated with `skip_deserializing`,  I would remove it here

jan-auer:
For reviewers: A regression test for the legacy config is still missing.

Dav1dde (approved):

relay-protocol/src/condition.rs
jan-auer: I didn't wanna add this since the semantics of a comparison with an array are questionable. For `eq`, we defined it as "one of", but for comparisons this seemed weird to me. Thoughts?

relay-sampling/src/config.rs
jan-auer: I'll skip this one, especially since we use `.rules` directly most of the time.

relay-sampling/src/config.rs
jan-auer: Legacy actually required serializing an empty `rules` array since in even earlier versions of Relay, that `skip_deserializing` annotation was missing. I'll keep it here to reflect the actual payload that's sent by these versions. 

This is important, because without these empty rules we could've just used a `serde(alias)`. Clarified this in a doc comment now.

relay-sampling/src/config.rs
```diff
@@ -7,22 +7,36 @@ use serde::{Deserialize, Serialize};
...
 
-    /// The ordered sampling rules v2 for the project.
+    /// **Deprecated**. The ordered sampling rules for the project in legacy format.
+    ///
+    /// Removed in favor of `Self::rules` in version `2`. This field remains here to parse rules
+    /// from old Sentry instances and convert them into the new format. The legacy format contained
+    /// both an empty `rules` as well as the actual rules in `rules_v2`. During normalization, these
+    /// two arrays are merged together.
+    #[serde(default, skip_serializing)]
     pub rules_v2: Vec<SamplingRule>,
```
iker-barriocanal, line 40: In some enums we've renamed the option to `deprecated` and set a serde alias on it. Could that be helpful here?

iker-barriocanal (approved):

olksdr (approved):

relay-sampling/src/config.rs
Dav1dde: I like this

relay-protocol/src/condition.rs
Dav1dde: I personally found it already surprising for `Eq`, usually this would be a `In` operator. So for me then I assumed the value can always be an array and it turns into something like `value <OPERATOR> ANY(value)`. 

I agree though this is weird but we already have a precedent for the array. Though I think we just leave it as is and implement it when we need it (we can bump versions now, yay)

relay-sampling/src/config.rs
Dav1dde: Thanks, I confused `skip_deserializing` with `skip_serializing` here.

relay-sampling/src/config.rs
jan-auer: It's a good idea, however it creates a difference between how Sentry serializes the schema and how it looks in Relay. I'm afraid that it causes more confusion when looking at the code.

PR #2723, opened 2023-11-14
ref(crons): Include message_type in kafka message

evanpurkhiser:
There are two types of messages that end up in the ingest-monitors kafka topic, "check_in" (the ones produced here in relay) and "clock_pulse" messages, which are produced externally and are intended to ensure the clock continues to run even when ingestion volume drops.

Currently we have some shim logic in our producer which handles adding in this message type when it is missing

https://github.com/getsentry/sentry/blob/07522eb8c43828deee6a74de2bb72cf989ec907d/src/sentry/monitors/consumers/monitor_consumer.py#L591-L595

This change introduces the `message_type` key for the check-in messages so we no longer need to shim this.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/store.rs
+++ b/relay-server/src/actors/store.rs
@@ -720,6 +720,7 @@ impl StoreService {
         item: &Item,
     ) -> Result<(), StoreError> {
         let message = KafkaMessage::CheckIn(CheckInKafkaMessage {
+            message_type: CheckInMessageType::CheckIn,
             project_id,
             retention_days,
             start_time: UnixTimestamp::from_instant(start_time).as_secs(),
@@ -1023,11 +1024,21 @@ struct ProfileKafkaMessage {
     payload: Bytes,
 }
 
+#[allow(dead_code)]
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum CheckInMessageType {
+    ClockPulse,
+    CheckIn,
+}
+
 #[derive(Debug, Serialize)]
 struct CheckInKafkaMessage {
     #[serde(skip)]
     routing_key_hint: Option<Uuid>,
 
+    /// Used by the consumer to discrinminate the message.
+    message_type: CheckInMessageType,
     /// Raw event payload.
     payload: Bytes,
     /// Time at which the event was received by Relay.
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/store.rs
```diff
@@ -1028,6 +1029,9 @@ struct CheckInKafkaMessage {
     #[serde(skip)]
     routing_key_hint: Option<Uuid>,
 
+    /// Discriminate tag used to differentiate these messages from clock pulse messages that are
+    /// also placed into the topic.
+    message_type: String,
```
Dav1dde: Should we use an enum for this?

evanpurkhiser: Oh yeah, I like that idea. 👍

relay-server/src/actors/store.rs
```diff
@@ -1028,6 +1029,9 @@ struct CheckInKafkaMessage {
     #[serde(skip)]
     routing_key_hint: Option<Uuid>,
 
+    /// Discriminate tag used to differentiate these messages from clock pulse messages that are
+    /// also placed into the topic.
```
iker-barriocanal: Could we reword this to meet [relay coding guidelines (internal)](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#987c0416a9f54674999eed59eecbf846)? Thanks!

> ensure that there is a precise first line, followed by more information

iker-barriocanal (approved):

relay-server/src/actors/store.rs
evanpurkhiser: I think I fixed this

evanpurkhiser:
Gonna get one more look over this just to be safe.

relay-server/src/actors/store.rs
```diff
@@ -1023,11 +1024,21 @@ struct ProfileKafkaMessage {
     payload: Bytes,
 }
 
+#[allow(dead_code)]
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum CheckInMessageType {
```
Dav1dde, line 1030: The first sentence of your issue would be nice here. Documenting the use of the values and that relay is only supposed to ever send `CheckIn`'s

Dav1dde (approved):

iker-barriocanal (approved):
🚀

relay-server/src/actors/store.rs
evanpurkhiser: I'll add it in in a diff PR 👍

PR #2724, opened 2023-11-15
fix(pii): Replace only user in the userpath

### LINKED ISSUE #2722: Large/Wrong truncation of `userpath` on PII on some path patterns
### Environment

relay 23.10.1, Rust 1.73.0

### Steps to Reproduce

1. Have a `PiiProcessor` compiled with `@userpath` in the config applications
2. Have this kind of string to do PII removal on
    ```
    make
    export PATH=$PATH:$HOME/collectd_exporter
    line
    line
    line
    line
    line
    ```

### Expected Result

Have 

    ```
    make
    export PATH=$PATH:$HOME/[user]
    line
    line
    line
    line
    line
    ```
### Actual Result

    ```
    make
    export PATH=$PATH:$HOME/[user]
    ```
    
 ### Code to reproduce
 
 ```rs
         let mut config = PiiConfig::default();
        config.applications = {
            let mut map = BTreeMap::new();
            map.insert(
                ValueType::String.into(),
                vec![
                    "@iban".into(),
                    "@ip".into(),
                    "@email".into(),
                    "@creditcard".into(),
                    "@pemkey".into(),
                    "@urlauth".into(),
                    "@userpath".into(),
                    "@password".into(),
                    "@usssn".into(),
                ],
            );
            map
        };

        let state = &STATE;
        let input = "make
export PATH=$PATH:$HOME/collectd_exporter
log
log
log
log
"
        .to_string();
        let compiled = config.compiled();
        let mut processor = PiiProcessor::new(compiled);
        let mut root = Annotated::new(input);
        process_value(&mut root, &mut processor, state).unwrap();
        println!("{root:?}");
 ```

olksdr:
Explicitly use only allowed characters for the username. 

Currently we replace everything till between the slashes, and if the slash after username is missing we will replace everything till the end of the input string. 

fix: https://github.com/getsentry/relay/issues/2722

#skip-changelog

### DIFF
--- a/relay-pii/src/regexes.rs
+++ b/relay-pii/src/regexes.rs
@@ -231,7 +231,7 @@ static PATH_REGEX: Lazy<Regex> = Lazy::new(|| {
                 )
             )
             (
-                [^/\\]+
+                [A–Za-z0–9'\._!\#^~-]+
             )
         ",
     )
--- a/relay-pii/src/regexes.rs
+++ b/relay-pii/src/regexes.rs
@@ -231,7 +231,7 @@ static PATH_REGEX: Lazy<Regex> = Lazy::new(|| {
                 )
             )
             (
-                [A–Za-z0–9'\._!\#^~-]+
+                [A–Za-z0–9'\.\ _!\#^~-]+
             )
         ",
     )
--- a/relay-pii/src/builtin.rs
+++ b/relay-pii/src/builtin.rs
@@ -1256,6 +1256,24 @@ HdmUCGvfKiF2CodxyLon1XkK8pX+Ap86MbJhluqK
             ];
         );
         assert_text_rule!(
+            rule = "@userpath";
+            input = r#"make
+            export PATH=$PATH:$HOME/collectd_exporter
+            log
+            log
+            log
+            "#;
+            output = r#"make
+            export PATH=$PATH:$HOME/[user]
+            log
+            log
+            log
+            "#;
+            remarks = vec![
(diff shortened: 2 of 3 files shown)


relay-pii/src/regexes.rs
```diff
@@ -231,7 +231,7 @@ static PATH_REGEX: Lazy<Regex> = Lazy::new(|| {
                 )
             )
             (
-                [^/\\]+
+                [A–Za-z0–9'\._!\#^~-]+
```
TBS1996: I think this would stop at first space? and I believe spaces are allowed in windows usernames. Should it also be in the list of allowed characters? 

I checked on regex101 for this

Dav1dde: Maybe should match anything until `\` or newline.

Edit: did this ever work on Windows with backwards slashes for path separators?

olksdr: @TBS1996  Overall it's not advisable to have spaces in the usernames and I'm not sure we want to support it either. 
@Dav1dde  Also if we remove everything , or just will new line we again end up with a somewhat wrong scrubbing. 

See for example changes in `relay-server/tests/snapshots/test_fixtures__dotnet__pii_stripping.snap` below - it was scrubbed kind of a little bit too much. I do not mind that either, but we have to agree on the final rule 😄

olksdr: included also space in the regex now

Dav1dde (approved):

iker-barriocanal (approved):

PR #2729, opened 2023-11-16
ref(normalization): Reduce processor module visibility

iker-barriocanal:
Follow-up to https://github.com/getsentry/relay/pull/2714#discussion_r1395425833.

The are a lot of tests requiring `NormalizeProcessor` in `transactions/processor.rs`, forcing higher visibility than necessary on `normalize/processor.rs`. This PR moves required tests from one module to another, thus reducing the required module visibility. As we continue to refactor the processors, we'll find a better home for functions and tests.

This PR does not contain any functional changes or updates to tests.

#skip-changelog


relay-event-normalization/src/transactions/processor.rs
iker-barriocanal, line 1: This module now contains very little functionality. I'm going to keep it for now, as more things will be moving around and I don't want to set a clear definition of modules until functional refactors are finished. Let me know if you have a strong opinion that differs from this.

olksdr: I would still rename this module in this PR, just to reflect it does not contain `processor` anymore.

olksdr (approved):

relay-event-normalization/src/transactions/processor.rs
iker-barriocanal: Yeah that's fair. Addressed in [c31fea2](https://github.com/getsentry/relay/pull/2729/commits/c31fea235563507c0d5e72ea420b250e50feb3de).

PR #2731, opened 2023-11-16
ref(normalization): Replace LightNormalization with NormalizeProcessor

iker-barriocanal:
This PR removes light normalization in favor of using `NormalizeProcessor` directly, and updates all the calls to directly invoke `NormalizeProcessor`. There are no functional changes in this PR, just refactors. Note this PR doesn't get rid of "light normalization" as a concept entirely, as it's referenced in other parts of the codebase. That will be addressed in future PRs.

It's suggested to review commits sequentially.

#skip-changelog


TBS1996 (approved):
lgtm

seems pretty straightforward, and it touches a lot of tests that all pass 

nice job!

relay-cabi/src/processing.rs
```diff
@@ -137,7 +137,11 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
         enable_trimming: config.enable_trimming.unwrap_or_default(),
         measurements: None,
     };
-    light_normalize_event(&mut event, light_normalization_config)?;
+    process_value(
+        &mut event,
+        &mut NormalizeProcessor::new(normalization_config),
+        ProcessingState::root(),
+    )?;
```
olksdr, line 144: So we still use store processor here after the new `NormalizeProcessor`?

olksdr (approved):

relay-cabi/src/processing.rs
iker-barriocanal: Yes. We still need to run both processors in some circumstances until we unify them.

PR #2733, opened 2023-11-16
fix(profiling): Transfer profile ID when the event has been extracted

jjbayer:
Relay copies profile IDs into a transaction context in order to associate profiles to transactions.

In https://github.com/getsentry/relay/pull/2715, we tried to move this extraction to a point _before_ dynamic sampling, so we can define transaction sampling rules based on the profile ID. However, the PR tried to do this at a point where the event had not been parsed yet, resulting in the profile ID always missing.

As @olksdr pointed out,

> It might be a good idea to add a test for this, to make sure that profile id will be added where it's needed.

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -31,8 +31,8 @@ use relay_event_normalization::{GeoIpLookup, RawUserAgentInfo};
 use relay_event_schema::processor::{self, ProcessingAction, ProcessingState};
 use relay_event_schema::protocol::{
     Breadcrumb, ClientReport, Contexts, Csp, Event, EventType, ExpectCt, ExpectStaple, Hpkp,
-    IpAddr, LenientString, Metrics, NetworkReportError, OtelContext, RelayInfo, Replay,
-    SecurityReportType, SessionAggregates, SessionAttributes, SessionStatus, SessionUpdate,
+    IpAddr, LenientString, Metrics, NetworkReportError, OtelContext, ProfileContext, RelayInfo,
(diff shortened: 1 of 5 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -2756,6 +2769,7 @@ impl EnvelopeProcessorService {
             });
 
             self.extract_event(state)?;
+            self.transfer_profile_id(state);
```
jjbayer, line 2772: This is the main change of this PR: The transfer of the profile ID happens _after_ event extraction. The rest of the PR is largely identical with https://github.com/getsentry/relay/pull/2715.

relay-server/src/actors/processor.rs
```diff
@@ -4437,4 +4451,96 @@ mod tests {
 
         assert_eq!(get_sampling_match(res).sample_rate(), 0.4);
     }
+
+    #[tokio::test]
+    async fn test_profile_id_transfered() {
```
olksdr: does this have to be `tokio::test` and `async`?

olksdr (approved):
let's 🚢  it again!

iker-barriocanal (approved):

relay-server/src/actors/processor.rs
jjbayer: 👍 It can be sync if I use dummy `Addr`s instead of `mock_service`. Changed here: https://github.com/getsentry/relay/pull/2733/commits/e68e1badb5f48193c2c6aed46a56832d9a157155

PR #2734, opened 2023-11-16
feat(webvitals): Updates performance score values to use 0-1 cdf values instead of the actual raw component score

edwardgou-sentry:
Updates web vital performance score component values to use 0 - 1 cdf values instead of the actual raw component score. The raw component scores have dynamic upper bounds depending on the component weights provided, so we should use the cdf values since they're bound between 0 - 1 (which we can combine with the weight to get the actual score in the product).


relay-event-normalization/src/normalize/processor.rs
```diff
@@ -915,7 +916,7 @@ fn normalize_performance_score(
                     measurements.insert(
                         "score.total".to_owned(),
                         Measurement {
-                            value: score_total.into(),
+                            value: (score_total / weight_total).into(),
```
iker-barriocanal, line 919: Not actionable for this PR, but to not forget about if/how we can remove the `score.total` measurement.

iker-barriocanal (approved):

py/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- `normalize_performance_score` stores 0 to 1 cdf score instead of weighted score for each performance score component ([#2734](https://github.com/getsentry/relay/pull/2734))
```
jjbayer: Please also update the top-level changelog (this is the changelog for the python library).
```suggestion
- `normalize_performance_score` stores 0 to 1 cdf score instead of weighted score for each performance score component. ([#2734](https://github.com/getsentry/relay/pull/2734))
```

jjbayer (approved):

py/CHANGELOG.md
edwardgou-sentry: Updated!

k-fish:
If you're planning on doing `avg CDF * avg weight` for a component (eg. FCP) then I would triple check that the total score and each component score adds up every which way. It's a different calculation because each individual CDF has a difference on the sum of weighted CDF depending on that weight _for that particular profile_. 

I wrote storing weights / weighted scores as a one liner on the original doc, but it was almost a footnote, it really deserves it's own section, I added a `Example Score Calculations` section to walk through the calculations.

PR #2735, opened 2023-11-16
feat(spans): Extension tag for resource spans

jjbayer:
Add a file extension tag to span metrics for resource spans, so they can be filtered by it.

This does not contribute to cardinality, because the extension tag is derived from the description tag.

See https://github.com/getsentry/sentry/pull/60179 for sentry-side PR.

#skip-changelog


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -331,6 +334,17 @@ pub(crate) fn extract_tags(
...
             let truncated = truncate_string(scrubbed_desc, config.max_tag_value_size);
+            if span_op.starts_with("resource.") {
+                if let Some(ext) = truncated
+                    .rsplit('/')
+                    .next()
+                    .and_then(|last_segment| last_segment.rsplit_once('.'))
+                    .map(|(_, extension)| extension)
+                {
+                    span_tags.insert(SpanTagKey::Extension, ext.to_owned());
+                }
```
DominikB2014, line 345: Just to clarify here, if there is no extension, we simply do not insert it correct? This would allow to query for spans that don't have an extension

DominikB2014 (commented):
Lgtm

DominikB2014 (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -331,6 +334,17 @@ pub(crate) fn extract_tags(
             span_tags.insert(SpanTagKey::Group, span_group);
 
             let truncated = truncate_string(scrubbed_desc, config.max_tag_value_size);
+            if span_op.starts_with("resource.") {
+                if let Some(ext) = truncated
+                    .rsplit('/')
+                    .next()
+                    .and_then(|last_segment| last_segment.rsplit_once('.'))
```
Dav1dde, line 341:
```suggestion
                    .unwrap_or(truncated)
                    .rsplit_once('.')
```

Then this works also on resources which do not contain a slash

Dav1dde, line 341:
```suggestion
                    .unwrap_or(truncated.as_str())
                    .rsplit_once('.')
```

To make this work on resources with no slash

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -331,6 +334,17 @@ pub(crate) fn extract_tags(
             span_tags.insert(SpanTagKey::Group, span_group);
 
             let truncated = truncate_string(scrubbed_desc, config.max_tag_value_size);
+            if span_op.starts_with("resource.") {
```
Dav1dde, line 337: Should we move this to the below `span_op.starts_with("resource.")` block, so all the resource extraction is one place?

Dav1dde (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: That was the first thing I tried, but `truncated` is moved into `span_tags` on line 348 so I had to access it before that.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: The `rsplit('/').next()` above returns `Some` even if there is no slash, so that should be equivalent to your solution.

relay-dynamic-config/src/defaults.rs
```diff
@@ -152,6 +152,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 ("", "transaction.method"),
                 ("", "transaction.op"),
                 ("", "resource.render_blocking_status"), // only set for resource spans
+                ("", "file_extension"),                  // only set for resource spans
```
jjbayer, line 155: Had to use `file_extension` instead of `extension` because `extension` has already been used in the indexer db.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -83,6 +85,7 @@ impl SpanTagKey {
             SpanTagKey::System => "system",
             SpanTagKey::TimeToFullDisplay => "ttfd",
             SpanTagKey::TimeToInitialDisplay => "ttid",
+            SpanTagKey::Extension => "file_extension",
```
olksdr: Do we want to keep the naming consistent and have actually `FileExtension` enum variant?

olksdr (approved):
Left one question, otherwise lgtm

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -331,6 +334,17 @@ pub(crate) fn extract_tags(
...
 
             let truncated = truncate_string(scrubbed_desc, config.max_tag_value_size);
+            if span_op.starts_with("resource.") {
+                if let Some(ext) = truncated
+                    .rsplit('/')
+                    .next()
+                    .and_then(|last_segment| last_segment.rsplit_once('.'))
+                    .map(|(_, extension)| extension)
+                {
+                    span_tags.insert(SpanTagKey::Extension, ext.to_owned());
```
iker-barriocanal: Should we normalize the extension itself? For example, lowercasing or an allow-list of possible values.

iker-barriocanal (approved):
LGTM. Please update the PR description with the indexer PR when it's ready.

relay-event-normalization/src/normalize/span/tag_extraction.rs
Dav1dde: Oh right. Nice.

relay-event-normalization/src/normalize/span/tag_extraction.rs
DominikB2014: I'm a bit hesitant to put an allow list, its possible we might just not think of a certain extension and just not ingest it, for example a weird one is `.jsonp` which is technically js. Ingesting all extensions provides us the flexibility to decide what extensions we want to see on the query/UI side, and we can easily change what we display. 

Although I do think normalizing to lowercase makes sense.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: Done.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: Added lowercasing now. The allowlist would not improve cardinality, because the tag is derived from the existing `scrubbed.description` tag.

PR #2738, opened 2023-11-17
ref(normalization): Remove schema processor

iker-barriocanal:
`SchemaProcessor` is currently only an indirection step in `NormalizeProcessor` for schema checks. This PR removes the schema processor and performs these checks in `NormalizeProcessor`.

`SchemaProcessor` actions now run as part of processing child values of `NormalizeProcessor`. This order change doesn't matter, as `SchemaProcessor` adds remarks and/or meta errors on missing values. Decisions on rejecting events stay the same, and normalization steps won't be performed as the value is missing (same behavior).

#skip-changelog


relay-event-normalization/src/lib.rs
```diff
@@ -24,13 +24,12 @@ mod legacy;
...
 mod regexes;
 mod remove_other;
-pub mod replay;
-mod schema;
 mod statsd;
 mod timestamp;
 mod transactions;
 mod trimming;
 
+pub mod replay;
```
olksdr, line 32: Why has this been moved to imports instead of, as above, modules definitions?  I think this should stay where it was before.

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -224,9 +224,6 @@ impl<'a> Processor for NormalizeProcessor<'a> {
             return Ok(());
         }
 
-        // Check for required and non-empty values
-        schema::SchemaProcessor.process_event(event, meta, ProcessingState::root())?;
-
```
olksdr, line 229: Does the order when the schema processor was running and how the check runs now matters? 
Does it execute in the same order? Different?

olksdr (approved):
Overall lgtm. Left few questions.

relay-event-normalization/src/normalize/processor.rs
iker-barriocanal: The order is different: `SchemaProcessor` actions now run as part of processing child values. This doesn't matter, as `SchemaProcessor` adds remarks and/or meta errors on missing values. Decisions on rejecting events stay the same, and normalization steps won't be performed as the value is missing (same behavior).

I updated the PR description to reflect this.

relay-event-normalization/src/lib.rs
iker-barriocanal: In the codebase we separate `mod`s and `pub mod`s. Fwiw we also separate `pub mod`s and `pub use`s, which I haven't done in this PR, so happy to create a new section. What do you think?

PR #2743, opened 2023-11-20
Revert "feat(spool): Avoid evicting old project for incoming envelopes"

olksdr:
Reverts getsentry/relay#2581

This should be a temporary revert to validate the assumption that this change could cause the relay to keep the some of the incoming envelopes in memory for the longer period of time.  Which can cause very slow memory leak. 

related to: https://github.com/getsentry/team-ingest/issues/248

#skip-changelog


jjbayer (approved):
Could you add a description with why we're reverting this, and that the revert should be only temporary?

Issue #2717, opened 2023-11-13
[EPIC] Move the metrics cardinality limiter to Relay


jan-auer:
The performance of metrics queries depends on the number of time series they have to aggregate. To put an upper bound on the number of time series - and thus the number of metric buckets with a unique combination of tag values - the metrics ingestion pipeline has a **cardinality limiter**.

It resides in the metrics indexer consumer, currently causing more than half of its CPU usage and introducing additional latency due to a high number of Redis calls. There are two reasons to move the cardinality limiter out of the indexer consumers: 

1. The limiter currently does not apply to Relay.
2. The indexer consumers are a bottleneck and rely on low-latency high throughput.

# Stage 1

```[tasklist]
### Stage 1 / Re-Implementation
- [ ] https://github.com/getsentry/relay/issues/2719
- [x] Organize a Memstore/Redis Instance (We decided to use the existing Relay Redis for now)
- [x] Pass configuration via global config down from Sentry
- [ ] https://github.com/getsentry/sentry/pull/61028
- [ ] https://github.com/getsentry/sentry/pull/61024
- [ ] https://github.com/getsentry/getsentry/pull/12241
- [ ] https://github.com/getsentry/relay/pull/2745
- [x] ~~`SMISMEMBER` instead of loading the entire set~~ (Needs Redis 6.2)
- [ ] https://github.com/getsentry/relay/pull/2849
- [x] Hookup Cardinality Limiter into existing Processing Pipeline
- [x] Investiage alternative HashSet hasher, e.g. AHash for Redis implementation
- [ ] https://github.com/getsentry/snuba/pull/5349/
- [ ] https://github.com/getsentry/sentry-options-automator/pull/574
- [ ] https://github.com/getsentry/relay/pull/2947
- [ ] https://github.com/getsentry/sentry/pull/63562
- [ ] https://github.com/getsentry/relay/pull/2972
- [ ] https://github.com/getsentry/relay/pull/2990
- [ ] https://github.com/getsentry/relay/pull/2978
- [ ] https://github.com/getsentry/relay/pull/3008
- [ ] https://github.com/getsentry/relay/pull/3020
- [ ] https://github.com/getsentry/relay/pull/3018
- [ ] https://github.com/getsentry/relay/issues/3010
- [ ] https://github.com/getsentry/relay/pull/3029
- [ ] https://github.com/getsentry/relay/pull/3053
- [ ] https://github.com/getsentry/relay/pull/3071
[...]

PR #2745, opened 2023-11-20
feat(cardinality): Implement Redis set based cardinality limiter

Dav1dde:
Introduces a new `relay-cardinality` module, initially it is only housing the cardinality limiter, but at some point there may be more functionality related to cardinality (e.g. smart detection of high cardinality tags).

Implements a very basic cardinality limiter using one Redis set per organization and namespace to track cardinality.

API is still very much work in progress, this works for now but will definitely need to be changed in the future when the requirements for the cardinality limiter change. Outcomes, Tag erasure etc.

Epic: https://github.com/getsentry/relay/issues/2717


relay-metrics/src/aggregator.rs
```diff
@@ -891,11 +891,11 @@ impl fmt::Debug for Aggregator {
 /// Splits buckets into N logical partitions, determined by the bucket key.
 pub fn partition_buckets(
     project_key: ProjectKey,
-    buckets: Vec<Bucket>,
+    buckets: impl IntoIterator<Item = Bucket>,
     flush_partitions: Option<u64>,
 ) -> BTreeMap<Option<u64>, Vec<Bucket>> {
     let flush_partitions = match flush_partitions {
-        None => return BTreeMap::from([(None, buckets)]),
+        None => return BTreeMap::from([(None, buckets.into_iter().collect())]),
```
Dav1dde, line 898: I checked with [godbolt](https://godbolt.org/z/o41qrWbhj) this is optimized away when passing a vector already

relay-cardinality/src/limiter.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// Metric namespace.
+    pub namespace: MetricNamespace,
+    /// Hash of the metric.
+    pub hash: u32,
+
+    /// Implementation detail to optimize the implementation.
+    ///
+    /// Tracking the status directly on the entry means we do
+    /// not have to temporarily allocated a second data structure
```
```suggestion
    /// not have to temporarily allocate a second data structure
```

relay-cardinality/src/limiter.rs
```diff
@@ -0,0 +1,169 @@
...
+/// CardinalityLimiter configuration.
+#[derive(Debug, Clone)]
+pub struct Config {
+    /// The cardinality limit to enforce.
+    pub cardinality_limit: usize,
+}
+
+/// Cardinality Limiter enforcing cardinality limits on buckets.
+///
+/// Delegates enforcement to a [`Limiter`], see for example the
```
jjbayer: Dropped a line here?

relay-cardinality/src/limiter.rs
```diff
@@ -0,0 +1,169 @@
...
+
+/// A single entry to check cardinality for.
+#[derive(Clone, Copy, Debug)]
+pub struct Entry {
+    /// Opaque entry Id, used to keep track of indices and buckets.
+    pub id: EntryId,
+
+    /// Metric namespace.
+    pub namespace: MetricNamespace,
+    /// Hash of the metric.
```
```suggestion
    /// Hash of the metric name & tags.
```

relay-cardinality/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+publish = false
+
+[features]
+redis = ["relay-redis/impl"]
+
+[dependencies]
+hash32 = "0.3.1"
+itertools = { workspace = true }
+relay-log = { path = "../relay-log" }
+relay-metrics = { path = "../relay-metrics" }
```
jjbayer: nit: I think we should be able to avoid this dependency by having an API that works for any hashable item + a generic scope that can be serialized into a redis prefix.

Besides separation of concerns, this would make it easier to call the rate limiter from normalization in the future.

relay-cardinality/src/redis.rs
```diff
@@ -0,0 +1,427 @@
...
+            cmd.query(&mut self.0)
+                .map_err(relay_redis::RedisError::Redis)?;
+        }
+
+        redis::cmd("EXPIRE")
+            .arg(name)
+            .arg(expire)
+            // .arg("NX") - NX not supported for Redis < 7
+            .query(&mut self.0)
+            .map_err(relay_redis::RedisError::Redis)?;
```
jjbayer, line 265: Will this raise an error if the set was removed between redis commands? Or rather fail silently?

relay-cardinality/src/redis.rs
```diff
@@ -0,0 +1,427 @@
...
+
+        Ok(())
+    }
```
jjbayer, line 268: Should we use https://docs.rs/redis/latest/redis/struct.Pipeline.html here to reduce the number of requests? Or even a Lua script to get atomicity for `EXPIRE`.

jjbayer (commented):
I like the look of this! Just a bunch of random comments for now, will do a more in-depth review later.

relay-cardinality/Cargo.toml
Dav1dde: I agree, the API needs more work, it is intentionally stupid because I am not sure what the total set of requirements will be. So my approach was to do the dumbest thing first than gradually improve on it.

E.g. just accepting hashes isn't enough, it would need to include some kind of scope as well and we're not even sure yet what scopes we want to support or wether even the scope should be generic.

relay-cardinality/src/redis.rs
Dav1dde: I double checked, this fails silently and only raises an error when there is an issue with redis/the connection.

relay-cardinality/src/redis.rs
Dav1dde: Switched to a pipeline! I don't think we want/need an atomic pipeline, if the key already expired, we don't really care if the the expire command fails, because the window is no longer relevant.

relay-cardinality/src/limiter.rs
```diff
@@ -0,0 +1,169 @@
...
+            timestamp: UnixTimestamp::from_secs(0),
+            tags: BTreeMap::new(),
+            value: relay_metrics::BucketValue::Counter(0.0),
+            width: 0,
+        };
+
+        let accepted = self
+            .limiter
+            .check_cardinality_limits(organization, entries, self.config.cardinality_limit)?
+            .map(move |entry| std::mem::replace(&mut buckets[entry.id.0], EMPTY));
```
Dav1dde: The `swap` is a late change to be able to return owned buckets, initially the limiter took buckets by reference, but this allows us to just pass the iterator on into the partioning. 

This is a good sign that the API is not what it should be, but I think for now this is okay, I would like to use this to gather performance information and then re-work this when we come up with a plan on how to proceed with Stage 2.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 - Allow to ingest measurements on a span. ([#2792](https://github.com/getsentry/relay/pull/2792))
 - Extract size metrics for all resource spans when permitted. ([#2805](https://github.com/getsentry/relay/pull/2805))
 - Allow access to more fields in dynamic sampling and metric extraction. ([#2820](https://github.com/getsentry/relay/pull/2820))
+- Add Redis set based cardinality limiter. ([#2745](https://github.com/getsentry/relay/pull/2745))
```
```suggestion
- Add Redis set based cardinality limiter for metrics. ([#2745](https://github.com/getsentry/relay/pull/2745))
```

relay-cardinality/Cargo.toml
jjbayer: As discussed: The scope could be either a string that is then used as a redis key prefix, or a trait with a method that produces a prefix. But we can defer this to a later PR as you suggest.

relay-cardinality/src/limiter.rs
```diff
@@ -0,0 +1,169 @@
...
+pub trait Limiter {
+    /// Verifies cardinality limits.
+    ///
+    /// Returns an iterator containing only accepted entries.
+    fn check_cardinality_limits<I>(
+        &self,
+        organization: OrganizationId,
+        entries: I,
+        limit: usize,
+    ) -> Result<Box<dyn Iterator<Item = Entry>>>
```
jjbayer: Does this need to be `dyn`? Or could it be a type parameter?

relay-cardinality/src/limiter.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub(crate) fn is_accepted(&self) -> bool {
+        matches!(self.status, EntryStatus::Accepted)
+    }
+}
+
+/// CardinalityLimiter configuration.
+#[derive(Debug, Clone)]
+pub struct Config {
+    /// The cardinality limit to enforce.
+    pub cardinality_limit: usize,
```
jjbayer, line 88: IMO it would make sense to also have `time_window` in this config.

relay-cardinality/src/limiter.rs
jjbayer: Could `check_cardinality_limits` just toggle the entries in place? Then `accepted` could be a `.filter()` on the modified entries.

relay-cardinality/src/redis.rs
```diff
@@ -0,0 +1,429 @@
...
+use crate::{
+    limiter::{Entry, Limiter},
+    statsd::CardinalityLimiterCounters,
+    OrganizationId, Result,
+};
+
+/// Key prefix used for Redis keys.
+const REDIS_KEY_PREFIX: &str = "relay:cardinality";
+/// Maximum amount of arguments for a Redis `SADD` call.
+const REDIS_MAX_SADD_ARGS: usize = 1000;
```
jjbayer: nit: We're already in a module called `redis`, so we could remove the `REDIS_` prefix here.

relay-cardinality/src/redis.rs
```diff
@@ -0,0 +1,429 @@
...
+        Self {
+            redis,
+            window,
+            #[cfg(test)]
+            time_offset: 0,
+        }
+    }
+}
+
+impl RedisSetLimiter {
```
jjbayer:
```suggestion

```

relay-cardinality/src/redis.rs
```diff
@@ -0,0 +1,429 @@
...
+            counter(CardinalityLimiterCounters::RedisRead) += 1,
+            namespace = scope.namespace.as_str()
+        );
+        metric!(
+            counter(CardinalityLimiterCounters::RedisHashCheck) += entries.len() as i64,
+            namespace = scope.namespace.as_str()
+        );
+
+        let active_window = client
+            .read_set(&self.to_redis_key(&scope, self.window.active_time_bucket(timestamp)))?;
```
jjbayer: As discussed: We could go straight with `SMISMEMBERS` instead of downloading the entire set.

relay-cardinality/src/redis.rs
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    fn to_redis_key(&self, scope: &Scope, window: u64) -> String {
+        // Pattern match here, to not forget to update the key, when modifying the scope.
+        let Scope {
+            organization_id,
+            namespace,
+        } = scope;
+
+        format!("{REDIS_KEY_PREFIX}:scope-{{{organization_id}}}-{namespace}-{window}",)
```
jjbayer: Why are there escaped `{` in the key?

relay-cardinality/src/redis.rs
```diff
@@ -0,0 +1,429 @@
...
+        let mut client = self.redis.client()?;
+        let mut client = RedisHelper(client.connection()?);
+
+        let mut accepted = Vec::new();
+        for (scope, entries) in entries {
+            accepted.push(self.check_limits(&mut client, scope, entries, timestamp, limit)?);
+        }
+
+        Ok(Box::new(accepted.into_iter().flatten()))
+    }
```
jjbayer, line 220: Should we wrap this function body in a timer metric?

relay-server/src/actors/processor.rs
```diff
@@ -1261,13 +1281,49 @@ impl EnvelopeProcessorService {
...
 
+    fn check_cardinality_limits(
+        &self,
+        enable_cardinality_limiter: bool,
+        _organization_id: u64,
+        buckets: Vec<Bucket>,
+    ) -> Result<Box<dyn Iterator<Item = Bucket>>, relay_cardinality::Error> {
+        if !enable_cardinality_limiter {
+            return Ok(Box::new(buckets.into_iter()));
+        }
```
jjbayer, line 1292: nit: Why not check the flag outside of this function? Seems more straight-forward than passing a boolean that makes the function behave like a noop (but this might just be my personal preference).

relay-server/src/actors/processor.rs
```diff
@@ -1261,13 +1281,49 @@ impl EnvelopeProcessorService {
         }
     }
 
+    fn check_cardinality_limits(
+        &self,
+        enable_cardinality_limiter: bool,
+        _organization_id: u64,
```
jjbayer, line 1287: nit: Should make this function `#[cfg(feature = "processing")]`?

```suggestion
        organization_id: u64,
```

relay-cardinality/src/redis.rs
```diff
@@ -0,0 +1,429 @@
...
+            );
+        }
+
+        if !new_hashes.is_empty() {
+            for time_bucket in self.window.iter(timestamp) {
+                let key = self.to_redis_key(&scope, time_bucket);
+
+                // The expiry is a off by `window.granularity_seconds`,
+                // but since this is only used for cleanup, this is not an issue.
+                client.add_to_set(&key, &new_hashes, self.window.window_seconds)?;
```
jjbayer: We could start the redis pipeline outside of this loop and send all commands at once.

relay-cardinality/Cargo.toml
Dav1dde: Addressed with a new trait now `CardinalityItem`

relay-cardinality/src/limiter.rs
Dav1dde: Addressed now by tracking rejections in `BTreeSet` and an iterator which can use it to skip over rejections

relay-cardinality/src/limiter.rs
Dav1dde: Yes unfortunately, since this is the return value. I thought we can return `impl Trait` from traits since 1.74 but apparantely this isn't out yet.

relay-cardinality/src/limiter.rs
Dav1dde: It is an implementtion detail of the redis based implementation, that's why it is not in this config. But we may want to revisit the split of types in general.

relay-cardinality/src/redis.rs
Dav1dde: I did it to split public interface from private, but removed now

relay-cardinality/src/redis.rs
Dav1dde: Redis will use anything between `{}` for sharding, so one entire org is always on the same redis shard/node.

For everything executed in one pipeline and script it has to be executed on the same redis node/shard, this allows us to potentially do that.

relay-server/src/actors/processor.rs
Dav1dde: I implemented the function pretty much only for this check, so I can early return, otherwise it becomes very complex with the `cfg(feature = "processing")`

relay-server/src/actors/processor.rs
Dav1dde: Unfortunately this is needed for non-processing, the entire function just exists to handle the `if` case with an early return to make the code somewhat readable and not a pure if else cfg mess.

relay-cardinality/src/redis.rs
Dav1dde: Can the pipeline become too big?

relay-cardinality/src/redis.rs
Dav1dde: Unfortunately, we cannot do this, `SMISMEMBERS` was added in 6.2 and we still have instances on 5.X, another reason to look into the lua script

relay-cardinality/src/redis.rs
jjbayer: Oh nice, I had no idea!

relay-server/src/actors/processor.rs
jjbayer: Is it still a mess with the `if_processing!` macro?

relay-cardinality/src/redis.rs
jjbayer: Good question, did not consider that.

relay-cardinality/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! Metrics Cardinality Limiter
+//! Relay Cardinality Module
```
jjbayer, line 1: Not necessarily for this PR, but since this is a new crate, we should take a look at what `make doc-rust` produces and whether we're happy with the documentation. Other crates have a summary of what they do in the crate-level docs, e.g. https://github.com/getsentry/relay/blob/a9e7ef5e19918927a001d068391fa901f7841e9c/relay-metrics/src/lib.rs#L1-L64.

jjbayer (approved):
Let's ship it! Before opting in all orgs we should transition from copying entire redis sets to a Lua script, as you suggested.

relay-cardinality/src/lib.rs
Dav1dde: I will definitely do this in a follow up, I also want to get a proper documentation in for how the limiter works!

relay-cardinality/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+description = "Metrics Cardiniality Limiter"
+homepage = "https://getsentry.github.io/relay/"
+repository = "https://github.com/getsentry/relay"
+version = "23.10.1"
+edition = "2021"
+license-file = "../LICENSE"
+publish = false
+
+[features]
+redis = ["relay-redis/impl"]
```
olksdr: Should there be a `default` feature as well without redis? 

```suggestion
[features]
default = []
redis = ["relay-redis/impl"]
```

relay-cardinality/src/limiter.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Transforms this item into a consistent hash.
+    fn to_hash(&self) -> u32;
+
+    /// Extracts a scope to check cardinality limits for from this item.
+    fn to_scope(&self) -> Option<Self::Scope>;
+}
+
+// Implementation for a bucket, currently here, may also be implemented in `relay-metrics`, if it
+// makes sense to move it.
+impl CardinalityItem for Bucket {
```
olksdr: Yeah, having this implemented in metrics in my opinion would be cleaner.  Since relay-metrics would be responsible  to provide the implementation and this crate just provides the facilities for the entire thing to work.

relay-config/src/config.rs
```diff
@@ -1262,6 +1262,43 @@ pub struct GeoIpConfig {
...
+    /// sliding window, configure what the granularity of that window is.
+    ///
+    /// If this is equal to `window_seconds`, the quota resets to 0 every
+    /// `window_seconds`.  If this is a very small number, the window slides
+    /// "more smoothly" at the expense of having much more redis keys.
+    ///
+    /// The number of redis keys required to enforce a quota is `window_seconds /
+    /// granularity_seconds`.
+    ///
+    /// Defaults to: 10 minutes.
```
olksdr, line 1283: it seems that default is set to 60 seconds only.

Dav1dde: good catch, thanks, phew!

relay-cardinality/Cargo.toml
Dav1dde: Isn't that the implicit default behaviour, you always have a default feature which is empty?

relay-cardinality/src/limiter.rs
Dav1dde: I'll move it!

relay-cardinality/Cargo.toml
olksdr: I meant to have it explicitly.  We have it always defined in our crates. I'm not sure what the team prefer. Explicit or implicit more.

Dav1dde: Done!

PR #2746, opened 2023-11-20
Add DE Deployments

rgibert:
Add DE and it's POPs instances to the GoCD deployment logic.  Deployments to DE won't start until https://github.com/getsentry/gocd-jsonnet/pull/59 is merged.

#skip-changelog


gocd/templates/libs/relay-pops.libsonnet
```diff
@@ -70,8 +79,8 @@ local generic_pops_stage(region) =
 // of clusters, other regions only deploy to a single cluster.
 {
   stages(region)::
-    if region == 'us' then
-      [us_pops_stage()]
+    if region == 'us' || region == 'de' then
+      [pops_stage()]
```
olksdr: I think thi function should get a `region` input
```suggestion
      [pops_stage(region)]
```

rgibert: Thanks for catching that, fixed.

olksdr (approved):
lgtm

PR #2749, opened 2023-11-21
feat(spans): Base group on truncated description

jjbayer:
The span group was previously a hash of the full scrubbed span description, not the truncated description that ends up in the description tag. We want to change this because of two reasons:

* It is confusing to see two rows (i.e. groups) with the same description in the UI.
* Reduce cardinality.

NOTE: Based on the following query, this will change the group hash of ~13% of all span groups:

```sql
SELECT
uniq(tags.raw_value[indexOf(tags.key, 9223372036854775808 + 249)]) AS `span.description`,
length(tags.raw_value[indexOf(tags.key, 9223372036854775808 + 249)]) == 200 AS `full_length`
FROM generic_metric_distributions_aggregated_dist
WHERE (granularity = 2)
AND (timestamp >= now() - INTERVAL 2 HOUR)
AND (metric_id IN [9223372036854776214])
GROUP BY `full_length`
ORDER BY `span.description` DESC
LIMIT 100
```

ref: [internal issue](https://www.notion.so/sentry/Compute-group-hash-from-truncated-scrubbed-description-d23718a39d364f6ebb8198abd550105c?pvs=4)


olksdr (approved):

iker-barriocanal (approved):

jjbayer:
We concluded that we would not save much by performing this change, and that it might be preferable from the product perspective to have separate groups even if the truncated description looks the same.

PR #2753, opened 2023-11-22
fix(server): Avoid a race condition in dequeuing after global config

jan-auer:
`RequestUpdate` cannot be called directly, we need to go through
`Project::get_or_fetch_state` to ensure that internal invariants are not
violated. In this instance, the message does not create a required channel in
the `Project` instance.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/project_cache.rs
+++ b/relay-server/src/actors/project_cache.rs
@@ -498,7 +498,7 @@ impl ProjectCacheBroker {
         };
 
         relay_log::info!(
-            "global config received: {} project were pending",
+            "global config received: {} projects were pending",
             project_keys.len()
         );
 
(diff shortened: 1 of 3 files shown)


olksdr (approved):
:shipit:

relay-server/src/actors/project_cache.rs
```diff
@@ -490,23 +490,34 @@ impl GlobalConfigStatus {
...
-                let message = RequestUpdate {
-                    project_key,
-                    no_cache: false,
-                };
+        ) else {
+            return;
+        };
+
+        relay_log::info!(
+            "global config received: {} project were pending",
```
TBS1996:
```suggestion
            "global config received: {} projects were pending",
```

TBS1996 (approved):

PR #2754, opened 2023-11-22
ref(normalization): Restore SchemaProcessor

iker-barriocanal:
Reverts https://github.com/getsentry/relay/commit/e2f51fe646872d7ba193309508f3b2f27a4a4368.

After a short discussion with the team, we've agreed to have a processor that calls others processors instead of doing everything. This will make dealing with specific features for the whole event easier, instead of an on-visitor-item basis. This PR doesn't change any behavior.

#skip-changelog


olksdr (approved):
Looks ok. 
Just some imports hast to be fixed I think.

PR #2755, opened 2023-11-22
feat(spans): Sanitize file extension

jjbayer:
* Sanitize file extensions like `.js~some~data` to `.js`. 
* Drop file extensions that are longer than 10 chars after sanitation.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -271,11 +271,7 @@ fn scrub_resource_path(path: &str) -> String {
     }
 
     // Only accept short, clean file extensions.
-    if let Some((invalid, _)) = extension
-        .bytes()
-        .enumerate()
-        .find(|(_, c)| !c.is_ascii_alphanumeric())
-    {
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

TBS1996 (approved):

olksdr (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -267,6 +270,18 @@ fn scrub_resource_path(path: &str) -> String {
         extension = "";
     }
 
+    // Only accept short, clean file extensions.
+    if let Some((invalid, _)) = extension
+        .bytes()
+        .enumerate()
+        .find(|(_, c)| !c.is_ascii_alphanumeric())
```
Dav1dde: [`.position(fn)`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position) returns an index directly

Dav1dde (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
jjbayer: Thanks!

PR #2756, opened 2023-11-22
feat(webvitals): Update normalize_performance_score to accept components with 0 weight and automatically normalize weights to a sum of 1

edwardgou-sentry:
- Updates `normalize_performance_score` to accept a `PerformanceScoreProfile` with components of 0 weight. This is needed because we must store 0 weight for unused component to ensure that the sum of all `avg(measurements.score.weight.*)` equals `1` at any given time. Not storing 0 allows the possibility that the total sum of avg weights exceeds 1.
- Updates `normalize_performance_score` to accept a `PerformanceScoreProfile` with a component weights sum of any number greater than 0. Previously, we expected the component weights from the `PerformanceScoreProfile` config to always sum up to 1. This change automatically normalizes the stored component weights so that the sum of stored weights will always equal 1.

### DIFF
--- a/relay-event-normalization/src/normalize/processor.rs
+++ b/relay-event-normalization/src/normalize/processor.rs
@@ -886,11 +886,14 @@ fn normalize_performance_score(
                 let mut score_total = 0.0;
                 let mut weight_total = 0.0;
                 for component in &profile.score_components {
+                    weight_total += component.weight;
+                }
+                for component in &profile.score_components {
+                    let normalized_component_weight = component.weight / weight_total;
                     if let Some(value) = measurements.get_value(component.measurement.as_str()) {
                         let cdf = utils::calculate_cdf_score(value, component.p10, component.p50);
-                        let component_score = cdf * component.weight;
+                        let component_score = cdf * normalized_component_weight;
                         score_total += component_score;
-                        weight_total += component.weight;
                         should_add_total = true;
 
                         measurements.insert(
@@ -905,7 +908,7 @@ fn normalize_performance_score(
                     measurements.insert(
                         format!("score.weight.{}", component.measurement),
                         Measurement {
-                            value: component.weight.into(),
+                            value: normalized_component_weight.into(),
                             unit: (MetricUnit::Fraction(FractionUnit::Ratio)).into(),
                         }
                         .into(),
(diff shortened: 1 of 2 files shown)


relay-event-normalization/src/normalize/processor.rs
```diff
@@ -878,19 +878,22 @@ fn normalize_performance_score(
                 if !profile
                     .score_components
                     .iter()
-                    .all(|c| measurements.contains_key(c.measurement.as_str()))
+                    .all(|c| measurements.contains_key(c.measurement.as_str()) || c.weight == 0.0)
```
Dav1dde: So this float comparison is tricky. While the comparison here actually does work in most cases (even for `-0.0 == 0.0`) we do not actually know the value of `c.weight`. If weight is purely deserialized this code will probably work as intended, but the moment it is a result of a calculation, we cannot be sure 0 is actually 0.

Example:

```rs
let f: f32 = 0.3 / (0.9 / 3.0) - 1.0;
println!("{:.32} is 0? {}", f, f == 0.0);
```

This is trivially correct, but the code prints false.

jjbayer: nit: Took me a while to understand that this means "all measurements with a profile weight greater zero are required". Might be worth adding a comment about it, or rewriting as `.any()` (flip the condition).

jjbayer (approved):

relay-event-normalization/src/normalize/processor.rs
jjbayer: Agree with this, though we do `==` comparisons with floats parsed from config in other places as well, e.g.

https://github.com/getsentry/relay/blob/b239d2b09c9fcf39ccc40c4d0ac9b9cd8bd812ef/relay-sampling/src/evaluation.rs#L280

Dav1dde: Solution for 0 is actually pretty short: `f.abs() <= f32::EPSILON` (though a lot harder to grasp the intent)

iker-barriocanal (approved):

relay-event-normalization/src/normalize/processor.rs
edwardgou-sentry: Updated to make use of `f64::EPSILON`

edwardgou-sentry: Updated! I reversed the `all` to `any` and updated the comment

Issue #2716, opened 2023-11-13
[EPIC] Add throughput rate limits for metric buckets


jan-auer:
The metrics consumers scale with bucket throughput. As there have been challenges to scale the pipeline lately, we're looking to limit the throughput to what the pipeline can handle for the interim.

In longer run, the pipeline should scale to any volume needed. The rate limiter will then primarily ensure cost-effective metrics use cases by placing a reasonable limit on the product.

## Validation

- Metrics:
  - Number of buckets rate limited by scope.
  - Metrics messages sent to Kafka (already exists in Relay)
  - Metrics messages as seen by the consumers
- Alerts
  - Rate Limit hit
  - [Optional] Rate Limit capacity at X%

## Design

- Scoped Limits:
  - Global
  - Namespaces Total
  - Org Total
  - Namespace per Org
- Goal is to protect Kafka throughput -> Rate Limit by Amount of Buckets, not Values/Datapoints
- Use existing abuse quota system from Sentry to communicate down to relay

## Technical Design

- Extend Scopes to support namespaces and global, includes redis key
- Add Rate Limiting to the EnvelopeProcessor, after splitting
- Log outcomes for discarded/rate-limited Buckets
- Rate limit on batches not on buckets
  - Discard all remaining batches if one hits the rate limiter
  - Possibly: rate limit already on the entire flush, splitting will only increase the number of total buckets

## Limitations

- Best Effort rate limiting, we don't need exact rate limits and consistency. Implementation may be consistent.
- Processing Relays only
- No propagation of rate limits, because we can only propagate Rate Limit Backoffs (Stop sending me stuff now). To be changed later.

## Rollout

- Load Test (on Loadtesting Infra)
- S4S
- Enable it in Prod

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/relay/pull/2758
- [ ] https://github.com/getsentry/sentry/pull/61666
- [ ] https://github.com/getsentry/relay/pull/2928
- [ ] https://github.com/getsentry/relay/pull/2941
- [ ] https://github.com/getsentry/sentry/pull/64574
- [x] Test global abuse quota in S4S
- [ ] https://github.com/getsentry/sentry-options-automator/pull/1117
- [x] Implement rate limiting with org/namespace scope for metrics - https://github.com/getsentry/relay/pull/3090
- [ ] https://github.com/getsentry/sentry/pull/68686
```

```[tasklist]
### Followups
[...]

PR #2758, opened 2023-11-23
feat(server): Org rate limit per metric bucket

TBS1996:
part of: https://github.com/getsentry/relay/issues/2716

in order to protect our kafka metric consumers, we want to have a way of rate limiting based on the amount of buckets, as that's what decides the load placed on our kafka topics. 

We are starting out with just the org throughput limits but will be expanded upon further as outlined in the linked epic.


relay-server/src/actors/processor.rs
```diff
@@ -1926,8 +1930,70 @@ impl EnvelopeProcessorService {
             path: "".to_owned(),
             project_id: Some(scoping.project_id),
         };
-        for (partition_key, buckets) in partition_buckets(scoping.project_key, buckets, partitions)
+
+        let bucket_qty = buckets.len();
+        let bucket_partitions = partition_buckets(scoping.project_key, buckets, partitions);
+
+        let quotas = project_state.config.quotas.as_slice();
```
```rs
let quotas = &project_state.config.quotas;
```

relay-server/src/actors/processor.rs
```diff
@@ -1926,8 +1930,70 @@ impl EnvelopeProcessorService {
...
+        let item_scoping = ItemScoping {
+            category: DataCategory::Metrics,
+            scoping: &scoping,
+        };
+
+        if cached_rate_limits
+            .check_with_quotas(quotas, item_scoping)
+            .is_limited()
         {
+            relay_log::info!("dropping {bucket_qty} buckets due to throughput ratelimit");
```
Dav1dde: Shouldn't we still have outcomes for this, Metrics, Transactions and Profiles?

relay-server/src/actors/processor.rs
```diff
@@ -1926,8 +1930,70 @@ impl EnvelopeProcessorService {
...
+                        remote_addr: None,
+                        category: DataCategory::Metrics,
+                        quantity: total_batches as u32,
+                    });
+
+                    self.inner
+                        .project_cache
+                        .send(UpdateRateLimits::new(scoping.project_key, limits));
+
+                    relay_log::info!("dropping {bucket_qty} buckets due to throughput ratelimit");
```
Dav1dde: No Outcomes for Transactions/Profiles?

relay-server/src/actors/processor.rs
```diff
@@ -1926,8 +1930,70 @@ impl EnvelopeProcessorService {
...
+        let bucket_qty = buckets.len();
+        let bucket_partitions = partition_buckets(scoping.project_key, buckets, partitions);
+
+        let quotas = project_state.config.quotas.as_slice();
+        let item_scoping = ItemScoping {
+            category: DataCategory::Metrics,
+            scoping: &scoping,
+        };
+
+        if cached_rate_limits
```
Dav1dde, line 1375: I think we should move all the rate limit checks to a separate function, this `handle_encode_metrics` is getting quite crowded.

relay-server/src/actors/processor.rs
```diff
@@ -1992,7 +2061,47 @@ impl EnvelopeProcessorService {
...
+            scoping,
+            bucket_qty,
+            &bucket_partitions,
+            max_batch_size_bytes,
+            project_state,
+        ) {
+            return;
+        }
 
+        #[cfg(feature = "processing")]
```
Dav1dde: This is not good^^

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -106,6 +106,21 @@ pub struct TransactionCount {
     pub has_profile: bool,
 }
 
+pub fn get_transaction_and_profile_count(
+    buckets: &[Bucket],
```
Dav1dde: This can take a `BucketsView` then you also don't have to create a `BucketView` manually.

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -106,6 +106,21 @@ pub struct TransactionCount {
     pub has_profile: bool,
 }
 
+pub fn get_transaction_and_profile_count(
```
Dav1dde: This function already kind-of exists in the processors `create_metrics_item` and uses `SourceQuantities` to count, we can re-use this instead.

relay-server/src/actors/processor.rs
```diff
@@ -1992,7 +2061,47 @@ impl EnvelopeProcessorService {
...
+            extraction_mode,
+            project_state,
+            rate_limits: cached_rate_limits,
+        } = message;
+
+        let partitions = self.inner.config.metrics_partitions();
+        let max_batch_size_bytes = self.inner.config.metrics_max_batch_size_bytes();
+
+        let upstream = self.inner.config.upstream_descriptor();
+        let dsn = PartialDsn {
```
Dav1dde: We have `PartialDsn::outbound` now (when you update master)

Dav1dde (commented):
Just some quick nits, I'll have to look into the outcome logging more carefully later.

I would like if we didn't have to duplicate this and somehow can share as much code as possible with the `ManagedEnvelope` logic (when we drop the envelope transactions and profile outcomes are created). But I'll need more time to think about this.

relay-server/src/actors/processor.rs
TBS1996: wait, that wasnt supposed to be there...

relay-server/src/utils/metrics_rate_limits.rs
TBS1996: alright! will check that. 

unrelated to this pr, but i guess the same logic applies to MetricsLimiter::create?

relay-server/src/actors/processor.rs
```diff
@@ -1494,11 +1514,165 @@ impl EnvelopeProcessorService {
...
+            ExtractionMode::from_usage(usage)
+        };
+
+        let quotas = &project_state.config.quotas;
+        let item_scoping = ItemScoping {
+            category: DataCategory::Metrics,
+            scoping: &scoping,
+        };
+
+        if let Some(rate_limiter) = self.inner.rate_limiter.as_ref() {
```
Dav1dde: Can return early here, saves on indentation and mental overhead.

```rs
let Some(rate_limiter) = self.inner.rate_limiter.as_ref() else { return false; }
```

relay-server/src/actors/processor.rs
```diff
@@ -1494,11 +1514,165 @@ impl EnvelopeProcessorService {
...
+                );
+
+                return true;
+            }
+
+            match rate_limiter.is_rate_limited(quotas, item_scoping, total_batches, false) {
+                Ok(limits) if limits.is_limited() => {
+                    let timestamp = UnixTimestamp::now().as_datetime().unwrap_or_else(Utc::now);
+                    let reason_code = limits.longest().and_then(|limit| limit.reason_code.clone());
+                    self.inner.outcome_aggregator.send(TrackOutcome {
```
Dav1dde: Doesn't this get tracked twice now, with the call to `drop_buckets_with_outcomes` below?

Also the `reason_code` is calculated twice?

relay-server/src/actors/processor.rs
```diff
@@ -1507,8 +1681,22 @@ impl EnvelopeProcessorService {
         let upstream = self.inner.config.upstream_descriptor();
         let dsn = PartialDsn::outbound(&scoping, upstream);
 
-        for (partition_key, buckets) in partition_buckets(scoping.project_key, buckets, partitions)
-        {
+        let _bucket_qty = buckets.len();
+        let bucket_partitions = partition_buckets(scoping.project_key, buckets, partitions);
+
+        #[cfg(feature = "processing")]
```
Dav1dde, line 1445: I think it makes sense to check the cached rate limits even on non-processing relays

relay-server/src/actors/processor.rs
TBS1996: ah yeah, missed it while moving code to the function, thanks!

relay-server/src/actors/processor.rs
TBS1996: well atm, they just get updated from redis, which isnt available on non-processing relays, so im not sure if that would make sense

Dav1dde: I am not sure how propagation of quotas/limits to pops is exactly working, but I don't see a reason why we cant check it now and once we actually have a working propagation everything just works

relay-server/src/actors/processor.rs
```diff
@@ -1494,11 +1514,157 @@ impl EnvelopeProcessorService {
...
+                total_batches,
+                scoping,
+                bucket_partitions,
+                mode,
+            );
+
+            return true;
+        }
+
+        match rate_limiter.is_rate_limited(quotas, item_scoping, total_batches, false) {
```
jjbayer: Is this where the counter in redis is updated? Might be worth adding a comment about it.

relay-server/src/actors/processor.rs
```diff
@@ -1494,11 +1514,158 @@ impl EnvelopeProcessorService {
...
+        max_batch_size_bytes: usize,
+        project_state: Arc<ProjectState>,
+    ) -> bool {
+        let mode = {
+            let usage = match project_state.config.transaction_metrics {
+                Some(ErrorBoundary::Ok(ref c)) => c.usage_metric(),
+                _ => false,
+            };
+            ExtractionMode::from_usage(usage)
+        };
```
iker-barriocanal, line 1355: I haven't tried, but would something like this work?
```suggestion
        let mode = ExtractionMode::from_usage(
            project_state.config.transaction_metrics.map_or(false, |ErrorBoundary::Ok(ref c)| c.usage_metric())
        );
```

relay-server/src/actors/processor.rs
```diff
@@ -1494,11 +1514,158 @@ impl EnvelopeProcessorService {
...
+            .is_limited()
+        {
+            relay_log::info!("dropping {bucket_qty} buckets due to throughput ratelimit");
+            let reason_code = cached_rate_limits
+                .longest()
+                .and_then(|limit| limit.reason_code.clone());
+
+            self.drop_buckets_with_outcomes(
+                reason_code,
+                total_batches,
```
iker-barriocanal: What's the difference between `total_batches` and `bucket_qty` (function param), the maximum size of the bucket? Can we use just one of the two? It seems odd to me that sometimes we use one and sometimes the other (see when emitting outcomes and on log messages).

relay-server/src/actors/processor.rs
```diff
@@ -1494,11 +1514,158 @@ impl EnvelopeProcessorService {
...
+        }
+
+        self.inner.outcome_aggregator.send(TrackOutcome {
+            timestamp,
+            scoping,
+            outcome: Outcome::RateLimited(reason_code),
+            event_id: None,
+            remote_addr: None,
+            category: DataCategory::Metrics,
+            quantity: total_batches as u32,
```
iker-barriocanal: Should we exclude buckets with the `usage` metric from rate-limited outcomes? These metrics are data we (sentry) generate but the user doesn't care about, and I assume these outcomes will be presented to them.

iker-barriocanal (commented):
Does this PR rate-limit buckets on a binary basis? Do we want to consider scenarios in which some buckets are rate-limited but others aren't?

Dav1dde:
@iker-barriocanal 

> Does this PR rate-limit buckets on a binary basis? Do we want to consider scenarios in which some buckets are rate-limited but others aren't?

This is on the roadmap but not planned for this PR, the idea is to rate limit based on namespace (there is  some info in the epic: https://github.com/getsentry/relay/issues/2716)

relay-server/src/actors/processor.rs
Dav1dde: `bucket_qty` is the amount of total buckets in the flush, `total_batches` is the amount of batches we are writing upstream (aka. amount of requests to upstream or amount of messages in the kafka topic).

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Return global config ready status to downstream relays. ([#2765](https://github.com/getsentry/relay/pull/2765))
 - Add Mixed JS/Android Profiles events processing. ([#2706](https://github.com/getsentry/relay/pull/2706))
 - Allow to ingest measurements on a span. ([#2792](https://github.com/getsentry/relay/pull/2792))
+- Org rate limit metrics per batch. ([#2758](https://github.com/getsentry/relay/pull/2758))
```
jan-auer: - Let's update the entry to reflect that this is actually rate limiting per bucket
- I would propose moving this to "Internal"

relay-cabi/include/relay.h
```diff
@@ -97,6 +96,10 @@ enum RelayDataCategory
    * TODO(jferg): Rename this to UserFeedback once old UserReport is deprecated.
    */
   RELAY_DATA_CATEGORY_USER_REPORT_V2 = 14,
+  /**
+   * Metric bucket.
+   */
+  RELAY_DATA_CATEGORY_METRIC_BUCKET = 15,
```
jan-auer, line 102: Note that this will require a release of the python library and bump in Sentry and Snuba.

jan-auer (commented):
This looks good to me. Could you add an integration test?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 - Return global config ready status to downstream relays. ([#2765](https://github.com/getsentry/relay/pull/2765))
 - Add Mixed JS/Android Profiles events processing. ([#2706](https://github.com/getsentry/relay/pull/2706))
 - Allow to ingest measurements on a span. ([#2792](https://github.com/getsentry/relay/pull/2792))
+- Add size limits on metric related envelope items. ([#2800](https://github.com/getsentry/relay/pull/2800))
+- Include the size offending item in the size limit error message. ([#2801](https://github.com/getsentry/relay/pull/2801))
```
Dav1dde: These changes shouldnt be here

relay-server/src/actors/processor.rs
```diff
@@ -1261,11 +1280,158 @@ impl EnvelopeProcessorService {
...
+        bucket_partitions: &BTreeMap<Option<u64>, Vec<Bucket>>,
+        mode: ExtractionMode,
+    ) {
+        let mut source_quantities = SourceQuantities::default();
+
+        for buckets in bucket_partitions.values() {
+            source_quantities += source_quantities_from_buckets(&BucketsView::new(buckets), mode);
+        }
+
+        let timestamp = UnixTimestamp::now().as_datetime().unwrap_or_else(Utc::now);
```
Dav1dde: Is this not just `Utc::now()` why the roundtrip through `UnixTimestamp`?

tests/integration/test_store.py
```diff
@@ -499,6 +499,81 @@ def transform(e):
...
+            "scopeId": str(key_id),
+            "categories": ["metric_bucket"],
+            "limit": metric_bucket_limit,
+            "window": 86400,
+            "reasonCode": "throughput rate limiting",
+        }
+    ]
+
+    def generate_ticks():
+        # Generate a new timestamp for every bucket, so they do not get merged by the aggregator
```
Dav1dde, line 543: We can also use tags for this, should be easier than having timestamps tied to `bucket_interval`

Dav1dde (approved):

PR #2760, opened 2023-11-23
ref(normalization): Restore TransactionsProcessor

iker-barriocanal:
Restores the TransactionsProcessor, as we're moving towards separating the steps in a different way. This PR doesn't have any functional changes and just moves a few things around; the large diff is caused by renaming `transactions/utils.rs` to `transactions/processor.rs` and then reintroducing the transactions processor.

#skip-changelog


jjbayer (approved):
Rubber stamping this, I trust that there are no functional changes.

olksdr (approved):

PR #2761, opened 2023-11-23
feat(spans): Extract main_thread tag for spans

shruthilayaj:
Extracts main_thread tag for spans based on span[data][thread.name].
Right now, we only expect to extract this if thread name is `main`.

#skip-changelog


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -384,6 +388,17 @@ pub(crate) fn extract_tags(
...
     }
 
+    if let Some(thread_name) = span
+        .data
+        .value()
+        .and_then(|data| data.get("thread.name"))
+        .and_then(|value| value.as_str())
+    {
+        if thread_name == MAIN_THREAD_NAME {
+            span_tags.insert(SpanTagKey::MainTread, "true".to_owned());
```
jjbayer: Can we limit this to `MOBILE_SDKS.contains(...)`, like the `Mobile` tag?

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -83,6 +86,7 @@ impl SpanTagKey {
             SpanTagKey::TimeToFullDisplay => "ttfd",
             SpanTagKey::TimeToInitialDisplay => "ttid",
             SpanTagKey::FileExtension => "file_extension",
+            SpanTagKey::MainTread => "main_thread",
```
jjbayer, line 89: If we need this tag on metrics as well, please add it to the metrics config as well, for example

https://github.com/getsentry/relay/blob/9bf1570bfe712a1e9a3540f6c2849af546ea8307/relay-dynamic-config/src/defaults.rs#L161-L162

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer, line 1: If we need this tag on metrics as well, please add it to the metrics config, for example

https://github.com/getsentry/relay/blob/9bf1570bfe712a1e9a3540f6c2849af546ea8307/relay-dynamic-config/src/defaults.rs#L161-L162

Because there's already a `span.main_thread` metrics tag, let's add that prefix to the metrics config, i.e. `("span.", "main_thread")`. https://github.com/getsentry/sentry/blob/8d216f916c08e0e40cfd3427fed560a8dd1ce591/src/sentry/sentry_metrics/indexer/strings.py#L170

shruthilayaj: addressed in https://github.com/getsentry/relay/pull/2761/commits/ae31b460004a39b98eb08f79314f4c5b5ff5a39b

relay-event-normalization/src/normalize/span/tag_extraction.rs
shruthilayaj: yup 👍, https://github.com/getsentry/relay/pull/2761/commits/ae31b460004a39b98eb08f79314f4c5b5ff5a39b

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -128,6 +132,7 @@ pub(crate) fn extract_span_tags(event: &mut Event, config: &Config) {
     // TODO: To prevent differences between metrics and payloads, we should not extract tags here
     // when they have already been extracted by a downstream relay.
     let shared_tags = extract_shared_tags(event);
+    let sdk_name = event.sdk_name().to_owned();
```
jjbayer: nit: It would be slightly cheaper to set `let is_mobile = MOBILE_SDKS.contains(event.sdk_name())` here, and plug that into `extract_tags`.

Or even reuse the value that `extract_shared_tags` wrote to `shared_tags.get(SpanTagKey::Mobile)`.

jjbayer (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
shruthilayaj: good idea! just passing an is_mobile boolean now https://github.com/getsentry/relay/pull/2761/files#diff-313f39c52849630c10c4f6da270c2ed0f99d7deac8894a838bd166dddc94b19dR135

relay-event-normalization/src/normalize/utils.rs
```diff
@@ -59,6 +59,9 @@ pub const MOBILE_SDKS: [&str; 4] = [
     "sentry.javascript.react-native",
 ];
 
+/// Allowed value for main thread name
```
jjbayer:
```suggestion
/// Allowed value for main thread name.
```

jjbayer (approved):

PR #2762, opened 2023-11-23
feat(webvitals): Move back to storing weighted Performance scores instead of cdf

edwardgou-sentry:
Updates `normalize_performance_score` to go back to storing weighted performance scores.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/processor.rs
+++ b/relay-event-normalization/src/normalize/processor.rs
@@ -795,7 +795,7 @@ fn normalize_performance_score(
                         measurements.insert(
                             format!("score.{}", component.measurement),
                             Measurement {
-                                value: cdf.into(),
+                                value: component_score.into(),
                                 unit: (MetricUnit::Fraction(FractionUnit::Ratio)).into(),
                             }
                             .into(),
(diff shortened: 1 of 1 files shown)


edwardgou-sentry:
Sorry for the reversal, turns out that we can't put together the `avg` the cdf and weights since it won't properly form the correct aggregate scores

jjbayer (approved):
Did not validate the math. I guess we have to be careful when querying over time ranges where the definition of this value changed? Or will we just create a new profile?

edwardgou-sentry:
> Did not validate the math. I guess we have to be careful when querying over time ranges where the definition of this value changed? Or will we just create a new profile?

Yeah it has changed a few times now, just need to be aware of the timeranges. This is only being used by the sentry org as well so the impact should be minimal.

edwardgou-sentry:
> Did not validate the math. I guess we have to be careful when querying over time ranges where the definition of this value changed? Or will we just create a new profile?

Yeah, it has changed a few times now, so just need to be aware of the timeranges. This is only being used by the Sentry org as well, so there should be less impact.

PR #2751, opened 2023-11-21
feat(metric-meta): Add support for metric metadata

Dav1dde:
Implements `metric_meta` envelope item and storage in Redis.

Epic: https://github.com/getsentry/sentry/issues/60260

PR #2763, opened 2023-11-24
ref(metrics): Allow MRI to be owned

Dav1dde:
This allows us to pass a MRI as an owned structure and allows for less conversions between String <-> MRI down the line. This will also make it possible to implement Deserialize for the MRI

Needed for:  #2751

#skip-changelog


jan-auer (approved):

olksdr (approved):
:shipit:

PR #2765, opened 2023-11-27
feat(server): Return global config status for downstream requests

TBS1996:
follow-up to https://github.com/getsentry/relay/pull/2697

we now have a mechanism to delay envelope processing until a global config is available, however, we currently send back a default global config if one is requested from downstream, which kind of defeats the purpose if the downstream relay believes it has a valid global config but it doesn't. 
 
To fix this, this pr will return an additional flag whether the config it sends is ready or not, up to date downstream relays will not use them but instead keep trying to get a ready global config and not process envelopes until they do. Older relays won't deserialize the status flag and their behaviour will be as before.


relay-server/src/actors/global_config.rs
```diff
@@ -256,17 +266,28 @@ impl GlobalConfigService {
         match result {
             Ok(Ok(config)) => {
                 let mut success = false;
+                // No global status implies an outdated upstream, in that case, we fall back on
+                // the old behaviour of using the global config that we receive even if it might
+                // not be the valid one.
+                let is_ready = config
+                    .global_status
+                    .map(|stat| stat.is_ready())
+                    .unwrap_or(true);
```
TBS1996: unwrap_or(true) because we wanna keep the old behaviour of relaying the default global config if the upstream is outdated (not having global_status implies its outdated)

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
 - `normalize_performance_score` now handles `PerformanceScoreProfile` configs with zero weight components and component weight sums of any number greater than 0. ([#2756](https://github.com/getsentry/relay/pull/2756))
+- Return global config ready status to downstream relays  ([#2765](https://github.com/getsentry/relay/pull/2765))
```
```suggestion
- Return global config ready status to downstream relays. ([#2765](https://github.com/getsentry/relay/pull/2765))
```

relay-server/src/actors/global_config.rs
```diff
@@ -35,11 +35,29 @@ type UpstreamQueryResult =
     Result<Result<GetGlobalConfigResponse, UpstreamRequestError>, relay_system::SendError>;
 
 /// The response of a fetch of a global config from upstream.
+///
+/// Instead of using global_config::Status, we have the separate StatusResponse in order to not
```
Dav1dde: I would put this as a code comment on `global_status`

relay-server/src/actors/global_config.rs
```diff
@@ -35,11 +35,29 @@ type UpstreamQueryResult =
...
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 struct GetGlobalConfigResponse {
     #[serde(default)]
     global: Option<GlobalConfig>,
+    global_status: Option<StatusResponse>,
+}
+
+/// A mirror of [`Status`] without the associated data for use in serialization.
+#[derive(Debug, Deserialize, Serialize)]
```
Dav1dde: Can be `Copy, Clone` then `is_ready` can take self.

I am fine with how it is, but we have a bunch of enums that are `Copy, Clone`

relay-server/src/actors/global_config.rs
```diff
@@ -256,17 +266,28 @@ impl GlobalConfigService {
...
                 match config.global {
-                    Some(global_config) => {
+                    Some(global_config) if is_ready => {
                         // Log the first time we receive a global config from upstream.
                         if !self.global_config_watch.borrow().is_ready() {
                             relay_log::info!("received global config from upstream");
                         }
                         self.global_config_watch
-                            .send_replace(Status::Ready(Arc::new(global_config)));
+                            .send_replace(Status::Ready(global_config.into()));
```
Dav1dde: Can this not stay `Arc::new`

relay-server/src/endpoints/project_configs.rs
```diff
@@ -129,15 +131,19 @@ async fn inner(
...
-        true => Some(
-            state
-                .global_config()
-                .send(global_config::Get)
-                .await?
-                .get_ready_or_default(),
-        ),
-        false => None,
+    let (global, global_status) = match inner.global {
+        true => match state.global_config().send(global_config::Get).await? {
```
Dav1dde: Maybe make this an `if`, I think we talked about this and we prefer `if` over match on boolean

relay-server/src/actors/global_config.rs
```diff
@@ -256,17 +266,28 @@ impl GlobalConfigService {
         match result {
             Ok(Ok(config)) => {
                 let mut success = false;
+                // No global status implies an outdated upstream, in that case, we fall back on
+                // the old behaviour of using the global config that we receive even if it might
+                // not be the valid one.
+                let is_ready = config
+                    .global_status
+                    .map(|stat| stat.is_ready())
```
Dav1dde: This could be a `map_or(true, |stat| stat.is_ready()))`

Dav1dde (approved):

relay-server/src/actors/global_config.rs
TBS1996: i dont really see the point of it taking self 🤔

relay-server/src/actors/global_config.rs
TBS1996: yeah good idea!

relay-server/src/actors/global_config.rs
TBS1996: nice tip! dealing with options is really an art in rust lol

relay-server/src/actors/global_config.rs
TBS1996: oh yeah, didnt have the old diff in mind when i was changing the code many times

relay-server/src/endpoints/project_configs.rs
```diff
@@ -127,19 +128,22 @@ async fn inner(
...
+                Some(GlobalConfig::default().into()),
+                Some(StatusResponse::Pending),
+            ),
+        }
+    } else {
+        (None, None)
     };
 
+    let mut pending = Vec::with_capacity(keys_len);
+    let mut configs = HashMap::with_capacity(keys_len);
```
TBS1996, line 147: i think its better when variables are declared closer to where theyre being used

relay-server/src/actors/global_config.rs
Dav1dde: So generally types should implement Copy if they can: https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy

Once your type is copy, it doesnt really make sense to take a reference, especially for small types where a pointer is bigger than the type itself (compiler does what it wants though and may optimize to still use a reference if it makes sense etc.).

But here it also allows you to match without referencing `&Self::Ready`, which means it actually looks a bit nicer too imo

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## 23.11.2
 
+**Features**:
```
Dav1dde, line 5: This is in 23.11.2 now

relay-server/src/actors/global_config.rs
TBS1996: ok makes sense, it is done now

CHANGELOG.md
TBS1996: damn

CHANGELOG.md
```diff
@@ -1,7 +1,13 @@
 # Changelog
 
+**Unreleased**
+
+- Return global config ready status to downstream relays.  ([#2765](https://github.com/getsentry/relay/pull/2765))
```
iker-barriocanal:
```suggestion
- Return global config ready status to downstream relays. ([#2765](https://github.com/getsentry/relay/pull/2765))
```

relay-server/src/actors/global_config.rs
```diff
@@ -40,6 +40,23 @@ type UpstreamQueryResult =
 struct GetGlobalConfigResponse {
     #[serde(default)]
     global: Option<GlobalConfig>,
+    /// Instead of using [`Status`], we use StatusResponse as a separate field in order to not
+    /// make breaking changes to the api.
```
iker-barriocanal: nit: I think the context this comment provides is better suited in a GitHub PR comment rather than code comment.

relay-server/src/actors/global_config.rs
```diff
@@ -256,8 +265,13 @@ impl GlobalConfigService {
         match result {
             Ok(Ok(config)) => {
                 let mut success = false;
+                // No global status implies an outdated upstream, in that case, we fall back on
```
iker-barriocanal: Could we make this comment a bit clearer? "outdated" here refers to any relay that doesn't have a ready global config and not an "old" relay, and this can happen in new relays too (e.g. spin up a new relay that can't connect to upstream because the network is down).

CHANGELOG.md
```diff
@@ -1,7 +1,13 @@
...
+
 ## 23.11.2
 
+**Features**:
```
iker-barriocanal: Could we revert the changes in released changelogs?

iker-barriocanal (approved):
LGTM! Please see the comments before merging.

relay-server/src/actors/global_config.rs
Dav1dde: A missing `global_status` field in the response does indicate the relay (or sentry) we are talking to is on an older version (pre `global_status` field). An updated version would send a default config but also the status as `pending`. But I agree maybe we can phrase this better.

iker-barriocanal: Oh right, I confused the status in the global config with the state of relay with a global config. Thanks for pointing it out!

CHANGELOG.md
TBS1996: i guess not, unfixed it now!

relay-server/src/actors/global_config.rs
TBS1996: I changed it to normal comment rather than doc comment, cause I think it's not directly intuitive why we do it like this, so its nice for reader to know it's because of API

relay-server/src/actors/global_config.rs
```diff
@@ -40,6 +40,23 @@ type UpstreamQueryResult =
 struct GetGlobalConfigResponse {
     #[serde(default)]
     global: Option<GlobalConfig>,
+    // Instead of using [`Status`], we use StatusResponse as a separate field in order to not
+    // make breaking changes to the api.
+    global_status: Option<StatusResponse>,
```
jjbayer, line 46: Doesn't this also need a `#[serde(default)]` attribute (like `global` above)?

relay-server/src/actors/global_config.rs
```diff
@@ -256,8 +265,12 @@ impl GlobalConfigService {
         match result {
             Ok(Ok(config)) => {
                 let mut success = false;
+                // Older relays won't send a global status, in that case, we will pretend like the
+                // default global config is an up to date one, because that was the old behaviour.
+                let is_ready = config.global_status.map_or(true, |stat| stat.is_ready());
```
jjbayer: nit: Should we rename `config` to something like `response` here? Just to make it extra clear where it comes from.

jjbayer (commented):
Nice, looking forward to this!

relay-server/src/actors/global_config.rs
TBS1996: oh yes, that makes sense!

relay-server/src/actors/global_config.rs
TBS1996: huh id think that was necessary yeah, but it's weird integration tests didnt fail because of that, since minisentry don't return a global_status

TBS1996: anyway i added it

Dav1dde: We could remove the top `#[serde(default)]`, serde implicitly has the default behaviour for Option's, they are special in that sense.

PR #2768, opened 2023-11-27
ref(processor): Move sessions related code into separate sub-module

olksdr:
Create separate module in the `processor` for sessions related code.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor/session.rs
+++ b/relay-server/src/actors/processor/session.rs
@@ -1,3 +1,5 @@
+//! Contains the sessions related processor code.
+
 use std::error::Error;
 use std::net;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/processor/session.rs
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    true
+}
+
+/// Validates all sessions and session aggregates in the envelope, if any.
+///
+/// Both are removed from the envelope if they contain invalid JSON or if their timestamps
+/// are out of range after clock drift correction.
+pub fn process(state: &mut ProcessEnvelopeState, config: Arc<Config>) {
```
jjbayer: Should we move this to the top of the file? We've been slowly changing our convention to pub-first. This doc only mentions it for `impl` blocks, but I guess the same applies to standalone functions: https://develop.sentry.dev/rust/#declaration-order

jjbayer (approved):

relay-server/src/actors/processor/session.rs
olksdr: yeah, sounds good! Will do

iker-barriocanal (approved):
Solid stuff!

PR #2770, opened 2023-11-27
feat(spans): Allow well-known path segments in resource URLs

jjbayer:
We currently erase all but the last path segment for resource URLs to keep cardinality down. This leads to very minimal span descriptions, especially when the URL is relative.

This PR allowlists a set of widely used path segments (`static`, `chunks`, `media`, etc.) to give the span description a little more context.

Example:

```shell
# before
/*/*-*.js

# after
/_next/static/chunks/*-*.js
```

Fixes https://github.com/getsentry/team-ingest/issues/249

#skip-changelog


relay-event-normalization/src/normalize/span/description/resource.rs
```diff
@@ -0,0 +1,57 @@
+//! Code for scrubbing resource span description.
+use std::collections::BTreeSet;
+
+use once_cell::sync::Lazy;
+
+// TODO: move functions from parent module here.
```
jjbayer, line 6: I will move resource-related functions in a follow-up PR, so this one retains a readable diff.

relay-event-normalization/src/normalize/span/description/resource.rs
```diff
@@ -0,0 +1,57 @@
...
+        "dist",
+        "dms",
+        "files",
+        "icons",
+        "image",
+        "images",
+        "img",
+        "immutable",
+        "js",
+        "libs",
```
AbhiPrasad, line 38:
```suggestion
        "lib",
        "libs",
```

olksdr (approved):

AbhiPrasad (approved):

iker-barriocanal (approved):

PR #2774, opened 2023-11-27
feat(normalization): Introduce event normalization entry point

iker-barriocanal:
This PR introduces an entry point for event normalization. It currently only runs what previously was "light normalization", but it should contain all normalization logic in the future. This PR doesn't contain any functional changes.

Previous PRs approached the goal by introducing a new processor and calling it directly. This not only makes a more difficult to use API (`process_value` with a processor, a processing state, and the processor itself), but also allows running normalization with non-top-level items; for example, it'd be possible to run normalization on a standalone span. Normalization must be controlled from top-level items and a processor doesn't provide such control a regular function does. Internally, this function uses other processors depending on what the goals are.

Summary of the changes in this PR:
- Removes `NormalizeProcessor` in favor of introducing `normalize_event`.
- Renames the config to `NormalizationConfig`, as there's no such processor anymore. No properties have been updated.
- `normalize_event` calls `event::normalize`. The former checks the renormalization flag to decide whether to call the latter, and the latter runs the old "light normalization" (without the redundant flag check).
- Moving normalization code around changed how some modules are used, and these have been moved accordingly: `logentry`, `mechanism`, and `stacktrace`. These don't contain any changes.

Reviewing the commits in order may ease the review.

#skip-changelog


relay-event-normalization/src/event.rs
```diff
@@ -1,23 +1,13 @@
-//! Event normalization processor.
-//!
-//! This processor is work in progress. The intention is to have a single
-//! processor to deal with all event normalization. Currently, the normalization
-//! logic is split across several processing steps running at different times
-//! and under different conditions, like light normalization and store
-//! processing. Having a single processor will make things simpler.
-
```
olksdr, line 8: It's still would be great to keep some docs  describing what this module is responsible for.

relay-event-normalization/src/event.rs
```diff
@@ -29,288 +19,133 @@ use relay_event_schema::protocol::{
...
+        config.max_secs_in_past,
+        config.max_secs_in_future,
+    )?; // Timestamps are core in the metrics extraction
+    normalize_event_tags(event)?; // Tags are added to every metric
+
+    // TODO: Consider moving to store normalization
+    if config.device_class_synthesis_config {
+        normalize_device_class(event);
+    }
+    light_normalize_stacktraces(event)?;
```
olksdr: Do we still want to call this `light_normalize_...` ?

relay-event-normalization/src/event.rs
```diff
@@ -29,288 +19,133 @@ use relay_event_schema::protocol::{
...
+        config.measurements.clone(),
+        config.max_name_and_unit_len,
+    ); // Measurements are part of the metric extraction
+    normalize_performance_score(event, config.performance_score);
+    normalize_breakdowns(event, config.breakdowns_config); // Breakdowns are part of the metric extraction too
+
+    // Some contexts need to be normalized before metrics extraction takes place.
+    processor::apply(&mut event.contexts, normalize_contexts)?;
+
+    if config.light_normalize_spans && event.ty.value() == Some(&EventType::Transaction) {
```
olksdr: Same here, does it make sense to rename this config option and remove `light_normalize_...` from it?

relay-event-normalization/src/lib.rs
```diff
@@ -207,3 +214,153 @@ impl<'a> Processor for StoreProcessor<'a> {
...
+/// The returned [`ProcessingResult`] indicates whether the passed event should
+/// be ingested or dropped.
+pub fn normalize_event(
+    event: &mut Annotated<Event>,
+    config: &NormalizationConfig,
+) -> ProcessingResult {
+    let Annotated(ref mut event, ref mut meta) = event;
+    let Some(ref mut event) = event else {
+        return Ok(());
+    };
```
olksdr: Will this also work here?

```suggestion
    let Annotated(Some(ref mut event), ref mut meta) = event else {
        return Ok(());
    };
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -28,13 +28,7 @@ pub mod user_agent;
 pub mod utils;
 
 mod contexts;
-mod logentry;
-mod mechanism;
-mod processor;
 mod request;
-mod stacktrace;
```
olksdr, line 35: wondering why this mode where moved out of `normalize` module to top level?

olksdr (commented):
I remember we discussed making normalization step infallible and introduce verification method which actually will be responsible for checking the event. 
Will this change come in the followup PRs?

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: The `event` module uses `stacktrace::process_non_raw_frame` for stacktrace normalization. This results in 3 options on `stacktrace`:

1. Make it public in `normalize`. `normalize` is fully exported, so that'd result in `stacktrace` being unnecessarily exposed outside of the crate.
2. Make it `pub(crate)`. Workaround to option 1, but probably unnecessary as it's only used in `event`.
3. Make it a top-level module. I _think_ `normalize` will become redundant and we'll be able to merge it into the top-level module, so this seemed reasonable.

What do you think? Am I missing something?

iker-barriocanal:
> making normalization step infallible and introduce verification method which actually will be responsible for checking the event

That's coming (already WIP) in a next PR! Creating the entry point first makes separating validation easier, as invoking a processor directly obscures it.

There are still references to light normalization in the codebase. I'll address all of them at a later point.

relay-event-normalization/src/normalize/mod.rs
olksdr: I would basically keep everything in `normalize` module together with new `event` module. That would produce smaller diff and for now it would make sense to keep all these modules in there. 

The `lib.rs` already re-exports almost everything and it would be easy to use it.

relay-event-normalization/src/lib.rs
```diff
@@ -207,3 +214,152 @@ impl<'a> Processor for StoreProcessor<'a> {
         Ok(())
     }
 }
+
+/// Configuration for [`normalize_event`].
+#[derive(Clone, Debug)]
+pub struct NormalizationConfig<'a> {
```
olksdr: I would move this in to `event` module, since that where it mostly used and can be re-exported if needed for outside of the crate.

relay-server/src/actors/processor.rs
```diff
@@ -1465,7 +1465,7 @@ impl EnvelopeProcessorService {
         let request_meta = state.managed_envelope.envelope().meta();
         let client_ipaddr = request_meta.client_addr().map(IpAddr::from);
 
-        let light_normalize_spans = state
+        let normalize_spans = state
```
olksdr, line 902: It would be great to rename `light_normalize_event` function to just `normalize_event`, but can be done in followup

olksdr (approved):
Beginning  looks great! Looking forward to the followups! :shipit:

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: Do you feel strongly opinionated about it? Considering that the crate's purpose is to validate and normalize/process, what's the benefit of having a `normalize` submodule and what would go in or out of it?

relay-event-normalization/src/lib.rs
iker-barriocanal: What would you do with `normalize_event`? IMO it makes sense to keep the entry point and its config together. Note that `NormalizationConfig` will be expanded in the future with the config of `StoreNormalizeProcessor` and it won't only be used in `event.rs` anymore.

relay-event-normalization/src/normalize/mod.rs
olksdr: I'm fine with moving forward with this, just so we can deliver the changes faster. 
But I prefer better code organization in the future PRs.

relay-event-normalization/src/lib.rs
olksdr: Same comment, the win is the better code organization. 
I wouldn't keep the code in the `lib.rs` itself but used it for re-exports from the modules which has to be public etc. 
But it's fine to move with the current setup and revisit this later.

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: Yeah makes sense and I agree, should be addressed in [04f1415](https://github.com/getsentry/relay/pull/2774/commits/04f14156b6ccf084976d15709b4a81084f8d543c).

relay-event-normalization/src/event.rs
```diff
@@ -166,151 +162,141 @@ impl<'a> Default for NormalizeProcessorConfig<'a> {
...
+    event: &mut Annotated<Event>,
+    config: &NormalizationConfig,
+) -> ProcessingResult {
+    let Annotated(Some(ref mut event), ref mut meta) = event else {
+        return Ok(());
+    };
 
-        // Process security reports first to ensure all props.
-        normalize_security_report(event, self.config.client_ip, &self.config.user_agent);
+    let is_renormalize = config.is_renormalize;
```
TBS1996, line 178: seems a bit redundant imo

relay-event-normalization/src/event.rs
```diff
@@ -438,7 +424,7 @@ fn normalize_user_geoinfo(geoip_lookup: &GeoIpLookup, user: &mut User) {
 }
 
 fn normalize_logentry(logentry: &mut Annotated<LogEntry>, _meta: &mut Meta) -> ProcessingResult {
-    processor::apply(logentry, crate::normalize::logentry::normalize_logentry)
+    processor::apply(logentry, crate::logentry::normalize_logentry)
```
TBS1996, line 427: would the precedent be that we would use the namespace->verb instead of verb->noun where applicable ? like here doing logentry::normalize instead of logentry::normalize_logentry?

relay-event-normalization/src/event.rs
iker-barriocanal: I find this a bit easier to read and reuse, but I'll revisit this decision in the future if it's no longer applicable.

relay-event-normalization/src/event.rs
iker-barriocanal: I think that'd make sense if we were refactoring it, but I just moved the module one level up so changing the name feels out of scope.

PR #2777, opened 2023-11-28
feat(spans): Normalization for span ingestion

jjbayer:
Run full normalization before extracting metrics from a standalone span.

#skip-changelog


tests/integration/test_store.py
```diff
@@ -1340,8 +1343,8 @@ def test_span_ingestion(
                         "traceId": "89143b0763095bd9c9955e8175d1fb23",
                         "spanId": "e342abb1214ca181",
                         "name": "my 1st OTel span",
-                        "startTimeUnixNano": 1697620454980000000,
-                        "endTimeUnixNano": 1697620454980078800,
+                        "startTimeUnixNano": int(start.timestamp() * 1_000_000_000),
+                        "endTimeUnixNano": int(end.timestamp() * 1_000_000_000),
```
jjbayer, line 1348: Our clock drift processor fixes very old timestamps, so we have to use dynamic timestamps here.

relay-server/src/actors/processor.rs
```diff
@@ -2110,6 +2112,7 @@ impl EnvelopeProcessorService {
 
     #[cfg(feature = "processing")]
     fn extract_spans(&self, state: &mut ProcessEnvelopeState) {
+        use relay_event_normalization::span::tag_extraction;
```
olksdr: This should be moved to above into the 

```rust
use {
```

relay-server/src/actors/processor.rs
```diff
@@ -2279,59 +2282,95 @@ impl EnvelopeProcessorService {
 
     #[cfg(feature = "processing")]
     fn process_spans(&self, state: &mut ProcessEnvelopeState) {
+        use crate::actors::processor::span::{normalize_span, NormalizeSpanConfig};
```
olksdr: Same here for this 2 imports

relay-server/src/actors/processor.rs
```diff
@@ -2279,59 +2284,92 @@ impl EnvelopeProcessorService {
...
+                    .aggregator_config_for(MetricNamespace::Spans)
+                    .max_tag_value_length,
+            };
+            let mut annotated_span = match normalize_span(annotated_span, config) {
+                Ok(s) => s,
                 Err(e) => {
-                    relay_log::error!("invalid span: {e}");
-                    return;
+                    relay_log::debug!("failed to normalize span: {}", e);
+                    return ItemAction::DropSilently;
```
iker-barriocanal: Why the error -> debug downgrade, considering that spans are silently dropped?

relay-server/src/actors/processor.rs
```diff
@@ -2279,59 +2284,92 @@ impl EnvelopeProcessorService {
...
-                let mut item = Item::new(ItemType::Span);
-                item.set_payload(ContentType::Json, payload);
-                items.push(item);
+                let metrics = extract_metrics(span, config);
+                state.extracted_metrics.project_metrics.extend(metrics);
             }
-        };
-        let standalone_span_ingestion_enabled = state
-            .project_state
-            .has_feature(Feature::StandaloneSpanIngestion);
```
iker-barriocanal: Is the feature flag no longer useful? Without this check, we'll proceed with `ItemType::Span`s too.

relay-server/src/actors/processor.rs
jjbayer: Good question! `normalize_span` returns an error when the span is invalid (e.g. missing trace ID), so we don't want to be notified about that. I changed the wording to make the failure case more clear.

relay-server/src/actors/processor.rs
phacops: It's already checked in `span::filter` so we didn't think we needed to have it twice.

relay-server/src/actors/processor/span.rs
```diff
@@ -1,18 +1,30 @@
...
-use crate::envelope::ItemType;
-use crate::utils::ItemAction;
-use relay_dynamic_config::Feature;
+use crate::{actors::processor::ProcessEnvelopeState, envelope::ItemType, utils::ItemAction};
+#[cfg(feature = "processing")]
+use {
+    crate::actors::processor::ProcessingError,
+    crate::envelope::{ContentType, Item},
+    crate::metrics_extraction::generic::extract_metrics,
+};
```
iker-barriocanal, line 27: Should we merge the `processing` import blocks into one?

iker-barriocanal (approved):

relay-server/src/actors/processor/span.rs
jjbayer: We don't have a best practice for this yet. My reasoning was to keep the partition into std lib - other crates - current crate.

PR #2783, opened 2023-11-29
feat(webvitals): adds optional performance score components

edwardgou-sentry:
Adds support for an `optional` PerformanceScoreWeightedComponent. Measurements that are marked optional will not block performance score processing if they are missing, and will have their weight set to 0.


relay-event-normalization/src/normalize/mod.rs
```diff
@@ -346,6 +346,9 @@ pub struct PerformanceScoreWeightedComponent {
     pub p10: f64,
     /// Median used to define the log-normal for calculation
     pub p50: f64,
+    // Whether the measurement is optional. If the measurement is missing, performance score processing
+    // may still continue, and the weight will be 0.
+    pub optional: Option<bool>,
```
Dav1dde: If there is no difference between `None` and `false`:

```rs
#[serde(default)]
pub optional: bool,
```

relay-event-normalization/src/normalize/processor.rs
```diff
@@ -736,20 +736,30 @@ fn normalize_performance_score(
...
+                    // All non-optional measurements with a profile weight greater than 0 are
+                    // required to exist on the event. Skip calculating performance scores if
+                    // a measurement with weight is missing.
                     break;
                 }
                 let mut score_total = 0.0;
                 let mut weight_total = 0.0;
                 for component in &profile.score_components {
+                    // Skip optional components if they are not present on the event.
+                    if component.optional.is_some()
```
Dav1dde: Can be `unwrap_or(false)` as above

relay-event-normalization/src/normalize/mod.rs
edwardgou-sentry: Thanks! There's no difference so i've updated to using this

relay-event-normalization/src/normalize/processor.rs
edwardgou-sentry: Updated to just `component.optional` since I'm using a default value for optional now

iker-barriocanal (approved):

Dav1dde (approved):

PR #2784, opened 2023-11-29
ref(metric-meta): Add metric for total incoming metric meta

Dav1dde:
Tracks the total number of code location items in a statsd metric so we can estimate the load generated in SDKs.

Epic: https://github.com/getsentry/sentry/issues/60260
#skip-changelog

### DIFF
--- a/relay-metrics/src/meta/aggregator.rs
+++ b/relay-metrics/src/meta/aggregator.rs
@@ -48,6 +48,7 @@ impl MetaAggregator {
     pub fn add(&mut self, project_key: ProjectKey, meta: MetricMeta) -> Option<MetricMeta> {
         let mut send_upstream = HashMap::new();
 
+        let mut total = 0;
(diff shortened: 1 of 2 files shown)


relay-metrics/src/meta/aggregator.rs
```diff
@@ -58,8 +59,12 @@ impl MetaAggregator {
             if let Some(items) = self.add_scoped(&scope, items) {
                 send_upstream.insert(scope.mri, items);
             }
+
+            total += 1;
         }
 
+        relay_statsd::metric!(counter(MetricCounters::MetaAggregatorTotalIncoming) += total);
```
jan-auer: Right now, this is tracking `mapping.len()`. Should we instead track the number of items? 

_nit_: In its current form, this can be simplified to track `meta.mapping.len()`.

jan-auer (approved):

PR #2792, opened 2023-11-29
feat(spans): Add a measurements field to span protocol

phacops:
As we move towards standalone span ingestion, we need to support passing measurements to spans directly. This change will allow us to receive measurements when ingesting a span.

The side-effect is spans in transactions will also be able to ingest measurements but they won't be used in the product.

### DIFF
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -77,9 +77,6 @@ pub struct Span {
     pub received: Annotated<Timestamp>,
 
     /// Measurements which holds observed values such as web vitals.
-    ///
-    /// Measurements are only available on transactions. They contain measurement values of observed
-    /// values such as Largest Contentful Paint (LCP).
     #[metastructure(skip_serialization = "empty")]
     #[metastructure(omit_from_schema)] // we only document error events for now
     pub measurements: Annotated<Measurements>,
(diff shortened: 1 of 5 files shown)


relay-event-schema/src/protocol/span.rs
```diff
@@ -76,6 +76,14 @@ pub struct Span {
     /// Timestamp when the span has been received by Sentry.
     pub received: Annotated<Timestamp>,
 
+    /// Measurements which holds observed values such as web vitals.
+    ///
+    /// Measurements are only available on transactions. They contain measurement values of observed
+    /// values such as Largest Contentful Paint (LCP).
+    #[metastructure(skip_serialization = "empty")]
+    #[metastructure(omit_from_schema)] // we only document error events for now
```
jjbayer, line 81:
```suggestion
```

Not sure if this is necessary, we don't set it on the other fields in this struct.

relay-event-schema/src/protocol/span.rs
```diff
@@ -76,6 +76,14 @@ pub struct Span {
     /// Timestamp when the span has been received by Sentry.
     pub received: Annotated<Timestamp>,
 
+    /// Measurements which holds observed values such as web vitals.
+    ///
+    /// Measurements are only available on transactions. They contain measurement values of observed
+    /// values such as Largest Contentful Paint (LCP).
```
jjbayer:
```suggestion
```

jjbayer (approved):
LGTM, wanna extend the serialization test at the bottom of this file with some `measurements`?

As discussed, let's get buy in from SDK teams before merging this.

relay-event-schema/src/protocol/span.rs
phacops: This is necessary since the `JSONSchema` trait is not added on `Measurements`.

phacops:
TSC said OK for the Relay part but want to see an RFC in case it's made into a public API. It's not the plan yet so we can go ahead and merge this.

PR #2794, opened 2023-11-29
fix(integration): Fix integration test

TBS1996:
the real sentry wouldn't return a "global" key with an empty val, but no key at all, so this makes it more aligned to that. 

#skip-changelog


iker-barriocanal (approved):
Thanks for getting to this!
nit: we have a `test` type for commits that deal with tests: https://develop.sentry.dev/commit-messages/#type

PR #2798, opened 2023-11-30
fix(user-report): Accept user reports with trailing whitespaces

Dav1dde:
We have a customer with a forked version of the native SDK which is sending invalid user reports with a trailing `\n`. Trim the payload before parsing it to allow user reports to go through.


olksdr (approved):

Dav1dde:
Okay so this is weird, the native SDK does not even implement the functionality to send user feedback reports and we are seeing this from a version that is >1 year outdated. Maybe the customer made some modifications to their SDK. I want to investigate this first before we go ahead with this.

PR #2799, opened 2023-11-30
ref(processor): Move unreal and dynamic sampling into separate sub-modules

olksdr:
This PR contains few following changes:

* moves `unreal` processor related code into separate sub-module
* moves `dynamic sampling` processor related code (and also related tests) into separate sub-module
* instead of cloning and passing the `Arc<Config>` just use `&Config` for all sub-modules

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor/unreal.rs
+++ b/relay-server/src/actors/processor/unreal.rs
@@ -18,7 +18,7 @@ use crate::utils;
 /// this includes cases where a single attachment file exceeds the maximum file size. This is in
 /// line with the behavior of the envelope endpoint.
 ///
-/// After this, [`EnvelopeProcessorService`] should be able to process the envelope the same
+/// After this, [`crate::actors::processor::EnvelopeProcessorService`] should be able to process the envelope the same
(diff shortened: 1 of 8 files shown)


Dav1dde (approved):
Assuming the code was just moved 👍

relay-server/src/actors/processor/unreal.rs
```diff
@@ -0,0 +1,40 @@
+//! Unreal 4 processort related code.
```
iker-barriocanal, line 1:
```suggestion
//! Unreal 4 processor related code.
```

relay-server/src/actors/processor/unreal.rs
```diff
@@ -0,0 +1,40 @@
...
+//!
+//! These functions are included only in the processing mode.
```
iker-barriocanal, line 3: Q: does it make sense to add the `processing` mode flag checks here, over this comment? It'd be harder to get outdated.

iker-barriocanal (approved):

PR #2800, opened 2023-11-30
feat(metrics): Add size limits for metric envelope items

Dav1dde:
Introduce the missing size limits for the metric envelope items.

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -563,6 +563,12 @@ struct Limits {
     max_profile_size: ByteSize,
     /// The maximum payload size for a span.
     max_span_size: ByteSize,
+    /// The maximum payload size for a stasd metric.
+    max_statsd_size: ByteSize,
+    /// The maximum payload size for metric buckets.
+    max_metric_buckets_size: ByteSize,
+    /// The maximum payload size for metric metadata.
+    max_metric_meta_size: ByteSize,
     /// The maximum payload size for a compressed replay.
     max_replay_compressed_size: ByteSize,
     /// The maximum payload size for an uncompressed replay.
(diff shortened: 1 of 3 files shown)


relay-config/src/config.rs
```diff
@@ -604,6 +610,9 @@ impl Default for Limits {
             max_api_chunk_upload_size: ByteSize::mebibytes(100),
             max_profile_size: ByteSize::mebibytes(50),
             max_span_size: ByteSize::mebibytes(1),
+            max_statsd_size: ByteSize::kibibytes(100),
```
Dav1dde: Not sure if this is a good limit?

relay-config/src/config.rs
```diff
@@ -604,6 +610,9 @@ impl Default for Limits {
...
+            max_metric_buckets_size: ByteSize::mebibytes(1),
```
Dav1dde, line 614: This was discussed with Jan, apparantely the SDKs already have that limit

relay-config/src/config.rs
```diff
@@ -563,6 +563,12 @@ struct Limits {
     max_profile_size: ByteSize,
     /// The maximum payload size for a span.
     max_span_size: ByteSize,
+    /// The maximum payload size for a stasd metric.
```
TBS1996:
```suggestion
    /// The maximum payload size for a statsd metric.
```

TBS1996 (commented):
lgtm, we should probably figure out what a good statsd limit is before going forward though

relay-config/src/config.rs
jan-auer: Let's set 1MB for all of these, DDM sends items larger than 100kB.

jan-auer (approved):

olksdr (approved):

PR #2801, opened 2023-11-30
feat(size-limits): Include offender in error message

### LINKED ISSUE #2771: Better error message instead of `envelope size exceeded size limits`
It would be great to also include the limit of what kind of envelope item was exceeded. Something like `envelope size exceeded size limits for type "check-in"` or something along those lines. (I would love to keep the existing error message intact because we have some queries in GCP that search for "envelope size exceeded size limits".)

Here: 
https://github.com/getsentry/relay/blob/master/relay-server/src/endpoints/common.rs#L80C1-L80C1

Dav1dde:
Includes the first size limit offending item in the error message.

Depends on: #2800
Fixes: #2771


olksdr (approved):

jan-auer (commented):
~Should we also include that information in the error serialized to the web api?~ That is the case through the `Display` implementation.

jan-auer (approved):

PR #2804, opened 2023-11-30
ref(config): Cleanup metric config

Dav1dde:
The metric metadata config is a brand new config for a very experimental feature (we can assume no one uses it), unfortunately I wasn't careful and added it to the `metrics:` key, which is already being used for statsd metrics emitted by relay, not related to metric processing.

#skip-changelog

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -517,7 +517,7 @@ impl Default for Metrics {
     }
 }
 
-/// Controls processing of metrics and metric metadata.
+/// Controls processing of Sentry metrics and metric metadata.
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(default)]
 struct SentryMetrics {
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -502,14 +502,6 @@ struct Metrics {
     /// For example, a value of `0.3` means that only 30% of the emitted metrics will be sent.
     /// Defaults to `1.0` (100%).
     sample_rate: f32,
-    /// Code locations expiry in seconds.
-    ///
-    /// Defaults to 15 days.
-    meta_locations_expiry: u64,
-    /// Maximum amount of code locations to store per metric.
-    ///
-    /// Defaults to 5.
-    meta_locations_max: usize,
 }
 
 impl Default for Metrics {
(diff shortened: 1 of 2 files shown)


relay-config/src/config.rs
```diff
@@ -502,14 +502,6 @@ struct Metrics {
...
     /// Defaults to `1.0` (100%).
     sample_rate: f32,
-    /// Code locations expiry in seconds.
-    ///
-    /// Defaults to 15 days.
-    meta_locations_expiry: u64,
-    /// Maximum amount of code locations to store per metric.
-    ///
-    /// Defaults to 5.
-    meta_locations_max: usize,
```
iker-barriocanal, line 512: Have we considered self-hosted users (or us) that may be using this config here?

I think doing this is fine as long as we document it properly in the changelog.

relay-config/src/config.rs
```diff
@@ -521,6 +513,46 @@ impl Default for Metrics {
             hostname_tag: None,
             buffering: true,
             sample_rate: 1.0,
+        }
+    }
+}
+
+/// Controls processing of metrics and metric metadata.
```
```suggestion
/// Controls processing of Sentry metrics and metric metadata.
```
(not sure about this suggestion)

iker-barriocanal (approved):

relay-config/src/config.rs
Dav1dde: Nobody is using this config here, we just added it and it is still a very experimental feature, but this is a valid concern for the partitions and batch size settings, we will also have to update these in our prod configs before/after merging this.

jjbayer (changes requested):
Could you add a motivation to the PR description?

This is a breaking change (`aggregator.max_flush_bytes` is no longer respected), so I would feel more comfortable if we kept the old fields and added a deprecation warning for them.

Dav1dde:
@jjbayer added the description, hopefully touching on all the points from the original slack and offline conversations

relay-config/src/config.rs
```diff
@@ -521,6 +513,46 @@ impl Default for Metrics {
...
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(default)]
+struct SentryMetrics {
```
olksdr, line 523: `SentryMetrics` seems ok.  But what about `DeveloperMetrics` or `SentryDeveloperMetrics`?

olksdr (commented):
Overall looks ok.

relay-config/src/config.rs
Dav1dde: I feel like `Developer` implies DDM/custom metrics but this is supposed to be for config for all metrics.

Dav1dde:
Updated the PR and removed the breaking changes.

iker-barriocanal (approved):

jjbayer (approved):
LGTM, please update the PR description!

PR #2808, opened 2023-12-01
ref(processor): Add PartialDsn ctor for outbound requests

Dav1dde:
Follow-Up from https://github.com/getsentry/relay/pull/2751#discussion_r1404280608

#skip-changelog

### DIFF
--- a/relay-server/src/extractors/request_meta.rs
+++ b/relay-server/src/extractors/request_meta.rs
@@ -113,7 +113,7 @@ impl PartialDsn {
         })
     }
 
-    /// Creates a new [`PartialDsn`] for a relay outbound request.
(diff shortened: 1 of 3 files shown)


relay-server/src/extractors/request_meta.rs
```diff
@@ -112,6 +113,18 @@ impl PartialDsn {
         })
     }
 
+    /// Creates a new [`PartialDsn`] for a relay outbound request.
```
olksdr:
```suggestion
    /// Creates a new [`PartialDsn`] for a Relay outbound request.
```

olksdr (approved):

PR #2813, opened 2023-12-04
feat(global-config): Add support for options in global config

Dav1dde:
Sentry PR: https://github.com/getsentry/sentry/pull/61024

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -1,9 +1,10 @@
-use std::fs::File;
 use std::io::BufReader;
 use std::path::Path;
+use std::{collections::HashMap, fs::File};
 
 use relay_event_normalization::MeasurementsConfig;
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
 
 /// A dynamic configuration for all Relays passed down from Sentry.
 ///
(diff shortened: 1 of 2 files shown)


relay-server/src/actors/global_config.rs
```diff
@@ -38,11 +38,9 @@ type UpstreamQueryResult =
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 struct GetGlobalConfigResponse {
-    #[serde(default)]
```
Dav1dde, line 41: This is redundant and just cleanup

relay-dynamic-config/src/global.rs
```diff
@@ -14,7 +15,11 @@ use serde::{Deserialize, Serialize};
 #[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
 pub struct GlobalConfig {
     /// Configuration for measurements normalization.
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub measurements: Option<MeasurementsConfig>,
+    /// Sentry options passed down to relay.
```
olksdr:
```suggestion
    /// Sentry options passed down to Relay.
```

relay-dynamic-config/src/global.rs
```diff
@@ -1,9 +1,10 @@
-use std::fs::File;
 use std::io::BufReader;
 use std::path::Path;
+use std::{collections::HashMap, fs::File};
```
olksdr: Let's split it into the separate imports.

olksdr (approved):

relay-dynamic-config/src/global.rs
```diff
@@ -34,13 +40,27 @@ impl GlobalConfig {
...
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct Options {
+    // Example:
+    // ```rs
+    // #[serde(default, rename = "relay.some-option.name")]
+    // pub some_option: Vec<u32>,
+    // ```
+    /// All other unknown options.
+    #[serde(flatten)]
+    other: HashMap<String, Value>,
```
iker-barriocanal, line 55: nit: I assume this `other` is for backward compatibility -- does using `RawValue` avoid deserializing the options we don't need? Haven't checked how serde deals with this exactly.

PR #2816, opened 2023-12-04
feat(ai): add LLM token metric span metric extraction

JoshFerge:
As part of our LLM monitoring, we are capturing token counter metrics from Open AI completion spans. These three metrics are `prompt`, `completion`, and `total`. These numbers will be added to the span data fields. It is unclear if `total` is needed, as it is just a sum of `prompt` and `completion`, but we can always drop it if doing a sum query is trivial there.

We are tagging these counters with:

`group, description, op, language_model, environment, transaction`.

Note that any field with the word `token` is by default scrubbed, so we are going with `tken`. We can map this back to `Token` either in the UI, or if we build out specific product oriented API endpoints, we can do it there. 

This is intended for an Internal Release, so only the Sentry org will be sending these spans.

See spec document here: https://www.notion.so/sentry/LLM-v0-Tech-Spec-81a1aaae56b54c0e93606edcd882cfcd
Part of https://github.com/getsentry/getsentry/issues/12171

The associated SDK change is here for now https://github.com/getsentry/sentry/pull/61086

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -154,7 +154,6 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
             condition: Some(RuleCondition::eq("span.op", "ai.llm.completion")),
             tags: [
                 ("span.sentry_tags.", "group"),
-                ("span.sentry_tags.", "description"),
                 ("span.sentry_tags.", "op"),
                 ("span.data.", "language_model"),
                 ("span.sentry_tags.", "environment"),
@@ -174,7 +173,6 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
             condition: Some(RuleCondition::eq("span.op", "ai.llm.completion")),
             tags: [
                 ("span.sentry_tags.", "group"),
-                ("span.sentry_tags.", "description"),
                 ("span.sentry_tags.", "op"),
                 ("span.data.", "language_model"),
                 ("span.sentry_tags.", "environment"),
@@ -194,7 +192,6 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
             condition: Some(RuleCondition::eq("span.op", "ai.llm.completion")),
             tags: [
                 ("span.sentry_tags.", "group"),
-                ("span.sentry_tags.", "description"),
                 ("span.sentry_tags.", "op"),
                 ("span.data.", "language_model"),
                 ("span.sentry_tags.", "environment"),
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -20,6 +20,9 @@ const DISABLED_DATABASES: &[&str] = &[
 /// A list of span.op` patterns we want to enable for mobile.
 const MOBILE_OPS: &[&str] = &["app.*", "ui.load*"];
 
+/// A list of AI Module span ops we want to ingest metrics for.
+const LLM_OPS: &[&str] = &["ai.llm.*"];
+
 /// A list of patterns found in MongoDB queries
 const MONGODB_QUERIES: &[&str] = &["*\"$*", "{*", "*({*", "*[{*"];
 
@@ -63,6 +66,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
         (
             RuleCondition::eq("span.op", "http.client")
                 | RuleCondition::glob("span.op", MOBILE_OPS)
+                | RuleCondition::glob("span.op", LLM_OPS)
                 | (RuleCondition::glob("span.op", "db*") & !is_disabled & !is_mongo)
                 | resource_condition.clone(),
             resource_condition,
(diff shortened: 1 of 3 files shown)


relay-dynamic-config/src/defaults.rs
```diff
@@ -143,6 +147,66 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
                 .into(),
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "c:spans/data.llm.llm_completion_tkens@none".into(),
+            field: Some("span.data.llm_completion_tkens".into()),
+            condition: Some(RuleCondition::eq("span.op", "ai.llm.completion")),
+            tags: [
+                ("span.sentry_tags.", "group"),
+                ("span.sentry_tags.", "description"),
```
phacops: This looks like it contains the prompt and I would think it's high cardinality since it could contain user input. Is this correct?

If yes, we can't use the description or group as a tag.

relay-dynamic-config/src/defaults.rs
```diff
@@ -143,6 +147,66 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 })
                 .into(),
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "c:spans/data.llm.llm_completion_tkens@none".into(),
```
phacops, line 152: Is the missing `o` in `tokens` a typo or the correct metric name?

relay-dynamic-config/src/defaults.rs
```diff
@@ -63,6 +66,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
         (
             RuleCondition::eq("span.op", "http.client")
                 | RuleCondition::glob("span.op", MOBILE_OPS)
+                | RuleCondition::glob("span.op", LLM_OPS)
```
phacops, line 69: This would start ingesting `exclusive_time` in 2 different distributions. Is this what you want? Also, see the comment below about `description`. If it turns out to be high cardinality, we should not ingest those 2 metrics for the LLM ops.

relay-dynamic-config/src/defaults.rs
JoshFerge: mentioned in description -- `token` is one of our default field regexes that get scrubbed, so the name cannot have the word `token` in it. (Scrubbing happens after metrics extraction, but we want this value to be on the indexed data as well).

relay-dynamic-config/src/defaults.rs
JoshFerge: In general the way OpenAI recommends one to set up their prompts is to have the first prompt be a "system" prompt, which sets the behavior of the assistant. For example, in our own [use case,](https://github.com/getsentry/sentry/blob/c8071e2283a5b14bba9180dddc2fedc654177277/src/sentry/api/endpoints/event_ai_suggested_fix.py#L256) we only have 4 prompts, so the grouping should work in basic use cases.

Misuses of the API yes would resust in high cardinality descriptions, but this shouldn't be the norm.

phacops: As we also identified on Slack, you're going to hit truncation issues. I would not tag with the description, keep hashing the full description for the group though (so beware of https://github.com/getsentry/relay/pull/2749).

I would also suggest creating your own `exclusive_time` metrics as the tags will be very different from the current `exclusive_time` metrics. Or at least add a condition to not emit one of them since you likely don't need both.

JoshFerge: I've gone ahead and removed description from the metrics tags here. Yes, I think for LMM prompts we'd want the group to be based off of the un-truncated description.

relay-dynamic-config/src/defaults.rs
```diff
@@ -143,6 +147,63 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "c:spans/data.llm.llm_completion_tkens@none".into(),
+            field: Some("span.data.llm_completion_tkens".into()),
+            condition: Some(RuleCondition::eq("span.op", "ai.llm.completion")),
+            tags: [
+                ("span.sentry_tags.", "group"),
+                ("span.sentry_tags.", "op"),
+                ("span.data.", "language_model"),
```
jjbayer, line 158: Do you want to be able to query `language_model` from the spans dataset as well as from metrics? In that case, you'd have to add it to `sentry_tags` in [tag_extraction.rs](https://github.com/getsentry/relay/blob/19ab9c3d4990963a1dd065f41517426b5f07232d/relay-event-normalization/src/normalize/span/tag_extraction.rs#L1).

relay-dynamic-config/src/defaults.rs
```diff
@@ -143,6 +147,63 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
...
+                ("span.sentry_tags.", "environment"),
```
jjbayer, line 159: This would end up as tag `span.environment`, but we already have a precedent for naming that tag `environment`. Probably easiest to list `TagSpec` objects explicitly here instead of using `.map()`.

relay-dynamic-config/src/defaults.rs
jjbayer, line 158: Please also add an entry to [strings.py](https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py) for `language_model`, such that this string does not get duplicated in the indexer db for every org.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -83,6 +83,7 @@ pub(crate) fn scrub_span_description(span: &Span) -> Option<String> {
                 Some(description.to_owned())
             }
             ("file", _) => scrub_file(description),
+            ("ai", _) => Some(description.to_owned()),
```
jjbayer, line 86: Same question as @phacops: Can we expect description to be low cardinality?

relay-dynamic-config/src/defaults.rs
jjbayer: I think I have the same question: Is the `exclusive_time` metric required for this feature?

Dav1dde:
@JoshFerge Is this still something you need? Feel free to leave the PR open, but I'll drop it from our board for now.

PR #2818, opened 2023-12-05
feat(protocol): Implement Getter for request.headers

jan-auer:
This exposes `request.headers` via `Getter` so that it can be accessed in metric
extraction and dynamic sampling. The product needs to access the `Referer`
header.

Note that the referer is a common attribute which will be extracted into a
dedicated field in a separate PR.

#skip-changelog


olksdr (approved):

Dav1dde (approved):

iker-barriocanal (approved):
Should we add a test?

PR #2821, opened 2023-12-06
Revert "feat(server): Org rate limit per metric bucket"

jjbayer:
Reverts getsentry/relay#2758

This was deployed before sentry updated librelay, so sentry crashed with unknown data category.

#skip-changelog


iker-barriocanal (approved):

Dav1dde (approved):

jjbayer:
@TBS1996 before we ship this again, let's figure out why Relay produced outcomes with this data category at all if no quota config was configured for it yet.

After that I recommend the order:

1. Add a PR for just the data category.
2. Release librelay and bump it in sentry.
3. Re-introduce the business logic of this PR.

PR #2822, opened 2023-12-06
feat(spans): Extract `os.name` tag on mobile spans

shruthilayaj:
We need to tag "platform" mobile spans originated from, specifically for cross platform use cases. 
Cross platform apps (React Native and Flutter) can run on Apple and Android phones, so we want
to be able to differentiate which os the span originated from.
We don't want to just limit this to `sentry.javascript.react-native` or `sentry.dart.flutter` SDKs
because it is possible users can initialize a Native SDK in addition to the hybdrid SDKs for a
particular project.

We already [extract](https://github.com/getsentry/relay/blob/master/relay-server/src/metrics_extraction/transactions/mod.rs#L174) os.name on transaction metrics so os.name exists on [indexer/strings.py](https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py#L128).

To determine the "platform" a mobile span originated from:
1. check if contexts.app exists - this tells us if the span originated from an app (as opposed to mobile browser - since we are currently focused on app use cases only)
2. extract the value on `os.name` (iOS, Android) on the os context

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -187,7 +187,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 ("", "ttfd"),
                 ("", "ttid"),
                 ("span.", "main_thread"),
-                ("", "platform"),
+                ("", "os.name"),
             ]
             .map(|(prefix, key)| TagSpec {
                 key: format!("{prefix}{key}"),
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -187,6 +187,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 ("", "ttfd"),
                 ("", "ttid"),
                 ("span.", "main_thread"),
+                ("", "platform"),
             ]
             .map(|(prefix, key)| TagSpec {
                 key: format!("{prefix}{key}"),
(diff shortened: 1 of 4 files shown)


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -201,6 +204,14 @@ pub fn extract_shared_tags(event: &Event) -> BTreeMap<SpanTagKey, String> {
 
     if MOBILE_SDKS.contains(&event.sdk_name()) {
         tags.insert(SpanTagKey::Mobile, "true".to_owned());
+
+        if event.context::<AppContext>().is_some() {
```
jjbayer, line 211: Might be worth adding a comment here why we limit this to apps.

> check if contexts.app exists - this tells us if the span originated from an app (as opposed to mobile browser - since we are currently focused on app use cases only)

relay-dynamic-config/src/defaults.rs
```diff
@@ -187,6 +187,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 ("", "ttfd"),
                 ("", "ttid"),
                 ("span.", "main_thread"),
+                ("", "os.name"),
```
jjbayer, line 190: Do we need this tag on both `exclusive_time` and `exclusive_time_light`?

jjbayer (approved):
Thanks for contributing! OS name should be low-cardinality so this should be fine. See question about what metrics need this tag.

relay-dynamic-config/src/defaults.rs
shruthilayaj: Good question! not right now, but would be nice to have for some of the app.start* spans that we don't plan to facet by transaction in the app startup module. 

Related to this however, the `ttid` and `ttfd` tag will never need `exclusive_time_light` since the tags will only ever be used in the context of a "transaction".

PR #2826, opened 2023-12-07
feat(spans): Keep domain + extension for image resources

jjbayer:
Group `resource.img` spans by domain + extension. Known path segments like `data` are also allowed.

This PR does not yet enable span ingestion for all orgs (still restricted to `all-modules`).

ref: https://github.com/getsentry/sentry/issues/61161

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -562,6 +562,13 @@ mod tests {
 
     span_description_test!(
         resource_query_params2,
+        "https://data.domain.com/data/guide123.gif?jzb=3f535634H467g5-2f256f&ct=1234567890&v=1.203.0_prod",
+        "resource.img",
+        "https://*.domain.com/data/*.gif"
+    );
+
+    span_description_test!(
+        resource_query_params2_script,
         "https://data.domain.com/data/guide123.js?jzb=3f535634H467g5-2f256f&ct=1234567890&v=1.203.0_prod",
         "resource.script",
         "https://*.domain.com/data/guide*.js"
@@ -625,6 +632,13 @@ mod tests {
 
     span_description_test!(
         random_string1,
+        "https://static.domain.com/6gezWf_qs4Wc12Nz9rpLOx2aw2k/foo-99",
+        "resource.img",
+        "https://*.domain.com/*/*"
+    );
+
+    span_description_test!(
+        random_string1_script,
         "https://static.domain.com/6gezWf_qs4Wc12Nz9rpLOx2aw2k/foo-99",
         "resource.script",
         "https://*.domain.com/*/foo-*"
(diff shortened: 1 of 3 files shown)


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -298,14 +301,20 @@ fn scrub_resource_filename(path: &str) -> String {
...
-    let mut joined = segments.join("/");
-    if !extension.is_empty() {
-        joined.push('.');
-        joined.push_str(extension);
-    }
+    let basename = if ty == "img" {
+        Cow::Borrowed("*")
+    } else {
+        scrub_resource_segment(basename)
+    };
```
jjbayer, line 308: This is the main functional change in this PR: for `img` spans, only keep the extension.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -298,14 +301,20 @@ fn scrub_resource_filename(path: &str) -> String {
         extension = "";
     }
 
-    let mut segments = base_path.split('/').map(scrub_resource_segment);
-    let mut joined = segments.join("/");
```
jjbayer, line 302: Drive-by fix: Segments are already handled in `scrub_resource`, so no need to do it again here.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -553,16 +562,30 @@ mod tests {
...
         "resource.img",
-        "https://*.domain.com/data/guide.gif"
+        "https://*.domain.com/data/*.gif"
+    );
+
+    span_description_test!(
+        resource_no_ids_img,
+        "https://data.domain.com/something/guide.gif",
+        "resource.img",
+        "https://*.domain.com/*/*.gif"
```
jjbayer, line 595: @DominikB2014 these tests illustrate what the new span descriptions will look like.

olksdr (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -723,15 +746,15 @@ mod tests {
 
     span_description_test!(
         resource_img_comma_with_extension,
-        "https://example.org/p/fit=cover,width=150,height=150,format=auto,quality=90/media/photosV2/weird-stuff-123-234-456.jpg",
-        "resource.img",
-        "https://example.org/*/media/*/weird-stuff-*-*-*.jpg"
+        "https://example.org/p/fit=cover,width=150,height=150,format=auto,quality=90/media/photosV2/weird-stuff-123-234-456.js",
+        "resource.script",
```
iker-barriocanal, line 771: Why are we changing existing tests, instead of adding new ones? Should we update the test names if we update what they're testing?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
jjbayer: The intention of the test was to make sure commas don't make it into the scrubbed description, and that is now trivially true for `resource.img` because the entire base name gets removed. But you are right, I will duplicate the tests instead.

DominikB2014 (approved):

DominikB2014:
One thing we might have to consider in the future is that we can still have images that don't have the span.op:resource.img, once merged, I can monitor how it looks in this case.

PR #2827, opened 2023-12-07
feat(perf): Fewer tags on performance score metrics

jjbayer:
Performance score measurements are special measurements derived by relay. As such, they do not require the full set of metric tags that SDK-side measurements get.

Limiting the number of tags will reduce cardinality by ~20%, according to my analysis.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/transactions/mod.rs
+++ b/relay-server/src/metrics_extraction/transactions/mod.rs
@@ -28,11 +28,11 @@ const PLACEHOLDER_UNPARAMETERIZED: &str = "<< unparameterized >>";
 ///
 /// These are a subset of "universal" tags.
 const PERFORMANCE_SCORE_TAGS: [CommonTag; 6] = [
-    CommonTag::Release,
-    CommonTag::Environment,
-    CommonTag::Transaction,
-    CommonTag::TransactionOp,
     CommonTag::BrowserName,
+    CommonTag::Environment,
+    CommonTag::GeoCountryCode,
+    CommonTag::Release,
+    CommonTag::Transaction,
     CommonTag::TransactionOp,
 ];
 
@@ -692,6 +692,7 @@ mod tests {
                 tags: {
                     "browser.name": "Chrome",
                     "environment": "fake_environment",
+                    "geo.country_code": "US",
                     "release": "1.2.3",
                     "transaction": "gEt /api/:version/users/",
                     "transaction.op": "mYOp",
@@ -709,6 +710,7 @@ mod tests {
                 tags: {
                     "browser.name": "Chrome",
                     "environment": "fake_environment",
+                    "geo.country_code": "US",
                     "release": "1.2.3",
                     "transaction": "gEt /api/:version/users/",
                     "transaction.op": "mYOp",
@@ -726,6 +728,7 @@ mod tests {
                 tags: {
                     "browser.name": "Chrome",
                     "environment": "fake_environment",
+                    "geo.country_code": "US",
                     "release": "1.2.3",
                     "transaction": "gEt /api/:version/users/",
                     "transaction.op": "mYOp",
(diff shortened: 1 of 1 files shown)


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -629,6 +680,57 @@ mod tests {
...
+                    [
+                        0.0,
+                    ],
+                ),
+                tags: {
+                    "browser.name": "Chrome",
+                    "environment": "fake_environment",
+                    "release": "1.2.3",
+                    "transaction": "gEt /api/:version/users/",
+                    "transaction.op": "mYOp",
```
jjbayer: Without the tag filter, this would include platform, dist, etc.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -275,9 +294,27 @@ impl TransactionExtractor<'_> {
                     None => continue,
                 };
 
+                // We treat a measurement as "performance score" if its name is the name of another
+                // measurement prefixed by `score.`.
+                let is_performance_score = name == "score.total"
+                    || name
+                        .strip_prefix("score.weight.")
+                        .or_else(|| name.strip_prefix("score."))
+                        .map_or(false, |suffix| measurement_names.contains(suffix));
```
jjbayer, line 303: This is clumsy: It would be better to mark measurements as "performance score" somehow, but they might originate from a downstream relay so any kind of flag would have to be serialized into the event JSON.

relay-server/src/metrics_extraction/transactions/mod.rs
jjbayer: @edwardgou-sentry @k-fish apart from the performance score UI, could there be a more generic measurements-related part of the UI that breaks if these metrics come with only a subset of tags?

iker-barriocanal (approved):

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -24,6 +24,18 @@ pub mod types;
...
+/// Tags we set on metrics for performance score measurements (e.g. `score.lcp.weight`).
+///
+/// These are a subset of "universal" tags.
+const PERFORMANCE_SCORE_TAGS: [CommonTag; 6] = [
+    CommonTag::Release,
+    CommonTag::Environment,
+    CommonTag::Transaction,
+    CommonTag::TransactionOp,
+    CommonTag::BrowserName,
+    CommonTag::TransactionOp,
```
edwardgou-sentry: Seems like TransactionOp appears here twice, is it intentional? Also, not a hard requirement but I think we may want to include GeoCountryCode in case we may need to facet by country code in the future

relay-server/src/metrics_extraction/transactions/mod.rs
edwardgou-sentry: As far as I know, Dashboards is the only other place users would be able to write a query for Performance Scores using metrics. But I'm pretty sure that if a metrics query breaks due to missing tags, it will fall back to the indexed dataset instead. I think this behaviour should be ok

relay-server/src/metrics_extraction/transactions/mod.rs
jjbayer: Whoops, looks like I forgot to replace that second one with country code. Fixed now, thanks for catching it!

edwardgou-sentry (approved):

k-fish (approved):
👍 looks good. On top of limiting tags, I also mentioned in my original doc that it's possible we don't store distributions for perf score since it just needs sum + count (avg), which are both counters, which could drastically lower storage.

jjbayer:
> it's possible we don't store distributions for perf score since it just needs sum + count (avg), which are both counters, which could drastically lower storage.

@k-fish good point. Could we build this feature by just using the "counter" metric type (which supports floats)?

k-fish:
@jjbayer yes, it could be count (sum) and count (count) then divide them.

PR #2829, opened 2023-12-07
test(flaky): Try to fix flaky project config test

Dav1dde:
Failed here: https://github.com/getsentry/relay/actions/runs/7127328182/job/19407012315

Assumption is that there is a slight race condition until the error is added as a test failure, in a separate test we have an additional delay before clearing test failures. Moved the check and clear after an already existing sleep to hopefully fix this test.

#skip-changelog

### DIFF
--- a/tests/integration/test_projectconfigs.py
+++ b/tests/integration/test_projectconfigs.py
@@ -294,34 +294,31 @@ def test_unparsable_project_config(buffer_config, mini_sentry, relay):
         }
     } == data
 
+    def assert_clear_test_failures():
+        try:
+            assert {str(e) for _, e in mini_sentry.test_failures} == {
+                f"Relay sent us event: error fetching project state {public_key}: invalid type: integer `99`, expected a string",
+            }
+        finally:
+            mini_sentry.test_failures.clear()
+
     # Event is not propagated, relay logs an error:
-    try:
-        relay.send_event(project_key)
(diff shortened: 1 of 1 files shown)


jjbayer (approved):
Let's try it!

olksdr (approved):

tests/integration/test_projectconfigs.py
```diff
@@ -294,34 +294,31 @@ def test_unparsable_project_config(buffer_config, mini_sentry, relay):
         }
     } == data
 
+    def assert_clear_test_failures():
+        try:
+            assert {str(e) for _, e in mini_sentry.test_failures} == {
+                f"Relay sent us event: error fetching project state {public_key}: invalid type: integer `99`, expected a string",
+            }
+        finally:
+            mini_sentry.test_failures.clear()
```
iker-barriocanal, line 303: Not related to the flakiness, but why do we need `try..finally` here? Does it make any difference in a test?

iker-barriocanal (approved):

tests/integration/test_projectconfigs.py
Dav1dde: I copied it and wasn't sure myself. I **think** it is here in case there are other failures, they would already raise an error from the assert and keeping the test failures it would show them again when exiting the test (through the normal assertion behaviour).

PR #2831, opened 2023-12-07
chore(spans): Only tag `ttid` & `ttfd` with `exclusive_time` metric

shruthilayaj:
The ttid, ttfd and span.main_thread tag will never need exclusive_time_light since the 
tags will only ever be used in the context of a "transaction".
Updated snapshot test to show an `exclusive_time_light` bucket
isn't created.

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -184,8 +184,6 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
             tags: [
                 ("", "device.class"),
                 ("", "release"),
-                ("", "ttfd"),
-                ("", "ttid"),
                 ("span.", "main_thread"),
                 ("", "os.name"),
             ]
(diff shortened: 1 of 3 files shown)


relay-dynamic-config/src/defaults.rs
```diff
@@ -197,6 +195,17 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
             })
             .into(),
         },
+        TagMapping {
+            metrics: vec![LazyGlob::new("d:spans/exclusive_time@millisecond".into())],
+            tags: [("", "ttfd"), ("", "ttid")]
+                .map(|(prefix, key)| TagSpec {
+                    key: format!("{prefix}{key}"),
```
```suggestion
            tags: ["ttfd", "ttid"]
                .map(|key| TagSpec {
                    key: key.to_owned(),
```

relay-server/src/metrics_extraction/event.rs
```diff
@@ -1209,7 +1216,9 @@ mod tests {
 
         assert!(!metrics.is_empty());
         for metric in metrics {
-            if metric.name == "c:spans/count_per_op@none" {
+            if metric.name == "c:spans/count_per_op@none"
+                || metric.name == "d:spans/exclusive_time_light@millisecond"
```
jjbayer: Instead of skipping this one, should we assert that the tags are not set on `exclusive_time_light`?

jjbayer (approved):
Thanks for doing this!

relay-dynamic-config/src/defaults.rs
shruthilayaj: also moved `span.main_thread` to not tag `exclusive_time_light`, addressed this change [here](https://github.com/getsentry/relay/pull/2831/commits/d3ab814059f8a20ee66b70a152d643402f4ea42e?diff=split&w=1#diff-b391d3acb7ee266caa09903f615c9e66567dd2a2a630d07435971931175adb5eR184).

relay-server/src/metrics_extraction/event.rs
shruthilayaj: Addressed [here](https://github.com/getsentry/relay/pull/2831/commits/d3ab814059f8a20ee66b70a152d643402f4ea42e?diff=split&w=1#diff-d25a118d71db14a3d77408d5cb5a71753d453771343e3fc77d84305036a04e81R1219-R1224)!

PR #2832, opened 2023-12-07
ref(metrics): Store MRI instead of String on metrics Bucket

Dav1dde:
Store the parsed MRI in the Bucket instead of the plain string. We re-parse the MRI in multiple spots which means we always have to do the error handling even though this will never fail.

Storing the MRI instead of the string form allows us to get rid of the surplus parses and just generally gives us more assertions and information in the type system.

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregator.rs
+++ b/relay-metrics/src/aggregator.rs
@@ -35,12 +35,9 @@ impl From<AggregateMetricsErrorKind> for AggregateMetricsError {
 #[derive(Debug, Error, PartialEq)]
 #[allow(clippy::enum_variant_names)]
 enum AggregateMetricsErrorKind {
-    /// A metric bucket had invalid characters in the metric name.
-    #[error("found invalid characters: {0}")]
-    InvalidCharacters(String),
     /// A metric bucket had an unknown namespace in the metric name.
-    #[error("found unsupported namespace: {0}")]
-    UnsupportedNamespace(MetricNamespace),
+    #[error("found unsupported namespace")]
+    UnsupportedNamespace,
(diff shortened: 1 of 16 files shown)


relay-metrics/src/aggregator.rs
```diff
@@ -35,12 +35,9 @@ impl From<AggregateMetricsErrorKind> for AggregateMetricsError {
 #[derive(Debug, Error, PartialEq)]
 #[allow(clippy::enum_variant_names)]
 enum AggregateMetricsErrorKind {
-    /// A metric bucket had invalid characters in the metric name.
-    #[error("found invalid characters: {0}")]
-    InvalidCharacters(String),
     /// A metric bucket had an unknown namespace in the metric name.
-    #[error("found unsupported namespace: {0}")]
-    UnsupportedNamespace(MetricNamespace),
```
Dav1dde, line 43: This is always `MetricNamespace::Unsupported`, no point in having it as an argument

Dav1dde:
This is tricky because the type of the bucket comes from its value and the MRI is carrying a type as well. In the future we should maybe store the bucket differently:

```rs
struct Bucket {
    name: Cow<'a, str>,
    namespace: MetricNamespace, 
    unit: MetricUnit, 
    ...
}
```

Dav1dde:
This is tricky because the type of the bucket comes from its value and the MRI is carrying a type as well. In the future we should maybe store the bucket differently:

```rs
struct Bucket {
    name: Cow<'a, str>,
    namespace: MetricNamespace, 
    unit: MetricUnit, 
    ...
}
```

PR #2834, opened 2023-12-10
ref(spans): List metric tags explicitly

jjbayer:
List every tag for every metric explicitly. This nonfunctional change is the first step towards more selective tagging (tag only when you need it, and only on the metrics that actually need it).

ref: https://github.com/getsentry/team-ingest/issues/252

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -341,10 +341,12 @@ pub struct Tag {
 }
 
 impl Tag {
+    /// Prepares a tag with a given tag name.
     pub fn with_key(key: impl Into<String>) -> Self {
-        Self { key }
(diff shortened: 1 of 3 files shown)


relay-dynamic-config/src/metrics.rs
```diff
@@ -335,6 +335,71 @@ impl TagSpec {
...
+    pub fn when(self, condition: RuleCondition) -> TagSpec {
+        let Self { key, field, value } = self;
+        TagSpec {
+            key,
+            field,
+            value,
+            condition: Some(condition),
+        }
+    }
+}
```
jjbayer, line 401: These are just helpers to build a `TagSpec` in code.

olksdr (approved):

PR #2841, opened 2023-12-11
ref(crons): Rename DataCategory::MONITOR -> DataCategory::MonitorCheckIn

evanpurkhiser:
Before we start using this everywhere let's fix this naming so we don't
have to deal with this being confusing for the rest of time.

#skip-changelog

### DIFF
--- a/relay-base-schema/src/data_category.rs
+++ b/relay-base-schema/src/data_category.rs
@@ -42,7 +42,7 @@ pub enum DataCategory {
     /// contrast, `transaction` only guarantees that metrics have been accepted for the transaction.
     TransactionIndexed = 9,
     /// Monitor check-ins.
-    Monitor = 10,
+    MonitorCheckIn = 10,
     /// Indexed Profile
     ///
     /// This is the category for indexed profiles that will be stored later.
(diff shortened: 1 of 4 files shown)


evanpurkhiser:
This would be a breaking change for the exported python library. AFAICT we haven't used it at all yet in sentry.

I think it would be good to get this in before we actually start using it

rjo100:
Need to update this as well?
https://github.com/getsentry/relay/blob/master/py/sentry_relay/consts.py

evanpurkhiser:
Good call out. This appears to be auto-generated. Will have to figure out what's going on here

evanpurkhiser:
I think this would be a breaking change for https://develop.sentry.dev/sdk/rate-limiting/ as well.

We would need to return both `monitor` and `monitor_checkin` when we rate-limit them.

evanpurkhiser:
Going to leave this open, but I think it's unlikely we'll change this right now.

evanpurkhiser:
@jan-auer could you comment on if changing this is feasible or even worthwhile? Maybe the naming change isn't necessary.

evanpurkhiser:
Closing for now. May revisit in the future.

PR #2844, opened 2023-12-12
build: Update axum and http

jan-auer:
We would like to update the Sentry SDK, which requires to update `axum` and related dependencies. As `reqwest` has not updated to the latest version of `http` and `hyper` yet, we will temporarily run two versions of the frameworks side-by-side and convert between their types in the forward endpoint.

The `axum-server` library also has not released their latest version yet. Since the only thing needed from axum-server is graceful shutdown, we implement this ourselves now following the example in the `axum` repository.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/server.rs
+++ b/relay-server/src/actors/server.rs
@@ -54,6 +54,7 @@ fn listen(config: &Config) -> Result<TcpListener, ServerError> {
 
 type App = NormalizePath<axum::Router>;
 
+/// Build the axum application with all routes and middleware.
 fn make_app(service: ServiceState) -> App {
     // Build the router middleware into a single service which runs _after_ routing. Service
     // builder order defines layers added first will be called first. This means:
@@ -80,6 +81,8 @@ fn make_app(service: ServiceState) -> App {
         .layer(middleware)
         .with_state(service);
 
+    // Add middlewares that need to run _before_ routing, which need to wrap the router. This are
+    // especially middlewares that modify the request path for the router:
     NormalizePath::new(router)
 }
 
--- a/relay-server/src/actors/server.rs
+++ b/relay-server/src/actors/server.rs
@@ -96,7 +96,6 @@ pub struct HttpServer {
 impl HttpServer {
     pub fn new(config: Arc<Config>, service: ServiceState) -> Result<Self, ServerError> {
         let listener = listen(&config)?;
-        // let server = build_server(listener, &config);
 
         Ok(Self {
             config,
@@ -132,7 +131,7 @@ async fn serve(mut listener: TcpListener, app: App, config: Arc<Config>) {
     let mut graceful_shutdown = Controller::shutdown_handle();
     let (connections_tx, connections_rx) = watch::channel(());
 
-    // Create a connection builder to reuse across connections
+    // Create a connection builder to reuse across connections.
     let builder = Arc::new(connection_builder(&config));
 
     loop {
--- a/relay-server/src/actors/server.rs
+++ b/relay-server/src/actors/server.rs
@@ -86,7 +86,7 @@ fn make_app(service: ServiceState) -> App {
 /// HTTP server service.
 ///
 /// This is the main HTTP server of Relay which hosts all [services](ServiceState) and dispatches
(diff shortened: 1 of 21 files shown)


jan-auer:
There are still test failures with:
> failed to parse JSON response, EOF while parsing a value at line 1 column 0

relay-server/src/actors/server.rs
```diff
@@ -1,143 +1,231 @@
...
+    Ok(listener.try_into()?)
+}
+
+type App = NormalizePath<axum::Router>;
+
+fn make_app(service: ServiceState) -> App {
+    // Build the router middleware into a single service which runs _after_ routing. Service
+    // builder order defines layers added first will be called first. This means:
+    //  - Requests go from top to bottom
+    //  - Responses go from bottom to top
```
olksdr, line 62: Maybe we can move this to the function docs instead?

relay-server/src/actors/server.rs
```diff
@@ -1,143 +1,231 @@
...
-        if config.tls_listen_addr().is_some()
-            || config.tls_identity_password().is_some()
-            || config.tls_identity_path().is_some()
-        {
-            return Err(ServerError::TlsNotSupported);
-        }
-
-        Ok(Self { config, service })
+        let listener = listen(&config)?;
+        // let server = build_server(listener, &config);
```
olksdr: Commented code to be removed?

relay-server/src/actors/server.rs
```diff
@@ -1,143 +1,231 @@
...
-            match timeout {
-                Some(timeout) => handle.graceful_shutdown(Some(timeout)),
-                None => handle.shutdown(),
+async fn serve(mut listener: TcpListener, app: App, config: Arc<Config>) {
+    // Create channels to track shutdown and open connections.
+    let hard_shutdown = monitor_hard_shutdown();
+    let mut graceful_shutdown = Controller::shutdown_handle();
+    let (connections_tx, connections_rx) = watch::channel(());
+
+    // Create a connection builder to reuse across connections
```
olksdr:
```suggestion
    // Create a connection builder to reuse across connections.
```

relay-server/src/endpoints/forward.rs
```diff
@@ -112,8 +146,8 @@ impl fmt::Debug for ForwardRequest {
 }
 
 impl UpstreamRequest for ForwardRequest {
-    fn method(&self) -> Method {
-        self.method.clone()
+    fn method(&self) -> reqwest::Method {
+        self.method.as_str().parse().unwrap()
```
olksdr, line 150: should we fallback to some default method instead of `unwrap` here ?

olksdr (approved):
LGTM! 

Using new request extractor helped to simplify code a lot, removing all explicit generics and bounds. Nice!

Left few comments/questions.

relay-server/src/endpoints/forward.rs
jan-auer: Good question. The panic here is unreachable, since the parsed method is already valid. A default would also create an invalid request.

relay-server/src/actors/server.rs
jan-auer: Can do, does it make more sense for you there? The order of the middlewares is something that's important to the `ServiceBuilder`, hence the placement within the function body.

jan-auer:
> Using new request extractor helped to simplify code a lot

That's right. We might have even more potential to simplify code now, since we no longer need to extract a generic Request, which means we can derive `FromRequest` on structs containing `Request` again.

relay-server/src/actors/server.rs
olksdr: For me yes. 
Since the entire function is actually dedicated to putting app together and the docs will do just good if they will be on the function level.

jan-auer: As I was moving it and adding back an old comment that got lost, I looked at it again but it simply doesn't make sense to me. This code comment is relevant only for the implementation of the function, as it clarifies on two specifics:
1. In which order the middlewares are registered and why
2. Why some middlewares wrap the router and some others do not. 

To a user of this function, none of these matter. In fact, the user of this function does not even see that router middlewares are applied.

PR #2845, opened 2023-12-12
feat(spans): Resource response sizes as measurements

jjbayer:
Until now, we were able to annotate span metrics and span kafka messages with arbitrary key value pairs using `span.sentry_tags`. But `sentry_tags` only supports string values, which is why `span.measurements` was introduced.

Copy resource response sizes into measurements so we can sort by them in the spans dataset.

#skip-changelog


DominikB2014 (approved):
This is awesome!

olksdr (approved):

PR #2846, opened 2023-12-12
fix(spans): Parse quotes in MySQL

jjbayer:
Some of the dialect-specific parsing logic did not apply because we use a wrapped dialect `DialectWithParameters`, which has the wrong `type_id()`.

Until https://github.com/sqlparser-rs/sqlparser-rs/pull/1065 gets merged, use a fork of `sqlparser` to make `DialectWithParameters` behave like the underlying dialect.

ref: https://github.com/getsentry/relay/issues/2847

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/sql/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/mod.rs
@@ -617,6 +617,28 @@ mod tests {
         "DELETE FROM some_table WHERE id IN (%s)"
     );
 
+    scrub_sql_test!(escape_quote, r#"SELECT 'Wayne\'s World'"#, "SELECT %s");
+
+    scrub_sql_test!(
+        escape_double_quote,
+        r#"SELECT '{"json": "yes"}'"#,
+        "SELECT %s"
+    );
+
+    scrub_sql_test_with_dialect!(
+        mysql_escape_quote,
+        "mysql",
+        r#"SELECT "Wayne's World"#,
+        "SELECT %s"
+    );
+
+    scrub_sql_test_with_dialect!(
+        mysql_escape_double_quote,
+        "mysql",
+        r#"SELECT "{\"json\": \"yes\"}""#,
+        "SELECT %s"
+    );
+
     scrub_sql_test!(
         bytesa,
         r#"SELECT "t"."x", "t"."arr"::bytea, "t"."c" WHERE "t"."id" IN (%s, %s)"#,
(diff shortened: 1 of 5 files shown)


jjbayer:
Blocked on https://github.com/sqlparser-rs/sqlparser-rs/pull/1065.

relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -617,6 +617,28 @@ mod tests {
...
+        r#"SELECT "Wayne's World""#,
+        "SELECT %s"
+    );
+
+    scrub_sql_test_with_dialect!(
+        mysql_escape_double_quote,
+        "mysql",
+        r#"SELECT "{\"json\": \"yes\"}""#,
+        "SELECT %s"
+    );
```
jjbayer, line 640: These tests would previously fail.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -429,6 +429,10 @@ impl DialectWithParameters {
 }
 
 impl Dialect for DialectWithParameters {
+    fn dialect(&self) -> std::any::TypeId {
+        self.0.dialect()
+    }
```
jjbayer, line 434: This informs the parser to treat the current dialect the same as the wrapped dialect, for example MySQL.

olksdr (approved):

Dav1dde (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -429,6 +429,10 @@ impl DialectWithParameters {
 }
 
 impl Dialect for DialectWithParameters {
+    fn dialect(&self) -> std::any::TypeId {
```
iker-barriocanal, line 432: nit: should we `#[inline]`?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: Should we?

> Note: The rustc compiler automatically inlines functions based on internal heuristics. Incorrectly inlining functions can make the program slower, so this attribute should be used with care.

https://doc.rust-lang.org/stable/reference/attributes/codegen.html?highlight=inline#the-inline-attribute

PR #2849, opened 2023-12-13
ref(cardinality): Use a Lua script and in-memory cache for the cardinality limiter

Dav1dde:
Changes the cardinality limiter implementation to use a Lua script. The Lua script evaluates each individual hash and returns a decision wether this hash is accepted or rejected by the cardinality limiter, this allows us to save one additional round trip to Redis.

Additionally there is an in memory cache infront of Redis remembering previous decisions from the Redis script, allowing us to skip Redis calls alltogether.

Epic: https://github.com/getsentry/relay/issues/2717


relay-cardinality/src/statsd.rs
```diff
@@ -8,7 +8,7 @@ pub enum CardinalityLimiterCounters {
     ///  - `scope`: The scope of check operation.
     #[cfg(feature = "redis")]
     Rejected,
-    /// Incremented for every loaded Redis set.
+    /// Incremented for every redis cardinality check.
```
Dav1dde: Maybe we should rename the entire metric? We dont have dashboards yet, but `Read` also somewhat fits, that is how many times we went to redis to read cardiniality data

relay-cardinality/src/redis.rs
```diff
@@ -59,58 +164,41 @@ impl RedisSetLimiter {
...
-        ))?;
-        let mut new_hashes = HashSet::new();
+        let keys = self
+            .window
+            .iter(timestamp)
+            .map(|time_bucket| self.to_redis_key(organization_id, &item_scope, time_bucket));
 
-        let mut new_cardinality = active_window.len();
-        let mut rejected = 0;
+        let hashes = entries.iter().map(|entry| entry.hash);
```
Dav1dde, line 216: Histogram for `entries.len()`?

relay-cardinality/src/cardinality.lua
```diff
@@ -0,0 +1,99 @@
...
+        if is_new_item then
+            current_cardinality = current_cardinality + 1
+            needs_expire = true
+        end
+
+        -- add to the remainings sets
+        for i = 2, #KEYS do
+            redis.call('SADD', KEYS[i], value)
+        end
+    elseif redis.call('SISMEMBER', working_set, value) == 1 then
```
jjbayer: Should we use batch calls here (`SMISMEMBER`)? Something like

1. Check `budget := max_cardinality - current_cardinality`
2. If budget >= len(values), batch-add and be happy.
3. Else, run `SMISMEMBER` to determine membership for all values.
4. Take at most `budget` values for which `SMISMEMBER` returned `0`, and batch-add them to the set.
5. Return `ACCEPTED` for values for which either membership was `1`, or was added in step 4. `REJECTED` for remaining values.

relay-cardinality/src/redis.rs
```diff
@@ -1,27 +1,131 @@
...
+            Self::Rejected
+        })
+    }
+}
+
+#[derive(Debug)]
+struct CardinalityScriptResult {
+    cardinality: u64,
+    rejected: usize,
+    bitmap: Vec<Status>,
```
jjbayer: nit

```suggestion
    statuses: Vec<Status>,
```

relay-cardinality/src/redis.rs
```diff
@@ -1,27 +1,131 @@
...
+            .next()
+            .ok_or_else(|| {
+                redis::RedisError::from((
+                    redis::ErrorKind::TypeError,
+                    "Expected cardinality as the first result from the cardinality script",
+                ))
+            })
+            .and_then(FromRedisValue::from_redis_value)?;
+
+        let mut bitmap = Vec::with_capacity(iter.len() - 1);
```
jjbayer: Does `iter.len()` return the original length, even if the iterator already consumed one element?

relay-cardinality/src/redis.rs
```diff
@@ -1,27 +1,131 @@
...
+        Self(Script::new(include_str!("cardinality.lua")))
+    }
+
+    fn invoke(
+        &self,
+        con: &mut Connection,
+        limit: usize,
+        expire: u64,
+        keys: impl Iterator<Item = String>,
+        hashes: impl Iterator<Item = u32>,
```
jjbayer: nit: Should we use the same order of argments here as in the redis script?

relay-cardinality/src/redis.rs
jjbayer: I would like that (unless we can already get that information indirectly).

relay-cardinality/src/redis.rs
```diff
@@ -59,58 +164,41 @@ impl RedisSetLimiter {
...
-            // The expiry is a off by `window.granularity_seconds`,
-            // but since this is only used for cleanup, this is not an issue.
-            client.add_to_sets(keys, &new_hashes, self.window.window_seconds)?;
-
-            metric!(
-                counter(CardinalityLimiterCounters::RedisHashUpdate) += new_hashes.len() as i64,
-                scope = &item_scope,
-            );
-        }
+        let result = std::iter::zip(entries, result.bitmap)
```
jjbayer: Should we verify that entries and result.bitmap have the same length?

relay-cardinality/src/statsd.rs
jjbayer: Is the number of times that we call the Lua script? Then maybe `RedisCheck` makes sense?

We could also get rid of `RedisHashCheck`, in favor of a histogram metric that tracks the number of hashes per check.

jjbayer (commented):
Looks good, see comment on batch calls. I assume that getting reducing the number of redis commands is still beneficial, even within a Lua script.

relay-cardinality/src/redis.rs
Dav1dde: Thanks, that is a refactoring bug and was supposed to use the underlaying source not the iterator!

relay-cardinality/src/redis.rs
Dav1dde: We kind of can get the average by dividing the total amount of hashes by the amount of set invocations.

relay-cardinality/src/cardinality.lua
Dav1dde: This is implemented now, thanks for making me look into this!

relay-cardinality/src/cardinality.lua
```diff
@@ -0,0 +1,174 @@
...
+-- to all sets in `KEYS`.
+--
+-- Returns the total amount of values added to the 'working set'.
+local function sadd(t, offset, max)
+    local added = 0;
+
+    for i = 1, #KEYS do
+        local is_working_set = i == 1
+
+        for from, to in batches(#t, 7000, offset, max) do
```
jjbayer, line 94: I think we should batch exclusively on the Relay side. That would simplify this script, and we need some sort of batching there anyway.

relay-cardinality/src/cardinality.lua
```diff
@@ -0,0 +1,174 @@
...
+end
+
+
+local ACCEPTED = true
+local REJECTED = false
+local HASHES_OFFSET = 2 -- arg1: max_cardinality, arg2: expiry
+
+local working_set = KEYS[1]
+local max_cardinality = tonumber(ARGV[1])
+local expire = tonumber(ARGV[2])
```
jjbayer: nit

```suggestion
local expiry = tonumber(ARGV[2])
```

relay-cardinality/src/cardinality.lua
```diff
@@ -0,0 +1,174 @@
...
+-- Update cardinality, at this point cardinality cannot change anymore,
+-- the maximum amount of keys that can be added, have been added to the set.
+results[1] = current_cardinality
+
+-- If we ran out of budget, check the remaining items for membership
+if budget <= 0 and offset < #ARGV then
+    for arg_i = offset + 1, #ARGV do
+        local value = ARGV[arg_i]
+
+        if redis.call('SISMEMBER', working_set, value) == 1 then
```
jjbayer, line 163: Can we leave a code comment here that we should use `SMISMEMBER` as soon as it's available on our cluster?

relay-cardinality/src/redis.rs
```diff
@@ -1,27 +1,145 @@
...
+            Self::Rejected
+        })
+    }
+}
+
+#[derive(Debug)]
+struct CardinalityScriptResult {
+    cardinality: u64,
+    rejected: usize,
+    stati: Vec<Status>,
```
jjbayer: As much as I like Latin: https://en.wiktionary.org/wiki/statuses#English

```suggestion
    statuses: Vec<Status>,
```

jjbayer (approved):
Still some open comments, but no blockers as far as I am concerned!

relay-cardinality/src/cardinality.lua
Dav1dde: I like batching in the lua script, makes the relay side easier to read and we have a guarantee that this never fails in lua (one less error to handle and invariant).

The lua implementation would get slightly easier, but we would still need a slicing function like `batches` (to slice `offset` and `max`), which now is just a minimal overhead to the `batches` function (just a change of the max value and the offset).

So overall imo we don't gain much by moving it to relay.

relay-cardinality/src/statsd.rs
Dav1dde: I believe we can get this number from a histogram as well, right?

If we add the histrogram for the amount of hashes sent to redis, we can also get the count from the histogram. -> I'll switch to the histrogram

relay-cardinality/src/cardinality.lua
jjbayer: > but we would still need a slicing function like batches

That makes sense!

> So overall imo we don't gain much by moving it to relay.

Even if we keep batching on the Lua side, don't we also need batching on the Rust side? Not sure if we would run into size limits or network timeouts otherwise.

relay-cardinality/src/redis.rs
```diff
@@ -135,11 +135,40 @@ impl CardinalityScript {
...
+        std::iter::zip(self.entries, self.statuses)
+    }
+}
+
 /// Implementation uses Redis sets to keep track of cardinality.
 pub struct RedisSetLimiter {
     redis: RedisPool,
     window: SlidingWindow,
     script: CardinalityScript,
+    cache: Cache,
```
jjbayer, line 171: nit: Does the cache have to be a property of `RedisSetLimiter`? Since it's independent of redis, could it be part of `CardinalityLimiter`?

Or go fancy and use the [Composite Pattern](https://en.wikipedia.org/wiki/Composite_pattern):

```mermaid
classDiagram
    Limiter <|-- CachedLimiter
    Limiter <|-- RedisSetLimiter
    CachedLimiter "1" *-- "2" Limiter
```

relay-cardinality/src/redis.rs
```diff
@@ -278,22 +283,94 @@ impl Limiter for RedisSetLimiter {
...
                 organization_id,
-                item_scope.to_string(),
+                item_scope.clone(),
                 entries,
                 timestamp,
                 limit,
-            )?);
+            )?;
+
+            // This always acquires a write lock, but since
```
jjbayer: This comment looks incomplete.

relay-cardinality/src/redis.rs
```diff
@@ -341,6 +418,250 @@ impl SlidingWindow {
...
+        item_scope: &'a str,
+        window: u64,
+    ) -> CacheUpdate<'a> {
+        let mut inner = self.inner.write().unwrap_or_else(PoisonError::into_inner);
+
+        // If the passed window is newer then the current window, reset the cache to the new window.
+        if window > inner.current_window {
+            inner.current_window = window;
+            inner.cache.clear();
+        } else if window != inner.current_window {
```
jjbayer: nit: I find this slightly easier to understand

```suggestion
        } else if window < inner.current_window {
```

relay-cardinality/src/redis.rs
```diff
@@ -341,6 +418,250 @@ impl SlidingWindow {
...
+            item_scope,
+        } = key;
+
+        self.organization_id == *organization_id && self.item_scope == item_scope
+    }
+}
+
+/// Helper to aggregate cache hits by scope before reporting them to statsd.
+#[derive(Debug)]
+struct HitCounter(hashbrown::HashMap<String, (i64, i64)>);
```
jjbayer: What's the reason for using a non-standard map implementation here?

relay-cardinality/src/redis.rs
```diff
@@ -341,6 +418,250 @@ impl SlidingWindow {
...
+/// Critical section of the [`Cache`].
+#[derive(Debug, Default)]
+struct CacheInner {
+    cache: hashbrown::HashMap<CacheKey, CacheValue>,
+    current_window: u64,
+}
+
+/// Scope specific information of the cache.
+#[derive(Debug, Default)]
+struct CacheValue(hashbrown::HashSet<u32>);
```
jjbayer: I was confused by the term `Value` here for a moment. Given that this _is_ the cache for a given scope, should we name it `ScopedCache` or something like that?

jjbayer: Could we add a comment with the reasoning why we're using `hashbrown` here?

jjbayer (approved):
Couple of questions, but no blockers!

Please update the PR title & description to mention the cache as well as the Lua script.

relay-cardinality/src/redis.rs
Dav1dde: Needed for `entry_ref`, the alternative is pretty awkward (`get_mut` followed by an insert if `None`)

relay-cardinality/src/cardinality.lua
Dav1dde: I am not sure how Redis and the network behaves here, in my local tests it didn't make a difference but there is also no latency and Redis does not do anything else (e.g. rate limiting etc.)

Dav1dde: Would we batch and use a pipeline, I feel like that would make no difference then?

relay-cardinality/src/redis.rs
Dav1dde: 👍 definitely can be improved, initially this was heavily tied into the Lua logic, but it is pretty separate now. Ill do this in a followup PR.

It is still a bit tricky since it does require the same sliding window logic and ties into the splitting by scope logic which is Redis specific.

relay-cardinality/src/redis.rs
Dav1dde: Will add a comment, the `big_set_small_query` benchmark shows a 10% speedup when using `hashbrown`

PR #2852, opened 2023-12-14
ref(cardinality): Recover buckets on cardinality limiter failure

Dav1dde:
Recover buckets on cardinality limiter error instead of discarding all

#skip-changelog

### DIFF
--- a/relay-cardinality/src/limiter.rs
+++ b/relay-cardinality/src/limiter.rs
@@ -115,13 +115,13 @@ impl<T: Limiter> CardinalityLimiter<T> {
                 Some(Entry::new(EntryId(id), item.to_scope()?, item.to_hash()))
             });
 
-            let rejections = self.limiter.check_cardinality_limits(
+            let result = self.limiter.check_cardinality_limits(
                 organization,
                 entries,
                 self.config.cardinality_limit,
             );
 
-            let rejections = match rejections {
+            let rejections = match result {
                 Ok(rejections) => rejections
                     .map(|rejection| rejection.id.0)
                     .collect::<BTreeSet<usize>>(),
--- a/relay-cardinality/src/limiter.rs
+++ b/relay-cardinality/src/limiter.rs
@@ -7,7 +7,7 @@ use std::hash::Hash;
 use relay_statsd::metric;
 
 use crate::statsd::CardinalityLimiterTimers;
-use crate::{OrganizationId, Result};
+use crate::{Error, OrganizationId, Result};
 
 /// Limiter responsible to enforce limits.
 pub trait Limiter {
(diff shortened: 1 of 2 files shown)


relay-cardinality/src/limiter.rs
```diff
@@ -109,17 +109,24 @@ impl<T: Limiter> CardinalityLimiter<T> {
...
             let entries = items.iter().enumerate().filter_map(|(id, item)| {
                 Some(Entry::new(EntryId(id), item.to_scope()?, item.to_hash()))
             });
 
-            let rejections = self
-                .limiter
-                .check_cardinality_limits(organization, entries, self.config.cardinality_limit)?
-                .map(|rejection| rejection.id.0)
-                .collect::<BTreeSet<usize>>();
+            let rejections = self.limiter.check_cardinality_limits(
```
jjbayer: nit

```suggestion
            let result = self.limiter.check_cardinality_limits(
```

relay-server/src/actors/processor.rs
```diff
@@ -1449,21 +1450,29 @@ impl EnvelopeProcessorService {
         enable_cardinality_limiter: bool,
         _organization_id: u64,
         buckets: Vec<Bucket>,
-    ) -> Result<Box<dyn Iterator<Item = Bucket>>, relay_cardinality::Error> {
+    ) -> impl Iterator<Item = Bucket> {
         if !enable_cardinality_limiter {
-            return Ok(Box::new(buckets.into_iter()));
+            return Either::Left(buckets.into_iter());
```
iker-barriocanal, line 1456: IIUC `Left` is for buckets that are accepted and `Right` for those that are rate-limited. Is this correct? Is it worth documenting it?

iker-barriocanal (approved):

relay-server/src/actors/processor.rs
```diff
@@ -1449,21 +1450,29 @@ impl EnvelopeProcessorService {
         enable_cardinality_limiter: bool,
         _organization_id: u64,
         buckets: Vec<Bucket>,
-    ) -> Result<Box<dyn Iterator<Item = Bucket>>, relay_cardinality::Error> {
+    ) -> impl Iterator<Item = Bucket> {
```
jjbayer, line 1453: Nice!

relay-server/src/actors/processor.rs
```diff
@@ -1449,21 +1450,29 @@ impl EnvelopeProcessorService {
...
+                        "cardinality limiter failed"
+                    );
+
+                    Either::Left(buckets.into_iter())
+                }
+            };
         }
 
-        Ok(Box::new(buckets.into_iter()))
+        Either::<_, relay_cardinality::limiter::Accepted<_>>::Left(buckets.into_iter())
```
jjbayer, line 1476: Why does this need type annotations?

jjbayer (approved):

relay-server/src/actors/processor.rs
Dav1dde: For when we compile without `feature = processing`, then the `Right` side is never constructed -> error without annotations

relay-server/src/actors/processor.rs
Dav1dde: Exactly, but it doesn't have to be like that, `Left` and `Right` can contain [any iterator](https://docs.rs/itertools/0.12.0/itertools/enum.Either.html), it's just a helper to be able to return 2 distinct iterator types as one concrete type

PR #2853, opened 2023-12-14
feat(spans): Remove unused transaction tag from resource metrics

jjbayer:
We currently set the high cardinality `transaction` tag on all of the resource size metrics:

- `http.response_content_length` (encoded size)
- `http.decoded_response_content_length`
- `http.response_transfer_size`

However, we only show `http.response_content_length` split by transaction in the UI, so for now the tag can be removed on the other two metrics.

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -286,9 +286,6 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 Tag::with_key("span.op")
                     .from_field("span.sentry_tags.op")
                     .always(), // already guarded by condition on metric
-                Tag::with_key("transaction")
-                    .from_field("span.sentry_tags.transaction")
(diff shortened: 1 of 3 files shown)


relay-dynamic-config/src/defaults.rs
```diff
@@ -286,9 +286,6 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                 Tag::with_key("span.op")
                     .from_field("span.sentry_tags.op")
                     .always(), // already guarded by condition on metric
-                Tag::with_key("transaction")
-                    .from_field("span.sentry_tags.transaction")
-                    .always(), // already guarded by condition on metric
```
jjbayer, line 291: Press "expand" to see the context on these changes.

olksdr (approved):

DominikB2014 (approved):

PR #2854, opened 2023-12-15
feat(quotas): Add global throughput limit

TBS1996:
Adds a mechanism to limit the throughput of metric buckets. 

It's done globally, meaning across different relays, using redis.

In order to not call redis every time, we use a budget system (the name quota already taken).  We "take" a certain budget by incrementing a redis counter for the given global quota in the given slot, put it in a local counter, and count down to zero before asking for more from redis. 
  
 
The setup should be very performant, we use an RWLock together with atomics, the atomics can be mutated with a shared reference, so the different threads won't have to wait to acquire a lock. We only need exclusive reference to the rwlock when we insert a new quota, which should be a rare event.   
 
 
I used Ordering::SeqCst to be on the safe side, looking for input on which we can downgrade!


relay-quotas/src/global_quota.lua
```diff
@@ -0,0 +1,54 @@
...
+local default_budget = 100
+
+-- The reason to override the default max budget sometimes is to ensure the ratelimiter can handle
+-- batches higher than `default_budget`.
+local max_budget = math.max(quantity, default_budget)
+
+local redis_count = tonumber(redis.call('GET', key) or 0)
+
+
+if redis_count > limit then
```
jjbayer: Should this be `>=`? If `redis_count == limit` the else branch does an INCRBY 0.

relay-quotas/src/global_quota.lua
```diff
@@ -0,0 +1,54 @@
...
+-- When the redis key/val should be deleted. Should be current_time() + window + grace.
+local expiry = tonumber(ARGV[2])
+-- The minimum amount of budget needed for a specific batch.
+local quantity = tonumber(ARGV[3])
+
+local default_budget = 100
+
+-- The reason to override the default max budget sometimes is to ensure the ratelimiter can handle
+-- batches higher than `default_budget`.
+local max_budget = math.max(quantity, default_budget)
```
jjbayer: IMO it should be the caller's responsibility to figure out what budget they want to take. That would simplify the Lua script and make it easier to make default_budget configurable.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +137,189 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+}
+
+/// Counters used as a cache for global quotas.
+///
+/// We might have different global counters for different conditions (e.g. different namespaces).
+/// We use an RwLock because the counter itself is an `AtomicUsize`. Atomics can be modified with
+/// a shared reference, meaning that we can have multiple threads updating the counter concurrently,
+/// as long as there's no writers.
+#[derive(Clone, Default)]
+pub struct GlobalCounters(Arc<RwLock<hashbrown::HashMap<BudgetKey, BudgetState>>>);
```
jjbayer: Does this need to be pub? Its members seem to be private.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +137,189 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+            .write()
+            .map_err(|_| GlobalRateLimitError::PoisonedLock)?;
+
+        // If there's still some budget left in the same slot we will add it to the new budget we took.
+        // This can happen if for example the budget is 30, then we try to decrement by 50, so we
+        // ask for 100 more in budget but the 30 budget is still available so the local count is 130.
+        let old_budget: usize = map
+            .get(&budget_key)
+            .filter(|val| val.slot == current_slot)
+            .map(|val| val.count.load(Ordering::SeqCst))
```
jjbayer: Why do we have both atomic counters _and_ a `RwLock`? Wouldn't the outer lock be enough to protect the data?

TBS1996: If we didn't have atomics within the lock, we would need write access to modify the budget, which would slow down multithreading a lot. For now, we only need write access on the lock when calling redis_sync. Now that i think of it, I could optimize it further by making `slot` an atomic, and only getting write access when we insert a new key.

relay-quotas/src/redis.rs
TBS1996: oh, good call!

relay-quotas/src/global_quota.lua
TBS1996: oh yeah that makes way more sense i think. changing it!

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,197 @@ impl std::ops::Deref for RedisQuota<'_> {
     }
 }
 
+/// Possible outcomes from attemping to decrement a global counter budget.
+#[derive(Debug)]
+enum DecOutcome {
```
TBS1996: Looking for a better name x)

relay-quotas/src/redis.rs
TBS1996: ^ implemented said optimization

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,300 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+struct BudgetKey {
+    prefix: String,
+    window: u64,
+}
+
+/// Used to look up a hashmap of [`BudgetKey`]-keys without a string allocation.
+///
+/// This works due to the 'Equivalent' trait in the hashbrown crate.
+#[derive(Clone, Copy, Hash)]
+struct BudgetKeyRef<'a> {
```
TBS1996: ~Is it overkill to have a separate type just to avoid allocations?~   
 
I followed this precedence: https://github.com/getsentry/relay/blob/82a0409ca2ab8abb685a21e3aaa6de2f1a792aaa/relay-cardinality/src/cache.rs#L242

relay-quotas/src/redis.rs
```diff
@@ -200,19 +511,36 @@ impl RedisRateLimiter {
...
-                invocation.key(key);
-                invocation.key(refund_key);
-
-                invocation.arg(quota.limit());
-                invocation.arg(quota.expiry().as_secs() + GRACE);
-                invocation.arg(quantity);
-                invocation.arg(over_accept_once);
-
-                tracked_quotas.push(quota);
+                if quota.scope == QuotaScope::Global {
```
TBS1996, line 228: the condition here is a bit ugly, as it's just incidental that the new global scope always uses the budget cache. We talked briefly before about if we could use the budget system for all quotas, but i have yet to explore that possibility.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,300 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+        quantity: usize,
+    ) -> Result<bool, GlobalRateLimitError> {
+        use GlobalRateLimitError as E;
+
+        match self.decrement_budget(quota, quantity) {
+            ok @ Ok(_) => ok,
+            err @ Err(E::Redis | E::PoisonedLock | E::LoopLimitExceeded) => err,
+            Err(E::BudgetEmpty | E::SlotExpired | E::KeyMissing) => {
+                self.redis_sync(client, quota, quantity)?;
+                self.decrement_budget(quota, quantity)
```
TBS1996: the slot could technically have expired between the two calls, but the chance of that is too small to bother handling imo

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,300 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+                    Ordering::SeqCst,
+                    Ordering::SeqCst,
+                )
+                .is_ok()
+            {
+                return Ok(false);
+            }
+        }
+
+        Err(GlobalRateLimitError::LoopLimitExceeded)
```
TBS1996: hopefully never happens, I was just nervous to have a potentially infinite loop in prod

relay-quotas/src/global_quota.lua
```diff
@@ -9,44 +9,31 @@
...
--- The reason to override the default max budget sometimes is to ensure the ratelimiter can handle
--- batches higher than `default_budget`.
-local max_budget = math.max(quantity, default_budget)
+-- The budget that the caller intends to take. Will be capped if too close to the limit.
+local requested_budget = tonumber(ARGV[3])
 
 local redis_count = tonumber(redis.call('GET', key) or 0)
 
+if redis_count == 0 then
+    redis.call('EXPIREAT', key, expiry)
```
jjbayer: AFAIK this has no effect, because the key does not exist.

relay-quotas/src/redis.rs
```diff
@@ -137,155 +139,228 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+    ) -> Result<bool, GlobalRateLimitError> {
+        use GlobalRateLimitError as E;
+
+        match self.is_rate_limited(quota, quantity) {
+            ok @ Ok(_) => ok,
+            err @ Err(E::Redis | E::LoopLimitExceeded) => err,
+            Err(E::BudgetEmpty | E::SlotExpired | E::KeyMissing) => {
+                self.redis_sync(client, quota, quantity)?;
+                self.is_rate_limited(quota, quantity)
+            }
```
jjbayer: These are not really error cases, so I would prefer if `is_rate_limited` returned a custom enum instead of a `Result`.

relay-quotas/src/redis.rs
```diff
@@ -137,155 +139,228 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+        quantity: usize,
+    ) -> Result<bool, GlobalRateLimitError> {
+        use GlobalRateLimitError as E;
+
+        match self.is_rate_limited(quota, quantity) {
+            ok @ Ok(_) => ok,
+            err @ Err(E::Redis | E::LoopLimitExceeded) => err,
+            Err(E::BudgetEmpty | E::SlotExpired | E::KeyMissing) => {
+                self.redis_sync(client, quota, quantity)?;
+                self.is_rate_limited(quota, quantity)
```
jjbayer: There's an implicit infinite loop here.

relay-quotas/src/redis.rs
```diff
@@ -137,155 +139,228 @@ impl std::ops::Deref for RedisQuota<'_> {
...
 #[derive(Debug, thiserror::Error)]
 enum GlobalRateLimitError {
     #[error("quota key missing")]
     KeyMissing,
-    #[error("failed to access RWLock")]
-    PoisonedLock,
     #[error("slot expired")]
     SlotExpired,
     #[error("budget empty")]
     BudgetEmpty,
```
jjbayer:
```suggestion
    LocalBudgetExhausted,
```

relay-quotas/src/redis.rs
```diff
@@ -137,155 +139,228 @@ impl std::ops::Deref for RedisQuota<'_> {
...
-            // Expired slot implies the the time window has passed, and we need to
-            // ask for a new budget for the new slot.
-            return DecOutcome::SlotExpired;
-        }
+        let val = map
+            .get(&BudgetKeyRef::new(quota))
+            .ok_or(GlobalRateLimitError::KeyMissing)?;
+
+        let max_loop_iterations = 10;
+        for _ in 0..max_loop_iterations {
```
jjbayer: Is this loop here because of `compare_exchange_weak`? I wonder if this is really more performant than using a lock which we acquire to read and write, which would be the simpler implementation.

relay-quotas/src/redis.rs
```diff
@@ -137,155 +139,228 @@ impl std::ops::Deref for RedisQuota<'_> {
...
-        let mut current_budget = val.count.load(Ordering::SeqCst);
+            let redis_count = val.last_seen_redis_value.load(Ordering::SeqCst);
+            let current_budget = val.budget.load(Ordering::SeqCst);
 
-        // Since the redis count was pre-incremented by the amount of budget we took, the actual
-        // total count have to subtract the current budget.
-        let total_count = redis_count.saturating_sub(current_budget as u64);
+            // Since the redis count was pre-incremented by the amount of budget we took, the actual
+            // total count have to subtract the current budget.
+            let total_count = redis_count.saturating_sub(current_budget);
```
jjbayer: I think there's also a race condition here: Because `val.budget` and `val.last_seen_redis_value` are not read together, their values could go out of sync.

relay-quotas/src/redis.rs
```diff
@@ -137,155 +139,228 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+        let val = map
+            .get(&BudgetKeyRef::new(quota))
+            .ok_or(GlobalRateLimitError::KeyMissing)?;
+
+        let max_loop_iterations = 10;
+        for _ in 0..max_loop_iterations {
+            if current_slot != val.slot.load(Ordering::SeqCst) {
+                // Expired slot implies the the time window has passed, and we need to
+                // ask for a new budget for the new slot.
+                return Err(GlobalRateLimitError::SlotExpired);
```
jjbayer: I don't understand this:

1. Either we can only request a specific slot. In that case, there should be no retry after the slot expires.
2. Or we can request a slot even after its redis key expired. In this case, the Lua script should just recreate the key.

relay-quotas/src/redis.rs
TBS1996: fixed now by making redis_value and slot normal usize. this means they can't be mutated when we have the read lock.

relay-quotas/src/redis.rs
TBS1996: im trying to look for more examples, but I found this: https://speakerdeck.com/komamitsu/performance-comparison-of-mutex-rwlock-and-atomic-types-in-rust?slide=12

relay-quotas/src/redis.rs
TBS1996: what do you mean? it just calls it twice

relay-quotas/src/redis.rs
TBS1996: that was the previous implementation and I wasn't happy with it.  The thing is that many many of the outcomes are fine the first time you call the function, but if it's not resolved after the redis_sync, it's more of a hard error. So I ended up creating a custom enum and an error enum with 90% overlap.  
  
Personally I think it's fine to see them as errors, we attempt to learn if it's ratelimited by getting a true/false value, and if some condition stops us from getting that value, you can say it's an error case

relay-quotas/src/global_quota.lua
TBS1996: oh good point, thanks!

relay-quotas/src/redis.rs
jjbayer: It's recursive, right? So this call to `is_rate_limited` might end up calling `is_rate_limited` again, etc.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+    #[error("slot expired")]
+    SlotExpired,
+    #[error("budget empty")]
+    BudgetEmpty,
+    #[error("failed to communicate with redis")]
+    Redis,
+}
+
+fn current_slot(window: u64) -> u64 {
+    UnixTimestamp::now().as_secs() / window
```
jjbayer: Can `window` be zero, e.g. by misconfiguration? Should we use `checked_sub` to prevent a panic?

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+    ) -> Result<bool, GlobalRateLimitError> {
+        let Some(limit) = quota.limit.map(|limit| limit as usize) else {
+            return Ok(false);
+        };
+
+        let current_slot = current_slot(quota.window) as usize;
+        let map = self.counters.read().unwrap_or_else(|e| e.into_inner());
+        let val = map
+            .get(&BudgetKeyRef::new(quota))
+            .ok_or(GlobalRateLimitError::KeyMissing)?
```
jjbayer: If we wrap `BudgetState` in an `Arc` + `RwLock` / `Mutex`, we could acquire a write lock here, do `get_or_insert`, and release the lock immediately. That would make the `KeyMissing` state unnecessary.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+            .read()
+            .unwrap_or_else(|e| e.into_inner());
+
+        if current_slot != val.slot {
+            return Err(GlobalRateLimitError::SlotExpired);
```
jjbayer: Similar to above, could we remove the old `val` and `insert` a new one (protected by the write lock mentioned above)?

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+/// When we want to ratelimit across all relay-instances, we need to use redis to synchronize.
+/// Calling Redis every time we want to check if an item should be ratelimited would be very expensive,
+/// which is why we have this cache. It works by 'taking' a certain budget from redis, by pre-incrementing
+/// a global counter. We Put the amount we pre-incremented into this local cache and count down until
+/// we have no more budget, then we ask for more from redis. If we find the global counter is above
+/// the quota limit, we will ratelimit the item.
+#[derive(Clone)]
+struct GlobalCounters {
+    script: Arc<Script>,
+    counters: Arc<RwLock<hashbrown::HashMap<BudgetKey, RwLock<BudgetState>>>>,
```
jjbayer: `counters` does not need to be `Arc`. I tried removing `derive(Clone)` from `RedisRateLimiter` and the compiler did not complain.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+            }
+        } else {
+            debug_struct.field("counters", &"Lock Poisoned");
+        }
+
+        debug_struct.finish()
+    }
+}
+*/
+
```
jjbayer:
```suggestion
```

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+            .read()
+            .unwrap_or_else(|e| e.into_inner());
+
+        if current_slot != val.slot {
+            return Err(GlobalRateLimitError::SlotExpired);
+        }
+
+        let redis_count = val.last_seen_redis_value;
+
+        loop {
```
jjbayer: I would prefer an implementation without atomics (see other comments), but if we're going there, you can use [`fetch_update`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#considerations), which wraps `compare_exchange_weak` internally.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+                    Ordering::SeqCst,
+                )
+                .is_ok()
+            {
+                return Ok(false);
+            }
+        }
+    }
+
+    fn call_redis(
```
jjbayer: nit: Move this function below `redis_sync` to keep the order of high level -> low level functions.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+        loop {
+            let current_budget = val.budget.load(Ordering::SeqCst);
+            let total_count = redis_count.saturating_sub(current_budget);
+
+            if total_count + quantity > limit {
+                return Ok(true);
+            }
+
+            if current_budget < quantity {
+                return Err(GlobalRateLimitError::BudgetEmpty);
```
jjbayer: Proposal for simplification:

1. Get a write lock instead of a read lock in this function.
2. No need for atomics or a retry loop because we then have exclusive access to `val`.
3. On this line, instead of returning an error, go to redis to get a fresh budget. That makes the `BudgetEmpty` state unnecessary.

The lock will be released quickly, except when we need to sync with redis. But in that case, we're already holding an exclusive lock that locks other threads out, so the contention on the lock should be comparable.

This would also simplify `redis_sync`.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+        &self,
+        client: &mut PooledClient,
+        quota: &RedisQuota,
+        quantity: usize,
+    ) -> Result<(), GlobalRateLimitError> {
+        let shared_counter_map = self.counters.read().unwrap_or_else(|e| e.into_inner());
+        let current_slot = current_slot(quota.window) as usize;
+        let budget_key = BudgetKeyRef::new(quota);
+
+        match shared_counter_map.get(&budget_key) {
```
jjbayer: nit: You could match on `shared_counter_map.entry()` instead.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,291 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+                        self.counters.write().unwrap_or_else(|e| e.into_inner());
+
+                    let (new_budget, redis_value) = self.call_redis(quota, client, quantity)?;
+
+                    exclusive_map.insert(
+                        budget_key.into_owned(),
+                        RwLock::new(BudgetState::new(new_budget, redis_value, current_slot)),
+                    );
+                } else {
+                    drop(self.counters.write().unwrap_or_else(|e| e.into_inner()));
```
jjbayer: What does this do? It acquires a write lock which it immediately releases?

relay-quotas/src/redis.rs
TBS1996: so im exploring the idea of getting rid of the atomics (although personally i like them),  but regarding the fetch_update, I already tried it and I didn't like it. Mainly because it takes in a closure which just return an option, even though it can fail for two reasons, so we have to check the condition after.  
 
look: https://gist.github.com/TBS1996/a4deed8bb3794be9b5966e922b85137e

relay-quotas/src/redis.rs
TBS1996: neat! assumed it had to be arc but it makes sense we don't need since innerprocessor is arced

relay-quotas/src/redis.rs
TBS1996: so yes, this would indeed make it more simple, although with a decent performance loss. Looking forward to the irl discussion on whether it's worth it or not. 
 
About removing the BudgetEmpty and the KeyMissing (as mentioned in other place), i dont see the point of that, they're essentially just there for readability as keymissing/budgetempty/slotexceeded are all fixed by syncing to redis. so it's not really 'state' in practice

relay-quotas/src/redis.rs
TBS1996: the idea was that if another thread is currently writing, it means they are calling redis, so it would be better to return when that thread is done with calling redis. if you call .write() it'd block until that thread is finished updating the redis results. 
  
now i have a atomicbool flag instead though

relay-quotas/src/redis.rs
TBS1996: we could, but would that really simplify it? right now the structure is kinda simple in that this function doesn't call any other functions

relay-quotas/src/redis.rs
TBS1996: oh nice catch, didnt think of that!

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+
+**Features**:
+
+- Add a global throughput rate limiter for metric buckets. ([#2854](https://github.com/getsentry/relay/pull/2854))
+
+## 23.12.1
```
Dav1dde, line 10: This is the wrong spot

relay-quotas/Cargo.toml
```diff
@@ -14,6 +14,7 @@ default = []
 redis = ["dep:thiserror", "dep:relay-log", "relay-redis/impl"]
 
 [dependencies]
+hashbrown = {workspace = true}
```
Dav1dde, line 17: Formatting

relay-quotas/src/global_quota.lua
```diff
@@ -0,0 +1,40 @@
...
+--
+-- Made to work with a local cache of quota limit. The caller will "take" a certain budget
+-- from the global redis counter, given that the limit have not been exceeded. This is to dramatically
+-- reduce the amount of redis calls while at the same time not overshooting the quota.
+-- We return both the size of the taken budget and the size of the redis count.
+-- The reason we return the size of the redis count is to avoid asking for more budget when
+-- the previously seen redis count is still bigger than the limit.
+--
+-- The redis keys are unique to their timeslot, which is why we let them expire in order to not
+-- fill up redis with dead keys.
```
Dav1dde, line 11: Would be nice to have arguments and possible return values listed here

relay-quotas/src/redis.rs
```diff
@@ -16,6 +17,8 @@ use crate::REJECT_ALL_SECS;
 /// metrics may not be in sync with the computer running this code.
 const GRACE: u64 = 60;
 
+const DEFAULT_BUDGET_REQUEST: usize = 100;
```
Dav1dde: Need to make this configurable and probably percentage based.

E.g. we would want to configure that Relays take 0.1% of the Quota Limit to cache:

```
Quota: 1M
Default Request: 0.1% (number to be defined)
-> Cached: 10k
```

We also need to come up with some numbers that make sense for us.

relay-quotas/src/redis.rs
```diff
@@ -797,4 +1082,58 @@ mod tests {
...
+        };
+
+        let inner_scoping = Scoping {
+            organization_id: 69420,
+            project_id: ProjectId::new(42),
+            project_key: ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap(),
+            key_id: Some(4711),
+        };
+
+        let static_inner_scoping: &'static Scoping = Box::leak(Box::new(inner_scoping));
```
Dav1dde, line 861: Even in tests we shouldn't resort to leaking memory. If you split quota creation and `RedisQuota` creation in 2 functions this is unnecessary.

relay-quotas/src/redis.rs
Dav1dde: I'd just use an `if let Some(..) = .. { .. } else { .. }` here

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,258 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+    fn fetch_block_redis(&self) -> bool {
+        self.redis_call_in_progress.swap(true, Ordering::SeqCst)
+    }
+
+    fn unblock_redis(&self) {
+        self.redis_call_in_progress.store(false, Ordering::SeqCst);
+    }
+
+    fn update(&mut self, new_budget: usize, redis_value: usize, current_slot: usize) {
+        let is_slot_expired = self.slot != current_slot;
```
Dav1dde: Need to be careful here, the `current_slot` can be older than `self.slot`!! We probably should reject anything which has an older slot than `this.slot`.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,258 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+        }
+    }
+}
+
+impl GlobalCounters {
+    /// Returns `true` if the global quota should be ratelimited.
+    ///
+    /// Certain errors can be resolved by syncing to redis, so in those cases
+    /// we try again to decrement the budget after syncing.
+    fn evaluate_quota(
```
Dav1dde: nit: the Method returns a boolean, the name of the function should be worded that the return value can be understood without reading the docs, something like `is_rate_limited()`

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,258 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+                BO::NotRateLimited => return Ok(false),
+                BO::KeyMissing | BO::SlotExpired | BO::LocalBudgetExhausted => {
+                    self.redis_sync(client, quota, quantity)?;
+                }
+            }
+        }
+    }
+
+    /// Returns `true` if items should be ratelimited.
+    fn decrement_budget(&self, quota: &RedisQuota, quantity: usize, limit: usize) -> BudgetOutcome {
```
Dav1dde: nit: I'd call this `try_use_budget`

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,258 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+        let map = self.counters.read().unwrap_or_else(|e| e.into_inner());
+        let val = match map.get(&BudgetKeyRef::new(quota)) {
+            Some(inner_lock) => inner_lock.read().unwrap_or_else(|e| e.into_inner()),
+            None => return BudgetOutcome::KeyMissing,
+        };
+
+        if current_slot(quota.window) != val.slot {
```
Dav1dde: As noted elsewhere it is possible that the `current_slot(quota.window)` is older than `val.slot`

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,258 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+            if budget < quantity {
+                return BudgetOutcome::LocalBudgetExhausted;
+            }
+
+            match val
+                .budget
+                .compare_exchange_weak(budget, budget - quantity, SeqCst, SeqCst)
+            {
+                Ok(_) => return BudgetOutcome::NotRateLimited,
+                Err(_) => continue,
```
Dav1dde: This loads the `budget` for you already, you continue then to just load the budget again.

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,258 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+            None => return BudgetOutcome::KeyMissing,
+        };
+
+        if current_slot(quota.window) != val.slot {
+            return BudgetOutcome::SlotExpired;
+        }
+
+        use Ordering::*;
+        loop {
+            let budget = val.budget.load(SeqCst);
```
Dav1dde: I believe this can be `Relaxed`

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,258 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+                return BudgetOutcome::RateLimited;
+            }
+
+            if budget < quantity {
+                return BudgetOutcome::LocalBudgetExhausted;
+            }
+
+            match val
+                .budget
+                .compare_exchange_weak(budget, budget - quantity, SeqCst, SeqCst)
```
Dav1dde: I believe this can be `SeqCst` for success and `Relaxed` for failure

relay-quotas/src/redis.rs
```diff
@@ -125,6 +139,258 @@ impl std::ops::Deref for RedisQuota<'_> {
...
+    ) -> Result<(), RedisError> {
+        let shared_counter_map = self.counters.read().unwrap_or_else(|e| e.into_inner());
+        let current_slot = current_slot(quota.window);
+        let budget_key = BudgetKeyRef::new(quota);
+
+        match shared_counter_map.get(&budget_key) {
+            Some(val) => {
+                let val_read = val.read().unwrap_or_else(|e| e.into_inner());
+
+                if val_read.fetch_block_redis() {
```
Dav1dde: As discussed this makes the code spin.

Also the problem with unblocking which is easy to get wrong/miss or fail (during panics).

relay-quotas/src/redis.rs
TBS1996: yup. in a follow-up i guess?

PR #2856, opened 2023-12-15
fix(spans): Scrub integer file extensions

jjbayer:
We've seen resource span descriptions like `http://domain.com/*.123` which cause high cardinality in the extension.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -293,13 +293,7 @@ fn scrub_resource_filename<'a>(ty: &str, path: &'a str) -> Cow<'a, str> {
         extension = "";
     }
 
-    // Only accept short, clean file extensions.
-    if let Some(invalid) = extension.bytes().position(|c| !c.is_ascii_alphanumeric()) {
-        extension = &extension[..invalid];
-    }
-    if extension.len() > MAX_EXTENSION_LENGTH {
-        extension = "";
-    }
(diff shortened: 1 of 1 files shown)


DominikB2014 (approved):

Dav1dde (approved):
Maybe we should just have a whitelist of extensions instead?

jjbayer:
> Maybe we should just have a whitelist of extensions instead?

@Dav1dde That's a good point. @DominikB2014 would the list of supported images be short enough to enumerate?

DominikB2014:
> > Maybe we should just have a whitelist of extensions instead?
> 
> 
> 
> @Dav1dde That's a good point. @DominikB2014 would the list of supported images be short enough to enumerate?

We do have a list of extensions that we look for in the FE

https://github.com/getsentry/sentry/blob/master/static/app/views/performance/browser/resources/shared/constants.ts

This mostly lines up with

https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types

But I'm not sure if this includes all types tho. Are we able to query for all extensions that we have seen to be sure?

jjbayer:
> But I'm not sure if this includes all types tho. Are we able to query for all extensions that we have seen to be sure?

@DominikB2014 I ran a query and posted the results [here](https://www.notion.so/sentry/Allowlist-for-file-extensions-65ce495452cc42f3be69062c7b6a47fa?pvs=4#df3c6dc55f8f478e830e631ca936738e), let me know which of those you would like to add to the allowlist!

PR #2862, opened 2023-12-18
ref(processor): Introduce processing groups pipelines

olksdr:
This change makes another step forward to typed processing pipeline:
* split the envelope when possible per item type
  * if transaction, all the item kept together in the same envelope
  * trying to keep event and attachment also in the same envelope when possible
* the rest of the items are move into separate envelopes (one item per one envelope)

#### Open question:
* ~should we create new event id for each newly separated envelope (or item type)?~
* is it fine to split incoming envelope upon the receiving one, mainly in `queue_envelope` function? 

Related to https://github.com/getsentry/team-ingest/issues/253

#skip-changelog


relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
+            relay_log::error!("There are not type and no items in the envelope");
+            return Err(ProcessingError::NoEventPayload);
+        };
 
-            if_processing!({
-                unreal::expand(state, &self.inner.config)?;
-            });
+        // Defined processing pipeilnes based on the event type and/or item type they contain.
+        match ty {
+            // This stil can contain attachements.
```
```suggestion
            // This can still contain attachements.
```

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
+
+                dynamic_sampling::sample_envelope(state)?;
+
+                if_processing!({
+                    event::store(state, &self.inner.config, self.inner.geoip_lookup.as_ref())?;
+                    self.enforce_quotas(state)?;
+                    profile::process(state, &self.inner.config);
+                });
+
+                if state.has_event() {
```
Dav1dde: Isn't this always true for transactions?

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
+            .event_type()
+            // get the item type from the event type if there is an event
+            .map(ItemType::from_event_type)
+            // othrwise get the first item from the items list and assume it's the only one defined
+            // which we are waiting for.
+            .or_else(|| {
+                state
+                    .envelope()
+                    .items()
+                    .next()
```
Dav1dde: I assume the new code in the endpoint makes sure the envelope is already correctly split into "single" items?

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
+                event::finalize(state, &self.inner.config)?;
+                self.light_normalize_event(state)?;
+                event::filter(state)?;
+
+                if_processing!({
+                    event::store(state, &self.inner.config, self.inner.geoip_lookup.as_ref())?;
+                    self.enforce_quotas(state)?;
+                });
+
+                if state.has_event() {
```
Dav1dde: This is always false?

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
```
Dav1dde: This is always false?

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
```
Dav1dde: This is always true?

relay-server/src/endpoints/common.rs
```diff
@@ -286,44 +286,19 @@ fn queue_envelope(
...
     }
 
-    // Take all NEL reports and split them up into the separate envelopes with 1 item per
-    // envelope.
-    for nel_envelope in envelope.split_all_by(|item| matches!(item.ty(), ItemType::Nel)) {
-        relay_log::trace!("queueing separate envelopes for NEL report");
-        let buffer_guard = state.buffer_guard();
-        let nel_envelope = buffer_guard
+    // Split off the envelopes by item type.
+    let envelopes = managed_envelope.take_envelope().into_processible_parts();
```
Dav1dde: I wonder if we should stop modifying the envelope here and instead do that in the processor/`process_state`?

relay-server/src/envelope.rs
```diff
@@ -1188,31 +1188,51 @@ impl Envelope {
...
-            // Each item should get an envelope with the new event id.
-            headers.event_id = Some(EventId::new());
+    /// Some of the items still kept together, since they must be processed as one unit.
+    pub fn into_processible_parts(mut self) -> SmallVec<[Box<Self>; 3]> {
+        let headers = self.headers.clone();
+        let mut items = std::mem::take(&mut self.items);
+
+        // If we have a transaction item , we assume that it's a transaction, and all the related
+        // items are connected to it and a subject to dynamic sampling and related operations.
+        if items.iter().any(|item| item.ty() == &ItemType::Transaction) {
```
Dav1dde: Can this actually not be mixed, e.g. transaction + session etc.?

relay-server/src/envelope.rs
```diff
@@ -1188,31 +1188,51 @@ impl Envelope {
...
+            .collect::<Vec<_>>();
+
+        // Get the rest of the envelopes, one per item.
+        //
+        // TODO(olek): New event id per envelope?
+        let mut envelopes: SmallVec<[Box<Self>; 3]> = items
+            .into_iter()
+            .map(|item| {
+                let mut headers = headers.clone();
+                // NEL item type should always we recieving a new event id.
```
Dav1dde: > NEL item type should always we recieving a new event id.


Was that the case before as well?

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
-            self.light_normalize_event(state)?;
-            dynamic_sampling::normalize(state);
-            event::filter(state)?;
-            dynamic_sampling::run(state, &self.inner.config);
+                if_processing!({
+                    event::store(state, &self.inner.config, self.inner.geoip_lookup.as_ref())?;
+                    self.enforce_quotas(state)?;
+                });
+
+                if state.has_event() {
```
Dav1dde: Always true?

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
+
+                attachment::scrub(state);
+            }
+            ItemType::Session | ItemType::Sessions => {
+                session::process(state, &self.inner.config);
+                if_processing!({
+                    self.enforce_quotas(state)?;
+                });
+
+                if state.has_event() {
```
Dav1dde: Always false?

Dav1dde (commented):
Just some comments, with a bunch of questions if `state.has_event` doesn't always evalute to true or false. I mostly validated this against `EventType`.

Generally this is very hard for me to figure out since I have no idea when:

* a item type *must* have an associated event (also this would then always be processed as event?)
* an item type *can* have an associated event
* an item type *never* has an associated event

Do we have documentation about this somewhere?

Overall I really like the direction this is taking!

relay-server/src/actors/processor.rs
olksdr: In most cases it will be 1 item, yes, but not always.

relay-server/src/actors/processor.rs
olksdr: Sooo, it's tricky, the `enforce_quotas`  can actually remove the event from the processing state 😄  and this condition will be false. And same goes for every question you asked where we enforce quotas.

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
+                }
 
-            // We avoid extracting metrics if we are not sampling the event while in non-processing
-            // relays, in order to synchronize rate limits on indexed and processed transactions.
-            if self.inner.config.processing_enabled() || state.sampling_result.should_drop() {
-                self.extract_metrics(state)?;
+                attachment::scrub(state);
             }
+            // Contains data which belongs together with transactions, e.g. profiles attachments.
+            ItemType::Transaction | ItemType::Profile => {
```
jjbayer: Profiles are currently only allowed if they're in the same envelope as a transaction, in which case `ty` is always `Transaction`, right?
```suggestion
            ItemType::Transaction => {
```

relay-server/src/actors/processor.rs
```diff
@@ -936,68 +936,295 @@ impl EnvelopeProcessorService {
...
+            ItemType::Session | ItemType::Sessions => {
+                session::process(state, &self.inner.config);
+                if_processing!({
+                    self.enforce_quotas(state)?;
+                });
+
+                if state.has_event() {
+                    event::scrub(state)?;
+                    event::serialize(state)?;
+                }
```
```suggestion
```

Like @Dav1dde said.

relay-server/src/actors/processor/report.rs
```diff
@@ -469,26 +469,15 @@ mod tests {
             item
         });
 
-        envelope.add_item({
-            let mut item = Item::new(ItemType::Event);
-            item.set_payload(ContentType::Json, "{}");
-            item
-        });
```
jjbayer: Are we explicitly disallowing envelopes that contain an event and a user report now?

jjbayer (commented):
> should we create new event id for each newly separated envelope (or item type)?

@olksdr no, I think we should either copy the event ID (when an item relates to an event), or not set an event ID at all.

relay-server/src/actors/processor.rs
Dav1dde: Thanks!

relay-server/src/endpoints/common.rs
olksdr: I could do that too, I will have to add the `BufferGuard` to processor. Otherwise it wouldn't be that difficult

relay-server/src/envelope.rs
olksdr: Added more precise filtering

relay-server/src/envelope.rs
olksdr: yeah, nel comes from the browser and in one batch it can be many different events, which we had to separate.

relay-server/src/actors/processor.rs
olksdr: Yeah, the profile is in there, since I want to cover all the branches in match for `ItemType`

relay-server/src/actors/processor.rs
olksdr: @jjbayer  the sessions will never have an event ?

relay-server/src/actors/processor/report.rs
olksdr: Right now the processor will always get those items in the separate envelopes. Since we split the envelope in the endpoint. I can also move the split back into processor, and this test will still look as it did before.

relay-server/src/actors/processor.rs
jjbayer: Every item that does not `require_event` is split into an event-less envelope in `queue_envelope` (see linked code), so yes I think that is safe to assume.

https://github.com/getsentry/relay/blob/e655ae18be1c4aa20742bc687a7e768b92ad2b9e/relay-server/src/endpoints/common.rs#L306-L322

relay-server/src/actors/processor.rs
```diff
@@ -950,68 +992,251 @@ impl EnvelopeProcessorService {
...
+
+                attachment::scrub(state);
```
jjbayer: nit: Should we put every one of these match arms into separate functions?

relay-server/src/actors/processor.rs
```diff
@@ -941,6 +941,48 @@ impl EnvelopeProcessorService {
         Ok(())
     }
 
+    /// Returns the _main_ item type from the envelope in the current state.
+    ///
+    /// This functions mimics [`event::extract`] function and returns the [`ItemType`] with
+    /// the same priority.
+    ///
+    /// Note: This function [`event::extract`] should be refactored and unified.
+    fn event_item_type(state: &ProcessEnvelopeState) -> Option<ItemType> {
```
jjbayer: Should we rename this to `main_item_type()`?

tests/integration/test_healthchecks.py
```diff
@@ -152,6 +133,7 @@ def store_internal_error_event():
         try:
             # These events will consume all the disk sapce and we will report not ready.
             relay.send_event(project_key)
+            relay.send_event(project_key)
```
jjbayer: Why did this test change?

olksdr: To unflaky it , but it was already done in the separate PR, after merging `master` this change will disappear.

relay-server/src/actors/processor.rs
olksdr: It's a good question. 
I personally wouldn't do that, because it would require more changes, like refactoring `if_processing!` macro somehow to make sure it works outside of the `process_state` function. 
But I will have a look not, how can I do it better.

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +962,333 @@ impl EnvelopeProcessorService {
...
 
-            if_processing!({
-                unreal::expand(state, &self.inner.config)?;
+        if_processing!(self.inner.config, {
+            event::store(state, &self.inner.config, self.inner.geoip_lookup.as_ref())?;
+            self.enforce_quotas(state)?;
+            profile::process(state, &self.inner.config);
+        });
+
+        if state.has_event() {
```
jan-auer, line 1187: This should be trivially true in `process_transaction`. Dynamic sampling can drop the event by returning an error. The same is true for all other process functions related to events.

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +962,333 @@ impl EnvelopeProcessorService {
...
+            event::serialize(state)?;
+            if_processing!(self.inner.config, {
+                span::extract_from_event(state);
             });
+        }
 
+        attachment::scrub(state);
+        Ok(())
+    }
+    fn process_attachments(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
```
jan-auer:
```suggestion
    }
    
    /// <- please add some doc comment describing under which condition this should be called
    fn process_attachments(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
```

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +962,333 @@ impl EnvelopeProcessorService {
...
-            // We avoid extracting metrics if we are not sampling the event while in non-processing
-            // relays, in order to synchronize rate limits on indexed and processed transactions.
-            if self.inner.config.processing_enabled() || state.sampling_result.should_drop() {
-                self.extract_metrics(state)?;
-            }
 
-            dynamic_sampling::sample_envelope(state)?;
-
-            if_processing!({
+            if_processing!(self.inner.config, {
```
jan-auer: Shouldn't this event technically also go through store normalization?

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +962,333 @@ impl EnvelopeProcessorService {
...
 
-        if_processing!({
+        if_processing!(self.inner.config, {
             self.enforce_quotas(state)?;
-            profile::process(state, &self.inner.config);
-            self.process_check_ins(state);
-            span::process(state, self.inner.config.clone());
         });
 
         if state.has_event() {
```
jan-auer: Can you double-check the conditions for this? There should be an invariant that `state.has_event() === state.creates_event()` earlier. To be precise, when the state creates an event, there should be a placeholder created, which must be serialized unconditionally. If this is not the case, we likely have an implementation bug in `creates_event()`.

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +962,333 @@ impl EnvelopeProcessorService {
...
+            event::serialize(state)?;
+        }
+        Ok(())
+    }
+
+    fn process_user_reports(
+        &self,
+        state: &mut ProcessEnvelopeState,
+    ) -> Result<(), ProcessingError> {
+        report::process(
```
jan-auer, line 1227: Can we move processing after quotas? 

This is not possible for regular events (errors, transactions, security, nel, ...) since inbound filters must run before quotas and require processed payloads. However, in this case the quotas do not depend on a processed payload, so we can save some work when the project is out of quota.

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +962,333 @@ impl EnvelopeProcessorService {
...
+    fn process_checkins(&self, _state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
+        if_processing!(self.inner.config, {
+            self.enforce_quotas(_state)?;
+            self.process_check_ins(_state);
+        });
+        Ok(())
+    }
+
+    fn process_spans(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
+        span::filter(state);
```
jan-auer, line 1256: Are you sure the spans filter does not require processing (i.e. normalization) to run first?

relay-server/src/endpoints/common.rs
jan-auer: Any changes we make to this have to consider the rate-limit fast path in the project cache. Since we're applying rate limits there, either we keep splitting envelopes first or we need to ensure that the rate limiting code in the project cache is aware of the dependencies between items.

relay-server/src/envelope.rs
```diff
@@ -1190,31 +1192,53 @@ impl Envelope {
...
-        };
-
-        let headers = &mut self.headers;
-
-        for item in split_items {
-            // Each item should get an envelope with the new event id.
-            headers.event_id = Some(EventId::new());
+    /// The items which require events are always kept together in one envelope, the reest items
+    /// will be split into separate envelope with one item type each.
+    pub fn into_processible_parts(mut self) -> SmallVec<[Box<Self>; 3]> {
```
jan-auer: This function contains queueing / processing business logic that exceeds what should be implemented on a data class like the Envelope. Let's move this function either to queueing (_endpoints/common.rs_) or to the processor.

relay-server/src/actors/processor.rs
olksdr: In the current implementation (`master`) the span filter runs before any normalization. Maybe @jjbayer  can chime in on this?

relay-server/src/actors/processor.rs
olksdr: Currently `enforce_quotas` can remove the event from the state, that's why this condition still can be `false` or `true`. We can address this in  the followup.

relay-server/src/endpoints/common.rs
```diff
@@ -286,44 +286,19 @@ fn queue_envelope(
...
     }
 
-    // Take all NEL reports and split them up into the separate envelopes with 1 item per
-    // envelope.
-    for nel_envelope in envelope.split_all_by(|item| matches!(item.ty(), ItemType::Nel)) {
-        relay_log::trace!("queueing separate envelopes for NEL report");
-        let buffer_guard = state.buffer_guard();
-        let nel_envelope = buffer_guard
+    // Split off the envelopes by item type.
+    let envelopes = ProcessingGroup::split_envelope(&mut managed_envelope.take_envelope());
```
jjbayer: Should `split_envelope` take ownership of the envelope (i.e. pass by value)?

relay-server/src/endpoints/common.rs
```diff
@@ -286,44 +286,19 @@ fn queue_envelope(
...
+    for (_, envelope) in envelopes {
```
jjbayer: Instead of providing a separate `from_envelope` method that analyses the envelope again, should we just pass the `ProcessingGroup` on to the next service here? For example as an attribute on the `ManagedEnvelope`.

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +1127,183 @@ impl EnvelopeProcessorService {
...
+        if_processing!(self.inner.config, {
+            self.enforce_quotas(state)?;
+            span::process(state, self.inner.config.clone());
+        });
+        Ok(())
+    }
+
+    fn process_state(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
+        let group = ProcessingGroup::from_envelope(state.envelope());
```
jjbayer: See comment in `common.rs`: Ideally the group would already be set on the envelope here.

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +1127,183 @@ impl EnvelopeProcessorService {
...
+            ProcessingGroup::StandaloneAttachment => self.process_attachments(state)?,
+            ProcessingGroup::UserReport => self.process_user_reports(state)?,
+            ProcessingGroup::Replay => self.process_replays(state)?,
+            ProcessingGroup::CheckIn => self.process_checkins(state)?,
+            ProcessingGroup::Span => self.process_spans(state)?,
+            ProcessingGroup::Unknown => {
+                relay_log::error!(
+                    tags.project = %state.project_id,
+                    "Could not identify the processing group based on the envelope's items."
+                );
```
jjbayer: Should we add the first item's type as a tag here? And/or attach a list of item types in the error's data. Will be interesting to see what kind of item is in the envelope.

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +1127,183 @@ impl EnvelopeProcessorService {
...
+        match group {
+            ProcessingGroup::Error => self.process_errors(state)?,
+            ProcessingGroup::Transaction => self.process_transactions(state)?,
+            ProcessingGroup::Session => self.process_sessions(state)?,
+            ProcessingGroup::StandaloneAttachment => self.process_attachments(state)?,
+            ProcessingGroup::UserReport => self.process_user_reports(state)?,
+            ProcessingGroup::Replay => self.process_replays(state)?,
+            ProcessingGroup::CheckIn => self.process_checkins(state)?,
+            ProcessingGroup::Span => self.process_spans(state)?,
+            ProcessingGroup::Unknown => {
```
jjbayer: I think we should play it safe and call the old implementation of `process_state` here (+ log the error).

We might have missed some edge cases where we rely on items being dropped (e.g. by `filter` functions), so calling the old `process_state` here could provide a safety net. Once we see no more errors (except for `unknown` items), we can remove the old `process_state`.

relay-server/src/actors/processor.rs
```diff
@@ -75,9 +78,191 @@ mod span;
...
+            && !envelope.items().any(Item::creates_event)
+        {
+            return Self::StandaloneAttachment;
+        }
+
+        if envelope.items().any(|item| {
+            item.ty() == &ItemType::ReplayEvent || item.ty() == &ItemType::ReplayRecording
+        }) {
+            return Self::Replay;
+        }
```
jjbayer: I noticed that `required_event` returns `true for `ReplayEvent`. I think that's wrong, @jan-auer replay events do not arrive in the same envelope as transactions or error events, do they?

relay-server/src/actors/processor.rs
```diff
@@ -75,9 +78,191 @@ mod span;
...
+
+        if envelope.items().any(Item::requires_event) {
+            return Self::Error;
+        }
```
jjbayer: I noticed one edge case here:

When a profile comes in without a transaction, `from_envelope` will return `Self::error`. But `process_error` does not call `profile::filter`, which takes care of dropping the profile!

https://github.com/getsentry/relay/blob/635a7cf051741f214a0c7cffd05f9593f1b65924/relay-server/src/actors/processor/profile.rs#L26-L27

I.e. we should return `ProcessingGroup::Transaction` for `ItemType::Profile`. There might be other cases like this, so maybe we should explicitly enumerate the item types that are allowed in each group, and send other items to the `Unknown` group. Together with the "safety net" I suggested below, this might help us catch bugs (if there are any).

relay-server/src/actors/processor.rs
jjbayer: This looks correct to me. `span::process` takes care of normalizing the span.

relay-server/src/actors/processor.rs
```diff
@@ -951,77 +1127,183 @@ impl EnvelopeProcessorService {
...
+    fn process_replays(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
+        replay::process(state, &self.inner.config)?;
+        if_processing!(self.inner.config, {
+            self.enforce_quotas(state)?;
+        });
+        Ok(())
+    }
+
+    /// Processes cron check-ins.
+    fn process_checkins(&self, _state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
```
jjbayer, line 1246:
```suggestion
    fn process_checkins(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
```

relay-server/src/envelope.rs
```diff
@@ -824,7 +824,7 @@ impl Item {
             ItemType::Attachment => true,
             ItemType::FormData => true,
             ItemType::RawSecurity => true,
-            ItemType::Nel => false,
+            ItemType::Nel => true,
```
iker-barriocanal: Why does NEL now require an event? `split_envelope` already extracts them independently.

relay-server/src/actors/processor.rs
```diff
@@ -75,9 +78,163 @@ mod span;
...
+            } else {
+                ProcessingGroup::Error
+            };
+            envelopes.push((
+                group,
+                Envelope::from_parts(headers.clone(), require_event_items),
+            ))
+        }
+
+        if !standalone_attachment_items.is_empty() {
```
iker-barriocanal: Why do we extract them into a variable above, and here append that variable to the result if it isn't empty? Is it possible to append the attachment to the result directly?

relay-server/src/utils/managed_envelope.rs
```diff
@@ -145,15 +161,21 @@ impl ManagedEnvelope {
...
-        Self::new_internal(envelope, Some(slot), outcome_aggregator, test_store)
+        Self::new_internal(envelope, Some(slot), outcome_aggregator, test_store, group)
     }
 
     /// Returns a reference to the contained [`Envelope`].
     pub fn envelope(&self) -> &Envelope {
         self.envelope.as_ref()
     }
 
+    /// Returns the optional group where this envelope belongs to.
```
iker-barriocanal: Is the group optional? It seems enforced to me.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -398,12 +399,18 @@ impl OnDisk {
...
 
-        let managed_envelope = self.buffer_guard.enter(
-            envelope,
-            services.outcome_aggregator.clone(),
-            services.test_store.clone(),
-        )?;
-        Ok(managed_envelope)
+        ProcessingGroup::split_envelope(*envelope)
+            .into_iter()
+            .map(|(group, envelope)| {
```
iker-barriocanal, line 404: What's the behavior when an old relay attempts to spool an envelope with a group that was added in a more recent version, and we don't fall back to `Unknown`? Maybe I'm getting too paranoid here.

iker-barriocanal (commented):
Looks solid! Leaving a few questions.

relay-server/src/actors/processor.rs
```diff
@@ -216,50 +137,92 @@ impl ProcessingGroup {
...
+        let span_items = envelope
+            .take_items_by(|item| item.ty() == &ItemType::Span || item.ty() == &ItemType::OtelSpan);
+        if !span_items.is_empty() {
+            grouped_envelopes.push((
+                ProcessingGroup::Span,
+                Envelope::from_parts(headers.clone(), span_items),
+            ))
+        }
+
+        // Extract all the items which require event into separate envelope.
```
```suggestion
        // Extract all the items which require an event into separate envelope.
```

relay-server/src/actors/processor.rs
```diff
@@ -216,50 +137,92 @@ impl ProcessingGroup {
...
+            let headers = headers.clone();
+            let items: SmallVec<[Item; 3]> = smallvec![item.clone()];
+            let envelope = Envelope::from_parts(headers, items);
+            let item_type = item.ty();
+            let group = if item_type == &ItemType::CheckIn {
+                ProcessingGroup::CheckIn
+            } else if item_type == &ItemType::UserReport || item_type == &ItemType::ClientReport {
+                ProcessingGroup::UserReport
+            } else {
+                relay_log::error!("Unsupported ItemType in the envelope: {item_type}");
```
```suggestion
                if item_type != &ItemType::Unknown { // Unknown item types are forwarded by design
                    relay_log::error!("Unsupported ItemType in the envelope: {item_type}");
                }
```

See https://github.com/getsentry/relay/blob/d17608ca5d0cb9b897d7c96e99b5a7cddb09ae82/relay-server/src/envelope.rs#L126-L129

relay-server/src/actors/processor.rs
```diff
@@ -1279,8 +1249,82 @@ impl EnvelopeProcessorService {
         Ok(())
     }
 
+    /// Previous / Old implementation of the `process_state` function, which is used for all
+    /// unknown [`ProcessingGroup`] variant.
+    ///
+    /// Note: this will be removed once we are confident in the new implementation and make sure
+    /// that all the groups properly covered.
+    fn process_state_pre(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
```
jjbayer: nit: I would rename this to `process_state_old` or `process_state_legacy`. When I first saw the function name I thought it was a pre-processor.

relay-server/src/actors/processor.rs
jjbayer: Better idea: We can have a separate enum variant for deliberately forwarded unknown item types, and one for ungroupble items. I.e.

```rust
enum ProcessingGroup {
    // ...
    ForwardUnknown, // This is where ItemType::Unknown goes. No processing is done to it.
    Ungrouped, // this is everything we failed to group -> should log an error.

File: relay-server/src/endpoints/common.rs [hunk 40]
```diff
@@ -356,6 +332,9 @@ pub async fn handle_envelope(
             envelope,
             state.outcome_aggregator().clone(),
             state.test_store().clone(),
+            // It's not clear at this point which group this envelope belongs to.
+            // The decission will be made while queueing in `queue_envelope` function.
+            ProcessingGroup::Unknown,
```
jjbayer: Ideally this would be reflected in the type somehow (for example `ManagedEnvelope` vs. `GroupedManagedEnvelope`), but that would be too big of a refactor now, so this is fine.

relay-server/src/actors/spooler/mod.rs
jjbayer: @iker-barriocanal how would that happen? Only the serializable envelope is written to disk (not the managed envelope or the group), so the Relay version that reads the envelope from disk decides how it wants to group the envelope. Which is as it should be IMO.

jjbayer (approved):
See comment about the "known unknown" group of items. Everything else looks good to me!

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: Oh right, I assumed it was the managed envelope I guess.

iker-barriocanal (approved):
🚀

relay-server/src/actors/processor.rs
olksdr: done

relay-server/src/actors/processor.rs
olksdr: done, now skipping processing for all the  unknown item types

relay-server/src/endpoints/common.rs
olksdr: Yeah, it's already too many changes now, in the follow-ups when we start to introduce types, we can take care of this.

jjbayer (approved):

Dav1dde (approved):
:shipit:

PR #2864, opened 2023-12-18
build: Update axum and http

jan-auer:
We would like to update the Sentry SDK, which requires to update `axum` and related dependencies. As `reqwest` has not updated to the latest version of `http` and `hyper` yet, we will temporarily run two versions of the frameworks side-by-side and convert between their types in the forward endpoint.

The `axum-server` library also has not released their latest version yet. Since the only thing needed from axum-server is graceful shutdown, we implement this ourselves now following the example in the `axum` repository.

#skip-changelog


jan-auer:
This PR caused an increase in CPU utilization, which needs to be fixed before merging this again.

iker-barriocanal:
~What's the benefit of this upgrade?~ _(found some more context)_ I wonder because this may cause some confusion:

> we will temporarily run two versions of the frameworks side-by-side and convert between their types in the forward endpoint.

olksdr (approved):
:shipit:  let's try it again

Dav1dde:
> :shipit: let's try it again

Do we have an idea what caused the spike?

jan-auer:
> What's the benefit of this upgrade?

See the pull request description. We'd like to upgrade the Sentry SDK and since the axum integration there already depends on version `0.7`, we'll have to do the update now. Alternatively, we could also vendor the integration and disable the feature flag, but there are some nice code improvements that come with the new `Request` type.

> Do we have an idea what caused the spike?

Not yet sure. In the last attempt, we did not have TCP keepalive configured on the socket, as hyper has dropped that functionality. In 56477c5, I'm doing this manually now in the same way that hyper configured it.

jan-auer:
Another test showed that there's a single thread in `main-rt` that is running at 100% CPU utilization. We have not been able to pull a stack trace from that thread yet.

PR #2867, opened 2023-12-19
fix(spans): Collapse constants in SQL select

jjbayer:
Before:

```sql
SELECT foo, bar, NULL, baz FROM my_table
-- became
SELECT  .., %s, baz FROM my_table
```

After:

```sql
SELECT foo, bar, NULL, baz FROM my_table
-- becomes
SELECT .. FROM my_table
```

#skip-changelog


olksdr (approved):

Dav1dde (approved):

iker-barriocanal (approved):

gggritso (approved):
👍🏻

PR #2868, opened 2023-12-19
fix(spans): Detect hex with fallback scrubber

jjbayer:
In some cases we still rely on regex scrubbing for SQL queries. Those regexes did not detect hex values (e.g. `0xA1B2C3D4`) until now.

#skip-changelog


olksdr (approved):

relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -34,6 +34,8 @@ static NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
         # Capture ODBC escape sequence.
         ((?-x)(?P<odbc_escape_sequence>\{(?:ts?|d)\s+'.+'\})) |
         ((?-x)(?P<number>(-?\b(?:[0-9]+\.)?[0-9]+(?:[eE][+-]?[0-9]+)?\b)(::\w+(\[\]?)?)?)) |
+        # Hex constants
+        ((?-x)(?P<hex>(\b0x[0-9a-f]+\b)(::\w+(\[\]?)?)?)) |
```
Dav1dde, line 38: Doesn't this need to be `[0-9a-fA-F]`?

jjbayer: There's `(?xi)` at the beginning of the regex. I updated the unit test to double check.

Dav1dde (approved):

PR #2869, opened 2023-12-19
feat(server): Add an internal endpoint for batched metrics

jan-auer:
Introduces an endpoint that allows Relays to send metrics from multiple projects and organizations in a single request. This endpoint requires Relay authentication.

The endpoint takes a JSON payload in the following format:

```
{
  "project_key1": [ /* buckets... */ ],
  "project_key2": [ /* buckets... */ ],
  // ...
}
```

Emission of such requests will be implemented in a separate PR.


relay-server/src/endpoints/batch_metrics.rs
```diff
@@ -0,0 +1,29 @@
...
+    body: SignedBytes,
+) -> impl IntoResponse {
+    if !body.relay.internal || state.config().emit_outcomes() != EmitOutcomes::AsOutcomes {
+        return StatusCode::FORBIDDEN.into_response();
+    }
+
+    state.processor().send(ProcessBatchedMetrics {
+        payload: body.body,
+        start_time: start_time.into_inner(),
+        sent_at: None,
```
iker-barriocanal, line 25: Any reason to not set `sent_at`?

relay-server/src/extractors/signed_json.rs
```diff
@@ -101,7 +100,36 @@ where
...
+    B::Data: Send,
+    B::Error: Into<axum::BoxError>,
+{
+    type Rejection = SignatureError;
+
+    async fn from_request(
+        request: Request<B>,
+        state: &ServiceState,
+    ) -> Result<Self, Self::Rejection> {
+        let SignedBytes { body, relay } = SignedBytes::from_request(request, state).await?;
```
iker-barriocanal, line 131: What do we need `SignedJson` for? We have `SignedBytes` and use it here, plus the actor gets bytes and assumes it's valid JSON.

relay-server/src/actors/processor.rs
```diff
@@ -1181,6 +1203,40 @@ impl EnvelopeProcessorService {
         }
     }
 
+    fn handle_process_batched_metrics(&self, message: ProcessBatchedMetrics) {
```
iker-barriocanal, line 1206: Is it worth extracting metric processing into a module? We've already done this for a few cases, [see](https://github.com/getsentry/relay/tree/eac1481fe37967269658fe85516404f096f696e2/relay-server/src/actors/processor).

relay-server/src/endpoints/batch_metrics.rs
jan-auer: We don't have it in that endpoint. Since this endpoint is just for inter-Relay communciation, there's no clock drift correction necessary.

@jjbayer suggested to still implement clock drift in the internal message, so we don't forget about it once we use the message from some other place.

relay-server/src/extractors/signed_json.rs
jan-auer: `SignedJson` is used by all other endpoints, which we're not refactoring in this PR. The usual pattern is to validate + deserialize JSON directly in the extractor.

relay-server/src/actors/processor.rs
jan-auer: Agreed. @olksdr would you like to move this afterwards as part of your refactors?

iker-barriocanal (approved):

relay-server/src/actors/processor.rs
olksdr: @jan-auer  yeah, I will have a look into this, and try to refactor into the separate module.

PR #2871, opened 2023-12-19
feat(spans): Group interactions by component name

jjbayer:
Spans with op `ui.interaction` or `ui.react` will be grouped by `span.data.ui.component_name`.

The potentially high-cardinality description is replaced with the low-cardinality component name.

#skip-changelog


relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_all_modules.snap
```diff
@@ -1692,4 +1692,52 @@ expression: metrics
...
+        width: 0,
+        name: "d:spans/exclusive_time@millisecond",
+        value: Distribution(
+            [
+                32.000065,
+            ],
+        ),
+        tags: {
+            "environment": "fake_environment",
+            "span.description": "my-component-name",
```
jjbayer: @0Calories the span description is overwritten by the component name in this implementation, because `span.description` on span metrics and in the spans dataset has so far always contained the human readable group name. This does mean that the original description is only available via nodestore though.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -76,6 +76,10 @@ pub(crate) fn scrub_span_description(span: &Span) -> Option<String> {
                 // be low-risk to start adding the description.
                 Some(description.to_owned())
             }
+            ("ui", sub) if sub.starts_with("interaction.") || sub.starts_with("react.") => data
+                .and_then(|data| data.get("ui.component_name"))
```
iker-barriocanal, line 80: Should we have a fallback in case `ui.component_name` doesn't exist?

olksdr (approved):

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
jjbayer: The fallback is `None`, so no description and no group hash. This is intentional. Good point though!

0Calories (approved):

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_all_modules.snap
```diff
@@ -951,4 +951,30 @@ expression: metrics
...
+        name: "d:spans/exclusive_time_light@millisecond",
+        value: Distribution(
+            [
+                32.000065,
+            ],
+        ),
+        tags: {
+            "span.description": "my-component-name",
+            "span.group": "e674f9eca1d88a4d",
+            "span.op": "ui.interaction.click",
```
jjbayer, line 966: @0Calories I forgot that we now have stricter metrics extraction rules, so we need to specify exactly what metrics and tags we want for these interaction spans. For now, I started with just these three, but let me know if other tags are needed in the product.

relay-dynamic-config/src/defaults.rs
```diff
@@ -98,6 +99,12 @@ fn span_metrics(is_extract_all: bool) -> impl IntoIterator<Item = MetricSpec> {
 
     let is_mobile = is_mobile_sdk.clone() & (is_mobile_op.clone() | is_screen);
 
+    let is_interaction = if is_extract_all {
+        RuleCondition::glob("span.op", "ui.interaction.*")
+    } else {
+        RuleCondition::never()
+    };
```
jjbayer, line 106: @0Calories this means that span metrics for interaction spans will only exist for our own orgs for now. We have to remove this condition before GA'ing.

jjbayer: Never mind, that feature flag is always enabled these days. https://github.com/getsentry/sentry-options-automator/pull/603

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_all_modules.snap
0Calories: thanks @jjbayer , these are fine for now but in the future when we need to support more usecases we may need more metrics. Will reach out to you later when that time comes

PR #2872, opened 2023-12-19
feat(spans): Extract span count per segment name and release

shruthilayaj:
For given screen, transaction.op and releases,
we want to be able to show a breakdown of the
% of spans we are not extracting metrics for.
This metric allows us to get the total span
count. We should be okay just extracting it for
mobile spans right now, no other use cases
in the product.

#skip-changelog


iker-barriocanal (approved):

relay-dynamic-config/src/defaults.rs
```diff
@@ -338,6 +338,23 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
                     .always(),
             ],
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "c:spans/count_per_segment@none".into(),
```
jjbayer, line 343: Did you add this string to https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py? Please make sure that's deployed before merging this PR.

jjbayer (approved):

relay-dynamic-config/src/defaults.rs
shruthilayaj: gosh darn it, here you go! https://github.com/getsentry/sentry/pull/62079

PR #2874, opened 2023-12-19
feat(spans): Extract metrics for file spans

phacops:


relay-dynamic-config/src/defaults.rs
```diff
@@ -63,6 +66,7 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
         (
             RuleCondition::eq("span.op", "http.client")
                 | RuleCondition::glob("span.op", MOBILE_OPS)
+                | RuleCondition::glob("span.op", FILE_OPS)
```
```suggestion
                | RuleCondition::eq("span.op", FILE_OPS)
```

jjbayer (approved):

relay-dynamic-config/src/defaults.rs
```diff
@@ -26,6 +26,9 @@ const MONGODB_QUERIES: &[&str] = &["*\"$*", "{*", "*({*", "*[{*"];
 /// A list of patterns for resource span ops we'd like to ingest.
 const RESOURCE_SPAN_OPS: &[&str] = &["resource.script", "resource.css", "resource.img"];
 
+/// A list of file span ops we'd like to ingest.
+const FILE_OPS: &[&str] = &["file.read", "file.write"];
```
iker-barriocanal: [Docs](https://develop.sentry.dev/sdk/performance/span-operations/#mobile) also mention `file.delete`, `file.open` and `file.rename`; not sure if we also want to include these.

iker-barriocanal (approved):

relay-dynamic-config/src/defaults.rs
phacops: Good point. We're indexing all of them even if we're targeting `read` and `write` for mobile, might as well extract metrics for all of them.

PR #2877, opened 2023-12-20
feat(processor): Add a metric to track the time of all messages

jan-auer:
In order to estimate the throughput and capacity of the envelope
processor service, we're tracking the number and duration of all
processed messages. This should allow us to assess when the CPU pool is
reaching its limit.

The metric is tracked as `processor.message.duration` and has a
`message` tag containing the name of the variant / struct that is being
processed. This allows to break down the volume and duration of each of
the messages by type.


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 **Internal**:
```
olksdr, line 5: Could you, please, add `Unreleased` section name?

olksdr (approved):

CHANGELOG.md
Dav1dde: I definitely didn't forget to add this in my last PR...

Dav1dde (approved):

iker-barriocanal (approved):

PR #2878, opened 2023-12-20
fix(normalization): Normalize timestamps before TimestampProcessor

### LINKED ISSUE #2851: Run ClockDriftProcessor before time range validation for transactions
https://github.com/getsentry/relay/pull/2406 added time range validation for transactions, but it runs before clock drift processing.

We should switch the order to allow clock drift-corrected transactions to be accepted.

ref: https://github.com/getsentry/sentry/issues/60739

iker-barriocanal:
This PR moves timestamp normalization to happen before running the `TimestampProcessor` during event normalization.

`TimestampProcessor` rejects events with stale timestamps. However, it's possible that the clock in a network isn't up-to-date and events are sent with a stale timestamp even if they are just generated, resulting in Relay rejecting these events. The timestamp normalization step solves the clock issue but must happen before validation.

Fixes https://github.com/getsentry/relay/issues/2851.

### DIFF
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - This release requires Python 3.9 or later. There are no intentionally breaking changes included in this release, but we stopped testing against Python 3.8.
-- Normalize event timestamps before validating them, fixing cases where Relay would drop valid events with reason "invalid_transaction". ([#2851](https://github.com/getsentry/relay/issues/2851))
+- Normalize event timestamps before validating them, fixing cases where Relay would drop valid events with reason "invalid_transaction". ([#2878](https://github.com/getsentry/relay/pull/2878))
 
 ## 0.8.39
 
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 This release requires Python 3.9 or later. There are no intentionally breaking changes included in this release, but we stopped testing against Python 3.8.
 
+- Normalize event timestamps before validating them, fixing cases where Relay would drop valid events with reason "invalid_transaction". ([#2851](https://github.com/getsentry/relay/issues/2851))
+
 ## 0.8.39
 
 - Add `_metrics_summary` as temporary key on `Event` for a DDM experiment. ([#2757](https://github.com/getsentry/relay/pull/2757))
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## Unreleased
 
-This release requires Python 3.9 or later. There are no intentionally breaking changes included in this release, but we stopped testing against Python 3.8.
-
+- This release requires Python 3.9 or later. There are no intentionally breaking changes included in this release, but we stopped testing against Python 3.8.
 - Normalize event timestamps before validating them, fixing cases where Relay would drop valid events with reason "invalid_transaction". ([#2851](https://github.com/getsentry/relay/issues/2851))
 
 ## 0.8.39
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -2344,20 +2344,24 @@ mod tests {
 
     #[test]
     fn test_reject_stale_transactions_after_timestamp_normalization() {
+        let now = Utc::now();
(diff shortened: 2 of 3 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -202,6 +202,13 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) -
     // Check for required and non-empty values
     schema::SchemaProcessor.process_event(event, meta, ProcessingState::root())?;
 
+    normalize_timestamps(
+        event,
+        meta,
+        config.received_at,
+        config.max_secs_in_past,
+        config.max_secs_in_future,
+    )?; // Timestamps are core in the metrics extraction
```
olksdr, line 211: After this normalization applied, the `TimestamProcessor::process_event` does not have anything todo, does it? 

Correct me if I'm missing something in the code, but it seems like the timestamps will already be set to correct values.

py/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 This release requires Python 3.9 or later. There are no intentionally breaking changes included in this release, but we stopped testing against Python 3.8.
 
+- Normalize event timestamps before validating them, fixing cases where Relay would drop valid events with reason "invalid_transaction". ([#2851](https://github.com/getsentry/relay/issues/2851))
```
olksdr: Could you , please fix the record above to start with `-` and remove space between them?

relay-event-normalization/src/event.rs
iker-barriocanal: Hmm, that makes sense. Timestamps were already processed when we introduced `TimestampProcessor`, so I wonder what I'm missing. I'll have a look at it, thanks for pointing it out!

iker-barriocanal: `TimestampProcessor` is still useful: timestamp normalization (mostly clockdrift) is sometimes not enough to accept a transaction. For example, when the end timestamp is up-to-date but the start timestamp is still stale after normalizing it. I added a test case covering it in [d09b833](https://github.com/getsentry/relay/pull/2878/commits/d09b8332ec2cf24920d9403f60153352e483a50e).

iker-barriocanal:
Half of the commits in this pr are updating and fixing the changelog 😂

relay-event-normalization/src/event.rs
```diff
@@ -2321,4 +2321,53 @@ mod tests {
...
+        let config = NormalizationConfig {
+            received_at: Some(Utc::now()),
+            max_secs_in_past: Some(2),
+            max_secs_in_future: Some(1),
+            ..Default::default()
+        };
+
+        let json = r#"{
+  "event_id": "52df9022835246eeb317dbd739ccd059",
+  "transaction": "i have an stale timestamp, but im recent!",
```
```suggestion
  "transaction": "i have a stale timestamp, but i'm recent!",
```

relay-event-normalization/src/event.rs
```diff
@@ -2321,4 +2321,53 @@ mod tests {
...
+            max_secs_in_future: Some(1),
+            ..Default::default()
+        };
+
+        let json = format!(
+            r#"{{
+          "event_id": "52df9022835246eeb317dbd739ccd059",
+          "transaction": "clockdrift is not enough to accept me :(",
+          "start_timestamp": -62135811111,
+          "timestamp": {}
```
jjbayer, line 2367: Why did you set `timestamp` to an empty object here?

relay-event-normalization/src/event.rs
```diff
@@ -2321,4 +2321,53 @@ mod tests {
...
+            max_secs_in_past: Some(2),
+            max_secs_in_future: Some(1),
+            ..Default::default()
+        };
+
+        let json = format!(
+            r#"{{
+          "event_id": "52df9022835246eeb317dbd739ccd059",
+          "transaction": "clockdrift is not enough to accept me :(",
+          "start_timestamp": -62135811111,
```
jjbayer, line 2366: Could you explain why this test fails and the other one passes? Both contain timestamps that are extremely out of date.

Maybe it also helps to assert on the corrected timestamp in the passing test.

jjbayer (approved):
Please add comments to the test cases as to why they pass / fail.

olksdr (approved):

PR #2879, opened 2023-12-20
ref(metric-meta): Optimize Redis rountrips by using a pipe

Dav1dde:
#skip-changelog


olksdr (approved):
Thanks for working on this! :shipit:

iker-barriocanal:
Nice! I wonder if we have any data somewhere with the before/after deltas of using pipes.

Dav1dde:
> Nice! I wonder if we have any data somewhere with the before/after deltas of using pipes.

We actually have now, Jan added some metrics and made some new dashboards, that's how we noticed that metric meta is slow.

PR #2881, opened 2023-12-20
feat(stats-detectors): Add transaction.op to duration light

Zylphrex:
We actually need to extract transaction.op as well since some transactions can have the same name but different ops.


phacops (approved):

jjbayer (changes requested):
Couldn't you query this using the metric `d:transactions/duration@millisecond`? Or do you need the `d:transactions/duration_light@millisecond` for performance reasons?

Zylphrex:
> Couldn't you query this using the metric `d:transactions/duration@millisecond`? Or do you need the `d:transactions/duration_light@millisecond` for performance reasons?

`d:transactions/duration_light@millisecond` was added for performance reasons since we'll never filter any tags. However, one oversight was that we never distinguished between transactions with the same name but different operations.

The main use case for this metric is for statistical detectors where we query for the top transactions + their time series for every project to identify potential regressions.

jjbayer (approved):
@Zylphrex thanks for the extra context!

PR #2884, opened 2023-12-21
fix(spans): Scrub double-quoted strings in fallback mode

jjbayer:
Double quotes delimit identifiers (for example, column names) in most dialects, but in MySQL they can be also used for string values. When a query cannot be parsed for whatever reason, our fallback regex scrubber should remove these strings as it does for other dialects.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/sql/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/mod.rs
@@ -43,6 +43,17 @@ static NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     .unwrap()
 });
 
+/// For MySQL, also look for double quoted strings.
+static DOUBLE_QUOTED_STRING_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(
+        r#"(?xi)
+        # Capture double-quoted strings, including the remaining substring if `\'` is found.
+        ((?-x)(?P<single_quoted_strs>N?"(?:\\"|[^"])*(?:"|$)(::\w+(\[\]?)?)?))
+        "#,
+    )
+    .unwrap()
+});
+
 /// Removes extra whitespace and newlines.
 static WHITESPACE: Lazy<Regex> = Lazy::new(|| Regex::new(r"(\s*\n\s*)|(\s\s+)").unwrap());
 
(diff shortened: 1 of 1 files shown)


relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -43,6 +43,17 @@ static NORMALIZER_REGEX: Lazy<Regex> = Lazy::new(|| {
     .unwrap()
 });
 
+/// For MySQL, also look for double quoted strings.
+static DOUBLE_QUOTED_STRING_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(
+        r#"(?xi)
+        # Capture double-quoted strings, including the remaining substring if `\'` is found.
+        ((?-x)(?P<single_quoted_strs>N?"(?:\\"|[^"])*(?:"|$)(::\w+(\[\]?)?)?))
```
jjbayer, line 51: Copied and adapted from the regex defined above.

relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -133,22 +144,38 @@ fn scrub_queries_inner(db_system: Option<&str>, string: &str) -> (Option<String>
...
+    }
+    if let Cow::Owned(s) = STRIP_QUOTES.replace_all(&string, "$entity_name") {
+        string = Cow::Owned(s);
+    }
+    if let Cow::Owned(s) = COLLAPSE_ENTITIES.replace_all(&string, "$entity_name") {
+        string = Cow::Owned(s);
+    }
+    if let Cow::Owned(s) = COLLAPSE_COLUMNS.replace_all(&string, "$pre..$post") {
+        string = Cow::Owned(s);
+    }
```
jjbayer, line 178: This is more verbose, but also more in line with [our internal guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#941ff156a7064236910ce1d1684cdb30).

relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -768,6 +795,16 @@ mod tests {
         "SELECT %s, %s"
     );
 
+    scrub_sql_test_with_dialect!(
+        dont_fallback_to_regex,
+        "mysql",
+        // sqlparser cannot parse REPLACE INTO. If we know that
+        // a query is MySQL, we should give up rather than try to scrub
+        // with regex
```
jjbayer, line 803:
```suggestion
        // a query is MySQL, we should scrub double-quoted strings
```

olksdr (approved):

PR #2885, opened 2023-12-21
ref(server): Improve request body and signature handling

jan-auer:
Changes the signature of `UpstreamRequest` so that any request can be
configured to be signed now. Before, this was possible only for query
requests. The `UpstreamRequest` trait now has a `sign() -> bool` method
that returns whether or not a signature should be applied.

Along with this, there are changes to the request builder to make it
more ergonomic for the common use case:

- The builder uses interior mutability instead of moving now.
- `RequestBuilder` now takes `Into<Bytes>` as body, which allows for
  more efficient retries. However, it is still the responsibility of the
  request to ensure bytes are cloned. This is because usually we want
  expensive serialization operations outside of the Upstream's runtime.
- The `build()` function no longer receives the config, as this was
  needed just for internal query requests.
- The `build()` function does not need to finish the request, this is
  done by the upstream now.
- `HttpError::NoCredentials` has been removed. This was a hack to allow
  emitting this error from the build callback, but this is no longer
  required.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/upstream.rs
+++ b/relay-server/src/actors/upstream.rs
@@ -12,6 +12,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
 
+use bytes::Bytes;
 use itertools::Itertools;
 use relay_auth::{RegisterChallenge, RegisterRequest, RegisterResponse, Registration};
 use relay_config::{Config, Credentials, RelayMode};
@@ -306,21 +307,29 @@ pub trait UpstreamRequest: Send + Sync + fmt::Debug {
         true
     }
 
+    fn sign(&self) -> bool {
+        false
+    }
+
     /// Returns the name of the logical route.
     ///
     /// This is used for internal metrics and logging. Other than the path, this cannot contain
     /// dynamic elements and should be globally unique.
     fn route(&self) -> &'static str;
 
+    /// Callback to apply configuration to the request.
+    fn configure(&mut self, _config: &Config) {}
+
     /// Callback to build the outgoing web request.
     ///
-    /// This callback populates the initialized request with headers and a request body. To send an
-    /// empty request without additional headers, call [`RequestBuilder::finish`] directly.
+    /// This callback populates the initialized request with headers and a request body.
     ///
     /// Note that this function can be called repeatedly if [`retry`](UpstreamRequest::retry)
     /// returns `true`. This function should therefore not move out of the request struct, but can
     /// use it to memoize heavy computation.
-    fn build(&mut self, config: &Config, builder: RequestBuilder) -> Result<Request, HttpError>;
+    fn build(&mut self, _builder: &mut RequestBuilder) -> Result<(), HttpError> {
+        Ok(())
+    }
 
     /// Callback to complete an HTTP request.
     ///
@@ -391,7 +400,7 @@ type QuerySender<T> = Sender<Result<<T as UpstreamQuery>::Response, UpstreamRequ
 #[derive(Debug)]
 struct UpstreamQueryRequest<T: UpstreamQuery> {
     query: T,
-    compiled: Option<(Vec<u8>, String)>,
+    body: Option<Bytes>,
     max_response_size: usize,
     sender: QuerySender<T>,
 }
@@ -404,7 +413,7 @@ where
     pub fn new(query: T, sender: QuerySender<T>) -> Self {
         Self {
             query,
-            compiled: None,
+            body: None,
             max_response_size: 0,
             sender,
         }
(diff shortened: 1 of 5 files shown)


relay-server/src/http.rs
```diff
@@ -39,51 +38,72 @@ impl HttpError {
...
-            HttpError::NoCredentials => false,
         }
     }
 }
 
 pub struct Request(pub reqwest::Request);
 
 pub struct RequestBuilder {
-    builder: reqwest::RequestBuilder,
+    builder: Option<reqwest::RequestBuilder>,
```
jan-auer, line 48: This is not actually optional. If there's another safe way to move the request builder, I'd prefer that but I haven't found one.

relay-server/src/actors/upstream.rs
```diff
@@ -306,21 +297,29 @@ pub trait UpstreamRequest: Send + Sync + fmt::Debug {
         true
     }
 
+    fn sign(&self) -> bool {
```
Dav1dde, line 306: Need docs here

relay-server/src/http.rs
Dav1dde: Yeah this is a bit unfortunate, but any reason our `RequestBuilder` can't keep taking `self`?

I sketched this up here: https://gist.github.com/Dav1dde/f10c001462b0df03acfc241c16a6d697

jan-auer: This is exactly the kind of signature I wanted to avoid, as it requires to return the builder from the trait method. It's what we had before. Do you prefer that signature? I can revert that change.

relay-server/src/actors/upstream.rs
```diff
@@ -306,21 +297,29 @@ pub trait UpstreamRequest: Send + Sync + fmt::Debug {
...
+    }
+
     /// Returns the name of the logical route.
     ///
     /// This is used for internal metrics and logging. Other than the path, this cannot contain
     /// dynamic elements and should be globally unique.
     fn route(&self) -> &'static str;
 
+    /// Callback to apply configuration to the request.
+    fn configure(&mut self, _config: &Config) {}
```
Dav1dde, line 323: Is it just me or does this feel a bit dirty? Instead of passing the necessary config when creating the request we allow for this "backdoor". Then again, this does seem useful...

Dav1dde (approved):

relay-server/src/http.rs
Dav1dde: No not necessarily, probably nicer overall with the `&mut` except for the builder itself.

relay-server/src/actors/upstream.rs
```diff
@@ -835,6 +835,7 @@ impl SharedClient {
         &self,
         request: &mut dyn UpstreamRequest,
     ) -> Result<Response, UpstreamRequestError> {
+        request.configure(&self.config);
```
Dav1dde, line 850: I think we should only configure once or very very close to build (right before calling build) and document when `configure` is called in the trait method docs.

relay-server/src/actors/upstream.rs
jan-auer: IMO passing the config into the builder was dirty, so this improves on that. 

Unfortunately, we need this as we have to create the `UpstreamQueryRequest` (an internal request not visible in the public signature) in a trait impl that doesn't have access to the configuration. The `configure` hook is the most generic way to allow requests to customize their behavior based on config.

relay-server/src/http.rs
```diff
@@ -39,51 +38,72 @@ impl HttpError {
...
     }
 
-    /// Add a new header, not replacing existing ones.
-    pub fn header(mut self, key: impl AsRef<str>, value: impl AsRef<[u8]>) -> Self {
-        self.builder = self.builder.header(key.as_ref(), value.as_ref());
+    fn build<F>(&mut self, mut f: F) -> &mut Self
+    where
+        F: FnMut(reqwest::RequestBuilder) -> reqwest::RequestBuilder,
+    {
+        self.builder = Some(f(self.builder.take().unwrap()));
```
Dav1dde:
```suggestion
        self.builder = self.builder.take().map(f); 
```

relay-server/src/actors/upstream.rs
jan-auer: See this commit: 25a73d5fb7fecc2994bfe5fc7a718dc2ee48515e, this is exactly what's happening. I'll add a mention to the docs.

Dav1dde: Imo we should also move the `request.configure(&self.config);` call right before `request.build(&mut builder)?;`

So this isnt missed in a future refactor

PR #2886, opened 2023-12-21
ref(server): Create reusable utils for metric outcomes

jan-auer:
Moves utilities to collect `SourceQuantities` from metrics as well as logging
outcomes from `SourceQuantities` into a reusable place, so that we can check and
enforce rate limits outside of the processor.

This PR also fixes two issues with metric rate limiting:
- Emit the correct reason code for cached rate limits. Before, this would emit
  the reason code of the longest quota independently of whether that quota
  matched.
- Lower the log level for rate limited buckets to DEBUG

#skip-changelog

### DIFF
--- a/relay-metrics/src/view.rs
+++ b/relay-metrics/src/view.rs
@@ -96,12 +96,12 @@ impl<'a> BucketsView<'a> {
     }
 
     /// Iterator over all buckets in the view.
-    pub fn iter(&self) -> impl Iterator<Item = BucketView<'a>> {
+    pub fn iter(&self) -> BucketsViewIter<'a> {
         BucketsViewIter::new(self.inner, self.start, self.end)
     }
 
     /// Iterator which slices the source view into segments with an approximate size of `size_in_bytes`.
-    pub fn by_size(&self, size_in_bytes: usize) -> impl Iterator<Item = BucketsView<'a>> {
+    pub fn by_size(&self, size_in_bytes: usize) -> BucketsViewBySizeIter<'a> {
         BucketsViewBySizeIter::new(self.inner, self.start, self.end, size_in_bytes)
     }
 }
@@ -113,10 +113,19 @@ impl<'a> fmt::Debug for BucketsView<'a> {
     }
 }
 
+impl<'a> IntoIterator for &'_ BucketsView<'a> {
+    type Item = BucketView<'a>;
+    type IntoIter = BucketsViewIter<'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
 /// Iterator yielding all items contained in a [`BucketsView`].
 ///
 /// First and/or last item may be partial buckets.
-struct BucketsViewIter<'a> {
+pub struct BucketsViewIter<'a> {
     /// Source slice of buckets.
     inner: &'a [Bucket],
     /// Current index.
@@ -185,7 +194,7 @@ impl<'a> Iterator for BucketsViewIter<'a> {
 /// Iterator slicing a [`BucketsView`] into smaller views constrained by a given size in bytes.
 ///
 /// See [`estimate_size`] for how the size of a bucket is calculated.
-struct BucketsViewBySizeIter<'a> {
+pub struct BucketsViewBySizeIter<'a> {
     /// Source slice.
     inner: &'a [Bucket],
     /// Current position in the slice.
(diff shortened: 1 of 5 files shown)


relay-metrics/src/view.rs
```diff
@@ -96,12 +96,12 @@ impl<'a> BucketsView<'a> {
     }
 
     /// Iterator over all buckets in the view.
-    pub fn iter(&self) -> impl Iterator<Item = BucketView<'a>> {
+    pub fn iter(&self) -> BucketsViewIter<'a> {
```
jan-auer, line 99: Since the iterator types are now public, we can name them in these methods here, too. This is also in accordance with our style guide.

relay-server/src/actors/processor.rs
```diff
@@ -1459,21 +1389,16 @@ impl EnvelopeProcessorService {
...
-                .and_then(|limit| limit.reason_code.clone());
-
-            self.drop_buckets_with_outcomes(
-                reason_code,
-                total_buckets,
+        let limits = cached_rate_limits.check_with_quotas(quotas, item_scoping);
+        if limits.is_limited() {
+            relay_log::debug!("dropping {total_buckets} buckets due to throughput ratelimit");
+
+            let reason_code = limits.longest().and_then(|limit| limit.reason_code.clone());
```
jan-auer, line 1396: The bugfix is here. This uses the returned `limits` from `check_with_quotas` instead of the original `cached_rate_limits`.

relay-server/src/actors/processor.rs
```diff
@@ -1715,15 +1643,6 @@ impl Service for EnvelopeProcessorService {
     }
 }
 
-fn create_metrics_item(buckets: &BucketsView<'_>, extraction_mode: ExtractionMode) -> Item {
```
jan-auer, line 1718: This function could be trivially inlined and simplified

olksdr (approved):

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -36,18 +37,15 @@ pub struct MetricsLimiter<Q: AsRef<Vec<Quota>> = Vec<Quota>> {
...
 /// of datapoints contained in the bucket.
 ///
 /// Additionally tracks whether the transactions also contained profiling information.
 ///
 /// Returns `None` if the metric was not extracted from transactions.
-pub fn extract_transaction_count(
-    metric: &BucketView<'_>,
-    mode: ExtractionMode,
-) -> Option<TransactionCount> {
+fn count_metric_bucket(metric: BucketView<'_>, mode: ExtractionMode) -> Option<TransactionCount> {
```
jjbayer, line 48: Should we get rid of `TransactionCount` and just return a `SourceQuantities` here (with `buckets` set to 1)? Or rather, unite this function with `extract_metric_quantities`. The function name already suggests it counts more than just transaction and profile info.

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -73,6 +71,54 @@ pub fn extract_transaction_count(
...
+    let timestamp = Utc::now();
+
+    let categories = [
+        (DataCategory::Transaction, quantities.transactions as u32),
+        (DataCategory::Profile, quantities.profiles as u32),
+        (DataCategory::MetricBucket, quantities.buckets as u32),
+    ];
+
+    for (category, quantity) in categories {
+        addr.send(TrackOutcome {
```
jjbayer: Shouldn't we check for `quantity > 0` here?

jjbayer (approved):
Nice! See question about `quantity > 0`.

relay-server/src/utils/metrics_rate_limits.rs
jan-auer: Good catch, thank you!

relay-server/src/utils/metrics_rate_limits.rs
jan-auer: Discussed with @Dav1dde and decided that it's useful to have both functions still. I'm open to more refactoring but since both of those are internal functions it's not a priority for me right now.

jjbayer: Makes sense. I would personally rename it back to something like `get_transaction_count` then, but up to you.

Dav1dde: Agree with Joris here, I think the original name `extract_transaction_count` or `get_transaction_count` captures the intent of the function better.

We still need this behaviour for `MetricsLimiter` I believe (so can't just yet get rid of `TransactionCount` and combine it with `extract_metric_quantities`)

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -73,6 +71,56 @@ pub fn extract_transaction_count(
...
+            }
+        }
+    }
+
+    quantities
+}
+
+pub fn reject_metrics(
+    addr: Addr<TrackOutcome>,
+    quantities: SourceQuantities,
```
Dav1dde, line 97: Could take the `IntoIterator<Item = impl From<BucketView<'a>>>` directly here

Dav1dde (approved):

relay-server/src/utils/metrics_rate_limits.rs
jan-auer: That would be a bit less practical, as computing source quantities also needs the extraction mode from the project state. This signature should encourage reusing the quantities as much as possible, since those are expensive to create.

PR #2887, opened 2023-12-21
ref(normalization): Use numeric values in max_chars

iker-barriocanal:
This PR switches the expected value from a string to a usize for the `max_chars` field in an attribute's metastructure.

String values are confusing and may lead to investigating where the string mapping is happening (the `MaxChars` enum) and how these are applied. Further, updating `MaxChars` values impact multiple fields and may have unintended effects. Setting the limits individually solves these two problems.

### Minimal `MaxChars`

Some normalization steps require knowing the `max_chars` value of the field, so there's a minimal `MaxChars` enum to keep a duplicate of the limits. Adding multiple constants where required is another alternative, but spreading the thresholds makes identifying them harder to spot. There's a risk the enum gets out of sync, but failed tests should prevent regressions.

## Future work

- Run `TrimmingProcessor` for replays.
- Get rid of bag sizes.

#skip-changelog


relay-event-normalization/src/trimming.rs
```diff
@@ -430,20 +420,20 @@ mod tests {
         let mut processor = TrimmingProcessor::new();
 
         let mut event = Annotated::new(Event {
-            culprit: Annotated::new("x".repeat(300)),
+            logger: Annotated::new("x".repeat(300)),
```
iker-barriocanal, line 433: To avoid hardcoding a number in the `trim_string` below, I've switched the field to `logger` as it has a `max_chars` value in the metastructure and a `MaxChars` variant.

relay-event-derive/src/lib.rs
```diff
@@ -544,17 +524,16 @@ fn parse_field_attributes(
...
                                 match name_value.lit {
-                                    Lit::Str(litstr) => {
-                                        let attr = parse_max_chars(litstr.value().as_str());
-                                        rv.max_chars = Some(quote!(#attr));
+                                    Lit::Int(litint) => {
+                                        rv.max_chars = Some(quote!(#litint));
                                     }
                                     _ => {
-                                        panic!("Got non string literal for max_chars");
+                                        panic!("Got non usize literal for max_chars");
```
jjbayer: nit

```suggestion
                                        panic!("Got non integer literal for max_chars");
```

relay-event-normalization/src/trimming.rs
```diff
@@ -280,37 +277,29 @@ fn trim_string(value: &mut String, meta: &mut Meta, max_chars: MaxChars) {
             let chunk_chars = chunk.count();
 
             // if the entire chunk fits, just put it in
-            if length + chunk_chars < soft_limit {
+            if length + chunk_chars < max_chars {
```
jjbayer, line 282: Shouldn't we distinguish between hard and soft limit here somehow? Or was `trim_string` always called with `MaxChars::Hard`?

relay-event-normalization/src/trimming.rs
```diff
@@ -280,37 +277,29 @@ fn trim_string(value: &mut String, meta: &mut Meta, max_chars: MaxChars) {
...
                 new_chunks.push(chunk);
                 length += chunk_chars;
                 continue;
             }
 
-            match chunk {
-                // if there is enough space for this chunk and the 3 character
-                // ellipsis marker we can push the remaining chunk
-                Chunk::Redaction { .. } => {
```
jjbayer: I don't understand why this branch disappeared.

jjbayer (commented):
Looks good, couple of questions.

relay-event-normalization/src/trimming.rs
iker-barriocanal: It's back in [6c5e0d7](https://github.com/getsentry/relay/pull/2887/commits/6c5e0d7923dc75ff1d53046fd1f5adb9cd2839a0).

relay-event-normalization/src/trimming.rs
iker-barriocanal: They're back in [6c5e0d7](https://github.com/getsentry/relay/pull/2887/commits/6c5e0d7923dc75ff1d53046fd1f5adb9cd2839a0).

relay-event-derive/src/lib.rs
iker-barriocanal: Applied in [6c5e0d7](https://github.com/getsentry/relay/pull/2887/commits/6c5e0d7923dc75ff1d53046fd1f5adb9cd2839a0).

relay-event-schema/src/processor/attrs.rs
```diff
@@ -91,76 +91,6 @@ relay_common::derive_fromstr_and_display!(ValueType, UnknownValueTypeError, {
     ValueType::StackMemory => "stack_memory",
 });
 
-/// The maximum length of a field.
-#[derive(Debug, Clone, Copy, PartialEq, Hash)]
-pub enum MaxChars {
```
jjbayer, line 96: nit: Would be kind of nice to still have a `MaxChars` object, with `value` or and `allowance` fields and a `hard_limit()` method.

That would also help get around the `.unwrap_or(0)`

jjbayer (approved):
Have a suggestion to keep a minimal `MaxChars` struct, but other than that this looks good!

relay-event-schema/src/processor/attrs.rs
iker-barriocanal: Done in [a29758a](https://github.com/getsentry/relay/pull/2887/commits/a29758a9d88ef8af4ec94cc29c702e165fdb6303).

PR #2888, opened 2023-12-21
ref(server): Rate limit metrics in project cache

jan-auer:
Moves rate limiting of metric buckets from cached rate limits and static quotas
into the project cache. This avoids messages to the processor in the case that a
project is rate limited.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 use std::error::Error;
 use std::io::Write;
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


relay-server/src/actors/processor.rs
```diff
@@ -1356,88 +1351,62 @@ impl EnvelopeProcessorService {
...
-            // Check with redis if the throughput limit has been exceeded, while also updating
-            // the count so that other relays will be updated too.
-            match rate_limiter.is_rate_limited(quotas, item_scoping, total_buckets, false) {
-                Ok(limits) if limits.is_limited() => {
-                    relay_log::debug!(
-                        "dropping {total_buckets} buckets due to throughput ratelimit"
-                    );
+        // Check with redis if the throughput limit has been exceeded, while also updating
+        // the count so that other relays will be updated too.
+        match rate_limiter.is_rate_limited(quotas, item_scoping, quantities.buckets, false) {
```
jan-auer, line 1381: Note to reviewers: Look at this without whitespace diffs.

relay-server/src/actors/processor.rs
```diff
@@ -1272,7 +1267,7 @@ impl EnvelopeProcessorService {
         let scoping = *bucket_limiter.scoping();
 
         if let Some(rate_limiter) = self.inner.rate_limiter.as_ref() {
-            let item_scoping = ItemScoping {
+            let item_scoping = relay_quotas::ItemScoping {
```
jjbayer, line 1270: Why is `Scoping` imported, but `ItemScoping` isn't?

jjbayer (approved):
Nice!

relay-server/src/actors/processor.rs
jan-auer: It's used exclusively with `#[cfg(feature = "processing")]`, so the local qualifier was the simplest fix. We could also do a conditional import or a local import

jjbayer: Ah, right.

PR #2890, opened 2023-12-21
ref(server): Separate methods for encoding metrics

jan-auer:
This introduces two methods for flushing metric buckets:

1. The version for external and PoP Relays that performs partitioning,
   batching, and sends envelopes.
2. The version for processing Relays that performs cardinality limiting,
   rate limiting, and sends kafka messages directly. Batching occurs in
   the store forwarder.

To submit metric buckets to kafka more efficiently, there's a new
message `StoreMetrics` that can take in-memory buckets. It splits
buckets internally so that the caller doesn't have to take care of that
anymore. This is a fire-and-forget message that does not require the
caller to await the result.

The `StoreForwarder` therefore is now responsible to log outcomes for
metric buckets that cannot be produced.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/processor.rs
+++ b/relay-server/src/actors/processor.rs
@@ -10,7 +10,6 @@ use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use flate2::write::{GzEncoder, ZlibEncoder};
 use flate2::Compression;
-use itertools::Either;
 use relay_base_schema::project::{ProjectId, ProjectKey};
 use relay_common::time::UnixTimestamp;
 use relay_config::{Config, HttpEncoding};
@@ -38,12 +37,12 @@ use tokio::sync::Semaphore;
 #[cfg(feature = "processing")]
 use {
     crate::actors::project_cache::UpdateRateLimits,
+    crate::actors::store::Store,
     crate::utils::{EnvelopeLimiter, ItemAction, MetricsLimiter},
     relay_cardinality::{CardinalityLimiter, RedisSetLimiter, SlidingWindow},
     relay_metrics::{Aggregator, RedisMetricMetaStore},
     relay_quotas::{RateLimitingError, RedisRateLimiter},
     relay_redis::RedisPool,
-    std::collections::BTreeMap,
     symbolic_unreal::{Unreal4Error, Unreal4ErrorKind},
 };
 
@@ -596,6 +595,8 @@ struct InnerProcessor {
     metric_meta_store: Option<RedisMetricMetaStore>,
     #[cfg(feature = "processing")]
     cardinality_limiter: Option<CardinalityLimiter>,
+    #[cfg(feature = "processing")]
+    store_forwarder: Option<Addr<Store>>,
 }
 
 impl EnvelopeProcessorService {
@@ -610,6 +611,7 @@ impl EnvelopeProcessorService {
         upstream_relay: Addr<UpstreamRelay>,
         test_store: Addr<TestStore>,
         #[cfg(feature = "processing")] aggregator: Addr<Aggregator>,
+        #[cfg(feature = "processing")] store_forwarder: Option<Addr<Store>>,
     ) -> Self {
         let geoip_lookup = config.geoip_path().and_then(|p| {
             match GeoIpLookup::open(p).context(ServiceError::GeoIp) {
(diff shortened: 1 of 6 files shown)


relay-server/src/actors/store.rs
```diff
@@ -254,6 +281,38 @@ impl StoreService {
...
+                    MetricKafkaMessage {
+                        org_id: scoping.organization_id,
+                        project_id: scoping.project_id,
+                        name: bucket.name(),
+                        value: bucket.value(),
+                        timestamp: bucket.timestamp(),
+                        tags: bucket.tags(),
+                        retention_days: retention,
+                    },
+                )?;
```
jan-auer: We should handle the error here and log outcomes.

relay-server/src/actors/store.rs
```diff
@@ -78,35 +82,65 @@ pub struct StoreEnvelope {
     pub scoping: Scoping,
 }
 
+/// TODO
+#[derive(Clone, Debug)]
+pub struct StoreMetrics {
+    pub buckets: Vec<Bucket>,
+    pub scoping: Scoping,
+    pub retention: u16,
+    pub mode: ExtractionMode,
```
jan-auer, line 91: The store actor now logs outcomes internally for metric buckets. I think this is generally desirable, as this means we do not have to await the message sent in. Ideally, I'd like to do the same for envelopes.

However, this means we have to add things like the `ExtractionMode` now to the message so that it can accurately track how much was dropped. 

We could also just send SourceQuantities in, but then we do not know the precise number of transactions or profiles dropped if only some metrics don't make it in. This is not worse than the Envelope path right now, as that path doesn't account for errors of individual buckets either.

We basically have a choice now:
1. Keep the current implementation with the added fields and logic
2. Make it simpler again and log the total quantities of all buckets if some buckets fail to produce. It is very likely that such errors are not isolated to single messages, so likely either all or no messages fail.

relay-server/src/actors/processor.rs
```diff
@@ -1505,6 +1511,15 @@ impl EnvelopeProcessorService {
         }
     }
 
+    fn handle_encode_metrics(&self, message: EncodeMetrics) {
+        #[cfg(feature = "processing")]
+        if let Some(ref store_forwarder) = self.inner.store_forwarder {
```
Dav1dde: Should we check explicitly for wether processing is enabled here or is this implicitly already the case through the store_forwarder being `Some`?

In any case I think there should still be an explicit check for the boolean.

relay-server/src/actors/store.rs
```diff
@@ -254,6 +288,52 @@ impl StoreService {
...
+
+        let batch_size = self.config.metrics_max_batch_size_bytes();
+        let mut dropped = SourceQuantities::default();
+        let mut error = None;
+
+        for bucket in buckets {
+            // Create a local bucket view to avoid splitting buckets unnecessarily. Since we produce
+            // each bucket separately, we only need to split buckets that exceed the size, but not
+            // batches.
+            for view in BucketsView::new(&[bucket]).by_size(batch_size).flatten() {
```
Dav1dde, line 307: nit: Could hide this behind `bucket.split(max_size)`, which returns the iterator.

relay-server/src/actors/processor.rs
```diff
@@ -1505,6 +1511,15 @@ impl EnvelopeProcessorService {
         }
     }
 
+    fn handle_encode_metrics(&self, message: EncodeMetrics) {
```
Dav1dde, line 1533: I think some docs or a short comment would be nice here, why we can split logic in two different methods.

E.g. the processing one:
1. no partitioning required
2. we only have to do rate limits here
3. we only have to do cardinality limiting here

The PoP/non-processing one:
1. no rate limits
2. no cardinality limits
3. needs to partition

Dav1dde (approved):
Nice, this finally gets rid of the `serialize -> deserialize -> serialize into Kafka` with a bunch of overhead in between (partitioning, splitting, building envelopes ...)

relay-server/src/actors/store.rs
jan-auer: Let's leave this for a follow-up, I believe we might need such a signature anyway for the cross-org endpoint.

PR #2893, opened 2023-12-22
ref(cardinality): Add a timer for Redis call duration

Dav1dde:
#skip-changelog


olksdr (approved):

jan-auer (approved):
Please check during the deploy what overhead the timer itself introduces. We might want to back it out again after your measurement is complete.

PR #2894, opened 2023-12-22
ref(metrics): Change flush into a cross-project message

jan-auer:
The metrics aggregator service now flushes the buckets of all projects
in a single `FlushBuckets` message. Previously, this message was emitted
separately for every project.

This change allows the processor to batch buckets from multiple projects
into a single request. This change is not yet included in this PR and
will be added in a follow-up.

#skip-changelog

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -757,6 +757,13 @@ struct Http {
     ///  - `gzip` (default): Compression using gzip.
     ///  - `br`: Compression using the brotli algorithm.
     encoding: HttpEncoding,
+    /// Submit metrics globally through a shared endpoint.
+    ///
+    /// As opposed to regular envelopes which are sent to an endpoint inferred from the project's
+    /// DSN, this submits metrics to the global endpoint with Relay authentication.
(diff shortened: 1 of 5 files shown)


relay-config/src/config.rs
```diff
@@ -757,6 +757,13 @@ struct Http {
     ///  - `gzip` (default): Compression using gzip.
     ///  - `br`: Compression using the brotli algorithm.
     encoding: HttpEncoding,
+    /// Submit metrics globally through a shared endpoint.
+    ///
+    /// As opposed to regular envelopes which are sent to an endpoint inferred from the project's
+    /// DSN, this submits metrics to the global endpoint with Relay authentication.
+    ///
+    /// This option does not have any effect on processing mode.
+    global_metrics: bool,
```
jan-auer, line 766: This option is still inactive. Originally, I was going to implement the functionality in the same PR. Let me know if you prefer if I move this to the next PR.

iker-barriocanal (approved):

Dav1dde (approved):

PR #2895, opened 2023-12-22
test(spooler): Fix healthcheck flaky test

### LINKED ISSUE #2883: [Flaky CI]: test_healthchecks.py::test_readiness_disk_spool
### Flakiness Type

Assertion failure

### Name of Test

test_healthchecks.py::test_readiness_disk_spool

### Link to Test Run

https://github.com/getsentry/relay/actions/runs/7285837429/job/19853444134

### Details

_No response_

iker-barriocanal:
The test attempted to have a broken config to not dequeue envelopes, in a key with an `ErrorBoundary`. This PR sets a broken config in a key without an `ErrorBoundary` to make effective the no-dequeuing.

### Why was it flaky?

The config was not effectively broken for Relay due to the `ErrorBoundary`, thus Relay attempted dequeuing.

As the test sets a maximum of 0 for the memory and disk, Relay considers these always full, and transitions between `MemoryFileStandby` (memory+file) and `Disk` states constantly. In combination with the above, a race condition happens: if Tokio runs spooler jobs first, Relay will have time to make two state transitions and end up at `MemoryFileStandby` for handling the last health readiness check; but Relay will only transition once if Tokio runs the project cache jobs first.

`MemoryFileStandby` and `Disk` handle readiness checks differently: `Disk` reports ready if the disk is not full, but other states always report ready even if the memory is full too. (see https://github.com/getsentry/relay/issues/2896)

Resolves https://github.com/getsentry/relay/issues/2883.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/spooler/mod.rs
+++ b/relay-server/src/actors/spooler/mod.rs
@@ -653,7 +653,7 @@ impl BufferState {
         }
     }
 
-    /// Becomes a different state, depdending on the current state and the current conditions of
+    /// Becomes a different state, depending on the current state and the current conditions of
     /// underlying spool.
     async fn transition(self, config: &Arc<Config>) -> Self {
         match self {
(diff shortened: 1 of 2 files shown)


Dav1dde (approved):
🥳

PR #2899, opened 2023-12-22
chore(trace): Add extra "data" field to the trace context

olksdr:
Adds extra `data` field to the trace context with nested types for better schema definition. 

#skip-changelog

### DIFF
--- a/relay-event-schema/src/protocol/contexts/trace.rs
+++ b/relay-event-schema/src/protocol/contexts/trace.rs
@@ -118,6 +118,10 @@ pub struct TraceContext {
     /// This flag only applies to events with [`Error`] type that have an associated dynamic sampling context.
     pub sampled: Annotated<bool>,
 
+    /// Arbitrary additional data on a trace.
+    #[metastructure(pii = "true")]
+    pub data: Annotated<Object<Value>>,
+
     /// Additional arbitrary fields for forwards compatibility.
     #[metastructure(additional_properties, retain = "true", pii = "maybe")]
     pub other: Object<Value>,
@@ -170,6 +174,9 @@ mod tests {
   "exclusive_time": 0.0,
   "client_sample_rate": 0.5,
   "origin": "auto.http",
+  "data": {
+    "route": "/path"
+  },
   "other": "value",
   "type": "trace"
 }"#;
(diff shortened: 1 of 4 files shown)


relay-event-schema/src/protocol/contexts/trace.rs
```diff
@@ -118,6 +118,10 @@ pub struct TraceContext {
     /// This flag only applies to events with [`Error`] type that have an associated dynamic sampling context.
     pub sampled: Annotated<bool>,
 
+    /// Arbitrary additional data on a trace.
+    #[metastructure(pii = "true")]
```
jjbayer:
```suggestion
    #[metastructure(pii = "true", bag_size = "medium")]
    #[metastructure(skip_serialization = "empty")]
```

I don't know if this is the right bag size, but I noticed other data bags have it + the `skip_serialization` attr.

jjbayer (approved):

relay-event-schema/src/protocol/contexts/trace.rs
olksdr: @jjbayer  changed the annotations in [e2a23c3](https://github.com/getsentry/relay/pull/2899/commits/e2a23c3dc2a355701135cdd6fd749a10641cb76d)

iker-barriocanal (commented):
Do we know what lives in `data`, across all SDKs? Could this result in false positives (scrubbing what shouldn't be) that degrade the product experience?

olksdr:
@jjbayer  PTAL! 
Added explicit types for the routes in the `data` field.

relay-event-schema/src/protocol/contexts/trace.rs
```diff
@@ -118,11 +118,58 @@ pub struct TraceContext {
...
+    pub other: Object<Value>,
+}
+
+/// The route in the application, set by React Native SDK.
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct Route {
+    /// Parameters assigned to this route.
+    #[metastructure(pii = "true", skip_serialization = "empty", bag_size = "medium")]
+    params: Annotated<Object<Value>>,
```
jjbayer, line 161: Not 100% sure if a `pii=true` child of a `pii=maybe` parent gets scrubbed automatically (I hope so). Should we add a test to `relay-pii` for an example input like `{"route": {"params": {"password": "pw-leaked"}}}`.

jjbayer (approved):

relay-event-schema/src/protocol/contexts/trace.rs
olksdr: yep, added test in [a325880](https://github.com/getsentry/relay/pull/2899/commits/a325880608e6552739f796d6889a67e36a3a2a1a), seems like scrubbing is working by default as expected.

PR #2900, opened 2023-12-22
ref(cardinality): Optimize iterator for only rejections or all accepted

Dav1dde:
Optimization tested by tracking allocations via a custom allocator: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=a512f449788d9ca8885e299045270b5d

#skip-changelog

### DIFF
--- a/relay-cardinality/src/limiter.rs
+++ b/relay-cardinality/src/limiter.rs
@@ -138,6 +138,7 @@ impl<T: Limiter> CardinalityLimiter<T> {
 }
 
 /// Result of [`CardinalityLimiter::check_cardinality_limits`].
+#[derive(Debug)]
 pub struct CardinalityLimits<T> {
     source: Vec<T>,
     rejections: BTreeSet<usize>,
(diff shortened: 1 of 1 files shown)


jjbayer (approved):

relay-cardinality/src/limiter.rs
```diff
@@ -159,10 +161,19 @@ pub struct Accepted<T> {
 
 impl<T> Accepted<T> {
     fn new(source: Vec<T>, rejections: BTreeSet<usize>) -> Self {
-        Self {
-            source: source.into_iter(),
-            rejections: rejections.into_iter().peekable(),
-            next_index: 0,
+        if rejections.len() == source.len() {
```
iker-barriocanal, line 164: This checks whether both have the same length and assumes it's always 0. How does this code behave when `source.len() == 1` and `rejections.len() == 1`?

Dav1dde: When we have the same amount of items and rejections it means all items have been rejected and we have 0 items that are accepted -> short circuit with empty.

PR #2902, opened 2023-12-22
feat(server): Send metrics via the global endpoint

jan-auer:
Introduces an option `http.global_metrics`. When enabled and not in
processing mode, Relay sends metrics to the global batch endpoint at
`/api/0/relays/metrics/` instead of envelopes. This endpoint allows for
batched submission of metrics from multiple projects, which should
reduce the overall number of requests.

### Bug Fixes

This change contains additional bug fixes that were discoverd during
implementation:

- The batch metrics endpoint no longer requires the `emit_outcomes`
  outcomes flag to be set. This was invalid copy & paste from the
  outcomes endpoint.
- Requests with HTTP encoding received a signature computed from the
  compressed body. However, Relay requires the signature on the
  uncompressed body.

### Details

Building the request occurs in the `EnvelopeProcessor` in place of
building envelopes in the following steps:

1.  Iterate buckets from all projects and stream them into partitions.
2.  While collecting, check the size of the partition. As soon as a
    partition reaches the batch size limit, flush the partition eagerly.
    Buckets at the border may be split.
3.  At the end flush all remaining partitions with data.
4.  To flush a partition, JSON-encode the list of buckets and optionally
    apply HTTP encoding (compression).
5.  Submit the `SendMetricsRequest` with the payload and outcome
    metadata directly to the upstream.
6.  In the response callback, log outcomes directly. Therefore, the
    request does not have to be awaited.

In processing mode, Relay still produces to Kafka. The configuration
option does not have an effect in processing mode.

A note on stability: This endpoint and functionality is meant for
operation at scale within a distributed Sentry installation. At this
moment, it is not recommended to enable this option for external Relays.

### Tasks

- [x] PR description
- [x] Changelog entry with option
- [x] Integration test
- [x] Doc comments
- [x] **BUG**: Batch metrics endpoint checks outcomes flag
- [x] **BUG**: Invalid signature applied to compressed requests

### Possible Improvements

The changes in this PR allow for further optimization and changes. In
order to keep the scope of this PR to a necessary minimum, these have
not been included yet:

- Submit regular envelopes directly from the `EnvelopeProcessor`,
  avoiding the roundtrip through the `EnvelopeManager`. This also avoids
  a redundant call back to compress envelopes in the processor, should
  HTTP encoding be enabled.
- Improve handling of the `ExtractionMode`. It can be accessed more
  conveniently through a getter. Additionally, it resides on project
  config where it should actually be included in global config.
  Currently it is being passed around in excess.

### DIFF
--- a/relay-metrics/src/aggregator.rs
+++ b/relay-metrics/src/aggregator.rs
@@ -71,13 +71,9 @@ impl BucketKey {
     ///
     /// This is used for partition key computation and statsd logging.
     fn hash64(&self) -> u64 {
-        BucketKeyRef {
-            project_key: self.project_key,
-            timestamp: self.timestamp,
-            metric_name: &self.metric_name,
-            tags: &self.tags,
-        }
-        .hash64()
+        let mut hasher = FnvHasher::default();
+        std::hash::Hash::hash(self, &mut hasher);
+        hasher.finish()
     }
 
     /// Estimates the number of bytes needed to encode the bucket key.
@@ -94,29 +90,6 @@ impl BucketKey {
             Ok(mri) => mri.namespace,
             Err(_) => MetricNamespace::Unsupported,
         }
-    }
-}
-
-/// Pendant to [`BucketKey`] for referenced data, not owned data.
-///
-/// This makes it possible to compute a hash for a [`Bucket`]
-/// without destructing the bucket into a [`BucketKey`].
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-struct BucketKeyRef<'a> {
-    project_key: ProjectKey,
-    timestamp: UnixTimestamp,
-    metric_name: &'a str,
-    tags: &'a BTreeMap<String, String>,
-}
-
-impl<'a> BucketKeyRef<'a> {
-    /// Creates a 64-bit hash of the bucket key using FnvHasher.
-    ///
-    /// This is used for partition key computation and statsd logging.
-    fn hash64(&self) -> u64 {
-        let mut hasher = FnvHasher::default();
-        std::hash::Hash::hash(self, &mut hasher);
-        hasher.finish()
     }
 }
 
@@ -886,36 +859,6 @@ impl fmt::Debug for Aggregator {
             .field("receiver", &format_args!("Recipient<FlushBuckets>"))
             .finish()
     }
-}
-
-/// Splits buckets into N logical partitions, determined by the bucket key.
-pub fn partition_buckets(
-    project_key: ProjectKey,
-    buckets: impl IntoIterator<Item = Bucket>,
-    flush_partitions: Option<u64>,
-) -> BTreeMap<Option<u64>, Vec<Bucket>> {
-    let flush_partitions = match flush_partitions {
-        None => return BTreeMap::from([(None, buckets.into_iter().collect())]),
-        Some(x) => x.max(1), // handle 0,
-    };
-    let mut partitions = BTreeMap::<_, Vec<Bucket>>::new();
-    for bucket in buckets {
-        let key = BucketKeyRef {
-            project_key,
-            timestamp: bucket.timestamp,
-            metric_name: &bucket.name,
-            tags: &bucket.tags,
-        };
-
-        let partition_key = key.hash64() % flush_partitions;
-        partitions
-            .entry(Some(partition_key))
-            .or_default()
-            .push(bucket);
-
-        relay_statsd::metric!(histogram(MetricHistograms::PartitionKeys) = partition_key);
-    }
-    partitions
 }
 
 #[cfg(test)]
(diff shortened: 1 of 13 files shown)


relay-metrics/src/aggregator.rs
```diff
@@ -71,13 +71,9 @@ impl BucketKey {
...
     /// This is used for partition key computation and statsd logging.
     fn hash64(&self) -> u64 {
-        BucketKeyRef {
-            project_key: self.project_key,
-            timestamp: self.timestamp,
-            metric_name: &self.metric_name,
-            tags: &self.tags,
-        }
-        .hash64()
+        let mut hasher = FnvHasher::default();
```
jan-auer, line 74: `BucketKeyRef` is no longer required. The hash created here and the hash used for partitioning do not have to be the same, which is why we can hash differently.

relay-server/src/statsd.rs
```diff
@@ -158,6 +158,12 @@ pub enum RelayHistograms {
     /// Size of queries (projectconfig queries, i.e. the request payload, not the response) sent by
     /// Relay over HTTP in bytes.
     UpstreamEnvelopeBodySize,
+
+    /// Distribution of flush buckets over partition keys.
+    ///
+    /// The distribution of buckets should be even.
+    /// If it is not, this metric should expose it.
+    PartitionKeys,
```
jan-auer, line 166: As partitioning is now exclusively in the processor, this metric has moved here from `relay-metrics`. Its key is still the same.

relay-metrics/src/view.rs
```diff
@@ -276,14 +276,10 @@ impl<'a> Iterator for BucketsViewBySizeIter<'a> {
...
                 SplitDecision::Split(at) => {
-                    // Only certain buckets can be split, if the bucket can't be split,
-                    // move it to the next batch.
-                    if bucket.can_split() {
-                        self.current = Index {
-                            slice: self.current.slice,
-                            bucket: self.current.bucket + at,
-                        };
-                    }
+                    self.current = Index {
```
jan-auer, line 284: The check for whether the bucket can split has been moved into `split_at`. Invariants for splitting are still checked in the inner iterator.

relay-metrics/src/view.rs
```diff
@@ -427,6 +424,46 @@ impl<'a> BucketView<'a> {
         Some(self)
     }
 
+    /// Estimates the number of bytes needed to serialize the bucket without value.
+    ///
+    /// Note that this does not match the exact size of the serialized payload. Instead, the size is
+    /// approximated through tags and a static overhead.
+    fn estimated_base_size(&self) -> usize {
+        50 + self.name().len() + aggregator::tags_cost(self.tags())
```
jjbayer: Should we put this in a constant? For example

```suggestion
        BUCKET_OVERHEAD + self.name().len() + aggregator::tags_cost(self.tags())
```

relay-server/src/http.rs
```diff
@@ -46,28 +46,22 @@ pub struct Request(pub reqwest::Request);
...
         }
     }
 
     pub fn finish(self) -> Result<Request, HttpError> {
-        let mut builder = self.builder.unwrap();
-        if let Some(body) = self.body {
-            builder = builder.body(body);
-        }
-        Ok(Request(builder.build()?))
+        Ok(Request(self.builder.unwrap().build()?))
```
jjbayer, line 61: Why is unwrapping OK here?

relay-server/src/actors/processor.rs
```diff
@@ -1655,6 +1732,192 @@ impl Service for EnvelopeProcessorService {
...
+                        );
+                    }
+                }
+                // Upstream is responsible to log outcomes.
+                Err(_received) => (),
+            }
+        })
+    }
+}
+
```
jjbayer, line 1931: There's a lot of metrics-specific methods here now, maybe we should move them to a submodule (as standalone functions) in a follow-up PR.

relay-server/src/actors/processor.rs
```diff
@@ -1540,6 +1528,91 @@ impl EnvelopeProcessorService {
...
+                        .entry(partition_key)
+                        .or_insert_with(|| Partition::new(batch_size, mode));
+
+                    if let Some(next) = partition.insert(bucket, *scoping) {
+                        // A part of the bucket could not be inserted. Take the partition and submit
+                        // it immediately. Repeat until the final part was inserted. This should
+                        // always result in a request, otherwise we would enter an endless loop.
+                        self.send_global_partition(partition_key, partition);
+                        remaining = Some(next);
+                    }
```
jjbayer, line 1611: Do we have test coverage for this case?

jjbayer (approved):
See comment about test coverage, apart from that LGTM!

relay-server/src/http.rs
jan-auer: The builder will never be set to `None`. The option is there to allow moving reqwest's inner builder in the builder methods. Adding a comment to clarify this.

jjbayer: 👍 , might be worth a code comment!

relay-server/src/actors/processor.rs
jan-auer: Added in 36159cce4

jjbayer (approved):

PR #2903, opened 2024-01-02
chore(beta): Address new beta lints

olksdr:
This [lint](https://rust-lang.github.io/rust-clippy/master/index.html#/test_attr_in_doctest) is producing warning (error in our case).  
Let's `ignore` the doc tests with `#[test]` attribute, in our case those used just as an example. 

#skip-changelog

### DIFF
--- a/relay-log/src/lib.rs
+++ b/relay-log/src/lib.rs
@@ -98,7 +98,7 @@
 //! at the beginning of test method. It enables test mode of the logger and customizes log levels
 //! for the current crate.
 //!
-//! ```
+//! ```ignore
 //! #[test]
 //! fn test_something() {
 //!     relay_log::init_test!();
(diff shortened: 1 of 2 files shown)


relay-log/src/lib.rs
```diff
@@ -98,7 +98,7 @@
 //! at the beginning of test method. It enables test mode of the logger and customizes log levels
 //! for the current crate.
 //!
-//! ```
+//! ```ignore
```
jjbayer:
```suggestion
//! ```no_run
```

No strong preference, but following the [docs](https://rust-lang.github.io/rust-clippy/master/index.html#/test_attr_in_doctest): 

> If you really want to show how to test stuff in an example, mark it no_run to make the intent clear.

jjbayer (approved):

PR #2904, opened 2024-01-02
ref(metrics): Better align flush times with intervals

jan-auer:
Align the flush times of backdated metric buckets with bucket intervals
so that global flushing becomes more effective. This is done through two
changes:

1. Allow to configure `ShiftKey::None` to disable shifting of flushed
   buckets. If set to `none`, buckets from all projects are flushed at
   the same time.
2. For backdated buckets, do not use the precise current timestamp.
   Instead, round it to the current bucket interval, then add the
   debounce delay. When shifting is configured, the shift is now also
   added to backdated buckets.

Integration tests by default use `shift_key: none` now to improve their
execution time, which also helps reduce flakiness.

#skip-changelog

### DIFF
--- a/relay-common/src/time.rs
+++ b/relay-common/src/time.rs
@@ -61,20 +61,6 @@ pub fn chrono_to_positive_millis(duration: chrono::Duration) -> f64 {
     duration_to_millis(duration.to_std().unwrap_or_default())
 }
 
-/// The conversion result of [`UnixTimestamp::to_instant`].
-///
-/// If the time is outside of what can be represented in an [`Instant`], this is `Past` or
-/// `Future`.
-#[derive(Clone, Copy, Debug)]
-pub enum MonotonicResult {
-    /// A time before the earliest representable `Instant`.
-    Past,
-    /// A representable `Instant`.
-    Instant(Instant),
-    /// A time after the latest representable `Instant`.
-    Future,
-}
-
 /// A unix timestamp (full seconds elapsed since 1970-01-01 00:00 UTC).
 #[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct UnixTimestamp(u64);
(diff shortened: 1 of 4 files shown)


relay-metrics/src/aggregator.rs
```diff
@@ -696,26 +691,27 @@ impl Aggregator {
...
+        let bucket_ts = self.config.get_bucket_timestamp(timestamp, bucket_width);
+
+        if !self.config.timestamp_range().contains(&bucket_ts) {
+            let delta = (bucket_ts.as_secs() as i64) - (UnixTimestamp::now().as_secs() as i64);
             relay_statsd::metric!(
                 histogram(MetricHistograms::InvalidBucketTimestamp) = delta as f64,
                 aggregator = &self.name,
             );
+
+            return Err(AggregateMetricsErrorKind::InvalidTimestamp(timestamp).into());
```
jjbayer, line 714: Nice, this is better.

jjbayer (approved):
Not sure about the failing test, IMO it should not have passed even before this change.

olksdr (approved):

relay-metrics/src/aggregatorservice.rs
```diff
@@ -498,40 +497,4 @@ mod tests {
         // receiver must have 1 bucket flushed
         assert_eq!(receiver.bucket_count(), 1);
     }
-
-    #[tokio::test]
-    async fn test_merge_back() {
```
jan-auer, line 503: Turns out this test was entirely broken. The regular flush cycle is tested by the function above, this one wasn't testing anything anymore.

relay-common/src/time.rs
```diff
@@ -133,41 +119,6 @@ impl UnixTimestamp {
...
-    /// ```
-    /// use std::time::{Duration, Instant};
-    /// use relay_common::time::{MonotonicResult, UnixTimestamp};
-    ///
-    /// let timestamp = UnixTimestamp::now();
-    /// if let MonotonicResult::Instant(instant) = timestamp.to_instant() {
-    ///    assert!((Instant::now() - instant) < Duration::from_millis(1));
-    /// }
-    /// ```
-    pub fn to_instant(self) -> MonotonicResult {
```
jan-auer, line 156: This was used in exactly one place, so we inlined it instead.

PR #2905, opened 2024-01-03
ref(store): Split buckets directly for kafka

jan-auer:
Minor refactor that changes how buckets are split before producing them to Kafka. Previously, the store service constructed a `BucketsView` with a single bucket, that it then splits and iterates. As the `BucketView` now has a direct split method, we can skip the iterator and split the view directly.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/store.rs
+++ b/relay-server/src/actors/store.rs
@@ -15,7 +15,7 @@ use relay_event_schema::protocol::{
 };
 use relay_kafka::{ClientError, KafkaClient, KafkaTopic, Message};
 use relay_metrics::{
-    Bucket, BucketViewValue, BucketsView, MetricNamespace, MetricResourceIdentifier,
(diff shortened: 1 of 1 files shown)


relay-server/src/actors/store.rs
```diff
@@ -301,10 +301,13 @@ impl StoreService {
         let mut error = None;
 
         for bucket in buckets {
-            // Create a local bucket view to avoid splitting buckets unnecessarily. Since we produce
-            // each bucket separately, we only need to split buckets that exceed the size, but not
-            // batches.
-            for view in BucketsView::new(&[bucket]).by_size(batch_size).flatten() {
+            let mut remaining = Some(BucketView::new(&bucket));
```
jan-auer, line 304: Introducing a new iterator just for this one usage didn't seem worth the abstraction. What do you think of this contraption?

olksdr (approved):

PR #2906, opened 2024-01-03
feat(app-start): Add duration metric for top-level app start spans

narsaynorath:
Adds a new MetricSpec and conditions for retrieving tags for a duration metric.

For nested app start spans we can use self-time, but at the top-level we need to expose duration and filter by op and description (e.g. `app.start.cold` and `Cold Start`)


relay-server/src/metrics_extraction/event.rs
```diff
@@ -1002,6 +1002,23 @@ mod tests {
                         "resource.render_blocking_status": "blocking"
                     },
                     "hash": "e2fae740cccd3789"
+                },
+                {
+                    "timestamp": 1694732408.3145,
```
narsaynorath: @jjbayer I ran `cargo insta review` to try to regenerate snapshots, but it didn't seem to produce anything. Could this be because my test case is not fulfilling the proper conditions?

narsaynorath:
Should this have a `CHANGELOG.md` entry?

jjbayer (approved):
Nice, thanks for contributing!

Issue #2910, opened 2024-01-04
Remove invalid values from distributions


jan-auer:
Distribution buckets containing `f64::INFINITY` (which can be the result of a division by zero) or `f64::NAN` serialize with `null` values to JSON. This creates errors in storage. Relay should drop such values before serializing.

We still have to determine where these values originate. If they are the result of erroneous computations in metrics extraction, the original bug needs to be fixed in addition. It is not impossible to ingest these values since `null` does not deserialize to `f64`.

See [SNUBA-426](https://sentry.sentry.io/issues/4747328935/)

jan-auer:
There are three options to fix this:

- https://github.com/getsentry/relay/pull/2943
- https://github.com/getsentry/relay/pull/2958
- https://github.com/getsentry/relay/pull/2957

jan-auer:
We decided to move ahead with https://github.com/getsentry/relay/pull/2958, introducing a strictly typed `FiniteF64` type.

jan-auer:
The downstream bug is mitigated, as no more `null` values are produced. 

To track the root cause of the invalid values, we'll start to log Sentry errors during metric extraction. https://github.com/getsentry/relay/pull/2963

jan-auer:
Closing this, as @edwardgou-sentry is looking into fixing performance scores now.

PR #2911, opened 2024-01-04
fix(normalize): Do not divide by zero in perf scores

jan-auer:
Resolve a division by zero in performance score computation that leads to dropped metrics for transactions. This could occur if either all components are optional, or the weight of all components is `0`.

Potential fix for [SNUBA-426](https://sentry.sentry.io/issues/4747328935/)


olksdr (approved):
Makes sense to me

PR #2914, opened 2024-01-05
fix(ddm): Fix the MRI normalization of metrics summaries

iambriccardo:
This PR normalizes all incoming metric identifiers for the metrics in the `_metrics_summary` by adding the default `custom` namespace.

Note that this normalization **won't** be required when the `metrics_summary` is properly typed. This is because when an MRI is deserialized and then serialized, Relay will automatically add `custom` if the namespace is not specified (assuming all spans need to go through one deser/ser step).


relay-event-normalization/Cargo.toml
```diff
@@ -22,6 +22,7 @@ relay-base-schema = { path = "../relay-base-schema" }
 relay-common = { path = "../relay-common" }
 relay-event-schema = { path = "../relay-event-schema" }
 relay-log = { path = "../relay-log" }
+relay-metrics = { path = "../relay-metrics" }
```
iambriccardo: We might want to export `MetricResourceIdentifier` to avoid introducing this dependency.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -2358,6 +2401,145 @@ mod tests {
...
+            spans: Annotated::new(vec![Annotated::new(Span {
+                op: Annotated::new("my_span".to_owned()),
+                _metrics_summary: Annotated::new(Value::Object(metrics_summary.clone())),
+                ..Default::default()
+            })]),
+            _metrics_summary: Annotated::new(Value::Object(metrics_summary)),
+            ..Default::default()
+        };
+        normalize_all_metrics_summaries(&mut event);
+        assert_debug_snapshot!(event, @r###"
```
Dav1dde: Should only assert the summaries subsets (spans and event_summaries) to make it shorter. We also have `assert_json_snapshot`.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -2358,6 +2401,145 @@ mod tests {
...
+        metrics_summary.insert(
+            "c:page_click@none".to_string(),
+            Annotated::new(Value::Array(Vec::new())),
+        );
+        metrics_summary.insert(
+            "s:user@none".to_string(),
+            Annotated::new(Value::Array(Vec::new())),
+        );
+        metrics_summary.insert(
+            "g:page_load@second".to_string(),
```
Dav1dde, line 2430: Maybe also add a single value with a non default namespace and assert that it is kept.

There is also the question if the namespace here has to be `custom` always.

relay-event-normalization/Cargo.toml
jjbayer: Could you move MRI to `relay-base-schema/metrics` instead?

iambriccardo: I could, I would expect it to be better to make a separate PR for doing so. WDYT?

jjbayer: Sounds good!

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -207,6 +210,46 @@ fn normalize_app_start_spans(event: &mut Event) {
...
+    *value = Value::Object(metrics);
+}
+
+/// Normalizes all the metrics summaries across the event payload.
+fn normalize_all_metrics_summaries(event: &mut Event) {
+    if let Some(metrics_summary) = event._metrics_summary.value_mut().as_mut() {
+        normalize_metrics_summary_mris(metrics_summary)
+    }
+
+    if let Some(spans) = event.spans.value_mut() {
```
Dav1dde: Early return with `let .. else` would make this less indentation heavy.

```rs
    let Some(spans) = event.spans.value_mut() else {
        return;
    };
```

Could also consider combining the inner `if let`'s:

```rs
    let Some(spans) = event.spans.value_mut() else {
        return;
    };

    for span in spans.iter_mut() {
        let metrics_summary = span
            .value_mut()
            .as_mut()
            .and_then(|span| span._metrics_summary.value_mut().as_mut());

        if let Some(ms) = metrics_summary {
            normalize_metrics_summary_mris(ms)
        }
    }
```

relay-event-normalization/Cargo.toml
Dav1dde: https://github.com/getsentry/relay/pull/2921

relay-event-normalization/src/normalize/mod.rs
jjbayer: +1

Dav1dde (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Introduce the configuration option `http.global_metrics`. When enabled, Relay submits metric buckets not through regular project-scoped Envelopes, but instead through the global endpoint. When this Relay serves a high number of projects, this can reduce the overall request volume. ([#2902](https://github.com/getsentry/relay/pull/2902))
 - Record the size of global metrics requests in statsd as `upstream.metrics.body_size`. ([#2908](https://github.com/getsentry/relay/pull/2908))
 - Only extract span metrics / tags when they are needed. ([#2907](https://github.com/getsentry/relay/pull/2907))
+- Normalize `_metrics_summary` metric identifiers. ([#2914](https://github.com/getsentry/relay/pull/2914))
```
```suggestion
- Normalize metric resource identifiers in `event._metrics_summary` and `span._metrics_summary`. ([#2914](https://github.com/getsentry/relay/pull/2914))
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -116,6 +118,7 @@ impl<'a> StoreNormalizeProcessor<'a> {
     fn normalize_spans(&self, event: &mut Event) {
         if event.ty.value() == Some(&EventType::Transaction) {
             normalize_app_start_spans(event);
+            normalize_all_metrics_summaries(event);
```
jjbayer: This function also modifies `event._metrics_summary`, so we should move it out of `normalize_spans` IMO.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -1,9 +1,11 @@
 use std::hash::Hash;
+use std::mem;
 use std::sync::Arc;
 
 use itertools::Itertools;
 use once_cell::sync::OnceCell;
 use regex::Regex;
+use relay_base_schema::metrics::MetricResourceIdentifier;
 use relay_base_schema::metrics::MetricUnit;
```
Dav1dde:
```suggestion
use relay_base_schema::metrics::{MetricUnit, MetricResourceIdentifier};
```

And remove the import above

jjbayer (commented):
I would personally rather type out the `_metrics_summary` field instead of writing a custom normalization function for it (can still be "forgiving", for example by using an untagged enum that is either `BTreeMap<MRI, Value>` or a plain `Value`, as @Dav1dde suggested.

PR #2917, opened 2024-01-05
feat(spans): Convert a Sentry span to the Snuba span schema

phacops:
We're working on pushing spans directly to Snuba instead of going through a Sentry consumer. This will produce to Kafka the proper schema understood by the Snuba consumer.

The way I did that is a bit convoluted right now. The envelope contains a Sentry span we parse using the Span protocol which needs to then be converted to the Kafka message which will use serde. This can be simplified by parsing the span in the envelope directly into a serde-compatible struct (which means when we modify the span, we now have to modify 2 places). Do you think the trade-off is worth it?

I would have reused the annotated protocol but we then can't pass it nicely to the producer to be serialized.

It will still push to the `ingest-spans` topic until we change the config so we're going to need this PR merged and deployed before we can deploy this change: https://github.com/getsentry/sentry/pull/62527.


relay-server/src/actors/store.rs
```diff
@@ -1138,22 +1251,31 @@ struct CheckInKafkaMessage {
...
+    _metrics_summary: &'a RawValue,
+    description: String,
+    duration_ms: u32,
     /// The ID of the transaction event associated to this span, if any.
     #[serde(skip_serializing_if = "Option::is_none")]
     event_id: Option<EventId>,
-    /// The numeric ID of the organization.
-    organization_id: u64,
+    exclusive_time_ms: f64,
+    group_raw: String,
```
jjbayer: This will never be set by Relay, and it's [not required](https://github.com/getsentry/sentry-kafka-schemas/blob/fd9fc502294ff12cb206951c92408d748caba624/schemas/snuba-spans.v1.schema.json#L97-L106), so we might as well omit it. 
```suggestion
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Add `duration` metric for mobile app start spans. ([#2906](https://github.com/getsentry/relay/pull/2906))
 - Introduce the configuration option `http.global_metrics`. When enabled, Relay submits metric buckets not through regular project-scoped Envelopes, but instead through the global endpoint. When this Relay serves a high number of projects, this can reduce the overall request volume. ([#2902](https://github.com/getsentry/relay/pull/2902))
 - Record the size of global metrics requests in statsd as `upstream.metrics.body_size`. ([#2908](https://github.com/getsentry/relay/pull/2908))
+- Convert a Sentry span to the Snuba span schema. ([#2917](https://github.com/getsentry/relay/pull/2917))
```
```suggestion
- Make Kafka spans compatible with the Snuba span schema. ([#2917](https://github.com/getsentry/relay/pull/2917))
```

relay-server/src/actors/store.rs
```diff
@@ -839,13 +849,116 @@ impl StoreService {
                 return Ok(());
             }
         };
+
+        let Some(span) = annotated_span.value() else {
```
jjbayer: This should make the `.clone()`s below unnecessary.

```suggestion
        let Some(span) = annotated_span.into_value() else {
```

relay-server/src/actors/store.rs
```diff
@@ -839,13 +849,116 @@ impl StoreService {
...
+            return Ok(());
+        };
+        let Some(start_timestamp) = span.start_timestamp.clone().into_value() else {
+            return Err(StoreError::NoStartTimestamp);
+        };
+
+        let _metrics_summary = match span._metrics_summary.to_json() {
```
```suggestion
        let metrics_summary = match span._metrics_summary.to_json() {
```

relay-server/src/actors/store.rs
```diff
@@ -828,8 +839,7 @@ impl StoreService {
         retention_days: u16,
         item: &Item,
     ) -> Result<(), StoreError> {
-        let payload = item.payload();
-        let span = match serde_json::from_slice(&payload) {
+        let annotated_span = match Annotated::<Span>::from_json_bytes(&item.payload()) {
```
jjbayer: Instead of going the detour via `Annotated`, we can deserialize into `SpanKafkaMessage` directly, borrowing as much as possible and using `RawValue` as you're already doing, with `#[serde(default)]`and `#[serde(default = "fn")]` attributes for default values and `#[serde(deserialize_with = "fn"]` for the timestamps.

relay-server/src/actors/store.rs
```diff
@@ -839,13 +849,116 @@ impl StoreService {
...
+        };
+
+        let _metrics_summary = match span._metrics_summary.to_json() {
+            Ok(_metrics_summary) => match RawValue::from_string(_metrics_summary) {
+                Ok(value) => value,
+                Err(_) => Default::default(),
+            },
+            Err(_) => Default::default(),
+        };
+        let description = span.description.clone().into_value().unwrap_or_default();
```
jjbayer: nit: Might be just my personal preference, but I would inline most of these variables in the `SpanKafkaMessage`.

jjbayer (changes requested):
The integration test will need updating.

relay-server/src/actors/store.rs
```diff
@@ -1136,24 +1133,50 @@ struct CheckInKafkaMessage {
...
-    /// Time at which the event was received by Relay. Not to be confused with `start_timestamp_ms`.
-    start_time: u64,
+#[derive(Debug, Deserialize, Serialize)]
+struct SpanKafkaMessage {
+    #[serde(skip_serializing)]
+    start_timestamp: f64,
+    #[serde(rename(deserialize = "timestamp"), skip_serializing)]
+    end_timestamp: f64,
+
+    description: String,
```
jjbayer: It should be possible to make the fields borrowing, i.e. `&'de str` instead of `String`, `&'de RawValue` instead of `Value`.

relay-server/src/actors/store.rs
```diff
@@ -824,12 +825,13 @@ impl StoreService {
         organization_id: u64,
         project_id: ProjectId,
         event_id: Option<EventId>,
-        start_time: Instant,
+        _: Instant,
```
jjbayer: Might as well remove this argument, since the received time is already set on `span.received`.

relay-server/src/actors/store.rs
```diff
@@ -1136,24 +1133,53 @@ struct CheckInKafkaMessage {
...
+        default,
+        rename = "_metrics_summary",
+        skip_serializing_if = "Option::is_none"
+    )]
+    metrics_summary: Option<&'a RawValue>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    parent_span_id: Option<&'a str>,
     /// The numeric ID of the project.
-    project_id: ProjectId,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
jjbayer, line 1161: Comment above should be two lines below.

relay-server/src/actors/store.rs
```diff
@@ -1136,24 +1133,53 @@ struct CheckInKafkaMessage {
...
     retention_days: u16,
-    /// Fields from the original span payload.
-    /// See [`relay-event-schema::protocol::span::Span`] for schema.
-    ///
-    /// By using a [`RawValue`] here, we can embed the span's JSON without additional parsing.
-    span: &'a RawValue,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    segment_id: Option<&'a str>,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    sentry_tags: BTreeMap<&'a str, &'a str>,
```
jjbayer: Any reason this is not a `RawValue`?  Is it because of the `skip_serializing_if`? Could we live without `skip_serializing_if`?

tests/integration/test_store.py
```diff
@@ -1538,77 +1529,60 @@ def test_span_ingestion(
...
+        span.pop("received", None)
 
     spans.sort(
-        key=lambda msg: msg["span"].get("description", "")
+        key=lambda msg: msg.get("description", "")
     )  # endpoint might overtake envelope
 
     assert spans == [
         {
-            "organization_id": 1,
```
jjbayer, line 1550: I think we still need the organization ID, right?

jjbayer (commented):
See comment about `organization_id`. Rest looks good to me!

tests/integration/test_store.py
phacops: We don't.

phacops: No, we don't need `organization_id` anymore.

relay-server/src/actors/store.rs
```diff
@@ -1136,24 +1133,53 @@ struct CheckInKafkaMessage {
...
+    span_id: &'a str,
+    #[serde(default)]
+    start_timestamp_ms: u64,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    tags: BTreeMap<&'a str, &'a str>,
```
jjbayer: Same as above

tests/integration/test_store.py
jjbayer: Alright.

jjbayer (approved):
See comments about RawValue, but no blockers!

This definitely needs a Canary deploy, that is

1. Pause `deploy-relay`
2. Deploy `deploy-relay-experimental`
3. Let run for ~1h and watch for consumer errors.
4. If no errors and span volume is the same as before, unpause `deploy-relay` and deploy there as well.

Canary can also be reverted quickly.

tests/integration/test_store.py
phacops: Even if the field is described in the schema, the Snuba consumer doesn't need it to insert data: https://github.com/getsentry/snuba/blob/master/rust_snuba/src/processors/spans.rs#L28-L53.

PR #2919, opened 2024-01-08
fix(server): Fix early return. 

TBS1996:
This return should probably have turned to a continue in this PR: https://github.com/getsentry/relay/pull/2894/files 

it turned to a loop over different vecs of buckets and the return statement would make it skip the following vectors. 

#skip-changelog


iker-barriocanal (commented):
My brain may still be on vacation, but it seems it's iterating through the scopes of a single message and the previous PR had also a return, so this could be ok? Could you add some more context and a test case for the bug?

jjbayer (changes requested):
Could you add a test that proves the old implementation was wrong?

Dav1dde (approved):

Dav1dde:
@iker-barriocanal  Since in theory we can have different quotas per project and org and the at this point there is no guarantee that the scopings are all confined to the same org (practically they are) this would stop processing all buckets for all projects/orgs just because we hit the rate limit for one project.

@jan-auer should probably have a quick check as well.

jan-auer (approved):

TBS1996:
@jjbayer  i've added a test now. not sure how to prove the old implementation is wrong other than changing the 'continue' to 'return' and observing that the test fail

relay-server/src/actors/processor.rs
```diff
@@ -594,7 +594,7 @@ pub struct ProjectMetrics {
 /// Encodes metrics into an envelope ready to be sent upstream.
 #[derive(Debug)]
 pub struct EncodeMetrics {
-    pub scopes: HashMap<Scoping, ProjectMetrics>,
+    pub scopes: Vec<(Scoping, ProjectMetrics)>,
```
iker-barriocanal: What's the reason for this change?

relay-server/src/actors/processor.rs
```diff
@@ -2439,6 +2439,94 @@ mod tests {
...
+    /// Ensures that if we ratelimit one batch of buckets in [`EncodeMetrics`] message, it won't
+    /// also ratelimit the next batches in the same message automatically.
+    #[cfg(feature = "processing")]
+    #[tokio::test]
+    async fn test_ratelimit_per_batch() {
+        use relay_metrics::BucketValue;
+        use relay_quotas::{Quota, ReasonCode};
+        use relay_test::mock_service;
+        use std::sync::atomic::AtomicUsize;
+        use std::sync::atomic::Ordering::SeqCst;
```
iker-barriocanal: Is there a reason to not move these imports to the tests module?

relay-server/src/actors/processor.rs
```diff
@@ -2439,6 +2439,94 @@ mod tests {
...
+        let f = |x: &mut Arc<AtomicUsize>, _: Store| {
+            x.fetch_add(1, SeqCst); // counts the number of not-ratelimited batches.
+        };
+
+        let (store, handle) = mock_service("store_forwarder", received_batches.clone(), f);
+        create_test_processor(Default::default()).encode_metrics_processing(message, &store);
+        drop(store);
+        handle.await.unwrap();
+
+        let received_batches = received_batches.load(SeqCst);
```
iker-barriocanal: For extra safety: could we validate that the batch we get here is the expected above (i.e. the one with org id 0)?

iker-barriocanal (commented):
> not sure how to prove the old implementation is wrong other than changing the 'continue' to 'return' and observing that the test fail

If a test that validates a correct use case fails with the previous code but passes with the current one, it means a bug was fixed 😄

relay-server/src/actors/processor.rs
TBS1996: guaranteeing consistency when testing, the test case depends on which scoping->projectmetrics pair come first, but the hashmap doesn't guarantee order.

relay-server/src/actors/processor.rs
TBS1996: they're just used in processing mode so in non-processing i'd get unused_import warnings. could have put them outside and under a feature flag too of course, I thought both ways are fine

relay-server/src/actors/processor.rs
TBS1996: done ^^

relay-server/src/actors/processor.rs
iker-barriocanal: Could we validate that the elements we're looking for exist in the hashmap instead? Assuming this happens with the newly added test `test_ratelimit_per_batch`.

relay-server/src/actors/processor.rs
iker-barriocanal: We have some tests that only run in processing mode. Adding a test for processing relays is alright.

relay-server/src/testutils.rs
```diff
@@ -114,10 +114,18 @@ pub fn create_test_processor(config: Config) -> EnvelopeProcessorService {
     #[cfg(feature = "processing")]
     let (_aggregator, _) = mock_service("aggregator", (), |&mut (), _| {});
 
+    #[cfg(feature = "processing")]
+    let redis = {
+        let url = std::env::var("RELAY_REDIS_URL")
+            .unwrap_or_else(|_| "redis://127.0.0.1:6379".to_owned());
```
olksdr: This seems a bit sketchy. 

Can the `redis` be just defaulted to `Option`?  If the config not provided, just make it `None` and/or also produce the error from inside of the tests, to make sure that it won't get un-notice if someone changes anything.

relay-server/src/actors/processor.rs
olksdr: Does it change anything, e.g. performance by changing the underlying container `Vec` vs. `HashMap`?  Or anything else between those lines?

TBS1996: could do it, but then the test wouldn't consistently fail without this change though, not sure how important that is

Dav1dde: Could use a `BTreeMap` for consistent ordering, I'd prefer staying with the map just so it is clear the scopes don't repeat.

TBS1996: @olksdr should be faster from what i know, since we only used it to iterate  
 
@Dav1dde yeah sure, could do that

iker-barriocanal: I'd stick to the hashmap if possible. I don't think the test should impact the data structure in this case

TBS1996: how would I verify it works if i use a hashmap?

relay-server/src/testutils.rs
TBS1996: I've now changed to what i think you had in mind

relay-server/src/actors/processor.rs
TBS1996: not sure what you mean, is there another way to mark that it'll run just for processing relays?

relay-server/src/actors/processor.rs
iker-barriocanal: Verify that the map contains all the items you expect, and nothing more. For example, if you expect `{'a': 1, 'b': 2}`, assert that the map contains exactly 2 items and that those items are `'a': 1` and `'b': 2`.

relay-server/src/actors/processor.rs
iker-barriocanal: The code with a `#[cfg(feature = "processing")]` flag, including imports and tests, is only available in processing relays. A test with that flag that uses imports with also that flag should result in no unused import warnings, unless I'm missing something. It's possible to group imports with the `cfg` with something like the following:
```rs
use {
  module_a::A;
  module_b::B;
}
```

TBS1996: ok yeah, thats what I was thinking about. As I said i thought both ways are fine but i did it the way you showed now

relay-server/src/actors/processor.rs
TBS1996: without this PR the order of how the hashmap iterates over the keys matters in prod, as it'll short-circuit on the first is_rate_limited == true, so to demonstrate that it's no longer the case with 'continue' i changed it to btreemap. In general we also even prefer btreemap on small maps like this

relay-server/src/actors/processor.rs
```diff
@@ -2434,11 +2439,115 @@ mod tests {
...
+                }
+            });
+            Config::from_json_value(config_json).unwrap()
+        };
+
+        let (store, handle) = {
+            let f = |org_ids: &mut Vec<u64>, msg: Store| {
+                let org_id = match msg {
+                    Store::Metrics(x) => x.scoping.organization_id,
+                    Store::Envelope(_) => panic!(),
```
iker-barriocanal: Could we set a message in this panic? If this test detects a regression in the future, a nice error message can help scope down the cause.

relay-server/src/actors/processor.rs
```diff
@@ -2434,11 +2439,115 @@ mod tests {
...
+    };
+
     use super::*;
 
+    /// Ensures that if we ratelimit one batch of buckets in [`EncodeMetrics`] message, it won't
+    /// also ratelimit the next batches in the same message automatically.
+    #[cfg(feature = "processing")]
+    #[tokio::test]
+    async fn test_ratelimit_per_batch() {
+        let rate_limited_org = 0;
```
iker-barriocanal: IIRC we set org_id=0 to define some rate limits, so it'd be great if we don't use that ID for this test.

iker-barriocanal (approved):

relay-server/src/actors/processor.rs
TBS1996: ok, changed it ^^

jjbayer (approved):
> i've added a test now. not sure how to prove the old implementation is wrong other than changing the 'continue' to 'return' and observing that the test fail

@TBS1996 thanks, that's exactly the proof I was looking for!

PR #2921, opened 2024-01-08
ref(metrics): Move MRI into relay-base-schema module

Dav1dde:
Moves the `MetricResourceIdentifier` into the `relay-base-schema` module next to all the metric units. As a preparation for https://github.com/getsentry/relay/pull/2914 to be able to use the `MRI` either directly in the event schema or for validations.
 
#skip-changelog


relay-base-schema/src/metrics/mri.rs
```diff
@@ -0,0 +1,471 @@
...
+
+    /// Parses an MRI from a string and a separate type.
+    ///
+    /// The given string must be a part of the MRI, including the following components:
+    ///  - (optional) The namespace. If missing, it is defaulted to `"custom"`
+    ///  - (required) The metric name.
+    ///  - (optional) The unit. If missing, it is defaulted to "none".
+    ///
+    /// The metric type is never part of this string and must be supplied separately.
+    pub fn parse_with_type(string: &'a str, ty: MetricType) -> Result<Self, ParseMetricError> {
```
Dav1dde, line 244: This is no public, we need it for `Bucket`, was `pub(crate)` before

relay-base-schema/src/metrics/mri.rs
```diff
@@ -0,0 +1,471 @@
...
+            _ => return Err(ParseMetricError),
+        })
+    }
+}
+
+relay_common::impl_str_serde!(MetricType, "a metric type string");
+
+/// An error returned when metrics or MRIs cannot be parsed.
+#[derive(Clone, Copy, Debug)]
+pub struct ParseMetricError;
```
Dav1dde, line 78: Was `pub struct ParseMetricError(pub(crate) ());` but this is kinda awkward since we need the error as well in `Bucket` (`relay-metrics`)

jjbayer (approved):
Could you run `make doc-rust` to check if the docs are still readable (especially the metrics crate docs)?

iker-barriocanal (approved):
Thanks for doing this!

Dav1dde:
Docs look good 👍

PR #2923, opened 2024-01-08
fix(spans): Restore domain tag on resource spans

jjbayer:
Required in the product.


relay-dynamic-config/src/defaults.rs
```diff
@@ -539,7 +539,7 @@ fn span_metrics_reduced() -> impl IntoIterator<Item = MetricSpec> {
                     .always(),
                 Tag::with_key("span.domain")
                     .from_field("span.sentry_tags.domain")
-                    .when(is_db.clone()),
+                    .when(is_db.clone() | is_resource.clone()),
```
jjbayer, line 542: It might not be necessary on `exclusive_time` (only on `exclusive_time_light`), but it will not increase cardinality because it's derived from description.

DominikB2014 (approved):

olksdr (approved):

iker-barriocanal (approved):

PR #2926, opened 2024-01-09
fix(spans): Handle strings with escaped characters

phacops:
Force the ordering by `span_id` in the integration tests.


jjbayer (approved):
There's 4 other `&'a str` in `SpanKafkaMessage`, but they're all IDs so I hope they're fine 🔥 .

PR #2924, opened 2024-01-09
ref(spans): Reduced set of metrics/tags for all orgs

jjbayer:
Follow-up to https://github.com/getsentry/relay/pull/2907: Apply reduced set of span metrics and tags regardless of feature flag.

ref: https://github.com/getsentry/team-ingest/issues/252

PR #2927, opened 2024-01-09
feat(app-start): Add Android app start-related spans

narsaynorath:
These spans are nested operations under the app.start.* spans on Android. Start collecting them for the Sentry product so we have some data before moving into the default config.

Note: This only adds the application portion of the spans. We are currently awaiting a new span that tracks the system portion of the app start and I will add that as another span op afterwards.

These spans are listed out [here](https://www.notion.so/sentry/70c42197962a41198992ae89383ceb57?v=af03325d6d3e4168bdb13bd3269fe3f1&pvs=4) for further information:

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -82,6 +82,26 @@ pub(crate) fn scrub_span_description(span: &Span) -> Option<String> {
                 // They are low-cardinality.
                 Some(description.to_owned())
             }
+            ("contentprovider", "load") => {
+                // `contentprovider.load` spans contain paths of third party framework components
+                // and their onCreate method such as
+                // `io.sentry.android.core.SentryPerformanceProvider.onCreate`, which
+                // _should_ be low-cardinality, on the order of 10s per project.
+                Some(description.to_owned())
+            }
+            ("application", "load") => {
+                // `application.load` spans contain paths of app components and their
+                // onCreate method such as
+                // `io.sentry.samples.android.MyApplication.onCreate`, which _should_ be
+                // low-cardinality.
+                Some(description.to_owned())
+            }
+            ("activity", "load") => {
+                // `activity.load` spans contain paths of app components and their onCreate/onStart
+                // method such as `io.sentry.samples.android.MainActivity.onCreate`, which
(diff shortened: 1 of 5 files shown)


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -82,6 +82,26 @@ pub(crate) fn scrub_span_description(span: &Span) -> Option<String> {
                 // They are low-cardinality.
                 Some(description.to_owned())
             }
+            ("contentprovider", "load") => {
```
narsaynorath, line 85: @jjbayer You pointed me to this file initially, is this the only change I need to make to have these new span ops ingested internally?

jjbayer: You'll also want to extend the `MOBILE_OPS` constant in `defaults.rs`: https://github.com/getsentry/relay/blob/7606bce0845179da2799179aa6cea855d3859b4d/relay-dynamic-config/src/defaults.rs#L20-L21

narsaynorath: Sounds good, I was thinking about doing that but I was worried adding the entry to `default.rs` would enable it incorrectly for all orgs. I've added them now and also regenerated some snapshots with these new ops

jjbayer: That is a good point! I just merged https://github.com/getsentry/relay/pull/2924, so the `Feature::SpanMetricsExtractionAllModules` flag and only extend the mobile ops conditionally.

narsaynorath:
@jjbayer Just a heads up I've addressed your comment in this PR and implemented the change we discussed to update the mobile ops conditionally. Totally open to seeing alternative implementations if you have suggestions on how to do this part better, but I found a way to do it in a single block 👍

jjbayer (approved):

jjbayer:
> @jjbayer Just a heads up I've addressed your comment in this PR and implemented the change we discussed to update the mobile ops conditionally. Totally open to seeing alternative implementations if you have suggestions on how to do this part better, but I found a way to do it in a single block 👍

@narsaynorath looks good to me. There's definitely a way to do it with iterator chaining, but IMO not worth the effort and it would not increase readability.

PR #2928, opened 2024-01-10
feat(quotas): Add global throughput limit

Dav1dde:
Similar to  #2854 but with locks instead of atomics.

Adds a mechanism to limit the throughput of metric buckets.

It's done globally, meaning across different relays, using redis.

In order to not call redis every time, we use a budget system (the name quota already taken). We "take" a certain budget by incrementing a redis counter for the given global quota in the given slot, put it in a local counter, and count down to zero before asking for more from redis.


relay-quotas/src/global.rs
```diff
@@ -0,0 +1,256 @@
...
+
+        let key = BudgetKeyRef::new(quota);
+
+        let opt_val = self.counters.read().unwrap().get(&key).cloned();
+        let val = match opt_val {
+            Some(val) => val,
+            None => Arc::clone(self.counters.write().unwrap().entry_ref(&key).or_default()),
+        };
+
+        let mut lock = val.lock().unwrap();
```
jjbayer: Do tokio worker threads automatically `catch_unwind` panics? I am not confident that Relay can actually recover from a panic here.

Dav1dde: We should deal with poisened locks properly now `.unwrap_or_else(PoisonError::into_inner)`

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,256 @@
...
+        quota: &RedisQuota,
+        quantity: usize,
+    ) -> Result<bool, RedisError> {
+        let Some(limit) = quota.limit else {
+            return Ok(false);
+        };
+
+        let key = BudgetKeyRef::new(quota);
+
+        let opt_val = self.counters.read().unwrap().get(&key).cloned();
```
Dav1dde: @jjbayer suggested to use a `Mutex` here since all we do is just a lookup anyways:

```rs
let val = { 
    let counters = self.counters.lock().unwrap_or_else(PoisonError::into_inner);
    Arc::clone(counters.entry_ref(&key).or_default())
};
```

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,256 @@
...
+        let reserved = self
+            .redis_counter
+            .try_reserve(client, budget_to_reserve, limit, quota)? as usize;
+
+        self.local_counter.increase_budget(reserved);
+        Ok(!self.local_counter.try_consume(quantity))
+    }
+
+    fn default_request_size_based_on_limit(&self) -> usize {
+        100
```
Dav1dde: We need to actually implement this, use a percentage we grab from the config

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub fn is_rate_limited(
+        &mut self,
+        client: &mut PooledClient,
+        quota: &RedisQuota,
+        quantity: usize,
+        limit: u64,
+    ) -> Result<bool, RedisError> {
+        let current_slot = current_slot(quota.window());
+
+        match self.slot.cmp(&current_slot) {
```
Dav1dde: As briefly discussed we could move this logic into the `Counter` impl, since we have interiror mutability instead of re-creating the counter we would just reset the in-memory counter.

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,256 @@
...
+            Ordering::Greater => {
+                // TODO double check logic here
+                // in theory this should never happen only if someone messes with the system time
+                // be safe and dont rate limit
+                relay_log::error!("budget slot ahead of current slot");
```
Dav1dde: This is risky, if this actually happens this will happen for a longer period of time with a very high volume.

Ideally we would only log this error like once per second.

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,256 @@
...
+        &mut self,
+        client: &mut PooledClient,
+        quantity: usize,
+        limit: u64,
+        quota: &RedisQuota,
+    ) -> Result<u64, RedisError> {
+        let script = load_global_lua_script();
+
+        let redis_key = quota.key();
+        let expiry = UnixTimestamp::now().as_secs() + quota.window();
```
Dav1dde: `quota.expiry()` exists already, is there a reason we cannot use that?

In `redis.rs` we use `GRACE`, we should also do it here. To not have it in multiple spots, we can just add the `GRACE` to the return value of `quota.expiry()` directly.

CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
 # Changelog
 
-## Unreleased
+## 23.12.1
```
jjbayer: This is definitely wrong, `23.12.1` is already listed below.

```suggestion
## Unreleased
```

relay-quotas/src/redis.rs
```diff
@@ -99,18 +110,26 @@ impl<'a> RedisQuota<'a> {
...
+    pub fn key(&self) -> String {
         // The subscope id is only formatted into the key if the quota is not organization-scoped.
         // The organization id is always included.
         let subscope = match self.quota.scope {
             QuotaScope::Organization => None,
             scope => self.scoping.scope_id(scope),
         };
 
+        // 0 is arbitrary, we just need to ensure global quotas from different orgs have the same key.
+        let org = if self.quota.scope == QuotaScope::Global {
```
Dav1dde: We need this logic everywhere we access `self.scoping.organization_id` -> in `shift`

relay-quotas/src/lib.rs
```diff
@@ -11,9 +11,14 @@
 /// typically happens for disabled keys, projects, or organizations.
 const REJECT_ALL_SECS: u64 = 60;
 
+#[cfg(feature = "redis")]
```
Dav1dde: Would move this together with:

```rs
#[cfg(feature = "redis")]
pub use self::global::*;
```

Directly before the `mod redis` block. All `redis` features together and alphabetically sorted.

relay-quotas/src/redis.rs
```diff
@@ -797,4 +835,58 @@ mod tests {
...
+        };
+
+        let inner_scoping = Scoping {
+            organization_id: 69420,
+            project_id: ProjectId::new(42),
+            project_key: ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap(),
+            key_id: Some(4711),
+        };
+
+        let static_inner_scoping: &'static Scoping = Box::leak(Box::new(inner_scoping));
```
Dav1dde: Still not a fan of the `Box::leak`, just splitting this setup function in 2 gets rid of the lifetime issues:

```
let quota = build_quota();
let quota = build_redis_quota(&quota);
```

relay-quotas/src/redis.rs
```diff
@@ -797,4 +835,58 @@ mod tests {
...
+        let scoping = ItemScoping {
+            category: DataCategory::MetricBucket,
+            scoping: static_inner_scoping,
+        };
+
+        RedisQuota::new(static_quota, scoping, UnixTimestamp::now()).unwrap()
+    }
+
+    #[test]
+    fn test_global_ratelimit() {
```
Dav1dde: We should also have a test using at least two separate `GlobalRateLimits` accessing the same Redis and making sure they are correct. This doesn't necessarily allow us to test perfect parallelism but should allow us to test multiple (hundreds) of rate limit attempts through multiple limiters and verify at the end that the total count is correct.

relay-quotas/src/global.rs
TBS1996: even though it's pretty short isn't it still better to not block other threads while we get the counter?

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,231 @@
...
+        quota: &RedisQuota,
+        quantity: usize,
+    ) -> Result<bool, RedisError> {
+        let Some(limit) = quota.limit else {
+            return Ok(false);
+        };
+
+        let key = BudgetKeyRef::new(quota);
+        let val = {
+            let mut limits = self.limits.lock().unwrap_or_else(PoisonError::into_inner);
```
jjbayer, line 35: Does it actually make sense for our business logic to use the value if another thread panicked? From the [docs](https://doc.rust-lang.org/nomicon/poisoning.html):

> However if the Mutex contained, say, a BinaryHeap that does not actually have the heap property, it's unlikely that any code that uses it will do what the author intended. As such, the program should not proceed normally. Still, if you're double-plus-sure that you can do something with the value, the Mutex exposes a method to get the lock anyway. It is safe, after all. Just maybe nonsense.

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,231 @@
...
+fn load_global_lua_script() -> &'static Script {
+    static SCRIPT: OnceLock<Script> = OnceLock::new();
+    SCRIPT.get_or_init(|| Script::new(include_str!("global_quota.lua")))
+}
+
+fn current_slot(window: u64) -> usize {
+    UnixTimestamp::now()
+        .as_secs()
+        .checked_div(window)
+        .unwrap_or_default() as usize
```
jjbayer: Should we log an error here? Might be weird to silently use a zero slot when the window is misconfigured to `0`.

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    fn increase_budget(&mut self, quantity: usize) {
+        self.budget = self.budget.saturating_add(quantity);
+    }
+}
+
+struct RedisCounter {
+    last_seen: u64,
```
jjbayer: nit: Might be just me but I associate `last_seen` with a timestamp.
```suggestion
    latest: u64,
```

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,231 @@
...
+            self.slot = current_slot;
+        }
+
+        if self.local_counter.try_consume(quantity) {
+            return Ok(false);
+        }
+
+        if !self.redis_counter.can_satisfy(quantity, limit) {
+            return Ok(true);
+        }
```
jjbayer: Can we flip this around and call the method `exceeds_limit`?

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,231 @@
...
+    fn can_satisfy(&self, quantity: usize, limit: u64) -> bool {
+        self.last_seen + quantity as u64 <= limit
+    }
+
+    fn try_reserve(
+        &mut self,
+        client: &mut PooledClient,
+        quantity: usize,
+        limit: u64,
+        quota: &RedisQuota,
```
jjbayer: nit: `quota` determines the redis key, which comes before redis args, so i would put the parameter between `client` and `quantity`.

relay-quotas/src/global_quota.lua
```diff
@@ -0,0 +1,66 @@
...
+-- fill up redis with dead keys.
+
+
+-- The key to the global quota.
+local key = KEYS[1]
+-- The max amount that we want to take within the given slot. We won't take a budget if
+-- the count is higher than the limit.
+local limit = tonumber(ARGV[1])
+-- When the redis key/val should be deleted.
+local expiry = tonumber(ARGV[2])
```
jjbayer: nit: I would move `expiry` to the last position, as the least relevant arg.

relay-quotas/src/redis.rs
```diff
@@ -135,11 +154,11 @@ impl std::ops::Deref for RedisQuota<'_> {
 /// attachments. For more information on quota parameters, see `QuotaConfig`.
 ///
 /// Requires the `redis` feature.
-#[derive(Clone)]
 pub struct RedisRateLimiter {
     pool: RedisPool,
     script: Arc<Script>,
     max_limit: Option<u64>,
+    counters: GlobalRateLimits,
```
jjbayer: nit: 
```suggestion
    global_limits: GlobalRateLimits,
```

relay-quotas/src/quota.rs
```diff
@@ -71,6 +71,7 @@ impl ItemScoping<'_> {
     /// Returns the identifier of the given scope.
     pub fn scope_id(&self, scope: QuotaScope) -> Option<u64> {
         match scope {
+            QuotaScope::Global => Some(0),
```
jjbayer: Why not `None`?

relay-quotas/src/redis.rs
```diff
@@ -797,4 +835,58 @@ mod tests {
...
+            .unwrap();
+        let expected_ratelimit_result = [false, false, true, true].to_vec();
+
+        // The limit is 200, while we take 90 at a time. So the first two times we call, we'll
+        // still be under the limit. 90 < 200 -> 180 < 200 -> 270 > 200 -> 360 > 200.
+        for should_ratelimit in expected_ratelimit_result {
+            let is_ratelimited = counter
+                .is_rate_limited(&mut client, &redis_quota, 90)
+                .unwrap();
+
```
jjbayer: Should we test that the result flips back to true after the slot expires?

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,231 @@
...
+use crate::RedisQuota;
+
+fn load_global_lua_script() -> &'static Script {
+    static SCRIPT: OnceLock<Script> = OnceLock::new();
+    SCRIPT.get_or_init(|| Script::new(include_str!("global_quota.lua")))
+}
+
+fn current_slot(window: u64) -> usize {
+    UnixTimestamp::now()
+        .as_secs()
```
jjbayer: Is discussed in standup, we should use the data's `received` timestamp instead of `now`.

jjbayer (approved):
General approach looks good to me. See comment about `received` time.

relay-quotas/src/redis.rs
```diff
@@ -99,18 +110,26 @@ impl<'a> RedisQuota<'a> {
...
             QuotaScope::Organization => None,
             scope => self.scoping.scope_id(scope),
         };
 
+        // 0 is arbitrary, we just need to ensure global quotas from different orgs have the same key.
+        let org = if self.quota.scope == QuotaScope::Global {
+            0
+        } else {
+            self.scoping.organization_id
+        };
```
iker-barriocanal: Is this the same as `self.scoping.scope_id(self.quota.scope)`? I think setting the arbitrary `0` here and `scope_id` makes introducing breaking changes too easy.

relay-quotas/src/redis.rs
```diff
@@ -200,19 +221,36 @@ impl RedisRateLimiter {
...
+                                &quota,
+                                item_scoping.scoping,
+                                self.retry_after((quota.expiry() - timestamp).as_secs()),
+                            ));
+                        }
+                        Ok(false) => continue,
+                        Err(e) => relay_log::error!(
+                            error = &e as &dyn std::error::Error,
+                            "failed to check global rate limit"
+                        ),
```
iker-barriocanal: IIUC, no rate limiting would happen in this case. What do you think about updating the function's docstring?

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/// A rate limiter for global rate limits.
+#[derive(Default)]
+pub struct GlobalRateLimits {
+    limits: Mutex<hashbrown::HashMap<BudgetKey, Arc<Mutex<GlobalRateLimit>>>>,
+}
+
+impl GlobalRateLimits {
+    /// Returns `true` if the global quota should be ratelimited.
```
iker-barriocanal:
```suggestion
    /// Returns `Some(true)` if the global quota should be ratelimited.
```
?

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,231 @@
...
+        self.prefix == key.prefix && self.window == key.window
+    }
+}
+
+/// A single global rate limit.
+///
+/// When we want to ratelimit across all relay-instances, we need to use Redis to synchronize.
+/// Calling Redis every time we want to check if an item should be ratelimited would be very expensive,
+/// which is why we have this cache. It works by 'taking' a certain budget from Redis, by pre-incrementing
+/// a global counter. We Put the amount we pre-incremented into this local cache and count down until
```
iker-barriocanal:
```suggestion
/// a global counter. We put the amount we pre-incremented into this local cache and count down until
```

relay-quotas/src/redis.rs
```diff
@@ -77,40 +82,71 @@ impl<'a> RedisQuota<'a> {
...
         let next_slot = self.slot() + 1;
         let next_start = next_slot * self.window + self.shift();
         UnixTimestamp::from_secs(next_start)
     }
 
-    fn key(&self) -> String {
+    /// Returns when the key should expire in Redis.
+    ///
+    /// Like [`Self::expiry()`] but adds an additional grace period for the key.
+    pub fn key_expiry(&self) -> u64 {
```
iker-barriocanal, line 126:
```suggestion
    pub fn expiry_with_grace(&self) -> u64 {
```
What do you think?

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// we try again to decrement the budget after syncing.
+    pub fn is_rate_limited(
+        &self,
+        client: &mut PooledClient,
+        quota: &RedisQuota,
+        quantity: usize,
+    ) -> Result<bool, RedisError> {
+        let key = KeyRef::new(quota);
+        let val = {
+            let mut limits = self.limits.lock().unwrap_or_else(PoisonError::into_inner);
```
iker-barriocanal, line 35: Is it worth it to add some observability to identify potential bugs when we see plenty of poisoning errors?

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,349 @@
...
+        let max_available_budget = quota
+            .limit
+            .unwrap_or(u64::MAX)
+            .saturating_sub(self.last_seen_redis_value);
+
+        if min_required_budget > max_available_budget {
+            return Ok(0);
+        }
+
+        let budget_to_reserve = min_required_budget.max(self.default_request_size(quantity, quota));
```
iker-barriocanal, line 165: Why do we care about the minimum required budget, and not request a default amount directly? Requesting minimum amounts may result in more Redis checks (do we know the impact?) which I believe we're trying to partially avoid.

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,349 @@
...
+            self.last_seen_redis_value = 0;
+            self.slot = quota_slot;
+        }
+
+        if self.budget >= quantity {
+            self.budget -= quantity;
+            return Ok(false);
+        }
+
+        let reserved = self.try_reserve(client, quantity, quota)?;
```
iker-barriocanal, line 138: Do we have mechanisms to avoid requesting from Redis too much? IIUC, we'll make a Redis request every tie there's no budget in the local cache.

relay-quotas/src/global.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    #[test]
+    fn test_global_ratelimit() {
+        let window = 10;
+        let limit = 200;
+
+        let quota = build_quota(window, limit);
+        let scoping = build_scoping();
+        let redis_quota = build_redis_quota(&quota, &scoping);
```
iker-barriocanal: I like this approach more than the `Box::leak` 👍

iker-barriocanal (commented):
Overall looks good to me, see questions on the requesting budget and making too many calls to Redis.

relay-quotas/src/global.rs
Dav1dde: We prefer `last_seen` because it makes it clear the value may be out of date and not in sync with Redis.

relay-quotas/src/global.rs
Dav1dde: Yeah that is fine, we never actually panic in the critical section, but even if we do, there is nothing that is left in an unsafe or uncertain state, since all we do is either insert something into the map or clone an Arc.

relay-quotas/src/global.rs
Dav1dde: We're properly recovering from the error now.

relay-quotas/src/global.rs
Dav1dde: Implemented with a global constant for now.

Might in the future want to use some kind of moving average instead.

relay-quotas/src/redis.rs
Dav1dde: I prefer the `key_expiry`, we're only interested to get an expiry for the Redis key which was created by the `key` method, this makes it clear they go together. Also that there is a "grace" is not relevant for the caller, the caller is just interested to know when to expire the Redis key.

relay-quotas/src/global.rs
Dav1dde: Effectively this lock will never be poisned, this only happens through a panic, which we would already notice. This is unfortunately just an API design error from the std lib we have to deal with here.

relay-quotas/src/global.rs
Dav1dde: If the minimum is higher than the default, we still need to fetch the minimum to not have a bug, in practice this probably never happens, but there might be an edge case where we have a huge quantity once.

relay-quotas/src/global.rs
Dav1dde: This is 'solved' by having a high default we reserve from Redis instead of just reserving the quantity. We currently reserve `0.01%` of the limit.

TBS1996 (approved):

relay-quotas/src/redis.rs
Dav1dde: No, this is purely a default for the Redis key, the values are unrelated.

PR #2929, opened 2024-01-10
feat(spans): De-duplicate SQL conditions

jjbayer:
SQL queries are sometimes generated with dynamic conditions (`.. OR .. OR .. OR ..`). After parameter scrubbing, these operands look the same, for example `id = %s OR id = %s OR id = %s`. In this case, simplify the condition to `id = %s`. 

ref: [internal issue](https://www.notion.so/sentry/Collapse-identical-OR-conditions-in-queries-40786f8cddb145e398e8ad2d5eb66522)

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/sql/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/mod.rs
@@ -704,12 +704,12 @@ mod tests {
         FROM a
         WHERE a.status = %s
         AND (
-        (a.id = %s AND a.org = %s)
-        OR (a.id = %s AND a.org = %s)
-        OR (a.id = %s AND a.org = %s)
-        OR (a.id = %s AND a.org = %s)
+            (a.id = %s AND a.org = %s)
+            OR (a.id = %s AND a.org = %s)
+            OR (a.id = %s AND a.org = %s)
+            OR (a.id = %s AND a.org = %s)
         )"#,
-        "SELECT * FROM a WHERE status = %s AND ((a.id = %s AND a.org = %s) OR ..)"
+        "SELECT * FROM a WHERE status = %s AND ((id = %s AND org = %s))"
     );
 
     scrub_sql_test!(
--- a/relay-event-normalization/src/normalize/span/description/sql/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/mod.rs
@@ -713,6 +713,32 @@ mod tests {
     );
 
     scrub_sql_test!(
+        duplicate_conditions_or,
+        r#"SELECT *
+        FROM a
+        WHERE a.status = %s
+        OR (
+            (a.id = %s OR a.org = %s)
+            OR (a.id = %s OR a.org = %s)
+            OR (a.id = %s OR a.org = %s)
+            OR (a.id = %s OR a.org = %s)
+        )"#,
+        "SELECT * FROM a WHERE status = %s OR ((id = %s OR org = %s))"
+    );
+
+    scrub_sql_test!(
+        non_duplicate_conditions,
+        r#"SELECT *
+        FROM a
+        WHERE a.status = %s
+        AND (
+            (a.id = %s AND a.org2 = %s)
+            OR (a.id = %s AND a.org = %s)
+        )"#,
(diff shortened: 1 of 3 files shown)


relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -698,6 +698,46 @@ mod tests {
...
+        r#"SELECT *
+        FROM a
+        WHERE a.status = %s
+        OR (
+            (a.id = %s OR a.org = %s)
+            OR (a.id = %s OR a.org = %s)
+            OR (a.id = %s OR a.org = %s)
+            OR (a.id = %s OR a.org = %s)
+        )"#,
+        "SELECT * FROM a WHERE status = %s OR ((id = %s OR org = %s))"
```
jjbayer: @gggritso this is not exactly the placeholder you suggested, but it does keep the implementation simple. Let me know if it's good enough.

relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -698,6 +698,46 @@ mod tests {
...
+        r#"SELECT *
+        FROM a
+        WHERE a.status = %s
+        AND (
+            (a.id = %s AND a.org = %s)
+            OR (a.id = %s AND a.org = %s)
+            OR (a.id = %s AND a.org = %s)
+            OR (a.id = %s AND a.org = %s)
+        )"#,
+        "SELECT * FROM a WHERE status = %s AND ((id = %s AND org = %s))"
```
olksdr: are the double parentheses `((` left on purpose ?

olksdr (approved):
one question, otherwise lgtm

gggritso:
@Zylphrex what do you think?

relay-event-normalization/src/normalize/span/description/sql/mod.rs
gggritso: Works for me 👍🏻 going to defer to @Zylphrex on this, since he found the original issue

Zylphrex: This looks fine to me as it covers the case I originally raised.

relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -211,8 +211,8 @@ mod tests {
 
     scrub_sql_test!(
         various_parameterized_ins_percentage,
-        "SELECT count() FROM table1 WHERE id IN (%s, %s) AND id IN (%s, %s, %s)",
-        "SELECT count() FROM table1 WHERE id IN (%s) AND id IN (%s)"
```
Zylphrex, line 215: I think this raises an edge case how queries like this one where we have the same condition twice joined by a `AND` operator.

Likely an uncommon use case, but the first list may be a subset of the second so it may be valid to transform it into just `id IN (%s)`.

Zylphrex (approved):

relay-event-normalization/src/normalize/span/description/sql/mod.rs
jjbayer: I think it's OK to pretend that they are the same for grouping purposes, with the same reasoning as for the ` OR ` clause: If the left operand (`id`) and the operator ` IN ` repeat multiple times, it is likely a dynamically generated query that we can simplify.

relay-event-normalization/src/normalize/span/description/sql/mod.rs
jjbayer: I did not intend to leave them initially, but they might help indicate to the user that something was scrubbed here. They should be easy to remove though, if we want to @gggritso any opinion?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/description/sql/mod.rs
iker-barriocanal: I'd personally remove them. I think it's easy to notice queries are getting scrubbed so it's no surprise.

jjbayer: Done.

gggritso: No strong opinion here 👍🏻

PR #2930, opened 2024-01-10
feat(spans): Send outcomes for spans

phacops:
We want to keep track of span outcomes in order to be able to bill them.

Since Relay is responsible to produce spans to Snuba directly, when we keep a span we need to also send an `Accepted` outcome.


jjbayer (changes requested):
From the moment we `process` a span until it's actually sent to kafka, there are still a few failure modes (`StoreService::produce_span` is not infallible). So I think we should log the outcome right after producing to kafka, i.e. here: https://github.com/getsentry/relay/blob/4bccbbb9326ad0b1f77e7414120b2eeff8a940f5/relay-server/src/actors/store.rs#L849

`StoreService` already has access to `outcome_aggregator`, so it should be easy to track the outcome from there.

relay-server/src/envelope.rs
```diff
@@ -617,9 +617,12 @@ impl Item {
             ItemType::ReplayEvent | ItemType::ReplayRecording => Some(DataCategory::Replay),
             ItemType::ClientReport => None,
             ItemType::CheckIn => Some(DataCategory::Monitor),
+            ItemType::Span | ItemType::OtelSpan => Some(if indexed {
```
jjbayer, line 620: For standalone spans, `indexed` will always be `false`, because it gets its value from https://github.com/getsentry/relay/blob/518148061bde526b7f6713b8e3c32c4660e4d53d/relay-server/src/utils/managed_envelope.rs#L299, which is only true if there was an event!

Can we live with that for now? I can try to make `indexed` depend on an item header instead of the event flag, then everything should fall into place.

relay-base-schema/src/data_category.rs
```diff
@@ -65,6 +65,10 @@ pub enum DataCategory {
     UserReportV2 = 14,
     /// Metric buckets.
     MetricBucket = 15,
+    /// SpanIndexed
+    ///
+    /// This is the category for spans we store in full.
+    SpanIndexed = 16,
```
jjbayer: As discussed: We need to release librelay with this new category and then bump it in sentry and snuba before we start producing outcomes with this ID.

jjbayer (commented):
We have a bunch of integration tests for outcomes in `test_outcome.py`, especially around indexed vs. non-indexed. Would be good to add one for spans.

relay-base-schema/src/data_category.rs
phacops: https://github.com/getsentry/relay/pull/2937

relay-server/src/envelope.rs
phacops: I tried something in that regard, let me know what you think (https://github.com/getsentry/relay/pull/2930/commits/872eb2e16e6afda0e15157e60d661a31739f5734, https://github.com/getsentry/relay/pull/2930/commits/5b2974396314e510d6f5149b80311d8043c528d1).

relay-server/src/actors/processor/span/processing.rs
```diff
@@ -110,6 +111,7 @@ pub fn process(state: &mut ProcessEnvelopeState, config: Arc<Config>) {
             }
         };
         new_item.set_payload(ContentType::Json, payload);
+        new_item.set_metrics_extracted(state.event_metrics_extracted);
```
jjbayer: This should take the value of the old item, not of the event.

relay-server/src/actors/processor/span/processing.rs
```diff
@@ -128,18 +130,30 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState) {
             Ok(span) => span,
             Err(e) => {
                 relay_log::error!("Invalid span: {e}");
+                state.managed_envelope.track_outcome(
+                    Outcome::Invalid(DiscardReason::InvalidSpan),
+                    relay_quotas::DataCategory::SpanIndexed,
+                    1,
+                );
```
jjbayer, line 137: This raises a new question: Do we want outcomes for spans that were extracted from transactions?

This question is especially relevant for the `Accepted` outcome. When it becomes time to charge customers for standalone spans, won't we need to differentiate between spans that were submitted as part of transactions vs. spans that were submitted standalone? cc @indragiek

relay-server/src/actors/processor/span/processing.rs
```diff
@@ -128,18 +130,30 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState) {
...
                 return;
             }
         };
         let mut item = Item::new(ItemType::Span);
         item.set_payload(ContentType::Json, span);
+        // At this point, metrics were extracted already.
+        item.set_metrics_extracted(true);
```
jjbayer:
```suggestion
        // If metrics extraction happened for the event, it also happened for its spans:
        item.set_metrics_extracted(state.event_metrics_extracted);
```

relay-server/src/utils/managed_envelope.rs
```diff
@@ -194,16 +194,23 @@ impl ManagedEnvelope {
     {
         let mut outcomes = vec![];
         let use_indexed = self.use_index_category();
+        let mut push_outcome =
+            |item: &mut Item, outcome: Outcome, quantity: usize, use_indexed: bool| {
+                if let Some(category) = item.outcome_category(use_indexed) {
+                    outcomes.push((outcome, category, quantity));
+                };
+            };
```
```suggestion
```

relay-server/src/utils/managed_envelope.rs
```diff
@@ -194,16 +194,23 @@ impl ManagedEnvelope {
...
         self.envelope.retain_items(|item| match f(item) {
             ItemAction::Keep => true,
+            ItemAction::DropSilently => false,
+            ItemAction::Drop(outcome) if item.ty() == &ItemType::Span => {
+                push_outcome(item, outcome, item.quantity(), item.metrics_extracted());
+                false
+            }
```
```suggestion
```

relay-server/src/utils/managed_envelope.rs
```diff
@@ -194,16 +194,23 @@ impl ManagedEnvelope {
...
             ItemAction::Drop(outcome) => {
-                if let Some(category) = item.outcome_category(use_indexed) {
-                    outcomes.push((outcome, category, item.quantity()));
-                }
-
+                push_outcome(item, outcome, item.quantity(), use_indexed);
```
```suggestion
                let use_indexed = if item.ty() == &ItemType::Span {
                    item.metrics_extracted()
                } else {
                    use_indexed
                };
                if let Some(category) = item.outcome_category(use_indexed) {
                    outcomes.push((outcome, category, quantity));
                };

                
```

relay-server/src/actors/processor/span/processing.rs
phacops: There's no difference between charging for extracted or standalone spans.

Once the customer choose the new plan, we'll only look at span outcomes to understand how to bill them. The customer might use a newer SDK, sending standalone spans, or an older SDK, sending transactions and in both cases, we'd want to know the number of accepted spans to bill them and that's why we want to count spans extracted from the transactions as well.

jjbayer: Alright!

relay-server/src/actors/store.rs
```diff
@@ -836,17 +840,40 @@ impl StoreService {
...
+            scoping.organization_id,
+            KafkaMessage::Span(span),
+        )?;
+
+        self.outcome_aggregator.send(TrackOutcome {
+            category: DataCategory::SpanIndexed,
+            event_id,
+            outcome: Outcome::Accepted,
+            quantity: 1,
+            remote_addr,
```
jjbayer: Let's set these fields (`event_id` and `remote_addr`) to `None` so that the outcome aggregator can summarize outcomes. It already drops these fields under certain conditions (e.g. `Invalid` outcome), but spans are very high volume so we don't want to ever generate a separate outcome kafka message for every span.

```suggestion
            event_id: None,
            outcome: Outcome::Accepted,
            quantity: 1,
            remote_addr: None,
```

jjbayer (approved):
Approving to unblock. Before you merge, please

* Remove `event_id` and `remote_addr` from the `Accepted` outcome.
* Make sure librelay versions in sentry and snuba are up-to-date with the new `SpanIndexed` category.
* Add an integration test.
* After merging, deploy to Canary and observe how many extra Kafka messages are being sent.

relay-server/src/actors/store.rs
jjbayer: Actually, could you also add `Accepted` to the `matches!` condition here, so that we don't mess this up in any future use cases of accepted? Thanks! https://github.com/getsentry/relay/blob/5cd6c1ed620f0ae6cf5823b98245c32582d45f60/relay-server/src/actors/outcome_aggregator.rs#L163-L172

PR #2932, opened 2024-01-10
fix(spans): Update extraction condition for count_per_segment

shruthilayaj:
We'd like to keep a count of _all_ mobile spans so we
can determine "other". Updates the condition to not just
include the mobile span ops.

#skip-changelog


relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_mobile.snap
```diff
@@ -330,5 +369,28 @@ expression: "(&event.value().unwrap().spans, metrics)"
...
+                1.0,
+            ),
+            tags: {
+                "span.op": "custom.op",
+            },
+        },
+        Bucket {
+            timestamp: UnixTimestamp(1597976303),
+            width: 0,
+            name: "c:spans/count_per_segment@none",
```
shruthilayaj, line 386: ensures we extract this metric for a custom op

jjbayer (approved):

PR #2934, opened 2024-01-11
ref(processor): Sample items, not envelopes

jjbayer:
Preparation for https://github.com/getsentry/relay/pull/2374.

In that PR need the ability to reject only a subset of envelope items in dynamic sampling, so we have to go from rejecting the entire envelope to rejecting individual items.

This PR does not yet change dynamic sampling behavior.

#skip-changelog


relay-server/src/actors/processor.rs
```diff
@@ -1418,13 +1414,6 @@ impl EnvelopeProcessorService {
                             state.managed_envelope.reject(outcome);
                         }
 
-                        if err.should_keep_metrics() {
-                            state.extracted_metrics.send_metrics(
-                                state.managed_envelope.envelope(),
-                                self.inner.project_cache.clone(),
-                            );
-                        }
```
jjbayer, line 1434: Isn't this nice: A negative sampling decision is not an error anymore, so no need to distinguish between "good errors" and "bad errors".

relay-server/src/actors/processor.rs
```diff
@@ -283,9 +283,6 @@ pub enum ProcessingError {
     #[error("failed to apply quotas")]
     QuotasFailed(#[from] RateLimitingError),
 
-    #[error("event dropped by sampling rule {0}")]
-    Sampled(MatchedRuleIds),
```
jjbayer, line 289: Dynamic sampling is now infallible.

relay-server/src/actors/processor.rs
```diff
@@ -498,10 +490,14 @@ impl<'a> ProcessEnvelopeState<'a> {
...
     /// Removes the event payload from this processing state.
-    #[cfg(feature = "processing")]
     fn remove_event(&mut self) {
         self.event = Annotated::empty();
     }
+
+    fn reject_event(&mut self, outcome: Outcome) {
+        self.remove_event();
+        self.managed_envelope.reject_event(outcome);
+    }
```
jjbayer, line 499: Because  we do not reject the entire envelope, we need to manually remove the event from the state & the envelope itself.

relay-server/src/actors/processor/dynamic_sampling.rs
```diff
@@ -96,20 +96,21 @@ pub fn run(state: &mut ProcessEnvelopeState, config: &Config) {
...
     if let SamplingResult::Match(sampling_match) = std::mem::take(&mut state.sampling_result) {
         // We assume that sampling is only supposed to work on transactions.
         if state.event_type() == Some(EventType::Transaction) && sampling_match.should_drop() {
             let matched_rules = sampling_match.into_matched_rules();
 
+            let outcome = Outcome::FilteredSampling(matched_rules.clone());
             state
                 .managed_envelope
-                .reject(Outcome::FilteredSampling(matched_rules.clone()));
+                .retain_items(|_| utils::ItemAction::Drop(outcome.clone()));
```
jjbayer, line 108: In this PR, we drop all items unconditionally because that is what happened before. In the next PR, we will introduce a condition here to keep profiles even when their transaction is dropped.

relay-server/src/utils/managed_envelope.rs
```diff
@@ -219,7 +219,7 @@ impl ManagedEnvelope {
         self.envelope.retain_items(|item| match f(item) {
             ItemAction::Keep => true,
             ItemAction::Drop(outcome) => {
-                if let Some(category) = item.outcome_category(use_indexed) {
+                if let Some(category) = dbg!(item.outcome_category(use_indexed)) {
```
```suggestion
                if let Some(category) = item.outcome_category(use_indexed) {
```

relay-server/src/utils/managed_envelope.rs
```diff
@@ -249,6 +249,16 @@ impl ManagedEnvelope {
         self
     }
 
+    /// Remove event item(s) and log an outcome.
```
olksdr:
```suggestion
    /// Removes event item(s) and log an outcome.
```

olksdr (approved):
lgtm.
Any specific tests we want to add before merging this?

jjbayer:
> Any specific tests we want to add before merging this?

@olksdr I'll try to think of one. The objective is that nothing changes, so I can only add coverage for edge cases that were not covered before.

relay-server/src/actors/processor/dynamic_sampling.rs
```diff
@@ -96,20 +96,21 @@ pub fn run(state: &mut ProcessEnvelopeState, config: &Config) {
...
 
-            return Err(ProcessingError::Sampled(matched_rules));
+            // The event is no longer in the envelope, so we need to handle it separately:
+            state.reject_event(outcome);
```
Dav1dde, line 111: I don't quite understand this. We drop the items in the managed envelope through retain_items which seems to already log outcomes through the managed envelope, why is this still needed?

And why the `event_category` conditional check in the implementation?

relay-server/src/actors/processor.rs
Dav1dde: Won't this change current behaviour? We would recover from errors before, now we don't because everything is dropped in the new sampling impl.

jjbayer: `should_keep_metrics` was only true for the `Sampled` error, so this code was a workaround to send the metrics anyway. Now that the `Sampled` error no longer exists, we can just rely on the `Ok` branch above.

relay-server/src/actors/processor/dynamic_sampling.rs
jjbayer: The event at this point is not an envelope item, but a separate field on the state (see `extract_event`).

> And why the event_category conditional check in the implementation?

We need an event category to log an outcome for it. When `event_category` is `None`, it means there is no event to remove.

iker-barriocanal (approved):

Dav1dde (approved):

PR #2935, opened 2024-01-11
ref(normalization): Remove unused normalization returns

iker-barriocanal:
This PR removes unused returns in event normalization. There are no functional changes.

Most functions that returned a `ProcessingAction` always return the `Ok` variant, not requiring significant code changes, and meaning the return is indeed not useful. The `DeleteValueHard` and `DeleteValueSoft` error variants are swallowed by the processor (including `processor::apply` calls) to add errors and remarks to `meta`, but these are [never bubbled up](https://github.com/getsentry/relay/blob/478c3ada166b108464a0971827433a8db090935e/relay-event-schema/src/processor/funcs.rs#L22-L25). The `InvalidTransaction` error variant, the [only one that's bubbled up](https://github.com/getsentry/relay/blob/478c3ada166b108464a0971827433a8db090935e/relay-event-schema/src/processor/funcs.rs#L26), is only returned by `TransactionsProcessor` and `TimestampProcessor`, and these remain unchanged.

For an easier review, each commit in the PR refactors a specific normalization step.

### Motivation

The goal is to have two separate steps: validation, validating the correctness of an event and rejecting it if it's invalid, and normalization, normalizing the payload without rejecting the event. In code, this means that the normalization function should not return a `ProcessingAction`; thus, each normalization step should not return a `ProcessingAction` either. The changes presented in this PR are required before taking the next step.

### Next steps

After this PR, only two remaining items returning a `ProcessingAction` are left: `TransactionsProcessor` and `TimestampProcessor`. The next step is to move the validation steps of these processors from normalization into a separate validation step.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -20,7 +20,7 @@ use relay_event_schema::processor::{
 use relay_event_schema::protocol::{
     AsPair, Context, ContextInner, Contexts, DeviceClass, Event, EventType, Exception, Headers,
     IpAddr, LogEntry, Measurement, Measurements, NelContext, Request, SpanAttribute, SpanStatus,
-    Tags, TraceContext, User,
+    Tags, User,
 };
 use relay_protocol::{Annotated, Empty, Error, ErrorKind, Meta, Object, Value};
 use smallvec::SmallVec;
@@ -280,7 +280,7 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) -
     .ok();
 
     // Some contexts need to be normalized before metrics extraction takes place.
-    processor::apply(&mut event.contexts, normalize_contexts)?;
+    normalize_contexts(&mut event.contexts);
 
     if config.normalize_spans && event.ty.value() == Some(&EventType::Transaction) {
         // XXX(iker): span normalization runs in the store processor, but
@@ -898,14 +898,17 @@ fn normalize_app_start_spans(event: &mut Event) {
 }
 
 /// Normalizes incoming contexts for the downstream metric extraction.
-fn normalize_contexts(contexts: &mut Contexts, _: &mut Meta) -> ProcessingResult {
-    for annotated in &mut contexts.0.values_mut() {
-        if let Some(ContextInner(Context::Trace(context))) = annotated.value_mut() {
-            normalize_trace_context(context)?
+fn normalize_contexts(contexts: &mut Annotated<Contexts>) {
+    processor::apply(contexts, |contexts, _meta| {
+        for annotated in &mut contexts.0.values_mut() {
+            if let Some(ContextInner(Context::Trace(context))) = annotated.value_mut() {
+                context.status.get_or_insert_with(|| SpanStatus::Unknown);
+            }
         }
-    }
 
-    Ok(())
+        Ok(())
+    })
+    .ok();
 }
 
 /// New SDKs do not send measurements when they exceed 180 seconds.
@@ -944,11 +947,6 @@ fn normalize_units(measurements: &mut Measurements) {
             .unit
             .set_value(Some(stated_unit.or(default_unit).unwrap_or_default()))
     }
-}
-
-fn normalize_trace_context(context: &mut TraceContext) -> ProcessingResult {
-    context.status.get_or_insert_with(|| SpanStatus::Unknown);
-    Ok(())
 }
 
 /// Remove measurements that do not conform to the given config.
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -250,7 +250,8 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) -
             meta.add_error(ErrorKind::InvalidData);
             Err(ProcessingAction::DeleteValueSoft)
         }
-    })?;
+    })
+    .ok();
 
     // Default required attributes, even if they have errors
     normalize_logentry(&mut event.logentry, meta)?;
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -263,7 +263,7 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) -
         normalize_device_class(event);
     }
     normalize_stacktraces(event);
-    normalize_exceptions(event)?; // Browser extension filters look at the stacktrace
+    normalize_exceptions(event); // Browser extension filters look at the stacktrace
     normalize_user_agent(event, config.normalize_user_agent); // Legacy browsers filter
     normalize_measurements(
         event,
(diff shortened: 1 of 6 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -614,32 +626,36 @@ fn normalize_device_class(event: &mut Event) {
 /// Process the last frame of the stacktrace of the first exception.
 ///
 /// No additional frames/stacktraces are normalized as they aren't required for metric extraction.
-fn normalize_stacktraces(event: &mut Event) -> ProcessingResult {
+fn normalize_stacktraces(event: &mut Event) {
```
iker-barriocanal, line 617: We could simplify this function with `get_value!`, but that's out of scope for this PR.

relay-event-normalization/src/stacktrace.rs
```diff
@@ -12,39 +12,38 @@ fn is_url(filename: &str) -> bool {
...
     if let Some(frames) = stacktrace.frames.value_mut() {
         for frame in frames.iter_mut() {
-            processor::apply(frame, process_non_raw_frame)?;
+            process_non_raw_frame(frame);
         }
     }
-
-    Ok(())
 }
+pub fn process_non_raw_frame(frame: &mut Annotated<Frame>) {
```
iker-barriocanal, line 23: This function remains functionally the same, its contents are now wrapped in a `processor::apply` call. The alternative is to wrap the calls to the function in a `processor::apply` that always returns `Ok(())` to satisfy the processor's API, which is unnecessarily complex (vs just a function call).

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -382,18 +376,17 @@ fn is_high_cardinality_transaction(event: &Event) -> bool {
...
-    scrub_identifiers_with_regex(string, &TRANSACTION_NAME_NORMALIZER_REGEX, "*")
+pub(crate) fn scrub_identifiers(string: &mut Annotated<String>) {
+    scrub_identifiers_with_regex(string, &TRANSACTION_NAME_NORMALIZER_REGEX, "*");
 }
 
 fn scrub_identifiers_with_regex(
     string: &mut Annotated<String>,
     pattern: &Lazy<Regex>,
     replacer: &str,
-) -> Result<bool, ProcessingAction> {
```
iker-barriocanal, line 393: This function is only called from the `scrub_identifiers` above and the `Result` is never used, so I removed it instead of adding a `.ok()` to keep an unused return.

relay-event-normalization/src/event.rs
```diff
@@ -270,11 +274,13 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) -
     normalize_breakdowns(event, config.breakdowns_config); // Breakdowns are part of the metric extraction too
 
     processor::apply(&mut event.request, |request, _| {
-        request::normalize_request(request)
-    })?;
+        request::normalize_request(request);
+        Ok(())
+    })
+    .ok();
```
iker-barriocanal: There are a few places where this pattern is repeated (vs just a function call like above), but this is out of scope of this PR.

relay-event-normalization/src/event.rs
```diff
@@ -200,15 +200,17 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) -
     transactions_processor.process_event(event, meta, ProcessingState::root())?;
 
     // Check for required and non-empty values
-    schema::SchemaProcessor.process_event(event, meta, ProcessingState::root())?;
+    schema::SchemaProcessor
+        .process_event(event, meta, ProcessingState::root())
+        .ok();
```
Dav1dde: Can we not make the processor not return a `Result` here, just because the processor currently never returns an error, doesn't mean it will forever and we have no guarantee for this.

Edit: This seems to be a pattern throughout the entire PR.

Dav1dde: Personally I really don't like ignoring errors via `.ok()` this is just a combinator that returns an `Option` instead of `Result` which happens to silence the `must_use`. The `let _ = ...` syntax is a lot more obvious imo.

Dav1dde (commented):
Instead of swallowing potential errors that never exist in quite a few places, can we not get rid of the source of the problem and make these functions not fallible in the first place?

There seem to be only very few offenders, 2x processors and `processor::apply`.

This would mean in the future no one can change the implementation of those functions and expect the result to actually do something.

iker-barriocanal:
> Instead of swallowing potential errors that never exist in quite a few places, can we not get rid of the source of the problem and make these functions not fallible in the first place?

Not sure if I understand the suggestion here. The [processors](https://github.com/getsentry/relay/blob/478c3ada166b108464a0971827433a8db090935e/relay-event-schema/src/processor/traits.rs#L54) and `processor::apply` return a `ProcessingResult`, and you either need to use the result or refactor it not to return it. Refactoring the processors is out of scope. Since at the top level the `ok` variant is always returned (see below), swallowing it should be ok.

The usage of `processor:apply` (and reading easiness) is not great in every case, but this is a different problem I'm not trying to solve right now.

> This would mean in the future no one can change the implementation of those functions and expect the result to actually do something.

These functions always return `Ok(())`, but when [rejecting events](https://github.com/getsentry/relay/blob/478c3ada166b108464a0971827433a8db090935e/relay-event-schema/src/processor/funcs.rs#L26). The next PR will split a validation function to reject events, so the current normalization function will not reject them, resulting in the `Ok(())` variant always being used.

@Dav1dde what do you think?

Dav1dde:
> Not sure if I understand the suggestion here. The [processors](https://github.com/getsentry/relay/blob/478c3ada166b108464a0971827433a8db090935e/relay-event-schema/src/processor/traits.rs#L54) and processor::apply return a ProcessingResult, and you either need to use the result or refactor it not to return it. Refactoring the processors is out of scope. Since at the top level the ok variant is always returned (see below), swallowing it should be ok.

Why is it out of scope? I feel like making the Result error type for the processor an associated type and making the 2 affected processors return `Infallible` should be relatively straight forward. For `processor::apply` we could just temporarily have a second function which does not return a Result (if there are uses where we still need to Result, otherwise could just make it infallible).

Dav1dde (approved):
Looked into it more, there would be a lot more changes than I anticipated necessary to make the processing infallible where we do not return actions anyways.

As noted I would prefer to ignore results via `let _ = ...` but I am not sure what Relay convention here is.

relay-event-normalization/src/event.rs
iker-barriocanal: I don't think there is a convention in Relay for this, but I agree `.ok()` is not very obvious. Addressed in [e1e008d](https://github.com/getsentry/relay/pull/2935/commits/e1e008dbabbe87140fdfaef9cdda5d34c148ae4b).

iker-barriocanal: As discussed offline (posting here for future reference): this is a great idea but requires a bigger refactor, and that is too much work to get into at this point.

PR #2936, opened 2024-01-11
ref(server): Submit envelopes directly from processor

jan-auer:
Removes circular roundtrips between the envelope processor service and
the now obsolete envelope manager by encoding and submitting envelopes
directly in the processor.

Previously, `EnvelopeManager` was the service responsible to supervise
the lifecycle of an envelope. This included dispatching it to the right
backend (Kafka or HTTP upstream) and logging outcomes in the event of an
error. However, in order to submit envelopes via HTTP, they
conditionally need to be compressed, which was sent back to the envelope
processor. The overall chain was therefore Processor -> manager ->
processor -> manager -> upstream -> manager. Additionally, the manager
had to maintain a long-running future to await the result of sending the
envelope upstream.

With this refactor, the envelope manager no longer exists. Instead, the
processor directly encodes the envelope and submits a message to the
correct backend. As a positive side-effect of this change, encoding of
large envelopes no longer blocks the main runtime. If submission fails,
either the request's `respond` method logs outcomes, or in case the
message is lost `ManagedEnvelope::drop` does so.

The public `SubmitEnvelope` message still exists and has moved to
`EnvelopeProcessor`, which allows other services to send envelopes.

#skip-changelog


relay-server/src/actors/mod.rs
```diff
@@ -28,7 +27,7 @@
 //! Controller::run(|| Server::start())
 //!     .expect("failed to start relay");
 //! ```
-pub mod envelopes;
+// pub mod envelopes;
```
Dav1dde: Leftover

relay-server/src/actors/processor.rs
```diff
@@ -1105,10 +1117,8 @@ impl EnvelopeProcessorService {
 
         match result {
             Ok(response) => {
-                if let Some(managed_envelope) = response.envelope {
-                    self.inner.envelope_manager.send(SubmitEnvelope {
-                        envelope: managed_envelope,
-                    })
+                if let Some(envelope) = response.envelope {
+                    self.handle_submit_envelope(SubmitEnvelope { envelope });
```
Dav1dde, line 1119: Should we instead send this message to ourselves? 

Metrics collected in the "event loop" for the `SubmitEnvelope` are incorrect now.
It's a little awkward since we don't currently have our own address.

relay-server/src/actors/processor.rs
```diff
@@ -1274,6 +1284,77 @@ impl EnvelopeProcessorService {
...
+            scoping,
+        } = message;
+
+        let upstream = self.inner.config.upstream_descriptor();
+        let dsn = PartialDsn::outbound(&scoping, upstream);
+
+        let mut envelope = Envelope::from_request(None, RequestMeta::outbound(dsn));
+        for client_report in client_reports {
+            let mut item = Item::new(ItemType::ClientReport);
+            item.set_payload(ContentType::Json, client_report.serialize().unwrap()); // TODO: unwrap OK?
```
Dav1dde, line 1352: re TODO: I would say so, we're serializing our own datastructure and this is already common practice in the rest of Relay.

relay-server/src/actors/processor.rs
```diff
@@ -1274,6 +1282,77 @@ impl EnvelopeProcessorService {
...
     }
 
+    fn handle_submit_envelope(&self, message: SubmitEnvelope) {
+        let SubmitEnvelope { mut envelope } = message;
+
+        #[cfg(feature = "processing")]
+        if self.inner.config.processing_enabled() {
+            if let Some(store_forwarder) = self.inner.store_forwarder.clone() {
+                relay_log::trace!("sending envelope to kafka");
+                store_forwarder.send(StoreEnvelope { envelope });
```
jan-auer, line 1292: The store forwarder used to return an async response with a potential error. Now it no longer has to be awaited, and instead logs outcomes internally through the managed envelope.

relay-server/src/actors/processor.rs
```diff
@@ -1274,6 +1282,77 @@ impl EnvelopeProcessorService {
...
+            }
+        }
+
+        // if we are in capture mode, we stash away the event instead of forwarding it.
+        if Capture::should_capture(&self.inner.config) {
+            relay_log::trace!("capturing envelope in memory");
+            self.inner
+                .test_store
+                .send(Capture::accepted(envelope.take_envelope()));
+            envelope.accept();
```
jan-auer: This has been pushed into the branch from `EnvelopeManager::handle_submit`

relay-server/src/actors/processor.rs
```diff
@@ -1274,6 +1282,77 @@ impl EnvelopeProcessorService {
...
+            return;
+        }
+
+        relay_log::trace!("sending envelope to sentry endpoint");
+        let http_encoding = self.inner.config.http_encoding();
+        match encode_envelope(envelope.envelope_mut(), http_encoding) {
```
jan-auer: Note that setting the sent_at timestamp (part of the body), serialization, and HTTP encoding are now a single fallible step. In a follow up, we can check if sent_at could be lifted out into a header.

relay-server/src/actors/processor.rs
```diff
@@ -1274,6 +1284,77 @@ impl EnvelopeProcessorService {
...
+        #[cfg(feature = "processing")]
+        if self.inner.config.processing_enabled() {
+            if let Some(store_forwarder) = self.inner.store_forwarder.clone() {
+                relay_log::trace!("sending envelope to kafka");
+                store_forwarder.send(StoreEnvelope { envelope });
+                return;
+            }
+        }
+
+        // if we are in capture mode, we stash away the event instead of forwarding it.
```
```suggestion
        // If we are in capture mode, we stash away the event instead of forwarding it.
```

relay-server/src/actors/processor.rs
```diff
@@ -1274,6 +1282,77 @@ impl EnvelopeProcessorService {
...
+        let dsn = PartialDsn::outbound(&scoping, upstream);
+
+        let mut envelope = Envelope::from_request(None, RequestMeta::outbound(dsn));
+        for client_report in client_reports {
+            let mut item = Item::new(ItemType::ClientReport);
+            item.set_payload(ContentType::Json, client_report.serialize().unwrap()); // TODO: unwrap OK?
+            envelope.add_item(item);
+        }
+
+        let envelope = ManagedEnvelope::standalone(
```
jan-auer, line 1356: As opposed to the previous implementation in `EnvelopeManager`, this must create a managed envelope now to conform to the interface. In the end, the managed envelope is redundant as we're not going to log any outcomes for it, but we will receive more error messages in case the message is dropped unintentionally.

relay-server/src/actors/processor.rs
```diff
@@ -1761,6 +1820,100 @@ fn encode_payload(body: &Bytes, http_encoding: HttpEncoding) -> Result<Bytes, st
     Ok(envelope_body.into())
 }
 
+/// Serializes the given Envelope and applies the configured HTTP encoding.
+///
+/// This function overrides the `sent_at` timestamp stored in the envelope. Since the envelope went
+/// through basic normalization, all timestamps have been corrected. We propagate the new `sent_at`
+/// to allow the next Relay to double-check this timestamp and potentially apply correction again.
+/// This is done as close to sending as possible so that we avoid internal delays.
+fn encode_envelope(
```
Dav1dde: Should we inline this instead? Would make the modification of the envelope's `sent_at` more obvious.

Dav1dde (approved):
Nice!

relay-server/src/actors/processor.rs
jan-auer: Services cannot send messages to themselves yet, we don't expose the address for that internally. I was thinking about this when writing it, but in a way envelope processing now simply also contains serializing the envelope. Otherwise, the actual submission still happens in another service. So, I think this is fine to leave as-is.

relay-server/src/actors/processor.rs
jan-auer: This is copied from the other location, I'd prefer not to change the implementation too much.

relay-server/src/actors/processor.rs
jan-auer: Implemented in 13f4955, what do you think?

relay-server/src/actors/processor.rs
Dav1dde: I just meant we should get rid of the `TODO` and phrased it horribly.

TBS1996:
nice! I remember we talked about this

maybe for a clean-up PR but there's a lot of docs still mentioning envelopemanager now

jan-auer:
@TBS1996 Thanks for pointing that out. I tried to update most docs now, but left the top-level mermaid diagram untouched. As it's completely outdated, this is really best done in a dedicated PR.

PR #2938, opened 2024-01-11
feat(app-start): Collect dynamically loaded library count and size metrics

narsaynorath:
App start transactions are being modified to send the dynamically loaded libraries.

From this we want to capture a count of loaded libraries as well as the size of the libraries.
We chose counter implementations because there isn't a need to represent this using
percentiles, min, or max. The calculation for average can be done in the backend by
dividing the values by the count of events.

For this implementation I added the metric to the measurements field and has to
add some code to retrieve measurements in the span Getter

TODO:
- [ ] Determine which tags I need to collect for these metrics and update the MetricSpecs
- [ ] Find out which DebugImage types are applicable and whether or not there are different access paths for `image_size`
- [x] CHANGELOG.md entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Validate error_id and trace_id vectors in replay deserializer. ([#2931](https://github.com/getsentry/relay/pull/2931))
 - Add a data category for indexed spans. ([#2937](https://github.com/getsentry/relay/pull/2937))
 - Add nested Android app start span ops to span ingestion ([#2927](https://github.com/getsentry/relay/pull/2927))
+- Add size and count metrics for dynamically loaded libraries on app start spans ([#2938](https://github.com/getsentry/relay/pull/2938))
 
 ## 23.12.1
 
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -383,5 +383,39 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
                     .always(), // already guarded by condition on metric
             ],
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "c:spans/count_dynamically_loaded_libraries@none".into(),
+            field: Some("span.measurements.count_dynamically_loaded_libraries".into()),
+            condition: Some(
+                duration_condition.clone() & is_mobile.clone() & app_start_condition.clone(),
+            ),
+            tags: vec![
+                Tag::with_key("span.op")
+                    .from_field("span.sentry_tags.op")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("span.description")
+                    .from_field("span.sentry_tags.description")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("span.group")
+                    .from_field("span.sentry_tags.group")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("transaction")
+                    .from_field("span.sentry_tags.transaction")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("device.class")
+                    .from_field("span.sentry_tags.device.class")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("release")
+                    .from_field("span.sentry_tags.release")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("os.name")
+                    .from_field("span.sentry_tags.os.name")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("environment")
+                    .from_field("span.sentry_tags.environment")
+                    .always(), // already guarded by condition on metric
+            ],
+        },
     ]
 }
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -385,8 +385,42 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
         },
         MetricSpec {
             category: DataCategory::Span,
-            mri: "c:spans/count_dynamically_loaded_libraries@none".into(),
-            field: Some("span.measurements.count_dynamically_loaded_libraries".into()),
+            mri: "c:spans/dynamically_loaded_libraries_count@none".into(),
+            field: Some("span.measurements.dynamically_loaded_libraries_count".into()),
+            condition: Some(
+                duration_condition.clone() & is_mobile.clone() & app_start_condition.clone(),
+            ),
+            tags: vec![
+                Tag::with_key("span.op")
+                    .from_field("span.sentry_tags.op")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("span.description")
+                    .from_field("span.sentry_tags.description")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("span.group")
+                    .from_field("span.sentry_tags.group")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("transaction")
+                    .from_field("span.sentry_tags.transaction")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("device.class")
+                    .from_field("span.sentry_tags.device.class")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("release")
+                    .from_field("span.sentry_tags.release")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("os.name")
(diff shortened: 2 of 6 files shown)


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -490,6 +534,42 @@ fn timestamp_by_op(spans: &[Annotated<Span>], op: &str) -> Option<Timestamp> {
...
+                .sum()
+        })
+}
+
+fn get_image_size(debug_image: &DebugImage) -> Option<Annotated<u64>> {
+    match debug_image {
+        DebugImage::Apple(image) => Some(image.image_size.clone()),
+        DebugImage::Symbolic(image) => Some(image.image_size.clone()),
+        DebugImage::MachO(image) => Some(image.image_size.clone()),
+        // TODO(nar): Add support for Android image types
```
narsaynorath, line 568: @jjbayer I noticed a few of these `DebugImage` types had `image_size` as properties whereas others like `Proguard` only had `other`. Do you know if in the cases of these "other" debug images, the image size could be stored in `other`?

I'm just wondering if you're familiar with this interface and if there's an easy way to optionally return the image size using a predictable access pattern like `image.image_size` or `image.other.image_size`.

I'm also waiting for word back from the SDKs about which of these images types are actually applicable to my use case so this list may grow/shrink, e.g. I don't know which types are typically sent for Android other than Proguard yet

narsaynorath:
@jjbayer If you had a chance, I'd appreciate some early feedback while I iron out some of the details in this PR. I think I've hooked everything up correctly, I just need to figure out the specifics of which tags I'd like to extract as well as ensure which debug image files apply to my use case.

relay-dynamic-config/src/defaults.rs
```diff
@@ -395,5 +395,73 @@ fn span_metrics(is_extract_all: bool) -> impl IntoIterator<Item = MetricSpec> {
                     .always(), // already guarded by condition on metric
             ],
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "c:spans/dynamically_loaded_libraries_count@none".into(),
```
jjbayer, line 400: We call the thing we're counting debug images or [DIFs](https://docs.sentry.io/product/cli/dif/) internally, so I would stick with that name, i.e. `c:spans/dif_count@none` or `c:spans/debug_image_count@none`.

Also make sure to register the name of the new metric in [strings.py](https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py) before merging this PR.

relay-dynamic-config/src/defaults.rs
```diff
@@ -395,5 +395,73 @@ fn span_metrics(is_extract_all: bool) -> impl IntoIterator<Item = MetricSpec> {
...
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("release")
+                    .from_field("span.sentry_tags.release")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("os.name")
+                    .from_field("span.sentry_tags.os.name")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("environment")
+                    .from_field("span.sentry_tags.environment")
+                    .always(), // already guarded by condition on metric
```
jjbayer, line 429: If we're unsure what tags we're going to need, let's start with the bare minimum and only add tags later if we really need them. This will result in fewer rows in clickhouse, i.e. faster queries & faster UX.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -93,6 +95,7 @@ impl SpanTagKey {
             SpanTagKey::FileExtension => "file_extension",
             SpanTagKey::MainTread => "main_thread",
             SpanTagKey::OsName => "os.name",
+            SpanTagKey::CountDynamicallyLoadedImages => "dynamically_loaded_libraries_count",
```
jjbayer, line 98: Same note on naming: I would go for "debug image" everywhere.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -133,7 +136,8 @@ pub struct Config {
     pub max_tag_value_size: usize,
 }
 
-/// Extracts tags from event and spans and materializes them into `span.data`.
+/// Extracts tags from event and spans and materializes them into `span.sentry_tags`,
+/// or `span.measurements`.
```
jjbayer, line 140: Thanks!

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -165,6 +171,44 @@ pub(crate) fn extract_span_tags(event: &mut Event, config: &Config) {
                 .collect(),
         );
 
+        // Only add dynamically loaded library information to root app start spans in mobile
+        if is_mobile {
+            if let Some(span_op) = span.op.value() {
+                if let Some(description) = span.description.value() {
+                    if (span_op.starts_with("app.start."))
+                        && (description == "Cold Start" || description == "Warm Start")
```
jjbayer, line 179: Might be worth putting this in a `is_app_start` helper function.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -490,6 +534,42 @@ fn timestamp_by_op(spans: &[Annotated<Span>], op: &str) -> Option<Timestamp> {
         .and_then(|span| span.timestamp.value().copied())
 }
 
+/// Counts the number of dynamically loaded images from debug meta.
+fn count_images(debug_meta: &Annotated<DebugMeta>) -> Option<f64> {
```
jjbayer, line 538: nit: I would convert to `f64` at the latest possible point, i.e. return `Option<usize>` here.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: Yeah check with the SDK folks if they send `image_size` (or look it up in a payload). If so, we should add it as a field to `ProguardDebugImage`. `other` is only there so that outdated customer relays can forward a new field.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -490,6 +534,42 @@ fn timestamp_by_op(spans: &[Annotated<Span>], op: &str) -> Option<Timestamp> {
...
+    debug_meta
+        .value()
+        .and_then(|debug_meta| debug_meta.images.value())
+        .map(|images| images.len() as f64)
+}
+
+/// Sum the sizes of dynamically loaded images from debug meta.
+fn sum_images_size(debug_meta: &Annotated<DebugMeta>) -> Option<f64> {
```
jjbayer, line 546: nit: I would convert to f64 at the latest possible point, i.e. return `Option<u64>` here. Unless we expect values so big that integer addition overflows.

relay-event-schema/src/protocol/span.rs
```diff
@@ -159,6 +159,8 @@ impl Getter for Span {
                     val.into()
                 } else if let Some(key) = path.strip_prefix("sentry_tags.") {
                     self.sentry_tags.value()?.get(key)?.as_str()?.into()
+                } else if let Some(key) = path.strip_prefix("measurements.") {
+                    self.measurements.value()?.get_value(key)?.into()
```
jjbayer, line 163: Realized after yesterday that we should insert an extra level here, i.e. `measurements.foo.value` and `measurements.foo.unit`. Just in case we ever want to access the unit in a condition.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Validate error_id and trace_id vectors in replay deserializer. ([#2931](https://github.com/getsentry/relay/pull/2931))
 - Add a data category for indexed spans. ([#2937](https://github.com/getsentry/relay/pull/2937))
 - Add nested Android app start span ops to span ingestion ([#2927](https://github.com/getsentry/relay/pull/2927))
+- Add size and count metrics for dynamically loaded libraries on app start spans ([#2938](https://github.com/getsentry/relay/pull/2938))
```
jjbayer, line 25:
```suggestion
- Add size and count metrics for dynamically loaded libraries on app start spans. ([#2938](https://github.com/getsentry/relay/pull/2938))
```

narsaynorath:
We've decided not to collect this metric at this time to simplify the app start workflow.

PR #2940, opened 2024-01-12
Forward profiles of non-sampled transactions

jjbayer:
Slowest function aggregation is a metrics-like profiling feature that requires symbolicated frames. Because of this dependency, it cannot (yet) be implemented as metrics aggregation in relay. Instead, forward all profiles to sentry, even if they would otherwise be dropped by dynamic sampling.

This feature is still experimental and guarded by a feature flag + global options.

This PR replaces https://github.com/getsentry/relay/pull/2374.


relay-server/src/services/processor.rs
```diff
@@ -499,6 +499,30 @@ impl<'a> ProcessEnvelopeState<'a> {
...
+            .and_then(|e| e.platform.as_str())
+            .unwrap_or("");
+
+        self.project_state
+            .has_feature(Feature::IngestUnsampledProfiles)
+            && global_options
+                .profile_metrics_allowed_platforms
+                .iter()
+                .any(|s| s == event_platform)
+            && rand::random::<f32>() < global_options.profile_metrics_sample_rate
```
jjbayer: @viglia I did not know how strict this check should be, so I defaulted to strictest possible (AND-condition). It's up to you to decide how you want to handle this (e.g. should a project with the feature flag enabled respect the sample rate? Or is the sample rate just for projects that do not have the feature flag enabled? Etc.)

tests/integration/test_dynamic_sampling.py
```diff
@@ -589,3 +590,185 @@ def test_relay_chain(
...
+
+    envelope = make_envelope(public_key)
+    relay.send_envelope(project_id, envelope)
+    envelope = mini_sentry.captured_events.get(timeout=1)
+
+    print(f"Envelope: {envelope}")
+    # profiles = list(
+    #     filter(lambda item: item.data_category == "profile", envelope.items)
+    # )
+    # print(f"Profiles: {json.loads(profiles[0].payload.bytes.decode('utf-8'))}")
```
jjbayer: @viglia With the changes I made, you should be able to continue work on this test.

relay-server/src/services/processor.rs
viglia: @jjbayer, that's correct. It should be the strictest possible.

So we want: 
1. global_options.unsampled_profiles_enabled == True
2. feature flag enabled
3. platform being allowed
4. sampling

tests/integration/test_dynamic_sampling.py
viglia: @jjbayer thank you 👍🏻

viglia: done!

tests/integration/test_dynamic_sampling.py
```diff
@@ -589,3 +599,198 @@ def test_relay_chain(
...
+
+    relay.send_envelope(project_id, envelope)
+    envelope = mini_sentry.captured_events.get(timeout=1)
+    envelope = mini_sentry.captured_events.get(timeout=1)
+
+    profiles = list(
+        filter(lambda item: item.data_category == "profile", envelope.items)
+    )
+    assert len(profiles) == 1
+    assert not profiles[0].headers.get("sampled")
```
jjbayer: We're looking for a very specific value, so I would make this check more strict:

```suggestion
    assert profiles[0].headers["sampled"] is False
```

relay-dynamic-config/src/feature.rs
```diff
@@ -42,6 +42,11 @@ pub enum Feature {
     /// Enable the Relay cardinality limiter.
     #[serde(rename = "organizations:relay-cardinality-limiter")]
     CardinalityLimiter,
+    /// Enable processing and extracting data from non dinamycally sampled profiles.
+    /// Only some data for slowest function aggregation will be used. The profile
+    /// itself won't be stored on GCS.
```
```suggestion
    /// Enable processing and extracting data from profiles that would normally be dropped by dynamic sampling.
    /// This is required for slowest function aggregation. The profile payload will be dropped on the sentry side.
```

jjbayer: @viglia could you link to slowest function aggregation code here?

relay-dynamic-config/src/global.rs
```diff
@@ -45,11 +45,28 @@ impl GlobalConfig {
 #[serde(default, rename_all = "camelCase")]
 #[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
 pub struct Options {
-    // Example:
-    // ```rs
-    // #[serde(default, rename = "relay.some-option.name")]
-    // pub some_option: Vec<u32>,
-    // ```
+    /// list of platform names for which we allow using unsampled profiles for the purpose
+    /// of improving profile (function) metrics
```
jjbayer:
```suggestion
    /// List of platform names for which we allow using unsampled profiles for the purpose
    /// of improving profile (function) metrics.
```

(same for comments below)

relay-protocol/src/value.rs
```diff
@@ -85,6 +85,14 @@ impl Value {
         }
     }
 
+    /// Returns the bool if this value is a bool, otherwise `None`.
+    pub fn as_bool(&self) -> Option<bool> {
+        match self {
+            Value::Bool(bool) => Some(*bool),
+            _ => None,
+        }
+    }
```
jjbayer: I _think_ we don't need this anymore.

```suggestion
```

jjbayer (commented):
I cannot formally approve, but this looks good to me. When we deploy this, let's do it carefully (`deploy-relay-experimental` first).

viglia (approved):

relay-server/src/services/processor.rs
```diff
@@ -504,6 +504,30 @@ impl<'a> ProcessEnvelopeState<'a> {
         self.remove_event();
         self.managed_envelope.reject_event(outcome);
     }
+
+    fn forward_unsampled_profiles(&self, global_config: &GlobalConfig) -> bool {
```
```suggestion
    /// Determines whether profiles that would otherwise be dropped by dynamic sampling should be kept.
    fn forward_unsampled_profiles(&self, global_config: &GlobalConfig) -> bool {
```

relay-server/src/services/processor.rs
```diff
@@ -1213,15 +1242,13 @@ impl EnvelopeProcessorService {
...
-        &self,
-        state: &mut ProcessEnvelopeState,
-    ) -> Result<(), ProcessingError> {
+    /// Processes standalone items that require an event ID, but do not have an event on the same envelope.
+    fn process_standalone(&self, state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
         profile::filter(state);
 
         if_processing!(self.inner.config, {
             self.enforce_quotas(state)?;
+            // profile::process(state, &self.inner.config); // TODO:
```
```suggestion
```

relay-server/src/services/processor/dynamic_sampling.rs
```diff
@@ -96,17 +97,32 @@ pub fn run(state: &mut ProcessEnvelopeState, config: &Config) {
...
+            let matched_rules = sampling_match.into_matched_rules();
             let outcome = Outcome::FilteredSampling(matched_rules.clone());
-            state
-                .managed_envelope
-                .retain_items(|_| utils::ItemAction::Drop(outcome.clone()));
-
+            state.managed_envelope.retain_items(|item| {
+                if unsampled_profiles_enabled && item.ty() == &ItemType::Profile {
+                    item.set_sampled(false);
+
```
jjbayer:
```suggestion
```

relay-dynamic-config/src/feature.rs
```diff
@@ -40,6 +40,12 @@ pub enum Feature {
     /// Enable the Relay cardinality limiter.
     #[serde(rename = "organizations:relay-cardinality-limiter")]
     CardinalityLimiter,
+    /// Enable processing and extracting data from non dinamycally sampled profiles.
```
Dav1dde:
```suggestion
    /// Enable processing and extracting data from non dynamically sampled profiles.
```

relay-dynamic-config/src/feature.rs
```diff
@@ -40,6 +40,12 @@ pub enum Feature {
...
+    /// Only some data for slowest function aggregation will be used. The profile
+    /// itself won't be stored on GCS.
+    /// `<https://github.com/getsentry/snuba/blob/master/snuba/snuba_migrations/functions/0001_functions.py#L209-L256>`
```
Dav1dde: If we link to code and line numbers, we should link to a specific commit hash imo

relay-dynamic-config/src/global.rs
```diff
@@ -45,11 +45,28 @@ impl GlobalConfig {
...
 pub struct Options {
-    // Example:
-    // ```rs
-    // #[serde(default, rename = "relay.some-option.name")]
-    // pub some_option: Vec<u32>,
-    // ```
+    /// List of platform names for which we allow using unsampled profiles for the purpose
+    /// of improving profile (function) metrics
+    #[serde(
+        default,
```
Dav1dde: The struct is already `serde(default)`, this isn't required.

relay-server/src/services/processor.rs
```diff
@@ -504,6 +504,31 @@ impl<'a> ProcessEnvelopeState<'a> {
...
+            .and_then(|e| e.platform.as_str())
+            .unwrap_or("");
+
+        self.project_state
+            .has_feature(Feature::IngestUnsampledProfiles)
+            && global_options
+                .profile_metrics_allowed_platforms
+                .iter()
+                .any(|s| s == event_platform)
+            && rand::random::<f32>() < global_options.profile_metrics_sample_rate
```
Dav1dde: Should use a `rand::thread_rng` here

relay-server/src/envelope.rs
```diff
@@ -517,6 +517,14 @@ pub struct ItemHeaders {
     #[serde(default, skip_serializing_if = "is_false")]
     metrics_extracted: bool,
 
+    /// `false` if the sampling decision is "drop".
+    ///
+    /// In the most common use case, the item is dropped when the sampling decision is "drop".
+    /// For profiles with the feature enabled, however, we keep all profile items and mark the ones
+    /// for which the transaction was dropped as `sampled: false`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
Dav1dde: `default` is implied for `Option`, but do we actually need `Option<bool>`, can't we use `bool` with a default of `true` and `skip_serializing_if = "if_true"` almost like we have for `metrics_extracted` above.

relay-server/src/services/processor.rs
```diff
@@ -504,6 +504,31 @@ impl<'a> ProcessEnvelopeState<'a> {
         self.remove_event();
         self.managed_envelope.reject_event(outcome);
     }
+
+    /// Determines whether profiles that would otherwise be dropped by dynamic sampling should be kept.
+    fn forward_unsampled_profiles(&self, global_config: &GlobalConfig) -> bool {
```
Dav1dde: I don't think this should be a method on the state. A free standing function in the dynamic sampling module seems more appropriate.

Dav1dde (commented):
Changes make sense to me, but I think this should be reviewed with someone with more insights into our processing.

relay-server/src/services/processor.rs
jjbayer: Done!

relay-server/src/envelope.rs
jjbayer: To me it feels wrong to have a flag that defaults to `sampled: true` even before the sampling decision has been made. If the meaning was flipped, i.e. the flag was called `should_be_dropped`, I would agree.

relay-server/src/services/processor.rs
jjbayer: `random()` is a [convenience wrapper](https://docs.rs/rand/latest/rand/fn.random.html) around `thread_rng()`. Do you mean we should create a dedicated RNG here?

relay-dynamic-config/src/feature.rs
jjbayer: Good catch!

relay-server/src/services/processor.rs
Dav1dde: oh TIL thanks 👍

relay-server/src/envelope.rs
Dav1dde: Fair enough, I don't have a strong opinion here. If you want to keep the `Option` you can remove the `default` from the serde annotation (`Option` always implies `default`).

relay-server/src/envelope.rs
```diff
@@ -739,6 +748,16 @@ impl Item {
         self.headers.metrics_extracted = metrics_extracted;
     }
 
+    /// Gets the `sampled` flag.
+    pub fn sampled(&self) -> Option<bool> {
```
Dav1dde: Could just return a boolean here, would make a check less awkward.

```rs
self.headers.sampled.unwrap_or(true)
```

Dav1dde (approved):

relay-server/src/envelope.rs
jjbayer: I simplified the header itself to a boolean now, as you suggested earlier. Don't like the semantics 100% but it does make everything easier.

relay-server/src/envelope.rs
```diff
@@ -522,14 +522,22 @@ pub struct ItemHeaders {
...
-    sampled: Option<bool>,
+    #[serde(default = "true_", skip_serializing_if = "is_true")]
+    sampled: bool,
 
     /// Other attributes for forward compatibility.
     #[serde(flatten)]
     other: BTreeMap<String, Value>,
 }
 
+fn true_() -> bool {
```
Dav1dde: What do you think?

```suggestion
fn default_true() -> bool {
```

Dav1dde (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Copy event measurements to span & normalize span measurements. ([#2953](https://github.com/getsentry/relay/pull/2953))
 - Add `allow_negative` to `BuiltinMeasurementKey`. Filter out negative BuiltinMeasurements if `allow_negative` is false. ([#2982](https://github.com/getsentry/relay/pull/2982))
 - Add possiblity to block metrics or their tags with glob-patterns. ([#2954](https://github.com/getsentry/relay/pull/2954), [#2973](https://github.com/getsentry/relay/pull/2973))
+- Forward profiles of non-sampled transactions ([#2940](https://github.com/getsentry/relay/pull/2940))
```
iker-barriocanal: nit
```suggestion
- Forward profiles of non-sampled transactions. ([#2940](https://github.com/getsentry/relay/pull/2940))
```

iker-barriocanal (approved):
Good stuff! 🚀

PR #2941, opened 2024-01-15
feat(server): Rate limit buckets by namespace

TBS1996:
part of: https://github.com/getsentry/relay/issues/2716

We want to also have global ratelimits based on namespace. This is done by grouping bucket-vectors by namespace and sending each group individually to check for ratelimiting.


relay-quotas/src/redis.rs
```diff
@@ -129,24 +129,14 @@ impl<'a> RedisQuota<'a> {
 
     /// Returns the key of the quota.
     pub fn key(&self) -> String {
-        if matches!(self.quota.scope, QuotaScope::Global) {
```
TBS1996, line 132: no longer necessary because we construct the redis key separately for global scoped quotas.

tests/integration/test_outcome.py
```diff
@@ -1927,3 +1927,116 @@ def make_envelope():
...
+    send_buckets(1, transaction_name, transaction_value, "d")
+
+    # assert we hit the transaction throughput limit configured.
+    outcomes = outcomes_consumer.get_outcomes()
+    assert len(outcomes) == 1
+    assert outcomes[0]["reason"] == transaction_reason_code
+
+    # Fill up the global limit
+    global_quota_remaining = metric_bucket_limit - transaction_limit
+    send_buckets(global_quota_remaining - 1, session_name, session_value, "s")
```
TBS1996: the '-1' is a problem I'm thinking of a solution for now.  
 
When multiple quotas matches the same item, like here in the case of both a namespace quota and one without, we count down when it matches the quota without namespace even though it gets ratelimited by the namespace quota right after.

TBS1996: The ugly immediate fix is like this:

``` Rust
        global_quotas.sort_by_key(|x| x.namespace.is_none());
        for quota in global_quotas {
            let is_rate_limited = self
                .global_limits
                .is_rate_limited(&mut client, &quota, quantity)
                .map_err(RateLimitingError::Redis)?;

            if is_rate_limited {
                let retry_after = self.retry_after((quota.expiry() - timestamp).as_secs());
                rate_limits.add(RateLimit::from_quota(&quota, &item_scoping, retry_after));
                break;
            }
        }
```

it works by ensuring the more restrictive global quotas come first, this won't scale though.  
 
working on something more generic now

TBS1996: done.

Cargo.lock
```diff
@@ -3853,6 +3853,7 @@ version = "24.1.1"
 dependencies = [
  "hashbrown 0.13.2",
  "insta",
+ "itertools",
```
TBS1996, line 3856: need it for group_by in filter_ratelimited

relay-metrics/src/bucket.rs
```diff
@@ -605,6 +605,11 @@ pub struct Bucket {
 }
 
 impl Bucket {
+    /// Returns the [`MetricNamespace`] of the bucket.
+    pub fn namespace(&self) -> Result<MetricNamespace, ParseMetricError> {
```
jjbayer: nit: If a function has to do work, it's good to reflect it in the name.

```suggestion
    pub fn parse_namespace(&self) -> Result<MetricNamespace, ParseMetricError> {
```

relay-quotas/Cargo.toml
```diff
@@ -22,6 +22,7 @@ relay-redis = { path = "../relay-redis", optional = true }
 serde = { workspace = true }
 smallvec = { workspace = true }
 thiserror = { workspace = true, optional = true }
+itertools = { workspace = true }
```
jjbayer, line 25: TODO: necessary?

relay-quotas/src/global.rs
```diff
@@ -48,40 +70,67 @@ impl GlobalRateLimits {
 struct Key {
     prefix: String,
     window: u64,
+    namespace: Option<MetricNamespace>,
```
jjbayer, line 73: I would make the namespace part of `prefix` here. The global quota logic does not need to know _what_ it is limiting, as long as it has a unique prefix to keep quotas apart.

relay-quotas/src/global.rs
```diff
@@ -16,28 +17,49 @@ fn load_global_lua_script() -> &'static Script {
...
-    /// Returns `Ok(true)` if the global quota should be ratelimited.
+    /// Returns a vector of the [`RedisQuota`]'s that should be ratelimited.
     ///
-    /// Certain errors can be resolved by syncing to redis, so in those cases
-    /// we try again to decrement the budget after syncing.
-    pub fn is_rate_limited(
+    /// We don't know if an item should be ratelimited or not until we've checked all the quotas.
+    /// Therefore we only start decrementing the budgets of the various quotas when we know
+    /// that None of the quotas hit the ratelimit.
+    pub fn filter_ratelimited<'a>(
```
jjbayer:
```suggestion
    pub fn filter_rate_limited<'a>(
```

relay-quotas/src/global.rs
```diff
@@ -48,40 +70,67 @@ impl GlobalRateLimits {
...
     prefix: &'a str,
     window: u64,
+    namespace: Option<MetricNamespace>,
 }
 
 impl<'a> KeyRef<'a> {
-    fn new(quota: &RedisQuota<'a>) -> Self {
+    fn redis_key(&self, slot: u64) -> RedisKey {
+        RedisKey::new(self, slot)
+    }
```
jjbayer: nit: I like having constructor methods (`new`) at the beginning of the `impl` block.

relay-quotas/src/quota.rs
```diff
@@ -340,15 +356,16 @@ mod tests {
 
         let quota = serde_json::from_str::<Quota>(json).expect("parse quota");
 
-        insta::assert_ron_snapshot!(quota, @r#"
+        insta::assert_ron_snapshot!(quota, @r###"
```
jjbayer, line 359: I guess `insta` did this, but I remember the beta linter complained about triple `###` a while back -> check if `cargo + nightly clippy` likes this.

relay-quotas/src/quota.rs
```diff
@@ -51,6 +53,9 @@ pub struct ItemScoping<'a> {
 
     /// Scoping of the data.
     pub scoping: &'a Scoping,
+
+    /// Namespace if applicable. `None` matches any namespace.
+    pub namespace: Option<MetricNamespace>,
```
jjbayer, line 58: I dislike adding the metrics namespace here, but I don't have a better idea at the moment.

relay-quotas/src/redis.rs
jjbayer: See comment about making namespace part of the prefix. With that, I think this code can remain as it is.

jjbayer (commented):
Overall looks good to me, see comment about putting namespace in the prefix.

relay-quotas/Cargo.toml
TBS1996: you mean if the group_by thing is necessary?

relay-quotas/src/global.rs
TBS1996: is it worth the extra string allocation though? We only use `Key` when we put a new key into the hashmap, otherwise we use keyref which is a cheap copy type. If we change the 'Key' here we'd have to do away with the cheap keyref type.

relay-quotas/src/quota.rs
TBS1996: i tried, it's fine

relay-quotas/src/redis.rs
TBS1996: left comment about prefix below

relay-quotas/src/quota.rs
TBS1996: yeah I have the same opinion

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Add possiblity to block metrics or their tags with glob-patterns. ([#2954](https://github.com/getsentry/relay/pull/2954), [#2973](https://github.com/getsentry/relay/pull/2973))
 - Forward profiles of non-sampled transactions. ([#2940](https://github.com/getsentry/relay/pull/2940))
 - Enable throttled periodic unspool of the buffered envelopes. ([#2993](https://github.com/getsentry/relay/pull/2993))
+- Add ability to rate limit buckets by namespace. ([#2941](https://github.com/getsentry/relay/pull/2941))
```
olksdr: Please, move this to `Unreleased` section

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Internal**
+
+- Add ability to rate limit buckets by namespace. ([#2941](https://github.com/getsentry/relay/pull/2941))
```
```suggestion
- Add ability to rate limit metric buckets by namespace. ([#2941](https://github.com/getsentry/relay/pull/2941))
```

relay-quotas/Cargo.toml
jjbayer: Yeah I was wondering whether we could work without the dependency. But it does improve readability, so it's fine with me.

relay-server/src/services/processor.rs
```diff
@@ -1685,29 +1687,53 @@ impl EnvelopeProcessorService {
...
+                    category: DataCategory::MetricBucket,
+                    scoping: &scoping,
+                    namespace: Some(namespace),
+                };
+
+                (!self.rate_limit_buckets(item_scoping, &buckets, quotas, mode, rate_limiter))
+                    .then_some(buckets)
+            })
+            .flatten()
+            .collect()
```
jjbayer, line 1715: We could probably prevent moving buckets around here, by working with a hashmap of references and only recreating the vector at the end. Not a blocker for me though.

jjbayer (approved):
@TBS1996 there's still at least one open comment, please check before merging.

Summary of pair review:

* Sprinkling `namespace` fields across types is not great because it introduces a datatype-specific attribute into a general type. The best approach would be to have a generic matching language for quota, similar to dynamic sampling, but that's out of scope for this PR.
* We discarded the possibility of a new `QuotaScope` `"global-namespace"` and encoding the namespace into the `scope_id`. This would allow us to reuse a lot of code, but it loses the flexibility of configuring per-org or per-project namespace limits.
* We can easily get rid of the namespace field(s) in the future if we want to reconsider, as long as we make sentry stop sending namespace-limits as well.

relay-server/src/services/processor.rs
TBS1996: gave it one try and the `BucketsView` taking in a `&[Bucket]` rather than a `&[&Buckets]` made it harder. Can probably make it work but don't wanna block this PR any longer atm

PR #2943, opened 2024-01-15
ref(metrics): Require finite floats in metric values

jan-auer:
PoC how float validation would look like at runtime without a dedicated type.
Initially this was expected to be less code, but the changes to the `IntoMetric`
trait are rather verbose already.

#skip-changelog


jan-auer:
Closing in favor of https://github.com/getsentry/relay/pull/2958

PR #2945, opened 2024-01-16
feat(ddm): Type metrics_summary fields

### LINKED ISSUE #2922: Type properly the `_metric_summary` field in Relay
Currently `_metric_summary` is untyped (`Value`) in Relay, we now need to normalize contained MRIs, we should now properly type the structure and remove/adjust the normalization code.

iambriccardo:
This PR types the `_metrics_summary` fields in both `Event` and `Span`. Since `Object` keys must be strings, normalization still has to be done manually.

Closes: https://github.com/getsentry/relay/issues/2922

#skip-changelog


relay-event-schema/src/protocol/span.rs
```diff
@@ -86,7 +109,7 @@ pub struct Span {
     /// This shall move to a stable location once we have stabilized the
     /// interface.  This is intentionally not typed today.
     #[metastructure(skip_serialization = "empty")]
-    pub _metrics_summary: Annotated<Value>,
+    pub _metrics_summary: Annotated<Object<Array<MetricSummary>>>,
```
Dav1dde: This should be `Annotated<MetricsSummary>` similar to `Measurements`. `Measurements` also does some validation in the `from_value` which we could also use for `MetricsSummary`.

I think ideally it could look like:

```rs
pub struct MetricsSummary(Map<MetricResourceIdentifier, ...>)
```

With a proper `FromValue` and `IntoValue` implementation.

relay-event-schema/src/protocol/event.rs
```diff
@@ -496,7 +496,7 @@ pub struct Event {
     /// This shall move to a stable location once we have stabilized the
     /// interface.  This is intentionally not typed today.
     #[metastructure(omit_from_schema)]
-    pub _metrics_summary: Annotated<Value>,
+    pub _metrics_summary: Annotated<Object<Array<MetricSummary>>>,
```
Dav1dde: dito typed `_metrics_summary: Annotated<MetricsSummary>`

relay-event-schema/src/protocol/metrics_summary.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// Minimum value of the metric.
+    pub min: Annotated<f64>,
+
+    /// Maximum value of the metric.
+    pub max: Annotated<f64>,
+
+    /// Sum of all metric values.
+    pub sum: Annotated<f64>,
+
+    /// Count of all metric values.
```
iker-barriocanal, line 37: nit(?)
```suggestion
    /// Count of all metrics.
```

relay-event-schema/src/protocol/metrics_summary.rs
```diff
@@ -0,0 +1,33 @@
...
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct MetricsSummary(pub Object<Array<MetricSummary>>);
+
+/// The metric summary of a single metric that is emitted by the SDK.
+///
+/// The summary contains specific aggregate values that the metric had during the span's lifetime. A single span can
+/// have the same metric emitted multiple times, which is the reason for aggregates being computed in each summary.
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct MetricSummary {
```
iker-barriocanal, line 27: Is this only for distribution metrics? If yes, I think it'd be useful to add it in the docstrings.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -219,22 +219,17 @@ fn normalize_app_start_spans(event: &mut Event) {
...
 /// identifiers in the form `metric_type:metric_name@metric_unit` and those identifiers need to be
 /// converted to MRIs in the form `metric_type:metric_namespace/metric_name@metric_unit`.
-fn normalize_metrics_summary_mris(value: &mut Value) {
-    let Value::Object(metrics) = value else {
-        return;
-    };
-
-    let metrics = mem::take(metrics)
+fn normalize_metrics_summary_mris(metrics_summary: &mut MetricsSummary) {
+    metrics_summary.0 = mem::take(&mut metrics_summary.0)
```
iker-barriocanal: What do you think if we add an accessor method to make the API friendlier to use? A `.value()` method in `MetricsSummary` that hides the `mem::take` would be friendlier to deal with. What do you think?

```rs
    metrics_summary.0 = metrics_summary.value().into_iter()...
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -219,22 +219,17 @@ fn normalize_app_start_spans(event: &mut Event) {
...
-            Some((
-                MetricResourceIdentifier::parse(&key).ok()?.to_string(),
-                value,
-            ))
+        .map(|(key, value)| match MetricResourceIdentifier::parse(&key) {
+            Ok(mri) => (mri.to_string(), value),
+            Err(err) => (
+                key,
+                Annotated::from_error(Error::invalid(err), value.0.map(IntoValue::into_value)),
+            ),
```
iker-barriocanal, line 228: Nice improvement!

iker-barriocanal (approved):

relay-event-schema/src/protocol/metrics_summary.rs
Dav1dde: It seems to be for all metrics: https://github.com/getsentry/sentry-python/blob/master/tests/test_metrics.py#L599-L627

relay-event-schema/src/protocol/metrics_summary.rs
Dav1dde: I think it actually counts all metric values.

iambriccardo: Yes it does

relay-event-normalization/src/normalize/mod.rs
iambriccardo: No strong opinions here, @Dav1dde what is your thought?

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -219,28 +218,22 @@ fn normalize_app_start_spans(event: &mut Event) {
...
-
-    let metrics = mem::take(metrics)
-        .into_iter()
-        .filter_map(|(key, value)| {
-            Some((
-                MetricResourceIdentifier::parse(&key).ok()?.to_string(),
-                value,
-            ))
+fn normalize_metrics_summary_mris(m: MetricSummaryMapping) -> MetricSummaryMapping {
+    m.into_iter()
```
iker-barriocanal, line 222: So clean! 💯

PR #2946, opened 2024-01-16
ref(global-config): Introduce `GlobalConfigHandle`

Dav1dde:
Make it possible to pass around an owned handle for the global config, instead of needing to query the current global config or a subscription from the global config service.

#skip-changelog

### DIFF
--- a/relay-server/src/actors/global_config.rs
+++ b/relay-server/src/actors/global_config.rs
@@ -10,6 +10,7 @@
 //! more details.
 
 use std::borrow::Cow;
+use std::fmt;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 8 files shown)


Dav1dde:
I a followup I want to replace the remaining usage of the `Get` and `Subscribe` usages.

relay-server/src/actors/global_config.rs
```diff
@@ -166,6 +167,39 @@ impl Status {
...
+
+    /// Returns the currently loaded or a default global config.
+    ///
+    /// When no global config has been received from upstream yet,
+    /// this will return a default global config.
+    pub fn current(&self) -> Arc<GlobalConfig> {
+        match &*self.watch.borrow() {
+            Status::Ready(config) => Arc::clone(config),
+            Status::Pending => Default::default(),
+        }
```
iker-barriocanal, line 191: IIUC, the purpose of `Pending` is to indicate there's no global config loaded yet, but this accessor is returning one. What do you think if we return `Option<Arc<GlobalConfig>>` instead, and return `None` if the status is pending? (or a similar approach)

Dav1dde: The problem is, how do we actually want to deal with that case? If all we do is `.unwrap_or_default()` then I prefer this.

It's a bit unfortunate that this is so much implicit behaviour, currently the only time we use this (in the processor) this should never be the case since envelopes are delayed until a global config is loaded.

Of course we can always fallback to this pattern:
```
let Some(global_config) = self.global_config.current() else { relay_log::error!(...); return; }
```
But that does make things a lot more annoying to deal with, e.g. what would we return from `light_normalize_event` , an error? Or do we need to go back to passing the global config around everywhere (which I kind of wanted to get rid of).

iker-barriocanal: Hmm, that makes sense. To be honest, I don't like any alternative I can think of nor do I have strong opinions.

iker-barriocanal (approved):

PR #2947, opened 2024-01-16
feat(cardinality): Create outcomes for cardinality limited metrics

Dav1dde:
Creates rate limited outcomes for cardinality limited items.

Maybe we should introduce a new outcome id for the cardinality limiter instead?

### DIFF
--- a/relay-cardinality/src/limiter.rs
+++ b/relay-cardinality/src/limiter.rs
@@ -207,7 +207,6 @@ mod tests {
             source: vec!['a', 'b', 'c', 'd', 'e'],
             rejections: BTreeSet::from([0, 1, 3]),
         };
-        dbg!(limits.rejected().collect::<Vec<_>>());
         assert!(limits.rejected().eq(['a', 'b', 'd'].iter()));
         assert_eq!(limits.into_accepted(), vec!['c', 'e']);
 
--- a/relay-cardinality/src/limiter.rs
+++ b/relay-cardinality/src/limiter.rs
@@ -207,21 +207,27 @@ mod tests {
             source: vec!['a', 'b', 'c', 'd', 'e'],
             rejections: BTreeSet::from([0, 1, 3]),
         };
-        assert!(limits.rejected().eq(['a', 'b', 'd'].iter()));
+        assert_eq!(
+            limits.rejected().copied().collect::<Vec<_>>(),
+            vec!['a', 'b', 'd']
+        );
         assert_eq!(limits.into_accepted(), vec!['c', 'e']);
 
         let limits = CardinalityLimits {
             source: vec!['a', 'b', 'c', 'd', 'e'],
             rejections: BTreeSet::from([]),
         };
-        assert!(limits.rejected().eq([].iter()));
+        assert_eq!(limits.rejected().copied().collect::<Vec<_>>(), vec![]);
         assert_eq!(limits.into_accepted(), vec!['a', 'b', 'c', 'd', 'e']);
 
         let limits = CardinalityLimits {
             source: vec!['a', 'b', 'c', 'd', 'e'],
             rejections: BTreeSet::from([0, 1, 2, 3, 4]),
         };
-        assert!(limits.rejected().eq(['a', 'b', 'c', 'd', 'e'].iter()));
+        assert_eq!(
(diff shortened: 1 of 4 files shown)


relay-server/src/actors/outcome.rs
```diff
@@ -180,6 +183,7 @@ impl Outcome {
         match self {
             Outcome::Filtered(_) | Outcome::FilteredSampling(_) => OutcomeId::FILTERED,
             Outcome::RateLimited(_) => OutcomeId::RATE_LIMITED,
+            Outcome::CardinalityLimited => OutcomeId::RATE_LIMITED,
```
Dav1dde, line 188: Should we introduce a new outcome instead?

relay-server/src/actors/processor.rs
```diff
@@ -1819,27 +1854,15 @@ impl EnvelopeProcessorService {
...
-                }
-            }
-
             let mode = match project_state.config.transaction_metrics {
                 Some(ErrorBoundary::Ok(ref c)) if c.usage_metric() => ExtractionMode::Usage,
                 _ => ExtractionMode::Duration,
             };
 
+            if project_state.has_feature(Feature::CardinalityLimiter) {
+                buckets = self.cardinality_limit_buckets(scoping, buckets, mode);
```
iker-barriocanal, line 1863: Nice!

relay-server/src/actors/outcome.rs
jan-auer: It's not strictly necessary; we could also differentiate it purely by reason code, which is what you're doing now. While it's not ideal that two variants to the same outcome ID, we already have a similar setup for dynamic sampling.

If you were to introduce a new outcome, please split this PR in two parts, and first bump all libraries to avoid errors in the consumers.

jan-auer (approved):

relay-server/src/actors/outcome.rs
iker-barriocanal: I think it'd be interesting to have this data available internally to evaluate how high/low the limits are, and the impact on costs. I don't know if other parts (e.g. consumers?) need to be updated to handle new outcomes.

relay-cardinality/src/limiter.rs
```diff
@@ -141,10 +139,15 @@ impl<T: Limiter> CardinalityLimiter<T> {
 #[derive(Debug)]
 pub struct CardinalityLimits<T> {
     source: Vec<T>,
-    rejections: HashSet<usize>,
+    rejections: BTreeSet<usize>,
 }
 
 impl<T> CardinalityLimits<T> {
+    /// Returns an iterator yielding only rejected items.
+    pub fn rejected(&self) -> impl Iterator<Item = &T> {
```
iker-barriocanal, line 147: Is this only for tests?
```suggestion
    /// Returns an iterator yielding only rejected items.
    #[cfg(test)]
    pub fn rejected(&self) -> impl Iterator<Item = &T> {
```

relay-cardinality/src/limiter.rs
```diff
@@ -202,20 +205,24 @@ mod tests {
     fn test_accepted() {
         let limits = CardinalityLimits {
             source: vec!['a', 'b', 'c', 'd', 'e'],
-            rejections: HashSet::from([0, 1, 3]),
+            rejections: BTreeSet::from([0, 1, 3]),
         };
+        dbg!(limits.rejected().collect::<Vec<_>>());
```
iker-barriocanal:
```suggestion
```

iker-barriocanal (approved):

relay-cardinality/src/limiter.rs
Dav1dde: No, only rejections are turned into outcomes. Only a subset of buckets may be cardinality limited while the rest is passed on.

relay-cardinality/src/limiter.rs
Dav1dde: Oops, thank you!

relay-server/src/actors/outcome.rs
Dav1dde: 👍 let's keep it like this and in the future we can split this into a separate outcome id

olksdr (approved):

PR #2948, opened 2024-01-16
feat(spans): Normalize and copy measurements to segments

phacops:


relay-event-normalization/src/event.rs
```diff
@@ -689,19 +689,38 @@ fn normalize_measurements(
...
+            max_mri_len,
+            event.start_timestamp.0,
+            event.timestamp.0,
+        );
+    }
+}
 
-        compute_measurements(duration_millis, measurements);
+/// Ensure only valid measurements are ingested.
+pub fn normalize_measurements_inner(
```
iker-barriocanal: Could we rename this function? I think it's not great to expose an "inner" function. I'm okay renaming the above to something else.
```suggestion
pub fn normalize_measurements(
```

iker-barriocanal (approved):

PR #2949, opened 2024-01-16
fix(spooler): Proactively move on-disk spool to memory when it fits

olksdr:
To make sure that we switch to in-memory spool faster we can proactively check if the size of the on-disk data fits into memory and move the spooled envelopes in one go.

- this will ensure that we eventually switch to in-memory spool, even though there are still some envelopes left unspooled
- this PR also makes sure to use better estimate of the used disk size for the spooled envelopes
- the test is updated to account for the change mentioned above

related to https://github.com/getsentry/team-ingest/issues/257


relay-server/src/actors/spooler/mod.rs
```diff
@@ -490,6 +527,60 @@ impl OnDisk {
...
+
+        loop {
+            let mut envelopes =
+                sql::delete_and_fetch_all(BATCH_SIZE.min(self.buffer_guard.available() as i64))
+                    .fetch(&self.db)
+                    .peekable();
+            relay_statsd::metric!(counter(RelayCounters::BufferReads) += 1);
+
+            // Stream is empty, we can break the loop, since we read everything by now.
+            if Pin::new(&mut envelopes).peek().await.is_none() {
```
Dav1dde: Should be able to pin on the stack here via [pin!](https://doc.rust-lang.org/std/pin/macro.pin.html)

relay-server/src/actors/spooler/mod.rs
```diff
@@ -443,12 +482,10 @@ impl OnDisk {
...
             // but only one batch, and the rest of them will stay on disk for the next iteration
             // to pick up.
-            //
-            // Right now we use 100 for batch size.
-            let batch_size = 100;
-            let mut envelopes = sql::delete_and_fetch(key, batch_size)
-                .fetch(&self.db)
-                .peekable();
+            let mut envelopes =
+                sql::delete_and_fetch(key, BATCH_SIZE.min(self.buffer_guard.available() as i64))
```
Dav1dde: Maybe move `BATCH_SIZE.min(self.buffer_guard.available() as i64)` into a separate method, it's used twice and then we can also give it a short doc string

relay-server/src/actors/spooler/mod.rs
```diff
@@ -490,6 +527,60 @@ impl OnDisk {
...
+                let envelope = match envelope {
+                    Ok(envelope) => envelope,
+
+                    // Bail if there are errors in the stream.
+                    Err(err) => {
+                        relay_log::error!(
+                            error = &err as &dyn Error,
+                            "failed to read the buffer stream from the disk",
+                        );
+                        self.track_count(-count);
```
Dav1dde: How does this work here, if the stream contains an error:

* does it end or can we just skip over the error?
* can there be more rows deleted than returned?

The count may not be correct here then, also we maybe shouldn't stop on error but just skip over them?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -490,6 +527,60 @@ impl OnDisk {
...
+    /// Unspools the entire contents of the on-disk spool.
+    async fn delete_and_fetch_all(
+        &mut self,
+        services: &Services,
+    ) -> Result<BTreeMap<QueueKey, Vec<ManagedEnvelope>>, BufferError> {
+        let mut result: BTreeMap<QueueKey, Vec<ManagedEnvelope>> = BTreeMap::new();
+
+        loop {
+            let mut envelopes =
+                sql::delete_and_fetch_all(BATCH_SIZE.min(self.buffer_guard.available() as i64))
```
Dav1dde: What if `buffer_guard.available()` keeps returning `0`, aren't we just spinning in this loop?

I assume this can't happen because we check for the watermark before calling this function?

Dav1dde (commented):
I think we should have a changelog entry for this, people in OPS may notice a different behaviour.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -636,17 +727,51 @@ impl BufferState {
                         "failed to spool the in-memory buffer to disk",
                     );
                 }
+                relay_log::trace!(
+                    "Transition to disk spool: # of envelopes = {}",
```
iker-barriocanal, line 745: How is this number of envelopes calculated? IIRC it was an estimation and not a super accurate number, I think we should note that in the trace message.

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: IIUC, not skipping a stream error may result in losing all envelopes?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -636,17 +727,51 @@ impl BufferState {
...
+                );
+
                 Self::Disk(disk)
             }
-            Self::Disk(disk) if disk.is_empty().await.unwrap_or_default() => {
-                Self::MemoryFileStandby {
-                    ram: InMemory::new(config.spool_envelopes_max_memory_size()),
-                    disk,
+            Self::Disk(mut disk) if Self::is_below_low_mem_watermark(config, &disk).await => {
+                match disk.delete_and_fetch_all(services).await {
```
iker-barriocanal, line 752: Is 30%, although low, an arbitrary number? What's the behavior if Relay attempts to unspool all envelopes to the memory when it goes into an outage and receives many envelopes?

What do you think about approaching this the other way around? Instead of "there now is x% of memory available, spool everything", a continuous check of "while there's a y% of memory left, unspool a few".

relay-server/src/actors/spooler/mod.rs
```diff
@@ -396,10 +425,19 @@ impl OnDisk {
             .try_get("received_at")
             .map_err(BufferError::FetchFailed)?;
         let start_time = StartTime::from_timestamp_millis(received_at as u64);
```
TBS1996, line 433: isnt it better to fetch received_at as u64 from the DB, i guess that would return an error rather than converting to huge number if negative?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -636,17 +727,51 @@ impl BufferState {
...
         }
     }
+
+    /// Returns `true` if the on-disk spooled data can fit in the memory.
+    ///
+    /// The spooled envelopes must:
+    /// * fit into memory and take not more than 30% of the configured space
+    /// * the used buffer guards also must be under the low watermark.
+    async fn is_below_low_mem_watermark(config: &Config, disk: &OnDisk) -> bool {
+        ((config.spool_envelopes_max_memory_size() as f64 * 0.3) as i64)
```
TBS1996: maybe ratio could be a const

relay-server/src/actors/spooler/mod.rs
```diff
@@ -396,10 +425,19 @@ impl OnDisk {
             .try_get("received_at")
             .map_err(BufferError::FetchFailed)?;
         let start_time = StartTime::from_timestamp_millis(received_at as u64);
+        let own_key: &str = row.try_get("own_key").map_err(BufferError::FetchFailed)?;
+        let sampling_key: &str = row
+            .try_get("sampling_key")
```
TBS1996, line 436: not really this PR, but thoughts on an enum that represents the columns? it could be a nice documentation for how the DB looks like. I assume try_get would accept it if it implements display

relay-server/src/actors/spooler/mod.rs
Dav1dde: I assume this is because SQLite does not know the `u64` datatype only `i64`

relay-server/src/actors/spooler/mod.rs
olksdr: done in [`b2b35f2`](https://github.com/getsentry/relay/pull/2949/commits/b2b35f2b50f796378f0b84da90c7061eaa9474b6)

relay-server/src/actors/spooler/mod.rs
olksdr: done in [`b2b35f2`](https://github.com/getsentry/relay/pull/2949/commits/b2b35f2b50f796378f0b84da90c7061eaa9474b6)

relay-server/src/actors/spooler/mod.rs
olksdr: skipping the error now and trying to read out all the envelopes instead , done in [`b2b35f2`](https://github.com/getsentry/relay/pull/2949/commits/b2b35f2b50f796378f0b84da90c7061eaa9474b6)

relay-server/src/actors/spooler/mod.rs
olksdr: added another check to make sure we won't have this situation, done in [`b2b35f2`](https://github.com/getsentry/relay/pull/2949/commits/b2b35f2b50f796378f0b84da90c7061eaa9474b6)

relay-server/src/actors/spooler/mod.rs
olksdr: This number here should be correct

relay-server/src/actors/spooler/mod.rs
olksdr: I do not follow your comment.  If I get you correctly, then the relay should not unspool anything if in the outage mode 

Could you, please, put a code example what you mean, maybe to illustrate the problem?

relay-server/src/actors/spooler/mod.rs
olksdr: this is not supported by sqlx, only `i64` in this case

relay-server/src/actors/spooler/mod.rs
olksdr: done in [`b2b35f2`](https://github.com/getsentry/relay/pull/2949/commits/b2b35f2b50f796378f0b84da90c7061eaa9474b6)

relay-server/src/actors/spooler/mod.rs
olksdr: maybe we can look into this in the followups 👍

relay-server/src/actors/spooler/mod.rs
```diff
@@ -59,7 +59,16 @@ use crate::utils::{BufferGuard, ManagedEnvelope};
...
 
-/// The set of errors which can happend while working the the buffer.
+/// The predefined batch size for the SQL queries, when fetching anything from the on-disk spool.
+const BATCH_SIZE: i64 = 200;
+
+/// The low memory watermark for spool.
+///
+/// This number is used to calculate how much memory should be taken by the on-disk spool if all
+/// the spooled envelopes will be moved to in-memory buffer.
+const LOW_SPOOL_MEMORY_WATERMARCK: f64 = 0.3;
```
Dav1dde:
```suggestion
const LOW_SPOOL_MEMORY_WATERMARK: f64 = 0.3;
```

Dav1dde (approved):
:shipit:

relay-server/src/actors/spooler/mod.rs
iker-barriocanal: Let's consider the following scenario: Relay has a low memory usage > starts unspooling > loses connection with upstream > plenty of envelopes get to Relay > spooler is still unspooling > memory usage is too high > not enough time to spool envelopes > start dropping them. Could this happen?

olksdr: the message processing is sequential, if there will be a problem again, the Relay will start spool again and will switch to disk eventually

PR #2950, opened 2024-01-16
feat(inbound-filters): Create new legacy browser filters

### LINKED ISSUE #63053: Update Relay to include new filters
Relay will need to be updated to have the new filters

roggenkemper:
this pr adds in new legacy browser filters. These new filters will be more up to date than the existing ones, and would allow for easier updates to the filters if needed. right now, the filters are defined as both a browser and a version ("safari_pre_6"). now the filters will just be for a browser, and we will have easier control over what versions to filter- we no longer have to make a new filter, we can just update what values get filtered. this can both be a short term and longer term fix to the problem with legacy browsers. It will allow us to move over to new filters in the short term, while supporting updates in the long term if needed (since the generic filters project doesn't have a timeline right now).

closes https://github.com/getsentry/sentry/issues/63053


roggenkemper:
@getsentry/ingest let me know your thoughts on this pr! was going off of previous PRs that added new filters since this isn't my area of expertise.

One question I have: Is it safe to merge this before any of the `sentry` changes that actually add in the ability to turn on these filters? or do we need to wait until the `sentry` changes to the settings/endpoint are merged in?

jjbayer (commented):
This looks good to me, @roggenkemper would you mind adding some tests? There should be some you can copy-paste in `legacy_browsers.rs`.

jjbayer:
> Is it safe to merge this before any of the `sentry` changes that actually add in the ability to turn on these filters?

That should be safe AFAIK.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Adds new legacy browser filters. ([#2950](https://github.com/getsentry/relay/pull/2950))
```
```suggestion
- Add new legacy browser filters. ([#2950](https://github.com/getsentry/relay/pull/2950))
```

jjbayer (approved):

PR #2953, opened 2024-01-16
feat(spans): Normalize and copy measurements to segments

phacops:
This will:
- Add measurements normalization to the span normalization function
- Copy measurements to the segment


jjbayer (changes requested):
As discussed, instead of running normalization on spans extracted from events twice, let's make `SpanKafkaMessage` as strict as the kafka schema and "normalize" the message before producing, like we do here:

https://github.com/getsentry/relay/blob/7970a7c13873b96b1084558f7e1ed9b89ec4444e/relay-server/src/services/processor/span/processing.rs#L398-L408

We should probably also make the numeric values in `SpanKafkaMessage` the new [`FiniteF64`](https://github.com/getsentry/relay/pull/2958), because `serde_json` will serialize infinity / `NaN` as `null` as well.

relay-event-normalization/src/event.rs
```diff
@@ -694,7 +694,7 @@ fn normalize_user_agent(_event: &mut Event, normalize_user_agent: Option<bool>)
 }
 
 /// Ensure measurements interface is only present for transaction events.
-fn normalize_measurements(
+fn normalize_measurements_from_event(
```
jjbayer: nit: 
```suggestion
fn normalize_event_measurements(
```

relay-event-normalization/src/event.rs
```diff
@@ -703,19 +703,38 @@ fn normalize_measurements(
...
+            measurements_config,
+            max_mri_len,
+            event.start_timestamp.0,
+            event.timestamp.0,
+        );
+    }
+}
 
-        compute_measurements(duration_millis, measurements);
+/// Ensure only valid measurements are ingested.
```
jjbayer, line 717:
```suggestion
/// Ensures only valid measurements are ingested.
```

relay-server/src/services/processor/span/processing.rs
```diff
@@ -20,24 +23,50 @@ use crate::services::outcome::{DiscardReason, Outcome};
...
 
-pub fn process(state: &mut ProcessEnvelopeState, config: Arc<Config>) {
+fn get_normalize_span_config<'a>(
+    config: Arc<Config>,
+    received_at: DateTime<Utc>,
+    global_measurements_config: Option<&'a MeasurementsConfig>,
+    project_measurements_config: Option<&'a MeasurementsConfig>,
+) -> NormalizeSpanConfig<'a> {
+    let transaction_aggregator_config =
+        AggregatorConfig::from(config.aggregator_config_for(MetricNamespace::Transactions));
```
jjbayer: I probably messed this up initially:

For transactions, this normalization step ensures that the payload aligns with the limits of the metrics aggregator. Span metrics potentially go into a different aggregator, so we should align with that.

```suggestion
    let aggregator_config =
        AggregatorConfig::from(config.aggregator_config_for(MetricNamespace::Spans));
```

relay-server/src/services/processor/span/processing.rs
```diff
@@ -20,24 +23,50 @@ use crate::services::outcome::{DiscardReason, Outcome};
 use crate::services::processor::{ProcessEnvelopeState, ProcessingError};
 use crate::utils::ItemAction;
 
-pub fn process(state: &mut ProcessEnvelopeState, config: Arc<Config>) {
+fn get_normalize_span_config<'a>(
```
jjbayer: nit: Move this private function below the `pub`.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -413,6 +467,73 @@ fn validate(mut span: Annotated<Span>) -> Result<Annotated<Span>, anyhow::Error>
     if let Some(tags) = tags.value_mut() {
         tags.retain(|_, value| !value.value().is_empty())
     }
+    if let Some(measurements) = measurements.value_mut() {
+        measurements.retain(|_, value| match value.value() {
+            Some(v) => !v.is_empty() && !v.value.is_empty(),
```
jjbayer: Is the measurement `unit` is allowed to be empty? This condition expands to

```rust
!v.value.is_empty() && !v.unit.is_empty() && !v.value.is_empty()
```

so I think you can simplify it.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -413,6 +467,73 @@ fn validate(mut span: Annotated<Span>) -> Result<Annotated<Span>, anyhow::Error>
...
+            None => false,
+        });
+    }
+    if let Some(metrics_summary) = _metrics_summary.value_mut() {
+        metrics_summary.0.retain(|_, value| match value.value() {
+            Some(v) => !v.is_empty() && v.iter().all(|v| !v.is_empty()),
```
jjbayer: Are the individual fields of a metrics summary allowed to be empty? (`min`, `max`, etc.). Do we need explicit checks on those?

https://github.com/getsentry/relay/blob/27963b14672270d8cb3645e6ca6571479d2afb34/relay-event-schema/src/protocol/metrics_summary.rs#L27-L42

relay-server/src/services/processor/span/processing.rs
jjbayer: We might also want an `is_finite` check on `measurement.value`, because `serde_json` serializes infinites as `null`.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -20,24 +23,50 @@ use crate::services::outcome::{DiscardReason, Outcome};
...
+                .max_name_length
+                .saturating_sub(MeasurementsConfig::MEASUREMENT_MRI_OVERHEAD),
+        ),
+    }
+}
+
+pub fn process(
+    state: &mut ProcessEnvelopeState,
+    config: Arc<Config>,
+    global_measurements_config: Option<MeasurementsConfig>,
```
jjbayer: There should be no reason to pass this by value.

```suggestion
    global_measurements_config: Option<&MeasurementsConfig>,
```

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 - Validate error_id and trace_id vectors in replay deserializer. ([#2931](https://github.com/getsentry/relay/pull/2931))
 - Add a data category for indexed spans. ([#2937](https://github.com/getsentry/relay/pull/2937))
 - Add nested Android app start span ops to span ingestion ([#2927](https://github.com/getsentry/relay/pull/2927))
+- Normalize and copy measurements on spans when appropriate. ([#2953](https://github.com/getsentry/relay/pull/2953))
```
```suggestion
- Copy event measurements to span & normalize span measurements. ([#2953](https://github.com/getsentry/relay/pull/2953))
```

jjbayer (approved):
Please see comments before merging. No blockers, but we still cannot be 100% sure that this won't generate new validation errors on the consumer side. To prevent that, I think we need to make `SpanKafkaMessage` more strict (this would sacrifice some performance) **and / or** make the consumer more lenient. See also [Sentry's development philosophy](https://develop.sentry.dev/philosophy/#correctness-can-be-a-trap):

> “be conservative in what you send, be liberal in what you accept”.

The second bullet point in the PR description can be removed I believe.

relay-server/src/services/processor/span/processing.rs
phacops: The field can be omitted but a `null` value can't be ingested.

relay-server/src/services/processor/span/processing.rs
phacops: They can be omitted but not set to `null`.

phacops: The Snuba consumers will require a value though, which is not compatible with the schema. I'll fix that in there though.

phacops: https://github.com/getsentry/snuba/pull/5428

relay-server/src/services/store.rs
```diff
@@ -1189,13 +1238,13 @@ struct SpanKafkaMessage<'a> {
     is_segment: bool,
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    measurements: Option<&'a RawValue>,
+    measurements: Option<BTreeMap<&'a str, Option<SpanMeasurement>>>,
```
jjbayer: Measurement keys _should_ be able to parse as borrowed, but I would not risk it:

```suggestion
    measurements: Option<BTreeMap<&'a RawValue, Option<SpanMeasurement>>>,
```

relay-server/src/services/store.rs
```diff
@@ -1189,13 +1238,13 @@ struct SpanKafkaMessage<'a> {
...
     #[serde(
         default,
         rename = "_metrics_summary",
         skip_serializing_if = "Option::is_none"
     )]
-    metrics_summary: Option<&'a RawValue>,
+    metrics_summary: Option<BTreeMap<&'a str, Option<Vec<Option<SpanMetricsSummary>>>>>,
```
jjbayer:
```suggestion
    metrics_summary: Option<BTreeMap<&'a RawValue, Option<Vec<Option<SpanMetricsSummary>>>>>,
```

tests/integration/test_store.py
```diff
@@ -1812,3 +1812,65 @@ def test_span_extraction_with_ddm(
...
+    event["_metrics_summary"] = metrics_summary
+
+    relay.send_event(project_id, event)
+
+    start_timestamp = datetime.fromisoformat(event["start_timestamp"])
+    end_timestamp = datetime.fromisoformat(event["timestamp"])
+    duration_ms = int((end_timestamp - start_timestamp).total_seconds() * 1e3)
+
+    metrics_summary["c:spans/some_metric@none"][0].pop("min", None)
+    metrics_summary["c:spans/some_metric@none"][0]["tags"].pop("random", None)
```
jjbayer: How does this line contribute to the test?

tests/integration/test_store.py
```diff
@@ -1812,3 +1812,65 @@ def test_span_extraction_with_ddm(
...
+                "min": None,
+                "max": 2.0,
+                "count": 4,
+                "tags": {
+                    "environment": "test",
+                },
+            },
+        ],
+    }
+    event["_metrics_summary"] = metrics_summary
```
jjbayer, line 1845: Wanna add some `measurements` to this test event as well?

relay-server/src/services/store.rs
```diff
@@ -865,6 +865,35 @@ impl StoreService {
             }
         };
 
+        if let Some(measurements) = &mut span.measurements {
+            measurements.retain(|_, v| {
+                if let Some(v) = v {
+                    v.value.map_or(false, f64::is_finite)
+                } else {
+                    false
+                }
```
jjbayer: Not sure if this is better:

```suggestion
                v.as_ref()
                    .and_then(|v| v.value)
                    .map_or(false, f64::is_finite)
```

relay-server/src/services/processor/span/processing.rs
```diff
@@ -220,13 +221,51 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState) {
...
+    received_at: DateTime<Utc>,
+    global_measurements_config: Option<&'a MeasurementsConfig>,
+    project_measurements_config: Option<&'a MeasurementsConfig>,
+) -> NormalizeSpanConfig<'a> {
+    let aggregator_config =
+        AggregatorConfig::from(config.aggregator_config_for(MetricNamespace::Spans));
+
+    NormalizeSpanConfig {
+        received_at,
+        transaction_range: aggregator_config.timestamp_range(),
```
jjbayer: nit: We should rename this field to `timestamp_range`.

jjbayer (approved):
See comments about `RawValue`.

jjbayer (approved):

Issue #2796, opened 2023-11-30
Block future ingestion of a metric


Dav1dde:
For this issue only blocking of the entire metric is required, in the future we want to also block by tag.
We will also have to modify the project config.

technical implementation:
- support globbing
- consider blocking tags in spec
- relay can block all namespaces

Epic: https://github.com/getsentry/sentry/issues/60263

PR #2954, opened 2024-01-17
feat(server): Denylist for metrics

TBS1996:
New addition to projectconfig that allow users to filter out certain metrics based on their name using glob patterns. 

#2796


relay-dynamic-config/src/project.rs
```diff
@@ -88,6 +88,9 @@ pub struct ProjectConfig {
     /// relays that might still need them.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub span_description_rules: Option<Vec<SpanDescriptionRule>>,
+    /// Configuration for filtering metrics.
+    #[serde(default, skip_serializing_if = "skip_deny_metrics")]
+    pub deny_metrics: ErrorBoundary<Metrics>,
```
Dav1dde: `metrics`

relay-dynamic-config/src/project.rs
```diff
@@ -88,6 +88,9 @@ pub struct ProjectConfig {
     /// relays that might still need them.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub span_description_rules: Option<Vec<SpanDescriptionRule>>,
+    /// Configuration for filtering metrics.
+    #[serde(default, skip_serializing_if = "skip_deny_metrics")]
```
Dav1dde: I'd rather make this `Option<ErrorBoundry<...>>`

Dav1dde (commented):
Just a quick check of things, I need in my PR as well

relay-dynamic-config/src/project.rs
TBS1996: well, atm it's just a vector underneath, and we don't really do Option\<Vec\> in the projectconfig if you check out the other values. I'm not sure what the difference between `None` and an empty vec would be, so I think we might as well not encode that.

relay-dynamic-config/src/project.rs
TBS1996: tl;dr

relay-dynamic-config/src/project.rs
Dav1dde: `Metrics` is an Object not a Vector? If you refer to the fields it contains, that does and should not matter.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 - Add automatic PII scrubbing to `logentry.params`. ([#2956](https://github.com/getsentry/relay/pull/2956))
 - Avoid producing `null` values in metric data. These values were the result of Infinity or NaN values extracted from event data. The values are now discarded during extraction. ([#2958](https://github.com/getsentry/relay/pull/2958))
 
+**Features**:
```
jan-auer:
```suggestion
**Internal**:
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -4,11 +4,40 @@ use std::collections::BTreeSet;
...
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct Metrics {
+    /// Patterns of names of metrics that we want to filter.
+    pub denied_names: GlobPatterns,
+}
+
+impl Metrics {
+    /// Filters metrics based on the metric name.
+    pub fn filter_metrics(&self, metrics: &mut Vec<Bucket>) {
```
jan-auer: Let's move this function into a logic module to keep the config types plain. For now, `processor.rs` would be the ideal place for this, until all metrics related processing functionality moves into a dedicated module.

This means you can also remove the dependency on `relay_metrics` from this crate.

relay-dynamic-config/src/metrics.rs
```diff
@@ -479,7 +508,9 @@ where
 
 #[cfg(test)]
 mod tests {
+
```
jan-auer:
```suggestion
```

relay-dynamic-config/src/project.rs
```diff
@@ -13,12 +13,12 @@ use relay_sampling::SamplingConfig;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
-use crate::defaults;
 use crate::error_boundary::ErrorBoundary;
 use crate::feature::FeatureSet;
 use crate::metrics::{
     self, MetricExtractionConfig, SessionMetricsConfig, TaggingRule, TransactionMetricsConfig,
 };
+use crate::{defaults, Metrics};
```
jan-auer: _nit_: There's already an import group for `crate::metrics::`, could you move `Metrics` there?

relay-dynamic-config/src/project.rs
jan-auer: After an offline conversation we decided on keeping this non-optional in favor of using `Metrics::default()`. Note that implementations should also use the default in case of an error in the boundary.

relay-dynamic-config/src/project.rs
jan-auer: This comment is still open.

relay-dynamic-config/src/metrics.rs
```diff
@@ -4,11 +4,40 @@ use std::collections::BTreeSet;
...
+}
+
+impl Metrics {
+    /// Filters metrics based on the metric name.
+    pub fn filter_metrics(&self, metrics: &mut Vec<Bucket>) {
+        metrics.retain(|bucket| !self.denied_names.is_match(&bucket.name));
+    }
+
+    /// Creates a new instance of [`Metrics`].
+    pub fn new(patterns: Vec<String>) -> Self {
```
jan-auer, line 23: _nit_: As this struct will get more fields, we will probably not need this constructor and instead construct it via:

```rust
Metrics {
  denied_names: vec![],
  ..Metrics::default()
}
```

relay-server/src/services/project.rs
```diff
@@ -585,6 +585,19 @@ impl Project {
             return;
         };
 
+        Self::apply_metrics_deny_list(&state.config.deny_metrics, metrics);
+        Self::filter_disabled_namespace(state, metrics);
+    }
+
+    fn apply_metrics_deny_list(deny_list: &ErrorBoundary<Metrics>, buckets: &mut Vec<Bucket>) {
```
jan-auer, line 592: _nit_: Would prefer to inline this function. It's not an associated function as it doesn't rely on the project, but you're passing fields of the project in. Since the function is rather small, it would lend itself for inlining:

```rust
if let ErrorBoundary::Ok(ref metrics) = state.config.deny_metrics.metrics {
  // TODO: Move filter_metrics into this file
  metrics.filter_metrics(buckets); 
}
```

jan-auer (changes requested):

relay-server/src/services/project.rs
TBS1996: rather than inling, I moved more responsibility to it, since I removed hte associated method on `Metrics` to avoid the `Bucket` dependency in the config crate. 
 
I assume thats's what you meant earlier, unless I should have multiple impl blocks to get around the dependency problem?

relay-dynamic-config/src/project.rs
```diff
@@ -139,6 +144,13 @@ fn skip_metrics_extraction(boundary: &ErrorBoundary<MetricExtractionConfig>) ->
     }
 }
 
+fn skip_deny_metrics(boundary: &ErrorBoundary<Metrics>) -> bool {
```
Dav1dde:
```suggestion
fn skip_metrics(boundary: &ErrorBoundary<Metrics>) -> bool {
```

relay-dynamic-config/src/metrics.rs
TBS1996: I just moved this logic into the apply_filter_metrics function in project.rs, how do you think about that?

relay-dynamic-config/src/metrics.rs
```diff
@@ -4,11 +4,27 @@ use std::collections::BTreeSet;
...
 use relay_protocol::RuleCondition;
 use serde::{Deserialize, Serialize};
 
 use crate::project::ProjectConfig;
 
+/// Configuration for metrics filtering.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct Metrics {
+    /// Patterns of names of metrics that we want to filter.
```
olksdr:
```suggestion
    /// The list of patterns which used to filter names of the received metrics.
```
Maybe something like this will sound better?

TBS1996: hmm, isnt it a bit misleading, like we want to say we filter metrics based on their name, not that we filter their names?

how about `/// List of patterns for blocking metrics based on their name.` ?

relay-dynamic-config/src/metrics.rs
```diff
@@ -4,11 +4,27 @@ use std::collections::BTreeSet;
...
 use relay_common::glob2::LazyGlob;
+use relay_common::glob3::GlobPatterns;
 use relay_protocol::RuleCondition;
 use serde::{Deserialize, Serialize};
 
 use crate::project::ProjectConfig;
 
+/// Configuration for metrics filtering.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(transparent)]
```
jan-auer:
```suggestion
```

jan-auer (approved):
Approved once `#[serde(transparent)]` is removed

relay-dynamic-config/src/metrics.rs
Dav1dde: This should be `#[serde(default, rename_all = "camelCase")]`

relay-dynamic-config/src/metrics.rs
```diff
@@ -4,11 +4,27 @@ use std::collections::BTreeSet;
...
+pub struct Metrics {
+    /// List of patterns for blocking metrics based on their name.
+    pub denied_names: GlobPatterns,
+}
+
+impl Metrics {
+    /// Returns `true` if it contains any names patterns to filter metric names.
```
Dav1dde, line 21: This comment is outdated

relay-dynamic-config/src/metrics.rs
```diff
@@ -4,11 +4,26 @@ use std::collections::BTreeSet;
...
 
+/// Configuration for metrics filtering.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct Metrics {
+    /// List of patterns for blocking metrics based on their name.
+    pub denied_names: GlobPatterns,
+}
+
+impl Metrics {
+    /// Returns `true` if it contains any names patterns to filter metric names.
```
Dav1dde: This comment is outdated, is_empty does more now

relay-dynamic-config/src/metrics.rs
```diff
@@ -4,11 +4,26 @@ use std::collections::BTreeSet;
...
 use relay_common::glob2::LazyGlob;
+use relay_common::glob3::GlobPatterns;
 use relay_protocol::RuleCondition;
 use serde::{Deserialize, Serialize};
 
 use crate::project::ProjectConfig;
 
+/// Configuration for metrics filtering.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct Metrics {
```
Dav1dde, line 16: This should be `#[serde(default, rename_all = "camelCase")]`

Dav1dde (changes requested):

relay-dynamic-config/src/project.rs
```diff
@@ -88,6 +89,9 @@ pub struct ProjectConfig {
     /// relays that might still need them.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub span_description_rules: Option<Vec<SpanDescriptionRule>>,
+    /// Configuration for filtering metrics.
```
Dav1dde:
```suggestion
    /// Configuration for metrics.
```

Dav1dde (commented):
This seems like a footgun that the there is no explicit filter for namespaces at least, but this was already discussed and marked as a future improvement.

Dav1dde (approved):
Does this get rid of `Request Changes` now?

PR #2955, opened 2024-01-17
ref(server): Remove redundant indirection.

TBS1996:
`fn rate_limit_metrics` doesn't do anything if the projectstate is not present, and the only time we used it was right before we check if we have the projectstate. Simply waiting to call the function until after we checked the projectstate removes some useless back-and-forth, which should make the code easier to reason about.  
 
With that it also made sense to make filter_metrics take in projectstate as an argument, maybe we should move this function now then.. 

#skip-changelog


relay-server/src/actors/project.rs
```diff
@@ -682,14 +680,16 @@ impl Project {
         buckets: Vec<Bucket>,
     ) {
         if self.metrics_allowed() {
-            let buckets = self.rate_limit_metrics(buckets, outcome_aggregator.clone());
+            match &mut self.state {
```
TBS1996: another bonus: less identation

jjbayer (approved):

PR #2957, opened 2024-01-17
ref: Serialize infinite as MAX/MIN

jan-auer:
Version that serializes infinite floats as MAX and MIN, ignoring NaN. This will create large floats, which can bee considered data corruption.

#skip-changelog


jan-auer:
Closing in favor of https://github.com/getsentry/relay/pull/2958

PR #2958, opened 2024-01-17
fix(metrics): Use a FiniteF64 type in metrics

jan-auer:
Introduces a strictly typed `FiniteF64` type to prevent storing non-finite floats in metrics. This type requires a runtime check to be constructed. All existing usage of this type has been updated:

- Metric extraction skips values that cannot be converted
- Counters and gauges use saturating addition (clamped at MAX / MIN)
- Tests use integers as mocks, which support infallible conversion
- The `relay_metrics::dist!` macro now takes values that implement `Into<FiniteF64>`

This PR does not yet log errors in metric extraction, this will be added in a follow-up.


olksdr (approved):
:shipit:

Dav1dde (approved):
:shipit:

PR #2959, opened 2024-01-17
ref(deploy): Add new PoP regional clusters to the deployment pipeline

beezz:
Adding new regional PoP clusters to the deployment pipeline. These clusters will eventually replace the current single-zone clusters.

Depends on https://github.com/getsentry/devinfra-deployment-service/pull/513

#skip-changelog

### DIFF
--- a/gocd/templates/libs/relay-pops.libsonnet
+++ b/gocd/templates/libs/relay-pops.libsonnet
@@ -11,6 +11,10 @@ local region_pops = {
     'us-pop-2',
     'us-pop-3',
     'us-pop-4',
+    'us-pop-regional-1',
+    'us-pop-regional-2',
+    'us-pop-regional-3',
+    'us-pop-regional-4',
   ],
 };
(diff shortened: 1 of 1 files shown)


Dav1dde (approved):

olksdr (approved):
🚢  it!

gocd/templates/libs/relay-pops.libsonnet
```diff
@@ -11,6 +11,10 @@ local region_pops = {
     'us-pop-2',
     'us-pop-3',
     'us-pop-4',
+    'us-pop-regional-1',
```
dmajere, line 14: why are they all us?
should it be asia, australia, europe and then us here?

beezz: that's how the files are named ~ https://github.com/getsentry/devinfra-deployment-service/pull/513/files

PR #2961, opened 2024-01-18
fix(metrics-extraction): Add defaulting of required attributes in light normalization

iambriccardo:
This PR adds a new normalization step in the light normalization pipeline which normalizes required attributes that have no value. The need for this change arises from metrics extraction, which is run after light normalization and requires the event payload to be normalized almost entirely to perform the correct conditional extraction.

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -19,8 +19,8 @@ use relay_event_schema::processor::{
 };
 use relay_event_schema::protocol::{
     AsPair, Context, ContextInner, Contexts, DeviceClass, Event, EventType, Exception, Headers,
-    IpAddr, LogEntry, Measurement, Measurements, NelContext, Request, SpanAttribute, SpanStatus,
-    Tags, User,
+    IpAddr, Level, LogEntry, Measurement, Measurements, NelContext, Request, SpanAttribute,
+    SpanStatus, Tags, User,
 };
 use relay_protocol::{Annotated, Empty, Error, ErrorKind, Meta, Object, Value};
 use smallvec::SmallVec;
@@ -201,6 +201,9 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) -
 
     // Check for required and non-empty values
     let _ = schema::SchemaProcessor.process_event(event, meta, ProcessingState::root());
+
+    // Set required attributes to default values if not present.
+    normalize_default_required_attrs(event);
 
     normalize_timestamps(
         event,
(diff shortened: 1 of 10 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -455,6 +458,18 @@ fn normalize_dist(distribution: &mut Annotated<String>) {
     });
 }
 
+/// Defaults required attributes to their identity value.
+fn normalize_default_required_attrs(event: &mut Event) {
+    // The defaulting behavior, was inherited from `StoreNormalizeProcessor` and it's put here since only light
+    // normalization happens before metrics extraction and we want the metrics extraction pipeline to already work
+    // on some normalized data.
+    event.platform.get_or_insert_with(|| "other".to_string());
+    event.level.get_or_insert_with(|| match event.ty.value() {
```
iambriccardo, line 465: I have moved both `platform` and `level` normalization, since both defaulted to values which might be useful for metrics extraction. I could technically move all the default normalizations, @iker-barriocanal what do you think?

relay-event-schema/src/protocol/event.rs
```diff
@@ -1146,6 +1147,8 @@ mod tests {
             ..Default::default()
         };
 
+        assert_eq!(Some(Val::String("info")), event.get_value("event.level"));
```
iambriccardo, line 1150: Added this test just to make sure the getter works as expected.

relay-event-normalization/src/event.rs
```diff
@@ -202,6 +202,9 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) -
     // Check for required and non-empty values
     let _ = schema::SchemaProcessor.process_event(event, meta, ProcessingState::root());
 
+    // Set required attributes to default values if not present.
+    normalize_default_required_attrs(event);
```
iker-barriocanal: Let's move this lower, as the event may still be rejected after this point. There are some required attribute normalization on line 255, that could be a good place.

relay-event-normalization/src/event.rs
iker-barriocanal: This is good enough for now, thanks!

iker-barriocanal (changes requested):
Thanks for working on this! Just one minor comment.

relay-event-normalization/src/event.rs
iambriccardo: Thanks for the tip! Will do

iker-barriocanal (approved):

PR #2962, opened 2024-01-18
fix(spooler): Proactively populate the project cache for envelopes spooled to disk

olksdr:
When the Relay restarts the index of the spooled to disk envelopes must be restored:
* read all the project keys we spooled to disk
* generate unique index
* make sure the project cache fetches them

Once the project state received Relay will start to unspool saved to disk envelopes. 

This should allow Relay to unspool faster, even for the projects which do not have the high volume traffic. 

related to https://github.com/getsentry/team-ingest/issues/257

#skip-changelog


relay-server/src/actors/spooler/mod.rs
```diff
@@ -684,6 +697,47 @@ impl OnDisk {
...
+
+        // Fold the list into the index format.
+        let index = extracted_keys?.into_iter().fold(
+            BTreeMap::new(),
+            |mut acc: BTreeMap<ProjectKey, BTreeSet<QueueKey>>, key| {
+                acc.entry(key.own_key).or_default().insert(key);
+                acc.entry(key.sampling_key).or_default().insert(key);
+                acc
+            },
+        );
```
olksdr: I could do this in one go, without allocating a vec, but that means I will have to also ignore the errors here, and filter them out, and report them in `extract_key` instead.

relay-server/src/actors/spooler/mod.rs
```diff
@@ -1028,13 +1091,34 @@ impl BufferService {
         Ok(())
     }
 
+    /// Handles the retrieving the index from the underlying spool.
+    async fn handle_get_index(&mut self, _: GetIndex) -> Result<(), BufferError> {
```
iker-barriocanal: Do you think what the "handling" is, is relevant enough to document it?

relay-server/src/actors/spooler/mod.rs
```diff
@@ -684,6 +697,55 @@ impl OnDisk {
...
+                None
+            }
+        }
+    }
+
+    /// Returns the index from the on-disk spool.
+    async fn get_spooled_index(
+        &self,
+    ) -> Result<BTreeMap<ProjectKey, BTreeSet<QueueKey>>, BufferError> {
+        let keys = sql::get_keys()
```
iker-barriocanal: Some time ago we discussed how querying the whole database during startup could be problematic for Relay. Is this no longer the case? Are there new considerations we've made?

iker-barriocanal (commented):
The code looks good to me. Leaving one question about the performance I think we should address before moving forward with this.

relay-server/src/actors/spooler/mod.rs
olksdr: Added some more docs. Let me know what exactly you would like to see here.

relay-server/src/actors/spooler/mod.rs
olksdr: I've changed this a little bit. We do not block the start up , and execute the query in the background, meaning we allow ourself to run it a bit longer and process the result async.

iker-barriocanal (approved):
Awesome! Thank you for the nice PR 😄

Dav1dde (approved):

relay-server/src/services/project_cache.rs
```diff
@@ -996,6 +1021,9 @@ impl Service for ProjectCacheService {
                 }
             };
 
+            // Reqest the existing index from the spooler.
```
jjbayer:
```suggestion
            // Request the existing index from the spooler.
```

relay-server/src/services/spooler/mod.rs
```diff
@@ -1029,13 +1092,49 @@ impl BufferService {
...
+                    match OnDisk::get_spooled_index(&db).await {
+                        Ok(index) => {
+                            relay_log::trace!(
+                                "recover index from disk with {} unique project keys",
+                                index.len()
+                            );
+                            project_cache.send(SpoolIndex(index))
+                        }
+                        Err(err) => {
+                            relay_log::error!("failed to retrive the index from the disk: {err}")
```
jjbayer:
```suggestion
                            relay_log::error!("failed to retrieve the index from the disk: {err}")
```

relay-server/src/services/spooler/mod.rs
```diff
@@ -183,6 +183,10 @@ impl RemoveMany {
 #[derive(Debug)]
 pub struct Health(pub Sender<bool>);
 
+/// Requests the index [`ProjectKey`] -> [`QueueKey`] of the data currently residing in the spool.
+#[derive(Debug)]
+pub struct GetIndex;
```
jjbayer: nit: Is there a more descriptive name we can use for this message. Something like `TriggerProjectStateFetchForSpooledEnvelopes`, but simpler?

jjbayer (commented):
Is there a way we can test this behavior?

relay-server/src/services/spooler/mod.rs
Dav1dde: ~~I like `GetIndex`, it just gives you the entire spool index and describes that operation, the project cache then just uses the index.~~

Oh actually I thought you would pass a return channel with the `GetIndex` but it actually just calls into the project cache directly. Agree then a more descriptive name wouldn't hurt.

olksdr: how about now, see changes in commit [`710e80f`](https://github.com/getsentry/relay/pull/2962/commits/710e80ffcb1b95e13b7b6ef1f8a9f6a9ebb0364a)

Added more docs and also rename the messages to hopefully give more intuition about what's going on

relay-server/src/services/spooler/mod.rs
```diff
@@ -1029,13 +1098,49 @@ impl BufferService {
...
+    /// Handles retrieving of the index from the underlying spool.
+    ///
+    /// To process this request we spawn the tokio task and once it's done the result will be sent
+    /// to [`ProjectCache`].
+    ///
+    /// If the spool is memory based, we ignore this request - the index in the
+    /// [`ProjectCache`] should be full and correct.
+    /// If the spool is located on the disk, we read up the keys and compile the index of the
+    /// spooled envelopes for [`ProjectCache`].
+    async fn handle_get_index(&mut self, _: RestoreIndex) -> Result<(), BufferError> {
```
jjbayer: Should we rename this to `handle_restore_index` as well?

jjbayer (approved):
> how about now, see changes in commit [710e80f](https://github.com/getsentry/relay/pull/2962/commits/710e80ffcb1b95e13b7b6ef1f8a9f6a9ebb0364a)

"Restore" sounds good to me!

relay-server/src/services/spooler/mod.rs
olksdr: yep, will do it now

PR #2964, opened 2024-01-18
ref(server): actors -> services

jjbayer:
Rename the `actors` module to `services`.

We migrated away from the actix actors framework over a year ago (see https://github.com/getsentry/relay/pull/1776). The replacement for actors are message handlers that implement the `Service` trait.

#skip-changelog


relay-server/src/lib.rs
```diff
@@ -262,6 +261,7 @@ mod http;
 mod metrics_extraction;
 mod middlewares;
 mod service;
+mod services;
```
jjbayer, line 264: To make naming less awkward here, we might want to move `service` to `services::utils` or something similar.

olksdr (approved):
:shipit:

iker-barriocanal (approved):

PR #2966, opened 2024-01-18
fix(spans): Deduplicate more repeated conditions

jjbayer:
https://github.com/getsentry/relay/issues/2929 partially scrubbed repeated conditions in SQL queries, but it did not consider the case where the repeated operands of an `OR` or `AND` operation are not siblings (see tree in code comments).

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/sql/parser.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/parser.rs
@@ -274,12 +274,32 @@ impl VisitorMut for NormalizeVisitor {
             }
             Expr::BinaryOp {
                 ref mut left,
-                op: BinaryOperator::Or | BinaryOperator::And,
+                op: op @ (BinaryOperator::Or | BinaryOperator::And),
                 ref right,
             } => {
-                let is_equal = left == right;
-                if is_equal {
+                if left == right {
+                    //     /\
+                    //    /  \
+                    //   /\   B
+                    //  /  \
+                    // A    A
                     *expr = take_expr(left);
+                } else {
+                    //     /\
+                    //    /  \
+                    //   /\   B
+                    //  /  \
+                    // A    B
+                    if let Expr::BinaryOp {
+                        left: left_left,
+                        op: left_op,
+                        right: left_right,
+                    } = left.as_mut()
+                    {
+                        if left_op == op && left_right == right {
+                            *left = Box::new(take_expr(left_left));
+                        }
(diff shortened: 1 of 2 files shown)


olksdr (approved):

Dav1dde (approved):

gggritso (approved):
👍🏻

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -274,12 +274,32 @@ impl VisitorMut for NormalizeVisitor {
...
+                    //   /\   B
+                    //  /  \
+                    // A    A
                     *expr = take_expr(left);
+                } else {
+                    //     /\
+                    //    /  \
+                    //   /\   B
+                    //  /  \
+                    // A    B
```
Zylphrex, line 292: A few questions, is the shape of the tree guaranteed to look like this? Are the following possible?

where the right side is a binary op
```
  /\
 /  \
A   /\
   /  \
  A    B
```

or both sides are a binary op

```
    / \
   /   \
  /\   /\
 /  \ /  \
A   B B   A
```

jjbayer: @Zylphrex Excellent question! The parser deterministically parses multiple ORs / ANDs into a left-leaning binary tree, that is `a OR b OR c` is parsed into

```rust
BinaryOp {
    left: BinaryOp {
        left: Identifier(
            Ident {
                value: "a",
            },
        ),
        op: Or,
        right: Identifier(
            Ident {
                value: "b",
            },
        ),
    },
    op: Or,
    right: Identifier(
        Ident {
            value: "c",
        },
    ),
},
```

Your first example is only possible with parentheses. The second example is possible when the inner and outer operand are different.

```
    / \
   /   \
  /\   /\
 /  \ /  \
A   B B   A
```

could theoretically be normalized into

```
    / \
   /   \
  /\   /\
 /  \ /  \
A   B A   B
```

and then further into

```
   /\
  /  \
 A    B
```


but that goes beyond finding simple forms of repetition and into full algebraic normalization, so I would not touch it unless you think it's worth it.

Zylphrex: Understood! I don't think it's necessary as these are edge cases that I don't expect to happen frequently but it's good to know that it's something we can solve if it does occur

PR #2968, opened 2024-01-18
fix(webvitals): Adds lower bound to component values in normalize_performance_score

edwardgou-sentry:
Adds a lower bound of 0 to component values in `normalize_performance_score`. `calculate_cdf_score` does not support negative values and web vitals are not expected to be negative.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -767,7 +767,11 @@ fn normalize_performance_score(
                     let mut normalized_component_weight = 0.0;
                     if let Some(value) = measurements.get_value(component.measurement.as_str()) {
                         normalized_component_weight = component.weight / weight_total;
-                        let cdf = utils::calculate_cdf_score(value, component.p10, component.p50);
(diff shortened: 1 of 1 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -767,7 +767,11 @@ fn normalize_performance_score(
                     let mut normalized_component_weight = 0.0;
                     if let Some(value) = measurements.get_value(component.measurement.as_str()) {
                         normalized_component_weight = component.weight / weight_total;
-                        let cdf = utils::calculate_cdf_score(value, component.p10, component.p50);
+                        let cdf = utils::calculate_cdf_score(
+                            value.max(0.0), // Webvitals can't be negative, but we need to clamp in case of bad data.
```
Dav1dde, line 771: I don't think this is enough, `utils::calculate_cdf_score` calculates `f64::ln(value)` which produces infinity for `0`.

Dav1dde (approved):

PR #2969, opened 2024-01-18
feat(spans): Extract INP metrics from spans

### LINKED ISSUE #2873: Extract INP and INP score from spans and into metrics
### Problem Statement
INP is a new core webvital that is due to replace FID in March 2024: https://web.dev/articles/inp

The Sentry **Web Vitals** module uses a concept of **Performance Scores** which uses FID as a component in it's calculation. We should replace FID with INP within the Performance Score concept. To do this, we need updates to Relay to calculate and store INP scores in indexed spans and metrics. More details below:

- INP is a span measurement that will come from the JS sdk. Other webvital measurements such as LCP, FCP, etc come from the JS sdk too, but they're on transactions instead.

- Each webvital on an transaction has a score measurement. For example measurement.score.lcp. These scores are calculated in relay in the event normalizer here: https://github.com/getsentry/relay/blob/3662370e488350987759bfa53bec5d362674bc62/relay-event-normalization/src/event.rs#L711

- The total score (ie measurements.score.total) is the sum of each Webvital score on an transaction, with weights applied. This also gets calculated in relay and is stored on the event and as a metric

- INP is on interaction spans, which do not have any other webvital attached except INP. So basically, the measurements.score.total of an INP span is just the same value as measurements.score.inp

-So the work remaining here is to bring over some of the score calculation logic from the event normalizer into spans so that we can get measurements.score.total and measurements.score.inp as metrics

### Additional Thoughts
- Since we are moving away from transactions and to exclusively spans, we may want to recreate the existing score metrics to be under the spans namespace instead. ie from `d:transactions/measurements.score.total@ratio` to something like  `d:spans/measurements.webvital.score.total@ratio` and `d:spans/measurements.webvital.score.inp@ratio` for INP . Also might want to prefix with `webvital` so theres more clarity around the score source.

- We only need a few tags. See here for what tags we used in transactions:
https://github.com/getsentry/relay/pull/2827/files

- For the **score** metrics, we also don't need all the distributions. `sum` and `count` _should_ be enough for the Web Vitals product purposes. For the raw INP measurement value, maybe we keep the standard distributions so that they can be queried in Dashboards? (this would be the existing behaviour of other webvital measurements)

phacops:
Fixes https://github.com/getsentry/relay/issues/2873.


relay-dynamic-config/src/defaults.rs
```diff
@@ -395,5 +395,57 @@ fn span_metrics(is_extract_all: bool) -> impl IntoIterator<Item = MetricSpec> {
...
+                Tag::with_key("environment")
+                    .from_field("span.sentry_tags.environment")
+                    .always(),
+                Tag::with_key("release")
+                    .from_field("span.sentry_tags.release")
+                    .always(),
+                Tag::with_key("browser.name")
+                    .from_field("span.browser.name")
+                    .always(),
+                Tag::with_key("geo.country_code")
```
Dav1dde: Browser name and geo code combined can be a really high cardinality (very easy to hit >2k per transaction on a global application), are they needed?

We're trying to get rid of them for transactions: https://github.com/getsentry/team-ingest/issues/234#issuecomment-1885211797

relay-event-schema/src/protocol/span.rs
```diff
@@ -159,6 +159,14 @@ impl Getter for Span {
...
                 } else if let Some(key) = path.strip_prefix("sentry_tags.") {
                     self.sentry_tags.value()?.get(key)?.as_str()?.into()
+                } else if let Some(key) = path.strip_prefix("measurements.") {
+                    self.measurements
+                        .value()?
+                        .get(key)?
+                        .value()?
+                        .value
+                        .value()?
+                        .into()
```
jjbayer: Let's distinguish between `measurement.value` and `measurement.unit` here, as we do for events: https://github.com/getsentry/relay/blob/190ac90b72f9f938e3e3f7a5c94acbf659f7adfa/relay-event-schema/src/protocol/event.rs#L771-L773

https://github.com/getsentry/relay/pull/2938/files#r1450208498

relay-dynamic-config/src/defaults.rs
jjbayer: +1, let's figure out whether we need all of these tags, especially whether we need them all in the same query. E.g. if `geo.country_code` is never queried together with `transaction`, we can split the metric into two.

phacops: That's a good question, I was merely replicating what's in transactions for this (as the task mentioned).

phacops: I removed them and limited cardinality for browser name.

relay-dynamic-config/src/defaults.rs
```diff
@@ -395,5 +400,51 @@ fn span_metrics(is_extract_all: bool) -> impl IntoIterator<Item = MetricSpec> {
                     .always(), // already guarded by condition on metric
             ],
         },
+        MetricSpec {
```
edwardgou-sentry, line 408: Can we also add `webvital.score.weight.inp`? 😄

relay-dynamic-config/src/defaults.rs
```diff
@@ -395,5 +400,51 @@ fn span_metrics(is_extract_all: bool) -> impl IntoIterator<Item = MetricSpec> {
...
+            category: DataCategory::Span,
+            mri: "d:spans/webvital.score.total@ratio".into(),
```
jjbayer, line 410: Please add the new MRIs to https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py before merging this.

relay-event-schema/src/protocol/span.rs
```diff
@@ -159,6 +159,15 @@ impl Getter for Span {
...
                     self.sentry_tags.value()?.get(key)?.as_str()?.into()
+                } else if let Some(rest) = path.strip_prefix("measurements.") {
+                    let name = rest.strip_suffix(".value")?;
+                    self.measurements
+                        .value()?
+                        .get(name)?
+                        .value()?
+                        .value
+                        .value()?
+                        .into()
```
jjbayer, line 171: Wanna add a unit test to this file?

relay-dynamic-config/src/defaults.rs
```diff
@@ -395,5 +400,51 @@ fn span_metrics(is_extract_all: bool) -> impl IntoIterator<Item = MetricSpec> {
...
+                Tag::with_key("browser.name")
+                    .from_field("span.browser.name")
+                    .always(), // already guarded by condition on metric
+            ],
+        },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "d:spans/webvital.score.inp@ratio".into(),
+            field: Some("span.measurements.score.inp.value".into()),
+            condition: Some(is_allowed_browser),
```
jjbayer, line 481: I would also add a test input to https://github.com/getsentry/relay/blob/f3ff94a1850be0b158250e454c6f650891c5931c/relay-server/src/metrics_extraction/event.rs (e.g. `test_extract_span_metrics`) to check if the condition actually works.

jjbayer (approved):
Do we know what volume we're expecting for this change initially? Do SDKs already send this measurement in practice?

See comment about `strings.py`.

relay-dynamic-config/src/defaults.rs
phacops: https://github.com/getsentry/sentry/pull/63699

relay-event-normalization/src/event.rs
```diff
@@ -753,80 +754,117 @@ fn normalize_performance_score(
...
-                let mut score_total = 0.0;
-                let mut weight_total = 0.0;
-                for component in &profile.score_components {
-                    // Skip optional components if they are not present on the event.
-                    if component.optional
-                        && !measurements.contains_key(component.measurement.as_str())
-                    {
+            let Some(measurements) = event.measurements.value_mut() else {
+                return;
+            };
```
jjbayer: We can move this out of the loop.

relay-event-normalization/src/event.rs
```diff
@@ -753,80 +754,117 @@ fn normalize_performance_score(
...
+            normalize_performance_score_inner(measurements, profile);
```
jjbayer: Once we found a matching profile, we can stop.

```suggestion
            normalize_performance_score_inner(measurements, profile);
            return;
```

relay-event-normalization/src/event.rs
```diff
@@ -753,80 +754,117 @@ fn normalize_performance_score(
...
                     }
-                    measurements.insert(
-                        format!("score.weight.{}", component.measurement),
-                        Measurement {
-                            value: normalized_component_weight.into(),
-                            unit: (MetricUnit::Fraction(FractionUnit::Ratio)).into(),
-                        }
-                        .into(),
-                    );
+                    normalize_performance_score_inner(measurements, profile);
```
```suggestion
                    normalize_performance_score_inner(measurements, profile);
                    return;
```

relay-event-normalization/src/event.rs
```diff
@@ -753,80 +754,117 @@ fn normalize_performance_score(
...
-                        measurements.insert(
-                            format!("score.{}", component.measurement),
-                            Measurement {
-                                value: component_score.into(),
-                                unit: (MetricUnit::Fraction(FractionUnit::Ratio)).into(),
-                            }
-                            .into(),
-                        );
+                    if condition.value != browser_name {
+                        continue;
```
jjbayer: If we're gonna hack anyway, I would wrap the span in a pseudo event that has `event.contexts.browser.name` and `measurements` set. That should make the rest of the refactor unnecessary.

Long-term, I would add a new key to the performance score config (e.g. define a field `span_conditions` on `PerformanceScoreProfile` in addition to `conditions` that formulates the condition on a span properly.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -332,6 +340,11 @@ fn normalize(
             .collect(),
     );
 
+    if let Some(measurements) = span.measurements.value_mut() {
+        // TODO: find the browser name by normalizing user-agent
+        normalize_span_performance_score("".to_string(), measurements, performance_score);
```
jjbayer: Here's some pointers:

https://github.com/getsentry/relay/blob/e27a4ea543931a0774961a1d57f63d9efe98fcd9/relay-event-normalization/src/normalize/user_agent.rs#L129-L155

https://github.com/getsentry/relay/blob/e27a4ea543931a0774961a1d57f63d9efe98fcd9/relay-event-normalization/src/normalize/user_agent.rs#L24-L39

phacops:
SDKs are not sending this measurement yet, volume will be 0 until the next release.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -342,10 +358,14 @@ fn normalize(
...
-    if let Some(measurements) = span.measurements.value_mut() {
-        // TODO: find the browser name by normalizing user-agent
-        normalize_span_performance_score("".to_string(), measurements, performance_score);
-    }
+    let mut event = Event {
+        contexts,
+        measurements: span.measurements.clone(),
+        spans: Annotated::from(vec![Annotated::new(span.clone())]),
+        ..Default::default()
+    };
```
jjbayer, line 366: This turned out hackier than I thought, with all the cloning. We could instead create the pseudo-event before the `retain_items` loop, pass that into `normalize()` instead, and `.take()` measurements & spans in and out of that to prevent cloning. But I would not invest time into that, we should instead have a generic `normalize_performance_score` function that works on both spans and events.

PR #2971, opened 2024-01-19
ref(spans): Remove redundant feature check

jjbayer:
This feature is now rolled out to 100% of projects on SaaS, so we can remove related check.

Using one feature flag for multiple features brings the danger of accidentally enabling a feature, so we should deprecate this one and introduce new feature flags for specific features instead.

#skip-changelog


olksdr (approved):

Dav1dde (approved):

phacops:
Let's wait this Monday to check if we can fully enable that and then good to go.

phacops (approved):

PR #2972, opened 2024-01-19
feat(cardinality): Implement quota/limit system for cardinality limiter

Dav1dde:
Implements a similar system we already have for rate limits for the cardinality limiter. Instead of a single global limit now quotas can be defined matching on different scopes (currently only org) and namespaces.

This allows us to support multiple scopes in the future and right now to support different limits per namespace.

In a follow up:
- I will add the cardinality limits to the project config and read them from there instead of the relay config.
- Update the module structure so we can get rid of a few `pub(crate)`'s, specifically the cache module should be a submodule from the Redis module. Keeping it as is for now to keep the diff smaller.

Metrics are now tagged with the limit id instead of the namespace of the metric being checked. Not sure if we want to change this in the future.

#skip-changelog


relay-server/src/services/processor.rs
```diff
@@ -835,6 +827,34 @@ impl EnvelopeProcessorService {
...
+            MetricNamespace::Transactions,
+            MetricNamespace::Custom,
+            MetricNamespace::Unsupported,
+        ]
+        .map(|ns| CardinalityLimit {
+            id: ns.as_str().to_owned(),
+            window,
+            limit,
+            scope,
+            namespace: Some(MetricNamespace::Sessions),
```
```suggestion
            namespace: Some(ns),
```
?

iker-barriocanal (approved):

relay-cardinality/benches/redis_impl.rs
jjbayer, line 1: I noticed the `#skip-changelog` on the PR description. I feel like this feature warrants a changelog entry, even if just under "internal".

relay-server/src/services/processor.rs
```diff
@@ -1812,14 +1832,20 @@ impl EnvelopeProcessorService {
...
+        limits: &[CardinalityLimit],
         buckets: Vec<Bucket>,
         mode: ExtractionMode,
     ) -> Vec<Bucket> {
         let Some(ref limiter) = self.inner.cardinality_limiter else {
             return buckets;
         };
 
-        let limits = match limiter.check_cardinality_limits(scoping.organization_id, buckets) {
+        let cs = relay_cardinality::Scoping {
```
```suggestion
        let cardinality_scoping = relay_cardinality::Scoping {
```

relay-cardinality/src/limiter.rs
```diff
@@ -1,55 +1,59 @@
...
 /// A single entry to check cardinality for.
 #[derive(Clone, Copy, Debug)]
-pub struct Entry<T: CardinalityScope> {
+pub struct Entry {
     /// Opaque entry Id, used to keep track of indices and buckets.
     pub id: EntryId,
 
     /// Cardinality scope.
-    pub scope: T,
+    pub namespace: MetricNamespace,
```
jjbayer, line 56: This simplification makes sense for now. At some point we will probably have cardinality limits for non-metrics though (e.g. span descriptions), and then we will have to generalize the scope again.

relay-cardinality/src/limiter.rs
```diff
@@ -1,55 +1,59 @@
...
 }
 
 /// A single entry to check cardinality for.
 #[derive(Clone, Copy, Debug)]
-pub struct Entry<T: CardinalityScope> {
+pub struct Entry {
     /// Opaque entry Id, used to keep track of indices and buckets.
     pub id: EntryId,
 
     /// Cardinality scope.
```
```suggestion
    /// Metric namespace to which the cardinality limit is scoped.
```

relay-cardinality/src/redis.rs
```diff
@@ -50,171 +49,250 @@ impl RedisSetLimiter {
...
+/// Internal state combining relevant entries for the respective quota and its limit.
+struct LimitState<'a> {
+    /// Id of the original limit.
+    pub id: &'a str,
+    /// Entries which are relevant for the quota.
+    pub entries: Vec<RedisEntry>,
+    /// The quota.
+    pub quota: RedisQuota,
+    /// The limit of the quota.
+    pub limit: u64,
```
jjbayer: Should the `limit` just be part of the quota? See https://github.com/getsentry/relay/blob/cac6d622f6e3f7bfdb89454ca8ae405dd9df5afc/relay-quotas/src/quota.rs#L233

relay-cardinality/src/redis.rs
```diff
@@ -50,171 +49,250 @@ impl RedisSetLimiter {
...
+
+    /// Amount of cache hits `(hits, misses)`.
+    cache_hits: (i64, i64),
+    /// Amount of acceptions and rejetions `(acceptions, reiections)`.
```
jjbayer:
```suggestion
    /// Amount of accepts and rejections `(accepts, rejections)`.
```

relay-cardinality/src/redis.rs
jjbayer: Alternative: rename `RedisQuota` to `QuotaScoping` or similar.

jjbayer (commented):
Leaving a random bunch of comments for now, will continue reviewing later but don't let that block you.

relay-cardinality/src/limiter.rs
Dav1dde: My idea for this was to at some point replace a `MetricNamespace` with a `CardinalityNamespace` and we could use a separate namespace for this usecase.

relay-cardinality/src/redis.rs
Dav1dde: Renamed to `QuotaScoping`

relay-cardinality/src/redis.rs
```diff
@@ -50,171 +49,256 @@ impl RedisSetLimiter {
...
     }
 }
 
+/// A quota scoping extracted from a [`CardinalityLimit`] and a [`Scoping`].
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub(crate) struct QuotaScoping {
+    pub window: SlidingWindow,
+    pub namespace: Option<MetricNamespace>,
+    pub organization_id: Option<OrganizationId>,
+    pub project_id: Option<ProjectId>,
```
jjbayer, line 184: nit: Are all combinations valid here? I.e. `(None, None, None)`, `(None, None, Some)`, etc.

relay-cardinality/src/redis.rs
```diff
@@ -50,171 +49,256 @@ impl RedisSetLimiter {
...
+        Some(Self {
+            window: limit.window,
+            namespace: limit.namespace,
+            organization_id,
+            project_id,
+        })
+    }
+
+    /// Wether the scoping applies to the passed entry.
+    pub fn applies_to(&self, entry: &Entry) -> bool {
```
jjbayer: nit: Similar functions are called `matches` elsewhere. Up to you whether you want to align it. Example: https://github.com/getsentry/relay/blob/2d7aa88d4c2251c428e7cf891ff8c9e7ebf131f1/relay-quotas/src/rate_limit.rs#L170-L171

relay-cardinality/src/redis.rs
```diff
@@ -50,171 +49,256 @@ impl RedisSetLimiter {
...
+        self.namespace.is_none() || self.namespace == Some(entry.namespace)
+    }
+
+    /// Returns all slots of the sliding window for a specific timestamp.
+    pub fn slots(&self, timestamp: u64) -> impl Iterator<Item = u64> {
+        self.window.iter(timestamp)
+    }
+
+    /// Turns the scoping into a Redis key for the passed slot.
+    fn to_redis_key(self, slot: u64) -> String {
```
jjbayer: nit

```suggestion
    fn into_redis_key(self, slot: u64) -> String {
```

relay-cardinality/src/redis.rs
```diff
@@ -50,171 +49,256 @@ impl RedisSetLimiter {
...
+    pub entries: Vec<RedisEntry>,
+    /// Scoping of the quota.
+    pub scope: QuotaScoping,
+    /// The limit of the quota.
+    pub limit: u64,
+
+    /// Amount of cache hits `(hits, misses)`.
+    cache_hits: (i64, i64),
+    /// Amount of accepts and rejections `(acceptions, rejections)`.
+    acceptions_rejections: (i64, i64),
```
jjbayer: Still skeptical about "acception", but apparently it is a word: https://en.wiktionary.org/wiki/acception.

relay-cardinality/src/statsd.rs
```diff
@@ -17,13 +17,13 @@ pub enum CardinalityLimiterCounters {
     /// Incremented for every hash which was served from the in memory cache.
     ///
     /// This metric is tagged with:
-    ///  - `scope`: The scope of check operation.
+    ///  - `id`: The id of the enfored limit.
```
jjbayer:
```suggestion
    ///  - `id`: The id of the enforced limit.
```

same below

jjbayer (approved):

relay-cardinality/src/redis.rs
Dav1dde: Yes they are, a `(None, None, None)` would be a total global cardinality limit, `(Some, Some, Some)` would be a project scoped limit for a certain namespace.

They only somewhat invalid config would be `(Some/None, None, Some)`, it works, just does not make sense, since a project will always have an org (hopefully).

relay-cardinality/src/redis.rs
Dav1dde: Ah sorry, forgot to adjust the names when changing the doc, will fix

Issue #2797, opened 2023-11-30
Untag during ingestion (ie. remove tag key, keep data)


Dav1dde:
We want the user to be able to retroactively fix problems with cardinality by dropping or removing high cardinality or unnecessary tags.

Epic: https://github.com/getsentry/sentry/issues/60263

PR #2973, opened 2024-01-19
feat(metrics): Block metric tags with glob pattern

TBS1996:
https://github.com/getsentry/relay/issues/2797
 
This PR introduces a way of deleting certain tags based on glob pattern matching.


relay-dynamic-config/src/metrics.rs
```diff
@@ -21,15 +21,29 @@ pub struct Metrics {
     /// List of patterns for blocking metrics based on their name.
     #[serde(skip_serializing_if = "GlobPatterns::is_empty")]
     pub denied_names: GlobPatterns,
+    /// List of patterns of tags to remove.
```
jan-auer: Could you add a bit more documentation on what that does? In contrast to the option above, it is worth pointing out that removing tags doesn't drop the overall metric bucket (i.e. the metric is still queryable without the removed tags).

You can also expand the docs on `TagBlock`, in which case it would be useful to actively refer to the docs on the struct.

relay-server/src/services/project.rs
```diff
@@ -579,52 +579,72 @@ impl Project {
...
-        };
+        let metrics_config = &state.config.metrics;
 
-        buckets.retain(|bucket| {
-            if metrics.denied_names.is_match(&bucket.name) {
+        metrics.retain_mut(|bucket| {
+            if Self::metric_name_denied(metrics_config, bucket) {
                 relay_log::trace!(mri = bucket.name, "dropping metrics due to block list");
                 false
+            } else if Self::metric_namespace_invalid(state, bucket) {
```
jan-auer: `metric_namespace_invalid` tracks trace logs internally when it drops the bucket, but `metric_name_denied` does not. It's easy to miss this difference, so could we align this?

This function actually drops due to two reasons: an MRI that cannot be parsed, or an unknown namespace. If you prefer to have this in sub functions, I'd suggest to split those into two functions that just return booleans, and then log the messages here at the upper level. This would require you to pass the parsed MRI into the namespace function instead of the entire bucket.

relay-server/src/services/project.rs
```diff
@@ -579,52 +579,72 @@ impl Project {
...
-    fn apply_metrics_deny_list(deny_list: &ErrorBoundary<Metrics>, buckets: &mut Vec<Bucket>) {
-        let ErrorBoundary::Ok(metrics) = deny_list else {
-            return;
-        };
+        let metrics_config = &state.config.metrics;
 
-        buckets.retain(|bucket| {
-            if metrics.denied_names.is_match(&bucket.name) {
+        metrics.retain_mut(|bucket| {
+            if Self::metric_name_denied(metrics_config, bucket) {
```
jan-auer: All of these functions are not associated with the `Project` struct in any way. Can we make them free functions?

relay-server/src/services/project.rs
```diff
@@ -579,52 +579,72 @@ impl Project {
...
-            let verdict = match mri.namespace {
-                MetricNamespace::Sessions => true,
-                MetricNamespace::Transactions => true,
-                MetricNamespace::Spans => state.has_feature(Feature::SpanMetricsExtraction),
-                MetricNamespace::Custom => state.has_feature(Feature::CustomMetrics),
-                MetricNamespace::Unsupported => false,
-            };
+        for tag_block_config in &metrics.denied_tags {
+            let name_pattern = &tag_block_config.name;
+            let tag_pattern = &tag_block_config.tag;
```
jan-auer: _nit_: Could be inlined for a smaller function body.

relay-server/src/services/project.rs
```diff
@@ -579,52 +579,72 @@ impl Project {
...
+                    .tags
+                    .retain(|tag_key, _| !tag_pattern.is_match(tag_key));
             }
+        }
+    }
 
-            verdict
-        });
+    fn metric_name_denied(deny_list: &ErrorBoundary<Metrics>, bucket: &Bucket) -> bool {
+        let ErrorBoundary::Ok(metrics) = deny_list else {
```
jan-auer: We now check the `ErrorBoundary` twice in two consecutive functions. If we pull this to the upper level, we could check it once, and then conditionally call the inner functions (or inline them, since we're really only talking about a single line of code, and a handful in the other case).

jan-auer (commented):
Could you add an integration test that shows this new config being used? It would be great to get some input from @obostjancic on the usability of this config.

relay-server/src/services/project.rs
TBS1996: agree thats better! I moved it up to the upper function now.

obostjancic:
> Could you add an integration test that shows this new config being used? It would be great to get some input from @obostjancic on the usability of this config.

cc @iambriccardo since he is implementing blocking endpoint in https://github.com/getsentry/sentry/pull/63598

TBS1996:
>Could you add an integration test that shows this new config being used?  

 
Done!

relay-dynamic-config/src/metrics.rs
```diff
@@ -21,15 +21,31 @@ pub struct Metrics {
...
+            && self.denied_names.is_empty()
+            && self.denied_tags.is_empty()
     }
 }
 
+/// Configuration for removing tags matching the `tag` pattern on metrics whose name matches the `name` pattern.
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct TagBlock {
+    /// Name of metric of which we want to remove certain tags.
+    pub name: GlobPatterns,
```
jan-auer, line 44: cc @obostjancic : If the product now wants to drop a tag for _all_ metrics in the `custom` namespace, you would configure the glob pattern `"?:custom/*"`.

jan-auer (approved):
This reads really well, thank you for applying the suggestions. Good to merge 👍

PR #2978, opened 2024-01-22
ref(cardinality): Track rejections in an accumulator

Dav1dde:
Instead of returning an iterator of accumulated rejections track all rejections in the accumulator directly. This allows the caller to pass in whatever accumulator it needs and skips at least one round of allocations (Before `Vec<Rejection> -> HashSet<usize>`, now it is just `HashSet<usize>`. This is also a preparation for when we want to track more than just which element was rejected, we can easily include the `CardinalityLimit` information in the argument to the accumulator.

tldr; removes 2 allocations

#skip-changelog


relay-cardinality/src/limiter.rs
```diff
@@ -20,19 +19,27 @@ pub struct Scoping {
...
     ///
     /// Returns an iterator containing only accepted entries.
-    fn check_cardinality_limits<I>(
+    fn check_cardinality_limits<I, T>(
         &self,
         scoping: Scoping,
         limits: &[CardinalityLimit],
         entries: I,
-    ) -> Result<Box<dyn Iterator<Item = Rejection>>>
+        outcomes: &mut T,
```
Dav1dde: I initially had this as `&mut dyn Outcomes`, get's rid of the generic at some dynamic dispatch. Not sure what to prefer here.

relay-cardinality/src/limiter.rs
```diff
@@ -20,19 +19,27 @@ pub struct Scoping {
     pub project_id: ProjectId,
 }
 
+/// Accumulator of all cardinality outcomes.
+pub trait Outcomes {
```
jjbayer: Do you actually need a trait on this level? Both implementations wrap a HashSet, so could we just make the set element a type parameter?

If we keep the trait: "outcome" is almost like a reserved word at sentry, can we call it `Rejections`?

Dav1dde: Yeah, it could just be a `HashSet`, the idea was that in the future we can do `outcomes.reject(entry, quota)` when we need the quota, which means we can immediately accumulate all the rejections and additionally do something with the quota info.

Also with the inner value of EntryId being hidden and we need the inner value in the HashSet I came up with this trait.

relay-cardinality/src/limiter.rs
```diff
@@ -20,19 +19,27 @@ pub struct Scoping {
...
+    fn reject(&mut self, entry_id: EntryId);
+}
+
 /// Limiter responsible to enforce limits.
 pub trait Limiter {
     /// Verifies cardinality limits.
     ///
     /// Returns an iterator containing only accepted entries.
-    fn check_cardinality_limits<I>(
+    fn check_cardinality_limits<I, T>(
```
jjbayer: nit: Why `I` and `T`? `E` and `R` would seem more obvious to me.

jjbayer (approved):

iker-barriocanal (approved):

Dav1dde:
Looks like this was 10% less CPU time spent in the cardinality limiter (excluding Redis) 💪

Issue #2980, opened 2024-01-22
Relay User Report Envelope API is broken


JoshFerge:
Customers are reporting the user report API is no longer working via javascript SDK.

https://github.com/getsentry/sentry-javascript/issues/10229
https://github.com/getsentry/sentry-javascript/issues/10280
https://www.notion.so/sentry/sentry-captureUserFeedback-Not-Working-d5aa7f6e2fd5416790820a0e2cf0cb47

I have traced this issue to Relay, where git-bisect reveals that the PR https://github.com/getsentry/relay/pull/2862 breaks the ingest of user report payloads (note these are the old user reports, captured via the javascript API call `Sentry.captureUserFeedback`).

The envelope I have used to test this is:

envelope 1 (event creation): 
```
{"event_id":"4cec9f3e1f214073b816e0f4de5f59b1","sent_at":"2024-01-22T17:16:32.564Z","sdk":{"name":"sentry.javascript.browser","version":"7.80.0"},"trace":{"environment":"development","public_key":"ac390a22a11167bbe4318c072a75fdd7","trace_id":"48a06ef0fc6448f69726dc5ac743f1e9"}}
{"type":"event"}
{"message":"User Feedback","level":"info","event_id":"4cec9f3e1f214073b816e0f4de5f59b1","platform":"javascript","request":{"url":"http://localhost:5173/","headers":{"User-Agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"}},"timestamp":1705943792.563,"environment":"development","sdk":{"integrations":["InboundFilters","FunctionToString","TryCatch","Breadcrumbs","GlobalHandlers","LinkedErrors","Dedupe","HttpContext"],"name":"sentry.javascript.browser","version":"7.80.0","packages":[{"name":"npm:@sentry/browser","version":"7.80.0"}]},"contexts":{"trace":{"trace_id":"48a06ef0fc6448f69726dc5ac743f1e9","span_id":"932fe0ad83d688bf"}}}
```

envelope 2 (user report):
```
{"event_id":"4cec9f3e1f214073b816e0f4de5f59b1","sent_at":"2024-01-22T17:16:32.564Z","sdk":{"name":"sentry.javascript.browser","version":"7.80.0"}}
{"type":"user_report"}
[...]

PR #2981, opened 2024-01-22
fix: Process user reports in separate group

olksdr:
Process user reports in separate processing group. 

Related https://github.com/getsentry/relay/issues/2980

### DIFF
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -192,6 +192,17 @@ impl ProcessingGroup {
             ))
         }
 
+        // Extract user report and user feedback.
+        let report_items = envelope.take_items_by(|item| {
+            matches!(item.ty(), &ItemType::UserReport | &ItemType::ClientReport)
+        });
+        if !report_items.is_empty() {
+            grouped_envelopes.push((
+                ProcessingGroup::UserReport,
+                Envelope::from_parts(headers.clone(), report_items),
+            ))
+        }
+
         // Extract all the items which require an event into separate envelope.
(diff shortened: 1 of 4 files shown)


JoshFerge (approved):
works with my local test.

olksdr:
@JoshFerge  thanks for testing this!

relay-server/src/services/processor.rs
```diff
@@ -192,6 +192,17 @@ impl ProcessingGroup {
             ))
         }
 
+        // Extract user report and user feedback.
+        let report_items = envelope.take_items_by(|item| {
+            matches!(item.ty(), &ItemType::UserReport | &ItemType::ClientReport)
```
jan-auer, line 197: Client reports are very different from user reports: They carry outcomes, so like metrics they can be processed either completely separately or in a common envelope with other ungrouped items. 

The similarity in the name `ClientReport` and `UserReport` is merely coincidental.

jan-auer, line 197: User reports are items that require event payloads, hence should be processed in the same group as attachments (or standalone attachments). We should follow up with refactoring the standalone attachment group above (really the "doesn't have an event"-group) to cover all items that have `requires_event` set to true. 

Note that the `requires_event` flag means that these items are existentially bound to an event instance in Sentry. That's the reason they are grouped together with events: If the event is dropped, so will be the other items. However, that doesn't mean the event payload must be in the same envelope. We should verify that the  other pipelines handle that correctly.

jan-auer (approved):

PR #2982, opened 2024-01-22
fix(webvitals): Drop negative web vitals from measurements

edwardgou-sentry:
Add `allow_negative` to `BuiltinMeasurementKey`. Filter out negative BuiltinMeasurements if `allow_negative` is false.

### DIFF
--- a/relay-event-normalization/src/normalize/mod.rs
+++ b/relay-event-normalization/src/normalize/mod.rs
@@ -36,6 +36,12 @@ mod contexts;
 pub struct BuiltinMeasurementKey {
     name: String,
     unit: MetricUnit,
+    #[serde(default, skip_serializing_if = "is_false")]
+    allow_negative: bool,
+}
+
+fn is_false(b: &bool) -> bool {
+    !b
 }
 
 impl BuiltinMeasurementKey {
@@ -44,6 +50,7 @@ impl BuiltinMeasurementKey {
         Self {
             name: name.into(),
             unit,
+            allow_negative: false,
         }
     }
 
@@ -55,6 +62,11 @@ impl BuiltinMeasurementKey {
     /// Returns the unit of the built in measurement key.
     pub fn unit(&self) -> &MetricUnit {
         &self.unit
+    }
+
+    /// Return true if the built in measurement key allows negative values.
+    pub fn allow_negative(&self) -> &bool {
+        &self.allow_negative
     }
 }
 
--- a/relay-event-normalization/src/normalize/mod.rs
+++ b/relay-event-normalization/src/normalize/mod.rs
@@ -36,7 +36,7 @@ mod contexts;
 pub struct BuiltinMeasurementKey {
     name: String,
     unit: MetricUnit,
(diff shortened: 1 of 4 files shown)


jan-auer (commented):
Thank you for this change, it makes sense to consider negative web vitals as invalid and remove them during normalization. 

Instead of hard-coding this, we can use [`builtin_measurements`](https://github.com/getsentry/relay/blob/c0b970edbc97f98e6f624510156810b08f2eb6c1/relay-event-normalization/src/normalize/mod.rs#L33-L39) from the global config. This way, the same normalization automatically applies to new measurements added in the future and extends beyond web vitals. Most of our measurements would require positive values (except for things measuring relative change, such as memory), so we can make this a new default:

```rust
/// Defines a builtin measurement.
#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Hash, Eq)]
pub struct BuiltinMeasurementKey {
    name: String,
    unit: MetricUnit,
    allow_negative: bool,
}

fn is_false(b: &bool) -> bool {
    !b
}
```

There's already a function called [`remove_invalid_measurements`](https://github.com/getsentry/relay/blob/c0b970edbc97f98e6f624510156810b08f2eb6c1/relay-event-normalization/src/event.rs#L965) that has access to this configuration and is used to prune invalid measurements. It is also called from within `normalize_measurments` ([here](https://github.com/getsentry/relay/blob/c0b970edbc97f98e6f624510156810b08f2eb6c1/relay-event-normalization/src/event.rs#L709)), so the only thing left to do is move it to the bottom of the function after calling `compute_measurements`.

By returning `false` from the [`retain` closure](https://github.com/getsentry/relay/blob/c0b970edbc97f98e6f624510156810b08f2eb6c1/relay-event-normalization/src/event.rs#L976), you can evict measurements from the list. An example where builtin measurement keys are checked is [here](https://github.com/getsentry/relay/blob/c0b970edbc97f98e6f624510156810b08f2eb6c1/relay-event-normalization/src/event.rs#L1008-L1009). You can add this additional check and return `false` if the measurement's value is negative and the matching config does not permit negative values.

edwardgou-sentry:
Thanks @jan-auer ! I've made updates to BuiltinMeasurementKey an moved the negative check to remove_invalid_measurements

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -36,6 +36,12 @@ mod contexts;
 pub struct BuiltinMeasurementKey {
     name: String,
     unit: MetricUnit,
+    #[serde(default, skip_serializing_if = "is_false")]
```
jan-auer: In my previous review I missed that there is `#[serde(default)]` on the struct, so the one here is redundant. 

```suggestion
    #[serde(skip_serializing_if = "is_false")]
```

jan-auer (approved):
Thank you!

relay-event-normalization/src/normalize/mod.rs
edwardgou-sentry: updated!

PR #2984, opened 2024-01-23
fix(user_report): Treat UserReport like standalone attachment

### LINKED ISSUE #2980: Relay User Report Envelope API is broken

olksdr:
If the user reports are send in separate envelope from the event, we treat it like a standalone attachment.

* changed the processing grouping 
* added few tests to verify the behaviour 
* client reports are in separate processing group, since they do not require any events

fix: https://github.com/getsentry/relay/issues/2980

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 - Add automatic PII scrubbing to `logentry.params`. ([#2956](https://github.com/getsentry/relay/pull/2956))
 - Avoid producing `null` values in metric data. These values were the result of Infinity or NaN values extracted from event data. The values are now discarded during extraction. ([#2958](https://github.com/getsentry/relay/pull/2958))
-- Add user reports to a correct processing group. ([#2981](https://github.com/getsentry/relay/pull/2981), [#2984](https://github.com/getsentry/relay/pull/2984))
+- Fix processing of user reports. ([#2981](https://github.com/getsentry/relay/pull/2981), [#2984](https://github.com/getsentry/relay/pull/2984))
 
 ## 24.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 - Add automatic PII scrubbing to `logentry.params`. ([#2956](https://github.com/getsentry/relay/pull/2956))
 - Avoid producing `null` values in metric data. These values were the result of Infinity or NaN values extracted from event data. The values are now discarded during extraction. ([#2958](https://github.com/getsentry/relay/pull/2958))
-- Process user reports in separate processing group. ([#2981](https://github.com/getsentry/relay/pull/2981))
+- Add user reports to a correct processing group. ([#2981](https://github.com/getsentry/relay/pull/2981), [#2984](https://github.com/getsentry/relay/pull/2984))
 
 ## 24.1.0
 
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -105,24 +105,26 @@ pub enum ProcessingGroup {
     Transaction,
     /// All the items which require (have or create) events.
     ///
-    /// This includes: errors, NEL, security reports, user and clients reports, some of the
+    /// This includes: errors, NEL, security reports, user reports, some of the
     /// attachments.
     Error,
     /// Session events.
     Session,
-    /// Attachments (including UserReports) which can be sent alone without any event attached to it in the current
-    /// envelope.
-    StandaloneAttachment,
+    /// Standalone items which can be sent alone without any event attached to it in the current
+    /// envelope e.g. some attachments, user reports.
+    Standalone,
     /// Outcomes.
     ClientReport,
+    /// Replays and ReplayRecordings.
     Replay,
     /// Crons.
     CheckIn,
+    /// Spans.
     Span,
-    /// The unknow item types will be forwarded upstream (to processing Relay), where we will
-    /// decide what to do with it.
+    /// Unknown item types will be forwarded upstream (to processing Relay), where we will
+    /// decide what to do with them.
     ForwardUnknown,
-    /// All the events in the envelope we failed to group.
+    /// All the items in the envelope that could not be grouped.
     Ungrouped,
 }
 
@@ -144,25 +146,6 @@ impl ProcessingGroup {
                 (ProcessingGroup::Error, envelope)
             });
         grouped_envelopes.extend(nel_envelopes);
-
-        // Extract all standalone attachments and reports.
-        //
-        // Note: only if there are no items in the envelope which can create events, otherwise they
-        // will be in the same envelope with all require event items.
-        if !envelope.items().any(Item::creates_event) {
-            let standalone_attachment_items = envelope.take_items_by(|item| {
-                matches!(
-                    item.ty(),
-                    &ItemType::Attachment | &ItemType::FormData | &ItemType::UserReport
-                )
-            });
-            if !standalone_attachment_items.is_empty() {
-                grouped_envelopes.push((
-                    ProcessingGroup::StandaloneAttachment,
-                    Envelope::from_parts(headers.clone(), standalone_attachment_items),
-                ))
-            }
-        };
 
         // Extract replays.
         let replay_items = envelope.take_items_by(|item| {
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -13,7 +13,7 @@
 
 - Add automatic PII scrubbing to `logentry.params`. ([#2956](https://github.com/getsentry/relay/pull/2956))
 - Avoid producing `null` values in metric data. These values were the result of Infinity or NaN values extracted from event data. The values are now discarded during extraction. ([#2958](https://github.com/getsentry/relay/pull/2958))
-- Process user reports in separate processing group. ([#2981](https://github.com/getsentry/relay/pull/2981))
+- Add user reports to a correct processing group. ([#2981](https://github.com/getsentry/relay/pull/2981), [#2984](https://github.com/getsentry/relay/pull/2984))
```
iker-barriocanal: nit: "separate" seems more explicit than "correct", could we find another term to make the changelog a bit more clear?

relay-server/src/services/processor.rs
```diff
@@ -144,11 +145,16 @@ impl ProcessingGroup {
             });
         grouped_envelopes.extend(nel_envelopes);
 
-        // Extract all standalone attachments.
-        // Note: only if there is no items in the envelope which can create events.
+        // Extract all standalone attachments and reports.
+        //
+        // Note: only if there is no items in the envelope which can create events, otherwise they
```
```suggestion
        // Note: only if there are no items in the envelope which can create events, otherwise they
```

relay-server/src/services/processor.rs
```diff
@@ -110,10 +110,11 @@ pub enum ProcessingGroup {
     Error,
     /// Session events.
     Session,
-    /// Attachments which can be sent alone without any event attached to it in the current
+    /// Attachments (including UserReports) which can be sent alone without any event attached to it in the current
```
Dav1dde: Would it make sense to have a separate group of `StandaloneUserReport`?

olksdr: The user reports are sent to the attachment kafka topic as well, and from what I understood we treat them like attachments, so the group name kind of reflects the purpose, I think. 

But maybe @jan-auer  correct me here and provide a little bit more context here.

tests/integration/test_user_report.py
```diff
@@ -0,0 +1,64 @@
...
+def test_user_report_with_event(
+    relay_with_processing, mini_sentry, attachments_consumer
+):
+    project_id = 42
+    relay = relay_with_processing()
+    mini_sentry.add_full_project_config(project_id)
+
+    event_id = uuid.uuid1().hex
+
+    error_payload = {
```
iker-barriocanal, line 40: I think we could benefit from a test where the error is dropped but not the user report, what do you think?

iker-barriocanal (approved):

tests/integration/test_user_report.py
olksdr: That's might be good, what condition do you have in mind for this to happen?

CHANGELOG.md
olksdr: I will think about this a bit hot to rephrase it.

tests/integration/test_user_report.py
olksdr: That's might be good, what condition do you have in mind for this to happen?

relay-server/src/services/processor.rs
Dav1dde: Makes sense 👍

jan-auer: Maybe we should simply rename this group. This is a group of standalone items that belong to events. Not sure what the best concise name is, but effectively everything that has `requires_event() = true` while there's no event in the envelope has to go in there.

relay-server/src/services/processor.rs
```diff
@@ -144,11 +145,16 @@ impl ProcessingGroup {
...
+        // Extract all standalone attachments and reports.
+        //
+        // Note: only if there are no items in the envelope which can create events, otherwise they
+        // will be in the same envelope with all require event items.
         if !envelope.items().any(Item::creates_event) {
             let standalone_attachment_items = envelope.take_items_by(|item| {
-                matches!(item.ty(), &ItemType::Attachment | &ItemType::FormData)
+                matches!(
+                    item.ty(),
+                    &ItemType::Attachment | &ItemType::FormData | &ItemType::UserReport
```
jan-auer: Can we replace this with `requires_event` to prevent future items missing here?

relay-server/src/services/processor.rs
```diff
@@ -110,10 +110,11 @@ pub enum ProcessingGroup {
...
     /// Session events.
     Session,
-    /// Attachments which can be sent alone without any event attached to it in the current
+    /// Attachments (including UserReports) which can be sent alone without any event attached to it in the current
     /// envelope.
     StandaloneAttachment,
-    UserReport,
+    /// Outcomes.
+    ClientReport,
     Replay,
```
jan-auer, line 119: While we're at it, could we add docs to the remaining group variants and fix typos in the existing docs?

```rust
    /// Unknown item types will be forwarded upstream (to processing Relay), where we will
    /// decide what to do with it.
    ForwardUnknown,
    /// All the items in the envelope that could not be grouped.
    Ungrouped,
```

relay-server/src/services/processor.rs
olksdr: Done and renamed it to `Standalone` for now.

relay-server/src/services/processor.rs
olksdr: Done

relay-server/src/services/processor.rs
olksdr: fixed I think 😄  thanks for pointing this out

tests/integration/test_user_report.py
iker-barriocanal: Sending an incorrect error should be good enough; but alternatively, send an error that's filtered out by e.g. inbound filters. The user report should be consumed in both cases.

jan-auer (approved):
Looks good, thank you for applying the suggestions! 

In a follow-up, we can clarify the semantics of `creates_event` and `requires_event`, and consider better alternatives to these functions. They do not live up to today's needs anymore.

Issue #2967, opened 2024-01-18
Session Aggregates are randomly dropped


Dav1dde:
Session aggregates are being dropped/lost before the reach Kafka/Storage.

We could verify by peeking into the Kafka Topic that the aggregates (as metrics) never make it into Kafka.

From the Kafka Topic: 

```
 (datetime.datetime(2024, 1, 18, 10, 9), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 11), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 12), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 14), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 17), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 19), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 21), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 23), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 25), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 29), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 31), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 32), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 34), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 35), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 40), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 49), 30.0),
 (datetime.datetime(2024, 1, 18, 10, 56), 30.0)]
```

In the UI:
![image](https://github.com/getsentry/relay/assets/255721/fae8b100-bd1d-4600-8a46-2df25218953d)

Expected to see one entry every minute with a value of 30.

Script which generates an envelop containing a session aggregate every minute:

```sh
#!/usr/bin/env bash


URL="https://<INGEST URL>/api/<PROJECT ID>/envelope/?sentry_key=<KEY>&sentry_version=7&sentry_client=sentry.javascript.browser%2F7.93.0"

RELEASE="r1"
ENVIRONMENT="production"
USER_AGENT="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"

while true; do
[...]

PR #2987, opened 2024-01-23
fix(metrics): Fetch project config when metrics are received

jjbayer:
Metrics that arrive through the global metrics endpoint do not currently trigger a refetch of the project config. If a single processing relay only receives metrics traffic (no envelopes) for a specific project, metrics might get stuck in the pre-aggregator a.k.a. metrics buffer.

ref: https://github.com/getsentry/relay/issues/2967.


relay-server/src/services/project_cache.rs
```diff
@@ -859,17 +859,19 @@ impl ProjectCacheBroker {
...
-            .merge_buckets(
-                aggregator,
-                outcome_aggregator,
-                envelope_processor,
-                message.buckets(),
-            );
+
+        let project = self.get_or_create_project(message.project_key());
+        project.prefetch(project_cache, false);
+        project.merge_buckets(
```
jjbayer, line 869: Without this change, the added integration test fails.

relay-server/src/services/project.rs
```diff
@@ -711,6 +711,10 @@ impl Project {
                         relay_log::debug!("sending metrics straight to aggregator");
                         let state = state.clone();
 
+                        // TODO: When the state is present but expired, we should send buckets
+                        // to the metrics buffer instead. In practice, the project state should be
+                        // refreshed at the time when the buckets emerge from the aggregator though.
```
jjbayer, line 716: I think we should do this, but I'll leave it for a follow-up.

tests/integration/test_metrics.py
```diff
@@ -241,6 +241,36 @@ def test_global_metrics(mini_sentry, relay):
...
+            "name": "c:transactions/foo@none",
+            "value": 17.0,
+            "type": "c",
+        }
+    ]
+    relay.send_metrics_batch(
+        {"buckets": {public_key: metrics}},
+    )
+
+    envelope = mini_sentry.captured_events.get(timeout=3)
```
jjbayer, line 265: This `.get()` fails without the `prefetch`.

Dav1dde (approved):

relay-server/src/services/project_cache.rs
```diff
@@ -859,17 +859,19 @@ impl ProjectCacheBroker {
...
-        // Only keep if we have an aggregator, otherwise drop because we know that we were disabled.
-        self.get_or_create_project(message.project_key())
-            .merge_buckets(
-                aggregator,
-                outcome_aggregator,
-                envelope_processor,
-                message.buckets(),
-            );
+
+        let project = self.get_or_create_project(message.project_key());
```
Dav1dde, line 867: I wonder if we should just always prefetch in `get_or_create_project`. I found the same "bug" for metric metadata as well.

TBS1996 (approved):

relay-server/src/services/project_cache.rs
jjbayer: @Dav1dde I would even go further and unify the multiple state enums we have for project configs at the moment:

* `GetOrFetch` - either a cached project config or a channel
* `State` - either a cached project config or a metrics buffer
* `ExpiryState` - either `Updated`, `Stale`, or `Expired`

We might be able to make a single enum out of these, and give them an interface that makes it impossible to access a project without prefetching.

PR #2990, opened 2024-01-23
ref(cardinality): Use cardinality limits from project config

Dav1dde:
Reads the cardinality config from the project config instead of generating them from the Relay config.

Sentry side implemented in: https://github.com/getsentry/sentry/pull/63562


olksdr (approved):
lgtm

Is there a test which checks the correctness of usage limits from the project config?

PR #2993, opened 2024-01-23
chore(spool): Enable periodic unspool

olksdr:
Currently in order to trigger unspool of buffered envelopes the new project state update must be received from the upstream, which sometimes can take longer time.

To make unspool action a bit more predictable let's enable periodic unspool initiated by project cache. This way:
* spool service does not have to know when to unspool
* spool service just has to serve the request and communicate back the results 
* project cache will take care of the initiating the action and only queue the envelopes when certain criteria are met

part of: https://github.com/getsentry/team-ingest/issues/267

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 - Copy event measurements to span & normalize span measurements. ([#2953](https://github.com/getsentry/relay/pull/2953))
 - Add `allow_negative` to `BuiltinMeasurementKey`. Filter out negative BuiltinMeasurements if `allow_negative` is false. ([#2982](https://github.com/getsentry/relay/pull/2982))
 - Add possiblity to block metrics or their tags with glob-patterns. ([#2954](https://github.com/getsentry/relay/pull/2954), [#2973](https://github.com/getsentry/relay/pull/2973))
-- Enable periodic unspool of the buffered envelopes. ([#2993](https://github.com/getsentry/relay/pull/2993))
+- Enable throttled periodic unspool of the buffered envelopes. ([#2993](https://github.com/getsentry/relay/pull/2993))
 
 **Bug Fixes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,8 @@
 - Proactively move on-disk spool to memory. ([#2949](https://github.com/getsentry/relay/pull/2949))
 - Default missing `Event.platform` and `Event.level` fields during light normalization. ([#2961](https://github.com/getsentry/relay/pull/2961))
 - Copy event measurements to span & normalize span measurements. ([#2953](https://github.com/getsentry/relay/pull/2953))
-- Add possiblity to block metrics with glob-patterns. ([#2954](https://github.com/getsentry/relay/pull/2954))
-- Enable periodic unspool of the buffered envelopes. ([#2993](https://github.com/getsentry/relay/pull/2993))
 - Add possiblity to block metrics or their tags with glob-patterns. ([#2954](https://github.com/getsentry/relay/pull/2954), [#2973](https://github.com/getsentry/relay/pull/2973))
+- Enable periodic unspool of the buffered envelopes. ([#2993](https://github.com/getsentry/relay/pull/2993))
 
 **Bug Fixes**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Default missing `Event.platform` and `Event.level` fields during light normalization. ([#2961](https://github.com/getsentry/relay/pull/2961))
 - Copy event measurements to span & normalize span measurements. ([#2953](https://github.com/getsentry/relay/pull/2953))
 - Add possiblity to block metrics with glob-patterns. ([#2954](https://github.com/getsentry/relay/pull/2954))
+- Enable periodic unspool of the buffered envelopes. ([#2993](https://github.com/getsentry/relay/pull/2993))
 
 **Bug Fixes**:
 
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -854,7 +854,7 @@ pub struct EnvelopeSpool {
     /// This is a hard upper bound and defaults to 524288000 bytes (500MB).
     #[serde(default = "spool_envelopes_max_memory_size")]
     max_memory_size: ByteSize,
-    /// The internal in milliseconds to trigger unspool.
+    /// The interval in milliseconds to trigger unspool.
     #[serde(default = "spool_envelopes_unspool_interval")]
     unspool_interval: u64,
 }
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -826,6 +826,11 @@ fn spool_envelopes_max_connections() -> u32 {
     20
 }
 
+/// Default interval to unspool buffered envelopes, 100ms.
+fn spool_envelopes_unspool_interval() -> u64 {
+    100
+}
+
 /// Persistent buffering configuration for incoming envelopes.
 #[derive(Debug, Serialize, Deserialize)]
 pub struct EnvelopeSpool {
@@ -849,6 +854,9 @@ pub struct EnvelopeSpool {
     /// This is a hard upper bound and defaults to 524288000 bytes (500MB).
     #[serde(default = "spool_envelopes_max_memory_size")]
     max_memory_size: ByteSize,
+    /// The internal in milliseconds to trigger unspool.
+    #[serde(default = "spool_envelopes_unspool_interval")]
+    unspool_interval: u64,
 }
 
 impl Default for EnvelopeSpool {
@@ -859,6 +867,7 @@ impl Default for EnvelopeSpool {
             min_connections: spool_envelopes_min_connections(),
             max_disk_size: spool_envelopes_max_disk_size(),
             max_memory_size: spool_envelopes_max_memory_size(),
+            unspool_interval: spool_envelopes_unspool_interval(), // 100ms
         }
     }
 }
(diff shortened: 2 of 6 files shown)


relay-config/src/config.rs
```diff
@@ -849,6 +854,9 @@ pub struct EnvelopeSpool {
     /// This is a hard upper bound and defaults to 524288000 bytes (500MB).
     #[serde(default = "spool_envelopes_max_memory_size")]
     max_memory_size: ByteSize,
+    /// The internal in milliseconds to trigger unspool.
```
Dav1dde:
```suggestion
    /// The interval in milliseconds to trigger unspool.
```

relay-server/src/services/project_cache.rs
```diff
@@ -913,6 +895,48 @@ impl ProjectCacheBroker {
...
+    }
+
+    /// Iterates the buffer index and tries to unspool the envelopes if the project state is still
+    /// valid.
+    ///
+    /// This makes sure we always moving the unspool forward, even if we do not fetch the project
+    /// states updates, but still can process data based on the existing cache.
+    fn handle_periodic_unspool(&mut self) {
+        self.buffer_unspool_handle.reset();
+        let keys = self.index.keys().cloned().collect::<Box<[_]>>();
```
Dav1dde: How big do we expect the index to be?

relay-server/src/services/project_cache.rs
```diff
@@ -913,6 +895,48 @@ impl ProjectCacheBroker {
...
+
+        // If there is nothing spooled, schedule the next check a little bit later.
+        if keys.is_empty() {
+            self.schedule_unspool();
```
Dav1dde, line 912: Is this to schedule the unspool before resetting?

Can we return in this branch to make it clear, e.g.
```rs
self.schedule_unspool();
self.buffer_unspool_backoff.reset();
return;
```

relay-server/src/services/project_cache.rs
```diff
@@ -520,36 +519,12 @@ impl GlobalConfigStatus {
 
 impl ProjectCacheBroker {
     fn set_global_config(&mut self, global_config: Arc<GlobalConfig>) {
-        let GlobalConfigStatus::Pending(project_keys) = std::mem::replace(
+        let GlobalConfigStatus::Pending = std::mem::replace(
```
Dav1dde: This is now just a `self.global_config = GlobalConfigStatus::Ready(global_config);`

But shouldn't we trigger an unspool here if it was pending so we don't have to wait for the backoff?

relay-server/src/services/project_cache.rs
```diff
@@ -502,14 +507,8 @@ struct ProjectCacheBroker {
...
     Ready(Arc<GlobalConfig>),
-    /// Project keys waiting for global config to dequeue from buffer.
-    ///
-    /// These project keys were used to try to dequeue the buffer, but were blocked because we
-    /// lacked a global config. When global config arrives, we will request project states for these keys again
-    /// which will trigger another dequeue that should be succesful. The reason we don't dequeue
-    /// directly when we receive [`GlobalConfig`] is that the [`ProjectState`]s might have expired in the meantime,
-    /// which would drop the envelopes when they are sent to ProjectCacheBroker::handle_processing.
-    Pending(BTreeSet<ProjectKey>),
+    /// The global config us not fetched yet.
```
jjbayer:
```suggestion
    /// The global config is not fetched yet.
```

relay-server/src/services/project_cache.rs
```diff
@@ -565,8 +540,13 @@ impl ProjectCacheBroker {
     /// forwarded to `handle_processing`.
     pub fn dequeue(&mut self, partial_key: ProjectKey) {
         // If we don't yet have the global config, we will defer dequeuing until we do.
-        if let GlobalConfigStatus::Pending(keys) = &mut self.global_config {
-            keys.insert(partial_key);
+        if let GlobalConfigStatus::Pending = self.global_config {
+            return;
+        }
+
+        // Do *not* attempt to unspool if there is more buffer guards are assigned than low
```
jjbayer:
```suggestion
        // Do *not* attempt to unspool if there is more buffer permits assigned than low
```

relay-server/src/services/project_cache.rs
```diff
@@ -913,6 +895,48 @@ impl ProjectCacheBroker {
         }
     }
 
+    /// Returns backoff timeout for an unspool attempt.
+    fn next_unspool_attempt(&mut self) -> Duration {
+        self.config.spool_envelopes_unspool_interval() + self.buffer_unspool_backoff.next_backoff()
```
jjbayer, line 891: nit: Instead of adding an interval to a backoff, could we use `ExponentialBackoff` instead of `RetryBackoff` and initialize it with an `initial_interval` of `self.config.spool_envelopes_unspool_interval()`?

relay-server/src/services/project_cache.rs
jjbayer: Could we use .`keys().peekable()` here?

relay-server/src/services/project_cache.rs
jjbayer: nit: Should the reset of the backoff be part of `schedule_unspool`?

relay-server/src/services/project_cache.rs
```diff
@@ -913,6 +895,48 @@ impl ProjectCacheBroker {
...
+        // If there is nothing spooled, schedule the next check a little bit later.
+        if keys.is_empty() {
+            self.schedule_unspool();
+        }
+
+        for project_key in keys.iter() {
+            if self
+                .projects
+                .get(project_key)
+                .map_or(false, |state| state.valid_state().is_some())
```
jjbayer: Is the project config actually being refreshed anywhere now? In the previous impl we had `.get_cached_state(project_cache, false)` which did an implicit fetch.

relay-server/src/services/project_cache.rs
Dav1dde: It's actually tricky, it doesn't work in the loop below because we need to hold a reference to the keys and then call `dequeue` which needs an exclusive reference to `self` (it also modifies the map we'd be iterating over). So without rewriting the `dequeue` method we have to clone all keys here.

relay-server/src/services/project_cache.rs
Dav1dde: It's actually tricky, it doesn't work in the loop below because we need to hold a reference to the keys and then call `dequeue` which needs an exclusive reference to `self` (it also modifies the map we'd be iterating over). So without rewriting the `dequeue` method we have to clone all keys here.

relay-server/src/services/project_cache.rs
```diff
@@ -683,8 +663,10 @@ impl ProjectCacheBroker {
             no_cache,
         );
 
-        if !state.invalid() {
-            self.dequeue(project_key);
+        // Schedule unspool if nothing is running at the moment.
```
iker-barriocanal, line 655: Should we check here too whether there are enough buffer guards available?

relay-server/src/services/project_cache.rs
olksdr: yeah. I have to get the copy of the project keys here, so I could iterate over them to schedule the dequeue.

relay-server/src/services/project_cache.rs
olksdr: nope, since if there is nothing to unspool, I want to wait a little bit longer before trying to unspool again. Once the unspool finishes with anything, we reset the backoff to initial state (which currently 100ms)

relay-server/src/services/project_cache.rs
olksdr: do you mean to use [`backoff::ExponentialBackoff`](https://docs.rs/backoff/latest/backoff/exponential/struct.ExponentialBackoff.html) directly?

We have a little better interface and we also could add some functions to make the setting of things for it more convenient.

relay-server/src/services/project_cache.rs
olksdr: It will be refreshed as it was before like normal refresh operation. 
Here I just want to check for the valid states, and if there are try to unspool them.

relay-server/src/services/project_cache.rs
olksdr: nope, the check is done in `handle_periodic_unspool` now.

relay-server/src/services/project_cache.rs
jjbayer: Makes sense. We could make `RetryBackoff` more flexible (i.e. accept a custom initial delay), but that's out of scope for this PR.

relay-server/src/services/project_cache.rs
jjbayer: Gotcha. It might not be related to the goal of this PR but I feel like we should trigger a prefetch at this point. Otherwise we still depend on incoming traffic to get old data unspooled, right?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Default missing `Event.platform` and `Event.level` fields during light normalization. ([#2961](https://github.com/getsentry/relay/pull/2961))
 - Copy event measurements to span & normalize span measurements. ([#2953](https://github.com/getsentry/relay/pull/2953))
 - Add possiblity to block metrics or their tags with glob-patterns. ([#2954](https://github.com/getsentry/relay/pull/2954), [#2973](https://github.com/getsentry/relay/pull/2973))
+- Enable periodic unspool of the buffered envelopes. ([#2993](https://github.com/getsentry/relay/pull/2993))
```
jjbayer: If I understand correctly, the main purpose of this PR is to not unspool everything at once, right?

```suggestion
- Enable throttled periodic unspool of the buffered envelopes. ([#2993](https://github.com/getsentry/relay/pull/2993))
```

jjbayer (approved):

relay-server/src/services/project_cache.rs
olksdr: it's a valid point. 
I've switched to using `get_cached_state` instead, which will trigger the refresh of the project state if it's not valid or expired.

Dav1dde (approved):

relay-server/src/services/project_cache.rs
```diff
@@ -1268,4 +1309,89 @@ mod tests {
             assert!(buffer_rx.try_recv().is_err())
         }
     }
+
+    #[tokio::test]
+    async fn periodic_unspool() {
```
olksdr, line 1312: @jjbayer  Added a test, if you want to have a look.

jjbayer: Nice!

PR #2995, opened 2024-01-24
test(metric-meta): Add integration test for metric meta

Dav1dde:
#skip-changelog

### DIFF
--- a/tests/integration/test_metric_meta.py
+++ b/tests/integration/test_metric_meta.py
@@ -1,0 +1,99 @@
+import json
+import time
+
+from sentry_sdk.envelope import Envelope, Item, PayloadRef
+from datetime import datetime, timezone, time as dttime
+
+
+def assert_location(expected, actual):
+    # Compare without the 'type' which is not persisted into storage.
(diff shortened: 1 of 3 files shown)


tests/integration/test_metric_meta.py
```diff
@@ -0,0 +1,99 @@
+import json
+import time
+
+from sentry_sdk.envelope import Envelope, Item, PayloadRef
+from datetime import datetime, timezone, time as dttime
+
+
+def assert_location(expected, actual):
+    # Compare without the 'type' which is not persisted into storage.
+    assert {k: expected[k] for k in expected.keys() - {"type"}} == actual
```
iker-barriocanal, line 10: I think this should be equivalent"
```suggestion
    # Compare without the 'type' which is not persisted into storage.
    expected.pop('type')
    assert expected == actual
```

olksdr (approved):

iker-barriocanal (approved):

tests/integration/test_metric_meta.py
Dav1dde: This modifies the passed in reference `expected` so we'd at least have to make a copy of it

PR #2998, opened 2024-01-25
fix(cardinality): Fix cardinality cache, introduce stronger typing

Dav1dde:
The cache incorrectly compared timestamps to slots, this was not caught because both used `u64`. To prevent future mistakes like this we're strongly typing `UnixTimestamp` and `Slot` now.

I also added new tests for the cache to catch these issues in the future.

#skip-changelog

### DIFF
--- a/relay-cardinality/src/cache.rs
+++ b/relay-cardinality/src/cache.rs
@@ -1,10 +1,14 @@
+use std::fmt;
 use std::sync::{PoisonError, RwLock, RwLockReadGuard, RwLockWriteGuard};
 
+use relay_common::time::UnixTimestamp;
+
 use crate::redis::QuotaScoping;
+use crate::window::Slot;
 
 /// Cached outcome, wether the item can be accepted, rejected or the cache has no information about
 /// this hash.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum CacheOutcome {
     /// Hash accepted by cache.
     Accepted,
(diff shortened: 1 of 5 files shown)


olksdr (approved):

relay-cardinality/src/window.rs
```diff
@@ -33,20 +34,27 @@ impl SlidingWindow {
...
-    pub fn active_slot(&self, timestamp: u64) -> u64 {
-        self.iter(timestamp).last().unwrap_or(0)
+    pub fn active_slot(&self, timestamp: UnixTimestamp) -> Slot {
+        self.iter(timestamp).last().unwrap_or(Slot(0))
     }
 }
 
+/// A single slot from a [`SlidingWindow`].
+#[derive(Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Slot(pub u64);
```
jjbayer: nit: Could we even make the `u64` private and instead implement `Display` for the redis key?

jjbayer (approved):
Types ❤️

relay-cardinality/src/cache.rs
```diff
@@ -119,22 +132,26 @@ impl<'a> CacheUpdate<'a> {
...
+    hashes: hashbrown::HashSet<u32>,
+    current_slot: Slot,
+}
 
 impl ScopedCache {
-    fn check(&self, hash: u32, limit: u64) -> CacheOutcome {
-        if self.0.contains(&hash) {
+    fn check(&self, slot: Slot, hash: u32, limit: u64) -> CacheOutcome {
+        if slot != self.current_slot {
+            return CacheOutcome::Unknown;
```
iker-barriocanal, line 148: We shouldn't hit this branch in prod, but should we emit e.g. an error in case we missed something?

iker-barriocanal (approved):

relay-cardinality/src/cache.rs
Dav1dde: We hit this branch when you try to read from the cache with a new (or old) slot.

This should always be `Some` since we hold a `RwLockGuard` and it was previously inserted if missing, but I decided to write tests instead:
```rs
            if let Some(cache) = inner.cache.get_mut(key) {
                cache.insert(hash);
            }
```

Ideally I would get rid of this entirely which we can if we wrap the map value with `Arc<RwLock<>>` which I plan to revisit in the future.

Issue #2426, opened 2023-08-25
Unify glob utilities


jjbayer:
Relay currently has three separate utility classes for glob matching:

| module                | translation to regex | regex caching           | capturing | configurable | multiple | builder | mode   |
| --------------------- | -------------------- | ----------------------- | --------- | ------------ | -------- | ------- | ------ |
| `relay_common::glob`  | globset              | global                  |           | x            |          |         | binary |
| `relay_common::glob2` | manual               | local (with `LazyGlob`) | x         |              |          |         | utf8   |
| `relay_common::glob3` | globset              | local                   |           |              | x        | x       | utf8   |

Attempt to unify these.

Dav1dde:
- `glob` is only used for `librelay` which has been fully moved into the ffi crate. 
- `glob3` is fully replaced with `relay-pattern`.
- `glob2` still needs work to implement separator matching.

PR #3001, opened 2024-01-25
ref(dynamic-config): Use single string for tag-name pattern.

TBS1996:
tag-name pattern gets serialized by a string of the pattern now, rather than a vector of strings.

Reason: we will probably only use a singular pattern for the 'name' in the tagblock config. It's a bit awkward to therefore deserialize it from a vector of strings.  
 
This should make it more intuitive on the Sentry side, at the cost of some more LOC in relay. 

#skip-changelog

### DIFF
--- a/relay-common/Cargo.toml
+++ b/relay-common/Cargo.toml
@@ -18,6 +18,7 @@ parking_lot = "0.12.1"
 regex = { workspace = true }
 sentry-types = "0.31.3"
 serde = { workspace = true }
+serde_json = { workspace = true }
 
 [dev-dependencies]
 serde_test = "1.0.125"
--- a/relay-common/Cargo.toml
+++ b/relay-common/Cargo.toml
@@ -18,7 +18,6 @@ parking_lot = "0.12.1"
 regex = { workspace = true }
 sentry-types = "0.31.3"
 serde = { workspace = true }
-serde_json = { workspace = true }
 
 [dev-dependencies]
 serde_test = "1.0.125"
--- a/relay-common/src/glob3.rs
+++ b/relay-common/src/glob3.rs
@@ -18,8 +18,8 @@ pub struct GlobPattern(GlobPatterns);
 
 impl GlobPattern {
     /// Returns a new instance of [`GlobPattern`].
-    pub fn new(pattern: impl Into<String>) -> Self {
-        Self(GlobPatterns::new(vec![pattern.into()]))
+    pub fn new(pattern: String) -> Self {
+        Self(GlobPatterns::new(vec![pattern]))
     }
 
     /// Returns `true` if the pattern matches the given message.
@@ -50,7 +50,7 @@ impl<'de> Deserialize<'de> for GlobPattern {
     where
         D: Deserializer<'de>,
     {
-        let pattern: String = Deserialize::deserialize(deserializer)?;
+        let pattern: String = dbg!(Deserialize::deserialize(deserializer))?;
         let glob_patterns = GlobPatterns::new(vec![pattern]);
         Ok(Self(glob_patterns))
     }
@@ -159,6 +159,28 @@ mod tests {
                 $($pattern.to_string()),*
             ])
         };
+    }
+
+    #[test]
+    fn test_serialize_glob_pattern() {
+        let pattern = "foobar";
+        let quoted_pattern = "\"foobar\"";
+
+        let glob_pattern: GlobPattern = serde_json::from_str(quoted_pattern).unwrap();
+
+        assert_eq!(glob_pattern, GlobPattern::new(pattern.to_string()));
+        assert_eq!(
+            serde_json::to_string(&glob_pattern).unwrap(),
+            quoted_pattern
+        );
+    }
+
+    #[test]
+    fn test_glob_pattern_match() {
+        assert!(GlobPattern::new("foo".to_string()).is_match("foo"));
+        assert!(!GlobPattern::new("foo".to_string()).is_match("nope"));
+        assert!(GlobPattern::new("foo*".to_string()).is_match("foobarblub"));
+        assert!(GlobPattern::new("*blub".to_string()).is_match("foobarblub"));
     }
 
     #[test]
--- a/relay-common/src/glob3.rs
+++ b/relay-common/src/glob3.rs
@@ -12,6 +12,50 @@ fn is_match(globs: &[Regex], message: &[u8]) -> bool {
     globs.iter().any(|regex| regex.is_match(message.as_ref()))
 }
 
+/// A singular glob-pattern.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct GlobPattern(GlobPatterns);
+
+impl GlobPattern {
+    /// Returns a new instance of [`GlobPattern`].
+    pub fn new(pattern: impl Into<String>) -> Self {
+        Self(GlobPatterns::new(vec![pattern.into()]))
+    }
+
+    /// Returns `true` if the pattern match the given message.
+    pub fn is_match<S>(&self, message: S) -> bool
+    where
+        S: AsRef<[u8]>,
+    {
+        self.0.is_match(message)
+    }
+}
+
+impl Serialize for GlobPattern {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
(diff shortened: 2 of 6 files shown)


iambriccardo (approved):
Thanks so much for doing this, it makes the API much cleaner!

iambriccardo:
JFYI, approved by mistake, I would prefer if someone from ingest checks it.

relay-common/src/glob3.rs
```diff
@@ -99,6 +99,33 @@ impl<'de> Deserialize<'de> for GlobPatterns {
     }
 }
 
+/// Deserializes the [`GlobPatterns`] to a singular string.
+///
+/// This assumes there's only one pattern. Use in conjunction with [`deserialize_singular_pattern`].
+pub fn serialize_singular_pattern<S>(name: &GlobPatterns, serializer: S) -> Result<S::Ok, S::Error>
```
Dav1dde: I would much prefer if that was a newtype (or just straight up a separate type) instead which can only ever hold one pattern, `GlobPattern`?

TBS1996: I was considering a separate type, but thought it'd add too much code repetition. But I like the newtype idea. Maybe something like this?

```Rust
/// A singular glob-pattern.
pub struct GlobPattern(GlobPatterns);

impl GlobPattern {
    ///
    pub fn new(pattern: String) -> Self {
        Self(GlobPatterns::new(vec![pattern]))
    }

    ///
    pub fn is_match<S>(&self, message: S) -> bool
    where
        S: AsRef<[u8]>,
    {
        self.0.is_match(message)
    }
}
```

TBS1996: ok i went ahead and did it, since it was so simple anyway

relay-common/src/glob3.rs
```diff
@@ -12,6 +12,50 @@ fn is_match(globs: &[Regex], message: &[u8]) -> bool {
...
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for GlobPattern {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let single_string: String = Deserialize::deserialize(deserializer)?;
```
Dav1dde:
```rs
Deserialize::deserialize(deserializer).map(Self::new)
```

Should be enough for the entire impl

relay-common/src/glob3.rs
```diff
@@ -12,6 +12,50 @@ fn is_match(globs: &[Regex], message: &[u8]) -> bool {
     globs.iter().any(|regex| regex.is_match(message.as_ref()))
 }
 
+/// A singular glob-pattern.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct GlobPattern(GlobPatterns);
+
+impl GlobPattern {
+    /// Returns a new instance of [`GlobPattern`].
+    pub fn new(pattern: impl Into<String>) -> Self {
```
Dav1dde:
```suggestion
    pub fn new(pattern: String) -> Self {
```

We usually don't do this.

Personally I would use this only for builders, it is otherwise barely useful (usually only in tests) and hides accidental allocations (passing `&foo` instead of `foo` by accident).

relay-common/src/glob3.rs
```diff
@@ -12,6 +12,50 @@ fn is_match(globs: &[Regex], message: &[u8]) -> bool {
...
+}
+
+impl Serialize for GlobPattern {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        debug_assert_eq!(self.0.patterns.len(), 1);
+
+        match self.0.patterns.first() {
```
Dav1dde: I would take the panic here, the newtype was made with exactly that constraint, that is will always contain exactly one pattern:

```rs
serializer.serialize_str(&self.0.patterns[0])
```

Should discuss this with the team though or just ignore this suggestion.

relay-common/src/glob3.rs
```diff
@@ -12,6 +12,50 @@ fn is_match(globs: &[Regex], message: &[u8]) -> bool {
     globs.iter().any(|regex| regex.is_match(message.as_ref()))
 }
 
+/// A singular glob-pattern.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct GlobPattern(GlobPatterns);
+
+impl GlobPattern {
```
Dav1dde: Need tests for this, `is_match`, `Deserialize` and `Serialize`

relay-common/src/glob3.rs
TBS1996: neat! thx

relay-common/Cargo.toml
```diff
@@ -18,6 +18,7 @@ parking_lot = "0.12.1"
 regex = { workspace = true }
 sentry-types = "0.31.3"
 serde = { workspace = true }
+serde_json = { workspace = true }
```
Dav1dde: Only used for test? -> `dev-dependency`

relay-common/src/glob3.rs
```diff
@@ -12,6 +12,43 @@ fn is_match(globs: &[Regex], message: &[u8]) -> bool {
     globs.iter().any(|regex| regex.is_match(message.as_ref()))
 }
 
+/// A singular glob-pattern.
+#[derive(Debug, Clone, PartialEq, Default)]
```
Dav1dde: Does default make sense here? What is a default `GlobPattern`, the empty String?

`Serialize` will fail if the inner is `[]`.

relay-common/src/glob3.rs
```diff
@@ -12,6 +12,43 @@ fn is_match(globs: &[Regex], message: &[u8]) -> bool {
...
+pub struct GlobPattern(GlobPatterns);
```
jjbayer: I would rather redefine `GlobPatterns` as a wrapper around `Vec<GlobPattern>`. That would save you the custom serialization. Question is, why not use `glob2::LazyGlob`?

relay-dynamic-config/src/metrics.rs
```diff
@@ -41,9 +41,9 @@ impl Metrics {
 #[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
 pub struct TagBlock {
     /// Name of metric of which we want to remove certain tags.
-    pub name: GlobPatterns,
+    pub name: GlobPattern,
```
jjbayer: Why not use `glob2::LazyGlob`? It's already imported in this file. Do we need specific properties of `glob3`?

relay-common/src/glob3.rs
TBS1996: right, clear oversight, thank you!

relay-dynamic-config/src/metrics.rs
TBS1996: simply I wasn't aware of it, I redefined to lazyglob now. about glob2 vs glob3 I need to read up on the differences here

relay-common/src/glob3.rs
TBS1996: `GlobPatterns` is an old type though, and when it comes to multiple patterns it does what we want it to do. How about keeping it as-is and just getting rid of `GlobPattern` and replace with `LazyGlob` as you suggested? i'm pushing those changes now

relay-dynamic-config/src/metrics.rs
Dav1dde: I did see that, but I assumed `glob2` was the old thing and it was replaced by `glob3` ...

jjbayer: Here's a cheat sheet: https://github.com/getsentry/relay/issues/2426

TBS1996: @jjbayer  so problem now is, lazyglob needs two asterisks to match multiple wildcards, globpatterns needs only one. It would be bad to have separate behaviour. We could create a new `GlobPatterns` type with glob2 in the background, but that'd be even more work.  
 
I'd prefer to keep the ability to send a vector of strings, it'd make the API easier. Maybe I should revert to previous with GlobPattern(GlobPatterns) ? what's your thoughts

jjbayer: Yeah, OK. I would make `GlobPattern` a private helper of the `metrics` config module though (no need to introduce yet another glob utility if we want to unify them anyway).

PR #3003, opened 2024-01-25
fix(spans): Apply max. SQL expression depth after simplification

jjbayer:
Repeating OR/AND conditions are collapsed, but if the maximum query depth is reached, the limit is applied before we can collapse it. This leads to scrubbed queries like:

```sql
WHERE ((.. OR (..) OR (.. = %s AND b = %s) OR (a = %s AND b = %s))
```

Fix by splitting the query visitor into two steps (first normalize, then reduce depth).

This PR also removes redundant parentheses around binary operations (e.g. `a OR (b OR c)`).

ref: [internal issue](https://www.notion.so/sentry/Collapse-identical-OR-conditions-in-queries-40786f8cddb145e398e8ad2d5eb66522)

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/sql/parser.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/parser.rs
@@ -452,7 +452,7 @@ fn take_expr(expr: &mut Expr) -> Expr {
     swapped
 }
 
-/// Remove parentheses for equal operators, e.g. `(a OR b) OR c`.
+/// Removes parentheses for equal operators, e.g. `(a OR b) OR c`.
 ///
 /// Only use this function on operations which have the
 /// [associative property](https://en.wikipedia.org/wiki/Associative_property).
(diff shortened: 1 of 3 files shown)


relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -465,6 +452,54 @@ fn take_expr(expr: &mut Expr) -> Expr {
     swapped
 }
 
+/// Remove parentheses for equal operators, e.g. `(a OR b) OR c`.
```
olksdr:
```suggestion
/// Removes parentheses for equal operators, e.g. `(a OR b) OR c`.
```

olksdr (approved):

gggritso (approved):
👍🏻 makes sense, though I guess we're losing the `(` and `)` that used to surround collapsed conditions, is that right?

iker-barriocanal (approved):

jjbayer:
> 👍🏻 makes sense, though I guess we're losing the `(` and `)` that used to surround collapsed conditions, is that right?

@gggritso Yes, we're losing that. Because this normalization is quite general, there might also be a bunch of other cases where we now remove parentheses, i.e. there will be grouping changes. If you feel strongly about I can revert this part, but it will result in duplicate groups for the case we're trying to fix.

gggritso:
@jjbayer 🚢 👍🏻

PR #3004, opened 2024-01-25
ref(normalization): Split validation and normalization

iker-barriocanal:
This PR splits the event validation from normalization, making normalization infallible. The goal is to have two separate steps: a validation step that errors on invalid events, and an infallible normalization with the single responsibility of normalizing events. There are no functional changes in the PR.

## Design choice: a 2-step validation
Validating timestamps requires some timestamp normalization (e.g. clock drift correction), meaning the original goal of having a single validation accepting an immutable event is not easily achievable in a clean way. As a result, validation consists of two steps: a transaction validation step accepting an immutable event, and a timestamp validation step performing _minimal_ normalization. Alternatively, a small pre-validation normalization step was considered, discarded for generating too much confusion.

Further, we rely on the developer to be aware of calling the right validation function(s) and not extending the minimal normalization during validation. This has been documented in the functions.

## Future work

- Although a 2-step validation moves us closer to the goal, it's not great. This decision may be revisited as part of a larger processing pipeline changes.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -7,16 +7,11 @@ use std::collections::BTreeSet;
 
 use std::hash::{Hash, Hasher};
 use std::mem;
-use std::ops::Range;
 
-use chrono::{DateTime, Duration, Utc};
 use relay_base_schema::metrics::{
     can_be_valid_metric_name, DurationUnit, FractionUnit, MetricUnit,
 };
-use relay_common::time::UnixTimestamp;
-use relay_event_schema::processor::{
-    self, MaxChars, ProcessingAction, ProcessingResult, ProcessingState, Processor,
-};
+use relay_event_schema::processor::{self, MaxChars, ProcessingAction, ProcessingState, Processor};
 use relay_event_schema::protocol::{
     AsPair, Context, ContextInner, Contexts, DeviceClass, Event, EventType, Exception, Headers,
     IpAddr, Level, LogEntry, Measurement, Measurements, NelContext, Request, SpanAttribute,
@@ -27,12 +22,11 @@ use smallvec::SmallVec;
 
 use crate::normalize::request;
 use crate::span::tag_extraction::{self, extract_span_tags};
-use crate::timestamp::TimestampProcessor;
 use crate::utils::{self, MAX_DURATION_MOBILE_MS};
 use crate::{
     breakdowns, mechanism, schema, span, stacktrace, transactions, trimming, user_agent,
-    BreakdownsConfig, ClockDriftProcessor, DynamicMeasurementsConfig, GeoIpLookup,
-    PerformanceScoreConfig, RawUserAgentInfo, SpanDescriptionRule, TransactionNameConfig,
+    BreakdownsConfig, DynamicMeasurementsConfig, GeoIpLookup, PerformanceScoreConfig,
+    RawUserAgentInfo, SpanDescriptionRule, TransactionNameConfig,
 };
 
 /// Configuration for [`normalize_event`].
@@ -50,27 +44,6 @@ pub struct NormalizationConfig<'a> {
     /// information should the event payload contain no such data. If no client hints are present,
     /// normalization falls back to the user agent.
     pub user_agent: RawUserAgentInfo<&'a str>,
-
-    /// The time at which the event was received in this Relay.
-    ///
-    /// This timestamp is persisted into the event payload.
-    pub received_at: Option<DateTime<Utc>>,
-
-    /// The maximum amount of seconds an event can be dated in the past.
-    ///
-    /// If the event's timestamp is older, the received timestamp is assumed.
-    pub max_secs_in_past: Option<i64>,
-
-    /// The maximum amount of seconds an event can be predated into the future.
-    ///
-    /// If the event's timestamp lies further into the future, the received timestamp is assumed.
-    pub max_secs_in_future: Option<i64>,
-
-    /// Timestamp range in which a transaction must end.
-    ///
-    /// Transactions that finish outside of this range are considered invalid.
-    /// This check is skipped if no range is provided.
-    pub transaction_range: Option<Range<UnixTimestamp>>,
 
     /// The maximum length for names of custom measurements.
     ///
@@ -138,10 +111,6 @@ impl<'a> Default for NormalizationConfig<'a> {
         Self {
             client_ip: Default::default(),
             user_agent: Default::default(),
-            received_at: Default::default(),
-            max_secs_in_past: Default::default(),
-            max_secs_in_future: Default::default(),
-            transaction_range: Default::default(),
             max_name_and_unit_len: Default::default(),
             breakdowns_config: Default::default(),
             normalize_user_agent: Default::default(),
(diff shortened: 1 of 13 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -160,57 +129,41 @@ impl<'a> Default for NormalizationConfig<'a> {
...
     // TODO: Parts of this processor should probably be a filter so we
     // can revert some changes to ProcessingAction)
-    let mut transactions_processor = transactions::TransactionsProcessor::new(
-        config.transaction_name_config.clone(),
-        config.transaction_range.clone(),
-    );
-    transactions_processor.process_event(event, meta, ProcessingState::root())?;
+    let mut transactions_processor =
+        transactions::TransactionsProcessor::new(config.transaction_name_config.clone());
+    let _ = transactions_processor.process_event(event, meta, ProcessingState::root());
```
iker-barriocanal, line 156: The `TransactionsProcessor` is now infallible and doesn't reject events.

relay-event-normalization/src/event.rs
```diff
@@ -2498,62 +2391,6 @@ mod tests {
         "###);
     }
 
-    /// Test that timestamp normalization updates a transaction's timestamps to
-    /// be acceptable, when both timestamps are similarly stale.
-    #[test]
-    fn test_accept_recent_transactions_with_stale_timestamps() {
```
iker-barriocanal, line 2504: These tests have been moved to `validation.rs`.

relay-event-normalization/src/event.rs
```diff
@@ -160,57 +129,41 @@ impl<'a> Default for NormalizationConfig<'a> {
...
-        config.transaction_name_config.clone(),
-        config.transaction_range.clone(),
-    );
-    transactions_processor.process_event(event, meta, ProcessingState::root())?;
+    let mut transactions_processor =
+        transactions::TransactionsProcessor::new(config.transaction_name_config.clone());
+    let _ = transactions_processor.process_event(event, meta, ProcessingState::root());
 
     // Check for required and non-empty values
     let _ = schema::SchemaProcessor.process_event(event, meta, ProcessingState::root());
```
iker-barriocanal, line 159: `SchemaProcessor` only modifies the payload to add errors and remarks to `meta`, so running it after validation and timestamp normalization is fine.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -2509,19 +2510,4 @@ mod tests {
         }
         "###);
     }
-
-    #[test]
-    fn test_reject_stale_transaction() {
```
iker-barriocanal, line 2514: This test has been moved to `validation.rs`.

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -210,8 +138,10 @@ impl Processor for TransactionsProcessor<'_> {
 
         set_default_transaction_source(event);
         self.normalize_transaction_name(event);
-        self.validate_transaction(event)?;
-        end_all_spans(event)?;
+        end_all_spans(event);
+        if let Some(trace_context) = event.context_mut::<TraceContext>() {
+            trace_context.op.get_or_insert_with(|| "default".to_owned());
+        }
```
iker-barriocanal, line 144: This was the only non-validation step in the old `validate_transaction`, so we keep it here.

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -438,13 +334,8 @@ fn end_all_spans(event: &mut Event) -> ProcessingResult {
                 span.timestamp.set_value(event.timestamp.value().cloned());
                 span.status = Annotated::new(SpanStatus::DeadlineExceeded);
             }
-        } else {
-            return Err(ProcessingAction::InvalidTransaction(
-                "spans must be valid in transaction event",
-            ));
```
iker-barriocanal, line 444: This check is now part of the validation, and all spans at this point are valid.

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -521,25 +412,6 @@ mod tests {
         assert!(event.value().is_some());
     }
 
-    #[test]
-    fn test_discards_when_missing_timestamp() {
```
iker-barriocanal, line 525: These red tests have been moved to `validation.rs`.

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -25,19 +23,12 @@ pub struct TransactionNameConfig<'r> {
...
 pub struct TransactionsProcessor<'r> {
     name_config: TransactionNameConfig<'r>,
-    timestamp_range: Option<Range<UnixTimestamp>>,
 }
 
 impl<'r> TransactionsProcessor<'r> {
     /// Creates a new `TransactionsProcessor` instance.
-    pub fn new(
-        name_config: TransactionNameConfig<'r>,
-        timestamp_range: Option<Range<UnixTimestamp>>,
```
iker-barriocanal, line 35: `timestamp_range` is used during validation, so not useful in the processor anymore.

relay-event-normalization/src/validation.rs
iker-barriocanal, line 1: Validation steps have been moved from `TransactionsProcessor`.

relay-server/src/services/processor.rs
```diff
@@ -1095,7 +1100,11 @@ impl EnvelopeProcessorService {
             };
 
             metric!(timer(RelayTimers::EventProcessingLightNormalization), {
-                normalize_event(event, &config).map_err(|_| ProcessingError::InvalidTransaction)
+                validate_transaction(event, &tx_validation_config)?;
+                validate_event_timestamps(event, &event_validation_config)
+                    .map_err(|_| ProcessingError::InvalidTransaction)?;
+                normalize_event(event, &normalization_config);
+                Result::<(), ProcessingError>::Ok(())
```
iker-barriocanal, line 1108: `Ok(())` isn't sufficiently explicit.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -293,9 +288,12 @@ fn normalize(
         ProcessingState::root(),
     )?;
 
+    if let Some(span) = annotated_span.value() {
+        validate_span(span)?;
+    }
     process_value(
         annotated_span,
-        &mut TransactionsProcessor::new(Default::default(), Some(timestamp_range)),
```
iker-barriocanal, line 298: Calling the processor with a span results in only running `process_span`, so `timestamp_range` checks were noops before. The remaining timestamp checks are done in the `validate_span` call above.
cc @jjbayer

relay-server/src/services/processor/span/processing.rs
```diff
@@ -216,8 +217,6 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState) {
 struct NormalizeSpanConfig<'a> {
     /// The time at which the event was received in this Relay.
     pub received_at: DateTime<Utc>,
-    /// Allowed time range for transactions.
-    pub timestamp_range: std::ops::Range<UnixTimestamp>,
```
iker-barriocanal, line 220: With the removal of `timestamp_range` below, the property is not used anymore.

iker-barriocanal (commented):
Although the diff is big there are three type of changes, for an easier review:
- Move all the validation from normalization to a separate module.
- Run validation in addition to normalization.
- Small code adaptations: removing unnecessary `unwrap`s, moving tests...

I've also added some comments. Let me know if there's anything else I can do to ease the process.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -719,16 +722,21 @@ mod tests {
...
                 }
 
-                // Normalize first, to make sure that all things are correct as in the real pipeline:
-                let res = normalize_event(
+                // Validate and normalize first, to make sure that all things
+                // are correct as in the real pipeline:
+                let res = validate_transaction(&mut event, &TransactionValidationConfig::default());
+                assert!(res.is_ok());
+                let res = validate_event_timestamps(&mut event, &EventValidationConfig::default());
+                assert!(res.is_ok());
```
```suggestion
                // Validate and normalize first, to make sure that all things
                // are correct as in the real pipeline:
                let res = validate_transaction(&mut event, &TransactionValidationConfig::default());
                assert!(res.is_ok());
                let res = validate_event_timestamps(&mut event, &EventValidationConfig::default());
                assert!(res.is_ok());
```

This should not be necessary for the tests that use this macro. Ideally we would even call only `extract_tags` here.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -497,16 +500,21 @@ mod tests {
...
         let features = FeatureSet(BTreeSet::from([Feature::SpanMetricsExtraction]));
 
-        // Normalize first, to make sure that all things are correct as in the real pipeline:
+        // Validate and normalize first, to make sure that all things are correct as in the real pipeline:
+        let res = validate_transaction(&event, &TransactionValidationConfig::default());
+        assert!(res.is_ok());
+
+        let res = validate_event_timestamps(&mut event, &EventValidationConfig::default());
+        assert!(res.is_ok());
+
```
jjbayer: I would try to run the test without this, and remove validation if it passes. Ideally tests only run the code that they're actually trying to test. See [internal guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#d8c37053239e46fbb319a06428fac18d).

```suggestion
        // Validate and normalize first, to make sure that all things are correct as in the real pipeline:
        let res = validate_transaction(&event, &TransactionValidationConfig::default());
        assert!(res.is_ok());

        let res = validate_event_timestamps(&mut event, &EventValidationConfig::default());
        assert!(res.is_ok());

```

relay-server/src/services/processor.rs
jjbayer: Why?

relay-server/src/services/processor/span/processing.rs
jjbayer: Thanks for this!

jjbayer (approved):
Please see comments before merging, but overall this looks good to me!

relay-server/src/services/processor.rs
iker-barriocanal: The LSP doesn't recognize what type of error the `Result` is, and being explicit makes it happy.

relay-event-normalization/src/normalize/span/tag_extraction.rs
iker-barriocanal: Removed validation calls, but not replacing with `extract_tags` as I don't want to overload this PR.

PR #3005, opened 2024-01-26
chore(processor): Remove legacy processe state function

olksdr:
Remove the call to legacy `process_state` function and also the function itself. 

Running the new grouped processing pipeline for the last few weeks showed that all the events are grouped and processed. But we still have the error in `Ungrouped` branch, and will see if something is regressed. 

The changes also include:
* adding `Metrics` processing group
* using `ProcessingGroup::split_envelope` function in test and also assert that correct group is assigned for the processing
* update `ManagedEnvelope::standalone` function to also accept the `ProcessingGroup` as an argument
 

related to: #2862 https://github.com/getsentry/team-ingest/issues/253

#skip-changelog


jjbayer:
Seems like there's client reports going through the ungrouped group, at least in a test case: https://github.com/getsentry/relay/actions/runs/7664823704/job/20889701617?pr=3005#step:5:1842

olksdr:
@jjbayer yeah, tests are wrong and has to be fixed. They all just use `ungrouped` group by default.

olksdr:
Updated description, since this PR includes a little bit more changes. To make sure that tests are also properly make use of `ProcessingGroup`

jjbayer (approved):

PR #3006, opened 2024-01-26
ref(metrics): Rename tag-field to tags

TBS1996:
It's a list, so the name should reflect that.

#skip-changelog

### DIFF
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -3,8 +3,11 @@
 ## Unreleased
 
 - Add `allow_negative` to `BuiltinMeasurementKey`. Filter out negative BuiltinMeasurements if `allow_negative` is false. ([#2982](https://github.com/getsentry/relay/pull/2982))
+- Add ability to block metric tags matching a glob pattern. ([#2973](https://github.com/getsentry/relay/pull/2973))
 
 ## 0.8.44
+
+- Add ability to block metrics matching a glob pattern. ([#2954](https://github.com/getsentry/relay/pull/2954))
 
 ### Various fixes & improvements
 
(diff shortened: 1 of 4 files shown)


py/CHANGELOG.md
```diff
@@ -3,9 +3,12 @@
 ## Unreleased
 
 - Add `allow_negative` to `BuiltinMeasurementKey`. Filter out negative BuiltinMeasurements if `allow_negative` is false. ([#2982](https://github.com/getsentry/relay/pull/2982))
+- Add ability to block metric tags matching a glob pattern. ([#2973](https://github.com/getsentry/relay/pull/2973))
 
 ## 0.8.44
 
+- Add ability to block metrics matching a glob pattern. ([#2954](https://github.com/getsentry/relay/pull/2954))
```
TBS1996, line 10: so this was added together with the 'various fixes and improvements', but this is more of a new feature. Not sure how to best retroactively add it

relay-dynamic-config/src/metrics.rs
```diff
@@ -43,7 +43,7 @@ pub struct TagBlock {
     /// Name of metric of which we want to remove certain tags.
     pub name: GlobPatterns,
```
jjbayer, line 44: Shouldn't this also be `"names"` then?

py/CHANGELOG.md
jjbayer: You did it right IMO. "Various fixes & improvements" seems to be a bot thing.

jjbayer (approved):

relay-dynamic-config/src/metrics.rs
TBS1996: atm, we think of it as it should just be a single name pattern there, it's just the type we use allows for multiple. So, depends on if we should name after what we intend to do, or name after what we are able to do

Issue #2992, opened 2024-01-23
Emit accepted outcomes for span metrics


jjbayer:
Just like for transactions, we need accepted outcomes for span metrics. Multiple solutions:

1. Create a new usage span metric and emit outcomes from the [billing metrics consumer](https://github.com/getsentry/sentry/blob/0b4b4a5278c69805f945bf6bf58d16f2f9f153a4/src/sentry/ingest/billing_metrics_consumer.py). This would be consistent with what we do for transactions.
1. Use the billing metrics consumer, but base the counts on an existing metric (e.g. exclusive_time). We moved away from this approach for transactions.
1. Log the "accepted" outcome directly from relay. This is nice & simple but has the downside that metrics could still be dropped at a later point. However, introducing a special "usage" metric has the same downside: What if the usage metric survives, but the other metrics are cardinality limited?

jjbayer:
Option 1 seems like the safe choice:  If for example a rate limit kicks in for all metrics of a project, Option 1 would reflect this in outcomes, while with Option 3, we would still happily emit accepted outcomes.

PR #3007, opened 2024-01-26
feat(spans): Span usage metric

### LINKED ISSUE #3009: Global Config options serialize defaults
When at least one global config option is set, all options will be serialized to downstream relays through the proxied project config endpoint.

Easily verifiable with the `test_get_global_config` test, adding a single option to `global_config` will fail the test because all defaults are serialized.

CoS: `test_get_global_config` passes with the following global config:

```py
    mini_sentry.global_config = {
        "measurements": {
            "builtinMeasurements": [
                {"name": "app_start_cold", "unit": "millisecond"},
                {"name": "app_start_warm", "unit": "millisecond"},
            ],
            "maxCustomMeasurements": 10,
        },
       "options": {"foo": "bar"}
    }
```

jjbayer:
Emit a usage metric for every span that gets _any_ metric extracted. As for transactions, this usage metric is then translated into `"Accepted"` outcomes in the [billing metrics consumer](https://github.com/getsentry/sentry/pull/63831).

ref: https://github.com/getsentry/relay/issues/2992

Also fixes https://github.com/getsentry/relay/issues/3009.


relay-dynamic-config/src/defaults.rs
```diff
@@ -105,6 +105,13 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
         & RuleCondition::eq("span.description", APP_START_ROOT_SPAN_DESCRIPTIONS);
 
     [
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "c:spans/usage@none".into(),
+            field: None,
+            condition: Some(duration_condition.clone()), // same as for count_per_op
```
jjbayer: @phacops this condition is open for debate: Does it make sense to count a span as "used" (and thereby create an accepted outcome for it) even if it's not used in any current module? Or should we rather use the condition from the `exclusive_time` metric?

iker-barriocanal (approved):
LGTM. Should we add a test case? I think it'd be interesting to ensure that a span gets exactly one usage metric after going through a couple of relays.

Dav1dde:
Not necessarily related to your PR, but we could also change:

```rs
    pub options: Option<Options>,
```

to:

```rs
    #[serde(default, skip_serializing_if = "is_default")]
    pub options: Options,
```

Makes using the options easier and avoid serializing an empty `{}` when all options are default.

relay-dynamic-config/src/global.rs
```diff
@@ -111,7 +134,8 @@ mod tests {
                 max_custom_measurements: 5,
             }),
             options: Some(Options {
-                other: HashMap::from([("relay.unknown".to_owned(), Value::Bool(true))]),
+                unsampled_profiles_enabled: true,
+                //other: HashMap::from([("relay.unknown".to_owned(), Value::Bool(true))]),
```
olksdr: should this commented out line be removed?

relay-dynamic-config/src/defaults.rs
phacops: To be accurate, we should not count a span as `Accepted` if the metric is internal (like `count_per_op`). So, I think using the `exclusive_time` metric's conditions is the right move here and any condition since it's the broader condition of all the metrics we use in the UI.

If we ever have a metric used in the UI with a broader condition than this, we'd have to update `usage` metric's condition as well.

jjbayer: Sounds good to me! Changed to `exclusive_time_light`, it's more lenient than `exclusive_time`.

jjbayer:
> LGTM. Should we add a test case? I think it'd be interesting to ensure that a span gets exactly one usage metric after going through a couple of relays.

@iker-barriocanal absolutely! I think the only-extract-metrics-once part has been sufficiently tested for other metrics, but I added a basic test for the metric itself.

> Makes using the options easier and avoid serializing an empty {} when all options are default.

@Dav1dde done!

Dav1dde (approved):
Nice, thanks for doing that cleanup 👍

relay-dynamic-config/src/defaults.rs
```diff
@@ -104,12 +104,17 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
     let app_start_condition = RuleCondition::glob("span.op", "app.start.*")
         & RuleCondition::eq("span.description", APP_START_ROOT_SPAN_DESCRIPTIONS);
 
+    // `exclusive_time_light` is the metric with the most lenient condition.
+    let exclusive_time_light_condition =
```
iker-barriocanal, line 108: nit: are these the exact conditions for the exclusive time metric? If not, these are the conditions for the usage metric, and I'd suggest renaming the variable in this case.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -1244,6 +1243,50 @@ mod tests {
...
+        );
+
+        let usage_metrics = metrics
+            .into_iter()
+            .filter(|b| b.name == "c:spans/usage@none")
+            .collect::<Vec<_>>();
+
+        assert_eq!(usage_metrics.len(), 7);
+        for m in usage_metrics {
+            assert!(m.tags.is_empty());
```
iker-barriocanal, line 1287: Should this metric have a tag for the project?

iker-barriocanal (approved):
Thanks for adding a test!

relay-dynamic-config/src/defaults.rs
jjbayer: They are the exact conditions for the `exclusive_time_light` metric.

relay-server/src/metrics_extraction/event.rs
jjbayer: No, the project ID itself is not part of the bucket.

PR #3008, opened 2024-01-26
feat(cardinality): Add cardinality kill switch to global config

Dav1dde:
Sentry PR: https://github.com/getsentry/sentry/pull/63942

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -83,7 +83,7 @@ pub enum CardinalityLimiterMode {
     /// Cardinality limiter is enabled.
     #[default]
     Enabled,
-    /// Cardinality limiter is enabled but no actions are taken.
+    /// Cardinality limiter is enabled but cardinality limits are not enforced.
     Passive,
     /// Cardinality limiter is disabled.
     Disabled,
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -38,6 +38,14 @@ impl GlobalConfig {
             Ok(None)
         }
     }
+
+    /// Returns the [`Options::cardinality_limiter_mode`] option.
+    pub fn cardinality_limiter_mode(&self) -> CardinalityLimiterMode {
+        self.options
+            .as_ref()
+            .map(|o| o.cardinality_limiter_mode)
+            .unwrap_or_default()
+    }
 }
 
 /// All options passed down from Sentry to Relay.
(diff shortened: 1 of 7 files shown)


relay-server/src/services/processor.rs
```diff
@@ -1811,6 +1811,15 @@ impl EnvelopeProcessorService {
         buckets: Vec<Bucket>,
         mode: ExtractionMode,
     ) -> Vec<Bucket> {
+        use relay_dynamic_config::CardinalityLimiterMode;
```
olksdr: we have something like imports only for processing , see 
https://github.com/getsentry/relay/blob/ef5111abf8273b0a3468d26a56867a7e4d0ff645/relay-server/src/services/processor.rs#L44
should this `use`  also be moved in there?

olksdr (approved):
looks good to me

relay-server/src/services/processor.rs
Dav1dde: I'll move it 👍

relay-dynamic-config/src/global.rs
```diff
@@ -58,11 +66,29 @@ pub struct Options {
...
 
+/// Kill switch for controlling the cardinality limiter.
+#[derive(Default, Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub enum CardinalityLimiterMode {
+    /// Cardinality limiter is enabled.
+    #[default]
+    Enabled,
+    /// Cardinality limiter is enabled but no actions are taken.
```
iker-barriocanal: What does "no actions are taken" mean? Could we elaborate that a bit?

Dav1dde: Rephrased to:

> Cardinality limiter is enabled but cardinality limits are not enforced.

Is that better to understand?

iker-barriocanal (approved):

relay-dynamic-config/src/global.rs
iker-barriocanal: Yep, thanks!

Dav1dde:
Don't merge this yet, for some reason I am seeing an empty string being sent as the option, this makes the global config fail to deserialize.

Dav1dde:
We serialize the option from an empty string as well, dealing with the broken values. For the future Sentry will send the correct value in the first place.

relay-dynamic-config/src/global.rs
```diff
@@ -58,11 +66,32 @@ pub struct Options {
...
+    /// Cardinality limiter is enabled.
+    #[default]
+    // De-serialize from the empty string, because the option was added to
+    // Sentry incorrectly which makes Sentry send the empty string as a default.
+    #[serde(alias = "")]
+    Enabled,
+    /// Cardinality limiter is enabled but cardinality limits are not enforced.
+    Passive,
+    /// Cardinality limiter is disabled.
+    Disabled,
```
jjbayer, line 92: Is there a chance this enum will be extended in the future? If so, we should add an `Unknown(String)` option, such that external Relays can still parse the config.

PR #3013, opened 2024-01-29
fix(spans): Reject spans with invalid timestamps

iker-barriocanal:
https://github.com/getsentry/relay/commit/90fa749f1c4af857fb0f0b78e517d0f9fed4b32a removed the `timestamp_range` from span normalization as it was not used. This PR introduces it back and validates a span's timestamps are valid. A span has valid timestamps when the following conditions are met:
- `start` and `end` timestamps exist.
- `start` <= `end` timestamps.
- `end` timestamp is within the valid timestamp range. This range is defined in relay's config, in the secondary aggregator.


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Internal**:
+
+- Drop spans ending outside the valid timestamp range. ([#3013](https://github.com/getsentry/relay/pull/3013))
```
iker-barriocanal, line 14: The change only applies to Relays and not librelay, so not updating py/Changelog.

relay-event-normalization/src/validation.rs
```diff
@@ -44,7 +44,9 @@ pub fn validate_transaction(
 
     validate_transaction_timestamps(event, config)?;
     validate_trace_context(event)?;
-    validate_spans(event)?;
+    // Transactions with spans with invalid timestamps are not expected to be
+    // rejected.
```
olksdr: This statement is not really true right now, since in the `validate_span` function we can return an error if either `start` or `end` timestamps are absent (which is kind of invalid too), so we will return the error from `validate_transaction` as well and reject the `transaction` as well.

It would be great to change this comment here to reflect what condition wouldn't actually reject the transaction.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -217,6 +218,8 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState) {
 struct NormalizeSpanConfig<'a> {
     /// The time at which the event was received in this Relay.
     pub received_at: DateTime<Utc>,
+    /// Allowed time range for spans.
+    pub timestmap_range: std::ops::Range<UnixTimestamp>,
```
```suggestion
    pub timestamp_range: std::ops::Range<UnixTimestamp>,
```

tests/integration/test_store.py
```diff
@@ -1881,3 +1881,74 @@ def test_span_extraction_with_ddm_missing_values(
     }
 
     spans_consumer.assert_empty()
+
+
+def test_span_reject_invalid_timestamps(
```
jjbayer, line 1886: Do we need an integration test for this? Could it be a unit test instead?

iker-barriocanal: A unit test is not as solid. The timestamp range on spans is only enforced in stand-alone spans and not on spans processed as part of transactions, meaning testing a single component is not enough and we should test the span processing pipeline instead.

relay-event-normalization/src/validation.rs
iker-barriocanal: Thanks for pointing it out! I reworded it in [b9b8b70](https://github.com/getsentry/relay/pull/3013/commits/b9b8b70620e9d5f55331e3c87c108f694ef64d17), let me know if this is still unclear.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -219,7 +219,7 @@ struct NormalizeSpanConfig<'a> {
     /// The time at which the event was received in this Relay.
     pub received_at: DateTime<Utc>,
     /// Allowed time range for spans.
-    pub timestmap_range: std::ops::Range<UnixTimestamp>,
+    pub timestamap_range: std::ops::Range<UnixTimestamp>,
```
jjbayer: LOL
```suggestion
    pub timestamp_range: std::ops::Range<UnixTimestamp>,
```

jjbayer (approved):
See rename typo, rest LGTM.

relay-server/src/services/processor/span/processing.rs
iker-barriocanal: lol I can't type, thanks for spotting it 👍

relay-event-normalization/src/validation.rs
```diff
@@ -44,7 +44,9 @@ pub fn validate_transaction(
 
     validate_transaction_timestamps(event, config)?;
     validate_trace_context(event)?;
-    validate_spans(event)?;
+    // There are no timestamp range requirements for span timestamps to reject
+    // the parent transaction.
```
olksdr: I would put this even more explicit. What do you think about this ? 
```suggestion
    // There are no timestamp range requirements for span timestamps.
    // Transaction will be rejected only if either end or start timestamp is missing. 
  
```

PR #3014, opened 2024-01-29
ref(normalization): Move LegacyProcessor to normalization

iker-barriocanal:
As we move towards the goal of removing `StoreProcessor` and merging its functionality into the new normalization, this PR moves the first step: `LegacyProcessor`. This is not a breaking change, as `normalize_event` always runs before `StoreProcessor`, including in [librelay](https://github.com/getsentry/relay/blob/0e7b0d022c49f20cd0c48c676ea7a4a86a1e590d/relay-cabi/src/processing.rs#L152).

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -24,7 +24,7 @@ use crate::normalize::request;
 use crate::span::tag_extraction::{self, extract_span_tags};
 use crate::utils::{self, MAX_DURATION_MOBILE_MS};
 use crate::{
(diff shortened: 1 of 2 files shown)


jjbayer (approved):
@iker-barriocanal Remind me, is it the goal that all normalization will run _before_ dynamic sampling eventually?

iker-barriocanal:
@jjbayer that will happen, but it's not the main goal. Having greater control over normalization will also allow defining when normalization runs; for example, only running it once in the ingestion pipeline instead of multiple times (e.g. run it in PoPs but not in processing relays).

PR #3015, opened 2024-01-29
ref(normalization): Move SchemaProcessor to normalize_event

iker-barriocanal:
Moves `SchemaProcessor` from `normalize` to `normalize_event`.

The goal of `normalize` is to normalize specific fields in the payload. The `SchemaProcessor` is a more generic step, and thus better suited for a level of abstraction above, in `normalize_event`. Placing correctly these steps is important as we merge `StoreProcessor`; providing a clear structure allows having greater control over the normalization steps and eases the understanding and updating of normalization as a whole, as agreed with the team in the [interface design (internal)](https://www.notion.so/sentry/Normalization-interface-design-5cf95c2bc9b94767bb7b64ff14528c47?pvs=4#8342024cd7364961ac56c19e02b04bfe).

#skip-changelog


jjbayer:
> Placing correctly these steps is important as we merge StoreProcessor.

@iker-barriocanal why is is important? Might make sense to put the answer to that question in the PR description.

iker-barriocanal:
@jjbayer good question! I expanded the PR description a bit. Let me know if it's still unclear or there are further questions.

relay-event-normalization/src/event.rs
```diff
@@ -144,6 +144,9 @@ pub fn normalize_event(event: &mut Annotated<Event>, config: &NormalizationConfi
     let _ = legacy::LegacyProcessor.process_event(event, meta, ProcessingState::root());
 
     if !is_renormalize {
+        // Check for required and non-empty values
+        let _ = schema::SchemaProcessor.process_event(event, meta, ProcessingState::root());
```
olksdr, line 148: And this is fine to run the schema processor before transactions processor which wasn't the case before this change?

olksdr (approved):

relay-event-normalization/src/event.rs
iker-barriocanal: The schema processor doesn't perform any normalization, all it does is add errors to the payload's meta when required attributes are missing or have invalid values.

PR #3018, opened 2024-01-30
ref(cardinality): Clean old values from cardinality cache

Dav1dde:
Old values may retain in cache when there are no longer arriving updates for these scopes. Periodically remove expired values from the cache.

#skip-changelog

### DIFF
--- a/relay-cardinality/src/redis/cache.rs
+++ b/relay-cardinality/src/redis/cache.rs
@@ -9,9 +9,6 @@ use crate::redis::QuotaScoping;
 use crate::statsd::{CardinalityLimiterCounters, CardinalityLimiterTimers};
 use crate::window::Slot;
 
-/// Interval in which the cache is checked for expired values.
-const CACHE_VACUUM_INTERVAL: Duration = Duration::from_secs(180);
-
 /// Cached outcome, wether the item can be accepted, rejected or the cache has no information about
 /// this hash.
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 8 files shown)


relay-cardinality/src/redis/cache.rs
```diff
@@ -1,11 +1,17 @@
...
 
 use relay_common::time::UnixTimestamp;
+use relay_statsd::metric;
 
 use crate::redis::QuotaScoping;
+use crate::statsd::{CardinalityLimiterCounters, CardinalityLimiterTimers};
 use crate::window::Slot;
 
+/// Interval in which the cache is checked for expired values.
+const CACHE_VACUUM_INTERVAL: Duration = Duration::from_secs(180);
```
jjbayer: Should we make this configurable right away? Seems like something that belongs in the static config.

relay-cardinality/src/redis/cache.rs
```diff
@@ -129,9 +137,37 @@ impl<'a> CacheUpdate<'a> {
...
+#[derive(Debug)]
 struct Inner {
     cache: hashbrown::HashMap<QuotaScoping, ScopedCache>,
+    last_vacuum: UnixTimestamp,
+}
+
+impl Inner {
+    fn vacuum(&mut self, ts: UnixTimestamp) {
+        // Debounce the vacuuming.
+        let secs_since_last_vacuum = ts.as_secs().saturating_sub(self.last_vacuum.as_secs());
```
jjbayer, line 156: Why are we using a virtual "now" time here? Shouldn't the debounce depend on the actual current time?

nit: If `timestamp` is just a buffered now time, should we rename it to `now`, or `current_time`?

jjbayer (approved):
See comment about config, rest LGTM!

relay-cardinality/src/redis/cache.rs
Dav1dde: Effectively it is always `now`, I was considering taking the `abs(ts - last_vacuum)` here, but since we only vacuum values which are in the past, based on the passed timestamp, I think it makes sense to only consider "the future".

Additionally using the passed in param also makes it a lot easier to test.

PR #3019, opened 2024-01-30
fix(metrics): Do not report dropped buckets for disabled project

jjbayer:
We started reporting errors for buckets dropped after aggregation in #2989. For disabled projects, we can drop silently though.

Fixes [POP-RELAY-2ZG](https://sentry.my.sentry.io/organizations/sentry/issues/541271/), [RELAY-2NQV](https://sentry.my.sentry.io/organizations/sentry/issues/541475/).

#skip-changelog

### DIFF
--- a/relay-server/src/services/project.rs
+++ b/relay-server/src/services/project.rs
@@ -1121,6 +1121,10 @@ impl Project {
         })
     }
 
+    /// Drops metrics buckets if they are not allowed for this project.
+    ///
+    /// Reasons for dropping can be rate limits or a disabled project.
+    /// Returns `Some` if metrics are currently allowed.
     pub fn check_buckets(
         &mut self,
         outcome_aggregator: Addr<TrackOutcome>,
@@ -1152,6 +1156,9 @@ impl Project {
         None
     }
 
+    /// Unit-testable helper function for [`Self::check_buckets`].
+    ///
+    /// Returns [`CheckedBuckets::Ok`] if metrics are currently allowed, or the reject reason otherwise.
     fn check_buckets_inner(
         &mut self,
         outcome_aggregator: Addr<TrackOutcome>,
@@ -1202,12 +1209,20 @@ impl Project {
     }
 }
 
+/// Result of a bucket check.
 #[derive(Debug)]
 enum CheckedBuckets {
+    /// Metrics are allowed. Contains the metrics plus scoping information.
     Ok(Scoping, ProjectMetrics),
+    /// The project has expired while the bucket was in aggregation.
+    ///
+    /// This should not happen as long as the aggregation time is shorter than the refresh time.
     ProjectExpired(usize),
+    /// The project is disabled or invalid.
     ProjectDisabled(usize),
+    /// The project has no project ID, even though it is not disabled. This should never happen.
     NoScoping(usize),
+    /// The buckets were rate limited.
     RateLimited(usize),
 }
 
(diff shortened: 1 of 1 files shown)


relay-server/src/services/project.rs
```diff
@@ -1126,21 +1126,51 @@ impl Project {
...
+                    CheckBucketsError::RateLimited(len) => {
+                        relay_log::debug!("dropping {len} buckets due to rate limit");
+                    }
+                };
+                e
+            })
+            .ok()
+    }
+
+    fn check_buckets_inner(
```
jjbayer, line 1162: I created this helper method to make the result unit-testable (specific error instead of `None`).

relay-server/src/services/project.rs
```diff
@@ -1126,21 +1126,51 @@ impl Project {
...
         let Some(project_state) = self.valid_state() else {
-            relay_log::error!(
-                tags.project_key = self.project_key.as_str(),
-                "there is no project state: dropping {len} buckets"
-            );
-            return None;
+            return Err(CheckBucketsError::ProjectExpired(len));
         };
 
+        if project_state.invalid() || project_state.disabled() {
```
jjbayer, line 1173: There are helper functions like `metrics_allowed()` and `check_disabled()`, but they would be redundant here because they both do an expiry check, which in this case already happened on line 1164.

relay-server/src/services/project.rs
```diff
@@ -1126,21 +1126,51 @@ impl Project {
         outcome_aggregator: Addr<TrackOutcome>,
         buckets: Vec<Bucket>,
     ) -> Option<(Scoping, ProjectMetrics)> {
+        self.check_buckets_inner(outcome_aggregator, buckets)
```
Dav1dde: I'd probably write this with a single match matching the result

```rs
match self.check_buckets_inner(outcome_aggregator, buckets) {
    Ok(r) => return Some(r),
    Err(CheckBucketsError::ProjectExpired(len)) => relay_log::error!(...),
    Err(CheckBucketsError::ProjectDisabled(len)) => ...,
    ...
}

None
```

Dav1dde (approved):

relay-server/src/services/project.rs
```diff
@@ -1164,18 +1192,26 @@ impl Project {
...
             buckets,
             project_state,
         };
 
-        Some((scoping, project_metrics))
+        Ok((scoping, project_metrics))
     }
 }
 
+#[derive(Debug)]
```
olksdr, line 1213: should this be actually a real error? 

You could derive `thiserror::Error` and add the proper display error for logging and wouldn't have to `map_err` but just log it once it's the `Err(err)`

Something like maybe ?
```rust
enum CheckBucketsError {
    #[error("there is no project state: dropping {0} buckets")]
    ProjectExpired(usize),
...
```

relay-server/src/services/project.rs
olksdr: I like the suggestion from David as well.

relay-server/src/services/project.rs
jjbayer: I like the idea and I tried it out, but it led me in the other direction: `ProjectDisabled` is not a real error, but I still want it as a separate variant to make the function unit-testable. So I went for a flat enum as return value instead, let me know what you think.

relay-server/src/services/project.rs
jjbayer: Done!

relay-server/src/services/project.rs
olksdr: Yeah, that also works.
 
Could you, please, just add few lines of docs here?

olksdr (approved):

relay-server/src/services/project.rs
jjbayer: Done!

PR #3020, opened 2024-01-30
ref(cardinality): Offset sliding window

Dav1dde:
Offsets the sliding window based on the org id of the scope. This should introduce a proper distribution of Redis calls and cache resets across the entire granule of the sliding window.

#skip-changelog


Dav1dde:
I need to double check this, I don't think this implementation works. I'll also see if I can come up with a nice test for this.

jjbayer:
Without having looked at the implementation, I have the concern that offsetting the sliding window will misalign cardinality for larger storage buckets.

```
shifted cardinality limiter:  -|---|---|---|---|---|---|---|---|---
storage:                      |-----------------|-----------------|
```

Dav1dde:
They will definitely be more likely to be misaligned, but the misalignment is already happening/possible since we cardinality limit based on now but storage persists with the bucket timestamp.

It also should be not relevant for 10s granularity and 1h/1d granularities.

relay-cardinality/src/redis/limiter.rs
```diff
@@ -210,9 +210,32 @@ impl QuotaScoping {
         self.namespace.is_none() || self.namespace == Some(entry.namespace)
     }
 
+    /// Returns the currently active slot.
+    pub fn active_slot(&self, timestamp: UnixTimestamp) -> Slot {
```
jjbayer: Are there any other places in the code that access `window.active_slot()` or `window.iter()` with a non-shifted timestamp? That would introduce a nasty bug. I'm almost considering a `ShiftedTimestamp` newtype.

jjbayer (approved):

relay-cardinality/src/redis/limiter.rs
Dav1dde: I made the `window` field on the `RedisQuota` no longer public to prevent this.

Unfortunatly this does not work  for the `redis/limiter.rs` file, since it is the same module, I considered moving it to its own module to have "working" visibility modifiers.

I am not opposed to adding another new-type though.

jjbayer: > I considered moving it to its own module to have "working" visibility modifiers.

This makes sense to me!

PR #3027, opened 2024-01-31
feat(app-start): Tag spans with start type

narsaynorath:
We want to tag all spans during an app start with the start type. This is because on Android, the span ops don't reflect the start type but we still want to filter spans by their type.

Start type can be derived from the transaction if there is a measurement associated, or in the future we will add `app_start_type` to span data and that should take precedence (see the GH issue that mentions this PR for further context).


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -143,6 +146,7 @@ pub(crate) fn extract_span_tags(event: &mut Event, config: &Config) {
     let is_mobile = shared_tags
         .get(&SpanTagKey::Mobile)
         .is_some_and(|v| v.as_str() == "true");
+    let start_type = get_event_start_type(event.clone(), is_mobile);
```
Dav1dde: Would move this directly above `ttid` or `ttfd` (after the spans check right below).

```suggestion
    let start_type = is_mobile.then(|| get_event_start_type(&event).flatten();
```

It's important that we do not clone the event, this is very costly. I would also not pass the `is_mobile` flag into the function, that is just hidden control flow the caller can take care of.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -674,6 +692,23 @@ fn span_op_to_category(op: &str) -> Option<&str> {
     }
 }
 
+/// Reads the event measurements to determine the start type of the event
+fn get_event_start_type(event: Event, is_mobile: bool) -> Option<String> {
```
Dav1dde: As noted above:

```suggestion
fn get_event_start_type(event: &Event) -> Option<String> {
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Add `raw_domain` tag to indexed spans. ([#2975](https://github.com/getsentry/relay/pull/2975))
 - Obtain `span.domain` field from the span data's `url.scheme` and `server.address` properties when applicable. ([#2975](https://github.com/getsentry/relay/pull/2975))
 - Do not truncate simplified SQL expressions. ([#3003](https://github.com/getsentry/relay/pull/3003))
+- Add `span.app_start_type` as self time tag for app start spans ([#3027](https://github.com/getsentry/relay/pull/3027))
```
Dav1dde: Tiny nit:

```suggestion
- Add `span.app_start_type` as self time tag for app start spans. ([#3027](https://github.com/getsentry/relay/pull/3027))
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -1,4 +1,4 @@
-//! Logic for persisting items into `span.data` fields.
+//! Logic for persisting items into `span.sentry_tags` and `span.measurements` fields.
```
jjbayer, line 1: Thanks!

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -60,6 +60,8 @@ pub enum SpanTagKey {
     FileExtension,
     /// Span started on main thread.
     MainTread,
+    /// The start type of the application when the span occurred
```
```suggestion
    /// The start type of the application when the span occurred.
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -674,6 +692,23 @@ fn span_op_to_category(op: &str) -> Option<&str> {
...
+/// Reads the event measurements to determine the start type of the event
+fn get_event_start_type(event: Event, is_mobile: bool) -> Option<String> {
+    // Only collect start type for mobile spans
+    if !is_mobile {
+        return None;
+    }
+
+    // Check the measurements on the event to determine what kind of start type the event is
+    if event.measurement("app_start_cold").is_some() {
+        Some("app.start.cold".to_owned())
```
jjbayer: nit: Up to you, but the `app.start.` prefix seems redundant to me. Why not just `"cold"` and `"warm"`?

relay-dynamic-config/src/defaults.rs
```diff
@@ -176,6 +176,9 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
                 Tag::with_key("span.main_thread")
                     .from_field("span.sentry_tags.main_thread")
                     .when(is_mobile.clone()),
+                Tag::with_key("span.app_start_type")
```
jjbayer: The `span.` prefix on other fields is historic, I think we can do without.

```suggestion
                Tag::with_key("app_start_type")
```

jjbayer: Please add the tag key to https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py before merging this PR, to prevent creating a copy of this tag key in the indexer db for every org.

relay-dynamic-config/src/defaults.rs
```diff
@@ -176,6 +176,9 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
...
+                    .from_field("span.sentry_tags.app_start_type")
+                    .when(is_mobile.clone()),
```
jjbayer: Do we need this tag on both `exclusive_time` and `exclusive_time_light` (i.e. both when we query with and without `transaction`)? Do we need it on any other metrics (e.g. `duration`)?

relay-dynamic-config/src/defaults.rs
jjbayer: I verified that `app_start_warm` does not exist yet in the indexer db:

```sql
select * from sentry_stringindexer where string = 'app_start_type'
select * from sentry_perfstringindexer where string = 'app_start_type'
```

cc @Dav1dde

relay-event-normalization/src/normalize/span/tag_extraction.rs
narsaynorath: @Dav1dde so I ended up doing this because I was getting the following error:
```rust
error[E0502]: cannot borrow `event` as immutable because it is also borrowed as mutable
   --> relay-event-normalization/src/normalize/span/tag_extraction.rs:154:37
    |
150 |     let Some(spans) = event.spans.value_mut() else {
    |                       ----------- mutable borrow occurs here
...
154 |     let start_type = is_mobile.then(|| get_event_start_type(&event)).flatten();
    |                                     ^^                       ----- second borrow occurs due to use of `event` in closure
    |                                     |
    |                                     immutable borrow occurs here
155 |     let ttid = timestamp_by_op(spans, "ui.load.initial_display");
    |                                ----- mutable borrow later used here
```

Is there a typical method for getting around this? I think I could store the measurements in a variable and pass that down into `get_event_start_type` but I like that `event` has a `measurements` method that does some of the nice retrieval for me. I'm not so familiar with Rust so please feel free to point me towards a more idiomatic approach

relay-event-normalization/src/normalize/span/tag_extraction.rs
narsaynorath: Good point, I replaced the values with "cold" and "warm". I see this is also consistent with what's being sent in app context for iOS

relay-dynamic-config/src/defaults.rs
narsaynorath: Could you actually clarify the purpose of `exclusive_time_light`? I think we might be able to drop it from there

I don't think we need it on `duration` though because that metric only applies to the "root" app start spans and in those cases we use a different filter. We only need it to filter a table that sorts by self time

jjbayer: `exclusive_time_light` is for when you query span metrics without the `transaction` tag.

narsaynorath: Thanks! Then I'll only store it for `exclusive_time`, this tag is only used with the `transaction` tag for now

relay-event-normalization/src/normalize/span/tag_extraction.rs
Dav1dde: Ah sorry, I missed that.

You're correct, we cannot do this, because a part of the event (spans field) is already mutably borrowed and then you pass the entire event to a function, that's why that is not allowed.

The Rust way of fixing this would be just passing the measurements of the event to the function, but in this case since we use the `event.measurement()` method this is not worth the trouble -> there is a good reason to keep it where it is 👍

relay-dynamic-config/src/defaults.rs
narsaynorath: PR is up for adding it here: https://github.com/getsentry/sentry/pull/64356

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Add `raw_domain` tag to indexed spans. ([#2975](https://github.com/getsentry/relay/pull/2975))
 - Obtain `span.domain` field from the span data's `url.scheme` and `server.address` properties when applicable. ([#2975](https://github.com/getsentry/relay/pull/2975))
 - Do not truncate simplified SQL expressions. ([#3003](https://github.com/getsentry/relay/pull/3003))
+- Add `span.app_start_type` as self time tag for app start spans. ([#3027](https://github.com/getsentry/relay/pull/3027))
```
```suggestion
- Add `app_start_type` as self time tag for app start spans. ([#3027](https://github.com/getsentry/relay/pull/3027))
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -156,7 +160,7 @@ pub(crate) fn extract_span_tags(event: &mut Event, config: &Config) {
             continue;
         };
 
-        let tags = extract_tags(span, config, ttid, ttfd, is_mobile);
+        let tags = extract_tags(span, config, ttid, ttfd, is_mobile, start_type.clone());
```
```suggestion
        let tags = extract_tags(span, config, ttid, ttfd, is_mobile, start_type);
```

see suggestions below.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -243,6 +247,7 @@ pub fn extract_tags(
     initial_display: Option<Timestamp>,
     full_display: Option<Timestamp>,
     is_mobile: bool,
+    start_type: Option<String>,
```
```suggestion
    start_type: Option<&str>,
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -674,6 +692,18 @@ fn span_op_to_category(op: &str) -> Option<&str> {
     }
 }
 
+/// Reads the event measurements to determine the start type of the event
+fn get_event_start_type(event: &Event) -> Option<String> {
```
```suggestion
fn get_event_start_type(event: &Event) -> Option<&'static str> {
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -674,6 +692,18 @@ fn span_op_to_category(op: &str) -> Option<&str> {
     }
 }
 
+/// Reads the event measurements to determine the start type of the event
```
```suggestion
/// Reads the event measurements to determine the start type of the event.
```

jjbayer (approved):
Please see comments before merging, and merge https://github.com/getsentry/sentry/pull/64356 before this one.

PR #3031, opened 2024-01-31
ref: python3.10+

asottile-sentry:
#skip-changelog


.github/workflows/build_library.yml
```diff
@@ -69,7 +69,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: '3.10'
```
iker-barriocanal, line 72: Is the `'` to make it a string and not be interpreted as `3.1`?

iker-barriocanal (approved):
🚀

.github/workflows/build_library.yml
asottile-sentry: yep because yaml sucks

PR #3032, opened 2024-01-31
ref(normalization): Move span exclusive time computation to normalization

iker-barriocanal:
This PR moves span normalization from `StoreNormalizeProcessor` to `normalize`, bringing us closer to having a single normalization step.

## Decisions

Normalization for the span attribute `ExclusiveTime` is run directly. This attribute is heavily required, the only existing one, and there are no plans to add more. This also simplifies the normalization config and logic.

The `normalize_spans` feature flag currently controls whether to run partial span normalization before store normalization (besides other non-normalization things). As we move all normalization to a single place and run it completely, there's no need to have a flag for this anymore and we can get rid of it.

#skip-changelog


relay-event-normalization/src/normalize/span/attributes.rs
```diff
@@ -220,51 +211,6 @@ mod tests {
         exclusive_times
     }
 
-    #[test]
-    fn test_skip_exclusive_time() {
```
iker-barriocanal, line 224: We don't want to skip exclusive time, so this test is stale.

relay-cabi/src/processing.rs
```diff
@@ -139,7 +139,6 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
         is_renormalize: config.is_renormalize.unwrap_or(false),
         device_class_synthesis_config: false, // only supported in relay
         enrich_spans: false,
-        normalize_spans: false,
```
iker-barriocanal, line 142: Actually I'm not sure if this is a breaking change, going to verify it.

relay-event-normalization/src/event.rs
```diff
@@ -228,14 +223,9 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) {
...
-    if config.normalize_spans && event.ty.value() == Some(&EventType::Transaction) {
-        // XXX(iker): span normalization runs in the store processor, but
-        // the exclusive time is required for span metrics. Most of
-        // transactions don't have many spans, but if this is no longer the
-        // case and we roll this flag out for most projects, we may want to
-        // reconsider this approach.
+    if event.ty.value() == Some(&EventType::Transaction) {
         crate::normalize::normalize_app_start_spans(event);
-        span::attributes::normalize_spans(event, &BTreeSet::from([SpanAttribute::ExclusiveTime]));
+        span::attributes::compute_span_exclusive_time(event);
```
jjbayer: Let's remove the module.

```suggestion
        span::exclusive_time::compute_span_exclusive_time(event);
```

relay-cabi/src/processing.rs
jjbayer: I noticed we always run exclusive_time computation, even if the exclusive_time is already there. We should either add a check on field presence to prevent recomputation, or check the `is_renormalize` flag instead.

jjbayer (commented):
We should rename `span_attributes` on the project config to `deprecated`, as we did here: https://github.com/getsentry/relay/blob/84f69b405a140e8ab8eaac740e4d7f8d4ed84eaa/relay-dynamic-config/src/feature.rs#L49-L51

iker-barriocanal:
> We should rename span_attributes on the project config to deprecated

I'll do that in a follow-up PR, because I'd also like to eliminate them in the project config, and scoping these two changes in a separate PR (vs in this one) is less obscure.

relay-cabi/src/processing.rs
iker-barriocanal: I ran a bunch of tests related to the event manager with librelay in a local sentry instance and all passed in case I was missing something (although not required as we are explicitly setting the value there), so this should be fine.

Agree on the `is_renormalize` flag, implemented in [4741bf7](https://github.com/getsentry/relay/pull/3032/commits/4741bf790f27edaed20127c59d42ab2ad9efda53).

tests/integration/test_envelope.py
```diff
@@ -367,119 +367,6 @@ def test_ops_breakdowns(mini_sentry, relay_with_processing, transactions_consume
     }
 
 
-def test_no_span_attributes(mini_sentry, relay_with_processing, transactions_consumer):
```
iker-barriocanal, line 370: `spanAttributes` is no longer used, so the test is stale.

tests/integration/test_envelope.py
```diff
@@ -367,119 +367,6 @@ def test_ops_breakdowns(mini_sentry, relay_with_processing, transactions_consume
...
-
-    event, _ = events_consumer.get_event()
-    assert event["transaction"] == "/organizations/:orgId/performance/:eventSlug/"
-    assert "trace" in event["contexts"]
-    assert "exclusive_time" not in event["contexts"]["trace"]
-    for span in event["spans"]:
-        assert "exclusive_time" not in span
-
-
-def test_empty_span_attributes(
```
iker-barriocanal, line 426: `spanAttributes` is no longer used, so the test is stale.

jjbayer (approved):

PR #3035, opened 2024-01-31
feat(replays): Combined Envelope Items (reopen)

JoshFerge:
Reopening https://github.com/getsentry/relay/pull/2170

- We originally closed it because there wasn't a clear use case. Now there is. https://github.com/getsentry/sentry/issues/60826

I've resolved merge conflicts, tests pass, and I believe comments from the previous PR have been addressed.


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Emit a usage metric for total spans. ([#3007](https://github.com/getsentry/relay/pull/3007))
 - Drop timestamp from metrics partition key. ([#3025](https://github.com/getsentry/relay/pull/3025))
 - Drop spans ending outside the valid timestamp range. ([#3013](https://github.com/getsentry/relay/pull/3013))
+- Adds support for combining replay envelope items ([#3035](https://github.com/getsentry/relay/pull/3035))
```
```suggestion
- Add support for combining replay envelope items. ([#3035](https://github.com/getsentry/relay/pull/3035))
```

relay-dynamic-config/src/feature.rs
```diff
@@ -11,6 +11,10 @@ pub enum Feature {
     /// Enables data scrubbing of replay recording payloads.
     #[serde(rename = "organizations:session-replay-recording-scrubbing")]
     SessionReplayRecordingScrubbing,
+    /// Enables combining session replay envelope item (Replay Recordings and Replay Events).
```
```suggestion
    /// Enables combining session replay envelope items (Replay Recordings and Replay Events).
```

relay-server/src/services/processor/replay.rs
```diff
@@ -151,3 +154,168 @@ fn process_replay_event(
 
     Ok(replay)
 }
+
+fn process_replays_combine_items(state: &mut ProcessEnvelopeState) -> Result<(), ProcessingError> {
+    let project_state = &state.project_state;
+    let combined_envelope_items =
+        project_state.has_feature(Feature::SessionReplayCombinedEnvelopeItems);
+
+    if combined_envelope_items {
```
jjbayer: nit: Would it make sense to move this check outside the function, or else turn it into a guard, i.e.

```rust
if !project_state.has_feature(Feature::SessionReplayCombinedEnvelopeItems) {
    return Ok(());
}

File: relay-server/src/services/processor/replay.rs [hunk 4]
```diff
@@ -151,3 +154,168 @@ fn process_replay_event(
...
+        {
+            if let Some(replay_recording_item) =
+                envelope.take_item_by(|item| item.ty() == &ItemType::ReplayRecording)
+            {
+                let mut data = Vec::new();
+                let mut combined_item_payload = BTreeMap::new();
+
+                combined_item_payload.insert("replay_event", replay_event_item.payload().to_vec());
+                combined_item_payload
+                    .insert("replay_recording", replay_recording_item.payload().to_vec());
```
jjbayer: I would create a type for this, something like
```rust
#[serde(Serialize, Deserialize)]
struct CombinedReplay {
    replay_event: Bytes,
    replay_recording: Bytes
}
```

relay-server/src/services/processor/replay.rs
```diff
@@ -151,3 +154,168 @@ fn process_replay_event(
...
+
+                if let Err(e) = rmp_serde::encode::write(&mut data, &combined_item_payload) {
+                    relay_log::error!(
+                        "failed to serialize combined replay event and recording: {}",
+                        e
+                    );
+                    // TODO: what to do here? Drop + emit outcome?
```
jjbayer: If we're adding back the original items then I would say no.

Otherwise you can probably use https://github.com/getsentry/relay/blob/c3ac7c0eba8c48587c7fab0621c741f5dc111aee/relay-server/src/utils/managed_envelope.rs#L265.

relay-server/src/services/processor/replay.rs
```diff
@@ -151,3 +154,168 @@ fn process_replay_event(
...
+                }
+
+                let mut combined_item = Item::new(ItemType::CombinedReplayEventAndRecording);
+
+                combined_item.set_payload(ContentType::MsgPack, data);
+                envelope.add_item(combined_item);
+            } else {
+                envelope.add_item(replay_event_item)
```
jjbayer: Should we also put back `replay_recording_item` here?

tests/integration/test_replay_combined_payload.py
```diff
@@ -0,0 +1,88 @@
...
+#     )
+#     envelope.add_item(Item(payload=PayloadRef(json=replay_event), type="replay_event"))
+
+#     relay.send_envelope(42, envelope)
+
+#     envelope = mini_sentry.captured_events.get(timeout=20)
+#     assert len(envelope.items) == 1
+
+#     replay_event = envelope.items[0]
+#     assert replay_event.type == "replay_recording_not_chunked"
```
jjbayer: Is this still relevant?

jjbayer (changes requested):
Is the goal of this PR purely to unite the two items in the kafka message? If so, I would not introduce a new item type, and implement the entire logic in the kafka producer (`store.rs`) instead.

If the goal is to send combined items from relay to relay (in a chain of relays), we need to add support for parsing a combined item and processing it correctly. Right now, I believe such an item would hit this branch and cause an error: https://github.com/getsentry/relay/blob/8411b1dfaaee645658cd5e9d5428058f2b19384f/relay-server/src/services/processor.rs#L233-L235

relay-server/src/services/store.rs
```diff
@@ -812,6 +821,38 @@ impl StoreService {
...
+        let message =
+            KafkaMessage::ReplayRecordingNotChunked(ReplayRecordingNotChunkedKafkaMessage {
+                replay_id,
+                project_id: scoping.project_id,
+                org_id: scoping.organization_id,
+                key_id: scoping.key_id,
+                retention_days,
+                received: UnixTimestamp::from_instant(start_time).as_secs(),
+                version: Some(1),
+                payload: item.payload(),
```
cmanallen, line 836: It would be nice if we could maintain backwards compatibility on this attribute by serializing the event on a separate key.  That would allow us to eventually rollout a consumer change rather than having to update the consumer preemptively to expect the combined type.

cmanallen: @jjbayer How simple is this task to accomplish?

JoshFerge: I've gone ahead and have just serialized the replay event on a separate key!

JoshFerge:
@jjbayer @cmanallen I've gone ahead and moved the logic to store -- now the change simply will look for the two replay items, and combine them if if finds them. A couple of questions to ponder:

1. How should we handle this roll-out? If we are able to use project features in `store` here, I would love to try this on the Sentry project first. If we cannot do that, it seems like we'd have to do some kind of hard cutover. 

My general idea here would be to:

- Double produce, and simply log and ignore the messages with a replay_event on them in our consumer, and ensure that everything there is working correctly.

- Remove the production of Combined Messages temporarily, let any combined messages drain out

- Then do a hard cutover, where we change relay to produce combined messages instead of stand-alone replay recording messages. 

- (Note, we've decided we're still going to produce `replay_events` from relay, so only recording messages will be impacted)


If we are able to use project features in store here, that would be more ideal, and also open to suggestions, perhaps we could use S4S if its set up with replays? Any other ideas?

relay-server/src/services/store.rs
```diff
@@ -399,6 +425,41 @@ impl StoreService {
...
 
+    fn extract_combined_replay_kafka_message(
+        event_id: EventId,
+        replay_items: Vec<&Item>,
+        scoping: Scoping,
+        start_time: Instant,
+        retention: u16,
+    ) -> Option<KafkaMessage> {
+        let mut replay_event_item = None;
+        let mut replay_recording_item = None;
```
jjbayer: nit: We could declare these variables in `store_envelope`, so we don't have to transform to and from a vector.

jjbayer (commented):
@JoshFerge in [`replay::process`](https://github.com/getsentry/relay/blob/0de38dcb15b2c4b20b03ba280eac8df67ecd43ea/relay-server/src/services/processor/replay.rs#L25), you have access to project features. So in that function you could set an internal item header similar to [this one](https://github.com/getsentry/relay/blob/0de38dcb15b2c4b20b03ba280eac8df67ecd43ea/relay-server/src/envelope.rs#L489-L490), and check that item header in `store_envelope`.

JoshFerge:
thank you @jjbayer. With the item header added now, we can try this out internally before releasing for everyone.

relay-server/src/envelope.rs
```diff
@@ -489,6 +489,11 @@ pub struct ItemHeaders {
     #[serde(default, skip)]
     rate_limited: bool,
 
+    /// Indicates that this item should be combined into one payload with othe replay item.
```
```suggestion
    /// Indicates that this item should be combined into one payload with other replay item.
```

relay-server/src/envelope.rs
```diff
@@ -739,6 +745,17 @@ impl Item {
         self.headers.source_quantities = Some(source_quantities);
     }
 
+    /// Returns if the payload's replay items should be combined into one kafka message.
+    #[cfg_attr(not(feature = "processing"), allow(dead_code))]
```
jjbayer: Why not just

```suggestion
    #[cfg(processing)]
```

?

relay-server/src/services/store.rs
```diff
@@ -269,6 +284,26 @@ impl StoreService {
...
+            let combined_replay_kafka_message = Self::extract_combined_replay_kafka_message(
+                event_id.ok_or(StoreError::NoEventId)?,
+                replay_event,
+                replay_recording,
+                scoping,
+                start_time,
+                retention,
+            );
+            self.produce(
+                KafkaTopic::ReplayRecordings,
```
jjbayer: I assume the recordings consumer will support the new format before we deploy relay?

jjbayer (approved):
Thanks for contributing! See remaining comments, and maybe wait until Monday before deploying.

relay-server/src/services/store.rs
```diff
@@ -269,6 +284,26 @@ impl StoreService {
             }
         }
 
+        if let (Some(replay_event), Some(replay_recording)) = (replay_event, replay_recording) {
```
cmanallen: This is not exhaustive.  Consider the states `(Some(replay_event), None)` and `(None, Some(replay_recording))`.  In these cases we would need to produce the individual events.

relay-server/src/services/store.rs
```diff
@@ -244,16 +247,28 @@ impl StoreService {
...
+                        replay_event = Some(item);
+                    } else {
+                        self.produce_replay_event(
+                            event_id.ok_or(StoreError::NoEventId)?,
+                            scoping.organization_id,
+                            scoping.project_id,
+                            start_time,
+                            retention,
+                            item,
+                        )?;
```
cmanallen: I'll preface this by saying this is a personal opinion.  I think this section needs to be refactored.  If we add the missing states from my previous comment we'll have duplicated producer code.  We'll also be producing effects in two different locations in the function.  Over the long term I think this is more error prone than doing everything close together.

My suggestion is: aggregate the state and evaluate that state in one place.  In psuedo code it looks like this:

```rust
fn store_envelope() {
    for ... {
        match ... {
            ItemType::ReplayEvent => {
                replay_event = Some(item);
            }
            ItemType::ReplayRecording => {
                replay_recording = Some(item);
            }
        }
    }

    if item.replay_combined_payload() {
        // Handles all the cases for combined types.
        self.produce_replay_messages(replay, replay_recording)
    } else {
        // Handles all the cases for separate types.
        self.produce_replay_messages(replay, None)
        self.produce_replay_messages(None, replay_recording)
    }
}

fn produce_replay_messages(x, y) {
    match (x, y) => {
        (Some(x), Some(y)) => {
            // produce combined
        },
        (Some(x), None) => {
            // produce x
        },
        (None, Some(y)) => {
            // produce y
        }
        _ => {
            // do nothing
        }
    }
}
```

relay-server/src/services/store.rs
cmanallen: We should also have test coverage for these branches.

cmanallen (changes requested):

relay-server/src/services/store.rs
JoshFerge: yes!

relay-server/src/services/store.rs
JoshFerge: thank you! will also change it as discussed so that we're still producing the replay event to the snuba topic directly from here as well.

relay-server/src/services/store.rs
```diff
@@ -812,6 +834,63 @@ impl StoreService {
         Ok(())
     }
 
+    #[allow(clippy::too_many_arguments)]
```
JoshFerge: not sure what to do about the too many args here

relay-server/src/services/store.rs
```diff
@@ -244,16 +248,17 @@ impl StoreService {
...
-                    self.produce_replay_recording(event_id, scoping, item, start_time, retention)?
+                    if item.replay_combined_payload() {
+                        send_combined_replay_envelope = true
+                    }
+                    replay_recording = Some(item);
+                }
+                ItemType::ReplayEvent => {
+                    if item.replay_combined_payload() {
+                        send_combined_replay_envelope = true
+                    }
```
cmanallen: Optional.  Looks prettier if you do this:

```rust
send_combined_replay_envelope = item.replay_combined_payload();
```

cmanallen (commented):
Whoops forgot to publish this review on Friday...

cmanallen (approved):

JoshFerge:
@cmanallen @jjbayer should be good for final review.

cmanallen (approved):

relay-server/src/services/store.rs
TBS1996: the functionality would differ then, if the next item is 'false' then itd override. you'd have to do `send_combined_replay_envelope |= item.replay_combined_payload();`

relay-server/src/services/store.rs
```diff
@@ -812,6 +836,47 @@ impl StoreService {
...
+
+            self.produce_replay_recording(
+                Some(replay_id),
+                scoping,
+                replay_recording,
+                combined_replay_event,
+                start_time,
+                retention_days,
+            )?;
+        }
```
TBS1996: the `replay_event.is_some()` check seems unnecesary, if it's none you can just use the variable directly anyway 

```suggestion
        if let Some(replay_recording) = replay_recording {
            let combined_replay_event = if send_combined_replay_envelope{
                replay_event
            } else {
                None
            };

            self.produce_replay_recording(
                Some(replay_id),
                scoping,
                replay_recording,
                combined_replay_event,
                start_time,
                retention_days,
            )?;
        }
```

relay-server/src/services/store.rs
```diff
@@ -838,6 +801,11 @@ impl StoreService {
         // Remaining bytes can be filled by the payload.
         let max_payload_size = self.config.max_replay_message_size() - max_message_metadata_size;
 
+        let mut replay_event_payload = None;
+        if let Some(replay_event) = replay_event {
+            replay_event_payload = Some(replay_event.payload());
+        }
```
jjbayer: nit: 
```suggestion
        let replay_event_payload = replay_event.map(|e| e.payload());
```

relay-server/src/services/store.rs
jjbayer: Instead of passing `replay_event`, `replay_recording`, and a boolean flag, the most idiomatic would probably be to pass an enum like 

```rust
enum Data {
    Event(&Item),
    Recording(&Item),
    Combined { event: &Item, recording: &Item }
}

That would also make states like `(None, None, true)` impossible.

cmanallen:
@jjbayer @TBS1996 Josh is out for the rest of the week.  I've updated the PR.  Please provide final approval.

Note: `replay_event` is only populated if the combined envelope feature is enabled.  This means we don't need to consider the combined envelope boolean in any other step.  Null is always sent if the feature is disabled or if replay_event was not present in the envelope.

Note: the kafka payload size calculation was reconfigured.  I'm using addition rather than subtraction.  This is to avoid situations where the replay_event is larger than the maximum size of a recording.  This would yield a negative number but since its of type `usize` the value wraps around and could give incorrect results.  Addition suffers from similar wrap around problems however because we our server's do not have 18 exabits of RAM I feel safe in using this method.

Note: oversized recording messages will now produce an outcome with discard reason `TooLarge`.

jjbayer (approved):

PR #3041, opened 2024-02-01
feat(spans): Add browser name to Sentry tags

phacops:
This is a follow-up to this previous PR (https://github.com/getsentry/relay/pull/2969). We need to insert `browser.name` in the span's `data` field and then extract the value to `sentry_tags`.


iker-barriocanal (approved):
LGTM. Please add a test before merging.

iker-barriocanal (approved):

PR #3044, opened 2024-02-01
Add protobuf support for onlyspans ingestion

sl0thentr0py:
* use official `prost` and `serde` compatible structs from `opentelemetry-proto`
* use `axum-extra` with the `protobuf` feature for the protobuf extractor
* removed our old custom structs and replaced the transformation logic

TODO
* [x] `trace_id/span_id` are `Vec<u8>`, fix that conversion
* [x] ~the date conversion helper from strings is not added to their serde derives, so I've ignored this test for now~ will need to do upstream later


relay-server/src/endpoints/spans.rs
```diff
@@ -1,37 +1,58 @@
...
+    B::Error: Into<axum::BoxError>,
+{
+    let mut trace = None;
+    if content_type.as_ref().starts_with("application/json") {
+        let json: Json<TracesData> = request.extract().await?;
+        trace = Some(json.0)
+    } else if content_type.as_ref().starts_with("application/x-protobuf") {
+        let protobuf: Protobuf<TracesData> = request.extract().await?;
+        trace = Some(protobuf.0)
+    }
```
iker-barriocanal: I think this should work:
```suggestion
    let trace = if content_type.as_ref().starts_with("application/json") {
        let json: Json<TracesData> = request.extract().await?;
        Some(json.0)
    } else if content_type.as_ref().starts_with("application/x-protobuf") {
        let protobuf: Protobuf<TracesData> = request.extract().await?;
        Some(protobuf.0)
    };
```

relay-server/src/endpoints/spans.rs
```diff
@@ -1,37 +1,58 @@
...
+
+    if let Some(trace) = trace {
```
iker-barriocanal: `let..else` allows you to remove one level of indentation:
```suggestion
    let Some(trace) = trace else {
        Ok(...)
    };
    // .......
```

relay-spans/Cargo.toml
```diff
@@ -13,8 +13,10 @@ publish = false
 chrono.workspace = true
 enumset = "1.0.4"
 once_cell.workspace = true
+opentelemetry-proto = { git = "https://github.com/open-telemetry/opentelemetry-rust", features = ["gen-tonic", "with-serde", "trace"] }
```
iker-barriocanal: The crate is available in [crates.io](https://crates.io/crates/opentelemetry), any reason to install it from git?

iker-barriocanal (commented):
Overall looks good to me, leaving the final review to @jjbayer as I don't have enough context to provide proper feedback.

relay-spans/Cargo.toml
sl0thentr0py: need a new PR that's not released yet, asked here
https://github.com/open-telemetry/opentelemetry-rust/pull/1462#issuecomment-1924135271

relay-spans/src/utils.rs
```diff
@@ -4,6 +4,8 @@ use std::str::FromStr;
 use serde::{de, Deserialize};
 use serde_json::{Map, Value};
 
+// TODO-neel-protobuf do this shit idk how
+#[allow(dead_code)]
```
phacops: You can remove the whole file, this is not needed anymore since you're using the generated code to deserialize.

relay-spans/src/span.rs
```diff
@@ -144,12 +41,18 @@ fn convert_from_otel_to_sentry_status(
...
 
-impl From<OtelSpan> for EventSpan {
-    fn from(from: OtelSpan) -> Self {
+/// Extension trait for OtelSpan.
+pub trait OtelSpanExt {
+    /// Transform an OtelSpan to an Sentry span.
+    fn to_sentry_span(self) -> EventSpan;
+}
+
+impl OtelSpanExt for OtelSpan {
```
phacops: Any reason not to keep the `From` trait?

sl0thentr0py: i can't implement a foreign trait on a foreign type, see
https://doc.rust-lang.org/error_codes/E0117.html

relay-spans/src/utils.rs
sl0thentr0py: ok but then i'll kill the test too, we can add try to add the string timestamp deser upstream

relay-spans/src/lib.rs
```diff
@@ -6,11 +6,11 @@
     html_favicon_url = "https://raw.githubusercontent.com/getsentry/relay/master/artwork/relay-icon.png"
 )]
 
-pub use crate::span::OtelSpan;
-pub use crate::trace::TracesData;
+pub use crate::span::OtelSpanExt;
+
+pub use opentelemetry_proto::tonic::common::v1 as OtelCommon;
```
Dav1dde:
```suggestion
pub use opentelemetry_proto::tonic::common::v1 as otel_common;
```

It's a module not a type

relay-spans/src/lib.rs
```diff
@@ -6,11 +6,11 @@
...
+pub use opentelemetry_proto::tonic::trace::v1 as OtelTrace;
```
Dav1dde:
```suggestion
pub use opentelemetry_proto::tonic::trace::v1 as otel_trace;
```

relay-spans/Cargo.toml
Dav1dde: Not a fan of a git dependency but if we have to it should always include a fixed commit hash (use `rev = `)!

relay-spans/src/span.rs
Dav1dde: Could just be a free standing function?

relay-spans/src/span.rs
```diff
@@ -241,140 +152,26 @@ impl From<OtelSpan> for EventSpan {
...
-    /// Indicates that the span describes a producer sending a message to a broker.
-    /// Unlike CLIENT and SERVER, there is often no direct critical path latency relationship
-    /// between producer and consumer spans. A PRODUCER span ends when the message was accepted
-    /// by the broker while the logical processing of the message might span a much longer time.
-    Producer = 4,
-    /// Indicates that the span describes consumer receiving a message from a broker.
-    /// Like the PRODUCER kind, there is often no direct critical path latency relationship
-    /// between producer and consumer spans.
-    Consumer = 5,
+trait OtelValueExt {
```
Dav1dde: Imo this should just be free standing functions, especially because `to_string` collides with the `Display` trait and it is easy to get that wrong

relay-spans/src/span.rs
```diff
@@ -241,140 +152,26 @@ impl From<OtelSpan> for EventSpan {
...
+    fn to_i64(&self) -> Option<i64> {
         match self {
-            AnyValue::Int(v) => Some(*v),
+            OtelValue::IntValue(v) => Some(*v),
             _ => None,
         }
     }
 
-    pub fn to_string(&self) -> Option<String> {
+    fn to_string(&self) -> Option<String> {
```
Dav1dde: I would prefer if this was something like `into_string` and takes `self`, that means in the `AnyValue::String` case we don't have to clone (and the caller can still clone if he has to, explicitly).

Dav1dde (changes requested):
We definitely have to make sure to not have an unpinned git dependency. Also would like to get a review from @jjbayer, I am missing context and was just looking for Rust stuff.

relay-spans/src/span.rs
sl0thentr0py: if y'all prefer that, sure, idk much rust convention

relay-spans/Cargo.toml
jjbayer: +1, please pin the version.

relay-spans/src/lib.rs
jjbayer: AFAICT `OtelCommon` is only used inside this crate, so I would remove the `pub use` completely.

relay-spans/src/span.rs
jjbayer: As long as it's an interface with one function and one implementation, I also prefer a standalone function.

jjbayer (commented):
@sl0thentr0py thanks for doing this! Nice to see some LoCs removed. Please add a test for binary protobuf ingestion to `test_store.py`. See https://github.com/getsentry/relay/blob/1f9b73145ec11cddae78d17f609665636a907e3a/tests/integration/test_store.py#L1412

relay-server/src/endpoints/spans.rs
```diff
@@ -1,22 +1,44 @@
...
-    Json(trace): Json<TracesData>,
-) -> Result<impl IntoResponse, BadStoreRequest> {
+    request: Request<B>,
+) -> axum::response::Result<impl IntoResponse>
+where
+    B: axum::body::HttpBody + Send + 'static,
+    B::Data: Send + Into<Bytes>,
+    B::Error: Into<axum::BoxError>,
+{
+    let trace = if content_type.as_ref().starts_with("application/json") {
```
Dav1dde: What do you think about:

```rs
async fn handle<B>(
    state: ServiceState,
    content_type: RawContentType,
    meta: RequestMeta,
    request: Request<B>,
) -> axum::response::Result<impl IntoResponse>
where
    B: axum::body::HttpBody + Send + 'static,
    B::Data: Send + Into<Bytes>,
    B::Error: Into<axum::BoxError>,
{
    let trace: TracesData = if content_type.as_ref().starts_with("application/json") {
        let Json(trace) = request.extract().await?;
        trace
    } else if content_type.as_ref().starts_with("application/x-protobuf") {
        let Protobuf(trace) = request.extract().await?;
        trace
    } else {
        return Ok(StatusCode::UNSUPPORTED_MEDIA_TYPE);
    };

    let mut envelope = Envelope::from_request(None, meta);
    for resource_span in trace.resource_spans {
        for scope_span in resource_span.scope_spans {
            for span in scope_span.spans {
                let Ok(payload) = serde_json::to_vec(&span) else {
                    continue;
                };
                let mut item = Item::new(ItemType::OtelSpan);
                item.set_payload(ContentType::Json, payload);
                envelope.add_item(item);
            }
        }
    }
    common::handle_envelope(&state, envelope).await?;

    Ok(StatusCode::ACCEPTED)
}
```

No need to have the option for `trace` and we can nicely pattern match to extract the data.

Dav1dde (commented):
Thanks looks good to me! Left a small improvement to the `handle` function.

It would also be nice to have at least one integration test which makes sure we're parsing the data correctly.

> the date conversion helper from strings is not added to their serde derives, so I've ignored this test for now

To you still want to address this todo from your PR description?

sl0thentr0py:
added integration test
the date stuff will need to be done upstream later, I think it's fine to remove that functionality for now since we're relying on official otel structs anyway and if it's in the spec they will eventually add it.

relay-spans/src/span.rs
```diff
@@ -144,281 +41,133 @@ fn convert_from_otel_to_sentry_status(
...
+        OtelValue::DoubleValue(v) => Some(v.to_string()),
+        OtelValue::BytesValue(v) => match String::from_utf8(v) {
+            Ok(v) => Some(v),
+            Err(_) => None,
+        },
+        _ => None,
+    }
+}
+
+/// Transform an OtelSpan to an Sentry span.
```
jjbayer:
```suggestion
/// Transform an OtelSpan to a Sentry span.
```

jjbayer (approved):

Dav1dde (approved):
👍

PR #3045, opened 2024-02-02
feat(metrics-summaries): Emit summaries to their own topic

phacops:
Instead of pushing metrics summaries as part of spans and having the Snuba consumer reads all spans, we're going to emit metrics summaries to their own topic.


relay-server/src/services/store.rs
```diff
@@ -896,13 +902,47 @@ impl StoreService {
...
+            for (mri, summaries) in metrics_summary {
+                let Some(summaries) = summaries else {
+                    continue;
+                };
+                for summary in summaries {
+                    let Some(summary) = summary else {
+                        continue;
+                    };
+                    // Ignore immedate errors on produce.
+                    let _ = self.produce(
```
jjbayer: Should we log an error here? Might be interesting to see if / how often this happened.

relay-server/src/services/store.rs
```diff
@@ -896,13 +902,47 @@ impl StoreService {
...
+                        KafkaTopic::MetricsSummaries,
+                        scoping.organization_id,
+                        KafkaMessage::MetricsSummary(MetricsSummaryKafkaMessage {
+                            count: summary.count.unwrap_or_default(),
```
jjbayer: I assume that the fields of `MetricsSummary` are `Optional` for a reason. IMO we should do one of two things:

1. Make the fields of `MetricsSummary` non-optional.
2. If it actually makes sense that these fields are empty, make the consumer accept missing / `null` values.

relay-server/src/services/store.rs
```diff
@@ -1269,7 +1309,7 @@ struct SpanKafkaMessage<'a> {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     segment_id: Option<&'a str>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    sentry_tags: Option<&'a RawValue>,
+    sentry_tags: Option<BTreeMap<&'a str, &'a str>>,
```
jjbayer: Remember special characters.

```suggestion
    sentry_tags: Option<BTreeMap<&'str, String>>,
```

relay-server/src/services/store.rs
```diff
@@ -1278,6 +1318,25 @@ struct SpanKafkaMessage<'a> {
...
+    is_segment: bool,
+    max: f64,
+    mri: &'a str,
+    min: f64,
+    project_id: u64,
+    retention_days: u16,
+    segment_id: &'a str,
+    span_id: &'a str,
+    sum: f64,
+    tags: BTreeMap<String, String>,
```
jjbayer: If we make `SpanMetricsSummary::tags` a non-optional `BTreeMap` with `skip_serializing_if=BTreeMap::is_empty`, we can borrow here:

```suggestion
    sentry_tags: &'a BTreeMap<String, String>>,
```

relay-server/src/services/store.rs
phacops: I think I forgot where I was going with this. The consumer supports missing values but not `null` values and I think I wanted to turn that into non-optional values.

relay-server/src/services/store.rs
phacops: I ended up making it a `RawValue` because I had trouble doing what you were suggesting as:
```
the trait `Deserialize<'_>` is not implemented for `&'a BTreeMap<&'a str, std::string::String>`
```

tests/integration/test_store.py
```diff
@@ -1886,7 +1885,6 @@ def test_span_extraction_with_ddm_missing_values(
             start_timestamp.replace(tzinfo=timezone.utc).timestamp() * 1e3
         ),
         "trace_id": "a0fa8803753e40fd8124b21eeb2986b5",
-        "_metrics_summary": metrics_summary,
```
jjbayer, line 1935: Please add a `metrics_summary_consumer` and check that it gets the message.

relay-kafka/src/config.rs
```diff
@@ -111,6 +114,8 @@ pub struct TopicAssignments {
     pub monitors: TopicAssignment,
     /// Standalone spans without a transaction.
     pub spans: TopicAssignment,
+    /// Summary for metrics collected during a span.
+    pub metrics_summaries: TopicAssignment,
```
jjbayer, line 118: Let's make sure we have the topic created and relay configuration in place before merging this, to not risk auto-creating the topic on the default cluster.

relay-server/src/services/store.rs
```diff
@@ -896,13 +902,52 @@ impl StoreService {
...
-        span.duration_ms = ((span.end_timestamp - span.start_timestamp) * 1e3) as u32;
-        span.event_id = event_id;
-        span.project_id = scoping.project_id.value();
-        span.retention_days = retention_days;
-        span.start_timestamp_ms = (span.start_timestamp * 1e3) as u64;
+            let group = span
+                .sentry_tags
+                .as_ref()
+                .and_then(|sentry_tags| sentry_tags.get("group").cloned())
+                .unwrap_or_default();
```
jjbayer: nit
```suggestion
                .and_then(|sentry_tags| sentry_tags.get("group"))
                .map_or("", String::as_str);
```

relay-server/src/services/store.rs
```diff
@@ -896,13 +902,52 @@ impl StoreService {
...
+                        }),
+                    ) {
+                        relay_log::error!(
+                            error = &error as &dyn std::error::Error,
+                            "failed to push metrics summary to kafka",
+                        );
+                    }
+                }
+            }
+            span.metrics_summary = None;
```
jjbayer: `SpanKafkaMessage` should document which fields are actually sent to Kafka on the `spans` topic, so I suggest we remove this field from the type and deserialize into a separate helper struct instead.

```rust
[derive(Deserialize)]
struct Helper {
    metrics_summary: Option<BTreeMap<Cow<'a, str>, Option<SpanMetricsSummaries<'a>>>>
}

File: relay-server/src/services/store.rs [hunk 9]
```diff
@@ -896,13 +902,52 @@ impl StoreService {
...
+                .as_ref()
+                .and_then(|sentry_tags| sentry_tags.get("group").cloned())
+                .unwrap_or_default();
+
+            for (mri, summaries) in metrics_summary {
+                let Some(summaries) = summaries else {
+                    continue;
+                };
+                for summary in summaries {
+                    let Some(summary) = summary else {
```
jjbayer: nit: You could [destructure](https://doc.rust-lang.org/rust-by-example/flow_control/match/destructuring/destructure_structures.html) `summary` here.

relay-server/src/services/store.rs
```diff
@@ -1253,7 +1298,7 @@ struct SpanKafkaMessage<'a> {
         rename = "_metrics_summary",
         skip_serializing_if = "Option::is_none"
     )]
-    metrics_summary: Option<BTreeMap<Cow<'a, str>, Option<SpanMetricsSummaries>>>,
+    metrics_summary: Option<BTreeMap<Cow<'a, str>, Option<SpanMetricsSummaries<'a>>>>,
```
jjbayer: AFAICT this `Cow` will always become `Cow::Owned`, so we might as well remove the  `#[serde(borrow)]` and use `String`.

Playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=63202f57b8a55bea32004f85b905d235

phacops: I think I still need the `borrow` because `'a` needs to outlive `'de` for the `tags: <&'a RawValue>`.

relay-kafka/src/config.rs
phacops: The topic was already created by @mwarkentin on the `spans` cluster but I should probably add some config to Relay in order to not auto-create it on the `events` cluster.

phacops: https://github.com/getsentry/ops/pull/9359

phacops:
Created a schema for this topic: https://github.com/getsentry/sentry-kafka-schemas/pull/212.

relay-server/src/services/store.rs
```diff
@@ -972,6 +910,121 @@ impl StoreService {
...
+                );
+                self.outcome_aggregator.send(TrackOutcome {
+                    category: DataCategory::SpanIndexed,
+                    event_id: None,
+                    outcome: Outcome::Invalid(DiscardReason::InvalidSpan),
+                    quantity: 1,
+                    remote_addr: None,
+                    scoping,
+                    timestamp: instant_to_date_time(start_time),
+                });
```
jjbayer: We should not emit an outcome here (at least not for spans), because at this point we have already parsed a `SpanKafkaMessage` successfully and there will be an accepted outcome for it in `produce_span`.

```suggestion
```

relay-server/src/services/store.rs
```diff
@@ -972,6 +910,121 @@ impl StoreService {
...
+    ) {
+        let payload = item.payload();
+        let d = &mut Deserializer::from_slice(&payload);
+        let mut metrics_summary: SpanWithMetricsSummary = match serde_path_to_error::deserialize(d)
+        {
+            Ok(span) => span,
+            Err(error) => {
+                relay_log::error!(
+                    error = &error as &dyn std::error::Error,
+                    "failed to parse span"
```
jjbayer:
```suggestion
                    "failed to parse metrics summary of span"
```

relay-server/src/services/store.rs
```diff
@@ -972,6 +910,121 @@ impl StoreService {
...
+                            || v.sum.is_some()
+                            || v.count.is_some();
+                    }
+                    false
+                });
+                !v.is_empty()
+            } else {
+                false
+            }
+        });
```
jjbayer: nit: Could this be inlined into the for-loop below?

jjbayer (approved):
See note about removing outcome, rest looks good to me! Is the config for SaaS already in place?

PR #3047, opened 2024-02-02
chore(spool): Improve unspool strategy 

olksdr:
The recent stats show that when we have a lot of records with only 1 envelope per key, the unspool is very slow, since the service has to go into DB and fetch 1 key at a time, which can block the `Buffer` and by accumulating the messages in the internal queue it can also use up more memory. 

This change is introducing batching for all the valid keys, which makes sure we can unspool many records at once, which allows us to decrease the number sent message between the services and also read from the disk more data in one go. 

Make sure we can unspool faster from the buffer:
* requests all the envelopes for all the valid projects
* batch those requests for all the incoming keys
* make sure we return keys back when is over low watermark

fix: https://github.com/getsentry/team-ingest/issues/257
also related to: https://github.com/getsentry/team-ingest/issues/267

#skip-changelog


relay-server/src/services/project_cache.rs
```diff
@@ -524,58 +522,17 @@ impl ProjectCacheBroker {
...
-        }
-
-        if !result.is_empty() {
-            self.buffer.send(DequeueMany::new(
-                partial_key,
-                result,
-                self.buffer_tx.clone(),
-            ))
-        }
+    pub fn dequeue(&mut self, keys: HashSet<QueueKey>) {
```
TBS1996: maybe could be inlined?

relay-server/src/services/project_cache.rs
```diff
@@ -907,40 +909,33 @@ impl ProjectCacheBroker {
...
+        // If we don't yet have the global config, we will defer dequeuing until we do.
+        if let GlobalConfigStatus::Pending = self.global_config {
+            self.buffer_unspool_backoff.reset();
+            self.schedule_unspool();
+            return;
+        }
         // If there is nothing spooled, schedule the next check a little bit later.
-        if self.index.is_empty() {
+        // And do *not* attempt to unspool if the assigned permits over low watermark.
+        if self.index.is_empty() && !self.buffer_guard.is_below_low_watermark() {
```
TBS1996: shouldn't this be or `||` ? Like isn't this two different conditions that both would imply we should skip spooling?

relay-server/src/services/project_cache.rs
```diff
@@ -899,6 +865,42 @@ impl ProjectCacheBroker {
...
+        } = key;
+
+        let is_own_state_valid = self.projects.get_mut(own_key).map_or(false, |project| {
+            // Returns `Some` if the project is cache otherwise None and also triggers refresh
+            // in background.
+            project
+                .get_cached_state(self.services.project_cache.clone(), false)
+                // Makes sure that the state also is valid.
+                .map_or(false, |state| !state.invalid())
+        });
```
TBS1996, line 886: not really significant but an early return here if own_state isn't valid should be a bit more performant

relay-server/src/services/project_cache.rs
```diff
@@ -899,6 +865,42 @@ impl ProjectCacheBroker {
...
+
+        let is_sampling_state_valid = if own_key != sampling_key {
+            self.projects
+                .get_mut(sampling_key)
+                .map_or(false, |project| {
+                    // Returns `Some` if the project is cache otherwise None and also triggers refresh
+                    // in background.
```
iker-barriocanal:
```suggestion
                    // Returns `Some` if the project is cached otherwise None and also triggers a refresh
                    // in the background.
```

relay-server/src/services/project_cache.rs
```diff
@@ -899,6 +865,42 @@ impl ProjectCacheBroker {
         }
     }
 
+    /// Returns `true` if the project state valid for the [`QueueKey`].
+    ///
+    /// Which includes the own key and the samplig key for the project.
+    fn is_state_valid(&mut self, key: &QueueKey) -> bool {
```
iker-barriocanal, line 873: Should we add to the docs that this method may trigger a refresh for all projects?

relay-server/src/services/spooler/mod.rs
```diff
@@ -342,9 +335,9 @@ impl InMemory {
     }
 
     /// Dequeues the envelopes from the in-memory buffer and send them to provided `sender`.
-    fn dequeue(&mut self, keys: &Vec<QueueKey>, sender: mpsc::UnboundedSender<ManagedEnvelope>) {
+    fn dequeue(&mut self, keys: HashSet<QueueKey>, sender: mpsc::UnboundedSender<ManagedEnvelope>) {
```
iker-barriocanal, line 338: nit: I wonder if we should type this as an iterator and not care about the actual data structure?

iker-barriocanal (approved):
🚀

relay-server/src/services/project_cache.rs
olksdr: yeah,  it should be `or` here. Thanks!

relay-server/src/services/project_cache.rs
olksdr: This is done on purpose, since I still want to trigger update for the `sampling_key` if it's there and also expired, so next iteration will be much faster

relay-server/src/services/project_cache.rs
olksdr: Added more docs

jjbayer (approved):

PR #3053, opened 2024-02-05
ref(cardinality): Log Sentry errors for cardinality limits

Dav1dde:
To also track which namespace/limit was affected, this PR also adds a hashset of limit IDs to the cardinality limiter return value.

#skip-changelog

### DIFF
--- a/relay-cardinality/src/limiter.rs
+++ b/relay-cardinality/src/limiter.rs
@@ -20,9 +20,9 @@ pub struct Scoping {
 }
 
 /// Accumulator of all cardinality limiter rejections.
-pub trait Rejections {
+pub trait Rejections<'a> {
     /// Called for ever [`Entry`] which was rejected from the [`Limiter`].
-    fn reject(&mut self, entry_id: EntryId);
+    fn reject(&mut self, limit_id: &'a str, entry_id: EntryId);
 }
 
 /// Limiter responsible to enforce limits.
@@ -30,16 +30,16 @@ pub trait Limiter {
     /// Verifies cardinality limits.
     ///
     /// Returns an iterator containing only accepted entries.
-    fn check_cardinality_limits<E, R>(
+    fn check_cardinality_limits<'a, E, R>(
         &self,
         scoping: Scoping,
-        limits: &[CardinalityLimit],
+        limits: &'a [CardinalityLimit],
         entries: E,
         rejections: &mut R,
     ) -> Result<()>
     where
         E: IntoIterator<Item = Entry>,
-        R: Rejections;
+        R: Rejections<'a>;
 }
 
 /// Unit of operation for the cardinality limiter.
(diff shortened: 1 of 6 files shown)


relay-config/src/config.rs
```diff
@@ -1291,12 +1291,21 @@ pub struct CardinalityLimiter {
...
     /// Defaults to 180 seconds, 3 minutes.
     pub cache_vacuum_interval: u64,
+
+    /// Sample rate for Cardinality Limiter Sentry errors.
+    ///
+    /// Rate needs to be between `0.0` and `1.0`.
+    /// If set to `1.0` all cardinality limiter rejections will be logged as a Sentry error.
+    ///
+    /// Defaults to 1% (0.01).
+    pub error_sample_rate: f32,
```
jjbayer: Not a blocker, but should we put this into global config instead? Would be nice to control it via automator.

jjbayer (approved):

relay-server/src/services/processor.rs
```diff
@@ -2251,6 +2263,13 @@ impl UpstreamRequest for SendEnvelope {
     }
 }
 
+/// Returns a boolean wether the current item should be sampled or discarded
+/// using the passed `rate` (0 <= rate <= 1).
```
```suggestion
/// Returns a boolean whether the current item should be sampled or discarded
/// using the passed `rate` (0 <= rate <= 1).
```

iker-barriocanal (approved):

relay-dynamic-config/src/global.rs
```diff
@@ -79,6 +74,16 @@ pub struct Options {
...
 
+    /// Sample rate for Cardinality Limiter Sentry errors.
+    ///
+    /// Rate needs to be between `0.0` and `1.0`.
+    /// If set to `1.0` all cardinality limiter rejections will be logged as a Sentry error.
+    #[serde(
+        rename = "relay.cardinality-limiter.error-sample-rate",
+        skip_serializing_if = "is_default"
+    )]
+    pub cardinality_limiter_error_sample_rate: f32,
```
jjbayer, line 85: We could theoretically make this `#[cfg(processing)]` (probably other cardinality limiter options as well).

relay-server/src/services/processor.rs
```diff
@@ -2251,6 +2262,13 @@ impl UpstreamRequest for SendEnvelope {
     }
 }
 
+/// Returns a boolean whether the current item should be sampled or discarded
```
jjbayer: nit: Didn't find a reference for this but this seems to be the Rust way.
```suggestion
/// Returns `true` if the current item should be sampled.
```

relay-cardinality/src/limiter.rs
```diff
@@ -136,30 +136,46 @@ impl<T: Limiter> CardinalityLimiter<T> {
...
+    fn new(source: Vec<T>, rejections: RejectionTracker<'a>) -> Self {
         Self {
             source,
-            rejections: rejections.0,
+            rejections: rejections.entries,
+            limits: rejections.limits,
         }
     }
 
+    /// Whether any items are rejected.
```
```suggestion
    /// Returns `true` if any items have been rejected.
```

jjbayer (approved):

relay-dynamic-config/src/global.rs
Dav1dde: Not a fan of this, it would just end up being collected in the fallback hashmap instead

PR #3054, opened 2024-02-05
ref(normalization): Move contexts to normalization

iker-barriocanal:
Moves the normalization of contexts from store normalization to regular normalization.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -775,7 +775,7 @@ fn normalize_contexts(contexts: &mut Annotated<Contexts>) {
     let _ = processor::apply(contexts, |contexts, _meta| {
         // Reprocessing context sent from SDKs must not be accepted, it is a Sentry-internal
         // construct.
-        // This does not run on renormalization anyway.
+        // [`normalize`] does not run on renormalization anyway.
         contexts.0.remove("reprocessing");
 
         for annotated in &mut contexts.0.values_mut() {
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -229,7 +229,7 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) {
     // Some contexts need to be normalized before metrics extraction takes place.
     normalize_contexts(&mut event.contexts);
 
-    if event.ty.value() == Some(&EventType::Transaction) && !config.is_renormalize {
+    if event.ty.value() == Some(&EventType::Transaction) {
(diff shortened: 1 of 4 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -229,7 +229,7 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) {
     // Some contexts need to be normalized before metrics extraction takes place.
     normalize_contexts(&mut event.contexts);
 
-    if event.ty.value() == Some(&EventType::Transaction) && !config.is_renormalize {
```
iker-barriocanal, line 232: `normalize` is not called when not renormalization, so this check is redundant (see line 141). I didn't realize about it when I added it last week.

relay-event-normalization/src/event.rs
```diff
@@ -2434,4 +2442,36 @@ mod tests {
         }
         "###);
     }
+
+    #[test]
+    fn test_normalization_removes_reprocessing_context() {
```
iker-barriocanal, line 2447: New tests, as I didn't see any and the feature seems somewhat important.

relay-event-normalization/src/event.rs
```diff
@@ -773,10 +773,18 @@ fn normalize_breakdowns(event: &mut Event, breakdowns_config: Option<&Breakdowns
 /// Normalizes incoming contexts for the downstream metric extraction.
 fn normalize_contexts(contexts: &mut Annotated<Contexts>) {
     let _ = processor::apply(contexts, |contexts, _meta| {
+        // Reprocessing context sent from SDKs must not be accepted, it is a Sentry-internal
+        // construct.
+        // This does not run on renormalization anyway.
```
```suggestion
        //[`normalize`] does not run on renormalization anyway.
```

Could we rename `normalize` to something like `primary_normalization` to make it more clear from the name that it's the part of normalization that runs when `! is_renormalize`? Might also be worth adding that piece of information to the function's doc comment.

jjbayer (approved):

relay-event-normalization/src/event.rs
iker-barriocanal: I agree it can be a bit confusing with all the normalization steps moving around. I think it should be clearer and easier to follow once there's a single normalization step. I'd like to avoid such names as it implies there are other normalization steps, or assume that's the only piece running on `!is_renormalize`. Happy to revisit this decision after migrating the store normalizer.

Addressed the comment suggestion in [389b5e0](https://github.com/getsentry/relay/pull/3054/commits/389b5e0c10ee504aa2e078b6c2ae42933beb1c04).

PR #3055, opened 2024-02-05
chore(logging): Add support to "off" for logging level

### LINKED ISSUE #3051: Can not turn off logging
### Environment
- latest `getsentry/relay` image
- Run on GKE Kubernetes

### Steps to reproduce
1. Run sentry with the following configuration
```
relay:
  mode: static
  upstream: https://<organization-domain>
  host: 0.0.0.0
  port: 3000
logging:
  level: off
```

### Expected Result
Stopping the log as described in https://docs.sentry.io/product/relay/options/#logging

### Actual Result
```
caused by: logging.level: invalid value: string "off", expected a log level at line 7 column 10
```
The relay works fine when I use `info` or `error` as the log level value, but it gives me that error when I use `off` log level value

olksdr:
Support all the logging levels mentioned in the docs https://docs.sentry.io/product/relay/options/#logging

This change basically addresses the support for `"off"` level.

fix: https://github.com/getsentry/relay/issues/3051

#skip-changelog

### DIFF
--- a/relay-log/src/setup.rs
+++ b/relay-log/src/setup.rs
@@ -1,12 +1,13 @@
 use std::borrow::Cow;
 use std::env;
-use std::fmt::Display;
+use std::fmt::{self, Display};
 use std::path::PathBuf;
 use std::str::FromStr;
 
+use relay_common::impl_str_serde;
 use sentry::types::Dsn;
 use serde::{Deserialize, Serialize};
-use tracing::{level_filters::LevelFilter, Level as TracingLevel};
+use tracing::level_filters::LevelFilter;
 use tracing_subscriber::{prelude::*, EnvFilter, Layer};
 
 #[cfg(feature = "dashboard")]
@@ -63,8 +64,7 @@ impl Display for LevelParseError {
     }
 }
 
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
-#[serde(rename_all = "lowercase")]
+#[derive(Clone, Copy, Debug)]
 pub enum Level {
     Error,
     Warn,
@@ -73,6 +73,8 @@ pub enum Level {
     Trace,
     Off,
 }
+
+impl_str_serde!(Level, "The logging level.");
 
 impl Level {
     /// Returns the tracing [`LevelFilter`].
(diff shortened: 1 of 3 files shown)


relay-log/src/setup.rs
```diff
@@ -47,53 +49,89 @@ pub enum LogFormat {
...
-            fn visit_str<E>(self, value: &str) -> Result<Level, E>
-            where
-                E: Error,
-            {
-                value
-                    .parse()
-                    .map_err(|_| Error::invalid_value(Unexpected::Str(value), &self))
-            }
+    /// Returns the tracing logging [`tracing::Level`].
+    pub fn level(&self) -> TracingLevel {
```
Dav1dde: Should we return `Option<TracingLevel>` here instead?

relay-log/src/setup.rs
```diff
@@ -47,53 +49,89 @@ pub enum LogFormat {
...
-    pub fn deserialize<'de, D>(deserializer: D) -> Result<Level, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        struct V;
-
-        impl<'de> Visitor<'de> for V {
-            type Value = Level;
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
Dav1dde: We have `relay_common::impl_str_serde` to make sure it uses the same impl as `FromStr`. E.g. the `FromStr` allows mixed case, this serde derive does not.

Dav1dde (approved):

relay-log/src/setup.rs
olksdr: removed this func

relay-log/src/setup.rs
olksdr: using `impl_str_serde` now

relay-log/src/setup.rs
```diff
@@ -47,53 +50,78 @@ pub enum LogFormat {
...
-        struct V;
-
-        impl<'de> Visitor<'de> for V {
-            type Value = Level;
-
-            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-                formatter.write_str("a log level")
-            }
+#[derive(Clone, Copy, Debug)]
+pub enum Level {
```
iker-barriocanal, line 68: nit(?)
```suggestion
pub enum LogLevel {
```

iker-barriocanal (approved):

relay-log/src/setup.rs
jjbayer: [Avoid redundant prefixes](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-%5Brfc-356%5D).

PR #3056, opened 2024-02-05
ref(normalization): Move user to normalization

iker-barriocanal:
Moves user normalization from store processing to normalization.

#skip-changelog

### DIFF
--- a/relay-cabi/src/processing.rs
+++ b/relay-cabi/src/processing.rs
@@ -85,11 +85,10 @@ pub unsafe extern "C" fn relay_valid_platforms(size_out: *mut usize) -> *const R
 #[relay_ffi::catch_unwind]
 pub unsafe extern "C" fn relay_store_normalizer_new(
     config: *const RelayStr,
-    geoip_lookup: *const RelayGeoIpLookup,
+    _geoip_lookup: *const RelayGeoIpLookup,
 ) -> *mut RelayStoreNormalizer {
     let config: StoreConfig = serde_json::from_str((*config).as_str())?;
-    let geoip_lookup = (geoip_lookup as *const GeoIpLookup).as_ref();
-    let normalizer = StoreProcessor::new(config, geoip_lookup);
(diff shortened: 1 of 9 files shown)


relay-cabi/src/processing.rs
```diff
@@ -85,11 +85,10 @@ pub unsafe extern "C" fn relay_valid_platforms(size_out: *mut usize) -> *const R
 #[relay_ffi::catch_unwind]
 pub unsafe extern "C" fn relay_store_normalizer_new(
     config: *const RelayStr,
-    geoip_lookup: *const RelayGeoIpLookup,
+    _geoip_lookup: *const RelayGeoIpLookup,
```
iker-barriocanal, line 88: GeoIP config was not used in the store processor: it only runs when `is_renormalize` is _not_ enabled, which should never be the case for calls from the C ABI. Removing this argument is a breaking change of the API, so I'm leaving it unused until `StoreNormalizer` is no longer needed so that we can remove this method and do a single breaking change.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -842,15 +817,21 @@ mod tests {
     #[test]
     fn test_geo_from_ip_address() {
```
iker-barriocanal, line 793: For now, I'm leaving geo-related tests in this module. I'll move them to a geo-specific module when I introduce it in a future PR.

relay-event-normalization/src/event.rs
```diff
@@ -368,6 +369,16 @@ pub fn normalize_user_geoinfo(geoip_lookup: &GeoIpLookup, user: &mut User) {
     }
 }
 
+fn normalize_user(user: &mut Annotated<User>) {
+    let Annotated(Some(user), _meta) = user else {
```
TBS1996:
```suggestion
    let Annotated(Some(user), _) = user else {
```

TBS1996 (approved):

PR #3058, opened 2024-02-06
fix(cardinality): Fix edge case where keys have no expiry

Dav1dde:
In the case with "perfect" cardinality, no additional values and below the limit, it was possible that new sets were correctly created but their expiry was never updated.

#skip-changelog

### DIFF
--- a/relay-cardinality/src/redis/cardinality.lua
+++ b/relay-cardinality/src/redis/cardinality.lua
@@ -86,7 +86,7 @@ local results = {
 --
 -- Returns the total amount of values added to the 'working set'.
 local function sadd(t, offset, max)
-    local working_set_cardinality = 0;
+    local working_set_cardinality = 0
     local any_modifications = false
 
     for i = 1, #KEYS do
--- a/relay-cardinality/src/redis/cardinality.lua
+++ b/relay-cardinality/src/redis/cardinality.lua
@@ -86,20 +86,23 @@ local results = {
 --
 -- Returns the total amount of values added to the 'working set'.
 local function sadd(t, offset, max)
-    local added = 0;
+    local working_set_cardinality = 0;
+    local any_modifications = false
 
     for i = 1, #KEYS do
         local is_working_set = i == 1
 
         for from, to in batches(#t, 7000, offset, max) do
             local r = redis.call('SADD', KEYS[i], unpack(t, from, to))
+
+            any_modifications = any_modifications or r > 0
             if is_working_set then
-                added = added + r
+                working_set_cardinality = working_set_cardinality + r
             end
         end
     end
 
-    return added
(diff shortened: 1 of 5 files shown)


relay-cardinality/src/redis/limiter.rs
```diff
@@ -320,7 +320,11 @@ mod tests {
         let url = std::env::var("RELAY_REDIS_URL")
             .unwrap_or_else(|_| "redis://127.0.0.1:6379".to_owned());
 
-        let redis = RedisPool::single(&url, RedisConfigOptions::default()).unwrap();
+        let opts = RedisConfigOptions {
+            max_connections: 1,
```
Dav1dde, line 324: Changed to `1` since the default is `24` and when re-running tests too quickly this caused some issues with my local Redis instance.

relay-cardinality/src/redis/cardinality.lua
```diff
@@ -86,20 +86,23 @@ local results = {
 --
 -- Returns the total amount of values added to the 'working set'.
 local function sadd(t, offset, max)
-    local added = 0;
+    local working_set_cardinality = 0;
```
jjbayer:
```suggestion
    local working_set_cardinality = 0
```

https://www.lua.org/pil/1.1.html

jjbayer (approved):

olksdr (approved):

relay-cardinality/src/redis/limiter.rs
jernejstrasner: Oh interesting. What issues?
Wouldn't this kind of negate the point of a pool? I mean, connection would still get recycled I guess but no parallelism.

Sorry for the noob questions :D

Dav1dde: With the default setting of 24 I got spurious failures when creating a new pool.

I assume the pool pre-allocates (up-to) `max_connections`, we create a new pool for every test, we have about 15 Redis related tests, which comes down to 360 connections per testrun. Something doesn't seem to be happy with creating/tearing down this amount of connections in a very short amount of time. Not sure what  the problem is. My Redis instance does have `maxclients` set to 10k, so in theory it should be fine.

Lowering the number fixed my issues so I didn't look further into why it started to fail sometimes.

jernejstrasner: The reason for me asking is that @anonrig pointed out to me yesterday that 8% of every request coming in is spent initiating a Redis connection so now I'm immediately interested when I hear Redis.

Dav1dde: If these are numbers from Relay, then there is something wrong. Relay should have a fixed pool of connections which only get recycled every 5 minutes.

Definitely interested in these numbers!

PR #3057, opened 2024-02-05
chore(spans): Upgrade sqlparser

jjbayer:
The `sqlparser` lib has been on a fork since https://github.com/getsentry/relay/pull/2846. We can now go back to the official version because https://github.com/sqlparser-rs/sqlparser-rs/pull/1065 has been merged & released.

ref: https://github.com/getsentry/team-ingest/issues/272

PR #3064, opened 2024-02-06
fix(spans): Catch panic when parsing SQL

jjbayer:
This PR reapplies #3057, which was [reverted](https://github.com/getsentry/relay/pull/3059) because of panics.

As a workaround, catch unwind panics so we can collect panic cases and report them upstream.

#skip-changelog (false positive)


relay-event-normalization/src/normalize/span/description/sql/mod.rs
```diff
@@ -876,13 +876,16 @@ mod tests {
...
-        // with regex
         r#"REPLACE INTO `foo` (`a`) VALUES ("abcd1234")"#,
-        "REPLACE INTO foo (a) VALUES (%s)"
+        "REPLACE INTO foo (..) VALUES (%s)"
+    );
+
+    scrub_sql_test!(
+        replace_into_set_dont_panic,
+        r#"REPLACE INTO `foo` SET x = y"#,
+        "REPLACE INTO foo SET x = y"
```
jjbayer, line 888: Without the `catch_unwind`, this test made `parse_query` panic.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -27,6 +27,25 @@ static TABLE_NAME_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i)[0-9a-f]{8,
...
     query: &str,
+) -> Result<Vec<Statement>, sqlparser::parser::ParserError> {
+    match std::panic::catch_unwind(|| parse_query_inner(db_system, query)) {
+        Ok(res) => res,
+        Err(_) => {
+            relay_log::error!(
+                tags.db_system = db_system,
+                query = query,
+                "parse_query panicked",
+            );
```
jjbayer, line 38: This will help us collect cases that we can then fix upstream.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -27,6 +27,25 @@ static TABLE_NAME_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i)[0-9a-f]{8,
 pub fn parse_query(
     db_system: Option<&str>,
     query: &str,
+) -> Result<Vec<Statement>, sqlparser::parser::ParserError> {
+    match std::panic::catch_unwind(|| parse_query_inner(db_system, query)) {
+        Ok(res) => res,
+        Err(_) => {
```
Dav1dde, line 33: Does the error not contain interesting information (maybe a stacktrace)?

Dav1dde (approved):

iker-barriocanal (approved):

olksdr (approved):
:shipit:

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: It probably does, but how do I extract it? The error type is `Box<dyn Any + Send>`.

Dav1dde: Actually misread the return type. It contains whatever was passed to `panic!()` (almost always some kind of string):

```rs
fn main() {
    if let Err(e) = std::panic::catch_unwind(|| panic!("test")) {
        println!("oops {:?}", e.downcast_ref::<&'static str>());
    }
    
    let r = 123;
    if let Err(e) = std::panic::catch_unwind(|| panic!("oops {r}")) {
        println!("oops {:?}", e.downcast_ref::<String>());
    }
}
```

So in our case it could be useful but most likely it's just a `unwrap()` somewhere.

PR #3065, opened 2024-02-06
ref(normalization): Move stacktrace to normalization

iker-barriocanal:
Moves stack trace normalization from StoreProcessor to normalization.

Currently, stack traces are normalized in a processor, so I've gone through all the references of `Stacktrace` and `RawStacktrace` to normalize them when needed. The `Exception` and `Thread` interfaces also contain a raw stack traces field that should not be normalized.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -12,9 +12,8 @@ use relay_base_schema::metrics::{
 };
 use relay_event_schema::processor::{self, MaxChars, ProcessingAction, ProcessingState, Processor};
 use relay_event_schema::protocol::{
-    AsPair, Context, ContextInner, Contexts, DeviceClass, Event, EventType, Exception, Headers,
-    IpAddr, Level, LogEntry, Measurement, Measurements, NelContext, Request, SpanStatus, Tags,
-    Timestamp, User,
+    AsPair, Context, ContextInner, Contexts, DeviceClass, Event, EventType, Headers, IpAddr, Level,
+    LogEntry, Measurement, Measurements, NelContext, Request, SpanStatus, Tags, Timestamp, User,
 };
 use relay_protocol::{Annotated, Empty, Error, ErrorKind, Meta, Object, Value};
 use smallvec::SmallVec;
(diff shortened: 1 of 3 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -504,35 +503,62 @@ fn normalize_device_class(event: &mut Event) {
...
-    match event.exceptions.value_mut() {
-        None => (),
-        Some(exception) => match exception.values.value_mut() {
-            None => (),
-            Some(exceptions) => match exceptions.first_mut() {
-                None => (),
-                Some(first) => normalize_last_stacktrace_frame(first),
-            },
-        },
+    normalize_event_stacktrace(event);
```
iker-barriocanal, line 522: I tried generating iterators of the stack traces to make their normalization generic (also useful in a future step), but the compiler is not happy as that'd result in 1+ mutable references. This alternative is the easier to read I could come up with.

relay-event-normalization/src/event.rs
```diff
@@ -504,35 +503,62 @@ fn normalize_device_class(event: &mut Event) {
     }
 }
 
-/// Process the last frame of the stacktrace of the first exception.
+/// Normalizes all the stack traces in the given event.
 ///
-/// No additional frames/stacktraces are normalized as they aren't required for metric extraction.
+/// Normalized stack traces are `event.stacktrace`, `event.exceptions.stacktrace`, and
+/// `event.thread.stacktrace`. Raw stack traces are not normalized.
```
iker-barriocanal, line 520: Raw stack traces are not processed, from the docstrings:

> Optional unprocessed stack trace.

relay-event-schema/src/protocol/types.rs
```diff
@@ -22,7 +22,7 @@ use crate::processor::{
     process_value, ProcessValue, ProcessingResult, ProcessingState, Processor, ValueType,
 };
 
-/// A array like wrapper used in various places.
+/// An array like wrapper used in various places.
```
TBS1996:
```suggestion
/// An array-like wrapper used in various places.
```  
my most pedantic comment yet

relay-event-normalization/src/event.rs
```diff
@@ -504,35 +503,62 @@ fn normalize_device_class(event: &mut Event) {
     }
 }
 
-/// Process the last frame of the stacktrace of the first exception.
+/// Normalizes all the stack traces in the given event.
```
TBS1996, line 517: so is this really a refactor ? this seems like a functional change

iker-barriocanal: Good question! Currently, light normalization normalizes the last frame of the stack trace of the first exception, and the rest of the stack trace is normalized in store processor. This PR unifies both steps, i.e. running normalization on all stack traces directly. You can see this in `normalize/mod.rs:548` below in the diff.

TBS1996: ah right i see. So I'm wondering why it was made like that in the first place, the comments said only one was needed for metric extraction but then we normalized all anyway? anyway, LGTM!

TBS1996 (approved):

relay-event-normalization/src/event.rs
iker-barriocanal: Only one is required for metrics extraction, so we normalized it before this happened. We still had to normalize all of them, so this normalization remained unchanged in the store processor.

PR #3068, opened 2024-02-07
ref(cardinality): Remove no longer necessary feature flag

Dav1dde:
Cardinality limits are now exposed via project config, and Sentry already does the feature flag check [there](https://github.com/getsentry/sentry/blob/master/src/sentry/relay/config/__init__.py#L204) -> we can remove the feature flag handling from Relay.

The feature flag was also only used by processing Relays, we don't have to consider compatibility here.

Should be merged together with https://github.com/getsentry/relay/pull/3067 but is not necessary.

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/feature.rs
+++ b/relay-dynamic-config/src/feature.rs
@@ -37,9 +37,6 @@ pub enum Feature {
     /// Enable metric metadata.
     #[serde(rename = "organizations:metric-meta")]
     MetricMeta,
-    /// Enable the Relay cardinality limiter.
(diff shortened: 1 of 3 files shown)


iker-barriocanal (approved):
> The feature flag was also only used by processing Relays

Can we remove the feature flag from `sentry` too?

Dav1dde:
@iker-barriocanal Yep, here https://github.com/getsentry/sentry/pull/64762 - You were just too fast with the review :D

olksdr (approved):

PR #3070, opened 2024-02-07
ref(normalization): Move debug image to normalization

iker-barriocanal:
Moves the normalization of debug images from `StoreNormalizeProcessor` to normalization. This brings us one step closer to getting rid of the store processor.

#skip-changelog


relay-event-normalization/src/event.rs
```diff
@@ -385,6 +386,26 @@ fn normalize_logentry(logentry: &mut Annotated<LogEntry>, _meta: &mut Meta) {
...
 }
 
+/// Normalizes the debug images in the event's debug meta.
+fn normalize_debug_meta(event: &mut Event) {
+    let Annotated(Some(debug_meta), _) = &mut event.debug_meta else {
+        return;
+    };
+    let Annotated(Some(debug_images), _) = &mut debug_meta.images else {
+        return;
+    };
```
olksdr, line 399: Maybe we can do it in one go here, destructure something like this 

```suggestion
        let Annotated(
        Some(DebugMeta {
            images: Annotated(Some(debug_images), _),
            ..
        }),
        _,
    ) = &mut event.debug_meta
    else {
        return;
    };

```

Not sure if this looks better though.. Maybe to `if` is better.

olksdr (approved):

relay-event-normalization/src/event.rs
iker-barriocanal: I find too many nested types in the suggestion, and formatting doesn't help either. I'm not sure if I've been looking into the same `let..else` pattern too much this couple of days so I may be biased though. I'm going to leave the `let..else` unless somebody feels strongly opinionated.

PR #3072, opened 2024-02-07
ref(normalization): Move exception to normalization

iker-barriocanal:
Moves the normalization of exceptions from StoreProcessor to normalization. This brings us one step closer to removing StoreProcessor.

#skip-changelog


relay-event-normalization/src/normalize/mod.rs
```diff
@@ -716,79 +680,6 @@ mod tests {
         assert_eq!(dynamic_config.max_custom_measurements().unwrap(), 3);
     }
 
-    #[test]
-    fn test_handles_type_in_value() {
```
iker-barriocanal, line 674: These tests have been moved to `event.rs`.

This one in particular, `test_handles_type_in_value`, was testing two things, so I split it into `test_handle_types_in_spaced_exception_values` and `test_handle_types_in_non_spaced_excepton_values`.

relay-event-normalization/src/event.rs
```diff
@@ -2515,4 +2548,75 @@ mod tests {
 
         assert_eq!(user.other, Object::new());
     }
+
```
olksdr, line 2578: Are there any tests for `normalize_exceptions` function which runs on the entire `Event`?  So we could cover the entire normalization step. I can't see any. 

Can some be added as well in this PR?

iker-barriocanal: Yep yep, added in [7738516](https://github.com/getsentry/relay/pull/3072/commits/7738516faf7616524a873160509feb826a629a5c).

olksdr (approved):

PR #3073, opened 2024-02-07
ref(normalization): Move frames to normalization

iker-barriocanal:
Move frame normalization from StoreNormalization to normalization. This brings us one step closer to removing StoreNormalization.

This PR also changes the name `process_stacktrace` -> `normalize_stacktrace`. "Process" has a specific meaning with processors, which are not used here and could generate some confusion, and "normalize" aligns with the naming used in all normalization steps.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/stacktrace.rs
+++ b/relay-event-normalization/src/stacktrace.rs
@@ -12,15 +12,16 @@ fn is_url(filename: &str) -> bool {
         || filename.starts_with("applewebdata:")
 }
 
-pub fn process_stacktrace(stacktrace: &mut RawStacktrace, _meta: &mut Meta) {
+pub fn normalize_stacktrace(stacktrace: &mut RawStacktrace, _meta: &mut Meta) {
     // This processing is only done for non raw frames (i.e. not for exception.raw_stacktrace).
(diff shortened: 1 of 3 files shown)


relay-event-normalization/src/stacktrace.rs
```diff
@@ -41,6 +42,33 @@ pub fn process_non_raw_frame(frame: &mut Annotated<Frame>) {
                 }
             }
         }
+
+        if frame.function.as_str() == Some("?") {
```
iker-barriocanal, line 46: The order is the same as in the processor: process the stack trace first (code above this +diff), then process frames (this +diff).

jjbayer: The call to `process_child_values` is gone here though. Are we sure that was a noop?

iker-barriocanal: There are no `process_*` calls for a `Frame`'s properties, so these aren't further normalized.

jjbayer (approved):
@iker-barriocanal light normalization runs before inbound filters. Some events have a lot of frames so this could be a heavy step. We should keep a close eye on the `EventProcessingLightNormalization` metric after deploying this.

PR #3076, opened 2024-02-08
fix(cardinality): Keep sliding window updated with good cardinality items

Dav1dde:
After checking for membership items were not added to new granules of the sliding window:

```
[ ][ ][ ]{ }{ }{ }[ ][ ][ ] - Start
[A][A][A]{ }{ }{ }[ ][ ][ ] - Insert A
   [A][A]{A}{ }{ }[ ][ ][ ] - Time passed, refresh A
          \-- This value was missing before.
```

#skip-changelog

### DIFF
--- a/relay-cardinality/src/redis/cardinality.lua
+++ b/relay-cardinality/src/redis/cardinality.lua
@@ -85,11 +85,16 @@ local results = {
 -- to all sets in `KEYS`.
 --
 -- Returns the total amount of values added to the 'working set'.
-local function sadd(t, offset, max)
+local function sadd(t, opts)
     local working_set_cardinality = 0
     local any_modifications = false
 
-    for i = 1, #KEYS do
+    opts = opts or {}
+    local key_offset = opts.skip_working_set and 1 or 0
+    local offset = opts.offset or 0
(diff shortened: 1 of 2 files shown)


relay-cardinality/src/redis/cardinality.lua
```diff
@@ -85,11 +85,16 @@ local results = {
...
-local function sadd(t, offset, max)
+local function sadd(t, opts)
     local working_set_cardinality = 0
     local any_modifications = false
 
-    for i = 1, #KEYS do
+    opts = opts or {}
+    local key_offset = opts.skip_working_set and 1 or 0
+    local offset = opts.offset or 0
+    local max = opts.max
```
olksdr, line 95: do you also need default for `max` here ? Seems like it's not always provided.

olksdr (approved):

relay-cardinality/src/redis/cardinality.lua
Dav1dde: It's a bit of a mess, but `max` being `nil` is fine, it's checked in the batches function and has a different behaviour if it's not set.

PR #3052, opened 2024-02-05
feat(processor): Introduce initial type system

olksdr:
Add initial interfaces (public contract) to processing state. 

@Dav1dde is this something you had in mind?  See `processing/state.rs` module. 

related: https://github.com/getsentry/team-ingest/issues/253

#skip-changelog

PR #3078, opened 2024-02-09
feat(processor): Add initial types into processing pipeline

olksdr:
This PR puts types restrictions on some functions, which makes it harder to get wrong.  There is no functional changes over here. The code that was moved around wasn't changed, it's just to make it easier to create the current state in more control manner. 

This is just initial setup, which should evolve once we move to `TypedEnvelope`  (will be described in https://github.com/getsentry/team-ingest/issues/281)

Supersedes https://github.com/getsentry/relay/pull/3052 , which was a test PR to see how different approach would look like, where we also can come back later.

Related to https://github.com/getsentry/team-ingest/issues/253

#skip-changelog


relay-server/src/services/processor.rs
```diff
@@ -1274,55 +1334,222 @@ impl EnvelopeProcessorService {
...
-            ProcessingGroup::Span => self.process_spans(state)?,
+            ProcessingGroup::Error(ErrorGroup) => {
+                let mut state = self.prepare_state::<ErrorGroup>(
+                    managed_envelope,
+                    project_id,
+                    project_state,
+                    sampling_project_state,
+                    reservoir_counters,
+                );
+                match self.process_errors(&mut state) {
```
Dav1dde: This is very repetitive, should be able to either do the conversion via `macro_rules` or via `map` and `map_err`.

Skteched up:

```rs
        macro_rules! process {
            ($f:ident) => {{
                let mut state = self.prepare_state(
                    managed_envelope,
                    project_id,
                    project_state,
                    sampling_project_state,
                    reservoir_counters,
                );
                match self.$f(&mut state) {
                    Ok(()) => Ok(ProcessingStateResult {
                        managed_envelope: state.managed_envelope,
                        extracted_metrics: state.extracted_metrics,
                    }),
                    Err(e) => Err((e, state.managed_envelope)),
                }
            }}
        }

        match group {
            ProcessingGroup::Error(_) => process!(process_errors),
            ProcessingGroup::Transaction(_) => process!(process_transactions),
            ...
        }
```

relay-server/src/services/processor.rs
```diff
@@ -101,38 +103,82 @@ macro_rules! if_processing {
...
+pub struct Ungrouped;
+
 /// Describes the groups of the processable items.
 #[derive(Clone, Copy, Debug)]
 pub enum ProcessingGroup {
     /// All the transaction related items.
     ///
     /// Includes transactions, related attachments, profiles.
-    Transaction,
+    Transaction(TransactionGroup),
```
Dav1dde, line 156: I don't think this does anything for us now, do you just want to have it here for the future?

Dav1dde (approved):
👍

relay-server/src/services/processor.rs
olksdr: Yeah! 
the `map` and `map_err` does not work because of the borrow checker, but `macro_rules` seems good to me

relay-server/src/services/processor.rs
olksdr: Yeah, it's the preparation for the followups

relay-server/src/services/processor.rs
```diff
@@ -1041,9 +1083,9 @@ impl EnvelopeProcessorService {
         Ok(())
     }
 
-    fn light_normalize_event(
+    fn light_normalize_event<G>(
```
jjbayer, line 1086: nit: We could already introduce marker traits here (without any methods), e.g. `EventGroup` implemented by  `TransactionGroup` and `ErrorGroup`.

```suggestion
    fn light_normalize_event<G: EventGroup>(
```

jjbayer (approved):

relay-server/src/services/processor.rs
olksdr: yap, that what I was mentioning today in the morning.  I can also do in in the followups, to have fewer changes in this PR.

Dav1dde: I'd prefer follow up 👍

PR #3081, opened 2024-02-09
ref(spans): Parse SQL only once and deduplicate errors

jjbayer:
An unparsable SQL query in a span description currently causes four sentry errors:

* A panic and a manual error when we parse to scrub.
* A panic and a manual error when we parse to extract table names.

This PR

- makes `sql_tables_from_query` reuse the parsed SQL query from `scrub_span_description`,
- removes the manual error, setting necessary tags & data on the sentry scope instead.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/sql/parser.rs
+++ b/relay-event-normalization/src/normalize/span/description/sql/parser.rs
@@ -28,19 +28,18 @@ pub fn parse_query(
     db_system: Option<&str>,
     query: &str,
 ) -> Result<Vec<Statement>, sqlparser::parser::ParserError> {
-    match std::panic::catch_unwind(|| parse_query_inner(db_system, query)) {
-        Ok(res) => res,
-        Err(_) => {
-            relay_log::error!(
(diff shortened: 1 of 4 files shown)


relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -58,8 +57,14 @@ fn parse_query_inner(
 }
 
 /// Tries to parse a series of SQL queries into an AST and normalize it.
-pub fn normalize_parsed_queries(db_system: Option<&str>, string: &str) -> Result<String, ()> {
+///
+///
```
olksdr: Are there supposed to be docs here ?

olksdr (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: Thanks, removing these.

```suggestion
```

Dav1dde (approved):

PR #3086, opened 2024-02-12
feat(global): Add quotas to global config

TBS1996:
https://github.com/getsentry/relay/issues/2716

This PR allows us to dynamically add quotas through global config.


relay-server/src/services/project.rs
```diff
@@ -692,39 +692,41 @@ impl Project {
...
-                            envelope_processor,
-                            outcome_aggregator,
-                        );
-                    }
-                }
-                State::Pending(inner_agg) => {
-                    // We need to queue the metrics in a temporary aggregator until the project state becomes available.
-                    relay_log::debug!("sending metrics to metrics-buffer");
-                    inner_agg.merge_all(self.project_key, buckets, None);
+        if !self.metrics_allowed() {
```
TBS1996, line 695: removed ~two~ one layers of identations with early returns. imo makes it easier to read and also is in line with how it's done in other places.

relay-dynamic-config/src/global.rs
```diff
@@ -117,13 +121,53 @@ fn is_default<T: Default + PartialEq>(t: &T) -> bool {
     *t == T::default()
 }
 
+/// Container for global and project level [`Quota`].
```
TBS1996: followed example of `DynamicMeasurementsConfig`.

relay-quotas/src/redis.rs
```diff
@@ -211,9 +211,9 @@ impl RedisRateLimiter {
     /// The passed `quantity` may be `0`. In this case, the rate limiter will check if the quota
     /// limit has been reached or exceeded without incrementing it in the success case. This can be
     /// useful to check for required quotas in a different data category.
-    pub fn is_rate_limited(
+    pub fn is_rate_limited<'a>(
         &self,
-        quotas: &[Quota],
+        quotas: impl Iterator<Item = &'a Quota>,
```
Dav1dde:
```suggestion
        quotas: impl IntoIterator<Item = &'a Quota>,
```

Means you can still pass in a slice directly and usually what you do when you want to accept anything that can yield items.

relay-dynamic-config/src/global.rs
```diff
@@ -117,13 +121,53 @@ fn is_default<T: Default + PartialEq>(t: &T) -> bool {
...
+pub struct DynamicQuotas<'a> {
```
Dav1dde: I'd move this into the `EnvelopeProcessor` or one of its associated utility modules.

Also rename to something like `CombinedQuotas`, I don't really know what `Dynamic` means here.

relay-dynamic-config/src/global.rs
```diff
@@ -117,13 +121,53 @@ fn is_default<T: Default + PartialEq>(t: &T) -> bool {
...
+    /// Returns a new [`DynamicQuotas`]
+    pub fn new(global_config: &'a GlobalConfig, quotas: &'a [Quota]) -> Self {
+        Self {
+            global_quotas: &global_config.quotas,
+            project_quotas: quotas,
+        }
+    }
+
+    /// Returns the combination of global and project quotas.
+    pub fn iter(&self) -> impl Iterator<Item = &'a Quota> {
```
Dav1dde: This should implement `IntoIterator` instead and it can be `Copy` (though if useful we can also keep the `iter()` around which would just forward into `self.into_iter()`)

relay-dynamic-config/src/global.rs
```diff
@@ -117,13 +121,53 @@ fn is_default<T: Default + PartialEq>(t: &T) -> bool {
...
 
+/// Container for global and project level [`Quota`].
+pub struct DynamicQuotas<'a> {
+    global_quotas: &'a [Quota],
+    project_quotas: &'a [Quota],
+}
+
+impl<'a> DynamicQuotas<'a> {
+    /// Returns a new [`DynamicQuotas`]
+    pub fn new(global_config: &'a GlobalConfig, quotas: &'a [Quota]) -> Self {
```
Dav1dde: After moving, this can take a reference to the global config and the project state.

relay-dynamic-config/src/global.rs
```diff
@@ -117,13 +121,53 @@ fn is_default<T: Default + PartialEq>(t: &T) -> bool {
...
+        }
+    }
+
+    /// Returns the combination of global and project quotas.
+    pub fn iter(&self) -> impl Iterator<Item = &'a Quota> {
+        self.global_quotas.iter().chain(self.project_quotas)
+    }
+
+    /// Returns `true` if both global quotas and project quotas are empty.
+    pub fn is_empty(&self) -> bool {
```
Dav1dde: `is_empty` usually comes with `len`

relay-server/src/services/processor.rs
```diff
@@ -1839,20 +1844,26 @@ impl EnvelopeProcessorService {
...
             })
             .flatten()
             .collect()
     }
 
     /// Returns `true` if the batches should be rate limited.
     #[cfg(feature = "processing")]
-    fn rate_limit_buckets(
-        &self,
+    fn rate_limit_buckets<'a>(
```
Dav1dde: Don't think this is necessary

relay-server/src/services/project.rs
```diff
@@ -692,39 +692,41 @@ impl Project {
...
     ) {
-        if self.metrics_allowed() {
-            match &mut self.state {
-                State::Cached(state) => {
-                    let state = Arc::clone(state);
-
-                    let buckets =
-                        self.rate_limit_metrics(&state, buckets, outcome_aggregator.clone());
-
-                    if !buckets.is_empty() {
```
Dav1dde, line 703: I would keep this. Returns in already (deeply) nested paths mean no other later code is run. E.g. if we add code after the match statement (logging, other logic) this wouldn't run anymore. 

So in order to understand if this return is correct I first need to check the rest of the function and evaluate if this is still correct behaviour.

relay-server/src/services/processor.rs
```diff
@@ -1814,11 +1819,11 @@ impl EnvelopeProcessorService {
...
 
     #[cfg(feature = "processing")]
-    fn rate_limit_buckets_by_namespace(
-        &self,
+    fn rate_limit_buckets_by_namespace<'a>(
+        &'a self,
         scoping: Scoping,
         buckets: Vec<Bucket>,
-        quotas: &[relay_quotas::Quota],
+        dynamic_quotas: DynamicQuotas<'a>,
```
```suggestion
        quotas: DynamicQuotas<'a>,
```

relay-dynamic-config/src/global.rs
TBS1996: well dynamic refers to the [dynamic config](https://github.com/getsentry/relay/blob/ca1612de2c2d29257b619774f68df20bbe705bad/relay-dynamic-config/src/lib.rs) crate, which is also how we named the `DynamicMeasurementConfig`. I Agree it sounds a bit confusing but we'll have more combined types like this in the future and then it'll be more clear to people reading that dynamic_x refers to combined global and project types. 
 
 
I'm open to changing ofc but then we should probably have a discussion on naming conventions regarding global config types in general

TBS1996: about where to put it, this is also something where we should have a precedent. To me it makes sense to either put it where quotas are defined, or keep it in the dynamic config crate. I prefer the latter tbh, like maybe making a `quotas.rs` there. thoughts?   
 
`processor.rs` seems too big already imo. if we wanna move it to relay-server then maybe to the global_config module you made there?

relay-dynamic-config/src/global.rs
TBS1996: added it now, although i thought in general we don't add helper's before we need them

relay-server/src/services/processor.rs
TBS1996: the lifetime?

otherwise i get this: `anonymous lifetimes in 'impl Trait' are unstable`

relay-dynamic-config/src/global.rs
Dav1dde: My thought is what I wrote, it should go where we use it with a name people understand.

We don't even need it in the first place, it is just a wrapper for a single iterator operation.

relay-dynamic-config/src/global.rs
Dav1dde: There is a [clippy lint](https://rust-lang.github.io/rust-clippy/master/#/len_without_is_empty) for the opposite ...

relay-server/src/services/processor.rs
Dav1dde: Now you do, it used to be a concrete type, which it still should be here to prevent mistakes, at this point we always want combined quotas, the type system should enforce that.

relay-dynamic-config/src/global.rs
Dav1dde: There is a [clippy lint](https://rust-lang.github.io/rust-clippy/master/#/len_without_is_empty) for the opposite ...

relay-dynamic-config/src/global.rs
Dav1dde: My thought is what I wrote, it should go where we use it (the module which actually cares about which quotas to pass along) with a name people understand.

We don't even need it in the first place, it is just a wrapper for a single iterator operation. The upside is we can enforce in the processor via the type system that the quotas have been combined.

relay-dynamic-config/src/global.rs
TBS1996: problem is, in some cases we get the quotas from `BucketLimiter` 

``` Rust
    /// Check and apply rate limits to metrics buckets.
    fn handle_rate_limit_buckets(&self, message: RateLimitBuckets) {
        let RateLimitBuckets { mut bucket_limiter } = message;

        let scoping = *bucket_limiter.scoping();

        if let Some(rate_limiter) = self.inner.rate_limiter.as_ref() {
            let item_scoping = relay_quotas::ItemScoping {
                category: DataCategory::Transaction,
                scoping: &scoping,
                namespace: None,
            };

            let global_config = self.inner.global_config.current();
            let quotas = DynamicQuotas::new(&global_config, bucket_limiter.quotas());

...
```

relay-dynamic-config/src/global.rs
TBS1996: moved to processor.rs

relay-quotas/src/redis.rs
```diff
@@ -603,7 +603,7 @@ mod tests {
         };
 
         let rate_limits: Vec<RateLimit> = build_rate_limiter()
-            .is_rate_limited(&[], scoping, 1, false)
+            .is_rate_limited([].iter(), scoping, 1, false)
```
Dav1dde: Seems to compile without change (`[]` also works).

relay-server/src/services/processor.rs
```diff
@@ -1880,7 +1886,7 @@ impl EnvelopeProcessorService {
         &self,
         scoping: Scoping,
         buckets: Vec<Bucket>,
-        quotas: &[relay_quotas::Quota],
+        quotas: DynamicQuotas,
```
Dav1dde: micro nit, I like having the lifetime explicitly listed:
```suggestion
        quotas: DynamicQuotas<'_>,
```

relay-server/src/services/processor.rs
```diff
@@ -2645,6 +2649,43 @@ impl UpstreamRequest for SendMetricsRequest {
...
+        self.len() == 0
+    }
+
+    /// Returns the number of both global and project quotas.
+    pub fn len(&self) -> usize {
+        self.global_quotas.len() + self.project_quotas.len()
+    }
+}
+
+// Implementing IntoIterator for &DynamicQuotas to allow iterating over the quotas.
```
Dav1dde: It's implemented for `DynamicQuotas` (not the reference), but imo we can also drop the comment, the trait is pretty self descriptive.

relay-server/src/services/processor.rs
```diff
@@ -2645,6 +2649,43 @@ impl UpstreamRequest for SendMetricsRequest {
     }
 }
 
+/// Container for global and project level [`Quota`].
+#[derive(Copy, Clone)]
+pub struct DynamicQuotas<'a> {
```
Dav1dde: Wouldn't make this pub, it's just a helper for the processor.

Also the reason I thought we want to change the name to something more descriptive, but the argument that `Dynamic` is a convention for this sort of thing is good 👍 .

relay-server/src/services/processor.rs
```diff
@@ -2677,6 +2719,36 @@ mod tests {
...
+            measurements: None,
+            quotas: vec![mock_quota("foo"), mock_quota("bar")],
+            options: Options::default(),
+        };
+
+        let project_quotas = vec![mock_quota("baz"), mock_quota("qux")];
+
+        let dynamic_quotas = DynamicQuotas::new(&global_config, &project_quotas);
+
+        for (expected_id, quota) in ["foo", "bar", "baz", "qux"].iter().zip(dynamic_quotas) {
```
```rs
        let quota_ids = dynamic_quotas.into_iter().filter_map(|q| q.id.as_deref());
        assert!(quota_ids.eq(["foo", "bar", "baz", "qux"]));
```

relay-server/src/services/processor.rs
```diff
@@ -2677,6 +2719,36 @@ mod tests {
...
+    fn test_dynamic_quotas() {
+        let global_config = GlobalConfig {
+            measurements: None,
+            quotas: vec![mock_quota("foo"), mock_quota("bar")],
+            options: Options::default(),
+        };
+
+        let project_quotas = vec![mock_quota("baz"), mock_quota("qux")];
+
+        let dynamic_quotas = DynamicQuotas::new(&global_config, &project_quotas);
```
Dav1dde, line 2749: Might as well test `len` and `is_empty` here, also maybe a separate test which actually checks for the empty case could be appropiate.

relay-server/src/services/processor.rs
```diff
@@ -1914,7 +1920,7 @@ impl EnvelopeProcessorService {
         &self,
         item_scoping: relay_quotas::ItemScoping,
         buckets: &[Bucket],
-        quotas: &[relay_quotas::Quota],
+        quotas: DynamicQuotas,
```
Dav1dde: Same micro nit:

```suggestion
        quotas: DynamicQuotas<'_>,
```

relay-server/src/services/processor.rs
```diff
@@ -2041,13 +2049,9 @@ impl EnvelopeProcessorService {
             let limits = project_state.get_cardinality_limits();
 
             let buckets = self.cardinality_limit_buckets(scoping, limits, buckets, mode);
+            let quotas = DynamicQuotas::new(&global_config, &project_state.config.quotas);
```
```suggestion
            let quotas = DynamicQuotas::new(&global_config, project_state.get_quotas());
```

You use `get_quotas` elsewhere, might as well do it here.

Dav1dde (approved):
Mostly tiny nits 👍 

(and it doesn't compile, but that is an easy fix)

relay-server/src/services/processor.rs
```diff
@@ -2658,25 +2658,24 @@ pub struct DynamicQuotas<'a> {
 
 impl<'a> DynamicQuotas<'a> {
     /// Returns a new [`DynamicQuotas`]
-    pub fn new(global_config: &'a GlobalConfig, project_quotas: &'a [Quota]) -> Self {
```
Dav1dde: These can stay public, just the type shouldn't be public (so no one can import it and use it by accident).

relay-server/src/services/processor.rs
```diff
@@ -2669,14 +2710,61 @@ mod tests {
...
+        let global_config = GlobalConfig {
+            measurements: None,
+            quotas: vec![mock_quota("foo"), mock_quota("bar")],
+            options: Options::default(),
+        };
+
+        let quotas = vec![mock_quota("baz")];
+
+        let dynq = DynamicQuotas::new(&global_config, &quotas);
+        assert_eq!(dynq.len(), 3);
```
Dav1dde: I think it's fine if we just assert the `len()` and `is_empty()` methods in `test_dynamic_quotas`:

```rs
assert_eq!(foo.len(), 5);
assert!(!foo.is_empty());
```

relay-server/src/services/processor.rs
```diff
@@ -2669,14 +2710,61 @@ mod tests {
...
+            reason_code: None,
+            namespace: None,
+        }
+    }
+
+    #[test]
+    fn test_dynamic_quota_empty() {
+        let glob_config = GlobalConfig::default();
+        let dynq = DynamicQuotas::new(&glob_config, &[]);
+        assert!(dynq.is_empty());
```
Dav1dde: Should also assert length here

relay-server/src/services/processor.rs
TBS1996: hm, how can the constructors be public but the type not? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9f89537d6f8d9660062c1df5b6eec9e5

relay-server/src/services/processor.rs
TBS1996: how can the type be private and its constructor not ? 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=537ebc15e46838757d1bb0295f59f73a#:~:text=Permalink%20to%20the%20playground

see comp error

Dav1dde: It's not used outside of this module (and that's what we want to prevent): https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9c751fa63c54def6ced6aeb184c6235b

Dav1dde: Constructor does not have to be public for this either, but I like still to keep the visibility modifier as a way of signalling "this function is intended to be called from outside the struct" vs "this method is an implementation detail and probably shouldnt be called".

TBS1996: alright, that makes sense to me, thanks!

PR #3089, opened 2024-02-12
ref(global-config): Don't fail on deserialization issues in global config options

Dav1dde:
Currently one broken option will take down Relay, and it is way too easy to accidentally configure the wrong option in the options automator.

This PR changes the behaviour to just use a default value when deserialization fails and logs an error.

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -19,7 +19,10 @@ pub struct GlobalConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub measurements: Option<MeasurementsConfig>,
     /// Sentry options passed down to Relay.
-    #[serde(skip_serializing_if = "is_default")]
+    #[serde(
+        deserialize_with = "default_on_error",
+        skip_serializing_if = "is_default"
+    )]
     pub options: Options,
 }
 
@@ -49,6 +52,7 @@ pub struct Options {
     /// of improving profile (function) metrics
     #[serde(
         rename = "profiling.profile_metrics.unsampled_profiles.platforms",
+        deserialize_with = "default_on_error",
         skip_serializing_if = "Vec::is_empty"
     )]
     pub profile_metrics_allowed_platforms: Vec<String>,
@@ -56,6 +60,7 @@ pub struct Options {
     /// Sample rate for tuning the amount of unsampled profiles that we "let through"
     #[serde(
         rename = "profiling.profile_metrics.unsampled_profiles.sample_rate",
+        deserialize_with = "default_on_error",
         skip_serializing_if = "is_default"
     )]
     pub profile_metrics_sample_rate: f32,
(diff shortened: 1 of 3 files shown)


TBS1996 (approved):
lgtm

relay-dynamic-config/src/global.rs
```diff
@@ -117,6 +129,24 @@ fn is_default<T: Default + PartialEq>(t: &T) -> bool {
...
+fn default_on_error<'de, D, T>(deserializer: D) -> Result<T, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+    T: Default + serde::de::DeserializeOwned,
+{
+    match T::deserialize(deserializer) {
+        Ok(value) => Ok(value),
+        Err(error) => {
+            relay_log::error!(
+                error = %error,
```
Dav1dde, line 141: Can't use `&error as &dyn Error` here because of lifetime issues, the error can borrow from `'de` and the macro does not accept errors that aren't bound by `'static`.

I also can't bound `D::Error: 'static` because that breaks the derive.

olksdr (commented):
My only question , how good are we will the defaults for those options? 

E.g. `profile_metrics_allowed_platforms` if that defaults to empty `Vec` - does it mean we allow everything through or nothing ?

Dav1dde:
@olksdr so the idea of these options was that they should always work even if they are missing/default. E.g. could be missing because someone is using an older Sentry version.

I assume for the `profile_metrics_allowed_platforms` key, this means the feature is disabled and since this was a recently added I think that is what is expected.

olksdr:
> @olksdr so the idea of these options was that they should always work even if they are missing/default. E.g. could be missing because someone is using an older Sentry version.
> 
> I assume for the `profile_metrics_allowed_platforms` key, this means the feature is disabled and since this was a recently added I think that is what is expected.

Thanks! 
Was just wondering, since there is not much info in the docs on those options.

relay-dynamic-config/src/global.rs
```diff
@@ -117,6 +129,24 @@ fn is_default<T: Default + PartialEq>(t: &T) -> bool {
...
+                "Error deserializing global config option: {}",
+                std::any::type_name::<T>(),
```
iker-barriocanal, line 143: Is there a way to know the field name, to make the error provide more info? Example:
```
// Current
"Error deserializing global config option: bool
// Alternative
"Error deserializing global config option: span_usage_metric"
```

iker-barriocanal (approved):

relay-dynamic-config/src/global.rs
Dav1dde: Only by deserializing with [serde_path_to_error](https://docs.rs/serde_path_to_error) then the error would contain the key. The deserializer here is not aware of its context

iker-barriocanal: It should be good enough to not introduce another dependency, ship it!

PR #3090, opened 2024-02-12
feat(quotas): Add ability to use namespace in non-global quotas.

TBS1996:
https://github.com/getsentry/relay/issues/2716

With this PR, we can also use namespace in non-global quotas. 

The reason for this PR was the requirement to be able to ratelimit based on namespace per org, but in reality it's pretty flexible so we can do namespace per project as well.

### DIFF
--- a/relay-quotas/src/redis.rs
+++ b/relay-quotas/src/redis.rs
@@ -388,14 +388,15 @@ mod tests {
     /// Tests that a quota with and without namespace are counted separately.
     #[test]
     fn test_non_global_namespace_quota() {
+        let quota_limit = 5;
         let get_quota = |namespace: Option<MetricNamespace>| -> Quota {
             Quota {
                 id: Some(format!("test_simple_quota_{}", uuid::Uuid::new_v4())),
                 categories: DataCategories::new(),
                 scope: QuotaScope::Organization,
                 scope_id: None,
-                limit: Some(5),
-                window: Some(60),
+                limit: Some(quota_limit),
+                window: Some(600),
                 reason_code: Some(ReasonCode::new(format!("ns: {:?}", namespace))),
                 namespace,
             }
@@ -417,6 +418,7 @@ mod tests {
 
         let rate_limiter = build_rate_limiter();
 
+        // First confirm normal behaviour without namespace.
         for i in 0..10 {
             let rate_limits: Vec<RateLimit> = rate_limiter
                 .is_rate_limited(quotas, scoping, 1, false)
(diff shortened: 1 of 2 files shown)


Dav1dde (approved):

jjbayer (approved):

relay-quotas/src/redis.rs
```diff
@@ -383,6 +385,73 @@ mod tests {
...
+                project_id: ProjectId::new(43),
+                project_key: ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap(),
+                key_id: Some(44),
+            },
+            namespace: Some(MetricNamespace::Transactions),
+        };
+
+        let rate_limiter = build_rate_limiter();
+
+        for i in 0..10 {
```
iker-barriocanal, line 422: You're effectively having two loops in one, and splitting it increases readability:

```rs
for i in 0..5 {
  assert_eq!(
    rate_limits[0].reason_code,
    Some(ReasonCode::new("ns: None"))
  );
}
for i in 0..5 {
  assert_eq!(rate_limits, vec![]);
}
```
(Same with the loop below)

relay-quotas/src/redis.rs
```diff
@@ -383,6 +385,73 @@ mod tests {
...
+                assert_eq!(
+                    rate_limits[0].reason_code,
+                    Some(ReasonCode::new("ns: None"))
+                );
+            } else {
+                assert_eq!(rate_limits, vec![]);
+            }
+        }
+
+        for i in 0..10 {
```
iker-barriocanal, line 440: These two loops are asserting validating two different features. Could we split them into two separate tests, as per [Relay guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#fd14645580f848d9969d124a59cc3f7d) indicate?

> Keep unit tests constrained to a single concern. Do not test multiple things in one test, do not test sequences of actions. Write multiple tests for each of the concerns.

This makes identifying the root cause easier on test failures.

iker-barriocanal (approved):
LGTM. Please address the suggestions below before merging.

relay-quotas/src/redis.rs
TBS1996: so I don't agree here, the two loops are there two assert that with and without namespace is counted differently, this is only possible if we have them both in the same test to ensure theyre both connected to the same redis instance. if you look below this test at `fn test_simple_quota` then you see the original test that checked the behaviour without namespace

relay-quotas/src/redis.rs
TBS1996: hmm i get your point but it also makes it a bit less clear the separation of the loop with and without the namespace, and we'd also have to duplicate the let ratelimits = ... block more.

I agree it's not as clear as can be though, I uploaded some changes. 

1. added comments
2. the 'quota_limit' variable should be more clear
3. flipped the if/else statements

iker-barriocanal: This is probably my personal opinion, and not going to block the PR for that, but this is one case in which I'm ok with some code duplication given I find less cognitive load and thus more readable.

iker-barriocanal (approved):

PR #3091, opened 2024-02-12
fix(spans): Scrub RELEASE SAVEPOINT statements

### LINKED ISSUE #3085: `RELEASE SAVEPOINT` and `ROLLBACK TO SAVEPOINT` SQL statements aren't scrubbed
The `sqlparser` upgrade broke the scrubbing of `RELEASE SAVEPOINT` and `ROLLBACK TO SAVEPOINT` SQL statements. Though even on the old versions those statements had some scrubbing issues. 

See also: https://github.com/getsentry/relay/pull/3084

Savepoint queries were disabled in #3083 as a quick fix.

jjbayer:
sqlparser improved parsing capabilities, which made us rely on the fallback scrubber less. Unfortunately, In the case of savepoints, the fallback scrubber caught more than the parser-based scrubber.

Reverts https://github.com/getsentry/relay/pull/3083
Fixes https://github.com/getsentry/relay/issues/3085

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -29,11 +29,8 @@ const MOBILE_OPS: &[&str] = &[
 /// A list of span descriptions that indicate top-level app start spans.
 const APP_START_ROOT_SPAN_DESCRIPTIONS: &[&str] = &["Cold Start", "Warm Start"];
 
-/// A list of patterns found in MongoDB queries.
+/// A list of patterns found in MongoDB queries
 const MONGODB_QUERIES: &[&str] = &["*\"$*", "{*", "*({*", "*[{*"];
-
-/// A list of PG queries we don't want to support.
-const DISABLE_SOME_PG_QUERIES: &[&str] = &["*SAVEPOINT*"];
 
 /// A list of patterns for resource span ops we'd like to ingest.
 const RESOURCE_SPAN_OPS: &[&str] = &["resource.script", "resource.css", "resource.img"];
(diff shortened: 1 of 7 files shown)


relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -492,7 +497,6 @@ impl VisitorMut for NormalizeVisitor {
                     }
                 }
             }
-
             _ => {}
```
Dav1dde: Can we get rid of this catch all, if in the future the library introduces new values to the enum it won't compile and we know we have to check something?

jjbayer: We absolutely should. I already started doing it and then gave up because there are so many match arms. But I guess I'll just power through.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -179,15 +180,25 @@ impl NormalizeVisitor {
...
     type Break = ();
 
-    fn pre_visit_relation(&mut self, relation: &mut ObjectName) -> ControlFlow<Self::Break> {
-        Self::simplify_compound_identifier(&mut relation.0);
-        ControlFlow::Continue(())
-    }
+    // fn pre_visit_relation(&mut self, relation: &mut ObjectName) -> ControlFlow<Self::Break> {
+    //     Self::simplify_compound_identifier(&mut relation.0);
+    //     ControlFlow::Continue(())
+    // }
```
jjbayer:
```suggestion
    // fn pre_visit_relation(&mut self, relation: &mut ObjectName) -> ControlFlow<Self::Break> {
    //     Self::simplify_compound_identifier(&mut relation.0);
    //     ControlFlow::Continue(())
    // }
```

relay-dynamic-config/src/defaults.rs
```diff
@@ -29,12 +29,9 @@ const MOBILE_OPS: &[&str] = &[
 /// A list of span descriptions that indicate top-level app start spans.
 const APP_START_ROOT_SPAN_DESCRIPTIONS: &[&str] = &["Cold Start", "Warm Start"];
 
-/// A list of patterns found in MongoDB queries.
+/// A list of patterns found in MongoDB queries
```
```suggestion
/// A list of patterns found in MongoDB queries.
```

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -492,12 +505,179 @@ impl VisitorMut for NormalizeVisitor {
...
+                        alias,
+                        lock_type: _,
+                    } = table;
+                    Self::scrub_name(table);
+                    if let Some(alias) = alias {
+                        Self::scrub_name(alias);
+                    }
+                }
+            }
+            Statement::UnlockTables => {}
```
jjbayer, line 671: These have not caused high cardinality in the past, but it's better to list them explicitly in case a new variant is introduced.

olksdr (approved):

Dav1dde (approved):

PR #3092, opened 2024-02-12
chore(build): Address beta Rust lint warnings

olksdr:
Make sure we can compile Relay with beta toolchain. 

#skip-changelog

### DIFF
--- a/relay-event-derive/src/lib.rs
+++ b/relay-event-derive/src/lib.rs
@@ -440,6 +440,7 @@ impl FieldAttrs {
 }
 
 #[derive(Copy, Clone, Default)]
+#[allow(dead_code)]
 enum SkipSerialization {
     #[default]
     Never,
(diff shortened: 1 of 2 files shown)


relay-event-derive/src/lib.rs
```diff
@@ -440,6 +440,7 @@ impl FieldAttrs {
 }
 
 #[derive(Copy, Clone, Default)]
+#[allow(dead_code)]
```
Dav1dde: Is this actually still used, Rust Analyzer tells me it's unused

Dav1dde (approved):

relay-event-derive/src/lib.rs
olksdr: That's a question, do we want to keep this around or just delete?

Dav1dde: If it is actually unused, I'd delete it.

olksdr: Makes sense. 
Done in [`57550e6`](https://github.com/getsentry/relay/pull/3092/commits/57550e6fad22a52509c83d2489d00501391acccd)

iker-barriocanal (approved):
Thank you for doing this!

PR #3096, opened 2024-02-12
ref(metrics): Add additional metrics to better estimate the usage of metrics by namespace

Dav1dde:
#skip-changelog

### DIFF
--- a/relay-server/src/envelope.rs
+++ b/relay-server/src/envelope.rs
@@ -142,35 +142,39 @@ impl ItemType {
             }
         }
     }
+
+    pub fn as_str(&self) -> &str {
+        match self {
+            Self::Event => "event",
+            Self::Transaction => "transaction",
+            Self::Security => "security",
+            Self::Attachment => "attachment",
+            Self::FormData => "form_data",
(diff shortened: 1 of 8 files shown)


relay-server/src/envelope.rs
```diff
@@ -142,35 +142,39 @@ impl ItemType {
...
+            Self::MetricBuckets => "metric_buckets",
+            Self::MetricMeta => "metric_meta",
+            Self::ClientReport => "client_report",
+            Self::Profile => "profile",
+            Self::ReplayEvent => "replay_event",
+            Self::ReplayRecording => "replay_recording",
+            Self::CheckIn => "check_in",
+            Self::Span => "span",
+            Self::OtelSpan => "otel_span",
+            Self::Unknown(ref s) => s,
```
Dav1dde, line 170: Potential cardinality explosion here, but should be fine?

jjbayer: I'd prefer to map this to unknown for the metric.

jjbayer (approved):
I would do a Canary deploy to make sure we don't take down Relay with the higher amount of metrics.

PR #3097, opened 2024-02-12
ref(normalization): Remove StoreProcessor

iker-barriocanal:
This PR removes the `StoreProcessor`, bringing us one step closer to removing the store normalizer entirely.

The biggest difference is that normalization of default attributes is now in `event::normalize` and runs in all relays, not just processing.

### Design choices

- There's a `StoreConfig` struct in the c-abi. Setting the configuration from librelay should happen with a single object and Relay uses 3 (2 for validation and 1 for normalization), so we can't get rid of it in the current setting.
- The normalization config has a `is_last_normalize`. Relay needs to know whether it's the last normalization step to e.g. remove unknown fields not to break future compatibility. This flag is currently enabled in processing relays and librelay calls, and it's disabled by default.

### Future work
- Move `normalize::mod` contents to their appropriate place. Currently, it contains a few types, a couple of functions used in other places, and tests validating the functionality in other modules.
- Make running normalization configurable, and only run it in PoP relays.

#skip-changelog


relay-cabi/src/processing.rs
```diff
@@ -126,7 +126,13 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
     validate_event_timestamps(&mut event, &event_validation_config)?;
 
     let normalization_config = NormalizationConfig {
+        project_id: config.project_id,
+        client: config.client.clone(),
+        protocol_version: config.protocol_version.clone(),
+        key_id: config.key_id.clone(),
+        grouping_config: config.grouping_config.clone(),
```
iker-barriocanal, line 133: Once we get rid of StoreProcessor, we'll also get rid of these `clone`s.

olksdr (commented):
@iker-barriocanal  are those failing integration test just flaky or failing for a reason? 
Could you, please, have a look into them before someone take a closer look to this PR? 
Thanks!

tests/integration/test_store.py
```diff
@@ -263,20 +263,6 @@ def store_event():
     store_count.acquire(timeout=2)
 
 
-def test_store_not_normalized(mini_sentry, relay):
```
iker-barriocanal, line 266: The test is stale as normalization runs for all relays, including the default attributes (like `key_id`).

relay-server/src/services/processor.rs
```diff
@@ -1208,8 +1208,33 @@ impl EnvelopeProcessorService {
                 max_secs_in_future: Some(self.inner.config.max_secs_in_future()),
                 is_validated: false,
             };
+
+            let key_id = state
+                .project_state
+                .get_public_key_config()
+                .and_then(|key| Some(key.numeric_id?.to_string()));
+            // TODO(iker): log the missing key_id error in all internal relays
```
iker-barriocanal, line 1216: When relay supports configuration to run normalization (run on pops, not processing), I'll remove this `todo`.

tests/integration/fixtures/security_report/csp_chrome_blocked_asset.no_processing.output.json
```diff
@@ -23,12 +23,18 @@
   },
   "culprit": "style-src cdn.example.com",
   "environment": "production",
+  "grouping_config": {
```
iker-barriocanal, line 26: These snapshot updates are the result of normalizing default attributes in regular normalization.

Dav1dde (approved):
As discussed offline, this is just an intermediate step to get rid of the `StoreProcessor` in the coming days which will cleanup a bunch of additional config and code which might be confusing for now (e.g. `event::store` and other callsites and misleading config).

relay-event-normalization/src/event.rs
```diff
@@ -156,6 +212,18 @@ pub fn normalize_event(event: &mut Annotated<Event>, config: &NormalizationConfi
         let _ =
             trimming::TrimmingProcessor::new().process_event(event, meta, ProcessingState::root());
     }
+
+    if config.is_last_normalize && config.remove_other.unwrap_or(!is_renormalize) {
+        // Remove unknown attributes at every level
+        let _ =
+            remove_other::RemoveOtherProcessor.process_event(event, meta, ProcessingState::root());
```
iker-barriocanal, line 216: These two processors don't return `ProcessingResult` errors that drop events, so it's safe to ignore the result.

relay-cabi/src/processing.rs
```diff
@@ -86,32 +206,31 @@ pub unsafe extern "C" fn relay_valid_platforms(size_out: *mut usize) -> *const R
...
 #[no_mangle]
 #[relay_ffi::catch_unwind]
 pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
-    normalizer: *mut RelayStoreNormalizer,
+    config: *mut RelayStoreConfig,
     event: *const RelayStr,
 ) -> RelayStr {
-    let processor = normalizer as *mut StoreProcessor;
+    let config = config as *mut StoreConfig;
+    let config = (*config).this();
```
iker-barriocanal: Making this in one line is less readable:
```rs
(*(config as *mut StoreConfig)).this()
```

relay-event-normalization/src/event.rs
Dav1dde: Would be good to add this comment to the code

relay-event-normalization/src/event.rs
```diff
@@ -201,6 +212,18 @@ pub fn normalize_event(event: &mut Annotated<Event>, config: &NormalizationConfi
         let _ =
             trimming::TrimmingProcessor::new().process_event(event, meta, ProcessingState::root());
     }
+
+    if config.is_last_normalize && config.remove_other.unwrap_or(!is_renormalize) {
```
Dav1dde: I find the behaviour strange if I explicitly state `remove_other: Some(true)` and then it does not run because it depends on `is_last_normalize` (additionally this is not documented).

Possibly include the check only in the `unwrap_or`.

relay-event-normalization/src/event.rs
```diff
@@ -109,6 +115,9 @@ pub struct NormalizationConfig<'a> {
     /// advanced normalizations such as inferring contexts or clock drift correction are disabled.
     pub is_renormalize: bool,
 
+    /// Overrides the default flag for other removal.
+    pub remove_other: Option<bool>,
```
Dav1dde: Maybe in a follow-up we can start removing these optionals?

The `StoreConfig` in the C-ABI can still have the optionals, but the NormalizationConfig doesn't have to have it.

relay-cabi/src/processing.rs
```diff
@@ -9,29 +9,149 @@ use std::os::raw::c_char;
...
 use relay_sampling::SamplingConfig;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
 
 use crate::core::{RelayBuf, RelayStr};
 
+/// Configuration for the store step -- validation and normalization.
+#[derive(Serialize, Deserialize, Debug, Default)]
+#[serde(default)]
+pub struct StoreConfig {
```
Dav1dde: I think we want to call this `StoreNormalizer` or wrap the config in a type called `StoreNormalizer` so we have still a consistent naming with the C-ABI functions.

relay-cabi/src/processing.rs
```diff
@@ -9,29 +9,149 @@ use std::os::raw::c_char;
...
+    /// Controls whether spans should be normalized (e.g. normalizing the exclusive time).
+    ///
+    /// To normalize spans in [`normalize_event`], `is_renormalize` must
+    /// be disabled _and_ `normalize_spans` enabled.
+    pub normalize_spans: bool,
+}
+
+impl StoreConfig {
+    /// Helper method to parse *mut StoreConfig -> &StoreConfig
+    pub fn this(&self) -> &Self {
```
Dav1dde:
```suggestion
   fn this(&self) -> &Self {
```

Looks like this doesn't need to be public

relay-event-normalization/src/event.rs
```diff
@@ -31,14 +31,20 @@ use crate::normalize::request;
...
+    MaxChars, PerformanceScoreConfig, RawUserAgentInfo, SpanDescriptionRule, TransactionNameConfig,
 };
 
 /// Configuration for [`normalize_event`].
 #[derive(Clone, Debug)]
 pub struct NormalizationConfig<'a> {
+    /// Whether it's the last normalization step.
+    ///
+    /// The last normalization run performs a few additional steps that
+    /// shouldn't happen in previous runs, like removing unknown attributes.
```
Dav1dde: Not quite true, it also needs `remove_other` to be enabled.

I think we can kill it in favor of having specialized bools like `remove_other` and `emit_events`

jjbayer: > I think we can kill it in favor of having specialized bools

+1, `is_last_normalize` could be misleading because the final normalization step actually happens in sentry.

relay-server/src/services/processor.rs
jjbayer, line 1: Should we rename `light_normalize_event` to `normalize_event` now?

relay-event-normalization/src/event.rs
```diff
@@ -894,6 +1019,210 @@ fn normalize_breakdowns(event: &mut Event, breakdowns_config: Option<&Breakdowns
...
+            let shortened = shorten_logger(event_logger, meta);
+            event.logger.set_value(Some(shortened));
+        }
+    }
+}
+
+/// Returns `true` if the given platform string is a known platform identifier.
+///
+/// See [`VALID_PLATFORMS`] for a list of all known platforms.
+pub fn is_valid_platform(platform: &str) -> bool {
```
jjbayer, line 1065: nit: Can we remove this to the top, because it's public?

relay-event-normalization/src/event.rs
iker-barriocanal: I find the comment redundant -- if you're reading it, you already see the result being ignored.

relay-cabi/src/processing.rs
```diff
@@ -80,38 +200,37 @@ pub unsafe extern "C" fn relay_valid_platforms(size_out: *mut usize) -> *const R
...
 #[no_mangle]
 #[relay_ffi::catch_unwind]
 pub unsafe extern "C" fn relay_store_normalizer_new(
     config: *const RelayStr,
     _geoip_lookup: *const RelayGeoIpLookup,
 ) -> *mut RelayStoreNormalizer {
-    let config: StoreConfig = serde_json::from_str((*config).as_str())?;
-    let normalizer = StoreProcessor::new(config);
-    Box::into_raw(Box::new(normalizer)) as *mut RelayStoreNormalizer
+    let config: StoreNormalizer = serde_json::from_str((*config).as_str())?;
```
jjbayer:
```suggestion
    let normalizer: StoreNormalizer = serde_json::from_str((*config).as_str())?;
```

jjbayer: same for functions below

relay-cabi/src/processing.rs
```diff
@@ -9,24 +9,144 @@ use std::os::raw::c_char;
...
+    ///
+    /// It is persisted into the event payload for correlation.
+    pub replay_id: Option<Uuid>,
+
+    /// Controls whether spans should be normalized (e.g. normalizing the exclusive time).
+    ///
+    /// To normalize spans in [`normalize_event`], `is_renormalize` must
+    /// be disabled _and_ `normalize_spans` enabled.
+    pub normalize_spans: bool,
+}
```
jjbayer, line 141: Is this a 1:1 copy of the old `StoreConfig`, or were there changes? Might be worth leaving the class definition in the old location to make the diff easier to read.

iker-barriocanal: Yep, it's a 1:1 copy

relay-event-normalization/src/event.rs
iker-barriocanal: All definitions above are also public. Where would you move it?

relay-cabi/src/processing.rs
Dav1dde: I like that is moved, Relay itself doesn't need it anymore.

relay-server/src/services/processor.rs
iker-barriocanal: Renamed most references in [59c0c95](https://github.com/getsentry/relay/pull/3097/commits/59c0c950929b63e64998e8ee817a21e7812b8dba). There are a couple left that I'll address in a follow-up as they'll require a bit more work than just renaming.

relay-cabi/src/processing.rs
iker-barriocanal: Done in [59c0c95](https://github.com/getsentry/relay/pull/3097/commits/59c0c950929b63e64998e8ee817a21e7812b8dba).

relay-event-normalization/src/event.rs
iker-barriocanal: Killed the config in [59c0c95](https://github.com/getsentry/relay/pull/3097/commits/59c0c950929b63e64998e8ee817a21e7812b8dba).

relay-event-normalization/src/event.rs
iker-barriocanal: Done in [59c0c95](https://github.com/getsentry/relay/pull/3097/commits/59c0c950929b63e64998e8ee817a21e7812b8dba).

relay-event-normalization/src/event.rs
iker-barriocanal: Using a single option in [59c0c95](https://github.com/getsentry/relay/pull/3097/commits/59c0c950929b63e64998e8ee817a21e7812b8dba).

Dav1dde (approved):

relay-cabi/src/processing.rs
```diff
@@ -127,8 +246,16 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
...
     validate_transaction(&mut event, &tx_validation_config)?;
 
+    let is_renormalize = config.is_renormalize.unwrap_or(false);
+
     let normalization_config = NormalizationConfig {
+        project_id: config.project_id,
+        client: config.client.clone(),
+        protocol_version: config.protocol_version.clone(),
+        key_id: config.key_id.clone(),
+        grouping_config: config.grouping_config.clone(),
```
jjbayer, line 256: Do these have to be clone?

relay-cabi/src/processing.rs
jjbayer: I like that it's moved too! My comment was about diff readability. But if there's no diff then it's fine.

relay-cabi/src/processing.rs
```diff
@@ -137,7 +264,9 @@ pub unsafe extern "C" fn relay_store_normalizer_normalize_event(
         breakdowns_config: None, // only supported in relay
         normalize_user_agent: config.normalize_user_agent,
         transaction_name_config: Default::default(), // only supported in relay
-        is_renormalize: config.is_renormalize.unwrap_or(false),
+        is_renormalize,
+        remove_other: config.remove_other.unwrap_or(!is_renormalize),
+        emit_error_events: !is_renormalize,
```
jjbayer:
```suggestion
        emit_event_errors: !is_renormalize,
```

jjbayer (approved):
There's a small chance that this breaks things, because processing steps running between light normalization and store normalization could in theory modify the event in ways that were previously caught by store normalization. But it's a risk worth taking IMO.

relay-cabi/src/processing.rs
iker-barriocanal: Yep, `NormalizationConfig` expects `String`s (and `serde_json::Value` for `grouping_config`). I'll look into the config to see if it needs to own the data.

PR #3102, opened 2024-02-13
ref: Coerce all Thread IDs to integers

untitaker:
While porting the Snuba consumers to Rust, I saw that thread IDs could
be strings. This isn't really aligned with how it's typed in our
frontend, typed in symbolicator, and appears to subvert user
expectations as well: https://github.com/getsentry/sentry-go/issues/767

It seems easiest to restrict the output type and silently coerce
strings to integers. This also gets rid of weird artifacts like people
sending in random (non-numerical) strings which is currently allowed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 **Internal**:
 
 - Set the span op on segments. ([#3082](https://github.com/getsentry/relay/pull/3082))
+- Coerce all Thread IDs to integers. ([#3102](https://github.com/getsentry/relay/pull/3102))
 
 ## 24.1.2
 
--- a/relay-event-schema/src/protocol/exception.rs
+++ b/relay-event-schema/src/protocol/exception.rs
@@ -93,7 +93,7 @@ mod tests {
             ty: Annotated::new("mytype".to_string()),
             value: Annotated::new("myvalue".to_string().into()),
             module: Annotated::new("mymodule".to_string()),
-            thread_id: Annotated::new(ThreadId::Int(42)),
+            thread_id: Annotated::new(ThreadId(42)),
             other: {
                 let mut map = Map::new();
                 map.insert(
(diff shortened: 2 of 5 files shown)


untitaker:
on second look, it seems that we sometimes get thread IDs like `$47F8`. is that a sensible value? @Swatinem any idea?

Swatinem:
The `$` prefix looks a bit bogus, but otherwise I believe the reason this is `u64 | String` is because you cannot represent a full `u64` in JSON.
Although we use `u64 | String` in the Rust SDK, and a full `u64` in Symbolicator without giving much thought to it.
On the other hand though, serde can deal with full `u64` without any problem, and it even has a `arbitrary_precision` feature which turns `serde_json::Value::Number` into a `String` internally.

Swatinem (approved):
lgtm, though I don’t have enough insight how we are treating these in the product.

iker-barriocanal:
The [thread interface](https://develop.sentry.dev/sdk/event-payloads/threads/#attributes) leaves the door open for non-numeric strings (typically = not required):

>id
>Required. The ID of the thread. Typically a number or numeric string. Needs to be unique among the threads. An exception can set the thread_id attribute to cross-reference this thread.

Is this aligned with the current behavior of SDKs, or is this a breaking change?

untitaker:
@iker-barriocanal generally this PR will convert e.g. `"42"` to `42`, and I thought that's enough to meet the external contract to SDKs.

The problem is with the string values we send in practice and that currently pass through unfiltered.

iker-barriocanal:
@untitaker yeah but for the example above, `$47F8`, this would be a breaking change. Do we know how often we get these values, or what the impact of a breaking change would be? I'm happy to simplify the protocol, but I'd like to avoid degrading user experience.

Dav1dde:
Do you want to completely get rid of strings or would it be enough if Relay tried to normalize strings to ints but if it fails we keep the id around as a string?

untitaker:
seems too hard i guess, fixing in snuba instead: https://github.com/getsentry/snuba/pull/5531

untitaker:
> Do you want to completely get rid of strings or would it be enough if Relay tried to normalize strings to ints but if it fails we keep the id around as a string?

either strings can happen or they can't, the middle-ground isn't particularly interesting to me. i think i'll just fix this in snuba, but i think it would still be good if somebody with more knowledge decided what to do here -- the UI is typed differently at least, and I haven't checked if it crashes or doesn't

PR #3104, opened 2024-02-13
fix: Normalize route path in trace context data

### LINKED ISSUE #3000: `contexts.trace.data.route` getting rejected by Sentry
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

4.13.1

### Steps to reproduce

1. Trigger an event that sends a trace to Sentry.
2. Observe the trace details in the Sentry issue interface.
3. The `contexts.trace.data.route` field does not contain the actual route name.

### Expected result

contexts.trace.data.route should contain route name

### Actual result

obrazek

```
{
http.flavor: 1.1,
 http.request.method: GET, 
 http.url: ...,
net.host.name: ..., 
net.host.port: 443, 
net.peer.ip: ..., 
route: None
}
```

olksdr:
Relay expects for `contexts.trace.data.route` to be a route, but some SDKs sent the string (for route names) in `route` and we want to handle this and normalize the data into the know types structure. 

This PR introduces the customer `FromValue` trait implementation for `Route` to handle:
* incoming object
* incoming string for `contexts.trace.data.route` field

fix: [#3000](https://github.com/getsentry/relay/issues/3000)

### DIFF
--- a/relay-event-schema/src/protocol/contexts/trace.rs
+++ b/relay-event-schema/src/protocol/contexts/trace.rs
@@ -203,7 +203,7 @@ impl FromValue for Route {
             }
             Annotated(None, meta) => Annotated(None, meta),
             Annotated(Some(value), mut meta) => {
-                meta.add_error(Error::expected("route"));
+                meta.add_error(Error::expected("route expected to be an object"));
                 meta.set_original_value(Some(value));
                 Annotated(None, meta)
             }
--- a/relay-event-schema/src/protocol/contexts/trace.rs
+++ b/relay-event-schema/src/protocol/contexts/trace.rs
@@ -153,9 +153,12 @@ pub struct Data {
 }
 
 /// The route in the application, set by React Native SDK.
-#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[derive(Clone, Debug, Default, PartialEq, Empty, IntoValue, ProcessValue)]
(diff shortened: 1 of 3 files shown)


relay-event-schema/src/protocol/contexts/trace.rs
```diff
@@ -170,6 +173,44 @@ pub struct Route {
...
+
+                if !values.is_empty() {
+                    route.other = values;
+                }
+
+                Annotated(Some(route), meta)
+            }
+            Annotated(None, meta) => Annotated(None, meta),
+            Annotated(Some(value), mut meta) => {
+                meta.add_error(Error::expected("route"));
```
iker-barriocanal: nit (not sure if the message is the most helpful one):
```suggestion
                meta.add_error(Error::expected("route expected to be an object"));
```

iker-barriocanal (approved):
Thanks for the fix!

I wonder if we should document this in the [trace context](https://develop.sentry.dev/sdk/event-payloads/contexts/#trace-context) to avoid similar issues in the future.

olksdr:
> Thanks for the fix!
> 
> I wonder if we should document this in the [trace context](https://develop.sentry.dev/sdk/event-payloads/contexts/#trace-context) to avoid similar issues in the future.

added docs https://github.com/getsentry/develop/pull/1157

iker-barriocanal (approved):

Dav1dde (approved):

PR #3105, opened 2024-02-13
feat(replays): Add ReplayVideo envelope item type

cmanallen:
Add support for replay-video event types.

Changes:

- Refactors event validation in the processor. This is to DRY the processing logic between the three replay event types.
- Adds new replay_video field to replay-recording Kafka message.
- Add outcomes.
- Configure rate-limits.

Relates to:
 * https://github.com/getsentry/sentry/issues/63255


jjbayer (commented):
> Is there a good pattern for handling envelope-items that depend on the presence of another, valid envelope-item? I.e. erroring if we see one but not the other.

There is no good pattern for that, but we do it, e.g. here: https://github.com/getsentry/relay/blob/a9baace4ae06626e29d3b1cdcb07b4113523a5d4/relay-server/src/services/processor/profile.rs#L29-L30

> How are rate-limits handled? These are logically one request but rate-limits are calculated per item.

If a video always has a recording, but a recording does not necessarily have a video, I would always drop the video with the recording (see code below), but only emit an outcome for the recording. https://github.com/getsentry/relay/blob/ac1ff144600f64f1ceac79b7298cd97b6545812b/relay-server/src/utils/rate_limits.rs#L640-L645

In general, if recordings and videos are tightly coupled, I would personally embed the video into the recording item, unless that is a problem size-wise.

cmanallen:
@jjbayer Updated the `ReplayVideo` item type to use a single envelope item.  I'm stashing the decomposed types on the item headers so I can avoid serializing it onto the payload and then deserializing it in `store.rs`.

relay-server/src/envelope.rs
```diff
@@ -771,6 +783,17 @@ impl Item {
         self.headers.replay_combined_payload = combined_payload;
     }
 
+    /// Returns the payload's replay video events.
+    #[cfg(feature = "processing")]
+    pub fn replay_video_events(&self) -> Option<(Vec<u8>, Vec<u8>)> {
+        self.headers.replay_video_events.clone()
```
TBS1996: clone here seems a bit too implicit, I get that we need an owned value to make a Bytes object, but maybe defer cloning to the caller so they know it's potentially expensive.  
 
we could also look into something like take_video_events if we only need to use it once and the payload is large

relay-server/src/envelope.rs
```diff
@@ -509,6 +513,11 @@ pub struct ItemHeaders {
     #[serde(default, skip)]
     replay_combined_payload: bool,
 
+    /// The parsed replay-event and replay-recording payloads.
+    /// NOTE: This is internal-only and not exposed into the Envelope.
+    #[serde(default, skip)]
+    replay_video_events: Option<(Vec<u8>, Vec<u8>)>,
```
jjbayer: I would prefer if we don't put this on the header, but make it part of the item payload somehow. We should either

1. manage to encode all information into a single item payload (e.g. `{"some": "json"}\x00vide0_bytes`),
2. let them be separate items after all, and only combine them in `store.rs`.

@cmanallen let me know if you would prefer to discuss this in a meeting.

relay-server/src/services/processor/replay.rs
```diff
@@ -54,82 +54,108 @@ pub fn process(
...
+            _ => ItemAction::Keep,
         }
-        _ => ItemAction::Keep,
     });
 
     Ok(())
 }
 
+enum ProcessingAction<T> {
+    Drop(ItemAction),
```
jjbayer:
```suggestion
    Drop(Outcome),
```

relay-server/src/envelope.rs
```diff
@@ -213,6 +216,7 @@ impl std::str::FromStr for ItemType {
             "profile" => Self::Profile,
             "replay_event" => Self::ReplayEvent,
             "replay_recording" => Self::ReplayRecording,
+            "replay_video" => Self::ReplayVideo,
```
jjbayer, line 228: How large would we expect these videos to be?

jjbayer (changes requested):

relay-server/src/envelope.rs
cmanallen: 10MB max.

relay-server/src/envelope.rs
cmanallen: Yes I would like to discuss in a meeting.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 **Features**:
 
 - Add protobuf support for ingesting OpenTelemetry spans and use official `opentelemetry-proto` generated structs. ([#3044](https://github.com/getsentry/relay/pull/3044))
+- Adds ReplayVideo envelope-item type. ([#3105](https://github.com/getsentry/relay/pull/3105))
```
jjbayer: Please move this to the `## Unreleased` section.

relay-server/Cargo.toml
```diff
@@ -97,6 +96,7 @@ reqwest = { version = "0.11.1", features = [
 rmp-serde = "1.1.1"
 rust-embed = { version = "8.0.0", optional = true }
 serde = { workspace = true }
+serde_bytes = "0.11"
```
jjbayer, line 99: This is a crate by the creators of serde, so adding it as dependency should be fine.

jjbayer (approved):

Dav1dde (approved):

relay-server/src/services/store.rs
```diff
@@ -891,6 +908,66 @@ impl StoreService {
...
+            Some(replay_event),
+            Some(replay_video),
+            start_time,
+            retention,
+        )?;
+
+        metric!(
+            counter(RelayCounters::ProcessingMessageProduced) += 1,
+            event_type = "replay_video"
+        );
```
jjbayer: `produce_replay_recording` already logs a metric. We should emit 1 metric for every Kafka message, this would emit two IIURC. What we could do is change the "event type" in `produce_replay_recording` to `recording_with_video` or similar if a video is passed with it.

relay-server/src/services/store.rs
```diff
@@ -891,6 +908,66 @@ impl StoreService {
...
+        };
+
+        self.produce_replay_event(
+            event_id.ok_or(StoreError::NoEventId)?,
+            scoping.organization_id,
+            scoping.project_id,
+            start_time,
+            retention,
+            Bytes::copy_from_slice(replay_event),
+        )?;
```
jjbayer: Why do we need to produce the event a second time? It's already published as part of the replay recording, right?

relay-server/src/services/store.rs
```diff
@@ -891,6 +908,66 @@ impl StoreService {
...
+            return Ok(());
+        };
+
+        self.produce_replay_event(
+            event_id.ok_or(StoreError::NoEventId)?,
+            scoping.organization_id,
+            scoping.project_id,
+            start_time,
+            retention,
+            Bytes::copy_from_slice(replay_event),
```
jjbayer: If we really need to duplicate the event, let's change the signature of `produce_replay_event` to take a reference, so we don't copy here.

jjbayer (changes requested):

relay-server/src/services/store.rs
cmanallen: Its being published to a different consumer.  We're producing one event to the replay_recording consumer and another event to the replay_event consumer.  The duplicated data being sent to the recording consumer for reasons other than storing the replay_event bytes.  Refer to this PR https://github.com/getsentry/relay/pull/3035 where we merged the payloads.

In the future we will drop the `produce_replay_event` function from Relay but we're not ready yet.

jjbayer (approved):

PR #3106, opened 2024-02-14
fix(server): Forward metrics in proxy mode

### LINKED ISSUE #3042: Proxy Relay drops Metrics because of missing feature flags
A Relay configured in Proxy mode does not forward metrics due to a missing metrics feature flag.

Since proxy Relays do not load project configs all metrics will be dropped. A proxy Relay should forward statsd metrics in an envelope.

TBS1996:
should fix: https://github.com/getsentry/relay/issues/3042

Problem was we're not able to send metrics through without scoping from the projectstate, but in proxy mode the projectstate isn't available for non-static projects. 

This PR simply skips the aggregation pipeline for metrics in proxy mode and send it to processing instead, which for metric groups does nothing and simply forwards it to upstream.


relay-server/src/services/project.rs
```diff
@@ -691,8 +694,17 @@ impl Project {
...
         envelope_processor: Addr<EnvelopeProcessor>,
         buckets: Vec<Bucket>,
+        project_cache: Addr<ProjectCache>,
     ) {
-        if self.metrics_allowed() {
+        if self.config.relay_mode() == RelayMode::Proxy {
+            if let Some(scoping) = self.partial_scoping() {
+                let mut bucket_map = HashMap::new();
+                bucket_map.insert(scoping.project_key, buckets);
+                project_cache.send(FlushBuckets {
```
TBS1996: skipped the aggregator for simplicity because it checks for scoping from projectstate and also features in `rate_limit_and_merge_buckets`. 
 
i wasnt sure if we should also skip `handle_flush_buckets`? it checks if it's ratelimited, do we care about ratelimiting in proxy mode?

relay-server/src/services/project.rs
```diff
@@ -1174,7 +1196,7 @@ impl Project {
             return CheckedBuckets::ProjectDisabled(len);
         }
 
-        let Some(scoping) = self.scoping() else {
+        let Some(scoping) = self.scoping().or(self.partial_scoping()) else {
```
TBS1996: so this is the most crucial line in this PR. If anyone have any ideas why it might be bad to use partial scoping here then please comment it

tests/integration/test_metrics.py
```diff
@@ -62,6 +62,52 @@ def metrics_by_name_group_by_project(metrics_consumer, timeout=None):
...
+
+    metrics_item = envelope.items[0]
+    assert metrics_item.type == "metric_buckets"
+
+    received_metrics = json.loads(metrics_item.get_bytes().decode())
+    received_metrics = sorted(received_metrics, key=lambda x: x["name"])
+    assert received_metrics == [
+        {
+            "timestamp": timestamp,
+            "width": 0,
```
TBS1996: width = 0 because we skip aggregation, is this a problem for later relays when they get it?

relay-server/src/services/project.rs
TBS1996: it'd be none if it's sent from the batch_metrics route though, unless an envelope from that project was sent earlier.  although i think the batch metrics  only gets sent from another relay. I guess it would be possible if you have multiple proxy relays in a chain and they have global metrics enabled?  
 
so we could also create a scoping from just the projectkey here for that scenario

tests/integration/test_metrics.py
```diff
@@ -62,6 +62,52 @@ def metrics_by_name_group_by_project(metrics_consumer, timeout=None):
             return metrics_by_project
 
 
+def test_metrics_proxy_mode(mini_sentry, relay):
```
TBS1996: this test fails without the changes in this PR

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 **Internal**:
 
 - Set the span op on segments. ([#3082](https://github.com/getsentry/relay/pull/3082))
+- Allow metrics to be forwarded with just partial scoping. ([#3106](https://github.com/getsentry/relay/pull/3106))
```
jjbayer: Please move this to **Bug Fixes**. Also note that we had a 24.2.0 release, so please merge master and then move this line to the Unreleased section.

```suggestion
- Allow metrics to be forwarded in proxy mode. ([#3106](https://github.com/getsentry/relay/pull/3106))
```

relay-server/src/endpoints/common.rs
```diff
@@ -262,28 +263,30 @@ fn queue_envelope(
...
-        relay_log::trace!("sending metrics into processing queue");
-        state.processor().send(ProcessMetrics {
-            items: metric_items.into_vec(),
-            project_key: envelope.meta().public_key(),
-            start_time: envelope.meta().start_time(),
-            sent_at: envelope.sent_at(),
-        });
-    }
+    if state.config().relay_mode() != RelayMode::Proxy {
+        // Remove metrics from the envelope and queue them directly on the project's `Aggregator`.
```
jjbayer, line 267:
```suggestion
        // Remove metrics from the envelope and queue them directly on the project's `Aggregator`. In proxy mode, we cannot aggregate metrics because we may not have a project ID.
```

relay-server/src/services/processor.rs
```diff
@@ -234,6 +234,15 @@ impl ProcessingGroup {
             ))
         }
 
+        let is_metric = |i: &Item| matches!(i.ty(), ItemType::Statsd | ItemType::MetricBuckets);
```
jjbayer: Might be worth adding metrics meta to the integration test as well:

```suggestion
        let is_metric = |i: &Item| matches!(i.ty(), ItemType::Statsd | ItemType::MetricBuckets | ItemType::MetricMeta);
```

We could add a `is_metrics()` method to `ItemType` and use it in `queue_envelope` and here.

tests/integration/test_metrics.py
```diff
@@ -62,6 +62,35 @@ def metrics_by_name_group_by_project(metrics_consumer, timeout=None):
...
+                "debounce_delay": 0,
+                "shift_key": "none",
+            },
+        },
+    )
+
+    project_id = 42
+
+    timestamp = int(datetime.now(tz=timezone.utc).timestamp())
+    metrics_payload = f"transactions/foo:42|c\ntransactions/bar:17|c|T{timestamp}"
```
jjbayer: Please also add a buckets payload and a metrics meta payload.

jjbayer (commented):
Nice! See comment about missing condition.

relay-server/src/endpoints/common.rs
```diff
@@ -262,28 +263,30 @@ fn queue_envelope(
...
-    // Remove metric meta from the envelope and send them directly to processing.
-    let metric_meta = envelope.take_items_by(|item| matches!(item.ty(), ItemType::MetricMeta));
-    if !metric_meta.is_empty() {
-        relay_log::trace!("sending metric meta into processing queue");
-        state.processor().send(ProcessMetricMeta {
-            items: metric_meta.into_vec(),
-            project_key: envelope.meta().public_key(),
-        })
+        // Remove metric meta from the envelope and send them directly to processing.
+        let metric_meta = envelope.take_items_by(|item| matches!(item.ty(), ItemType::MetricMeta));
```
Dav1dde, line 283: Nice! We didn't forget about metric meta.

jjbayer (approved):

Issue #3095, opened 2024-02-12
network outage, scheduling another check in 60 sec


pyr0hu:
Hello!

I've installed Relay into our infra to improve the response time for the HTTP API of our application.

We installed relay as the binary, added a static config and started a systemd service which keeps the relay running. After a minute or so, I start to get the network outage in the logs and the relay stops proxying the incoming requests. Enabling debug logs does not show anything out of order. It is deployed in AWS as an EC2 instance. The only third-party stuff on the instance is a Tailscale VPN connection. The sec groups should allow all outgoing connection so I don't think that's the problem as the sentry API calls seems to run OK. (See logs later)

Relay is running using the following systemd unit:

```
[Unit]
Description=Sentry Relay
After=network.target
StartLimitIntervalSec=0

[Service]
Type=simple
Restart=always
RestartSec=3
User=ec2-user
ExecStart=/usr/local/bin/sentry-relay run --config /home/ec2-user/.relay

[Install]
WantedBy=multi-user.target
```

The config folder contains the following:

```
.
├── config.yml
├── credentials.json
└── projects
    └── <project id>.json
```

config.yml

```
relay:
  mode: static
  upstream: https://sentry.io/
  host: 0.0.0.0
  port: 80
logging:
  level: debug
```

Project json is
```
{
  "slug": "<slug>",
  "publicKeys": [
    {
      "publicKey": "<pk>",
      "isEnabled": true
    }
  ],
  "config": {
    "allowedDomains": ["*"]
  }
}
```

Excerpt from the logs:

```
Feb 12 14:52:15 ip-10-0-1-183.eu-west-1.compute.internal sentry-relay[100660]: 2024-02-12T14:52:15.133025Z DEBUG relay_server::services::project_local: loading local states from file system directory="/home/ec2-user/.relay/projects"
Feb 12 14:52:25 ip-10-0-1-183.eu-west-1.compute.internal sentry-relay[100660]: 2024-02-12T14:52:25.132325Z DEBUG relay_server::services::project_local: loading local states from file system directory="/home/ec2-user/.relay/projects"
Feb 12 14:52:29 ip-10-0-1-183.eu-west-1.compute.internal sentry-relay[100660]: 2024-02-12T14:52:29.429381Z  WARN relay_server::services::upstream: network outage, scheduling another check in 60s
[...]

PR #3110, opened 2024-02-14
chore(tests): Add test against network outage in static mode

TBS1996:
test is due to this issue: https://github.com/getsentry/relay/issues/3095

#skip-changelog


TBS1996:
@iker-barriocanal  is this what you specified?

PR #2907, opened 2024-01-04
ref(spans): Reduce number of tags on span metrics

jjbayer:
During the development of Starfish we've repeatedly added new tags to span metrics without restricting the conditions under which a tag is set (that is, what queries actually require a tag).

This PR introduces a new config for the `exclusive_time` and `exclusive_time_light` metrics, which limits the number of cases in which the metrics and / or their tags are extracted. Some tags are removed entirely.

For now, the new config will only be enabled for projects that have the `all-modules` feature flag enabled, i.e. sentry-internal. After we've verified that the new config covers all required cases, we can enable the new config for all orgs.

ref: https://github.com/getsentry/team-ingest/issues/252

PR #3111, opened 2024-02-14
fix(spans): Ingest span metrics for http spans

phacops:
For an unknown reason, we stopped ingesting them on January 7th when we shouldn't have.

I also removed the test for `all_modules` as this feature doesn't exist anymore.


relay-dynamic-config/src/defaults.rs
```diff
@@ -73,6 +73,8 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
 
     let is_mobile_sdk = RuleCondition::eq("span.sentry_tags.mobile", "true");
 
+    let is_http = RuleCondition::eq("span.op", "http.client");
```
jjbayer, line 76: Do we need exclusive time metrics for all `http.client` spans? Or is there a way to narrow it down further?

relay-dynamic-config/src/defaults.rs
```diff
@@ -216,7 +225,7 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
                     .always(),
                 Tag::with_key("span.domain")
                     .from_field("span.sentry_tags.domain")
-                    .when(is_db.clone() | is_resource.clone()),
+                    .when(is_db.clone() | is_resource.clone() | is_http.clone()),
```
jjbayer, line 228: @gggritso Do we already know what tag combinations / filters we want to query `exclusive_time` by? Does this config accurately reflect those access patterns?

jjbayer (changes requested):
> For an unknown reason

@phacops the decision to stop ingesting them was deliberate & coordinated: https://github.com/getsentry/relay/pull/2907#discussion_r1441870906

Pressing "Request changes" because of open questions.

relay-dynamic-config/src/defaults.rs
gggritso: We have initial product requirements, so we do know, even though it might change:

- `environment` for environment filters
- `span.domain`, to break down and filter by domain in the table, and possibly render top 5 domains charts
- `span.status_code`, to show 2xx, 4xx, and 5xx counts by span domain and/or by span transaction
- `transaction`, to show a table of transactions for a given span domain, filter by transaction
- `transaction.method` so we can filter by correct endpoint method in the sample panel

The config looks good to me 👍🏻

relay-dynamic-config/src/defaults.rs
gggritso: 🤔 I think we need it for everything, yeah. Since the HTTP module needs to show the total time spent on `http` spans, I don't think we can omit metrics for any spans

jjbayer (approved):
Let's ship it, we'll see what impact this has on total cardinality.

PR #3112, opened 2024-02-15
fix(performance_score): Skip profiles without required measurements

phacops:
Currently, if a profile cannot be calculated because of missing measurements or bad weight, we'll stop calculating performance scores.

What if we want to have 2 configurations, one for INP and one for TTFD, CLS, etc for the same browser (same condition)?

This would allow us to go through all profiles matching the condition and not stop at the first matching profile where we'd miss some components.


Dav1dde (approved):

relay-event-normalization/src/event.rs
```diff
@@ -757,7 +757,7 @@ pub fn normalize_performance_score(
                     // All non-optional measurements with a profile weight greater than 0 are
                     // required to exist on the event. Skip calculating performance scores if
                     // a measurement with weight is missing.
```
```suggestion
                    // required to exist on the event. Skip this profile if
                    // a measurement with weight is missing.
```

jjbayer (approved):
Makes sense to me, @edwardgou-sentry please have a look as well.

edwardgou-sentry (approved):
Thanks @phacops ! This is needed because we will be storing a pageload profile and separate interaction profile per browser. Tested out this change locally and it works

relay-event-normalization/src/event.rs
```diff
@@ -755,9 +755,9 @@ pub fn normalize_performance_score(
                         && !c.optional
                 }) {
                     // All non-optional measurements with a profile weight greater than 0 are
-                    // required to exist on the event. Skip calculating performance scores if
+                    // required to exist on the event. Skip this profile if
                     // a measurement with weight is missing.
-                    break;
+                    continue;
```
iker-barriocanal, line 760: Do we need to update the `break` on line 776?

phacops: I think you're right and we also do not want the `break` at the end, we want to go through all profiles for which the event match the condition.

That way, with a profile for INP and a profile for CLS/TTFD/etc, if we have the right measurements, we'd calculating all instead of just one.

PR #3113, opened 2024-02-15
ref(healthcheck): Instrument healthcheck code path

iker-barriocanal:
This PR adds various metrics around the code path for a health check. We're seeing some health checks being slower than expected, and these metrics should provide visibility. Once the root cause is identified and addressed, we'll evaluate what metrics are worth keeping and which should be removed.

#skip-changelog

### DIFF
--- a/relay-server/src/endpoints/health_check.rs
+++ b/relay-server/src/endpoints/health_check.rs
@@ -4,9 +4,11 @@ use axum::extract::Path;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
 use serde::Serialize;
+use tokio::time::Instant;
 
 use crate::service::ServiceState;
-use crate::services::health_check::IsHealthy;
+use crate::services::health_check::{IsHealthy, IsHealthyWrapper};
+use crate::statsd::RelayTimers;
 
 #[derive(Serialize)]
 struct Status {
@@ -14,7 +16,16 @@ struct Status {
 }
 
 pub async fn handle(state: ServiceState, Path(kind): Path<IsHealthy>) -> impl IntoResponse {
-    match state.health_check().send(kind).await {
+    let received = Instant::now();
+
+    let wrapper = IsHealthyWrapper { kind, received };
+    let result = state.health_check().send(wrapper).await;
+
+    relay_statsd::metric!(
+        timer(RelayTimers::HealthcheckEndpointCheckDuration) = received.elapsed(),
+        kind = kind.variant()
+    );
+    match result {
         Ok(true) => (StatusCode::OK, axum::Json(Status { is_healthy: true })),
         _ => (
             StatusCode::SERVICE_UNAVAILABLE,
--- a/relay-server/src/services/health_check.rs
+++ b/relay-server/src/services/health_check.rs
@@ -5,10 +5,11 @@ use relay_config::{Config, RelayMode};
 use relay_metrics::{AcceptsMetrics, Aggregator};
 use relay_statsd::metric;
 use relay_system::{Addr, AsyncResponse, Controller, FromMessage, Interface, Sender, Service};
+use tokio::time::Instant;
 
-use crate::services::project_cache::{ProjectCache, SpoolHealth};
+use crate::services::project_cache::{ProjectCache, SpoolHealth, SpoolHealthWrapper};
 use crate::services::upstream::{IsAuthenticated, IsNetworkOutage, UpstreamRelay};
-use crate::statsd::RelayGauges;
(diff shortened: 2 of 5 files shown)


relay-server/src/endpoints/health_check.rs
```diff
@@ -4,17 +4,28 @@ use axum::extract::Path;
...
 
 pub async fn handle(state: ServiceState, Path(kind): Path<IsHealthy>) -> impl IntoResponse {
-    match state.health_check().send(kind).await {
+    let received = Instant::now();
+
+    let wrapper = IsHealthyWrapper { kind, received };
+    let result = state.health_check().send(wrapper).await;
+
+    relay_statsd::metric!(
+        timer(RelayTimers::HealthcheckEndpointCheckDuration) = received.elapsed(),
```
iker-barriocanal, line 25: The middleware metric should be pretty much the same as this one, but emitting a new one as I'm not sure what metrics we can trust.

relay-server/src/services/project_cache.rs
```diff
@@ -196,6 +196,11 @@ impl UpdateSpoolIndex {
 #[derive(Debug)]
 pub struct SpoolHealth;
 
+pub struct SpoolHealthWrapper {
+    pub message: SpoolHealth,
+    pub sent_instant: Instant,
```
Dav1dde, line 201: I really don't like this metric, this is adds a lot of changes just to carry some instant through a bunch of layers.

We already have the time how long it takes for the "answer" and we already have a timer which tells us how long the operation took. If they don't match up, the remaining time was the delay.

relay-server/src/services/health_check.rs
```diff
@@ -22,15 +23,29 @@ pub enum IsHealthy {
...
+        match self {
+            IsHealthy::Liveness => "liveness",
+            IsHealthy::Readiness => "readiness",
+        }
+    }
+}
+
+pub struct IsHealthyWrapper {
+    pub kind: IsHealthy,
+    pub received: Instant,
```
Dav1dde, line 37: Same here, we already have the metric how long the operation takes and how long we waited for a result, lots of changes for something we already know.

relay-server/src/services/spooler/mod.rs
```diff
@@ -1128,6 +1149,10 @@ impl BufferService {
 
             BufferState::Disk(ref disk) => health.0.send(!disk.is_full().await.unwrap_or_default()),
         }
+        relay_statsd::metric!(
+            timer(RelayTimers::SpoolerHealthDuration) = check_start.elapsed(),
```
Dav1dde, line 1153: Do we need this? We already have `handle_health` which tells us how long the operation takes and we have a good indication that this is very very fast. Also for project configs relays (where we have the issue) this is always Memory.

relay-server/src/services/health_check.rs
```diff
@@ -98,17 +113,37 @@ impl HealthCheckService {
...
+        let handle_start = Instant::now();
         let HealthCheck(message, sender) = message;
-        let response = self.handle_is_healthy(message).await;
+        relay_statsd::metric!(
+            timer(RelayTimers::HealthcheckServiceReceivedDelay) = message.received.elapsed(),
+        );
+
+        let response = self.handle_is_healthy(message.kind).await;
+        relay_statsd::metric!(
+            timer(RelayTimers::HealthcheckServiceHandleDuration) = handle_start.elapsed(),
```
Dav1dde, line 144: Do we need this dance with `Instant`?

Something like this is what we do everywhere else:
```rs
        metric!(
            timer(RelayTimers:: HealthcheckServiceReceivedDelay),
            message = ty, { self.handle_is_healthy(message.kind).await }
        )
```

relay-server/src/services/health_check.rs
```diff
@@ -98,17 +113,37 @@ impl HealthCheckService {
...
+                    .project_cache
+                    .send(SpoolHealthWrapper {
+                        message: SpoolHealth,
+                        sent_instant: project_cache_sent,
+                    })
                     .await
-                    .unwrap_or_default()
+                    .unwrap_or_default();
+                relay_statsd::metric!(
+                    timer(RelayTimers::HealthcheckServiceProjectcacheResponseDuration) =
```
Dav1dde, line 126: I think we should have one metric for health check probes and tag it with what operation, e.g. `project_cache`, meaning we can instrument all of our probes easily.

Dav1dde (changes requested):

iker-barriocanal:
Doesn't reveal new info

PR #3115, opened 2024-02-15
fix(spans): Max description length

jjbayer:
Do not attempt to parse very long span description.

This should prevent stack overflows in the SQL parser.

We could also limit the length of the description field by setting a `bag_size` attribute on it. Unfortunately, the `TrimmingProcessor` runs after other normalization steps, so it would not help in this case.


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -38,6 +41,14 @@ pub(crate) fn scrub_span_description(
         return (None, None);
     };
 
+    if description.len() > MAX_DESCRIPTION_LENGTH {
+        relay_log::error!(
+            description = description,
+            "Span description too large to parse"
+        );
```
jjbayer, line 48: We can remove this error at a later point.

iker-barriocanal (approved):

olksdr (approved):

PR #3116, opened 2024-02-15
fix(spans): Do not PII-scrub code locations by default

jjbayer:
Like any other field in `span.data`, `code.filepath` etc. were scrubbed by default.

This PR types out code location fields to declare them `pii = "maybe"`, which means they can be scrubbed by Advanced Data Scrubbing, but they won't be touched by default.


relay-event-schema/src/protocol/span.rs
```diff
@@ -177,6 +165,85 @@ impl Getter for Span {
...
+    /// Usually rightmost part of the code unit's name.
+    #[metastructure(field = "code.function", pii = "maybe")]
+    pub code_function: Annotated<Value>,
+    /// The "namespace" within which `code.function` is defined.
+    ///
+    /// Usually the qualified class or module name, such that
+    /// `code.namespace + some separator + code.function`
+    /// form a unique identifier for the code unit.
+    #[metastructure(field = "code.namespace", pii = "maybe")]
+    pub code_namespace: Annotated<Value>,
```
jjbayer, line 195: We could give these fields stricter types, but

1. We don't want parsing of `span.data` to suddenly fail because of stricter types,
2. Holding a `Value` allows us to return a `&Annotated<Value>` from `.get()`, to give `SpanData`  an `Object`-like interface.

relay-event-schema/src/protocol/span.rs
```diff
@@ -177,6 +165,85 @@ impl Getter for Span {
...
+            "code.filepath" => Some(&self.code_filepath),
+            "code.lineno" => Some(&self.code_lineno),
+            "code.function" => Some(&self.code_function),
+            "code.namespace" => Some(&self.code_namespace),
+            _ => self.other.get(key),
+        }
+    }
+
+    /// Setter for compatibility with [`Object`].
+    pub fn insert(&mut self, key: String, value: Annotated<Value>) {
```
iker-barriocanal, line 215: Do you think `set` could be more specific?

relay-event-schema/src/protocol/span.rs
```diff
@@ -177,6 +165,85 @@ impl Getter for Span {
...
+    pub code_namespace: Annotated<Value>,
+
+    /// Other fields in `span.data`.
+    #[metastructure(additional_properties, pii = "true", retain = "true")]
+    other: Object<Value>,
+}
+
+impl SpanData {
+    /// Getter for compatibility with [`Object`].
+    pub fn get(&self, key: &str) -> Option<&Annotated<Value>> {
```
iker-barriocanal, line 204: Implementing both `Getter` and this `get` seems redundant, could we have a single implementation?

Dav1dde (approved):

relay-event-schema/src/protocol/span.rs
jjbayer: This mimicks the interface of `Object = BTreeMap` (except it does not return the old value).

relay-event-schema/src/protocol/span.rs
jjbayer: We should probably just type out all the `.get()` calls in our code (i.e. replace with named fields), then we can remove this method. I can do that in a follow-up PR.

iker-barriocanal (approved):

relay-event-schema/src/protocol/span.rs
```diff
@@ -177,6 +165,85 @@ impl Getter for Span {
     }
 }
 
+#[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
+#[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
+pub struct SpanData {
```
TBS1996, line 177: maybe some docs on the type itself ?

relay-event-schema/src/protocol/span.rs
```diff
@@ -177,6 +165,85 @@ impl Getter for Span {
...
+                self.other.insert(key, value);
+            }
+        }
+    }
+}
+
+impl Getter for SpanData {
+    fn get_value(&self, path: &str) -> Option<Val<'_>> {
+        Some(match path {
+            "code\\.filepath" => self.code_filepath.as_str()?.into(),
```
TBS1996, line 231: where does the `\\.` pattern come from ?

Dav1dde: It's documented on the [Getter trait](https://getsentry.github.io/relay/relay_protocol/trait.Getter.html).

relay-event-schema/src/protocol/span.rs
jjbayer: Thanks, done!

PR #3117, opened 2024-02-15
ref(project-cache): Instrument task duration

iker-barriocanal:
We have observability into how long the project cache service takes to process each message, but not each task. In other words, we know how long it takes to process _one_ branch of the `tokio::select` but not the rest. This PR adds observability there.

#skip-changelog

### DIFF
--- a/relay-server/src/services/project_cache.rs
+++ b/relay-server/src/services/project_cache.rs
@@ -1113,7 +1113,7 @@ impl Service for ProjectCacheService {
                     biased;
 
                     Ok(()) = subscription.changed() => {
-                        metric!(timer(RelayTimers::EventProcessingDeserialize), ty = "update_global_config", {
+                        metric!(timer(RelayTimers::EventProcessingDeserialize), task = "update_global_config", {
                             match subscription.borrow().clone() {
                                 global_config::Status::Ready(_) => broker.set_global_config_ready(),
                                 // The watch should only be updated if it gets a new value.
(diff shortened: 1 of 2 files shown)


relay-server/src/services/project_cache.rs
```diff
@@ -1113,20 +1113,42 @@ impl Service for ProjectCacheService {
...
+                                global_config::Status::Ready(_) => broker.set_global_config_ready(),
+                                // The watch should only be updated if it gets a new value.
+                                // This would imply a logical bug.
+                                global_config::Status::Pending => relay_log::error!("still waiting for the global config"),
+                            }
+                        })
                     },
-                    Some(message) = state_rx.recv() => broker.merge_state(message),
+                    Some(message) = state_rx.recv() => {
+                        metric!(timer(RelayTimers::ProjectCacheTaskDuration), ty = "merge_state", {
```
iker-barriocanal: I think creating a macro for just the timer is an overkill as it's easy enough to read at this point. I'm open to updating it though.

relay-server/src/statsd.rs
```diff
@@ -359,6 +359,14 @@ pub enum RelayTimers {
     ///
     ///  - `message`: The type of message that was processed.
     BufferMessageProcessDuration,
+    /// Timing in milliseconds for processing a task in the project cache service.
+    ///
+    /// A task is a unit of work the service does. Each branch of the
+    /// `tokio::select` is a different task type.
+    ///
+    /// This metric is tagged with:
+    /// - `ty`: The type of the task the processor does.
```
Dav1dde: nit: we can call it `task`

Dav1dde (approved):

Issue #2880, opened 2023-12-20
Memory watermark in health check


jjbayer:
To prevent OOMs, reply with 429s and set health check to unhealthy when system memory goes over a configured limit.

Open questions:
* If memory is full because of project configs, we should _not_ stop accepting requests, but stop fetching project configs instead. But that is the less common scenario so we could ignore it in a first step.
* Instead of becoming unhealthy, should we go into spooling state?

Dav1dde:
Spooling changes are tracked here: https://github.com/getsentry/team-ingest/issues/131

Health check watermark is implemented.

PR #3118, opened 2024-02-15
ref(health): Include system memory thresholds in health check

Dav1dde:
Issue: https://github.com/getsentry/relay/issues/2880

Includes the system memory in the Relay health check, the intent is to signal orchestrator (Kubernetes) that this Relay should not receive any more requests since it is about to run out of memory instead of just crashing because of an OOM.

#skip-changelog

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -1304,6 +1304,38 @@ impl Default for CardinalityLimiter {
     }
 }
 
+/// Settings to control Relay's health checks.
+///
+/// After breaching one of the configured thresholds, Relay will
+/// return an `unhealthy` status from its health endpoint.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(default)]
+pub struct Health {
+    /// Interval in which Relay will refresh system information, like current memory usage.
+    ///
+    /// Defaults to 3 seconds.
+    pub sys_info_refresh_interval_secs: u64,
+    /// Maximum memory watermark in bytes.
+    ///
+    /// By default there is no absolute limit set and the watermark
+    /// is only controlled by setting [`Self::max_memory_percent`].
+    pub max_memory_bytes: Option<ByteSize>,
+    /// Maximum memory watermark as a percentage of maximum system memory.
+    ///
+    /// Defaults to `0.95` (95%).
+    pub max_memory_percent: f32,
+}
+
+impl Default for Health {
+    fn default() -> Self {
+        Self {
+            sys_info_refresh_interval_secs: 3,
+            max_memory_bytes: None,
+            max_memory_percent: 0.95,
+        }
+    }
+}
+
 #[derive(Serialize, Deserialize, Debug, Default)]
 struct ConfigValues {
     #[serde(default)]
@@ -1342,6 +1374,8 @@ struct ConfigValues {
     geoip: GeoIpConfig,
     #[serde(default)]
     cardinality_limiter: CardinalityLimiter,
+    #[serde(default)]
+    health: Health,
 }
 
 impl ConfigObject for ConfigValues {
@@ -2165,6 +2199,25 @@ impl Config {
     /// The cache will scan for expired values based on this interval.
     pub fn cardinality_limiter_cache_vacuum_interval(&self) -> Duration {
         Duration::from_secs(self.values.cardinality_limiter.cache_vacuum_interval)
+    }
+
+    /// Maximum memory watermark as a percentage of maximum system memory.
+    pub fn health_sys_info_refresh_interval(&self) -> Duration {
+        Duration::from_secs(self.values.health.sys_info_refresh_interval_secs)
+    }
+
+    /// Maximum memory watermark in bytes.
+    pub fn health_max_memory_watermark_bytes(&self) -> u64 {
+        self.values
+            .health
+            .max_memory_bytes
+            .as_ref()
+            .map_or(u64::MAX, |b| b.as_bytes() as u64)
+    }
+
+    /// Maximum memory watermark as a percentage of maximum system memory.
+    pub fn health_max_memory_watermark_percent(&self) -> f32 {
+        self.values.health.max_memory_percent
     }
 
     /// Creates an [`AggregatorConfig`] that is compatible with every other aggregator.
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -2201,7 +2201,7 @@ impl Config {
         Duration::from_secs(self.values.cardinality_limiter.cache_vacuum_interval)
     }
 
-    /// Maximum memory watermark as a percentage of maximum system memory.
+    /// Interval to refresh system information.
     pub fn health_sys_info_refresh_interval(&self) -> Duration {
         Duration::from_secs(self.values.health.sys_info_refresh_interval_secs)
     }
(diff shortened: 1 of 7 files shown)


relay-server/Cargo.toml
```diff
@@ -109,6 +109,7 @@ symbolic-common = { version = "12.1.2", optional = true, default-features = fals
 symbolic-unreal = { version = "12.1.2", optional = true, default-features = false, features = [
     "serde",
 ] }
+sysinfo = "0.30"
```
Dav1dde, line 112: I reviewed the code path Relay will be using (memory), the crate is maintained by an author of the Rust foundation.

relay-server/src/services/health_check.rs
```diff
@@ -138,3 +207,130 @@ impl Service for HealthCheckService {
...
+            system,
+            last_refresh: Instant::now(),
+            refresh_interval,
+        }
+    }
+
+    pub fn memory(&mut self) -> Memory {
+        self.refresh();
+
+        if let Some(cgroup) = self.system.cgroup_limits() {
```
Dav1dde, line 267: Need to use cgroups or we have a problem in containers

relay-config/src/config.rs
```diff
@@ -2159,6 +2193,25 @@ impl Config {
         Duration::from_secs(self.values.cardinality_limiter.cache_vacuum_interval)
     }
 
+    /// Maximum memory watermark as a percentage of maximum system memory.
```
olksdr: This doc has to be updated.

olksdr (approved):
Let's test it maybe on S4S and canary before fully rolling it out.

relay-server/src/services/health_check.rs
```diff
@@ -138,3 +240,95 @@ impl Service for HealthCheckService {
...
+        }
+    }
+
+    pub fn memory(&mut self) -> Memory {
+        self.refresh();
+
+        // Use the cgroup if available in case Relay is running in a container.
+        if let Some(cgroup) = self.system.cgroup_limits() {
+            Memory {
+                used: cgroup.total_memory.saturating_sub(cgroup.free_memory),
```
jjbayer, line 269: Not a blocker, but I wonder if we could base the health check on https://docs.rs/sysinfo/latest/sysinfo/struct.Process.html#method.memory, which returns the resident memory of the process.

This would require an additional indirection because we would need to configure what percentage of total memory our process is allowed to occupy. But it would prevent us from reporting unhealthy when Relay uses all the available memory for caching.

Dav1dde: We actually want to use system memory not process memory. It doesn't matter wether the container/host runs out of memory because Relay consumes all or something else is. In either case we want to stop accepting requests.

relay-server/src/services/health_check.rs
```diff
@@ -22,15 +24,60 @@ pub enum IsHealthy {
...
+pub enum Status {
+    /// Relay is healthy.
+    Healthy,
+    /// Relay is unhealthy.
+    Unhealthy,
+}
+
+impl Status {
+    fn combined(s: &[Status]) -> Self {
+        s.iter()
```
TBS1996: couldn't we just use `any` ? 
 

``` Rust
    fn combined(s: &[Status]) -> Self {
        match s.iter().any(|status| status == Self::Unhealthy) {
            true => Self::Unhealthy,
            false => Self::Healthy,
        }
    }
```

this seems easier to read 


alternatively:

``` Rust
    fn combined(s: &[Status]) -> Self {
        Self::from(s.iter().all(|status| status == Status::Healthy))
    }
```

relay-server/src/services/health_check.rs
```diff
@@ -138,3 +240,95 @@ impl Service for HealthCheckService {
...
+            self.system
+                .refresh_memory_specifics(MemoryRefreshKind::new().with_ram());
+
+            self.last_refresh = Instant::now();
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Memory {
```
TBS1996: docs?

tests/integration/test_healthchecks.py
TBS1996, line 1: how about a test that checks it can recover from OOM? like trigger oom, wait until the events left relay, verify that it becomes healthy again

relay-server/src/services/health_check.rs
```diff
@@ -76,39 +183,34 @@ impl HealthCheckService {
...
+        Status::combined(&[sys_mem, auth, agg, proj])
     }
 
     async fn handle_message(&self, message: HealthCheck) {
         let HealthCheck(message, sender) = message;
-        let response = self.handle_is_healthy(message).await;
+
+        let ty = message.variant();
+        let response = relay_statsd::metric!(
+            timer(RelayTimers::HealthCheckDuration),
```
iker-barriocanal, line 201: Good metric 😄

relay-server/src/services/health_check.rs
```diff
@@ -76,39 +183,34 @@ impl HealthCheckService {
...
 
-                if !self
-                    .aggregator
-                    .send(AcceptsMetrics)
-                    .await
-                    .unwrap_or_default()
-                {
-                    return false;
-                }
+        let (sys_mem, auth, agg, proj) = tokio::join!(
```
iker-barriocanal, line 186: nit: I wonder if we could use [`tokio::try_join`](https://docs.rs/tokio/latest/tokio/macro.try_join.html) to short circuit on a failed future and not wait on the others.

relay-server/src/services/health_check.rs
```diff
@@ -138,3 +240,95 @@ impl Service for HealthCheckService {
...
+            Memory {
+                used: self.system.used_memory(),
+                total: self.system.total_memory(),
+            }
+        }
+    }
+
+    fn refresh(&mut self) {
+        let now = Instant::now();
+        if (now - self.last_refresh) >= self.refresh_interval {
```
iker-barriocanal: Could this subtraction lead to a panic? I've checked some alternatives in the [docs](https://doc.rust-lang.org/std/time/struct.Instant.html), but there's a note in all of them that they may reintroduce panics in the future.

For readability I rather `self.last_refresh.elapsed()`, but not strongly opinionated.

iker-barriocanal (approved):

relay-server/src/services/health_check.rs
Dav1dde: Yeah, I thought about this too but didn't do it because of the awkward signature, would probably have to return `Result<(), ()>` from every probe.

Do you think I should change it?

relay-server/src/services/health_check.rs
Dav1dde: I forgot about `elapsed`, that is the winner here 👍

relay-server/src/services/health_check.rs
iker-barriocanal: Nah, I don't think it's worth the effort (that's why nit), I'd just ship it. If we bikeshed, a `Result` could make sense from a semantical point of view (a failed health check could be an error?), plus allow the `try_join`.

relay-server/src/services/health_check.rs
Dav1dde: Nice, thanks!

PR #3122, opened 2024-02-15
feat(on-demand): Extract user by normalizing data

### LINKED ISSUE #3121: Implement field getter for user using custom function
In on demand metrics we are extracting `user.id` as a field when trying to extract metrics for `user_misery()`. However, we noticed that `user.id` is not the same `user` field used by discover when querying `user_misery()` for this reason we would need to figure out a way to get the mutated user with this function (https://github.com/getsentry/relay/blame/01c7b2afdbfab7dc68a3539e1736aee6e395babf/relay-event-normalization/src/normalize/utils.rs#L134C9-L134C9) in the event getter.

iambriccardo:
This PR implements a normalized user field called `sentry_user` that has the goal of emulating the value that the `user` field has in the Sentry db. The rationale behind this implementation is that on demand metrics need to access the user field in a Sentry-specific format to have the same results as querying directly the data via Discover.

Closes: https://github.com/getsentry/relay/issues/3121


relay-event-schema/src/protocol/user.rs
```diff
@@ -76,6 +76,10 @@ pub struct User {
     #[metastructure(skip_serialization = "empty")]
     pub segment: Annotated<String>,
 
+    /// The user string representation as handled in Sentry.
```
jjbayer, line 79: nit: Could you add some examples here (as in the `get_event_user_tag` doc comment)?

relay-event-normalization/src/event.rs
```diff
@@ -2592,23 +2598,18 @@ mod tests {
...
-        });
-
-        normalize_user(&mut user);
-
-        let user = user.value().unwrap();
+    fn test_normalize_user() {
+        let json = r#"{
+            "user": {
+                "id": "123456",
+                "username": "john",
```
iambriccardo, line 2686: Added `username` to validate the priority of fields given in the `get_event_user_tag` function used in `normalize_user`.

relay-event-normalization/src/event.rs
```diff
@@ -373,14 +373,20 @@ pub fn normalize_user_geoinfo(geoip_lookup: &GeoIpLookup, user: &mut User) {
...
     };
+
     if !user.other.is_empty() {
         let data = user.data.value_mut().get_or_insert_with(Object::new);
         data.extend(std::mem::take(&mut user.other));
     }
+
+    // We set the `sentry_user` field in the `Event` payload in order to have it ready for the extraction
+    // pipeline.
+    let eventuser_tag = get_eventuser_tag(user);
```
jjbayer: Please also replace other uses of `get_eventuser_tag` with direct field access. We can even make the function private then.

jjbayer (approved):

relay-event-schema/src/protocol/user.rs
iambriccardo: Yess, will do. Should I also rename the function to `get_event_user_tag` it bugs me as it is written rn.

relay-event-normalization/src/event.rs
iambriccardo: Are we 100% sure that all the code paths where it's used, are happening afterwards light normalization? I checked the code and it seems like so.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 **Features**:
 
 - Add protobuf support for ingesting OpenTelemetry spans and use official `opentelemetry-proto` generated structs. ([#3044](https://github.com/getsentry/relay/pull/3044))
+- Add `user.sentry_user` computed field for the on demand metrics extraction pipeline. ([#3122](https://github.com/getsentry/relay/pull/3122))
```
jjbayer: Please merge master into your branch and move this to the "unreleased" section.

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -1891,13 +1895,13 @@ mod tests {
...
         assert_eq!(
-            normalize_utils::get_eventuser_tag(&user).unwrap(),
+            normalize_utils::get_event_user_tag(&user).unwrap(),
             "ip:127.0.0.1"
         );
 
         let user = User::default();
 
-        assert!(normalize_utils::get_eventuser_tag(&user).is_none());
+        assert!(normalize_utils::get_event_user_tag(&user).is_none());
```
jjbayer: Looks like these tests can all be moved to `normalize_utils`.

jjbayer (approved):

Dav1dde (approved):

PR #3124, opened 2024-02-16
feat(webvitals): Adds Google Chrome to the list of allowed browser name tags for extracting inp metrics from spans

edwardgou-sentry:
Adds "Google Chrome" to the list of allowed browser name tags for extracting inp metrics from spans. This is required because the extracted browser name using client hints is "Google Chrome" for chrome browsers.

#skip-changelog


edwardgou-sentry:
Chrome browser names for transactions is just "Chrome". Alternatively we can map "Google Chrome" to "Chrome" when parsing the user agent for browser name, but I feel like that would be more messy.

Dav1dde (approved):
I am curious where this difference between `Chrome` and `Google Chrome` is coming from (Chromium vs Google Chrome?) and wether it would make sense to unify them. 

Looking a bit through the code base, we already have `Google Chrome` elsewhere, unifying seems out of scope for now.

TBS1996 (approved):

iker-barriocanal (commented):
I'd look into normalizing "Google Chrome" to "Chrome" instead, as this also provides some indirect benefits (e.g. reduced cardinality in these fields). Is there any reason to not normalize the name, from the product pov?

edwardgou-sentry:
> I'd look into normalizing "Google Chrome" to "Chrome" instead, as this also provides some indirect benefits (e.g. reduced cardinality in these fields). Is there any reason to not normalize the name, from the product pov?

I don't think there is currently a product reason to have to use `Google Chrome`, but it feels more correct to me since that's the actual value that comes from the client hints. I don't have anything against normalizing to `Chrome` though.

I didn't look at how `browser.name` is generated for Transactions, but it seems to be done differently (because it's a different value). If we're abandoning Transactions in the future in favour of spans, maybe it makes sense to stick to the `Google Chrome` value from client hints, and then `Chrome` will naturally disappear? This might not be true or happen quickly though, because there could be existing Transaction product areas that rely on the browser name being `Chrome`.

@iker-barriocanal lmk if you think this is a blocker for this pr. I just wanted to add this in since it's blocking some of the work im doing with webvitals

DominikB2014 (approved):

iker-barriocanal (approved):
Thanks for the context @edwardgou-sentry! Not blocking. I'm not sure what it'll look like with only spans, but this aggregation improvement could have a nice positive impact.

Thanks for jumping and opening the PR too!

PR #3125, opened 2024-02-16
fix(spans): Do not copy parsed SQL

jjbayer:
All stack overflows after upgrading to sqlparser 0.43 occurred in `.clone()`, not during parsing (see [POP-RELAY-30P](https://sentry.my.sentry.io/organizations/sentry/issues/645286/)).

We cloned the parsed query to reuse it for table name extraction. But table name extraction should work just as well on the scrubbed query, so the clone is unnecessary.

This PR also reverts https://github.com/getsentry/relay/pull/3115.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -2,7 +2,8 @@
 mod resource;
 mod sql;
 use once_cell::sync::Lazy;
-pub use sql::parse_query;
+#[cfg(test)]
+pub use sql::{scrub_queries, Mode};
 
 use std::borrow::Cow;
 use std::path::Path;
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -20,9 +20,6 @@ use crate::span::tag_extraction::HTTP_METHOD_EXTRACTOR_REGEX;
 /// Dummy URL used to parse relative URLs.
 static DUMMY_BASE_URL: Lazy<Url> = Lazy::new(|| "http://replace_me".parse().unwrap());
 
-/// Very large SQL queries may cause stack overflows in the parser, so do not attempt to parse these.
-const MAX_DESCRIPTION_LENGTH: usize = 10_000;
-
 /// Maximum length of a resource URL segment.
 ///
 /// Segments longer than this are treated as identifiers.
@@ -40,14 +37,6 @@ pub(crate) fn scrub_span_description(
     let Some(description) = span.description.as_str() else {
         return (None, None);
     };
-
-    if description.len() > MAX_DESCRIPTION_LENGTH {
-        relay_log::error!(
-            description = description,
-            "Span description too large to parse"
-        );
-        return (None, None);
-    }
 
     let data = span.data.value();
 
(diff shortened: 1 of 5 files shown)


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -922,17 +912,6 @@ mod tests {
...
-    span_description_test!(
-        long_description_none,
-        // Do not attempt to parse very long descriptions.
-        {
-            let repeated = "+1".repeat(5000);
-            &("SELECT 1".to_string() + &repeated)
-        },
-        "db.query",
-        ""
-    );
```
jjbayer, line 934: This constructed case still causes a stack overflow in the visitor, but we haven't seen those cases in the wild. Nevertheless, I will report this problem upstream.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -41,14 +39,6 @@ pub(crate) fn scrub_span_description(
         return (None, None);
     };
 
-    if description.len() > MAX_DESCRIPTION_LENGTH {
-        relay_log::error!(
-            description = description,
-            "Span description too large to parse"
-        );
-        return (None, None);
-    }
```
jjbayer, line 50: Our `NormalizeVisitor` is actually quite good at reducing long queries, which usually contain lists of `(expr, expr, expr, expr)` that are collapsed to `..`. Only deep queries cause problems.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -63,7 +63,6 @@ pub fn normalize_parsed_queries(
     string: &str,
 ) -> Result<(String, Vec<Statement>), ()> {
     let mut parsed = parse_query(db_system, string).map_err(|_| ())?;
-    let original_ast = parsed.clone();
```
jjbayer, line 66: Cloning heavily nested queries causes stack overflows. So does dropping them (because of the `Box<..>` drop handler). But fortunately, the `NormalizeVisitor` and `MaxDepthVisitor` run before we drop the AST.

Dav1dde: So this woulld still overflow the stack if the two normalizers wouldn't actually end up throwing parts away already?

iker-barriocanal (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
```diff
@@ -75,7 +74,7 @@ pub fn normalize_parsed_queries(
     // Insert placeholders that the SQL serializer cannot provide.
     let replaced = concatenated.replace("___UPDATE_LHS___ = NULL", "..");
 
-    Ok((replaced, original_ast))
+    Ok((replaced, parsed))
```
Dav1dde, line 77: This is fine because the returned SQL is only ever used to extract table names from it?

I think it would be good to document that the returned value from `scrub_span_description` contains a parsed and modified SQL. This is quite a call chain to go through to figure out.

Dav1dde (approved):

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: The visitors would, yes. I'll suggest upstream to make the visitor non-recursive. https://github.com/getsentry/relay/pull/3125#discussion_r1492096260

Dav1dde: But also the drop impl?

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: 👍 I'll do a follow-up PR to simplify the call chain: parse & scrub the SQL in `tag_extraction`, then pass it into both `scrub_span_description` and `sql_action_from_query`.

relay-event-normalization/src/normalize/span/description/sql/parser.rs
jjbayer: Yes. Though if the visitor succeeds, the expression depth is limited, so it is less likely.

PR #3128, opened 2024-02-16
chore(processor): Introduce typed envelope for processing

olksdr:
This change adds a small typed wrapper for managed envelope which is used in the processor.
Also add a specific marker trait to tighten restriction for function which used only for event based processing. 

Also, 
- remove the types from within the `ProcessingGroup` enum variants, they are not used and it's not necessary for now
- ensures that the envelope can be stored or sent to upstream only if it's processed in the processor
 
There is no functional changes in this PR.

related: https://github.com/getsentry/team-ingest/issues/253

#skip-changelog


relay-server/src/services/processor.rs
```diff
@@ -147,38 +157,44 @@ pub struct ForwardUnknownGroup;
...
 /// Describes the groups of the processable items.
 #[derive(Clone, Copy, Debug)]
 pub enum ProcessingGroup {
     /// All the transaction related items.
     ///
     /// Includes transactions, related attachments, profiles.
-    Transaction(TransactionGroup),
-    /// All the items which require (have or create) events.
+    Transaction,
+    /// All the items which require g events.
```
jjbayer: False positive?

relay-server/src/services/processor.rs
```diff
@@ -147,38 +157,44 @@ pub struct ForwardUnknownGroup;
...
-    CheckIn(CheckInGroup),
+    CheckIn,
     /// Spans.
-    Span(SpanGroup),
+    Span,
     /// Metrics.
-    Metrics(MetricsGroup),
-    /// Unknown item types will be forwarded upstream (to processing Relay), where we will
+    Metrics,
+    /// Unknown item types will be forwarded upstream g, where we will
```
jjbayer: False positive?

jjbayer (approved):

Dav1dde (approved):

relay-server/src/services/processor.rs
olksdr: Yeah, find/replace didn't properly workout 😬  
Thanks!

relay-server/src/utils/managed_envelope.rs
```diff
@@ -65,6 +68,49 @@ struct EnvelopeContext {
     group: ProcessingGroup,
 }
 
+/// A wraper for [`ManagedEnvelope`] with assigned processing group type.
```
TBS1996, line 71:
```suggestion
/// A wrapper for [`ManagedEnvelope`] with assigned processing group type.
```

TBS1996 (approved):

PR #3129, opened 2024-02-16
ref(spans): Type out more data fields

jjbayer:
Follow up to https://github.com/getsentry/relay/pull/3116, based on discussion here: https://github.com/getsentry/relay/pull/3116#discussion_r1491069595.

> We should probably just type out all the .get() calls in our code (i.e. replace with named fields), then we can remove this method. I can do that in a follow-up PR.

#skip-changelog

### DIFF
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -175,11 +175,13 @@ impl Getter for Span {
 #[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
 #[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
 pub struct SpanData {
-    /// TODO: docs
+    /// Mobile app start variant.
+    ///
+    /// Can be either "cold" or "warm".
     #[metastructure(field = "app_start_type")] // TODO: no dot?
     pub app_start_type: Annotated<Value>,
 
-    /// TODO: docs
+    /// The client's browser name.
     #[metastructure(field = "browser.name")]
     pub browser_name: Annotated<Value>,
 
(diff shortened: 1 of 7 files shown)


relay-event-schema/src/protocol/span.rs
```diff
@@ -175,6 +175,16 @@ impl Getter for Span {
 #[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
 #[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
 pub struct SpanData {
+    /// Mobile app start variant.
+    ///
+    /// Can be either "cold" or "warm".
+    #[metastructure(field = "app_start_type")] // TODO: no dot?
```
jjbayer, line 181:
```suggestion
    #[metastructure(field = "app_start_type")]
```

relay-event-schema/src/protocol/span.rs
```diff
@@ -175,6 +175,16 @@ impl Getter for Span {
...
+    pub app_start_type: Annotated<Value>,
```
jjbayer, line 182: All these fields are weakly typed, because they can theoretically contain any data, even if we have conventions around them.

relay-event-schema/src/protocol/span.rs
```diff
@@ -194,44 +204,103 @@ pub struct SpanData {
...
-        match key.as_str() {
-            "code.filepath" => self.code_filepath = value,
-            "code.lineno" => self.code_lineno = value,
-            "code.function" => self.code_function = value,
-            "code.namespace" => self.code_namespace = value,
-            _ => {
-                self.other.insert(key, value);
-            }
-        }
-    }
```
jjbayer, line 225: We access fields directly now.

relay-event-schema/src/protocol/span.rs
```diff
@@ -194,44 +204,103 @@ pub struct SpanData {
...
+            "http\\.response.status_code" | "status_code" => {
+                self.http_response_status_code.value()?.into()
+            }
+            "resource\\.render_blocking_status" => {
+                self.resource_render_blocking_status.value()?.into()
+            }
+            "server\\.address" => self.server_address.value()?.into(),
+            "thread\\.name" => self.thread_name.value()?.into(),
+            "ui\\.component_name" => self.ui_component_name.value()?.into(),
+            "url\\.scheme" => self.url_scheme.value()?.into(),
```
jjbayer, line 302: Would be nice to write a `derive` for these getters.

relay-event-schema/src/protocol/span.rs
```diff
@@ -432,11 +501,10 @@ mod tests {
 
         assert_eq!(span.get_value("span.duration"), Some(Val::F64(477.800131)));
     }
-}
 
-#[test]
-fn test_span_data() {
-    let data = r#"{
+    #[test]
+    fn test_span_data() {
```
jjbayer, line 506: This test was outside `mod tests` by accident.

iker-barriocanal (approved):
Thank you for following up on this!

olksdr (approved):

Dav1dde (approved):

PR #3131, opened 2024-02-17
Rearchitect GoCD pipelines

### LINKED ISSUE #2939: Improve rollback and immediate rollout pipelines
The Relay deployment pipeline sequentially applies all clusters. For faster changes in one of the clusters or rollbacks, we'll need changes to this pipeline or a new pipeline. 

**Main Goals:**
```[tasklist]
### Tasks
- [x] Add a trigger for fast rollbacks
- [x] Introduce a canary stages for Processing and PoPs
```

**Testing the new pipelines:**
```[tasklist]
### Tasks
- [x] Create the new pipelines
- [x] Test S4S deploys via the new pipelines (redeploying what is currently deployed)
- [x] Test SaaS canary deploys via the new pipelines (redeploying what is currently deployed)
- [x] Test SaaS prod deploys via the new pipelines (redeploying what is currently deployed)
- [x] Test ST deploys via the new pipelines (redeploying what is currently deployed)
- [x] Test rollback pipeline functionality (redeploying what is currently deployed)
- [x] Test rollback pipelines simultaneously?
- [x] Get the pipelines in the hands of the ingest team
```

IanWoodard:
In this PR, I've restructured Relay's GoCD pipelines to separate Processing And PoPs into distinct pipelines. As we previously discussed, this change aims to improve efficiency and robustness of Relay's deployment process.

Key enhancements include:
* Addition of canary deployments for both Processing and PoPs, providing an early warning system for issues in our deployment process
* Introduction of soak-time stage with Sentry and DataDog checks, allowing us to detect and address issues early on and stop problematic deploys from rolling out further.
* Inclusion of dedicated rollback pipelines for both Processing and PoPs, increasing the speed at which we can mitigate issues.

While significant progress has been made, there are still areas that need finishing up before this can be rolled out to production. I've outlined these below as well as in the code via TODOs:
- [x] Adding DataDog monitor IDs
- [ ] Setting appropriate error limits for the Sentry checks
- [x] Double-check my changes : )

On my end, I need to update the Sentry checks to accommodate s4s, which is where Relay sends its Sentry events.
I am also aware that I didn't include an experimental deploy. To implement this without interfering with the new canary setup, we'll need to allocate separate pods.

#skip-changelog


olksdr (commented):
Before merging this I have few questions: 
* How do you plan to test this? 
* There are many `TODOs` still in here, are you planning to address all of them before merging or just update we you go with tests? 
* Can we keep our current deploy pipeline before this one production ready? - this one is  important since the team still has to work, deploy and debug things while the deploy pipeline is being prepared


But overall this looks ok to me, cannot comment much on the external lib (`pipedream`) though and what it does.

IanWoodard:
> Before merging this I have few questions:
> 
> * How do you plan to test this?
> * There are many `TODOs` still in here, are you planning to address all of them before merging or just update we you go with tests?
> * Can we keep our current deploy pipeline before this one production ready? - this one is  important since the team still has to work, deploy and debug things while the deploy pipeline is being prepared
> 
> But overall this looks ok to me, cannot comment much on the external lib (`pipedream`) though and what it does.

To answer your questions:

* My plan for testing the new Relay deployment pipelines, which primarily involve structural changes while retaining the same deploy scripts is to include both the new and old pipelines to enable new deploys to go out while the new pipelines are being tested. Initially, I plan on pausing all regions except for S4S in the new pipelines to prevent unintended deployments from going out. The stage of testing involves re-deploying S4S to the same version it is already on via the new pipelines to verify functionality Following this, we would proceed to test that the SaaS canaries for Processing and PoPs are able to deploy successfully. Similar to the S4S test, this would involve deploying the same version deployed elsewhere in SaaS since it is known to be stable. Success in these stages should allow us to extend testing to the rest of the pipelines. The logic behind the canary (besides the deploy portion) and soak stages is already well tested and shouldn't need additional testing. With that said, we will need to tune the thresholds for the monitoring to ensure it catches issues without blocking deploys unnecessarily. This doesn't need to be done before we start using the new pipelines, but we should get those tuned and turned on relatively quickly so we can reap the benefits. Lastly, we need to test the rollback pipeline, which is already well tested in other services, but should be tested nonetheless. To test this while minimizing disruptions I think we should rollback to what is currently deployed. While this isn't something we would ever do in reality, it does ensure the functionality works as expected. Assuming Ingest is onboard with this plan, I could carry out most of this testing during US hours since it should only involve redeploying the service and not deploying any new changes.
* I am cleaning up the remaining TODOs, some require Ingest's help, but I'll leave comments on the PR asking those question to make it easier.
* Of course! The plan is to keep the old pipeline around until the new ones are proven to work.

gocd/templates/libs/pops.libsonnet
```diff
@@ -0,0 +1,208 @@
+local utils = import './utils.libsonnet';
+local gocdtasks = import 'github.com/getsentry/gocd-jsonnet/libs/gocd-tasks.libsonnet';
+
+local canary_region_pops = {
+  de: [],
+  // TODO: Check that these are right
+  us: ['us-pop-1', 'us-pop-regional-1'],
```
IanWoodard: @beezz Does this look right?

olksdr:
@IanWoodard  thanks for the laying out  such a detailed plan. This sounds good to me! 

@getsentry/ingest does anyone have any concerns or followup questions?

iker-barriocanal:
The plan makes sense to me and seems well thought out. My only follow-up question: how long would it take to test the new pipelines and keep them paused?

gocd/templates/bash/check-sentry-errors.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# TODO: Make this work with s4s
+/devinfra/scripts/checks/sentry/release_error_events.py \
+  --project-id="${SENTRY_PROJECT_ID}" \
+  --project-slug="${SENTRY_PROJECT}" \
```
jjbayer: Is it possible to check multiple projects at once? PoPs could cause issues in processing relays, and vice versa.

gocd/templates/bash/check-sentry-errors.sh
```diff
@@ -0,0 +1,12 @@
...
+  --release="relay@${GO_REVISION_GETSENTRY_REPO}" \
+  --duration=5 \
+  --error-events-limit="${ERROR_LIMIT}" \
```
jjbayer: IMO the best would be to check for a _change_ in the number of error events. Would that be feasible?

jjbayer (approved):
Left two questions, but looks good to me overall! I think we should just start using the new pipelines and see how well they work in practice.

IanWoodard:
> The plan makes sense to me and seems well thought out. My only follow-up question: how long would it take to test the new pipelines and keep them paused?

The new pipelines should only take a few days to test. During this time, the old pipelines will remain fully functional.

gocd/templates/bash/check-sentry-errors.sh
IanWoodard: Yep! That is definitely possible! I'll update the code.

gocd/templates/bash/check-sentry-errors.sh
IanWoodard: While this is definitely feasible, it would require some additional work to get implemented. This could be a good candidate for a future improvement we make. How does that sound?

IanWoodard:
> Left two questions, but looks good to me overall! I think we should just start using the new pipelines and see how well they work in practice.

Yep! The plan is to get the new pipelines in your hands at some point next week.

gocd/templates/bash/check-sentry-errors.sh
jjbayer: Sounds good!

PR #3132, opened 2024-02-17
feat(performance): Extract transaction from span attributes if the span is an inp

edwardgou-sentry:
Extracts the `transaction` attribute from standalone INP spans as a tag.

#skip-changelog


edwardgou-sentry:
I'm not sure if this is where we want to handle extracting transaction name.

I added this to `extract_tags` because `extract_tags` is used for normalizing standalone spans.
Neither `extract_shared_tags` or `extract_span_tags` get used on standalone spans.

iker-barriocanal (approved):
The diff LGTM.

Do we have an estimation of the impact this could have on cardinality? What are the consequences if cardinality explodes, and what's the action plan to address these? Approving to unblock, but please have an estimation of the consequences and a way forward if the cardinality ends up being higher than expected.

jjbayer (commented):
@edwardgou-sentry could you please add a test for this? You could for example add another span to [`test_span_ingestion`](https://github.com/getsentry/relay/blob/751154e03eb55e27a29950c7126547c6090e634c/tests/integration/test_store.py#L1452-L1455) and see if the resulting spans & metrics match your expectation.

edwardgou-sentry:
> @edwardgou-sentry could you please add a test for this? You could for example add another span to [`test_span_ingestion`](https://github.com/getsentry/relay/blob/751154e03eb55e27a29950c7126547c6090e634c/tests/integration/test_store.py#L1452-L1455) and see if the resulting spans & metrics match your expectation.

Hey @jjbayer , so I added a test in this commit:
https://github.com/getsentry/relay/pull/3132/commits/73722bd35229bd07129a780195d3c079f85d45fd

However, it seems like CI is complaining:
```
    error[E0599]: no method named `get` found for reference `&SpanData` in the current scope
       --> relay-event-normalization/src/normalize/span/tag_extraction.rs:453:43
        |
    453 |                     .and_then(|data| data.get("transaction"))
        |                                           ^^^ method not found in `&SpanData`
```

What's odd is that this is from my initial commit, which passed CI just a couple days ago. I'm not really sure why it's failing now. Building, running, and running the tests on local seems to have no issues for me. Any ideas on what could be going on here?

edwardgou-sentry:
> The diff LGTM.
> 
> Do we have an estimation of the impact this could have on cardinality? What are the consequences if cardinality explodes, and what's the action plan to address these? Approving to unblock, but please have an estimation of the consequences and a way forward if the cardinality ends up being higher than expected.

I think we have some existing auto `transaction` parameterization(classifier?) that already happens on certain spans, like resource spans. This should help with reducing cardinality.

Would my changes here automatically benefit from the transaction name classifier? Maybe @jjbayer would know?

jjbayer:
> > @edwardgou-sentry could you please add a test for this? You could for example add another span to [`test_span_ingestion`](https://github.com/getsentry/relay/blob/751154e03eb55e27a29950c7126547c6090e634c/tests/integration/test_store.py#L1452-L1455) and see if the resulting spans & metrics match your expectation.
> 
> Hey @jjbayer , so I added a test in this commit: [73722bd](https://github.com/getsentry/relay/commit/73722bd35229bd07129a780195d3c079f85d45fd)
> 
> However, it seems like CI is complaining:
> 
> ```
>     error[E0599]: no method named `get` found for reference `&SpanData` in the current scope
>        --> relay-event-normalization/src/normalize/span/tag_extraction.rs:453:43
>         |
>     453 |                     .and_then(|data| data.get("transaction"))
>         |                                           ^^^ method not found in `&SpanData`
> ```
> 
> What's odd is that this is from my initial commit, which passed CI just a couple days ago. I'm not really sure why it's failing now. Building, running, and running the tests on local seems to have no issues for me. Any ideas on what could be going on here?

Did you rebase on master? `SpanData` does not have a `.get()` method any more, so you'll have to declare a `transaction` field on `SpanData` like the other fields there.

jjbayer:
> Would my changes here automatically benefit from the transaction name classifier? Maybe @jjbayer would know?

@edwardgou-sentry good catch, transaction name normalization does not run on spans yet. You'll have to pass the transaction rename rules from project config into `NormalizeSpanConfig` and then call a variant of [normalize_transaction_name](https://github.com/getsentry/relay/blob/73722bd35229bd07129a780195d3c079f85d45fd/relay-event-normalization/src/transactions/processor.rs#L90) from within [`span::process::normalize`](https://github.com/getsentry/relay/blob/73722bd35229bd07129a780195d3c079f85d45fd/relay-server/src/services/processor/span/processing.rs#L293). Let me know if you'd like more help on this.

https://github.com/getsentry/relay/blob/73722bd35229bd07129a780195d3c079f85d45fd/relay-dynamic-config/src/project.rs#L76

https://github.com/getsentry/relay/blob/73722bd35229bd07129a780195d3c079f85d45fd/relay-server/src/services/processor/span/processing.rs#L292-L293

relay-dynamic-config/src/defaults.rs
```diff
@@ -139,7 +139,11 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
             mri: "d:spans/exclusive_time@millisecond".into(),
             field: Some("span.exclusive_time".into()),
             condition: Some(
-                (is_db.clone() | is_resource.clone() | is_mobile.clone() | is_http.clone())
+                (is_db.clone()
+                    | is_resource.clone()
+                    | is_mobile.clone()
+                    | is_http.clone()
+                    | is_interaction)
```
jjbayer: @edwardgou-sentry `d:spans/exclusive_time@millisecond` is the metric that has the `"transaction"` tag. I assume we want that metric enabled for interactions?

relay-event-schema/src/protocol/span.rs
```diff
@@ -257,6 +257,10 @@ pub struct SpanData {
     #[metastructure(field = "thread.name")]
     pub thread_name: Annotated<Value>,
 
+    /// Origin Transaction name of the span.
+    /// For INP spans, this is the route name where the interaction occurred.
+    pub transaction: Annotated<String>,
```
jjbayer, line 263: Typing this as a `String` makes parsing fail if users put something else in there, but the normalization functions expect an `Annotated<String>`.

tests/integration/test_store.py
```diff
@@ -2081,6 +2106,30 @@ def test_span_ingestion_with_performance_scores(
...
         )
     )
+    envelope.add_item(
+        Item(
+            type="span",
+            payload=PayloadRef(
+                bytes=json.dumps(
+                    {
+                        "data": {
+                            "transaction": "/page/with/click/interaction/jane/123"
```
jjbayer, line 2116: `"123"` will be scrubbed by the generic scrubber, `"jane"` will be scrubbed by the project config rule defined in this test.

relay-event-schema/src/protocol/span.rs
```diff
@@ -257,6 +257,10 @@ pub struct SpanData {
     #[metastructure(field = "thread.name")]
     pub thread_name: Annotated<Value>,
 
+    /// Origin Transaction name of the span.
+    /// For INP spans, this is the route name where the interaction occurred.
```
iker-barriocanal: nit: the first line should be separted from the rest, according to [Relay code guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#987c0416a9f54674999eed59eecbf846):
```suggestion
    /// Origin Transaction name of the span.
    ///
    /// For INP spans, this is the route name where the interaction occurred.
```

relay-dynamic-config/src/defaults.rs
iker-barriocanal: Will this change result in extracting metrics for all `ui.interaction*`? Is this ok, or should we limit it to INP spans?

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -445,6 +445,15 @@ pub fn extract_tags(
                 }
             }
         }
+        if let Some(measurements) = span.measurements.value() {
+            if span_op.starts_with("ui.interaction") && measurements.contains_key("inp") {
```
iker-barriocanal: Note: the metric spec is defined for `ui.interaction.`:
```suggestion
            if span_op.starts_with("ui.interaction.") && measurements.contains_key("inp") {
```
It probably doesn't make much difference, but we should align these.

iker-barriocanal (approved):

relay-dynamic-config/src/defaults.rs
jjbayer: Good question, maybe adding `exclusive_time` light for interaction spans isn't even required, and the `"transaction"` tag is already on metrics like `d:spans/webvital.score.inp@ratio`.

I'll let @edwardgou-sentry decide.

edwardgou-sentry: Hey @jjbayer , I think I'll just need the following metrics to be tagged by `transaction`:
`d:spans/webvital.inp@millisecond`
`d:spans/webvital.score.weight.inp@ratio`
`d:spans/webvital.score.inp@ratio`
The metrics above should only exist on INP spans already

I don't think I'll be using `d:spans/exclusive_time@millisecond` so we can probably ignore that for now?

jjbayer: @edwardgou-sentry thanks, I reverted this change.

PR #3133, opened 2024-02-18
ref(tests): Rewrite dynamic sampling integration tests to rust

TBS1996:
rewrites the dynamic sampling tests, except for two of them because they use consumers and I haven't ported that to rust yet.  
 
Pretty big diff because i also had to make the basic testing framework in rust


iker-barriocanal:
Thanks for putting in the effort to make the PR. A couple of notes:
- Is there any chance to split the PR? The current scope is big enough to prevent us from making a decent review and providing meaningful feedback.
- Creating a new testing framework is an excellent opportunity to improve our existing Python integration tests, from addressing pain points to a simpler framework with a better Development Experience. Is there a document where I could read about these conclusions?

TBS1996:
>Is there any chance to split the PR? The current scope is big enough to prevent us from making a decent review and providing meaningful feedback.
 
 Hmm It's a bit hard to do that, the most basic tests all need a lot of boilerplate. the framework is bigger locally but I removed everything I didn't need for these tests.  
 
Although I could make the PR only focus on a single test and remove everything not needed and have a second PR doing the rest of the tests here.

>Creating a new testing framework is an excellent opportunity to improve our existing Python integration tests, from addressing pain points to a simpler framework with a better Development Experience. Is there a document where I could read about these conclusions?  
 
I don't think such a document exist no, I just tried to make the API as pleasant as possible while I was coding this, if it's of interest then as a team we can discuss what we want in a reworked testing framework and I'll do my best to implement it!

relay/tests/test_dynamic_sampling.rs
```diff
@@ -0,0 +1,326 @@
...
+        .add_basic_sampling_rule(RuleType::Transaction, 0.0)
+        .set_transaction_metrics_version(1);
+
+    let public_key = project_state.public_key();
+    let sentry = MiniSentry::new().add_project_state(project_state);
+    let relay = Relay::builder(&sentry).enable_outcomes().build();
+
+    // create an envelope with a trace context that is initiated by this project (for simplicity)
+    let envelope = Envelope::new()
+        .add_basic_transaction(None)
```
jjbayer, line 25: Might as well go full builder and do something like `.add_basic_transaction().with_transaction_name("foo")`

relay-test/src/mini_sentry.rs
```diff
@@ -0,0 +1,570 @@
...
+        }
+        self
+    }
+
+    pub fn wait_for_outcome(&self, timeout: u64) -> &Self {
+        for _ in 0..timeout {
+            if !self.is_empty() {
+                return self;
+            }
+            std::thread::sleep(Duration::from_secs(1));
```
jjbayer, line 87: I'm lacking full context, but is there a way we could spawn a bunch of async services in a single process instead of spawning a new process for relay, and then do everything async?

relay/tests/test_dynamic_sampling.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    // the event should be removed by Relay sampling
+    sentry.captured_envelopes().wait(2).assert_empty();
+
+    sentry
+        .captured_outcomes()
+        .wait_for_outcome(30)
+        .assert_outcome_qty(1)
+        .assert_all_outcome_id(Outcome::FILTERED)
+        .assert_all_outcome_reasons("Sampled:1");
```
jjbayer, line 39: Could this be more generic? The functions `assert*outcome*` seem too specific, would something like this make sense instead:

```suggestion
    let outcomes = 
        .captured_items::<Outcome>()
        .wait_for(30);
    let outcome = outcomes.next();
    assert_eq!(outcome.id, Outcome::Filtered);
    assert_eq!(outcome.reason, "Sampled:1");
    assert!(outcomes.next().is_none());
```

relay-test/src/mini_sentry.rs
jjbayer, line 87: Ideally, we don't wait a full second if the next item has already arrived. You could use condition variables for this, but there's probably a more idiomatic way to do it, e.g. channels or watchers. https://doc.rust-lang.org/std/sync/struct.Condvar.html

jjbayer (commented):
Thanks for picking this up! Just a bunch of random comments for now. I think the most important one is that we should never wait for something longer than necessary.

relay/tests/test_dynamic_sampling.rs
TBS1996: well that would be two different types i guess, so you mean a builder within the builder? 

`add_basic_transaction` sounds like it should be on the envelope type but `with_transaction_name` on the Item type

relay-test/src/mini_sentry.rs
TBS1996: interesting idea, i'll look into that!

relay/tests/test_dynamic_sampling.rs
TBS1996: well yes, could be more generic as the current python way of doing it, I liked this way of chaining asserts together though so it depends on the preference. I think we can discuss it as a team

relay-test/src/mini_sentry.rs
TBS1996: thanks! will look into this

olksdr:
Closing for now. 

We can revisit this later, when we have more time.

PR #3135, opened 2024-02-19
fix(tests): Fix flaky global ratelimit test

### LINKED ISSUE #3030: [Flaky CI]: test_global_rate_limit
### Flakiness Type

Assertion failure

### Name of Test

tests/integration/test_outcome.py::test_global_rate_limit

### Link to Test Run

https://github.com/getsentry/relay/actions/runs/7722991438/job/21052183827

### Details

_No response_

TBS1996:
it was flaky cause sometime we rolled over to another slot in the middle of the test, fixed by setting window to same as the current time so we start at beginning of a new very big slot

#skip-changelog


tests/integration/test_outcome.py
```diff
@@ -1648,19 +1648,21 @@ def test_global_rate_limit(
     projectconfig = mini_sentry.add_full_project_config(project_id)
     mini_sentry.add_dsn_key_to_project(project_id)
 
+    ts = datetime.utcnow().timestamp()
+    # Ensures we begin at the start of a slot so we don't go to next slot in the middle of the test
+    window = ts
```
iker-barriocanal: I'd suggest removing `ts` and using `window` directly on line 1671. Declaring a duplicated variable here that's not used until a later point makes the test harder to understand.

iker-barriocanal (approved):
Please see my comment below before merging.

tests/integration/test_outcome.py
TBS1996: hmm,  i mainly did it to make documentation easier, since it's a bit awkward to put a long doc sentence in the bucket json, but yeah I removed that variable now

Issue #3074, opened 2024-02-07
[EPIC] Metric Bucket JSON-Encoding for Sets and Distributions


Dav1dde:
Set and distribution metrics are currently being sent as float arrays in JSON which causes a significant overhead in various parts of the system. 

The float arrays are expensive to parse and consumers which are not interested in the individual values still need to parse each float (without a specialised JSON parser). Additionally in practice the float arrays are a very inefficient encoding and require much more than 8 bytes (f64) per value.

All array values are converted to little endian before compression and encoding.

To increase overall throughput we want to use a more efficient encoding of the float arrays for set and distribution metrics.

Before switching from JSON to a binary format like CBOR we want to explore different ways to encode and compress the data with JSON first.

Benchmarks regarding throughput and compression have been collected in [getsentry/bucket-compression](https://github.com/getsentry/bucket-compression).

We identified `zstd` to be a very good general purpose algorithm which should give us good wins across the board. To encode binary data (`zstd` compressed data) we will use Base64 (**without** padding).

*Note:* zstd is used in its streaming mode.

Proposed schema change for distribution and set values (`values` field):

```json
{
  "oneOf": [
    { "type": "array", "items": { "type": "number" } },
    {
      "type": "object",
      "properties": {
        "format": { "const": "array" },
        "data": { "type": "array", "items": { "type": "number" } }
      }
    },
    {
      "type": "object",
      "properties": {
        "format": { "const": "zstd" },
        "data": { "type": "string" }
      }
    }
  ]
}
```

Examples:

```json
  {
    "org_id": 1,
    "project_id": "12345...",
    "timestamp": 1615889440,
    "width": 10,
    "name": "endpoint.response_time",
    "tags": {
      "route": "user_index"
    },
    "type": "d",
    "value": {"format": "zstd", "data": "<base64>"}
  }
```

```json
  {
    "name": "endpoint.response_time",
    ...
    "type": "s",
    "value": {"format": "array", "data": [13.37, 42, 3.14159265358979323846264338327950288]},
  }
```

**Note:** We want to stay compatible with the current JSON float array values, since we components which directly write to Kafka and are not using Relay as their ingestion path.

## Milestone 1 - New JSON Schema

As a first milestone we want to support the new JSON schema with the `array` format. This does not change the encoding of the values yet.

We will have to add support for the new format in multiple systems:

```[tasklist]
### Relay
- [ ] https://github.com/getsentry/relay/pull/3137
[...]

PR #3137, opened 2024-02-19
feat(bucket-encoding): Implements dynamic bucket encoding

Dav1dde:
Adds support for the `array` metric bucket encoding, configurable through global configs.

Epic: #3074

### DIFF
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -162,11 +162,11 @@ impl MetricBucketEncodings {
 #[derive(Debug, Clone, Copy, Default, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "lowercase")]
 pub enum MetricEncoding {
-    /// The default compatibility encoding.
+    /// The default legacy encoding.
     ///
     /// A simple JSON array of numbers.
     #[default]
-    Compat,
+    Legacy,
     /// The array encoding.
     ///
     /// Uses already the dynamic value format but still encodes
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -3,6 +3,7 @@ use std::fs::File;
 use std::io::BufReader;
 use std::path::Path;
 
+use relay_base_schema::metrics::MetricNamespace;
 use relay_event_normalization::MeasurementsConfig;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
@@ -102,6 +103,14 @@ pub struct Options {
     )]
     pub span_usage_metric: bool,
 
+    /// Metric bucket encoding configuration by metric namespace.
+    #[serde(
+        rename = "relay.metric-bucket-encodings",
+        deserialize_with = "default_on_error",
+        skip_serializing_if = "is_default"
+    )]
+    pub metric_bucket_encodings: MetricBucketEncodings,
+
     /// All other unknown options.
     #[serde(flatten)]
     other: HashMap<String, Value>,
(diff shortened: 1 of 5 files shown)


relay-dynamic-config/src/global.rs
```diff
@@ -124,6 +133,47 @@ pub enum CardinalityLimiterMode {
...
+
+/// All supported metric bucket encodings.
+#[derive(Debug, Clone, Copy, Default, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum MetricEncoding {
+    /// The default compatibility encoding.
+    ///
+    /// A simple JSON array of numbers.
+    #[default]
+    Compat,
```
jjbayer: nit: Should we call this `Legacy`?

relay-server/src/services/store.rs
```diff
@@ -339,22 +345,27 @@ impl StoreService {
...
-                    project_id: scoping.project_id,
-                    name: view.name(),
-                    value: view.value(),
-                    timestamp: view.timestamp(),
-                    tags: view.tags(),
-                    retention_days: retention,
-                };
+                let namespace = MetricResourceIdentifier::parse(view.name())
+                    .map(|mri| mri.namespace)
+                    .unwrap_or(MetricNamespace::Unsupported);
```
jjbayer, line 357: Do we actually wan to create a metric message if we cannot parse the MRI? Or should we skip it instead?

relay-server/src/services/store.rs
```diff
@@ -695,20 +734,19 @@ impl StoreService {
...
-                    name: bucket.name.as_str(),
-                    value: (&bucket.value).into(),
-                    timestamp: bucket.timestamp,
-                    tags: &bucket.tags,
-                    retention_days: retention,
-                },
-            )?;
+            let view = BucketView::new(&bucket);
+
+            let namespace = MetricResourceIdentifier::parse(view.name())
```
jjbayer, line 742: Should we pass the parsed MRI from `handle_store_message` instead of parsing again?

relay-server/src/services/store.rs
```diff
@@ -1275,12 +1313,58 @@ struct MetricKafkaMessage<'a> {
...
 
+#[derive(Clone, Debug, Serialize)]
+#[serde(tag = "type", content = "value")]
+enum MetricValue<'a> {
+    #[serde(rename = "c")]
+    Counter(FiniteF64),
+    #[serde(rename = "d")]
+    Distribution(FloatArrayEncoding<&'a [FiniteF64]>),
+    #[serde(rename = "s")]
+    Set(FloatArrayEncoding<SetView<'a>>),
```
jjbayer: nit(naming): `Set` does not contain floats, so should we rename to `NumericArrayEncoding`?

relay-server/src/services/store.rs
Dav1dde: This will already skip unsupported namespaces in the store function. I opted not to make the decision here to skip so it is only in a single place (would be 3 otherwise).

relay-server/src/services/store.rs
Dav1dde: This is a separate code path to `handle_store_message` / `handle_store_metrics`, it's the old one where metrics are parsed from an envelope again.

relay-server/src/services/store.rs
Dav1dde: Renamed to just `ArrayEncoding`, since that is actually what it is, the code does not care about the payload 👍 .

jjbayer (approved):

olksdr (approved):

PR #3140, opened 2024-02-19
feat(processor): Add explitic types transition

olksdr:
We should be more explicit converting between types, which should be easier to spot in the PRs. 

These changes:
* introduce `TryFrom` implementation for processing group types and typed envelope
* type transition is more explicit and can fail if someone tries to convert something which is not supported
* use typed envelope in more places and remove managed envelope from imports

related to https://github.com/getsentry/team-ingest/issues/253
#skip-changelog


TBS1996:
will we remove the `group` field on envelope context later? I mean, if a function takes an ungrouped envelope it could just accept ManagedEnvelope, if it takes in grouped one it can take in either specific or generic

relay-server/src/endpoints/common.rs
```diff
@@ -303,17 +303,9 @@ fn queue_envelope(
...
-        managed_envelope.accept();
-    } else {
-        relay_log::trace!("queueing envelope");
-        state
-            .project_cache()
-            .send(ValidateEnvelope::new(managed_envelope));
-    }
+    // The entire envelope is taken for a split above, and it's empty at this point, we can just
+    // accept it without additional checks.
+    managed_envelope.accept();
```
jjbayer, line 308: Should we make `split_envelope` consume the managed envelope, so we don't have to handle its leftovers here?

relay-server/src/utils/managed_envelope.rs
```diff
@@ -68,26 +68,56 @@ struct EnvelopeContext {
...
+}
+
+impl std::error::Error for InvalidProcessingGroupType {}
+
+/// A wrapper for [`ManagedEnvelope`] with assigned processing group type.
 pub struct TypedEnvelope<G>(ManagedEnvelope, PhantomData<G>);
 
 impl<G> TypedEnvelope<G> {
-    pub fn new(managed_envelope: ManagedEnvelope) -> Self {
+    /// Create new typed envelope.
```
```suggestion
    /// Creates new typed envelope.
```

relay-server/src/utils/managed_envelope.rs
```diff
@@ -68,26 +68,56 @@ struct EnvelopeContext {
...
+    ///
+    /// Note: this method is private to make sure that only `TryFrom` implementation is used, which
+    /// requires the check for the error if conversion is failing.
+    fn new(managed_envelope: ManagedEnvelope, _ty: G) -> Self {
```
jjbayer: Should we move this below the public methods?

jjbayer (approved):

relay-server/src/endpoints/common.rs
olksdr: I might take this as followup, since right now this function also used in the place when we restore the envelopes from the spool, and there we have only boxed envelope. 
I could expose 2 function, once (which excepts the managed envelope) wraps another one (the current which jsut needs envelop) , it's that cleaner

Dav1dde (approved):

PR #3141, opened 2024-02-19
feat(server): parametrize dsc transaction

### LINKED ISSUE #2068: Match trace rules against the parametrized name of a transaction
At the moment it looks like dynamic sampling rules that apply on traces do not operate on the parametrised transaction names, i.e.  the transaction name is taken from the trace header and it is compared as is against the name specified in the rule.

This is a problem because we would like to create trace rules for rebalancing the transactions (root transactions) in a project and we are restricted to a small number of rules ( 30). Having hundreds of transaction names makes it hard to properly balance the transactions in a project.

What we need:
1. The trace transaction rules should be able to operate on parametrized transaction names ( not the original transaction name from the trace header).
2. The transaction name specified in the root transaction count metric should be the parametrized name ( since these are the names used to generate the rules in the first place), at the moment that doesn't seem to be the case (didn't check in the code just looked at the transaction names from the collected metrics).

TBS1996:
#2068

Parametrizes the root transaction of a DSC so that trace rules can match on the parametrized version of transaction names


relay-event-normalization/src/transactions/processor.rs
```diff
@@ -89,14 +107,7 @@ impl<'r> TransactionsProcessor<'r> {
...
         if self.treat_transaction_as_url(event) {
-            // Normalize transaction names for URLs and Sanitized transaction sources.
-            // This in addition to renaming rules can catch some high cardinality parts.
-            scrub_identifiers(&mut event.transaction);
-
-            // Apply rules discovered by the transaction clusterer in sentry.
-            if !self.name_config.rules.is_empty() {
-                self.apply_transaction_rename_rule(&mut event.transaction);
-            }
+            normalize_transaction_name(&mut event.transaction, self.name_config.rules);
```
TBS1996, line 110: extracted for re-use with dsc

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -19,6 +19,59 @@ pub struct TransactionNameConfig<'r> {
...
+/// Applies the rule if any found to the transaction name.
+///
+/// It find the first rule matching the criteria:
+/// - source matchining the one provided in the rule sorce
+/// - rule hasn't epired yet
+/// - glob pattern matches the transaction name
+///
+/// Note: we add `/` at the end of the transaction name if there isn't one, to make sure that
+/// patterns like `/<something>/*/**` where we have `**` at the end are a match.
+pub fn apply_transaction_rename_rules(
```
TBS1996, line 46: unchanged, just moved

relay-event-normalization/src/transactions/processor.rs
```diff
@@ -89,14 +107,7 @@ impl<'r> TransactionsProcessor<'r> {
 
     fn normalize_transaction_name(&self, event: &mut Event) {
         if self.treat_transaction_as_url(event) {
```
TBS1996, line 109: not sure how to deal with the `treat_transaction_as_url` logic here. It uses data in `transaction_info` field of `Event` which the DSC does not have access to. I only added the 'contains / ' condition from it

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - Forward metrics in proxy mode. ([#3106](https://github.com/getsentry/relay/pull/3106))
 - Do not PII-scrub code locations by default. ([#3116](https://github.com/getsentry/relay/pull/3116))
 
+**Features**;
+
+- Parametrize dsc root transactions. ([#3141](https://github.com/getsentry/relay/pull/3141))
```
```suggestion
- Parametrize transaction in dynamic sampling context. ([#3141](https://github.com/getsentry/relay/pull/3141))
```

relay-server/src/envelope.rs
```diff
@@ -1144,6 +1145,30 @@ impl Envelope {
...
+        if !transaction.contains('/') {
+            return;
+        }
+
+        let mut annotated = Annotated::new(transaction.clone());
+        normalize_transaction_name(&mut annotated, rules);
+
+        if let Some(parametrized) = annotated.into_value() {
+            *transaction = parametrized;
+        }
```
jjbayer: If the normalization decided to erase the transaction name, we should respect that decision, right?

```suggestion
    dsc.transaction = annotated.into_value();
```

relay-server/src/services/processor.rs
```diff
@@ -1129,6 +1128,12 @@ impl EnvelopeProcessorService {
         &self,
         state: &mut ProcessEnvelopeState<G>,
     ) -> Result<(), ProcessingError> {
+        let project_state = Arc::clone(&state.project_state);
```
jjbayer: Do we need the clone here?

relay-server/src/services/processor/dynamic_sampling.rs
```diff
@@ -743,4 +745,91 @@ mod tests {
 
         assert_eq!(get_sampling_match(res).sample_rate(), 0.4);
     }
+
+    // Checks that the trace rules matches on the parametrized name of the root ds ctransaction.
+    #[test]
+    fn test_trace_matches_parametrized_dsc_root_transaction() {
```
jjbayer: This test runs dynamic sampling, but what we actually need to test is

1. whether `parametrize_dsc_transaction` works as expected (could be a unit test)
2. whether the envelope processor, with a given project state and root project state, samples the transaction correctly.
3.  We should also check in a test case whether part 2 of the ticket is now given:
> The transaction name specified in the root transaction count metric should be the parametrized name ( since these are the names used to generate the rules in the first place), at the moment that doesn't seem to be the case (didn't check in the code just looked at the transaction names from the collected metrics).

relay-event-normalization/src/transactions/processor.rs
jjbayer: Yeah I think this is the best we can do for now. The chance that a rule derived from URL transactions accidentally also matches & messes up a non-URL transaction is low, so I think we can risk it.

relay-server/src/services/processor.rs
```diff
@@ -1129,6 +1128,12 @@ impl EnvelopeProcessorService {
...
+
+        state
+            .envelope_mut()
+            .parametrize_dsc_transaction(&project_state.config.tx_name_rules);
```
jjbayer: We need to pass the root project state here, not the event's project state.

jjbayer (changes requested):

relay-server/src/services/processor.rs
TBS1996: In the current setup, yes, because we need mutable reference to `ProcessEnvelopeState` to mutate the envelope, and the rules are inside the project state, so we'd get a `cannot borrow mutable because its also borrowed as immutable` error

relay-server/src/envelope.rs
TBS1996: hm yeah, that makes sense

relay-server/src/services/processor/dynamic_sampling.rs
TBS1996: 1. this test does check whether `parametrize_dsc_transaction` works though, the asserts are pretty explicit about it, although I I could also make one that only focus on that function as well.
2. so well I could make a test like that too, itd be essentially the same test just more of an integration test
3. again, doesn't this test check that? i think maybe there's a misunderstanding here, feel free to dm me about it

relay-server/src/services/processor.rs
TBS1996: oh yes, great catch!

relay-server/src/envelope.rs
```diff
@@ -1144,6 +1145,26 @@ impl Envelope {
...
+    /// Runs transaction parametrization on the DSC trace transaction.
+    ///
+    /// The purpose is for trace rules to match on the parametrized version of the transaction.
+    pub fn parametrize_dsc_transaction(&mut self, rules: &[TransactionNameRule]) {
+        let Some(ErrorBoundary::Ok(dsc)) = &mut self.headers.trace else {
+            return;
+        };
+
+        let parametrized_transaction = match &dsc.transaction {
+            Some(transaction) if transaction.contains('/') => {
```
jjbayer, line 1157:
```suggestion
            // Ideally we would only apply transaction rules to transactions with source `url`,
            // but the DSC does not contain this information. The chance of a transaction rename rule accidentially matching a non-URL transaction should be very low.
            Some(transaction) if transaction.contains('/') => {
```

relay-server/src/services/processor/dynamic_sampling.rs
jjbayer: Now that we have a unit test and an integration test, can we remove this one? Or does it still test something that the other two tests do not cover?

tests/integration/test_dynamic_sampling.py
```diff
@@ -348,6 +348,69 @@ def test_it_tags_error(mini_sentry, relay, expected_sampled, sample_rate):
...
+        "id": 1,
+    }
+
+    rules = ds.setdefault("rules", [sampling_rule])
+    rules.append(sampling_rule)
+
+    envelope = Envelope()
+    transaction_event, trace_id, _ = _create_transaction_item()
+    envelope.add_transaction(transaction_event)
+    _add_trace_info(envelope, trace_id, public_key, transaction=original_transaction)
```
jjbayer: We should add a second project config that contains the txNameRules and pass its project key here, to test that the normalization uses the correct config.

tests/integration/test_dynamic_sampling.py
```diff
@@ -348,6 +348,70 @@ def test_it_tags_error(mini_sentry, relay, expected_sampled, sample_rate):
...
+    parametrized_transaction = "/auth/login/*/"
+
+    config = mini_sentry.add_basic_project_config(project_id)
+    sampling_config = mini_sentry.add_basic_project_config(43)
+    public_key2 = sampling_config["publicKeys"][0]["publicKey"]
+
+    sampling_config["config"]["txNameRules"] = [
+        {
+            "pattern": pattern,
+            "expiry": "3022-11-30T00:00:00.000000Z",
```
TBS1996, line 374: note: test will fail 998 years from now

jjbayer (approved):

PR #3145, opened 2024-02-20
test(metrics): Increase timeout for transactions test

### LINKED ISSUE #2898: [Flaky CI]: test_transaction_metrics
### Flakiness Type

Assertion failure

### Name of Test

tests/integration/test_metrics.py::test_transaction_metrics[extract from transaction-sampled, external relay]

### Link to Test Run

https://github.com/getsentry/relay/actions/runs/7298623864/job/19889953446?pr=2895#step:9:1244

### Details

Different than https://github.com/getsentry/relay/issues/2865

Dav1dde:
The Relay logs show no differences to a successful run, in the end of the logs you can see that Relay emits 7 different metrics into the topic but yet the assertion fails with that there is a message missing. Seems like a spurious failure (Kafka being slow?), increased the timeout as a best effort fix (does not slow down tests, it's just for the error case).

Closes: #2898

#skip-changelog

### DIFF
--- a/tests/integration/test_metrics.py
+++ b/tests/integration/test_metrics.py
@@ -972,7 +972,7 @@ def test_transaction_metrics(
 
         return
 
-    metrics = metrics_by_name(metrics_consumer, 7)
+    metrics = metrics_by_name(metrics_consumer, 7, timeout=6)
     common = {
         "timestamp": int(timestamp.timestamp()),
         "org_id": 1,


iker-barriocanal (approved):

Dav1dde:
Okay so, reducing the timeout shows the same error we see in the failed test, all signs are pointing towards just a failed timeout.

Bumped from 3 to 6 seconds now.

olksdr (approved):

PR #3146, opened 2024-02-20
fix(config): Processing is only for managed mode

### LINKED ISSUE #3048: Forbid combination of proxy mode and processing in config
Apparently it is possible to run Relay with a combination of `mode: proxy` and `processing: enabled`. See https://github.com/getsentry/relay/issues/3033#issuecomment-1921441767.

We should catch this in a config validation step.

olksdr:
Proxy mode should not allow processing to be enabled. 

fix: https://github.com/getsentry/relay/issues/3048

#skip-changelog

### DIFF
--- a/relay/src/setup.rs
+++ b/relay/src/setup.rs
@@ -10,10 +10,7 @@ pub fn check_config(config: &Config) -> Result<()> {
     }
 
     if config.relay_mode() == RelayMode::Proxy && config.processing_enabled() {
-        anyhow::bail!(
-            "Processing mode cannot be enabled while in the proxy mode. \
-            Please disable processing or switch to managed mode."
-        );
+        anyhow::bail!("Processing cannot be enabled while in proxy mode.");
     }
 
     Ok(())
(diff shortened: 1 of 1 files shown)


Dav1dde (approved):

relay/src/setup.rs
```diff
@@ -9,6 +9,13 @@ pub fn check_config(config: &Config) -> Result<()> {
         );
     }
 
+    if config.relay_mode() == RelayMode::Proxy && config.processing_enabled() {
+        anyhow::bail!(
+            "Processing mode cannot be enabled while in the proxy mode. \
```
jjbayer:
```suggestion
            "Processing mode cannot be enabled while in proxy mode. \
```

relay/src/setup.rs
```diff
@@ -9,6 +9,13 @@ pub fn check_config(config: &Config) -> Result<()> {
...
+            Please disable processing or switch to managed mode."
```
jjbayer: This might incentivize users to enable processing in managed mode, so I would not even mention it here. Side note: Should `Processing` be a separate mode, i.e. variant of `RelayMode`, instead of a boolean flag?

```suggestion
            "
```

TBS1996 (approved):

PR #3148, opened 2024-02-21
instr(spans): Tag kafka metric with is_segment / has_parent

jjbayer:
For spans, tag the statsd metric that we emit for produced Kafka messages with `is_segment` and `has_parent`. 100% of transaction spans are currently converted to standalone spans, so this should give us an accurate estimate of the ratio of spans vs. segments (=transactions) vs. traces.

#skip-changelog

### DIFF
--- a/relay-spans/src/span.rs
+++ b/relay-spans/src/span.rs
@@ -84,6 +84,7 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
     let trace_id = hex::encode(trace_id);
     let parent_span_id = hex::encode(parent_span_id);
 
+    // TODO: This is wrong, a segment could still have a parent in the trace.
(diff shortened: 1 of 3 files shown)


relay-spans/src/span.rs
```diff
@@ -84,6 +84,7 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
     let trace_id = hex::encode(trace_id);
     let parent_span_id = hex::encode(parent_span_id);
 
+    // TODO: This is wrong, a segment could still have a parent in the trace.
```
jjbayer, line 87: Will fix this in a different PR. It will not skew the measurements by much because we barely ingest any standalone spans from SDKs yet.

mitsuhiko (approved):

olksdr (approved):

Dav1dde (approved):

PR #3149, opened 2024-02-21
ref(replays): video item

jjbayer:
Serialize processed replay video item back into the envelope instead of putting it on the header.

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor/replay.rs
+++ b/relay-server/src/services/processor/replay.rs
@@ -16,7 +16,7 @@ use relay_statsd::metric;
 use rmp_serde;
 use serde::{Deserialize, Serialize};
 
-use crate::envelope::{ContentType, Item, ItemType};
+use crate::envelope::{ContentType, ItemType};
 use crate::services::outcome::{DiscardReason, Outcome};
 use crate::services::processor::{ProcessEnvelopeState, ProcessingError, ReplayGroup};
 use crate::statsd::RelayTimers;
@@ -68,14 +68,13 @@ pub fn process(
         match item.ty() {
             ItemType::ReplayEvent => {
                 match handle_replay_event_item(
-                    &item.payload(),
+                    item.payload(),
                     project_config,
                     client_addr,
                     user_agent,
                 ) {
-                    ProcessingAction::Drop(action) => action,
-                    ProcessingAction::Keep => ItemAction::Keep,
-                    ProcessingAction::Replace(replay_event) => {
+                    Err(outcome) => ItemAction::Drop(outcome),
+                    Ok(replay_event) => {
                         item.set_payload(ContentType::Json, replay_event);
                         ItemAction::Keep
                     }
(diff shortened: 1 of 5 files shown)


relay-server/src/services/store.rs
```diff
@@ -878,6 +868,50 @@ impl StoreService {
...
+        payload: Bytes,
+        start_time: Instant,
+        retention: u16,
+    ) -> Result<(), StoreError> {
+        #[derive(Deserialize)]
+        struct VideoEvent<'a> {
+            replay_video: &'a [u8],
+            replay_event: &'a [u8],
+            replay_recording: &'a [u8],
+        }
```
jjbayer, line 884: Deserializing this is cheap, because it does not copy any data.

relay-server/src/services/processor/replay.rs
```diff
@@ -119,36 +121,30 @@ pub fn process(
...
 
 fn handle_replay_event_item(
-    payload: &[u8],
+    payload: Bytes,
     config: &ProjectConfig,
     client_ip: Option<IpAddr>,
     user_agent: &RawUserAgentInfo<&str>,
-) -> ProcessingAction<String> {
-    match process_replay_event(payload, config, client_ip, user_agent) {
+) -> Result<Bytes, Outcome> {
```
jjbayer, line 126: @cmanallen I refactored `ProcessingAction` to `Result<Bytes, Outcome>`, let me know if you like it. By using `Bytes`, which is a lightweight, reference-counted object, we can fold the `Keep` and `Replace` cases into one.

relay-server/src/services/store.rs
```diff
@@ -878,6 +868,50 @@ impl StoreService {
...
+        payload: Bytes,
+        start_time: Instant,
+        retention: u16,
+    ) -> Result<(), StoreError> {
+        #[derive(Deserialize)]
+        struct VideoEvent<'a> {
+            replay_event: &'a [u8],
+            replay_recording: &'a [u8],
+            replay_video: &'a [u8],
+        }
```
jjbayer, line 884: Deserializing this is cheap, because it does not copy any data.

relay-server/src/services/processor/replay.rs
```diff
@@ -226,63 +215,50 @@ fn handle_replay_recording_item(
...
-    if event.replay_video.is_empty() {
-        return ProcessingAction::Drop(ItemAction::Drop(Outcome::Invalid(
-            DiscardReason::InvalidReplayVideoEvent,
-        )));
+    if replay_video.is_empty() {
+        return Err(Outcome::Invalid(DiscardReason::InvalidReplayVideoEvent));
     }
 
-    ProcessingAction::Replace((replay_event, replay_recording, event.replay_video))
+    match rmp_serde::to_vec_named(&ReplayVideoEvent {
```
jjbayer, line 256: It's unfortunate that we need to serialize here, but it's a known problem that we plan to fix for other item types as well.

If we do not serialize back into the envelope, the data will be lost in PoP Relays.

olksdr (approved):

Dav1dde (approved):
Nice, as bad as the useless (but currently necessary) roundtrip through de- & serialization is with the zero copy deserialization is, it should be much less bad.

PR #3151, opened 2024-02-22
feat(processor): Introduce first step state wrapper

olksdr:
Introduced changes: 

* instead of passing mutable reference into all processing function - we mostly move the state and let the function own it
  * in case of error we return the state back together with error
  * in case of success the state is returned
  * most of the function accept only concrete state wrapper type and then return also either the same or different type
* introduced different states, which accepted and returned by different function, and initially predefines the order in which the functions can run
  * there are still few caveats which should be worked out, because of special behaviour in `processing` mode it's not that easy to return the proper type
  * in some cases the simple mutable reference to `ProcessEnvelopeState` still is passed around

There is no functional changes in this PR and mostly changing the signatures and return types of the functions.

related: https://github.com/getsentry/team-ingest/issues/253

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -90,6 +90,7 @@ mod replay;
 mod report;
 mod session;
 mod span;
+mod state;
 #[cfg(feature = "processing")]
 mod unreal;
 
@@ -469,6 +470,8 @@ impl From<ExtractMetricsError> for ProcessingError {
     }
 }
 
+pub type ProcessError<'a, G> = (ProcessEnvelopeState<'a, G>, ProcessingError);
+
 type ExtractedEvent = (Annotated<Event>, usize);
 
 impl ExtractedMetrics {
@@ -497,7 +500,7 @@ impl ExtractedMetrics {
 
 /// A state container for envelope processing.
 #[derive(Debug)]
-struct ProcessEnvelopeState<'a, Group> {
+pub struct ProcessEnvelopeState<'a, Group> {
     /// The extracted event payload.
     ///
     /// For Envelopes without event payloads, this contains `Annotated::empty`. If a single item has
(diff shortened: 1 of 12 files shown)


relay-server/src/services/processor.rs
```diff
@@ -1214,184 +1234,191 @@ impl EnvelopeProcessorService {
...
             &self.inner.global_config.current(),
         );
 
         if_processing!(self.inner.config, {
-            event::store(state, &self.inner.config)?;
-            self.enforce_quotas(state)?;
-            profile::process(state, &self.inner.config);
+            state = event::store(state, &self.inner.config)?;
+            let enforced_state = self.enforce_quotas(state)?;
+            state = profile::process(enforced_state, &self.inner.config);
```
jjbayer: It seems wrong to go back from `EnforcedQuotasState` to ProcessEnvelopeState` here. Since we care about maintaining order, maybe `enforce_quotas` is a bad place to start, because it is only called in processing relays. Something like `event::filter` with a resulting `FilteredState` would be easier to enforce later on.

jjbayer (commented):
Though I get the idea, I'm not sure if changing all signatures from `(&mut State) -> ()` to `(State) -> (State)` makes sense without actually changing the output type. We could instead only change the signature of the one function that actually transforms the state into a new type.

olksdr:
> Though I get the idea, I'm not sure if changing all signatures from `(&mut State) -> ()` to `(State) -> (State)` makes sense without actually changing the output type. We could instead only change the signature of the one function that actually transforms the state into a new type.

The idea was to transition to new state type in each function and make next function to consume only the concrete type it needs, that's why I made all the function to actually consume the provided state and return it back once the function done with it. 

This is basically preparation before introducing different types `EnforcedQuotasState` type here is just to illustrate where this is heading. 

I might overshoot here and we want to have much smaller change to begin with?

relay-server/src/services/processor.rs
olksdr: Yeah, the enforced quotas type might be a bad example to start with

relay-server/src/services/processor/state.rs
```diff
@@ -0,0 +1,94 @@
+//! This module contains types and utils used in processor to decribe and support the processing pipeline.
+//!
+
+use crate::services::processor::ProcessEnvelopeState;
+
+macro_rules! state_type {
+    ($name:ident) => {
+        pub struct $name<'a, G>(ProcessEnvelopeState<'a, G>);
```
Dav1dde, line 14: To allow running the same function from different groups, the inner would need to be generic and to enforce certain properties we'd have to use traits, is that still something you want to get to or do you have a new approach in mind?

olksdr:
Closing this out. After discussion we agreed this change too convoluted and does not work like an intermediate step. 
We will follow-up on this with the related initiatives and refactorings.

PR #3152, opened 2024-02-22
ref(health): Report failing health checks

Dav1dde:
Investigating why a health check failed is requiring more effort than anticipated, let's make it clear by logging and reporting an error.

#skip-changelog

### DIFF
--- a/relay-server/src/services/health_check.rs
+++ b/relay-server/src/services/health_check.rs
@@ -43,6 +43,17 @@ pub enum Status {
     Unhealthy,
 }
 
+impl Status {
+    /// Internal helper to report/log failing health checks.
+    fn report(self, probe: &'static str) -> Self {
+        if matches!(self, Self::Unhealthy) {
+            relay_log::error!("Health check probe '{probe}' failed");
+        }
+
+        self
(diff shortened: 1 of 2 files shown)


relay-server/src/services/health_check.rs
```diff
@@ -43,6 +43,17 @@ pub enum Status {
     Unhealthy,
 }
 
+impl Status {
+    /// Internal helper to report/log failing health checks.
+    fn report(self, probe: &'static str) -> Self {
+        if matches!(self, Self::Unhealthy) {
+            relay_log::error!("Health check probe '{probe}' failed");
```
iker-barriocanal, line 50: nit: should we make this info/warning? Not sure whether the macro reports these to Sentry; if that's the case ship the `error`.

iker-barriocanal (approved):

relay-server/src/services/health_check.rs
Dav1dde: I think `warn` would be the appropriate level, but they are not reported to Sentry iirc.

olksdr (approved):

PR #3153, opened 2024-02-22
feat(spans): Parse supabase span descriptions

jjbayer:
The description only contains the table name, e.g. `"from(users)"`.
In the future, we might want to parse `data.query` as well.

See https://github.com/supabase-community/sentry-integration-js/blob/bae59de7e86c916d9396e2e2de55e6ac17e646d6/index.js#L259

### DIFF
--- a/relay-event-normalization/src/normalize/span/tag_extraction.rs
+++ b/relay-event-normalization/src/normalize/span/tag_extraction.rs
@@ -365,6 +365,12 @@ pub fn extract_tags(
             } else {
                 None
             }
+        } else if span.origin.as_str() == Some("auto.db.supabase") {
+            scrubbed_description.as_deref().map(|s| {
+                s.trim_start_matches("from(")
+                    .trim_end_matches(')')
+                    .to_owned()
+            })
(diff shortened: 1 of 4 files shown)


olksdr (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -365,6 +365,12 @@ pub fn extract_tags(
             } else {
                 None
             }
+        } else if span.origin.as_str() == Some("auto.db.supabase") {
+            scrubbed_description.as_deref().map(|s| {
+                s.trim_start_matches("from(")
```
Dav1dde, line 370: Should we enforce that this description always starts with `from(`?

Dav1dde (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: Thanks, this actually surfaced a bug, see follow-up here: https://github.com/getsentry/relay/pull/3156

PR #3155, opened 2024-02-23
instr(server): Add namespace tag to kafka metrics-metric

TBS1996:
We now have the ability to limit metrics by namespace. To choose the right limit it would be helpful to see how often the different namespace are used. 

#skip-changelog

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -721,7 +721,8 @@ impl StoreService {
         )?;
         metric!(
             counter(RelayCounters::ProcessingMessageProduced) += 1,
-            event_type = "metric"
+            event_type = "metric",
+            namespace = namespace.as_str()
         );
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


relay-server/src/services/store.rs
```diff
@@ -721,7 +721,8 @@ impl StoreService {
         )?;
         metric!(
             counter(RelayCounters::ProcessingMessageProduced) += 1,
-            event_type = "metric"
+            event_type = "metric",
+            namespace = namespace.as_str()
```
jjbayer, line 725: Please document this tag here: https://github.com/getsentry/relay/blob/fbedc4878702ea405c6f741a07917a65da58a129/relay-server/src/statsd.rs#L540-L543

jjbayer (approved):

Dav1dde (approved):

relay-server/src/services/store.rs
TBS1996: thanks! i did

PR #3157, opened 2024-02-23
feat(cogs): Emit COGS

Dav1dde:
Emits COGS measurements from processing Relays into Kafka.

Currently only enables and implements COGS measurements for processing relays (with access to Kafka). There is currently no mechanism in Sentry to emit COGS via anything other than Kafka and PoP Relays do not have access to Kafka.

The `rust-usage-accountant` was prepared to support alternative transports than just Kafka also Relay (the COGS service) is prepared to emit measurements via a different mechanism 

COGS measurements are only relevant for Relay itself, `librelay` does not need them at this time, the callers themself should be COGS instrumented.

* Adds the `relay-cogs` module which contains the API we should use throughout Relay when we want to emit or deal with COGS data.
* Adds a COGS service which deals with actually receiving COGS measurements, aggregating, serializing and dispatching them.
* Adds config for the COGS Kafka Topic.

Depends on:
- https://github.com/getsentry/rust-usage-accountant/pull/8 
- https://github.com/getsentry/rust-usage-accountant/pull/9


Dav1dde:
Still missing tests, will add them next but I am already pretty happy with how it turned out.

relay-cogs/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "relay-cogs"
+authors = ["Sentry <oss@sentry.io>"]
+description = "Relay cost of goods module"
```
iker-barriocanal:
```suggestion
description = "Relay cost of goods sold module"
```

relay-config/src/config.rs
```diff
@@ -1336,6 +1336,41 @@ impl Default for Health {
...
+///
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(default)]
+pub struct Cogs {
+    /// Whether COGS measurements are enabled.
+    enabled: bool,
+    /// Granularity of the COGS measurements.
+    ///
+    /// Measurements are aggregated based on the granularity.
+    granularity: Option<Duration>,
```
iker-barriocanal: The config usually types time durations as `u32` and provides an accessor to build a `Duration` from it. We should keep it consistent.

relay-config/src/config.rs
```diff
@@ -1336,6 +1336,41 @@ impl Default for Health {
     }
 }
 
+///
```
iker-barriocanal: Something like:
```suggestion
/// COGS config.
```

iker-barriocanal (commented):
Very small review until the PR is ready.

Dav1dde:
Tests are in 🚀

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+    ///
+    ///     // App feature is only known after some computations.
+    ///     if do_something(item) {
+    ///         token.update(AppFeature::Spans);
+    ///     } else {
+    ///         token.update(AppFeature::Transactions);
+    ///     }
+    /// }
+    /// ```
+    pub fn update<T: Into<AppFeatures>>(&mut self, features: T) {
```
jjbayer: Instead of having an interface that overwrites all previous app features, could we have an interface that allows multiple updates? I.e.

```rust
let mut accountant = CogsToken::new();
// ...
for bucket in buckets {
    let feature = AppFeature::from(bucket.namespace);
    accountant.attribute(feature, bucket.size());
}
// in drop handler, accountant divides up elapsed time proportional to attributed numbers.
```

jjbayer: Hmm maybe I should keep on reading, seems like the builder does what I had in mind.

Dav1dde: The update was a bit of an afterthought, I was hoping I can set the weights once, when starting the measurement. Turns out we have some edge cases where that is not possible and we first have to do work before we can attribute the measurement.

relay-cogs/src/lib.rs
olksdr, line 1: Could you, please, add create level docs here? 

with following like all other crate have:
```
#![warn(missing_docs)]
#![doc(
)]
```

relay-cogs/src/cogs.rs
olksdr, line 1: It would be great also to have module level docs with short description what it does and what it is for.

olksdr (commented):
You might also want to mention this new crate in [the main](https://github.com/getsentry/relay/blob/master/relay/src/main.rs)

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+    ///
+    /// fn process_span(cogs: &Cogs, span: &mut Span) {
+    ///     let _token = cogs.timed(ResourceId::Relay, AppFeature::Spans);
+    ///
+    ///     scrub_sql(span);
+    ///     extract_tags(span);
+    /// }
+    ///
+    /// ```
+    pub fn timed<F: Into<AppFeatures>>(&self, resource: ResourceId, features: F) -> CogsToken {
```
iker-barriocanal: Any reason to go with the token approach over a closure? I find the increased readability worth the decreased flexibility. Something like the following:
```rs
Cogs::timed(some_features, || {
  // ...
});
```

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+    pub fn noop() -> Self {
+        Self::new(crate::NoopRecorder)
+    }
+}
+
+impl Cogs {
+    /// Starts a recording for a COGS measurement.
+    ///
+    /// When the returned token is dropped the measurement will be recorded
+    /// with the configured [recorder](CogsRecorder).
```
iker-barriocanal, line 42: Is it possible to generate multiple tokens at the same time? Are the features across all the tokens weighted automatically by `Cogs`, or does it double-count as if they were created sequentially?

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+impl CogsToken {
+    /// Cancels the COGS measurement.
+    pub fn cancel(&mut self) {
+        // No features -> nothing gets attributed.
+        self.update(AppFeatures::none());
+    }
+
+    /// Updates the app features to whicht the active measurement is attributed to.
```
iker-barriocanal:
```suggestion
    /// Updates the app features to which the active measurement is attributed to.
```

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+            }
+        }
+
+        self
+    }
+
+    /// Returns an iterator yielding an app feature and it's associated weight
+    /// normalized to the total stored weights in the range between `0.0` and `1.0`.
+    ///
+    /// Used to divide a measurement by the stored weights.
```
iker-barriocanal: nit: [docs](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#c66b1ab908674eb1a8ca25f3e7f6e423) should start with a single first line.

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+    pub fn weights(&self) -> impl Iterator<Item = (AppFeature, f32)> {
+        let total_weight: usize = self.0.into_iter().map(|(_, weight)| weight.get()).sum();
+
+        self.0.into_iter().filter_map(move |(feature, weight)| {
```
iker-barriocanal: nit: is it worth filtering out the features with a weight of 0 at the start of the method?

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+    /// Increases the `weight` of an [`AppFeature`].
+    pub fn add_weight(mut self, feature: AppFeature, weight: usize) -> Self {
+        let Some(weight) = NonZeroUsize::new(weight) else {
+            return self;
+        };
+
+        if let Some(previous) = self.0 .0.get_mut(feature) {
+            *previous = previous.saturating_add(weight.get());
+        } else {
+            self.0 .0.insert(feature, weight);
```
iker-barriocanal, line 244:
```suggestion
            self.0.0.insert(feature, weight);
```

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+        }
+
+        self
+    }
+
+    /// Sets the specified `weight` for an [`AppFeature`].
+    pub fn weight(mut self, feature: AppFeature, weight: usize) -> Self {
```
iker-barriocanal: nit and personal opinion, but I'd rather `with_weight`:
- `weight` seems like a getter.
- `set_weight` seems a bit off for a builder.

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+        let f = AppFeatures::builder()
+            .weight(AppFeature::Spans, 1)
+            .weight(AppFeature::Transactions, 1)
+            .weight(AppFeature::MetricsSpans, 0) // Noop
+            .add_weight(AppFeature::MetricsSpans, 1)
+            .weight(AppFeature::Transactions, 0) // Reset
+            .build();
+        {
+            let _token = cogs.timed(ResourceId::Relay, f);
+            std::thread::sleep(Duration::from_millis(50));
```
iker-barriocanal, line 300: Should we use tokio's sleep instead? IIUC tokio should handle the magic of knowing when to sleep and when to update other clocks, potentially increasing test speed.

relay-cogs/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+#[cfg(test)]
+pub use test::*;
+
+/// Resource ID as tracked in COGS.
+///
+/// Infrastructure costs are labeled with a resource id,
+/// these costs need to be broken down further by the application
+/// by [app features](AppFeature).
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum ResourceId {
```
iker-barriocanal, line 84: What other resource IDs may Relay have? If it's only going to be one, I'd hardcode it instead of having it as a variable.

relay-cogs/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+            Self::MetricsTransactions => 11,
+            Self::MetricsSpans => 12,
+            Self::MetricsSessions => 13,
+            Self::MetricsCustom => 14,
+            Self::MetricsUnsupported => 15,
+        }
+    }
+}
+
+// A COGS measurement.
```
iker-barriocanal:
```suggestion
/// A COGS measurement.
```

relay-cogs/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::utils::Enum;
+
+    use super::*;
+
+    #[test]
+    fn test_app_feature() {
+        for i in 0.. {
+            let Some(f) = AppFeature::from_index(i) else {
```
iker-barriocanal: Does this test only verify indexes in `from_index` -> `to_index`? Should we also verify the other way around?

relay-cogs/src/recorder.rs
```diff
@@ -0,0 +1,14 @@
+use crate::CogsMeasurement;
+
+/// Cogs consumer, recording actual measurements.
+pub trait CogsRecorder: Send + Sync {
```
iker-barriocanal: "consumer" is ambiguous from the reader's perspective, something like the following should be a bit more clear:
```suggestion
/// Cogs recorder, recording actual measurements.
pub trait CogsRecorder: Send + Sync {
```

relay-cogs/src/utils.rs
```diff
@@ -0,0 +1,97 @@
...
+            if let Some(value) = value {
+                return Some((E::from_index(index)?, value));
+            }
+        }
+
+        None
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (0, Some(self.map.0.len()))
```
iker-barriocanal: `size_hint` will always return the length of the whole map, not the remaining length of the iterator. According to [docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint), it should be the remaining length:

> Returns the bounds on the remaining length of the iterator.

iker-barriocanal (commented):
Partial review about `relay-cogs`, I'll continue later.

Open questions that may be addressed in the code I haven't reviewed yet:
- Should cogs be emitted in librelay calls? Should these be skipped, and we assume the components calling librelay are cogs-instrumented?
- Do we collect cogs measurements from external relays?

relay-config/src/config.rs
```diff
@@ -1336,6 +1336,43 @@ impl Default for Health {
...
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(default)]
+pub struct Cogs {
+    /// Whether COGS measurements are enabled.
+    enabled: bool,
+    /// Granularity of the COGS measurements.
+    ///
+    /// Measurements are aggregated based on the granularity.
+    ///
+    /// Defaults to 60 (1 minute).
```
iker-barriocanal: From the default it's kinda obvious what the unit is, but docs should still mention it:
```suggestion
    /// Granularity of the COGS measurements.
    ///
    /// Measurements are aggregated based on the granularity, in seconds.
    ///
    /// Defaults to 60 (1 minute).
```

relay-config/src/config.rs
```diff
@@ -1336,6 +1336,43 @@ impl Default for Health {
     }
 }
 
+/// COGS configuration.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(default)]
+pub struct Cogs {
+    /// Whether COGS measurements are enabled.
```
iker-barriocanal, line 1343:
```suggestion
    /// Whether COGS measurements are enabled.
    ///
    /// Disabled by default.
```

relay-metrics/src/cogs.rs
```diff
@@ -0,0 +1,50 @@
...
+/// COGS estimator based on the bucket count.
+pub struct ByCount<'a>(pub &'a [Bucket]);
+
+impl<'a> From<ByCount<'a>> for AppFeatures {
+    fn from(value: ByCount<'a>) -> Self {
+        metric_app_featurs(value.0, |_| 1)
+    }
+}
+
+fn metric_app_featurs(buckets: &[Bucket], f: impl Fn(&Bucket) -> usize) -> AppFeatures {
```
iker-barriocanal:
```suggestion
fn metric_app_features(buckets: &[Bucket], f: impl Fn(&Bucket) -> usize) -> AppFeatures {
```

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+            .field("features", &self.features)
+            .finish()
+    }
+}
+
+/// A collection of weighted [app features](AppFeature).
+///
+/// Used to attribute a single COGS measurement to multiple features.
+#[derive(Clone, Copy)]
+pub struct AppFeatures(EnumMap<16, AppFeature, NonZeroUsize>);
```
iker-barriocanal: Is it worth to mention in the docs what `AppFeatures` maps (feature -> weight)?

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+    /// Used to divide a measurement by the stored weights.
+    pub fn weights(&self) -> impl Iterator<Item = (AppFeature, f32)> {
+        let total_weight: usize = self.0.into_iter().map(|(_, weight)| weight.get()).sum();
+
+        self.0.into_iter().filter_map(move |(feature, weight)| {
+            if total_weight == 0 {
+                return None;
+            }
+
+            let ratio = (weight.get() as f32 / total_weight as f32).clamp(0.0, 1.0);
```
iker-barriocanal, line 203: Thinking out loud:

Are all ratios equivalent? Example: a ratio of 0.5 represents a usage that may (not) be the same as on high usage. Does this also depend on the configuration of k8s or the cloud provider?

relay-cogs/src/cogs.rs
olksdr: this is not a tokio test with `async` fn

relay-cogs/src/cogs.rs
Dav1dde: An additional indentation level hurts readability, your example is actually also missing a `move` and then internally the closure still needs to use a token to correctly handle all cases including panics.

relay-cogs/src/cogs.rs
Dav1dde: If you create multiple tokens then they are separate measurements and you will double count.

relay-config/src/config.rs
iker-barriocanal: IIUC the granularity also determines how often cogs are flushed. We should point that out.

relay-cogs/src/cogs.rs
Dav1dde: Open for suggestions how to word this better, this is the core of what the method does.

relay-cogs/src/cogs.rs
Dav1dde: There are no features with a weight of zero, the weight is a `NonZeroUsize`.

relay-cogs/src/cogs.rs
Dav1dde: For some reason cargo format formats it like this, it doesn't even accept the suggestion and corrects it.

relay-cogs/src/cogs.rs
Dav1dde: It follows builders like [`Command`](https://doc.rust-lang.org/std/process/struct.Command.html)

relay-cogs/src/lib.rs
Dav1dde: Once we want to start splitting costs like GCP network traffic, we'll need a second resource id. Also it does follow the [COGS Architecture](https://github.com/getsentry/etl/blob/master/workspace/dags/sql/cogs_v2/docs/architecture.md).

relay-server/src/services/processor.rs
```diff
@@ -2344,6 +2380,35 @@ impl EnvelopeProcessorService {
             }
         });
     }
+
+    fn app_features(&self, message: &EnvelopeProcessor) -> AppFeatures {
+        match message {
+            EnvelopeProcessor::ProcessEnvelope(v) => AppFeature::from(v.envelope.group()).into(),
+            EnvelopeProcessor::ProcessMetrics(_) => AppFeature::Unattributed.into(),
+            EnvelopeProcessor::ProcessBatchedMetrics(_) => AppFeature::Unattributed.into(),
```
iker-barriocanal, line 2385: Should these two be `UnattributedMetrics`?

relay-cogs/src/lib.rs
Dav1dde: It implicitly does, if `from_index` would yield the same variant twice, the assert below with `to_index` would at least fail once.

relay-cogs/src/utils.rs
Dav1dde: It's the upper bound, which is always bounded by the entire length, but you're right, we can actually give a better estimate, I'll update it.

relay-server/src/services/processor.rs
```diff
@@ -2848,7 +2913,7 @@ mod tests {
             let f = |org_ids: &mut Vec<u64>, msg: Store| {
                 let org_id = match msg {
                     Store::Metrics(x) => x.scoping.organization_id,
-                    Store::Envelope(_) => panic!("received envelope when expecting only metrics"),
+                    _ => panic!("received envelope when expecting only metrics"),
```
iker-barriocanal, line 2913: I'm not sure whether this is an improvement: extending `Store` with new variants results in this message being stale. I'd suggest either not having a catch-all, or updating the panic message to not tie it to "received envelope".

relay-cogs/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+            Self::Replays => 9,
+            Self::MetricMeta => 10,
+            Self::MetricsTransactions => 11,
+            Self::MetricsSpans => 12,
+            Self::MetricsSessions => 13,
+            Self::MetricsCustom => 14,
+            Self::MetricsUnsupported => 15,
+        }
+    }
+}
```
jjbayer: Given that we need to specify the indices + the enum size manually, why not go with

```rust
enum AppFeature {
    a = 0,
    b = 1,
    // ...
}

struct AppFeatures([NonZeroUsize; 16]);

impl AppFeatures {
    fn get(&self, index: AppFeature) -> Option<&NonZeroUsize> {
        self.0.get(index as usize)
    }
}
```

Would save a lot of LoC, plus the compiler even ensures that you do not use the same discriminant twice.

relay-cogs/src/cogs.rs
iker-barriocanal: Then `total_weight` can never be 0, and we can remove the `if` on line 183?

relay-server/src/services/cogs.rs
```diff
@@ -0,0 +1,183 @@
...
+impl CogsService {
+    pub fn new(config: &Config, #[cfg(feature = "processing")] store: Option<Addr<Store>>) -> Self {
+        #[cfg(feature = "processing")]
+        let producer = store
+            .map(RelayProducer::Store)
+            .filter(|_| config.cogs_enabled())
+            .unwrap_or(RelayProducer::Noop);
+
+        #[cfg(not(feature = "processing"))]
+        let producer = RelayProducer::Noop;
```
iker-barriocanal, line 62: Are there plans to extend the support for PoP relays too?

iker-barriocanal (approved):
LGTM, mostly minor comments and questions.

relay-cogs/src/cogs.rs
jjbayer: I personally like the RAII approach, but it's true that in other places in the Relay code base we use the lambda pattern (sentry SDK, metrics).

relay-metrics/src/cogs.rs
```diff
@@ -0,0 +1,50 @@
...
+use relay_cogs::{AppFeature, AppFeatures};
+
+use crate::{Bucket, BucketView};
+
+/// COGS estimator based on the estimated size of each bucket in bytes.
+pub struct BySize<'a>(pub &'a [Bucket]);
+
+impl<'a> From<BySize<'a>> for AppFeatures {
+    fn from(value: BySize<'a>) -> Self {
+        metric_app_featurs(value.0, |b| BucketView::new(b).estimated_size())
```
jjbayer:
```suggestion
        metric_app_features(value.0, |b| BucketView::new(b).estimated_size())
```

relay-cogs/src/cogs.rs
Dav1dde: The doc for `AppFeatures` starts with:

```
/// A collection of weighted [app features](AppFeature).
```

Do you think this needs more or a better explanation?

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+impl From<AppFeature> for AppFeatures {
+    fn from(value: AppFeature) -> Self {
+        Self::new(value)
+    }
+}
+
+/// A builder for [`AppFeatures`] which can be used to configure different weights per [`AppFeature`].
+pub struct AppFeaturesBuilder(AppFeatures);
```
jjbayer: Going off my previous comment, I still believe we do not need a builder pattern / a wide interface here. IMO a method `CogsToken::add_weight(&mut self, feature: AppFeature, weight: f32)` could be enough.

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,336 @@
...
+            .weight(AppFeature::Transactions, 0) // Reset
+            .build();
+        {
+            let _token = cogs.timed(ResourceId::Relay, f);
+            std::thread::sleep(Duration::from_millis(50));
+        }
+
+        let measurements = recorder.measurements();
+        assert_eq!(measurements.len(), 2);
+        dbg!(&measurements);
```
jjbayer:
```suggestion
        dbg!(&measurements);
```

relay-server/src/services/processor.rs
Dav1dde: No, this is actually a bit ugly.  They are unattributed because they are later updated with the "correct" weights, we only get these weights after we processed (parsed) the contained payload.

I kept the default here unattributed that if this shows up in the COGS somewhere we know that this shouldn't be there (either a bug or a small edge case), I tried explaining this in the `AppFeature::Unattributed` docs but maybe this is not clear enough.

relay-server/src/services/processor.rs
Dav1dde: That test only expects metrics, anything else should never arrive -> fail the test, I don't want to update the test every time we add more functionality to the store.

relay-cogs/src/cogs.rs
Dav1dde: It can, if there are zero elements

relay-server/src/services/cogs.rs
Dav1dde: Yes! One of the reasons the resource id is mapped from the config and just a constant.

relay-cogs/src/lib.rs
Dav1dde: That is already possible with the current `AppFeature`, it implicitly starts at 0 (I'll actually update `to_index` to use the cast, didn't think about that). 

But we still need to convert back from an index to the enum variant, which comes down to the same `from_index`. I initially had all of this inlined but it just became unreadable and I preferred having the array logic separate from the `AppFeatures`.

relay-cogs/src/cogs.rs
Dav1dde: Ideally I don't want the token to be mutated at all, unfortunately it's necessary, that's why there is the `update`. 

I see a problem when you can configure the individual weights on the token itself that it is very easy to mess up, tokens and individual weights can be dynamically added in different places, you can forget that there was already a weight and you just add where you should've replaced or cleared some other weight.

The current interface makes this impossible, that's also why `AppFeatures` is immutable, you can only build new weights from scratch (or merge). Hence the builder to have a separate phase of mutable weights, but once created it's immutable.

relay-cogs/src/cogs.rs
Dav1dde: The cogs stuff only works on measured, actual cost. If I rate limit metrics twice and the first time it may take 1ms and the second time 5 ms, each time that cost is divided according to the weights. For the first measurement 0.5ms will be attributed (0.5 ratio) and for the second one 2.5ms, which totals to 3ms.

At the end of the day you can sum up all measurements and calculate the final ratios of everything and these ratios you can then apply to your actual costs.

Dav1dde:
@iker-barriocanal I update the PR description to answer the question which came up

relay-server/src/services/processor.rs
iker-barriocanal: The explanation is good, I just was wondering about the default value.

relay-cogs/src/cogs.rs
Dav1dde: Your example becomes:

```rs
let mut accountant = CogsToken::new();

// ...
let features = AppFeatures::builder();
for bucket in buckets {
    let feature = AppFeature::from(bucket.namespace);
    features.add_weight(feature, bucket.size());
}
accountant.update(features);

// in drop handler, accountant divides up elapsed time proportional to attributed numbers.
```

Or better:

```rs
let mut accountant = CogsToken::new();

// ...
accountant.update(BySize(&bucktes));

// in drop handler, accountant divides up elapsed time proportional to attributed numbers.
```

relay-server/src/services/processor.rs
```diff
@@ -1616,38 +1638,35 @@ impl EnvelopeProcessorService {
...
-                self.inner
-                    .project_cache
-                    .send(MergeBuckets::new(public_key, buckets));
             } else if item.ty() == &ItemType::MetricBuckets {
                 match serde_json::from_slice::<Vec<Bucket>>(&payload) {
-                    Ok(mut buckets) => {
-                        for bucket in &mut buckets {
+                    Ok(mut b) => {
+                        for bucket in &mut b {
                             clock_drift_processor.process_timestamp(&mut bucket.timestamp);
```
jjbayer: Could we pull the clock_drift_processor call outside the loop as well?

relay-server/src/services/processor.rs
```diff
@@ -1616,38 +1638,35 @@ impl EnvelopeProcessorService {
         let clock_drift_processor =
             ClockDriftProcessor::new(sent_at, received).at_least(MINIMUM_CLOCK_DRIFT);
 
+        let mut buckets: Option<Vec<Bucket>> = None;
```
jjbayer: Would this work without the option? An empty vector does not allocate.

```suggestion
        let mut buckets = vec![];
```

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    /// Sets the specified `weight` for an [`AppFeature`].
+    pub fn weight(&mut self, feature: AppFeature, weight: usize) -> &mut Self {
+        if let Some(weight) = NonZeroUsize::new(weight) {
+            self.0 .0.insert(feature, weight);
+        } else {
+            self.0 .0.remove(feature);
+        }
+        self
+    }
```
jjbayer: Do we need this method? It's only used in tests and in `AppFeatures::new()` abd there we could use `add_weight`, if I understood the purpose of these methods correctly.

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,335 @@
...
+            .field("features", &self.features)
+            .finish()
+    }
+}
+
+/// A collection of weighted [app features](AppFeature).
+///
+/// Used to attribute a single COGS measurement to multiple features.
+#[derive(Clone, Copy)]
+pub struct AppFeatures(EnumMap<16, AppFeature, NonZeroUsize>);
```
jjbayer: nit

```suggestion
pub struct FeatureWeights(EnumMap<16, AppFeature, NonZeroUsize>);
```

relay-cogs/src/lib.rs
```diff
@@ -0,0 +1,240 @@
+//! Break down the cost of Relay by its components and individual features it handles.
+//!
+//! Relay is a one stop shop for all different kinds of events Sentry supports, Errors,
+//! Performance, Metrics, Replays, Crons and more. A single shared resource.
+//!
+//! This module intends to make it possible to give insights how much time and resources
+//! Relay spends processing individual features. The measurements collected can be later used
+//! for increased observability and accounting purposes.
+//!
+//! `relay-cogs` provides a way to to give an answer to the questions:
```
olksdr:
```suggestion
//! `relay-cogs` provides a way to give an answer to the questions:
```

relay-cogs/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    UnattributedMetrics,
+    /// When processing an envelope cannot be attributed or is not feasible to be attributed
+    /// to a more specific category, this app feature is emitted instead.
+    UnattributedEnvelope,
+
+    /// Transactions.
+    Transactions,
+    /// Errors.
+    Errors,
+    /// Spans.
```
olksdr, line 120: Is this granularity is enough ?  What I mean is e.g. transactions can also contain profiles  or transactions and errors can also contain attachments - do we want to have separate `AppFeature` for those as well?

Dav1dde: Probably makes sense at some point, I'll make some issues with improvements. 

For now these groups follow what we did before manually and hopefully the results match...

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,335 @@
...
+            }
+        }
+
+        self
+    }
+
+    /// Returns an iterator yielding an app feature and it's associated weight
+    /// normalized to the total stored weights in the range between `0.0` and `1.0`.
+    ///
+    /// Used to divide a measurement by the stored weights.
```
jjbayer:
```suggestion
    /// Returns an iterator yielding an app feature and its normalized associated weight.
    ///
    /// Weights are normalized such that the sum over all weights equals `1.0`.
```

relay-cogs/src/cogs.rs
```diff
@@ -0,0 +1,335 @@
...
+            recorder: Arc::clone(&self.recorder),
+        }
+    }
+}
+
+/// An in progress COGS measurement.
+///
+/// The measurement is recorded when the token is dropped.
+#[must_use]
+pub struct CogsToken {
```
jjbayer: nit: We have a guideline about avoiding redundant prefixes: https://develop.sentry.dev/rust/#naming

I think for the main type `Cogs` it makes sense to keep the name, but for all other types we could remove the prefix.

```suggestion
pub struct Token {
```

relay-server/src/services/processor.rs
jjbayer: @Dav1dde could you add a comment here to point out that their weights will be updated later?

relay-cogs/src/lib.rs
jjbayer: How about we just use a `BTreeMap` then? Do we have reason to believe that those few allocations will have a measurable effect on performance?

jjbayer (approved):
No blockers, but I'm kind of unhappy with the 1300 lines of code this adds 😄.

PR #3160, opened 2024-02-23
instr(kafka): platform tag

jjbayer:
Add a `platform` tag to the metric that counts span Kafka messages. We can remove this tag again once the investigation is done.

#skip-changelog

### DIFF
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -489,6 +489,7 @@ mod tests {
                     ],
                 },
             ),
+            platform: ~,
             other: {},
         }
         "###);
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -91,6 +91,12 @@ pub struct Span {
     #[metastructure(skip_serialization = "empty")]
     pub _metrics_summary: Annotated<MetricsSummary>,
 
+    /// Platform identifier.
+    ///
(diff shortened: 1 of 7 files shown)


relay-event-schema/src/protocol/span.rs
```diff
@@ -91,6 +91,12 @@ pub struct Span {
     #[metastructure(skip_serialization = "empty")]
     pub _metrics_summary: Annotated<MetricsSummary>,
 
+    /// Platform identifier.
+    ///
+    /// See [`Event::platform`].
+    #[metastructure(skip_serialization = "empty")]
+    pub platform: Annotated<String>,
```
jjbayer, line 98: It may seem wrong to add a new field to the span type for this, but

1. `Span` has `additional_properties` with `retain=true`, so SDKs can already send this field whether we declare it or not.
2. We only set this field in Relay on spans extracted from events, not in the event itself, so it won't end up in Nodestore.

relay-server/src/services/store.rs
```diff
@@ -968,6 +968,7 @@ impl StoreService {
 
         let is_segment = span.is_segment;
         let has_parent = span.parent_span_id.is_some();
+        let platform = VALID_PLATFORMS.iter().find(|p| *p == &span.platform);
```
jjbayer, line 971: Checking `VALID_PLATFORMS` to make sure cardinality is low, and to get a static `&str`.

olksdr (approved):

PR #3161, opened 2024-02-23
feat(global-filters): Introduce global generic filters

iker-barriocanal:
This PR adds support for global generic filters.

Generic filters are a new schema of inbound filters with generic conditions to avoid creating a new filter for each use case.
Project configs already support [project-specific generic filters](https://github.com/getsentry/relay/commit/ee3036cd33d6458f18c11edf4f5ba985bd03d14b). Global generic filters have been designed to be extendable, future-compatible, independent, and flexible to work well in isolation but also with project generic filters.

For event processing, Relay uses both the project config of the event and the global config. Before running the filters, Relay will merge the generic filters from the two configs and apply these to the event. If there's a match relay will drop the event and generate an outcome, or forward the event to the next processing step if there isn't.

## Why

Currently, adding a new generic filter results in an increase in the size of project configs (`config_size * number_of_filters`). This can result in a significant increase in memory usage in all the services that store project configs (`increase_per_project * number_of_projects`), especially Relay. Incorporating generic filters into global configs allows Relay to apply these filters similarly with minimal overhead.

There are additional, less important goals that this approach also achieves: not increasing network usage, not impacting project config computation time, etc.

## Previous work

Generic inbound filters were introduced in project configs in https://github.com/getsentry/relay/commit/ee3036cd33d6458f18c11edf4f5ba985bd03d14b. They are currently not enabled on `sentry` as supporting these filters in the global config was prioritized first.

## Protocol

Global generic filters have the same protocol as project generic filters, but they exist in the global config. A new optional `filters` key is added as a top-level key in global configs. This object expects the following properties:
- `version`: the version of global configs, for future-compatibility, as a `u16`.
- `filters`: a list of generic filters sorted by descending priority, with the same protocol as with the project generic filters.

Example of a full protocol:
```json
{
	// ... other global config keys ...
	"filters": {
		"version": 1,
		"filters": [
			{
				"id": "sample-errors",
				"isEnabled": true,
				"condition": {
					"op": "eq",
					"name": "event.exceptions",
					"value": "Sample Error"
				}
			}
		]
	}
}
```

### Protocol

Supporting external Relays with future compatibility to make updates safely:
- Introduced `version` to skip filters on a version not supported by Relay.
- When filters are unsupported, Relay forwards events upstream without extracting metrics. This allows the next updated Relay in the chain supporting the filters to run filtering and, in case there's a decision to keep the event, extract metrics from it.

Relay merges the configs prioritizing the project config. Each filter has an `isEnabled` flag to maximize flexibility and allow:
- Defining a filter for a single project in its config.
- Defining a filter for all projects in the global config.
- Defining a filter for most projects by enabling it in the global config and disabling it in the applicable project configs.
- Defining a filter for some projects by disabling it in the global config and enabling it in the applicable project configs.
This is the simplest alternative to support all the use cases.

### Relay implementation details

Relay types the filters internally as an ordered map with custom (de)serialization, keeping the order in the protocol, to benefit from:
- `O(1)` retrieval of filters given the ID.
- Removing duplicated filters.
Removing duplicates is a prerequisite to merge the filters from both configs properly. Since the list of filters in the payload is expected to be sorted by descending priority, only the first occurrence of a filter is selected and the rest are discarded during deserialization. Discarded filters are not forwarded downstream.

`filters` is typed as an `ErrorBoundary` instead of an `Option`:
- Errors during deserialization indicate breaking changes. Relay skips metric extraction and dynamic sampling in these cases, and the next relay in chain deals with them.
  - Note: processing relays will run dynamic sampling and extract metrics even if filters are broken.
- An empty entry in a payload is deserialized as an empty list of filters, effectively a noop.
- A list of filters in a supported version is appropriately deserialized and applied.


relay-filter/src/config.rs
```diff
@@ -248,13 +251,117 @@ impl GenericFilterConfig {
...
+///         "name": "event.exceptions",
+///         "value": "drop-error"
+///       }
+///     }
+///   ]
+/// }"#);
+/// ```
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GenericFiltersConfig {
```
iker-barriocanal, line 405: Since this is used in `should_filter`, which is public, should also be public.

relay-filter/src/config.rs
```diff
@@ -328,15 +482,209 @@ impl FiltersConfig {
     }
 }
 
+/// Configuration of generic filters combined from project and global config.
+///
+/// See [`CombinedFiltersConfigIter`] for details on how to iterate easily
+/// through the applicable filters.
+#[derive(Copy, Clone, Debug)]
+pub(crate) struct CombinedFiltersConfig<'a> {
```
iker-barriocanal: The module structure is not great, which required the `pub(crate)` here. Refactoring to a proper structure is out of scope of this PR though, I may approach it in the future.

tests/integration/test_metrics.py
```diff
@@ -1880,3 +1879,121 @@ def test_metric_bucket_encoding_dynamic_global_config_option(
...
+                "debounce_delay": 0,
+                "shift_key": "none",
+            }
+        }
+    )
+    transaction = generate_transaction_item()
+    relay.send_transaction(project_id, transaction)
+
+    tx, _ = tx_consumer.get_event()
+    assert tx is not None
```
iker-barriocanal: `get_event` already asserts the transaction is `None`, but that's not super clear so I added the check here for readability.

tests/integration/test_metrics.py
```diff
@@ -1851,8 +1852,6 @@ def test_metric_bucket_encoding_dynamic_global_config_option(
             namespace: "array"
         }
 
-    print(mini_sentry.global_config)
```
iker-barriocanal, line 1854: Note to self: investigate why this is not caught by automation.

relay-filter/src/config.rs
```diff
@@ -248,13 +251,117 @@ impl GenericFilterConfig {
...
+/// the filter's id and the value is the filter itself. The map is converted
+/// back to a list when serializing it, without the filters that were discarded
+/// as duplicates. See examples below.
+///
+/// # Examples
+///
+/// Deserialization:
+///
+/// ```
+/// # use relay_filter::GenericFiltersConfig;
```
iker-barriocanal, line 294: I decided to add the (de)serialization tests as doctests as the description above may not be 100% clear and I was going to add tests anyway.

relay-filter/src/config.rs
```diff
@@ -248,13 +251,117 @@ impl GenericFilterConfig {
...
-    pub filters: Vec<GenericFilterConfig>,
+    /// Map of generic filters, sorted by the order in the payload from upstream.
+    ///
+    /// The map contains unique filters, meaning there are no two filters with
+    /// the same id. See struct docs for more details.
+    #[serde(
+        deserialize_with = "deserialize_generic_filters",
+        serialize_with = "serialize_generic_filters"
+    )]
+    pub filters: IndexMap<String, GenericFilterConfig>,
```
olksdr: This might be already answered: Do we definitely want to keep the order of the rules? Does the order of the rules have some special meaning?

olksdr: Do we want to skip serialization when the `filters` is empty ?  Or your customer serialization function is handling this?

relay-filter/src/config.rs
olksdr: why not just `pub`?

relay-filter/src/config.rs
```diff
@@ -328,15 +482,208 @@ impl FiltersConfig {
...
+
+            if let Some(global_filters) = self.config.global_filters {
+                if self.global_index < global_filters.filters.len() {
+                    let filter = self.next_global_filter();
+                    if filter.is_some() {
+                        return filter;
+                    }
+                    continue;
+                }
+            }
```
olksdr: Does this mean we have one filter from project config and then from global config, and next iteration the same? 
Don't we want to iterate first all project config filters and only then all global config filters? 

Or maybe I"m getting this wrong?

olksdr (commented):
Seems like all your tests have max 2 filters in the config. Can we add at least 5 to each of the location and make sure to test that they are applied in according to the rules describes in this PR?

relay-filter/src/config.rs
```diff
@@ -328,15 +482,208 @@ impl FiltersConfig {
     }
 }
 
+/// Configuration of generic filters combined from project and global config.
+///
+/// See [`CombinedFiltersConfigIter`] for details on how to iterate easily
+/// through the applicable filters.
+#[derive(Copy, Clone, Debug)]
+pub(crate) struct CombinedFiltersConfig<'a> {
```
TBS1996: so currently we have used the `Dynamic` prefix for these combined global and project types. I'm fine with changing to `Combined` but then we should change the other instances too.

relay-dynamic-config/src/global.rs
```diff
@@ -23,6 +25,9 @@ pub struct GlobalConfig {
     /// Quotas that apply to all projects.
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub quotas: Vec<Quota>,
+    /// Configuration for global inbound filters.
+    #[serde(skip_serializing_if = "skip_default_error_boundary")]
+    pub filters: ErrorBoundary<GenericFiltersConfig>,
```
TBS1996, line 34: it seems similar to `MeasurementsConfig` in it's form. So maybe we should also add errorboundary to the measurementsconfig in another PR if it makes sense to do it in genericfiltersconfig 🤔

relay-filter/src/config.rs
iker-barriocanal: Rules are sorted by priority, so the order of the rules is preserved, and `IndexMap` ensures that (see [docs](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#order)):

> The key-value pairs have a consistent order that is determined by the sequence of insertion and removal calls on the map.

relay-filter/src/config.rs
iker-barriocanal: It's not needed. `CombinedFiltersConfig` (and `CombinedFiltersConfigIter`) is only used internally in `relay-filter`, so exposing it seems unnecessary. I'm ok making it `pub` if you have a preference for it.

relay-dynamic-config/src/global.rs
iker-barriocanal: What's the uncovered use case that adding `ErrorBoundary` to `MeasurementsConfig` covers?

relay-filter/src/config.rs
iker-barriocanal: The `loop` iterates through all project filters first, until `project_index` is too big and thus has evaluated all project filters. Then, the iterator moves on to global filters, and repeats the same approach. I added a small comment in [950325d](https://github.com/getsentry/relay/pull/3161/commits/950325dff6dc769f46dfcc6bb4122962dacd3f42) to provide a bit more visibility.

I acknowledge this piece of code is not straightforward to understand, so I'm open to suggestions on how to make it more readable!

relay-filter/src/config.rs
iker-barriocanal: Good point! Addressed in [950325d](https://github.com/getsentry/relay/pull/3161/commits/950325dff6dc769f46dfcc6bb4122962dacd3f42).

relay-filter/src/config.rs
iker-barriocanal: Good question!
- In `GenericFiltersConfig`, an empty filter map is serialized into an empty list.
- In the global config, filters are not serialized on `GenericFiltersConfig` deserialization errors or when it's the default value (no filters and `version` is 0 (default value of `u16`)).

Expanding on the second point: we shouldn't skip serialization on empty filters without considering the version, as a higher version can disable metric extraction in downstream relays. Sending empty filters doesn't cause deserialization errors.

Cargo.toml
```diff
@@ -26,6 +26,7 @@ insta = { version = "1.31.0", features = ["json", "redactions", "ron"] }
 hash32 = "0.3.1"
 hashbrown = "0.14.3"
 itertools = "0.10.5"
+indexmap = "2.0.0"
```
Dav1dde, line 28: We could bump the dependency to `2.2.3` (latest)

relay-dynamic-config/src/global.rs
```diff
@@ -23,6 +25,9 @@ pub struct GlobalConfig {
     /// Quotas that apply to all projects.
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub quotas: Vec<Quota>,
+    /// Configuration for global inbound filters.
```
Dav1dde, line 29:
```suggestion
    /// Configuration for global inbound filters.
    ///
    /// The configuration in the global config is later merged with the more specific
    /// configuration from the project state.
```

Something along the lines that this is only have the truth, what do you think?

relay-dynamic-config/src/global.rs
Dav1dde, line 34: Do we want to be a bit more specific here, like `inbound_filters`?

relay-dynamic-config/src/global.rs
```diff
@@ -46,6 +58,21 @@ impl GlobalConfig {
...
+        match &self.filters {
+            ErrorBoundary::Err(_) => None,
+            ErrorBoundary::Ok(f) => Some(f),
+        }
+    }
+
+    /// Returns the version of generic inbound filters.
+    ///
+    /// If the filters failed to deserialize, [`u16::MAX`] is returned.
+    pub fn generic_filters_version(&self) -> u16 {
```
Dav1dde: Any reason this doesn't just return an `Option`, the special case with `u16::MAX` seems non obvious at first and I think it makes it easier to understand if the caller has that logic.

Dav1dde: The other method is called `filters` we should align the naming.

relay-filter/src/config.rs
```diff
@@ -248,13 +251,117 @@ impl GenericFilterConfig {
     }
 }
 
+impl PartialEq for GenericFilterConfig {
+    fn eq(&self, other: &Self) -> bool {
+        self.id == other.id
```
Dav1dde: While this impl does make sense it is scary, it may not be immediately obvious that the impl only compares the IDs, e.g. if you want to compare an enabled to a disabled filter with the same ID you get `true`

relay-filter/src/config.rs
```diff
@@ -248,13 +251,117 @@ impl GenericFilterConfig {
...
+    }
+}
+
 /// Configuration for generic filters.
-#[derive(Clone, Debug, Default, Serialize, Deserialize)]
-pub(crate) struct GenericFiltersConfig {
+///
+/// # Deserialization
```
Dav1dde, line 257: Awesome docs!

We should doctests + examples more often.

relay-filter/src/config.rs
Dav1dde: @olksdr Do you mean skipping the entire object or just the `filters` field here? I think skipping the field we can always do.

Re skipping the object (which means we also skip the version) is a bit tricky and I'd probably default to always serializing the object if we got one from upstream.

relay-filter/src/config.rs
```diff
@@ -264,10 +371,57 @@ impl GenericFiltersConfig {
...
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            formatter.write_str("a vector of filters: Vec<GenericFilterConfig>")
+        }
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: de::SeqAccess<'de>,
+        {
+            let mut filters = IndexMap::new();
```
Dav1dde: `SeqAccess` gives you a `size` hint which we can pass on to `IndexMap::with_capacity`, something like `IndexMap::with_capacity(seq.size_hint().unwrap_or(0))`

relay-filter/src/config.rs
```diff
@@ -264,10 +371,57 @@ impl GenericFiltersConfig {
...
+
+fn serialize_generic_filters<S>(
+    index_map: &IndexMap<String, GenericFilterConfig>,
+    serializer: S,
+) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let mut seq = serializer.serialize_seq(Some(index_map.len()))?;
+    for (_, filter) in index_map.iter() {
```
Dav1dde:
```suggestion
    for filter in index_map.values() {
```

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+    /// It applies to all filters, from both project and global configs.
+    max_version: u16,
+}
+
+impl<'a> DynamicFiltersConfig<'a> {
+    /// Creates a [`DynamicFiltersConfig`] from the project and global configs.
+    pub fn new(
+        project_filters: &'a GenericFiltersConfig,
+        global_filters: Option<&'a GenericFiltersConfig>,
+        max_version: u16,
```
Dav1dde: This confused me at first, this is the supported version, initially I thought it was the `global_version.max(project_version)`.

```suggestion
        supported_version: u16,
```

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+impl<'a> IntoIterator for DynamicFiltersConfig<'a> {
+    type Item = &'a GenericFilterConfig;
+    type IntoIter = DynamicFiltersConfigIter<'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        DynamicFiltersConfigIter::new(self)
+    }
+}
+
+/// Iterator over [`GenericFilterConfig`] of a project and global config.
```
Dav1dde: Good docs, but I think they should be on `DynamicFiltersConfig` to have a higher chance to be seen. I'd probably never check the docs on the return value of `into_iter()`.

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+#[derive(Debug)]
+pub(crate) struct DynamicFiltersConfigIter<'a> {
+    /// Configuration of project and global filters.
+    config: DynamicFiltersConfig<'a>,
+    /// Index of the next filter in project config to evaluate.
+    project_index: usize,
+    /// Index of the next filter in global config to evaluate.
+    global_index: usize,
+    /// Filters that have been evaluated, either yielded or ignored.
+    evaluated: HashSet<&'a String>,
```
Dav1dde:
```suggestion
    evaluated: HashSet<&'a str>,
```

relay-dynamic-config/src/global.rs
jjbayer: If we treat `ErrorBoundary::Err` like `None`, why not use `default_on_error`? That could even make the getters below unnecessary.

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+    ///
+    /// Filters from the project and global configs are complementary, and in
+    /// isolation, they don't provide enough valuable information to perform the
+    /// filtering. Additionally, new versions may include features not supported
+    /// in the current Relay.
+    fn is_version_supported(&self) -> bool {
+        if self.config.max_version < self.config.project_filters.version {
+            return false;
+        }
+        if let Some(gc) = self.config.global_filters {
```
Dav1dde: Shorter and imo easier to understand:

```rs
self.config.project_filters.version <= self.config.max_version
    && self.config.global_filters.map_or(true, |gc| gc.version <= self.config.max_version)
```

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+
+        // This loop starts by iterating through all project filters: increments
+        // `project_index` on every iteration until it grows too much (i.e. all
+        // project filters are evaluated), yielding the evaluated filter if
+        // necessary.
+        // Then, the same approach is applied to global filters.
+        // `continue` restarts the loop, so global filters aren't evaluated
+        // until all project filters are.
+        loop {
+            if self.project_index < self.config.project_filters.filters.len() {
```
Dav1dde: I think this check should be in `self.next_project_filter`

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+                let filter = self.next_project_filter();
+                if filter.is_some() {
+                    return filter;
+                }
+                continue;
+            }
+
+            if let Some(global_filters) = self.config.global_filters {
+                if self.global_index < global_filters.filters.len() {
```
Dav1dde: If we also move this check into `self.next_global_filter`, the `if let Some()` is no longer necessary.

And we can write the entire loop as:

```rs
self.next_project_filter().or_else(self.next_global_filter)
```

relay-filter/src/config.rs
```diff
@@ -264,10 +371,57 @@ impl GenericFiltersConfig {
...
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: de::SeqAccess<'de>,
+        {
+            let mut filters = IndexMap::new();
+            while let Some(filter) = seq.next_element::<GenericFilterConfig>()? {
+                if filters.contains_key(&filter.id) {
+                    continue;
+                }
+                filters.insert(filter.id.clone(), filter);
```
jjbayer: nit: 
```suggestion
                if !filters.contains_key(&filter.id) {
                    filters.insert(filter.id.clone(), filter);
                }  
```

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+            .filters
+            .get_index(self.project_index)
+        else {
+            return None;
+        };
+        self.project_index += 1;
+
+        // Mark the filter as evaluated first. If the filter is disabled, it
+        // should not be applied from global configs.
+        let newly_evaluated = self.evaluated.insert(id);
```
Dav1dde: `newly_evaluated == false` is impossible right? Because it is the map key.

relay-filter/src/config.rs
```diff
@@ -248,13 +251,117 @@ impl GenericFilterConfig {
...
     pub version: u16,
-    /// List of generic filters.
-    pub filters: Vec<GenericFilterConfig>,
+    /// Map of generic filters, sorted by the order in the payload from upstream.
+    ///
+    /// The map contains unique filters, meaning there are no two filters with
+    /// the same id. See struct docs for more details.
+    #[serde(
+        deserialize_with = "deserialize_generic_filters",
+        serialize_with = "serialize_generic_filters"
```
jjbayer: nit: See https://serde.rs/field-attrs.html#with

```suggestion
        with = "module_name",
```

relay-filter/src/config.rs
```diff
@@ -264,10 +371,57 @@ impl GenericFiltersConfig {
...
+        seq.serialize_element(filter)?;
+    }
+    seq.end()
+}
+
+/// Configuration for all event filters from project configs.
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct FiltersConfig {
+pub struct ProjectFiltersConfig {
```
jjbayer, line 507: What's the reason for this rename?

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
     }
 }
 
+/// Configuration of generic filters combined from project and global config.
+///
+/// See [`DynamicFiltersConfigIter`] for details on how to iterate easily
+/// through the applicable filters.
+#[derive(Copy, Clone, Debug)]
+pub(crate) struct DynamicFiltersConfig<'a> {
```
jjbayer: Doesn't have to be exactly this name, but `Generic` should be part of it IMO.

```suggestion
pub(crate) struct DynamicGenericFiltersConfig<'a> {
```

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+            let Some(global_filters) = self.config.global_filters else {
+                return None;
+            };
+            let Some(filter) = global_filters.filters.get(id) else {
+                return None;
+            };
+            return Some(filter);
+        }
+
+        Some(filter)
```
Dav1dde:
```suggestion
        if filter.condition.is_some() {
            return Some(filter)
        } else {
            self.config.global_filters.and_then(|filters| filters.get(id))
            // .filter(|filter| filter.condition.is_some()) <- I think that makes sense but would be new behaviour
        }
```

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+        if filter.is_empty() {
+            return None;
+        }
+
+        let newly_added = self.evaluated.insert(id);
+        if !newly_added {
+            return None;
+        }
+
+        Some(filter)
```
Dav1dde:
```suggestion
        self.evaluated.insert(id).then_some(filter)
```

(Maybe I like combinators too much)

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+    }
+}
+
+impl<'a> Iterator for DynamicFiltersConfigIter<'a> {
```
Dav1dde: Iterator can be fused, we should implement it according to the [guidelines](https://develop.sentry.dev/rust/#iterators).

tests/integration/test_filters.py
```diff
@@ -282,3 +282,37 @@ def test_ignore_transactions_filters_are_applied(
     else:
         event, _ = transactions_consumer.get_event()
         assert event["transaction"] == transaction_name
+
+
+def test_global_filters_drop_events(
```
Dav1dde, line 287: An additional test that has a more complex scenario of combining global and project state would be nice.

Can be one test that tests a bunch of things like:
- Disabled in the global config but enabled in project state
- Enabled in global config, disabled in project state
- Enabled in both
- Disabled in both

Could also include version mismatches and differences in the tests but you already have separate tests for that 👍

tests/integration/test_metrics.py
Dav1dde: Oops, I think that was me 🤣

tests/integration/test_metrics.py
```diff
@@ -1880,3 +1879,121 @@ def test_metric_bucket_encoding_dynamic_global_config_option(
...
+def test_missing_global_filters_enables_metric_extraction(
+    mini_sentry,
+    relay_with_processing,
+    transactions_consumer,
+    metrics_consumer,
+):
+    metrics_consumer = metrics_consumer()
+    transactions_consumer = transactions_consumer()
+
+    mini_sentry.global_config.pop("filters")
```
Dav1dde, line 2030: This shouldn't be necessary, do the tests fail without?

relay-server/src/services/processor.rs
```diff
@@ -1274,12 +1274,22 @@ impl EnvelopeProcessorService {
         event::finalize(state, &self.inner.config)?;
         self.light_normalize_event(state)?;
         dynamic_sampling::normalize(state);
-        event::filter(state)?;
+        let global_config = self.inner.global_config.current();
+        event::filter(state, global_config.filters())?;
```
Dav1dde: I think I prefer `event::filter` takes a reference to the entire global config and then just picks the filters out of it, gives us a little bit more type safety.

relay-server/src/services/processor.rs
```diff
@@ -1274,12 +1274,22 @@ impl EnvelopeProcessorService {
...
         dynamic_sampling::run(state, &self.inner.config);
 
+        // Don't extract metrics if relay can't apply generic inbound filters.
+        // An inbound filter applied in another up-to-date relay in chain may
+        // need to drop the event, and there should not be metrics from dropped
+        // events.
+        let supported_generic_filters = state.project_state.config.filter_settings.generic.version
```
Dav1dde: This logic should be hidden away in the `relay-filters` module and `GENERIC_FILTERS_VERSION` shouldn't be public.

And maybe a crazy idea: what if `event::filter` returns an additional struct/enum containing that info?

relay-filter/src/config.rs
jjbayer: Might just be me but I find `if ... else if ... else` much easier to understand than multiple continue statements.

```suggestion
            } else if let Some(global_filters) = self.config.global_filters && self.global_index < global_filters.filters.len() {
```

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+                if self.global_index < global_filters.filters.len() {
+                    let filter = self.next_global_filter();
+                    if filter.is_some() {
+                        return filter;
+                    }
+                    continue;
+                }
+            }
+
+            return None;
```
jjbayer:
```suggestion
            } else {
               return None;
            }
```

relay-filter/src/config.rs
Dav1dde: I'd prefer to keep this scoped to the module, this has quite some heavy logic attached to it and shouldn't be used outside of the module.

relay-filter/src/config.rs
```diff
@@ -328,15 +482,215 @@ impl FiltersConfig {
...
+        DynamicFiltersConfig {
+            project_filters,
+            global_filters,
+            max_version,
+        }
+    }
+}
+
+impl<'a> IntoIterator for DynamicFiltersConfig<'a> {
+    type Item = &'a GenericFilterConfig;
```
jjbayer: I believe we can simplify the iterator logic by a lot if we had an `Item` type `MergedGenericFilterConfig<'a>` here, that offers methods to retrieve the merged config. Then the iterator would be something like

```rust
match project_filters.next() {
    Some(pf) => return MergedConfig::new(pf, global_filters.get(pf.id)),
    None => match global_filters.next() {
        Some(gf) => return MergedConfig::new(gf, None),
        None => None,
    }
}
```

relay-filter/src/config.rs
olksdr: Then I would keep it private all together.

relay-filter/src/config.rs
olksdr: ~Yeah, entire object~ scratch that..  To skip `filters` here .

relay-filter/src/config.rs
iker-barriocanal: Assuming the merging logic happens in `MergedGenericFilterConfig` or `MergedConfig`, how would the indexes for the next filter be updated? IIUC this can't happen in the `Item` of the iterator.

On the other hand, I don't like creating an abstraction that returns an abstraction over filters. As the client of the API, I just want to easily get the filters back in an easy single step.

relay-filter/src/config.rs
iker-barriocanal: This PR introduces some additional concepts around filters, and `FiltersConfig` is somewhat ambiguous about what the actual config is about. In the future, reading `FiltersConfig` would make me check the docs or go to the definition, whereas I believe `ProjectFiltersConfig` won't.

If there are opinions against the rename, I'm happy to revert it.

relay-dynamic-config/src/global.rs
iker-barriocanal: We need a way to tell the difference between not being able to deserialize the configs (e.g. unsupported) or not getting any config at all. This influences the decision to disable metric extraction. I tried to document this as bullet points in the last trade-off in the PR description, let me know if I could make it easier to read.

tests/integration/test_metrics.py
iker-barriocanal: Not sure if I understand the question. This test validates that missing global filters keep metric extraction enabled, and this line removes filters from the global config.

relay-dynamic-config/src/global.rs
iker-barriocanal: We can rename the variable and set a serde `rename`, but this seemed too much to me. What do you think?

The reason to keep `filters` in the protocol is because "inbound" is a `sentry` product term and Relay doesn't have to know about it. If new non-inbound filters are introduced, Relay may run them at this point.

Cargo.toml
iker-barriocanal: Happy to do that in a follow-up if the new version introduces some nice features. I don't want to increase the scope of this PR as it's already big and complex enough.

relay-dynamic-config/src/global.rs
Dav1dde: For my taste it's a bit too generic on the protocol level, but if there is a good reason to do it 👍

relay-filter/src/config.rs
iker-barriocanal: It's used in `generic.rs`, so it's either `pub(crate)` or `pub`. I'm going to leave it as `pub(crate)` unless there's a strong opinion on a different suggestion. We can revisit the module structure in the future.

relay-dynamic-config/src/global.rs
iker-barriocanal: In project configs, the field is `filterSettings`. I'm ok with updating the protocol to `inboundFilters`, but that also generates a consistency mismatch. What do you think?

relay-filter/src/config.rs
iker-barriocanal: Is the best place the `DynamicFiltersConfig` struct or the `into_iter` function of the `IntoIterator` implementation?

relay-filter/src/config.rs
iker-barriocanal: Good suggestion! I'm doing this refactor but keeping the `is_version_supported`, as removing this indirection decreases the readability and want to keep the docstrings.

relay-filter/src/config.rs
iker-barriocanal: This check outside the function allows skipping project filters, useful when the iterator has already exhausted them. Moving the check to the function would require returning an enum to know whether the evaluated filter should be yielded, skipped or all filters are exhausted.

(same response as in the comment below)

relay-filter/src/config.rs
iker-barriocanal: This check outside the function allows skipping project filters, useful when the iterator has already exhausted them. Moving the check to the function would require returning an enum to know whether the evaluated filter should be yielded, skipped or all filters are exhausted.

(same response as in the comment above)

relay-filter/src/config.rs
Dav1dde: Can you just move it to the generic module? Kind of fits there.

relay-dynamic-config/src/global.rs
Dav1dde: I'd prefer `inboundFilter` but maybe worth just bringing up with the team tomorrow.

tests/integration/test_filters.py
iker-barriocanal: I'll cover these cases with unit tests, as they are validating the iterator's implementation logic.

Edit: these edge cases should be covered with existing and newly added unit tests in [45e2f94](https://github.com/getsentry/relay/pull/3161/commits/45e2f94c76d625acc63521cd0f31e3308ffa365f).

Dav1dde: Fair enough, although that should also be nicely doable with pytest and parameterized tests.

relay-server/src/services/processor.rs
iker-barriocanal: Generally, replacing booleans with enums in function arguments increases readability. In this case, it's a supported/unsupported decision as the function return type, so I think a boolean is a better choice. What do you think?

relay-server/src/services/processor.rs
iker-barriocanal: `relay-dynamic-config` depends on `relay-filter` so passing the full configs isn't an option and we're limited to just passing the versions, unfortunately.

relay-filter/src/config.rs
iker-barriocanal: This seems to be an unstable feature, see https://github.com/rust-lang/rust/issues/53667. rust-analyzer also warns about the unnecessary `else` block.

relay-filter/src/config.rs
iker-barriocanal: The suggestion results in a different behavior: the iterator skips `None` return values and goes to another loop, instead of returning `None`. On the other hand, I think that a similar block for project and global filters makes the block easier to understand.

relay-filter/src/config.rs
iker-barriocanal: I'm not sure if this increases readability but applied the change in [45e2f94](https://github.com/getsentry/relay/pull/3161/commits/45e2f94c76d625acc63521cd0f31e3308ffa365f).

relay-filter/src/config.rs
iker-barriocanal: TIL fuse iterator, nice!

iker-barriocanal: Done in [45e2f94](https://github.com/getsentry/relay/pull/3161/commits/45e2f94c76d625acc63521cd0f31e3308ffa365f).

relay-filter/src/config.rs
iker-barriocanal: As comparing two configs is useful for tests, I added the full implementation just for `cfg=test` in [45e2f94](https://github.com/getsentry/relay/pull/3161/commits/45e2f94c76d625acc63521cd0f31e3308ffa365f).

relay-filter/src/config.rs
iker-barriocanal: Done in [45e2f94](https://github.com/getsentry/relay/pull/3161/commits/45e2f94c76d625acc63521cd0f31e3308ffa365f) (including a new doctest).

tests/integration/test_filters.py
iker-barriocanal: It's possible, but the unit tests are more scoped, easier to spot, faster to run, and less surface area to be flaky 😄

relay-dynamic-config/src/global.rs
iker-barriocanal: Agree, removed this method in [b57c122](https://github.com/getsentry/relay/pull/3161/commits/b57c12245eaaafd914ab1719e97e42dfd65356b8).

iker-barriocanal: Removed this method in [b57c122](https://github.com/getsentry/relay/pull/3161/commits/b57c12245eaaafd914ab1719e97e42dfd65356b8).

relay-filter/src/config.rs
iker-barriocanal: Actually yes, moved it in [b57c122](https://github.com/getsentry/relay/pull/3161/commits/b57c12245eaaafd914ab1719e97e42dfd65356b8).

relay-filter/src/config.rs
jjbayer: You wouldn't even need to track any indexes, the code above would already give you the body for the `next()` implementation.

Since there's a lot of comments on this PR, what do you think about a live discussion / some pair programming tomorrow?

tests/integration/test_metrics.py
Dav1dde: But you never added filters in the first place, I am probably missing something?

relay-server/src/services/processor.rs
Dav1dde: Isnt `event::filter` in `relay-server`?

relay-server/src/services/processor.rs
Dav1dde: I like the enum 👍

tests/integration/test_metrics.py
iker-barriocanal: Yep, I added them to the fixture in `mini_sentry`. All tests using that fixture now get the following config:

```txt
"filters": {"version": 1, "filters": []},
```

relay-server/src/services/processor.rs
iker-barriocanal: Oh right hehe yeah I agree, and applied in [8384897](https://github.com/getsentry/relay/pull/3161/commits/83848978c571ce5ad9a82cb47064a3bebc318a2a).

relay-dynamic-config/src/global.rs
iker-barriocanal: For consistency with the rest of Relay, we're sticking with `filters`.

relay-filter/src/generic.rs
```diff
@@ -27,44 +26,178 @@ fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
...
+            global,
+            global_index: 0,
+        }
+    }
+}
+
+impl<'a> Iterator for DynamicGenericFiltersConfigIter<'a> {
+    type Item = MergedFilter<'a>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
iker-barriocanal, line 117: Creating two iterators before `impl Iterator` makes the iterator implementation easier to understand, but adds unneeded complexity on typing these iterators in the struct: exact type of the iterator including the closure of the filtering that must be known at compile time and creating yet another struct is not worth it.

relay-filter/src/generic.rs
```diff
@@ -27,44 +26,178 @@ fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
...
+        if let Some((id, filter)) = self.project.get_index(self.project_index) {
+            self.project_index += 1;
+            let merged = merge_filters(filter, self.global.and_then(|gf| gf.get(id)));
```
iker-barriocanal, line 120: The biggest simplification from the previous revision is to not check `is_enabled` or `condition`. This increases the complexity of the iterator implementation enough to make it hard to understand, and dealing with these outside the iterator is not too hard. Thus, I'm not adding these checks now, and we can consider extending the abstraction in the future to deal with these, if necessary.

relay-filter/src/config.rs
iker-barriocanal: Followed the simplified approach in [a6db146](https://github.com/getsentry/relay/pull/3161/commits/a6db146fd2cbf1c653f989bbb05790e425e8023d).

relay-filter/src/config.rs
iker-barriocanal: Followed a different and simpler approach in [a6db146](https://github.com/getsentry/relay/pull/3161/commits/a6db146fd2cbf1c653f989bbb05790e425e8023d).

relay-filter/src/config.rs
iker-barriocanal: Followed a different and simpler approach in [a6db146](https://github.com/getsentry/relay/pull/3161/commits/a6db146fd2cbf1c653f989bbb05790e425e8023d).

relay-server/src/services/processor.rs
iker-barriocanal: I've looked into it, and still think the use-case is simple enough to deal with a boolean. Let me know if you're strongly opinionated on this and feel free to block the PR in that case.

relay-filter/src/config.rs
iker-barriocanal: In the new approach in [a6db146](https://github.com/getsentry/relay/pull/3161/commits/a6db146fd2cbf1c653f989bbb05790e425e8023d), I added the relevant documentation in the function generating the iterator.

relay-filter/src/generic.rs
```diff
@@ -27,44 +26,178 @@ fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
...
+            return Some(merged);
+        }
+
+        loop {
+            let Some((id, filter)) = self
+                .global
+                .and_then(|filters| filters.get_index(self.global_index))
```
Dav1dde:
```suggestion
            let (id, filter) = self.global?.get_index(self.global_index)?;
```

relay-dynamic-config/src/global.rs
```diff
@@ -23,6 +24,15 @@ pub struct GlobalConfig {
...
     pub quotas: Vec<Quota>,
+    /// Configuration for global inbound filters.
+    ///
+    /// These filters are merged with generic filters in project configs before
+    /// applying.
+    #[serde(
+        deserialize_with = "default_on_error",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub filters: Option<GenericFiltersConfig>,
```
Dav1dde: Wasn't there an `ErrorBoundary` before?

relay-filter/src/config.rs
```diff
@@ -248,13 +251,177 @@ impl GenericFilterConfig {
     }
 }
 
+#[cfg(test)]
+impl PartialEq for GenericFilterConfig {
```
Dav1dde: Should be able to just derive the `PartialEq`

relay-filter/src/config.rs
```diff
@@ -248,13 +251,177 @@ impl GenericFilterConfig {
...
+    /// Map of generic filters, sorted by the order in the payload from upstream.
+    ///
+    /// The map contains unique filters, meaning there are no two filters with
+    /// the same id. See struct docs for more details.
+    #[serde(
+        default,
+        skip_serializing_if = "IndexMap::is_empty",
+        with = "generic_filters_custom_serialization"
+    )]
+    pub filters: IndexMap<String, GenericFilterConfig>,
```
Dav1dde: I wonder if we should make this a newtype instead and hide the `IndexMap` which should make it impossible to construct an invalid index map. 

Can also implement Serialize/Deserialize directly on the newtype (cannot forget to add the annotation)

Would also simplify tests, could just be constructed from a list/vector of filter configs again.

relay-filter/src/generic.rs
```diff
@@ -27,44 +26,178 @@ fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
...
+        is_enabled: primary.is_enabled,
+        condition: primary
+            .condition
+            .as_ref()
+            .or(secondary.and_then(|filter| filter.condition.as_ref())),
+    }
+}
+
+#[derive(Debug, Default, PartialEq)]
+struct MergedFilter<'a> {
```
Dav1dde: nit: I think `GenericFilterConfigRef` is a better name

relay-dynamic-config/src/global.rs
iker-barriocanal: Yes, but it's no longer needed.

We only extract metrics in the following cases:
- In processing Relays. We don't want to limit that further.
- When events are going to be dropped due to dynamic sampling. Since there's a sampling decision outside generic filters, we don't want to limit this either.
In other cases, the events are forwarded upstream. This means that `Option` is good enough.

Cargo.toml
jjbayer: How would using the latest version increase the scope of this PR? It's a new dependency anyway?

jjbayer: nit: 
```suggestion
indexmap = "2.0.0"
itertools = "0.10.5"
```

relay-dynamic-config/src/global.rs
iker-barriocanal: Even if not needed, restored it in [d7790ea](https://github.com/getsentry/relay/pull/3161/commits/d7790eab2cb3ba91d9269a1bddb685314db7b632) to make the difference in the future.

relay-filter/src/config.rs
iker-barriocanal: Done in [d7790ea](https://github.com/getsentry/relay/pull/3161/commits/d7790eab2cb3ba91d9269a1bddb685314db7b632).

relay-filter/src/config.rs
```diff
@@ -416,34 +409,60 @@ pub struct GenericFiltersConfig {
...
+pub struct GenericFiltersMap(IndexMap<String, GenericFilterConfig>);
+
+impl GenericFiltersMap {
+    /// Returns an empty map.
+    pub fn new() -> Self {
+        GenericFiltersMap(IndexMap::new())
+    }
+
+    /// Returns whether the map is empty.
+    pub fn is_empty(&self) -> bool {
```
Dav1dde, line 434: With the deref this shouldnt be necessary

relay-filter/src/config.rs
```diff
@@ -416,34 +409,60 @@ pub struct GenericFiltersConfig {
...
+        self.0.is_empty()
+    }
+}
+
+impl From<Vec<GenericFilterConfig>> for GenericFiltersMap {
+    fn from(filters: Vec<GenericFilterConfig>) -> Self {
+        let mut map = IndexMap::with_capacity(filters.len());
+        for filter in filters {
+            map.insert(filter.id.clone(), filter.clone());
```
Dav1dde:
```
map.insert(filter.id.clone(), filter);
```

Dav1dde: You need to check if there is not already an item in the map, same thing you do in the Deserialize, order is important.

relay-filter/src/config.rs
Dav1dde: Needed for serde annotation

relay-dynamic-config/src/global.rs
```diff
@@ -46,6 +55,21 @@ impl GlobalConfig {
             Ok(None)
         }
     }
+
+    /// Returns the generic inbound filters.
+    pub fn filters(&self) -> Option<&GenericFiltersConfig> {
+        match &self.filters {
+            ErrorBoundary::Err(_) => None,
```
jjbayer, line 62: In this case, we filter nothing, right?

Not a strong opinion on this, because we can always bump the metrics extraction version, but Isn't the following scenario still relevant:

1. We make a breaking change to the filter config.
2. Outdated external relays stop filtering, and extract metrics from everything that was previously filtered out.

I'm also on board with just deciding that we'll never make a breaking change, but in that case, why do we need the error boundary?

relay-filter/src/config.rs
iker-barriocanal: It's used for skip serialization in serde.

relay-filter/src/config.rs
```diff
@@ -249,25 +253,258 @@ impl GenericFilterConfig {
...
+    /// the same id. See struct docs for more details.
+    #[serde(default, skip_serializing_if = "GenericFiltersMap::is_empty")]
+    pub filters: GenericFiltersMap,
 }
 
 impl GenericFiltersConfig {
     /// Returns true if the filters are not declared.
     pub fn is_empty(&self) -> bool {
-        self.filters.is_empty()
+        self.filters.0.is_empty()
```
Dav1dde:
```suggestion
        self.filters.is_empty()
```

relay-filter/src/generic.rs
```diff
@@ -27,32 +25,153 @@ fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
...
+        if let Some((id, filter)) = self.project.get_index(self.project_index) {
+            self.project_index += 1;
+            let merged = merge_filters(filter, self.global.and_then(|gf| gf.get(id)));
+            return Some(merged);
+        }
+
+        loop {
+            let Some((id, filter)) = self.global?.get_index(self.global_index) else {
+                return None;
+            };
```
Dav1dde:
```suggestion
            let (id, filter) = self.global?.get_index(self.global_index)?; 
```

relay-server/src/services/processor/event.rs
```diff
@@ -273,6 +274,7 @@ pub fn finalize<G: EventProcessing>(
 
 pub fn filter<G: EventProcessing>(
     state: &mut ProcessEnvelopeState<G>,
+    global_config: Arc<GlobalConfig>,
```
Dav1dde:
```suggestion
    global_config: &GlobalConfig,
```

Dav1dde (approved):
An integration test which tests an invalid global config would be nice to make sure DS is skipped even with a valid project config.

Cargo.toml
iker-barriocanal: 2.0.0 -> 2.2.3 should have no breaking changes, but there could be e.g. uncaught panics. If this is the case and we identify this PR as the root cause, narrowing the cause further to the dependency is more complicated. Separating it in a different PR is easier to spot, and I don't mind the extra bit of work.

relay-dynamic-config/src/global.rs
iker-barriocanal: Should be addressed in [bfb8657](https://github.com/getsentry/relay/pull/3161/commits/bfb8657364cfc5fa9739b625bf72f78aec987d20).

relay-server/src/services/processor.rs
```diff
@@ -1274,23 +1274,41 @@ impl EnvelopeProcessorService {
...
-        event::filter(state)?;
-        dynamic_sampling::run(state, &self.inner.config);
+        event::filter(state, &self.inner.global_config.current())?;
+        // Don't extract metrics if relay can't apply generic inbound filters.
+        // An inbound filter applied in another up-to-date relay in chain may
+        // need to drop the event, and there should not be metrics from dropped
+        // events.
+        // In processing relays, always extract metrics to avoid losing them.
+        let supported_generic_filters = self.inner.global_config.current().filters.is_ok()
+            && relay_filter::are_generic_filters_supported(
```
iker-barriocanal, line 1284: Unfortunately, passing the global config here creates a circular dependency with crates.

relay-dynamic-config/src/global.rs
```diff
@@ -46,6 +55,21 @@ impl GlobalConfig {
...
+            ErrorBoundary::Ok(f) => Some(f),
+        }
+    }
+}
+
+fn skip_generic_filters(filters_config: &ErrorBoundary<GenericFiltersConfig>) -> bool {
```
jjbayer: nit

```suggestion
fn is_err_or_empty(filters_config: &ErrorBoundary<GenericFiltersConfig>) -> bool {
```

relay-filter/src/config.rs
```diff
@@ -249,12 +253,164 @@ impl GenericFilterConfig {
 }
 
 /// Configuration for generic filters.
+///
+/// # Deserialization
+///
+/// `filters` is expected to be a [`Vec<GenericfilterConfig>`].
```
jjbayer: Should this say something like `filters` is expected to be a sequence of `GenericfilterConfig>`  in the serialized format?

relay-filter/src/config.rs
```diff
@@ -264,10 +420,91 @@ impl GenericFiltersConfig {
...
+        GenericFiltersMap(IndexMap::new())
+    }
+
+    /// Returns whether the map is empty.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+impl From<Vec<GenericFilterConfig>> for GenericFiltersMap {
```
jjbayer, line 439:
```suggestion
#[cfg(test)]
impl From<Vec<GenericFilterConfig>> for GenericFiltersMap {
```

relay-filter/src/generic.rs
```diff
@@ -27,32 +33,151 @@ fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
...
+/// configs.
+/// - No duplicates: once a filter is evaluated (yielded or skipped), no filter
+/// with the same id is evaluated again.
+/// - If a filter with the same id exists in projects and global configs, both
+/// are merged and the filter is yielded. Values from the filter in the project
+/// config are prioritized.
+fn merge_generic_filters<'a>(
+    project: &'a GenericFiltersConfig,
+    global: Option<&'a GenericFiltersConfig>,
+    #[cfg(test)] version: u16,
```
jjbayer: nit

```suggestion
    #[cfg(test)] max_supported_version: u16,
```

relay-filter/src/generic.rs
```diff
@@ -4,17 +4,23 @@
...
+
+use crate::{FilterStatKey, GenericFilterConfig, GenericFiltersConfig, GenericFiltersMap};
 use relay_event_schema::protocol::Event;
 use relay_protocol::RuleCondition;
 
 /// Maximum supported version of the generic filters schema.
 ///
 /// If the version in the project config is higher, no generic filters are applied.
-pub const VERSION: u16 = 1;
+const VERSION: u16 = 1;
```
jjbayer: nit

```suggestion
const MAX_SUPPORTED_VERSION: u16 = 1;
```

relay-filter/src/generic.rs
```diff
@@ -27,32 +33,151 @@ fn matches(event: &Event, condition: Option<&RuleCondition>) -> bool {
...
+///
+/// It's assumed both filters share the same id. The returned filter will have
+/// the primary filter's ID.
+fn merge_filters<'a>(
+    primary: &'a GenericFilterConfig,
+    secondary: Option<&'a GenericFilterConfig>,
+) -> GenericFilterConfigRef<'a> {
+    GenericFilterConfigRef {
+        id: primary.id.as_str(),
+        is_enabled: primary.is_enabled,
```
jjbayer, line 146: This means we cannot create "opt-out" filters that are enabled by default via the the global config. Not a blocker  but IMO the better solution would be to make `is_enabled` an `Option<bool>` in `ProjectFilterConfig` and use the global value when it it `None`.

relay-server/src/services/processor.rs
```diff
@@ -1274,23 +1274,41 @@ impl EnvelopeProcessorService {
...
+                    .global_config
+                    .current()
+                    .filters()
+                    .map(|f| f.version),
+                state.project_state.config.filter_settings.generic.version,
+            );
 
-        // Remember sampling decision, before it is reset in `dynamic_sampling::sample_envelope_items`.
-        let sampling_should_drop = state.sampling_result.should_drop();
+        let mut sampling_should_drop = false;
```
jjbayer, line 1293: Why is this declared outside of the `if`?

tests/integration/test_metrics.py
```diff
@@ -1880,3 +1878,172 @@ def test_metric_bucket_encoding_dynamic_global_config_option(
...
+    "global_generic_filters",
+    [
+        # Config is broken
+        {
+            "version": "Halp! I'm broken!",
+            "filters": [],
+        },
+        # Config is valid, but filters aren't supported
+        {
+            "version": "65535",
```
jjbayer: Don't we expect an integer here?

```suggestion
            "version": 65535,
```

tests/integration/test_metrics.py
```diff
@@ -1880,3 +1878,172 @@ def test_metric_bucket_encoding_dynamic_global_config_option(
...
+
+    transaction = generate_transaction_item()
+    relay.send_transaction(project_id, transaction)
+
+    if is_processing_relay:
+        tx, _ = tx_consumer.get_event()
+        assert tx is not None
+    else:
+        assert mini_sentry.captured_events.get() is not None
+    metrics_consumer.assert_empty()
```
jjbayer: I thought processing relays always extract metrics?

relay-filter/src/generic.rs
iker-barriocanal: If a filter is enabled by default in the global config, it's still possible to override it from the project config:
- Project config filters are evaluated first, along with the filters in the global configs, and have priority in this merging step.
- When going through the global config filters, those filters that exist in the project filters aren't evaluated again.
Let me know if there's a way to make this more clear.

relay-server/src/services/processor.rs
iker-barriocanal: - The value is used in the metric timer below, and only overwritten inside the `if`.
- If the `if` doesn't run, the event is kept and forwarded to the next relay, thus the default value `false`.

relay-filter/src/generic.rs
jjbayer: You're right, we don't need `is_enabled` to be `Option<bool>` because we can just omit the project-side config if we want the global default to work, and add a project-side config if we want to override. Sorry for the confusion.

relay-filter/src/config.rs
iker-barriocanal: It's also used in the doctest.

tests/integration/test_metrics.py
iker-barriocanal: They do! The test was inaccurate and there's something wrong with `assert_empty`. I adapted the test in [28ed597](https://github.com/getsentry/relay/pull/3161/commits/28ed597458dcf483fb76e8baa287b00b7a5c682b) to reflect the reality better.

jjbayer (approved):

relay-server/src/services/processor.rs
```diff
@@ -1274,23 +1274,41 @@ impl EnvelopeProcessorService {
...
 
-        // We avoid extracting metrics if we are not sampling the event while in non-processing
-        // relays, in order to synchronize rate limits on indexed and processed transactions.
-        if self.inner.config.processing_enabled() || sampling_should_drop {
-            self.extract_metrics(state)?;
-        }
+        if self.inner.config.processing_enabled() || supported_generic_filters {
```
Dav1dde, line 1295: Shouldn't this be:

```suggestion
        if self.inner.config.processing_enabled() && supported_generic_filters {
```

relay-server/src/services/processor.rs
```diff
@@ -1274,23 +1274,41 @@ impl EnvelopeProcessorService {
...
         dynamic_sampling::normalize(state);
-        event::filter(state)?;
-        dynamic_sampling::run(state, &self.inner.config);
+        event::filter(state, &self.inner.global_config.current())?;
+        // Don't extract metrics if relay can't apply generic inbound filters.
+        // An inbound filter applied in another up-to-date relay in chain may
+        // need to drop the event, and there should not be metrics from dropped
+        // events.
+        // In processing relays, always extract metrics to avoid losing them.
+        let supported_generic_filters = self.inner.global_config.current().filters.is_ok()
```
Dav1dde, line 1283: I really think `event::filter` should return this, this is filter related logic and just hard to read here.

or hide it behind some function in the `event` module, then you can also pass the global config.

relay-server/src/services/processor.rs
Dav1dde, line 1293: This doesn't need to be here and also not be mutable.

Just always use `sampling_should_drop = state.sampling_result.should_drop();`

relay-server/src/services/processor.rs
iker-barriocanal: Nope: processing relays should always run dynamic sampling, even if generic filters are unsupported.

Dav1dde: Can Managed Relays also have processing enabled?

relay-server/src/services/processor.rs
iker-barriocanal: `dynamic_sampling::sample_envelope_items` resets the value, see comment below. This change is not coming from this PR.

Dav1dde: Ah nvm sorry I misread the indentation

relay-server/src/services/processor.rs
Dav1dde: Nvm that wouldn't make sense.

Dav1dde (approved):
As noted before I think we can move the `supported_generic_filters` to a better place. Other than that :shipit:

relay-server/src/services/processor.rs
iker-barriocanal: Makes sense. I'll address this in a follow-up so that this PR can already land.

relay-server/src/services/processor.rs
jjbayer: They shouldn't, but we've seen customers do it. IMO `Processing` should be a mode, just like `Managed`.

PR #3162, opened 2024-02-26
fix(normalization): Accept transactions with unfinished spans

### LINKED ISSUE #3222: Sentry Invalidating Transactions for unfinished Spans
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.2.2

### Proguard

Enabled

### Version

7.3.0

### Steps to Reproduce

Start a transaction and hold a reference to it.  I've seen both scoped and non-scoped transactions have the effect, but I think scoped transaction are more likely.  With that transaction, start a child span, but don't finish it.   Finish the transaction.

### Expected Result

I would expect the child spans to be stopped when the transaction stops.  Their span status should probably be SpanStatus.ABORTED.

### Actual Result

If we query our project metrics from the statsV2 endpoint, there are a very high number of invalid_transaction. 

In 24 hours, here are the results of the [Stats V2 endpoint](https://docs.sentry.io/api/organizations/retrieve-event-counts-for-an-organization-v2/)

```
Outcome           Reason
accepted          none                   :  3356853
client_discard    cache_overflow         :  184
client_discard    sample_rate            :  704095
client_discard    network_error          :  50311
client_discard    queue_overflow         :  23
invalid           invalid_transaction    :  1893851
```

If the following code is used as the tracing callback, a coworker who experienced lost transactions was able to see them.

```kotlin
beforeSendTransaction = SentryOptions.BeforeSendTransactionCallback { transaction, _ ->
        
     val finishedVsUnfinishedSpans = transaction.spans.partition { it.isFinished }

     if (finishedVsUnfinishedSpans.second.isNotEmpty()) {
        with(transaction) {
              spans.clear()
              spans.addAll(finishedVsUnfinishedSpans.first)
        }
     }

     transaction
 }
```

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2113) by [Unito](https://www.unito.io)

iker-barriocanal:
Fixes https://github.com/getsentry/sentry-java/issues/3222.

This PR finishes all spans before validating them. This used to be the behavior before a regression was introduced [here](https://github.com/getsentry/relay/commit/90fa749f1c4af857fb0f0b78e517d0f9fed4b32a).

Additionally, event timestamp validation now runs before transaction validation. Event validation includes some timestamp normalization (like clock drift correction) that updates the event's timestamps. Normalizing a transaction's timestamps after updating a span's timestamps may result in spans being out of the transaction's range.

## Future work

- [ ] Consider having a single validation step. A single step removes the ambiguity and points of failures of running the two steps in a different order. Originally, the two steps were introduced to minimize validation on mutable references, but after this PR this is no longer the case.

### DIFF
--- a/relay-event-normalization/src/validation.rs
+++ b/relay-event-normalization/src/validation.rs
@@ -833,6 +833,6 @@ mod tests {
         let spans = &event.spans;
         let span = get_value!(spans[0]!);
 
-        assert_eq!(span.timestamp, event.timestamp);
+        assert_eq!(span.timestamp.value(), event.timestamp.value());
     }
 }
--- a/relay-event-normalization/src/validation.rs
+++ b/relay-event-normalization/src/validation.rs
@@ -32,10 +32,10 @@ pub struct TransactionValidationConfig {
 /// Note: this function does not validate a transaction's timestamp values are
 /// up-to-date, [`validate_event_timestamps`] should be used for that.
 pub fn validate_transaction(
-    event: &Annotated<Event>,
+    event: &mut Annotated<Event>,
     config: &TransactionValidationConfig,
 ) -> ProcessingResult {
-    let Annotated(Some(ref event), ref _meta) = event else {
+    let Annotated(Some(ref mut event), ref _meta) = event else {
         return Ok(());
     };
     if event.ty.value() != Some(&EventType::Transaction) {
@@ -46,6 +46,7 @@ pub fn validate_transaction(
     validate_trace_context(event)?;
     // There are no timestamp range requirements for span timestamps.
     // Transaction will be rejected only if either end or start timestamp is missing.
+    end_all_spans(event);
     validate_spans(event, None)?;
 
     Ok(())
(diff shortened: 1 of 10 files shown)


relay-event-normalization/src/validation.rs
```diff
@@ -229,8 +245,9 @@ pub struct EventValidationConfig {
 /// checking for validity, like clock drift correction. This normalization
 /// depends on the given configuration.
 ///
-/// Validation is checked individually on timestamps. Use
-/// [`validate_transaction`] to perform additional transaction-specific checks.
+/// Validation is checked individually on timestamps. For transaction events,
+/// [`validate_transaction`] should run after this method to perform additional
+/// transaction-specific checks.
```
iker-barriocanal, line 250: With the future action item, this docstring should be irrelevant.

olksdr (approved):

jjbayer (approved):

PR #3165, opened 2024-02-26
feat(server): Drop transactions in an only-spans world

jjbayer:
To test what our Performance product looks like without transactions (spans only), add the ability to drop transactions in Relay. The feature flag will only be enabled for a single experimental project for the time being.

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -1297,11 +1297,11 @@ impl EnvelopeProcessorService {
 
         if state.has_event() {
             event::scrub(state)?;
-            event::serialize(state)?;
             if_processing!(self.inner.config, {
                 span::extract_from_event(state);
-                span::maybe_remove_event(state);
+                span::maybe_discard_transaction(state);
             });
+            event::serialize(state)?;
         }
 
         attachment::scrub(state);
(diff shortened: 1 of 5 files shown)


relay-server/src/services/processor.rs
```diff
@@ -1300,10 +1300,11 @@ impl EnvelopeProcessorService {
 
                 if state.has_event() {
                     event::scrub(state)?;
-                    event::serialize(state)?;
                     if_processing!(self.inner.config, {
                         span::extract_from_event(state);
+                        span::maybe_discard_transaction(state);
                     });
+                    event::serialize(state)?;
```
jjbayer, line 1307: Need to drop before we serialize the event back into the envelope.

tests/integration/fixtures/processing.py
```diff
@@ -180,7 +180,9 @@ def assert_empty(self, timeout=None):
         self.test_producer.produce(self.topic_name, message)
         self.test_producer.flush(timeout=5)
 
+        print(timeout)
```
jjbayer:
```suggestion
```

tests/integration/fixtures/processing.py
```diff
@@ -180,7 +180,9 @@ def assert_empty(self, timeout=None):
...
         rv = self.poll(timeout=timeout)
+        print(rv)
```
jjbayer:
```suggestion
```

Dav1dde (approved):

PR #3166, opened 2024-02-26
feat(spans): Copies exclusive_time from span data into a top level attribute

edwardgou-sentry:
Takes `exclusive_time` from span data, and copies it as a top level attribute on the span.

For context, this is required because we enforce the top level `exclusive_time` attribute to be not empty when validating spans. For `otel_span`, we already do this on the `exclusive_time_ns` within the span data, but we do not do this on `span` type items (hence this pr).

#skip-changelog

### DIFF
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -559,6 +559,7 @@ mod tests {
             http_response_status_code: ~,
             thread_name: ~,
             ui_component_name: ~,
+            exclusive_time: ~,
             url_scheme: ~,
             other: {
                 "bar": String(
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -272,6 +272,10 @@ pub struct SpanData {
     #[metastructure(field = "url.scheme")]
     pub url_scheme: Annotated<Value>,
 
+    /// exclusive time
+    #[metastructure(field = "exclusive_time")]
+    pub exclusive_time: Annotated<Value>,
+
     /// Other fields in `span.data`.
     #[metastructure(additional_properties, pii = "true", retain = "true")]
     other: Object<Value>,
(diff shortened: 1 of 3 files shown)


edwardgou-sentry:
The reason why we are sending `exclusive_time` in `data` is because having `exclusive_time` as a toplevel attribute in the payload is less OTel compliant

jjbayer (changes requested):
> The reason why we are sending exclusive_time in data is because having exclusive_time as a toplevel attribute in the payload is less OTel compliant

@edwardgou-sentry I don't fully get this reasoning. Sentry spans have a different format than OTel spans. They are not OTel compatible, so why could an SDK not send the top-level field?

That said, if this is fix to make span ingestion work for existing SDKs that send `exclusive_time` in the data field, we can merge it.

edwardgou-sentry:
> > The reason why we are sending exclusive_time in data is because having exclusive_time as a toplevel attribute in the payload is less OTel compliant
> 
> @edwardgou-sentry I don't fully get this reasoning. Sentry spans have a different format than OTel spans. They are not OTel compatible, so why could an SDK not send the top-level field?
> 
> That said, if this is fix to make span ingestion work for existing SDKs that send `exclusive_time` in the data field, we can merge it.

Hey @jjbayer , there's some more context in this pr:
https://github.com/getsentry/sentry-javascript/pull/10704#discussion_r1496693949

We could send `exclusive_time` as a top level field, but we're preferring to send it in `data`. I think part of the reason is because in the sdks we don't separate types between OTel spans and Sentry spans?

I think we just need to decide whether we allow `exclusive_time` as a top level field in sentry spans from the sdk, or make the change in relay (ie this pr)

cc @AbhiPrasad

edwardgou-sentry:
Closing this PR since we are agreeing to pass `exclusive_time` as a top level field in envelopes from SDKs, so this normalization is no longer needed

PR #3167, opened 2024-02-27
chore(spooler): Change default connection pool to 1 connection

olksdr:
This sets the sqlite connections pool to use max only 1 connection with DB file.

Since we're processing only 1 message at a time right now, we do not really need to have many connections open and maintained at the same time. Which also will allow us to debug any weird looking issues behaviour if that can happen.

related to https://github.com/getsentry/ops/pull/9509

#skip-changelog

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -823,12 +823,12 @@ fn spool_envelopes_max_disk_size() -> ByteSize {
 
 /// Default for min connections to keep open in the pool.
 fn spool_envelopes_min_connections() -> u32 {
-    10
+    1
 }
 
 /// Default for max connections to keep open in the pool.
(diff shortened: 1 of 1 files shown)


jjbayer (approved):
I wonder if instead of .`await`ing every db op we should make large reads and writes exclusive, but still allow small requests like health checks to run concurrently. If we ever do that we would have to increase the number of connections again.

olksdr:
> I wonder if instead of .`await`ing every db op we should make large reads and writes exclusive, but still allow small requests like health checks to run concurrently. If we ever do that we would have to increase the number of connections again.

Yeah, it's a very good comment! 
And I think the config change should then go together if we end up doing the code changes.

iker-barriocanal (approved):

PR #3174, opened 2024-02-28
docs(metrics): allow empty space character in metrics tag values

viglia:
catching up with https://github.com/getsentry/develop/pull/1171

#skip-changelog


Dav1dde (commented):
Does Relay actually accept these values?

Also since this is docs, I don't think we need a changelog entry. Can just put `#skip-changelog` in the PR description.

viglia:
> Does Relay actually accept these values?
> 
> Also since this is docs, I don't think we need a changelog entry. Can just put `#skip-changelog` in the PR description.

👍🏻 removed the changelog entry

yes, I quickly checked [here](https://github.com/getsentry/relay/blob/master/relay-metrics/src/protocol.rs#L44-L46) and space character is not stripped out. Plus `php` is already sending tags that way.

Dav1dde (approved):

PR #3177, opened 2024-02-28
feat(performance): Extract replay and users for INP spans

edwardgou-sentry:
Extracts `replay_id` and `user` from span data as tags on indexed INP spans. 
~Extracts `profile_id` from span data as a column on indexed INP spans.~

#skip-changelog


relay-event-schema/src/protocol/span.rs
```diff
@@ -277,6 +277,18 @@ pub struct SpanData {
     #[metastructure(field = "url.scheme")]
     pub url_scheme: Annotated<Value>,
 
+    /// Profile ID
+    #[metastructure(field = "profile_id")]
+    pub profile_id: Annotated<Value>,
```
jjbayer: `profile_id` exists as a top-level field on the span, and it automatically gets picked up by kafka consumer. Should we use it?

edwardgou-sentry: Hey Joris, good point, I dropped the changes to `profile_id` in span data.
Should be able to just pass `profile_id` at the top level, will update the sdks.

jjbayer (approved):

PR #3178, opened 2024-02-28
feat(spans): Add SDK information to spans

phacops:
In order to make our some of our reporting dashboards, we'd need to know which SDKs generated the spans. This PR aims to add those 2 fields as Sentry tags. We'll then extract the values to store in BigQuery. They will be ingested by the Snuba consumer since they're part of Sentry tags now.

At the same time, in order to avoid a costly join on project ID, we're adding the organization ID to the Kafka payload.  This field will not be used by the Snuba consumer, only SBC for data ingestion into BigQuery.

We're also adding `platform`, a field necessary for performance issue detection later in the process. It will also be ingested as part of Sentry tags in ClickHouse.


iker-barriocanal (approved):

relay-server/src/services/store.rs
```diff
@@ -1521,6 +1522,8 @@ struct SpanKafkaMessage<'a> {
 
     #[serde(borrow, default, skip_serializing_if = "Option::is_none")]
     measurements: Option<BTreeMap<Cow<'a, str>, Option<SpanMeasurement>>>,
+    #[serde(default)]
+    organization_id: u64,
```
jjbayer, line 1526: The consumer allows unknown additional properties, right?

jjbayer (approved):

relay-server/src/services/store.rs
phacops: Yes.

PR #3182, opened 2024-02-29
feat(cli): Add command to clear the spooled on-disk data

olksdr:
Running `relay spool clear` command Relay will try to remove the spooled on disk data either using provided config file or the explicitly provided path, which will take priority. 

By default the `clear` operation will ask for confirmation. If the `force` argument is provided the command will execute without any interactive prompts.

fixes https://github.com/getsentry/team-ingest/issues/266

#skip-changelog


relay-server/src/services/spooler/spool_utils.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use crate::service::create_runtime;
+use crate::services::spooler::{sql, BufferError};
+
+/// Truncates the spool file deleting all the persisted on-disk data.
+pub fn truncate(path: PathBuf) -> Result<(), BufferError> {
+    let options = SqliteConnectOptions::new()
+        .filename(path)
+        .journal_mode(SqliteJournalMode::Wal)
+        .create_if_missing(true);
```
Dav1dde: Shouldn't we rather error if Relay can't find the spool?

relay-server/src/services/spooler/sql.rs
```diff
@@ -155,3 +155,8 @@ pub async fn do_insert(
 
     Ok(count)
 }
+
+/// Creates DELETE statement which truncates the entire `envelopes` table.
+pub fn truncate<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
+    sqlx::query("DELETE FROM envelopes;")
```
Dav1dde, line 164: Maybe should add a quick comment that sqlite does have a truncate optimization on where (I had to look up why we're not using `TRUNCATE` which I know from other DBs)

relay/src/cli.rs
```diff
@@ -314,6 +319,57 @@ pub fn init_config<P: AsRef<Path>>(config_path: P, _matches: &ArgMatches) -> Res
     Ok(())
 }
 
+/// Manages the on-disk spool file.
+pub fn manage_spool(config: &Config, matches: &ArgMatches) -> Result<()> {
+    if let Some(matches) = matches.subcommand_matches("clear") {
+        let path = match matches.get_one::<String>("path") {
```
Dav1dde: I think this should work:

```suggestion
        let path = match matches.get_one::<PathBuf>("path") {
```

relay/src/cli.rs
```diff
@@ -314,6 +319,57 @@ pub fn init_config<P: AsRef<Path>>(config_path: P, _matches: &ArgMatches) -> Res
...
+            Some(path) => {
+                let Ok(path) = PathBuf::from_str(path) else {
+                    bail!("Can't find the provided file path.")
+                };
+                path
+            }
+            None => {
+                let Some(path) = config.spool_envelopes_path() else {
```
Dav1dde: Can use anyhow's [Context](https://docs.rs/anyhow/latest/anyhow/trait.Context.html):

```suggestion
                let path = config.spool_envelopes_path().context("...")?;
```

relay/src/cli.rs
```diff
@@ -314,6 +319,57 @@ pub fn init_config<P: AsRef<Path>>(config_path: P, _matches: &ArgMatches) -> Res
...
+            for line in stdin.lock().lines().map_while(Result::ok) {
+                match line.trim().to_lowercase().as_str() {
+                    "yes" => break,
+                    "no" => bail!("Canceling the cleaning of the spooled data."),
+                    _ => eprintln!("Accepting only YES or NO: "),
+                }
+            }
+        }
+
+        relay_log::info!("Clearing the spool file: {}", path.to_string_lossy());
```
Dav1dde:
```suggestion
        relay_log::info!("Clearing the spool file: {}", path.display());
```

relay/src/cli.rs
```diff
@@ -314,6 +319,57 @@ pub fn init_config<P: AsRef<Path>>(config_path: P, _matches: &ArgMatches) -> Res
     Ok(())
 }
 
+/// Manages the on-disk spool file.
+pub fn manage_spool(config: &Config, matches: &ArgMatches) -> Result<()> {
+    if let Some(matches) = matches.subcommand_matches("clear") {
```
Dav1dde:
```suggestion
    let Some(matches) = matches.subcommand_matches("clear") else { unreachable!() }
```

relay/src/cli.rs
```diff
@@ -314,6 +319,57 @@ pub fn init_config<P: AsRef<Path>>(config_path: P, _matches: &ArgMatches) -> Res
...
+            }
+        }
+
+        relay_log::info!("Clearing the spool file: {}", path.to_string_lossy());
+
+        spool_utils::truncate(path)?;
+
+        relay_log::info!("On-disk spool emptied.")
+    } else {
+        unreachable!();
```
Dav1dde: Might as well return an error, via `bail` imo

relay-server/src/services/spooler/spool_utils.rs
olksdr: yeah, makes sense. Will do

relay/src/cli.rs
olksdr: Followed the pattern like other functions were done, but erroring here makes sense, will change.

olksdr: This also actually never happens that's why this is really unreachable , the `clap` does all the validation and checks before.

Dav1dde: Feel free to leave the `unreachable`, I didn't see that the rest of the code also does that and it does make sense.

relay-server/src/services/spooler/sql.rs
```diff
@@ -155,3 +155,11 @@ pub async fn do_insert(
 
     Ok(count)
 }
+
+/// Creates DELETE statement which truncates the entire `envelopes` table.
+///
+/// When the DELETE statement run without WHERE clause SQLite uses an optimization (so called "truncate" optimization)
+/// to erase the entire table content without having to visit each row.
```
iker-barriocanal:
```suggestion
/// When the DELETE statement runs without WHERE clause SQLite uses an optimization (so called "truncate" optimization)
/// to erase the entire table content without having to visit each row.
```

relay-server/src/services/spooler/spool_utils.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::service::create_runtime;
+use crate::services::spooler::{sql, BufferError};
+
+/// Truncates the spool file deleting all the persisted on-disk data.
+pub fn truncate(path: &PathBuf) -> Result<(), BufferError> {
+    let options = SqliteConnectOptions::new()
+        .filename(path)
+        .journal_mode(SqliteJournalMode::Wal);
+
+    let rt = create_runtime("truncator", 1);
```
iker-barriocanal, line 18: nit: adding the `spooler` prefix may help identify where it's coming from if we need to debug/profile (?)
```suggestion
    let rt = create_runtime("spooler-truncator", 1);
```

relay/src/cli.rs
```diff
@@ -314,6 +318,49 @@ pub fn init_config<P: AsRef<Path>>(config_path: P, _matches: &ArgMatches) -> Res
...
+                "yes" => break,
+                "no" => bail!("Canceling the cleaning of the spooled data."),
+                _ => eprintln!("Accepting only YES or NO: "),
+            }
+        }
+    }
+
+    relay_log::info!("Clearing the spool file: {}", path.to_string_lossy());
+
+    spool_utils::truncate(&path)?;
```
iker-barriocanal: Does this error provide enough info to the end user about what happened? Should Relay print a message that the clearing (partially) failed?

iker-barriocanal (approved):

relay/src/cli.rs
Dav1dde: I believe the `DELETE` returns the amount of rows deleted? Could be an interesting stat to display.

Dav1dde (approved):

relay/src/cli.rs
olksdr: done

relay-server/src/services/spooler/spool_utils.rs
olksdr: This shouldn't run together with anything else, but just execute one command. 
I can rename it later if it's really will be necessary

PR #3185, opened 2024-02-29
test(cardinality): Fixes flaky tests due to sliding window rollovers

Dav1dde:
Test was flaky before, properly fixed now. The `+1` second is not actually necessary and introduced the flakiness, e.g. `UnixTimestamp::from_secs(1799)` could consistently reproduce it. Fun fact: the test failed 1 out of 300 runs on average.

#skip-changelog


olksdr (approved):
:shipit:

PR #3189, opened 2024-03-01
instr(kafka): Tag producer error metric with topic

jjbayer:
There is an alert based on this metric. Tagging it by topic will make the problem faster to diagnose (no need to go and check the accompanying sentry issue).

#skip-changelog


Dav1dde (approved):

olksdr (approved):
🚢  it!

PR #3194, opened 2024-03-02
fix(gocd): Fixing sentry checks by ensuring they use the right release

IanWoodard:
This change stems from the realization that Relay doesn't simply use the SHA when creating a release, but also the version. As a result, the complexity of the scripts had to increase a bit to accommodate this.

Things worth mentioning:
* Since the Sentry project for PoPs is pop-relay, but the service is referred to as relay-pop in google storage, we needed another env var to represent the service rather than using the sentry-project's name.
* Due to there being three env vars that each contain multiple, space-delimited inputs, I felt it was worthwhile to combine them into tuples to mitigate confusion.
* To get the release name, I use the revision and service name to search google storage for the corresponding release name. If there is an easier way, please let me know.
* I realized that since Processing and PoP's pipelines are independent there is no guarantee that they will be deployed at the same time or will even be in sync (though they ideally should be). As a result, using the other service's Sentry release as a signal could be problematic, especially if the pipelines are not kept in sync. One step I took to help alleviate this issue is not cause the check to fail if a given release name is not found.
* Similar to the point directly above, assuming we use Sentry releases as signals from both services, the longer the period between their respective deploys, the more likely it is that a new error will occur and thus cause the new error check to fail. The new error check was originally designed to be run shortly after a deploy goes out since a new error occurring in that short time period is often indicative of a problematic change. As a result, the longer the release has been live before we check, the less useful this signal becomes. While this is not inherently problematic, it is something to consider when it comes to how best we should use these pipelines.

#skip-changelog

### DIFF
--- a/gocd/templates/bash/check-sentry-errors.sh
+++ b/gocd/templates/bash/check-sentry-errors.sh
@@ -1,21 +1,38 @@
 #!/bin/bash
 
+# Check Sentry for errors in the last 5 minutes for a given release (from GO_REVISION_RELAY_REPO)
+#
+# Required environment variables:
+#   DRY_RUN: When dry-run is 'true' this script will not fail if the checks indicate an issue
+#   ERROR_LIMIT: The number of error events to permit before failing
+#   GO_REVISION_RELAY_REPO: Git commit hash (provided by GoCD)
+#   SENTRY_AUTH_TOKEN: Sentry auth token (https://sentry.io/settings/account/api/auth-tokens/) (required by devinfra/scripts/checks/sentry/release_error_events.py)
+#   SENTRY_BASE: Sentry base API URL (e.g. https://sentry.io/api/0)
+#   SENTRY_PROJECTS: A space-separated list of <project_id>:<project_slug>:<service> tuples
+#   SENTRY_SINGLE_TENANT: When single-tenant is 'true' this script will use the sentry-st organization instead of sentry
+#   SKIP_CANARY_CHECKS: Whether to skip checks entirely (true/false)
+
 # shellcheck disable=SC2206
-project_ids=(${SENTRY_PROJECT_IDS})
-# shellcheck disable=SC2206
-project_slugs=(${SENTRY_PROJECTS})
+projects=(${SENTRY_PROJECTS})
 
+for project in "${projects[@]}"; do
+  IFS=':' read -r -a project_info <<<"$project"
+  project_id="${project_info[0]}"
+  project_slug="${project_info[1]}"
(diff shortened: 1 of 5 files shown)


olksdr (approved):
The change look ok to me. 
But how  do you feel if the bash scripts could be migrated to e.g. `python` instead? 

They do get a little bit more complicated with some `bash` specific magic, which might be hard to understand, especially if you do not deal with the language on the daily basis. 


The `check-sentry-errors.sh` and the  `check-sentry-new-errors.sh` seem to be very similar and maybe it would be replaced just by one python script instead?
Which we could also write some tests for. 

> SENTRY_PROJECTS: A space-separated list of <project_id>:<project_slug>:<service> tuples

 this also introduces a little bit complexity, and maybe more docs on why this is done directly in the code would help.

gocd/templates/bash/check-sentry-errors.sh
```diff
@@ -1,21 +1,38 @@
...
+    continue
+  fi
 
-for i in "${!project_ids[@]}"; do
   /devinfra/scripts/checks/sentry/release_error_events.py \
-    --project-id="${project_ids[i]}" \
-    --project-slug="${project_slugs[i]}" \
-    --release="relay@${GO_REVISION_RELAY_REPO}" \
+    --project-id="${project_id}" \
+    --project-slug="${project_slug}" \
```
jjbayer, line 36: Why does `release_error_events` need the project slug? Isn't the project ID enough to uniquely identify the project?

jjbayer (approved):

gocd/templates/bash/check-sentry-errors.sh
IanWoodard: It uses two different api calls each of which require different unique identifiers.

IanWoodard:
> The change look ok to me. But how do you feel if the bash scripts could be migrated to e.g. `python` instead?
> 
> They do get a little bit more complicated with some `bash` specific magic, which might be hard to understand, especially if you do not deal with the language on the daily basis.
> 
> The `check-sentry-errors.sh` and the `check-sentry-new-errors.sh` seem to be very similar and maybe it would be replaced just by one python script instead? Which we could also write some tests for.
> 
> > SENTRY_PROJECTS: A space-separated list of <project_id>:<project_slug>: tuples
> 
> this also introduces a little bit complexity, and maybe more docs on why this is done directly in the code would help.

I am all for moving these two python. I can put that on the roadmap for when I finish testing the new pipelines. I'll add some more comments in the meantime.

PR #3198, opened 2024-03-04
ref(store): Emit store metric from a single location

Dav1dde:
Emit the metrics from a single location, makes it impossible to forget to add a metric and cleans up the code a bit.

Re-uses the `variant` on the Kafka message which already yielded the same strings as previously manually as the `event_type`.

#skip-changelog

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -239,10 +239,6 @@ impl StoreService {
                         start_time,
                         item,
                     )?;
-                    metric!(
-                        counter(RelayCounters::ProcessingMessageProduced) += 1,
-                        event_type = "user_report"
-                    );
                 }
                 ItemType::Session | ItemType::Sessions => {
                     self.produce_sessions(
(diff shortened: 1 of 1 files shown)


relay-server/src/services/store.rs
```diff
@@ -511,6 +493,37 @@ impl StoreService {
...
+                    has_parent = bool_to_str(has_parent),
+                );
+            }
+            KafkaMessage::ReplayRecordingNotChunked(replay) => {
+                let has_video = replay.replay_video.is_some();
+
+                metric!(
+                    counter(RelayCounters::ProcessingMessageProduced) += 1,
+                    event_type = message.variant(),
+                    has_video = bool_to_str(has_video),
```
jjbayer, line 516: Is this new tag documented on `RelayCounters`? We should also let @cmanallen know that this is now reported in a different way.

jjbayer (approved):
Thanks for cleaning this up!

relay-server/src/services/store.rs
Dav1dde: Before there were two different event types, it's now the same event type but using a tag. I'll shoot him a message on Slack 👍

jjbayer: Thanks! Would be nice to add the `has_video` to https://github.com/getsentry/relay/blob/c63a2ed070b2a4b194c05575eed27a44debec429/relay-server/src/statsd.rs#L547-L552 for completeness.

PR #3199, opened 2024-03-04
ref(cardinality): Passive mode per cardinality limit

Dav1dde:
Makes it possible to mark individual cardinality limits per project to be marked as passive.

Initially the idea was not to add the `passive` flag to the individual limit and only enforce the passive property/mode in the processor. But it turns out this is more complicated than just directly discarding a passive result in the cardinality limiter. This was made easy due to already having the correct abstraction (`Rejections` trait) in place, it just needed to be slightly updated to be passed the limit instead of just the limit id. This approach also allows us to tag all metrics and cardinality errors with a passive tag.

#skip-changelog

### DIFF
--- a/relay-cardinality/src/config.rs
+++ b/relay-cardinality/src/config.rs
@@ -1,14 +1,21 @@
 use relay_base_schema::metrics::MetricNamespace;
 use serde::{Deserialize, Serialize};
 
-use crate::{CardinalityItem, SlidingWindow};
+use crate::SlidingWindow;
 
 /// A cardinality limit.
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CardinalityLimit {
     /// Unique identifier of the cardinality limit.
     pub id: String,
+    /// Wether this is a passive limit.
+    ///
+    /// Passive limits are tracked separately to normal limits
+    /// and are not enforced, but still evaluated.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub passive: bool,
+
     /// The sliding window to enforce the cardinality limits in.
     pub window: SlidingWindow,
     /// The cardinality limit.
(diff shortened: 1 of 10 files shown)


relay-cardinality/src/config.rs
```diff
@@ -4,11 +4,18 @@ use serde::{Deserialize, Serialize};
 use crate::SlidingWindow;
 
 /// A cardinality limit.
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CardinalityLimit {
     /// Unique identifier of the cardinality limit.
     pub id: String,
+    /// Wether this is a passive limit.
```
olksdr:
```suggestion
    /// Whether this is a passive limit.
```

olksdr (approved):

relay-cardinality/src/limiter.rs
```diff
@@ -434,4 +439,89 @@ mod tests {
...
+        let rejected = limited.rejected().collect::<HashSet<_>>();
+        assert_eq!(
+            rejected,
+            HashSet::from([
+                &Item::new(0, MetricNamespace::Custom),
+                &Item::new(2, MetricNamespace::Custom),
+                &Item::new(4, MetricNamespace::Custom),
+            ])
+        );
+        drop(rejected); // NLL are broken here without the explicit drop
```
jjbayer, line 517: What's NLL?

relay-cardinality/src/limiter.rs
```diff
@@ -176,7 +178,9 @@ impl<'a, T> CardinalityLimits<'a, T> {
     }
 
     /// Returns all id's of cardinality limits which were exceeded.
-    pub fn enforced_limits(&self) -> &HashSet<&'a str> {
+    ///
+    /// This includes passive limits.
+    pub fn limits(&self) -> &HashSet<&'a CardinalityLimit> {
```
jjbayer, line 183:
```suggestion
    pub fn exceeded_limits(&self) -> &HashSet<&'a CardinalityLimit> {
```

jjbayer (approved):
Might be worth a changelog entry, since it adds a new field to project configs.

relay-cardinality/src/limiter.rs
Dav1dde: Non lexical lifetimes.

The compiler should be smart enough to see that rejected is no longer used, but without the drop I cannot call `into_accepted` because there is still the reference to `rejected`.

relay-cardinality/src/limiter.rs
Dav1dde: I'll do that in a follow-up, I need. a librelay release to continue and I don't feel like waiting another 30 minutes...

PR #3201, opened 2024-03-04
feat(replays): Reject replay envelopes if any item contained within is invalid

### LINKED ISSUE #3180: Drop replay evelope if any of the items fails validation
Sometimes a replay-recording is rejected but the replay-event is allowed through.  This leads to confusing customer experiences.

cmanallen:
Closes: https://github.com/getsentry/relay/issues/3180

Drop the envelope in its entirety if any component fails validation.  This prevents confusing states where is shown a replay which was never ingested.


relay-server/src/services/processor/replay.rs
```diff
@@ -113,6 +130,12 @@ pub fn process(
         }
     });
 
+    if dropped_item {
+        state
+            .managed_envelope
+            .reject(Outcome::Invalid(DiscardReason::ReplayRejected));
+    }
```
jjbayer: This leads to different behavior depending on the order of items:

1. If the first item is dropped with an outcome, all subsequent items are dropped silently (one outcome in total).
2. If the second item is dropped with an outcome, the first outcome will still be in the envelope after `retain_items`. In that case, the first item will be rejected with `ReplayRejected` outcome (two outcomes in total).

Maybe we can replace `retain_items` with a simple for loop that breaks on the first failure, and then rejects the envelope with the outcome of the first failure.

jjbayer (changes requested):
See comment. We definitely need a few tests for this at well.

relay-server/src/services/processor/replay.rs
cmanallen: @jjbayer I thought `reject` logged and emitted no outcomes?  This is what I was referencing when writing the PR.  https://github.com/getsentry/relay/blob/c1f1aff8674345b8787739efbcf32296e8c140c2/relay-server/src/utils/managed_envelope.rs#L409

jjbayer: `reject` does log outcomes (the `Handling` is only used to decide whether or not to log an error). But it turns out we don't log outcomes for replays in `reject`, which is probably just oversight: https://github.com/getsentry/relay/blob/8eff86344e5e313583c131892f29bb825178d7da/relay-server/src/utils/managed_envelope.rs#L433-L468

Not sure what the best way forward is here, maybe the best would be to return a `ProcessingError`, which is then automatically reported as an outcome via `ProcessingError::to_outcome`.

cmanallen: I've added `retain_or_reject_all` methods to `ManagedEnvelope` and `Envelope`.  It guarantees one outcome is emitted, breaks early on processing error, ensures the envelope is empty after error, and uses all of the familiar APIs replay is accustomed to.

jjbayer: I looked into the code some more, I'm convinced now that the cleanest way to reject an entire envelope is to raise a `ProcessingError` and let it bubble up, because that's API that already exists for this purpose. And to make sure that `ManagedEnvelope::reject` emits outcomes for replays.

For the test, you can add the `outcomes_consumer` fixture in existing replay integration tests for that, or add new tests in `test_outcome.py`.

iker-barriocanal:
Is the goal to have a single outcome per replay? IIRC, in the past replays were split into two envelopes; how would outcome generation behave in that situation?

relay-server/src/services/processor/replay.rs
cmanallen: @jjbayer This commit implements the ProcessingError flow https://github.com/getsentry/relay/pull/3201/commits/27fcc83e9fa23cbd23b3435f570e80fb3db94aaa

cmanallen:
@iker-barriocanal This change doesn't handle that case but its not a common path.  It might not be used any more.

relay-server/src/services/processor/replay.rs
```diff
@@ -74,10 +73,13 @@ pub fn process(
                     client_addr,
                     user_agent,
                 ) {
-                    Err(outcome) => ItemAction::Drop(outcome),
+                    Err(outcome) => {
+                        if let Outcome::Invalid(reason) = outcome {
+                            return Err(ProcessingError::Invalid(reason));
+                        }
+                    }
```
jjbayer: You can now refactor `handle_replay_event_item` and other helper functions to return `Result<Bytes, DiscardReason>`, then this becomes
```suggestion
                    Err(reason) => {
                        return Err(ProcessingError::Invalid(reason));
                    }
```

Or even (haven't tested this)

```rust
let replay_event = handle_replay_event_item(...).map_err(ProcessingError::Invalid)?;
```

relay-server/src/envelope.rs
```diff
@@ -1319,6 +1319,11 @@ impl Envelope {
         self.items.retain(f)
     }
 
+    /// Drops every item in the envelope.
+    pub fn drop_items(&mut self) {
+        self.items.truncate(0);
```
```suggestion
        self.items.clear();
```

relay-server/src/services/processor/replay.rs
```diff
@@ -55,13 +54,13 @@ pub fn process(
...
 
-    state.managed_envelope.retain_items(|item| {
-        // If replays aren't enabled or an item was dropped - drop the remainder of the
-        // envelope.
-        if !replays_enabled {
-            return ItemAction::DropSilently;
-        }
+    // If the replay feature is not enabled drop the items silenty.
+    if !replays_enabled {
+        state.managed_envelope.drop_items();
```
jjbayer: This would also work, and would keep the API surface small. But `drop_items` is easier to read so I don't feel strongly about it:

```suggestion
        state
            .managed_envelope
            .retain_items(|_| ItemAction::DropSilently);
```

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@
 - Adds support for dynamic metric bucket encoding. ([#3137](https://github.com/getsentry/relay/pull/3137))
 - Use statsdproxy to pre-aggregate metrics. ([#2425](https://github.com/getsentry/relay/pull/2425))
 - Add SDK information to spans. ([#3178](https://github.com/getsentry/relay/pull/3178))
+- Drop replay envelopes if any item fails. ([#3201](https://github.com/getsentry/relay/pull/3201))
+- Add `retain_or_reject_all` methods. ([#3201](https://github.com/getsentry/relay/pull/3201))
```
```suggestion
```

jjbayer (commented):
Nice, with some tests this should be good to go!

relay-server/src/services/processor.rs
```diff
@@ -418,6 +418,9 @@ pub enum ProcessingError {
 
     #[error("invalid processing group type")]
     InvalidProcessingGroup(#[from] InvalidProcessingGroupType),
+
+    #[error("envelope invalid")]
+    Invalid(DiscardReason),
```
iker-barriocanal: Let's be a bit more explicit with the error as in the examples above, for example:
```suggestion
    #[error("invalid replay")]
    InvalidReplay(DiscardReason),
```
An ambiguous name could lead to making wrong assumptions when looking at errors.

relay-server/src/services/processor/replay.rs
iker-barriocanal: I'd either rename to `drop_items_silently` or use `retain_items` to make it clear there are no side-effects.

cmanallen:
@iker-barriocanal @jjbayer Adding test coverage for this.  I'm seeing in the logs `relay_server::utils::managed_envelope: dropped envelope: invalid data (invalid_replay)` but no outcomes emitted.  The reason is `ManagedEnvelope.context.summary.event_category` is `None`.  How can I populate this?

iker-barriocanal:
@cmanallen Relay doesn't infer event categories for replays, see
https://github.com/getsentry/relay/blob/912a697da8dee676d32b1d0cebf7942d14da900d/relay-server/src/utils/rate_limits.rs#L110-L112
I believe updating the applicable cases should populate the envelope context.
Edit: actually, this could have bad side-effects.

cmanallen:
@iker-barriocanal I updated those locally and it didn't change the test failure :\

jjbayer:
@cmanallen I added the missing replay rejection to `ManagedEnvelope::reject` and pushed a commit!

The quantity of the outcome will still be the number of items, because that is consistent with how we count replays for rate limiting purposes.

cmanallen:
@jjbayer Updated test coverage and changed quantity to only emit one outcome.  This should be ready.

relay-server/src/utils/rate_limits.rs
```diff
@@ -107,9 +107,9 @@ fn infer_event_category(item: &Item) -> Option<DataCategory> {
         ItemType::FormData => None,
         ItemType::UserReport => None,
         ItemType::Profile => None,
-        ItemType::ReplayEvent => None,
-        ItemType::ReplayRecording => None,
-        ItemType::ReplayVideo => None,
+        ItemType::ReplayEvent => Some(DataCategory::Replay),
+        ItemType::ReplayRecording => Some(DataCategory::Replay),
+        ItemType::ReplayVideo => Some(DataCategory::Replay),
```
cmanallen: Should I remove this?

jjbayer: Yes, please revert this change. This function should only be used for envelopes that contain sentry events i.e. errors, transactions, etc.

relay-server/src/utils/managed_envelope.rs
```diff
@@ -460,13 +465,17 @@ impl ManagedEnvelope {
...
-                outcome,
+                outcome.clone(),
                 // Secondary transaction counts are never indexed transactions
                 DataCategory::Transaction,
                 self.context.summary.secondary_transaction_quantity,
             );
         }
 
+        if self.context.summary.replay_quantity > 0 {
+            self.track_outcome(outcome.clone(), DataCategory::Replay, 1);
```
jjbayer: I strongly recommend tracking `replay_quantity` here instead of `1`. If we want to change the meaning of replay quantities, we should do it in a separate PR and align with both rate limits and the sentry component that reports `"accepted"` outcomes for replays.

jjbayer (approved):

PR #3204, opened 2024-03-04
feat(metrics_summaries): Allow to parse null tags and filter them

phacops:
We are getting some errors around null tag values when we're trying to deserialize metrics summaries to push them to their topic.

This will allow null values to be deserialized but still filtered before they end up on the topic.


jjbayer (approved):
A test case would be nice, but LGTM!

Dav1dde (changes requested):
I don't like doing data manipulation in the store service. I still think the consumers should be able to deal with it, but at the very least we should do these kinds of things in the processor/normalization step.

Doing it here in the store service is a slippery slope of just adding more and more and more.

PR #3206, opened 2024-03-04
fix(spans): Gate span metric extraction on the feature flag

phacops:
For a while, we were not ingesting all spans but we still wanted to ingest metrics summaries and the related spans.

Now we're ingesting all spans, we want to make sure it's still gated properly (only active for AM2 plans and above). This will stop ingesting segments and the related metrics summaries for AM1 customers.

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor/span/processing.rs
+++ b/relay-server/src/services/processor/span/processing.rs
@@ -148,6 +148,13 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState<TransactionGroup>) {
         return;
     };
 
+    if !state
+        .project_state
+        .has_feature(Feature::SpanMetricsExtraction)
+    {
+        return;
+    }
+
     let mut add_span = |span: Annotated<Span>| {
         let span = match validate(span) {
             Ok(span) => span,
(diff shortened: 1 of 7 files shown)


tests/integration/test_store.py
```diff
@@ -1845,6 +1845,7 @@ def test_span_extraction_with_metrics_summary(
     project_config = mini_sentry.add_full_project_config(project_id)
     project_config["config"]["features"] = [
         "organizations:custom-metrics",
+        "projects:span-metrics-extraction",
```
iker-barriocanal, line 1848: Bonus points for a test that validates that no feature flag results in no metrics.

iker-barriocanal (approved):

relay-server/src/services/processor/span/processing.rs
```diff
@@ -180,63 +187,50 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState<TransactionGroup>) {
...
-        .project_state
-        .has_feature(Feature::SpanMetricsExtraction);
-    let custom_metrics_enabled = state.project_state.has_feature(Feature::CustomMetrics);
-
     let Some(event) = state.event.value() else {
         return;
     };
 
-    let extract_transaction_span = span_metrics_extraction_enabled
-        || (custom_metrics_enabled && !event._metrics_summary.is_empty());
```
jjbayer, line 193: I assume this extra condition was here for a reason? I recently looked at this with @iambriccardo, did we want to enable transaction spans for anyone who emits custom metrics?

Anyway, I prefer the new version because it's simpler. Just wanted to double-check.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -148,6 +148,13 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState<TransactionGroup>) {
         return;
     };
 
+    if !state
+        .project_state
+        .has_feature(Feature::SpanMetricsExtraction)
```
jjbayer: We should document the `SpanMetricsExtraction` variant to reflect that it toggles both spans and span metrics. Maybe even rename the variant (but not the string representation).

jjbayer (approved):

relay-server/src/services/processor/span/processing.rs
phacops: We added this at the time where only the `all-modules` feature flag ingested segments, and metrics summaries were on segments only. This allowed us to bypass the span feature flags and ingest metrics summaries.

Now that we're indexing all spans, for all orgs, it's not needed anymore: if we see metrics summaries, we'll ingest them.

Part of the issue here is, as it was bypassing the span feature flag, it allowed us to ingest spans for AM1 customers, which we don't want.

If we ever go back to a mode where we do not index everything, we'll likely need something to ingest span + metrics summaries for selected spans again.

tests/integration/test_store.py
phacops: Bonus point unlocked.

PR #3209, opened 2024-03-04
fix(spans): Emit a usage metric for every span seen

phacops:
We've made it so we emit a usage metric for each span we extract a metric for which in turns will be used to emit a processed span outcome.

We'd like to change this to emit a usage metric for every span we see as Relay is still processing those spans even if it's not to store derived data from them. This has some impact for billing.

As we improve the product, we anticipate to store more and more information regarding those spans and the overall cost for the customer will not change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 - Use statsdproxy to pre-aggregate metrics. ([#2425](https://github.com/getsentry/relay/pull/2425))
 - Add SDK information to spans. ([#3178](https://github.com/getsentry/relay/pull/3178))
 - Filter null values from metrics summary tags. ([#3204](https://github.com/getsentry/relay/pull/3204))
-- Emit a processed span outcome for every span seen. ([#3209](https://github.com/getsentry/relay/pull/3209))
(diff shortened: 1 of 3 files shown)


mitsuhiko (approved):

iker-barriocanal (approved):
nit: could we reword the PR title and description? Outcomes have a specific meaning in Relay. "usage metric" suits better imo.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - Use statsdproxy to pre-aggregate metrics. ([#2425](https://github.com/getsentry/relay/pull/2425))
 - Add SDK information to spans. ([#3178](https://github.com/getsentry/relay/pull/3178))
 - Filter null values from metrics summary tags. ([#3204](https://github.com/getsentry/relay/pull/3204))
+- Emit a processed span outcome for every span seen. ([#3209](https://github.com/getsentry/relay/pull/3209))
```
```suggestion
- Emit a usage metric for every span seen. ([#3209](https://github.com/getsentry/relay/pull/3209))
```

Agree with @iker-barriocanal that we should talk about the usage metric here, not outcomes. The PR description can still mention that this usage metric is translated to accepted outcomes in sentry.

jjbayer (approved):
A

Dav1dde (approved):

PR #3213, opened 2024-03-05
feat(server): Drop transactions in a spans-only world

jjbayer:
To test what our Performance product looks like without transactions (spans only), add the ability to drop transactions in Relay. The feature flag will only be enabled for a single experimental project for the time being.

This PR re-applies https://github.com/getsentry/relay/pull/3165, and fixes the bug that caused empty events to end up on the errors topic.

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -1346,8 +1346,12 @@ impl EnvelopeProcessorService {
                     event::scrub(state)?;
                     if_processing!(self.inner.config, {
                         span::extract_from_event(state);
-                        span::maybe_discard_transaction(state);
                     });
+                }
+
+                span::maybe_discard_transaction(state);
+
+                if state.has_event() {
                     event::serialize(state)?;
                 }
 
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -1349,7 +1349,9 @@ impl EnvelopeProcessorService {
                     });
                 }
 
-                span::maybe_discard_transaction(state);
+                if_processing!(self.inner.config, {
+                    span::maybe_discard_transaction(state);
+                });
 
                 if state.has_event() {
                     event::serialize(state)?;
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -1344,10 +1344,11 @@ impl EnvelopeProcessorService {
 
                 if state.has_event() {
                     event::scrub(state)?;
-                    event::serialize(state)?;
                     if_processing!(self.inner.config, {
                         span::extract_from_event(state);
+                        span::maybe_discard_transaction(state);
                     });
+                    event::serialize(state)?;
                 }
 
                 attachment::scrub(state);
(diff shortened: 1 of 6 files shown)


relay-server/src/services/processor.rs
```diff
@@ -1344,12 +1344,19 @@ impl EnvelopeProcessorService {
...
                         span::extract_from_event(state);
                     });
                 }
 
+                if_processing!(self.inner.config, {
+                    span::maybe_discard_transaction(state);
+                });
+
+                if state.has_event() {
+                    event::serialize(state)?;
```
jjbayer, line 1357: In https://github.com/getsentry/relay/pull/3165, I did not check `has_event` again after `maybe_discard_transaction`, which made `event::serialize` assume there was an event.

relay-server/src/services/processor/event.rs
```diff
@@ -332,6 +332,11 @@ pub fn scrub<G: EventProcessing>(
 pub fn serialize<G: EventProcessing>(
     state: &mut ProcessEnvelopeState<G>,
 ) -> Result<(), ProcessingError> {
+    if state.event.is_empty() {
```
jjbayer, line 335: This check is redundant with the second `has_event` check in `process_transactions`, but it's worth checking here to be resilient against future refactors.

Ideally, we make `event` a return value of `event::extract`, and remove it from the state entirely to prevent future bugs where we use something that should not be accessible in the first place.

tests/integration/test_store.py
```diff
@@ -1366,6 +1374,15 @@ def test_span_extraction(
 
     relay.send_event(project_id, event)
 
+    if discard_transaction:
+        transactions_consumer.poll(timeout=2.0) is None
+
+        # We do not accidentally produce to the events topic:
+        assert events_consumer.poll(timeout=2.0) is None
```
jjbayer, line 1381: Without the extra `has_event` checks, the envelope processor would serialize an empty event that gets published to the events topic.

olksdr (approved):

Dav1dde (approved):
:shipit: what's the worst that can happen?

iker-barriocanal (approved):
LGTM. A couple of non-blocking questions:
- Should we continue extracting metrics from transactions that will be dropped?
- Should Relay emit outcomes for transactions that will be dropped?

jjbayer:
> Should we continue extracting metrics from transactions that will be dropped?

This is a good question. @shruthilayaj to get an accurate picture of what the product looks like without transactions, we should probably also drop any transaction metrics, correct? I did not consider this in the first PR.

shruthilayaj:
> > Should we continue extracting metrics from transactions that will be dropped?
> 
> This is a good question. @shruthilayaj to get an accurate picture of what the product looks like without transactions, we should probably also drop any transaction metrics, correct? I did not consider this in the first PR.

I had thought about it, if we can do this easily/safely please go ahead!

Dav1dde:
> I had thought about it, if we can do this easily/safely please go ahead!

Should we do that in a follow up, we had issues with this PR last time, maybe let's see first if everything is good now with this one before adding more to it.

jjbayer:
> Should we do that in a follow up, we had issues with this PR last time, maybe let's see first if everything is good now with this one before adding more to it.

Follow-up sounds good to me!

PR #3214, opened 2024-03-05
fix(normalization): Skip validation on renormalization

iker-barriocanal:
Previous normalization PRs run validation from librelay calls. However, some `sentry` tests run full normalization in librelay on partial (i.e. invalid) payloads, and running validation makes plenty of tests fail (see https://github.com/getsentry/sentry/pull/66134).

Similarly, many `sentry` tests expect span normalization to never happen. After spending significant time "fixing"* some tests, it'd require me even more time to "fix" the remaining ones as it requires a deeper understanding of other features/codebases. Thus, I'm reintroducing `normalize_spans` to disable span normalization (mostly exclusive time computation) from librelay calls and unblock the librelay release.

*Note: "fixing" a test means using data relay outputs from normalization. This doesn't align with other assumptions made in the tests, nor modifications to the event payload in the rest of the event processing pipeline.

### Future work
- Consider span normalization a regular step in event normalization behind the `is_renormalize` flag, removing the `normalize_spans` flag.

### DIFF
--- a/relay-event-normalization/src/validation.rs
+++ b/relay-event-normalization/src/validation.rs
@@ -19,6 +19,9 @@ pub struct TransactionValidationConfig {
     /// Transactions that finish outside this range are invalid. The check is
     /// skipped if no range is provided.
     pub timestamp_range: Option<Range<UnixTimestamp>>,
+
+    /// Controls whether the event has been validated before, in which case disables validation.
+    pub is_renormalize: bool,
 }
 
 /// Validates a transaction.
@@ -35,6 +38,10 @@ pub fn validate_transaction(
     event: &mut Annotated<Event>,
     config: &TransactionValidationConfig,
 ) -> ProcessingResult {
+    if config.is_renormalize {
+        return Ok(());
+    }
+
     let Annotated(Some(ref mut event), ref _meta) = event else {
         return Ok(());
     };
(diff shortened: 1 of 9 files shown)


relay-event-normalization/src/validation.rs
```diff
@@ -237,6 +244,9 @@ pub struct EventValidationConfig {
     ///
     /// If the event's timestamp lies further into the future, the received timestamp is assumed.
     pub max_secs_in_future: Option<i64>,
+
+    /// Controls whether the event has been validated before, in which case disables validation.
+    pub is_renormalize: bool,
```
olksdr: Have you considered to name this differently to make sure it's clear what the option means without reading the docs? 

To be honest even docs do not describe now what the option should be set to disable validation.

Maybe? 
```suggestion
    pub is_validated: bool
```

iker-barriocanal: I chose `is_renormalize` because it's a known name and concept we use in normalization and librelay. Is `is_validated` more clear? I can update the name and expand docs if it helps.

jjbayer (approved):
Approving to unblock. It would be nice to get rid of the config flag eventually.

Dav1dde (approved):
I don't fully understand the change and it seems to me we're working around bad tests in Sentry, but we are definitely blocked on this.

Have you considered exposing the flag from librelay via Keyword arg, so we have at least a chance of fixing the Sentry tests on a step by step basis?

relay-event-normalization/src/validation.rs
iker-barriocanal: Should be addressed in [bd72909](https://github.com/getsentry/relay/pull/3214/commits/bd729097df816356731b6c53541eea860106119d), let me know if it's still unclear or there's a better wording for it.

iker-barriocanal:
@Dav1dde great idea! Exposed the flag in [380895a](https://github.com/getsentry/relay/pull/3214/commits/380895abe8f8159d1dafc44dfe4046a63da21830). It can be enabled by initing the `StoreNormalizer` with `normalize_spans=True, is_renormalize=False`.

PR #3215, opened 2024-03-05
ref(cardinality): Adds passive mode changelog entries, renames fn

Dav1dde:
Followup of #3199 

Renames the `limits` fn to a better name, adds changelog and librelay changelog entry (project config is affected).

#skip-changelog


jjbayer (approved):
Thanks!

olksdr (approved):

iker-barriocanal (approved):

PR #3216, opened 2024-03-05
ref(filters): Move generic filter checks to event::filter

iker-barriocanal:
Follow-up to https://github.com/getsentry/relay/pull/3161#discussion_r1508884588.

This PR moves the check whether Relay supports generic filters from the processor to `event::filter`, which returns an enum of what Relay should do with the event. There are no functional changes.

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor/event.rs
+++ b/relay-server/src/services/processor/event.rs
@@ -271,7 +271,7 @@ pub fn finalize<G: EventProcessing>(
     Ok(())
 }
 
-/// Decision for applying some filters that don't drop the event.
+/// Status for applying some filters that don't drop the event.
 ///
 /// The enum represents either the success of running all filters and keeping
 /// the event, [`FiltersApplied::Ok`], or not running all the filters because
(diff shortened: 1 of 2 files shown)


relay-server/src/services/processor/event.rs
```diff
@@ -271,15 +271,35 @@ pub fn finalize<G: EventProcessing>(
     Ok(())
 }
 
+/// Decision for applying some filters that don't drop the event.
```
iker-barriocanal: Bikeshedding a bit: I don't really like "decision" here, but the alternatives "result" and "outcome" have other meanings that may generate some confusion. Open to suggestions.

relay-server/src/services/processor/event.rs
```diff
@@ -271,15 +271,35 @@ pub fn finalize<G: EventProcessing>(
...
+///
+/// The enum represents either the success of running all filters and keeping
+/// the event, [`FiltersApplied::Ok`], or not running all the filters because
+/// some are unsupported, [`FiltersApplied::Unsupported`].
+///
+/// If there are unsuppported filters, Relay should forward the event upstream
+/// so that a more up-to-date Relay can apply filters appropriately. Actions
+/// that depend on the outcome of event filtering, such as metric extraction,
+/// should be postponed until a filtering decision is made.
+pub enum FiltersApplied {
```
Dav1dde: What about:

```rs
enum FilterStatus {
    Ok,
    Unsupported,
}
```

Dav1dde:
```suggestion
#[must_use]
pub enum FiltersApplied {
```

Dav1dde (approved):
Thanks for following up on this.

PR #3217, opened 2024-03-05
Always print memory usage from readiness probe

beezz:
For debugging purposes, always log memory usage from the readiness probe.

### Legal Boilerplate


Dav1dde:
@beezz should we have a distribution metric instead?

beezz:
> @beezz should we have a distribution metric instead?

I did not mean to merge it like this, just to run it for a few tests but the distribution metric sounds much better 👍

Dav1dde:
> I did not mean to merge it like this, just to run it for a few tests

Ah I forgot you have OPS powers and can do that 👍

Dav1dde:
Closing this now, think this is no longer needed and we have metrics emitted from Relay now.

PR #3218, opened 2024-03-05
feat(bucket-encoding): Implements base64 and zstd encodings

Dav1dde:
- Implements Base64 and zstd metric bucket encodings.
- Moves all bucket encoding related logic into its own file.
- Splits the config options into two separate options to control sets from distributions separately.
- Makes it impossible for sessions to configure a bucket encoding other than `legacy`.

In a followup PR the encoding/compression of values should be moved into the processor.

### DIFF
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -182,7 +182,6 @@ pub enum CardinalityLimiterMode {
 #[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
 #[serde(default)]
 pub struct BucketEncodings {
-    sessions: BucketEncoding,
     transactions: BucketEncoding,
     spans: BucketEncoding,
     profiles: BucketEncoding,
@@ -193,12 +192,11 @@ impl BucketEncodings {
     /// Returns the configured encoding for a specific namespace.
     pub fn for_namespace(&self, namespace: MetricNamespace) -> BucketEncoding {
         match namespace {
-            MetricNamespace::Sessions => self.sessions,
             MetricNamespace::Transactions => self.transactions,
             MetricNamespace::Spans => self.spans,
             MetricNamespace::Profiles => self.profiles,
             MetricNamespace::Custom => self.custom,
-            MetricNamespace::Unsupported => BucketEncoding::default(),
+            _ => BucketEncoding::Legacy,
         }
     }
 }
@@ -225,7 +223,6 @@ where
         {
             let encoding = BucketEncoding::deserialize(de::value::StrDeserializer::new(v))?;
             Ok(BucketEncodings {
-                sessions: encoding,
                 transactions: encoding,
                 spans: encoding,
                 profiles: encoding,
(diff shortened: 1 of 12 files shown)


relay-dynamic-config/src/global.rs
```diff
@@ -170,37 +177,83 @@ pub enum CardinalityLimiterMode {
...
-    pub fn for_namespace(&self, namespace: MetricNamespace) -> MetricEncoding {
+    pub fn for_namespace(&self, namespace: MetricNamespace) -> BucketEncoding {
         match namespace {
-            MetricNamespace::Sessions => self.sessions,
             MetricNamespace::Transactions => self.transactions,
             MetricNamespace::Spans => self.spans,
             MetricNamespace::Profiles => self.profiles,
             MetricNamespace::Custom => self.custom,
-            MetricNamespace::Unsupported => self.unsupported,
+            _ => BucketEncoding::Legacy,
```
jjbayer, line 203: Should we add an explicit arm for `::Sessions` with a comment explaining that the session metrics consumer does not support any encoding other than legacy?

relay-server/src/services/store.rs
```diff
@@ -803,19 +812,17 @@ impl StoreService {
...
-            let encoding = encodings.for_namespace(namespace);
+        for mut bucket in serde_json::from_slice::<Vec<Bucket>>(&payload).unwrap_or_default() {
+            let namespace = encoder.prepare(&mut bucket);
 
-            let message =
-                self.create_metric_message(org_id, project_id, encoding, &view, retention);
-            self.send_metric_message(namespace, message)?;
+            let view = BucketView::new(&bucket);
+            let message = self
+                .create_metric_message(org_id, project_id, &encoder, namespace, &view, retention);
```
jjbayer, line 824: Should we be conservative here and fall back to legacy or array encoding when encoding fails?

relay-server/src/utils/bucket_encoding.rs
```diff
@@ -0,0 +1,201 @@
...
+pub enum DynamicArrayEncoding<T> {
+    /// Array encoding.
+    ///
+    /// Encodes all items as an array.
+    Array { data: T },
+    /// Base64 (with padding) encoding.
+    ///
+    /// Converts all items to little endian byte sequences
+    /// and Base64 encodes the raw little endian bytes.
+    Base64 { data: String },
```
jjbayer, line 116: Do we even need this mode? Even if it seems like a sub-mode of `zstd`, it's still a different case that the consumer needs to be able to support correctly -> more room for errors.

Dav1dde: My plan was to use it for sets in the beginning.

relay-server/src/services/store.rs
Dav1dde: We could, do you see any scenario where this would make sense? 

I am pretty certain that it can't error in practice since the only error should be coming from the `Write` and we're writing to an in memory buffer, which should never fail.

jjbayer: Alright, if failure is improbable then a fallback is probably overkill.

jjbayer (approved):

PR #3220, opened 2024-03-05
fix(health): Switch to sysinfo fork for correct cgroup v2 mem limits

Dav1dde:
The memory health check is broken for cgroup v2, see https://github.com/GuillaumeGomez/sysinfo/issues/1219

Switches to a fork until the upstream issue is fixed.

#skip-changelog


iker-barriocanal (approved):

olksdr (commented):
Before merging it would be great to test it with cgroup v1 and cgroup v2 and both work exactly the same and return expected results. So there are no unexpected surprises again.

Dav1dde:
@olksdr agreed, to compare the values #3223

Issue #3224, opened 2024-03-06
Pass on multipart attachments without a content-type


supervacuus:
This relates to https://github.com/getsentry/sentry-native/issues/955.

A summary of the above issue is:
* attachments sent via the `crashpad_handler` (which sends them as part of a multipart request) cannot be previewed in the web-ui because they all have content-type `application/octet-stream` by default.
* attachments sent as envelope items typically do not get a content-type header, in which case they will be MIME detected in the backend and only assigned `application/octet-stream` [if no MIME type could be identified](https://github.com/getsentry/sentry/blob/42050bfa4f3fcc6b19c9f104eba868dcc8535eaa/src/sentry/models/eventattachment.py#L184-L187).
* users would like to see the same behavior for attachments sent via the crashpad backend.

To achieve this in the Native SDK, we'd remove the content-type header from the multipart attachments sent via the `crashpad_handler`. This would reflect how we send attachment envelop items, which we also send without content type (this seems to be the default for other SDKs).

But to make this change in the Native SDK meaningful, we'd need Relay to pass on multipart attachments without content type if the SDK sent them this way (similar to how envelope attachments' content types are untouched). Right now, Relay assigns `application/octet-stream` to incoming multipart attachments without a content type.

Are there currently any components that rely on the automatic content type assignment in Relay, rather than deferring to the MIME detector in the Python code?

olksdr:
Hey @supervacuus 

thanks for opening an issue and a PR. We will discuss this with the team and post the update here or on your PR.

supervacuus:
Since the associated PR has been merged and deployed and preliminary tests show that multipart attachments are correctly detected and thus can be previewed in the web UI, we can close this. Thanks for the support!

PR #3225, opened 2024-03-06
fix(server): pass on multipart attachments without content type

supervacuus:
This PR visualizes the change required to go forward with the approach suggested here:  #3224

I introduced that `set_payload()` interface asymmetry only to avoid changing all the call sites for this PR.


supervacuus:
A few integration tests fail due to assertions that expect `application/octet-stream` when no incoming content type was set. I can fix those, but whether this behavior should be changed is at the core of this issue/PR. Could you let me know if I should fix them anyway?

relay-server/src/envelope.rs
```diff
@@ -710,20 +710,31 @@ impl Item {
     }
 
     /// Sets the payload and content-type of this envelope.
-    pub fn set_payload<B>(&mut self, content_type: ContentType, payload: B)
-    where
+    pub(crate) fn set_payload_optional_content_type<B>(
```
iker-barriocanal: The original `set_payload` does two things: set the `content_type` and truncate the payload. In the PR, the content type is only set from `set_payload` but never directly. Splitting both actions makes the `Item`'s API usage easier to understand (vs wondering what a `None` is for). For example:

```rs
// envelope.rs

fn set_payload(...) {
  self.headers.content_type = content_type;
  self.set_truncated_payload(...);
}

fn set_truncated_payload<B>(&mut self, payload: B) where B: Into<Bytes> {
  // ...
}

// multipart.rs

if let Some(content_type) = field.content_type() {
  item.set_payload(content_type, field_data);
} else {
  item.set_truncated_payload(field_data);
}
```

(not opinionated on the function name.)

Extending the docstrings to point to the other function, depending on the `content_type` needs, would be great.

relay-server/src/envelope.rs
```diff
@@ -710,20 +710,31 @@ impl Item {
...
+        let length = std::cmp::min(u32::MAX as usize, payload.len());
         payload.truncate(length);
 
         self.headers.length = Some(length as u32);
-        self.headers.content_type = Some(content_type);
+        self.headers.content_type = content_type;
         self.payload = payload;
     }
 
+    /// The primary payload setter requires a content-type, since most call-sites set it explicitly.
```
iker-barriocanal: The original docstring is more descriptive:
```suggestion
    /// Sets the payload and content-type of this envelope.
```

iker-barriocanal (changes requested):
Overall LGTM, and thanks for the context on the original issue!

Let's add some tests too, and it's good to go. (There are some tests using attachments in `relay-server/src/envelope.rs`, a similar structure is great.)

supervacuus:
> Let's add some tests too, and it's good to go. (There are some tests using attachments in `relay-server/src/envelope.rs`, a similar structure is great.)

Thanks, @iker-barriocanal.

To be clear, would you like to add attachment tests for `multipart.rs` analogous to the tests in `envelope.rs`? A test covering the `multipart_items()` content-type handling would be a good start. 

Is it okay to assert the attachment's `content-type` properties for failing integration tests as `None`?

iker-barriocanal:
> To be clear, would you like to add attachment tests for multipart.rs analogous to the tests in envelope.rs? A test covering the multipart_items() content-type handling would be a good start.

Yep, test coverage for both branches of `multipart_items` based on `content-type` is good enough IMO.

> Is it okay to assert the attachment's content-type properties for failing integration tests as None?

Yes, that's what we're looking for in the end -- no `content-type` should result in no value set, with no defaults.

Thanks for working on this, and let me know if you have further questions!

supervacuus:
Hi @iker-barriocanal, I tried to cover the mentioned issues. I used a "real-life" multipart payload for the unit test and minimized the contents to data required for the assertions. It should cover all PR-touched paths in `multipart_items()` (and more), but it is written from the minidump perspective, so I added the test there. Let me know if I should add assertions/payload items or split the payload into multiple tests.

relay-server/src/utils/multipart.rs
```diff
@@ -185,12 +185,15 @@ where
...
-            };
             // Extract the body after the immutable borrow on `file_name` is gone.
-            item.set_payload(content_type, field_data(&mut field, item_limit).await?);
+            if let Some(content_type) = field.content_type() {
+                item.set_payload(
+                    content_type.into(),
+                    field_data(&mut field, item_limit).await?,
+                );
+            } else {
+                item.set_truncated_payload(field_data(&mut field, item_limit).await?);
```
supervacuus: Nit: While the interface parameters make more sense, the name `set_payload()` vs `set_truncated_payload()` still seems weirdly asymmetric since both truncate... but I have no better name, except for using the verbose option of `set_payload_without_content_type()` (which was the name before I created the PR, and chose the intentionally tacky variant with the `None` option). I am open to suggestions; maybe verbosity is okay if only one call site exists.

tests/integration/test_minidump.py
```diff
@@ -422,7 +422,7 @@ def test_minidump_with_processing(
         {
             "id": attachment_id,
             "name": "minidump.dmp",
-            "content_type": "application/octet-stream",
+            "content_type": None,
```
iker-barriocanal: Q: Does `sentry` always expect the `content_type` field to exist?

Looking at [this code](https://github.com/getsentry/sentry/blob/42050bfa4f3fcc6b19c9f104eba868dcc8535eaa/src/sentry/attachments/base.py#L58) maybe it does, but I'm not familiar with that part of the codebase. If it's ok to not include the field on null values, let's skip serializing it by adding a `#[serde(skip_serializing_if = "Option::is_none")]` to `ChunkedAttachment.content_type` on `relay-server/src/services/store.rs`.

relay-server/src/endpoints/minidump.rs
```diff
@@ -185,4 +192,110 @@ mod tests {
         let garbage = b"xxxxxx";
         assert!(validate_minidump(garbage).is_err());
     }
+
+    #[tokio::test]
+    async fn test_minidump_multipart_attachments() -> anyhow::Result<()> {
+        let _body_data2 : &[u8] =
```
iker-barriocanal: nit: I think `_` isn't required:
```suggestion
        let body_data2 : &[u8] =
```

iker-barriocanal (commented):
Thanks for your effort and tests! One last question before merging 😄

relay-server/src/utils/multipart.rs
jjbayer: I would prefer `set_payload_without_content_type`, for what it's worth. `set_truncated_payload()` seems to imply that `set_payload` does not truncate, as you noted.

tests/integration/test_minidump.py
jjbayer: +1, sentry might expect the content type, we should make sure it can handle missing / `null` values.

relay-server/src/endpoints/minidump.rs
supervacuus: Sorry, that was a leftover from testing with various multipart bodies.

tests/integration/test_minidump.py
supervacuus: I am also not familiar with the code base. Still, when looking at the handling of attachments processed from envelope items (directly vs. via multipart), a non-existent content type appears to be valid input.

iker-barriocanal: Let me check internally and I'll get back to you

iker-barriocanal: It seems it's safe for `content_type` to be null or not exist (see [sentry test](https://github.com/getsentry/sentry/blob/6030fda66f27cfb3b4dfee3018a978fb9752a4c1/tests/sentry/api/endpoints/test_event_attachments.py#L28)), and sentry will also [default to `application/octet-stream`](https://github.com/getsentry/sentry/blob/dd0edf1fcc309765d94505f1bac758d7831dcbbf/src/sentry/models/eventattachment.py#L187) if it can't guess the type.

@supervacuus let's skip the serialization when the field is empty as I posted on my [comment above](https://github.com/getsentry/relay/pull/3225#discussion_r1530010648), and I'll merge the PR afterward.

iker-barriocanal (approved):
Thanks for putting in the effort and working on this!

I'll merge and deploy on Monday to avoid surprises on a Friday and weekend.

iker-barriocanal:
There were some difficulties in the infrastructure today and this PR is not live. I'll make another attempt tomorrow and post an update here when the change is live.

iker-barriocanal:
@supervacuus the PR is deployed!

supervacuus:
> @supervacuus the PR is deployed!

Thanks for the update. I tested this with a patched `crashpad_handler`; all attachments are correctly detected, and the preview works.

PR #3226, opened 2024-03-06
ref(server): Remove sampling_result from state

jjbayer:
Make `sampling_result` a plain old return value, to prevent bugs like the one fixed in https://github.com/getsentry/relay/pull/3186.

ref: https://github.com/getsentry/team-ingest/issues/299

#skip-changelog


olksdr (approved):

iker-barriocanal (approved):
Nice!!

PR #3229, opened 2024-03-06
feat(profiling): add namespace for profile metrics

viglia:
Sentry PR that added the `profles` use case: https://github.com/getsentry/sentry/pull/65152/files

### DIFF
--- a/relay-server/src/services/project.rs
+++ b/relay-server/src/services/project.rs
@@ -1246,6 +1246,7 @@ fn is_metric_namespace_valid(state: &ProjectState, namespace: &MetricNamespace) 
             state.has_feature(Feature::ExtractSpansAndSpanMetricsFromEvent)
                 || state.has_feature(Feature::StandaloneSpanIngestion)
         }
+        MetricNamespace::Profiles => true,
         MetricNamespace::Custom => state.has_feature(Feature::CustomMetrics),
         MetricNamespace::Unsupported => false,
     }
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -769,12 +769,6 @@ impl StoreService {
                     .options
                     .profiles_function_generic_metrics_enabled
                 {
-                    relay_log::with_scope(
-                        |scope| scope.set_extra("metric_message.name", message.name.into()),
-                        || {
-                            relay_log::error!("store service dropping profiles metric usecase: currently disabled")
-                        },
-                    );
                     return Ok(());
                 }
                 KafkaTopic::MetricsGeneric
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -762,6 +762,23 @@ impl StoreService {
                 );
                 return Ok(());
             }
+            MetricNamespace::Profiles => {
+                if !self
+                    .global_config
+                    .current()
+                    .options
+                    .profiles_function_generic_metrics_enabled
+                {
+                    relay_log::with_scope(
+                        |scope| scope.set_extra("metric_message.name", message.name.into()),
+                        || {
+                            relay_log::error!("store service dropping profiles metric usecase: currently disabled")
+                        },
+                    );
+                    return Ok(());
+                }
+                KafkaTopic::MetricsGeneric
+            }
             _ => KafkaTopic::MetricsGeneric,
         };
         let headers = BTreeMap::from([("namespace".to_string(), namespace.to_string())]);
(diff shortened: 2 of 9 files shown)


jjbayer (approved):
This might be worth adding an integration test, like this one:

https://github.com/getsentry/relay/blob/3dff617a1a5ff01f6617de82807bc50c7d695a2c/tests/integration/test_metrics.py#L169-L204

Dav1dde (changes requested):
We need to make sure every other piece of the pipeline already supports the namespace!

Are Sentry and the consumers already able to process the profile metrics without crashing?

viglia:
@Dav1dde I followed this [doc](https://www.notion.so/sentry/Creating-a-New-Use-Case-ID-for-the-Generic-Metric-Pipeline-Locally-d9bdeaac34c14ea88f1ff0f9dbd6e90a) and added the use case as suggested ~2 weeks ago in this [PR](https://github.com/getsentry/sentry/pull/65152/files)

ps. in the doc there is no mention about other consumer-specific changes. If you're aware of other changes that are documented somewhere else, please, link them here and I'll follow up.

Dav1dde:
@viglia I am just a bit paranoid, when we add it to Relay this means anyone can ingest these messages on any of our instances.

I don't know what needs to be added to Sentry, but it seems like there were no changes made to rate and cardinality limits, which eventually should be added.

relay-server/src/services/project.rs
```diff
@@ -1246,6 +1246,7 @@ fn is_metric_namespace_valid(state: &ProjectState, namespace: &MetricNamespace)
             state.has_feature(Feature::ExtractSpansAndSpanMetricsFromEvent)
                 || state.has_feature(Feature::StandaloneSpanIngestion)
         }
+        MetricNamespace::Profiles => true,
```
Dav1dde, line 1249: I'd be more comfortable if the profiles were gated by a feature flag, a feature flag we can only enable in S4S first.

iker-barriocanal (approved):
Could we link in the description to the PRs/code where the namespace has been added in sentry/snuba/etc? Thanks beforehand!

relay-server/src/services/project.rs
viglia: @Dav1dde I've added the kill-switch option instead, as we discussed offline.

viglia:
> @viglia I am just a bit paranoid, when we add it to Relay this means anyone can ingest these messages on any of our instances.
> 
> I don't know what needs to be added to Sentry, but it seems like there were no changes made to rate and cardinality limits, which eventually should be added.

@Dav1dde keep in mind that setting those limits is optional.

```
By default, a database writes quota will be provisioned by the following sentry options. 
They will be shared amongst all use case IDs without a dedicated write limit option name.

sentry-metrics.writes-limiter.limits.generic-metrics.global
sentry-metrics.writes-limiter.limits.generic-metrics.per-orgCardinality limit
```

So we have a default rate limiting value, both for [global](https://github.com/getsentry/sentry-options-automator/blob/main/options/regions/us/genericmetrics.yaml#L30) and for the [per-org](https://github.com/getsentry/sentry-options-automator/blob/main/options/regions/us/genericmetrics.yaml#L14)

Should we need customized values, we'll only find out with time and can make those changes to enforce different limits then.

relay-server/src/services/store.rs
```diff
@@ -762,6 +762,23 @@ impl StoreService {
...
+                    .current()
+                    .options
+                    .profiles_function_generic_metrics_enabled
+                {
+                    relay_log::with_scope(
+                        |scope| scope.set_extra("metric_message.name", message.name.into()),
+                        || {
+                            relay_log::error!("store service dropping profiles metric usecase: currently disabled")
+                        },
+                    );
```
Dav1dde: This error is too much volume for a case we expect to happen.

Dav1dde (approved):
> @Dav1dde keep in mind that setting those limits is optional.

Maybe, but these settings exist to protect our infrastructure. And in the incident case the necessary code in Sentry is missing to even configure the limits.

These two configs you pointed out are not the correct ones.

I urge you to follow up on these things, this is incident territory.

I don't want to block you, the Relay changes are fine the, just please drop the error.
The Rest I don't want to worry about.

viglia:
> Maybe, but these settings exist to protect our infrastructure. And in the incident case the necessary code in Sentry is missing to even configure the limits.

yes, that's why there are default cardinality & write limits that are shared among use cases that don't specify a custom one.


> These two configs you pointed out are not the correct ones.

Before merging I'll follow up with `sns` to double check everything, but those options I've shared come directly from the Doc that outlines how to add a new use case --> https://www.notion.so/sentry/Creating-a-New-Use-Case-ID-for-the-Generic-Metric-Pipeline-Locally-d9bdeaac34c14ea88f1ff0f9dbd6e90a?pvs=4#e0e24ab1731340c0aef682eefe8112b3

PR #3231, opened 2024-03-07
feat(inbound filters): Add cyptocurrency "cannot redefine property" to browser extension filter

scttcper:
These cryptocurrencies commonly have wallets that inject javascript into the page making noise. See the potential number of projects affected in redash https://redash.getsentry.net/queries/6220/source

or an issue within our own projects https://sentry.sentry.io/issues/4724125675/

#skip-changelog

### DIFF
--- a/relay-filter/src/browser_extensions.rs
+++ b/relay-filter/src/browser_extensions.rs
@@ -43,7 +43,9 @@ static EXTENSION_EXC_VALUES: Lazy<Regex> = Lazy::new(|| {
         webkit-masked-url:|
         # Firefox message when an extension tries to modify a no-longer-existing DOM node
         # See https://blog.mozilla.org/addons/2012/09/12/what-does-cant-access-dead-object-mean/
-        can't\saccess\sdead\sobject
+        can't\saccess\sdead\sobject|
+        # Crypocurrency related extension errors
+        Cannot\sredefine\sproperty:\s(solana|ethereum)
     "#,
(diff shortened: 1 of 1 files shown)


jjbayer (approved):

Dav1dde (approved):

relay-filter/src/browser_extensions.rs
```diff
@@ -43,7 +43,9 @@ static EXTENSION_EXC_VALUES: Lazy<Regex> = Lazy::new(|| {
         webkit-masked-url:|
         # Firefox message when an extension tries to modify a no-longer-existing DOM node
         # See https://blog.mozilla.org/addons/2012/09/12/what-does-cant-access-dead-object-mean/
-        can't\saccess\sdead\sobject
+        can't\saccess\sdead\sobject|
+        # Crypocurrency related extension errors
+        Cannot\sredefine\sproperty:\s(solana|ethereum)
```
iker-barriocanal, line 48: Should we include `*wallet*` matches too? Not as many matches with `"%Cannot redefine property:%wallet%"`, but it's another improvement.

iker-barriocanal (approved):

relay-filter/src/browser_extensions.rs
scttcper: Let me follow up with the team, i think i'd like to continue matching very specific words like okxwallet

Issue #1361, opened 2022-07-29
Relay connect to kafka with ssl auth


bumarcell:
We are using Sentry on-premise on Kubernetes ([this helm chart](https://github.com/sentry-kubernetes/charts)) with external Kafka. Our production Kafka uses ssl authentication, which I think is a common practice.

I noticed this feature was requested (and implemented) 2 years ago [here](https://github.com/getsentry/relay/issues/879) however it was reverted due to some regression problem (and the feature being regarded as an "edge case")  

Would it still be problematic to implement it with the current Sentry/Relay version 2 years after?

untitaker:
@bumarcell feel free to come up with a solution that solves the concern of not being able to use `--all-features`: https://github.com/getsentry/relay/pull/889

The concrete problem: We want to be able to run `cargo build --all-features` on macos, which is our primary development platform. The initial implementation broke that because the new `kafka-ssl` feature didn't compile on macos (iirc). I don't think we would object to additional, mandatory dependencies for as long as they're easy to manage (i.e.: they compile on macos).

We're calling it an "edge case" in that we're not going to try and spend more time on implementing this but I don't think we'd reject a patch that reintroduces this with those problems fixed.

PR #3232, opened 2024-03-07
feat(server): Kafka SSL support

jjbayer:
Previous attempts of compiling SSL support for Kafka failed on macOS (see [comment](https://github.com/getsentry/relay/issues/1361#issuecomment-1199120293)). Now it seems to work.

### DIFF
--- a/relay-kafka/Cargo.toml
+++ b/relay-kafka/Cargo.toml
@@ -10,7 +10,7 @@ license-file = "../LICENSE.md"
 publish = false
 
 [dependencies]
-rdkafka = { version = "0.29.0", optional = true, features = ["tracing"] }
+rdkafka = { version = "0.29.0", optional = true, features = ["tracing", "ssl"] }
 rdkafka-sys = { version = "4.3.0", optional = true }
 relay-log = { path = "../relay-log", optional = true }
 relay-statsd = { path = "../relay-statsd", optional = true }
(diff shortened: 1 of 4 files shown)


relay-kafka/Cargo.toml
```diff
@@ -10,7 +10,7 @@ license-file = "../LICENSE.md"
 publish = false
 
 [dependencies]
-rdkafka = { version = "0.29.0", optional = true, features = ["tracing"] }
+rdkafka = { version = "0.29.0", optional = true, features = ["tracing", "ssl"] }
```
jjbayer, line 13: Not sure if we want to add this feature unconditionally. But since `rdkafka` is already optional itself, it does not matter much IMO.

tests/integration/test_store.py
```diff
@@ -2297,3 +2297,9 @@ def test_span_ingestion_with_performance_scores(
             },
         },
     ]
+
+
+def test_kafka_ssl(relay_with_processing):
+    relay_with_processing(
+        options={"kafka_config": [{"name": "ssl.key.password", "value": "foo"}]}
+    )
```
jjbayer, line 2305: On master, Relay would fail with error

> Configuration property "ssl.key.password" not supported in this build: OpenSSL not available at build time

olksdr (approved):
If we sure nothing breaks by enabling ssl support in existing setups, it's LGTM 

Maybe it would be also great to add some tests to use SSL cert to connect to Kafka, which would require more work on CI side and we could follow up on later.

Dav1dde (approved):
👍 builds for me as well

PR #3235, opened 2024-03-07
ref(bucket-encoding): Tag kafka metrics with metric type and encoding

Dav1dde:
#skip-changelog

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -1439,8 +1439,8 @@ impl<'a> MetricValue<'a> {
 
     fn encoding(&self) -> Option<&'static str> {
         match self {
-            Self::Distribution(ae) => Some(ae.encoding()),
-            Self::Set(ae) => Some(ae.encoding()),
+            Self::Distribution(ae) => Some(ae.name()),
+            Self::Set(ae) => Some(ae.name()),
             _ => None,
         }
     }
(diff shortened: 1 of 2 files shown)


relay-server/src/services/store.rs
```diff
@@ -1438,6 +1467,13 @@ impl<T> ArrayEncoding<T> {
             MetricEncoding::Array => Self::Dynamic(DynamicArrayEncoding::Array { data }),
         }
     }
+
+    fn encoding(&self) -> &'static str {
```
jjbayer: nit: 
```suggestion
    fn name(&self) -> &'static str {
```

relay-server/src/services/store.rs
```diff
@@ -494,6 +494,16 @@ impl StoreService {
...
 
         match &message {
+            KafkaMessage::Metric {
+                message: metric, ..
+            } => {
+                metric!(
+                    counter(RelayCounters::ProcessingMessageProduced) += 1,
+                    event_type = message.variant(),
+                    metric_type = metric.value.variant(),
+                    metric_encoding = metric.value.encoding().unwrap_or(""),
```
jjbayer, line 504: Could you document these here? https://github.com/getsentry/relay/blob/3dff617a1a5ff01f6617de82807bc50c7d695a2c/relay-server/src/statsd.rs#L557-L562

Also curious that we documented a `namespace` tag there which we apparently dropped.

jjbayer (approved):

Issue #3079, opened 2024-02-09
Rate limit span items


jjbayer:
Code hint: https://github.com/getsentry/relay/blob/d3864a0633506ca5afe4b1d8462d0933e18634f6/relay-server/src/utils/rate_limits.rs#L24

PR #3238, opened 2024-03-07
feat(server): Enforce rate limits for standalone spans

jjbayer:
Rate limit `Span` and `OtelSpan` items when a rate limit for either `Span` or `SpanIndexed` data categories exist.

This PR does **not** yet rate limit

* span _metrics_, which I will implement in a separate PR.
* Quota enforcement on spans extracted from transactions (requires https://github.com/getsentry/relay/pull/3248).

ref: https://github.com/getsentry/relay/issues/3079


relay-server/src/services/outcome_aggregator.rs
```diff
@@ -155,7 +151,6 @@ impl OutcomeAggregator {
                 quantity,
             };
 
-            relay_log::trace!("Flushing outcome for timestamp {timestamp}");
```
jjbayer, line 158: I removed these because they produce a lot of noise when integration testing locally.

relay-server/src/utils/rate_limits.rs
```diff
@@ -474,7 +496,7 @@ where
     ///    are exhausted, both the event and metrics are dropped.
     ///  - `DataCategory::TransactionIndexed` counts ingested and stored events. If quotas with this
     ///    category are exhausted, just the event payload is dropped, but metrics are kept.
-    fn index_category(&self, category: DataCategory) -> Option<DataCategory> {
+    fn transaction_index_category(&self, category: DataCategory) -> Option<DataCategory> {
```
jjbayer, line 498: This function explicitly checks for `DataType::Transaction`, so I renamed it to make its purpose more clear.

relay-server/src/services/outcome_aggregator.rs
iker-barriocanal: I think a log for ack-ing outcomes are doing something is valuable, but not strongly opinionated.

Dav1dde: They are trace logs after all (fine by me though)

relay-server/src/utils/rate_limits.rs
```diff
@@ -153,12 +153,18 @@ pub struct EnvelopeSummary {
...
+    pub span_quantity: usize,
+
     /// Indicates that the envelope contains regular attachments that do not create event payloads.
     pub has_plain_attachments: bool,
 
     /// Whether the envelope contains an event which already had the metrics extracted.
     pub event_metrics_extracted: bool,
 
+    /// Whether the envelope contains spans which already had metrics extracted.
+    pub span_metrics_extracted: bool,
```
Dav1dde, line 166: Why do we need this for spans but not transactions?

relay-server/src/services/processor.rs
```diff
@@ -1431,19 +1429,21 @@ impl EnvelopeProcessorService {
...
     ) -> Result<(), ProcessingError> {
         span::filter(state);
         if_processing!(self.inner.config, {
-            self.enforce_quotas(state)?;
             span::process(
                 state,
                 self.inner.config.clone(),
                 &self.inner.global_config.current(),
             );
+            self.enforce_quotas(state)?;
```
iker-barriocanal, line 1446: Should we extract metrics from spans that are rate-limited? Do span metrics only depend on span metrics rate limits?

relay-server/src/utils/rate_limits.rs
```diff
@@ -610,6 +632,34 @@ where
...
+            // quota. Quota will be consumed by the metrics rate limiter instead.
+            let limits = (self.check)(scoping.item(DataCategory::Span), 0)?;
+            let longest = limits.longest();
+
+            // Only enforce and record an outcome if metrics haven't been extracted yet.
+            // Otherwise, the outcome is logged by the metrics rate limiter.
+            if !summary.span_metrics_extracted {
+                enforcement.span_metrics =
+                    CategoryLimit::new(DataCategory::Span, summary.span_quantity, longest);
+            } else if longest.is_none() {
```
Dav1dde: I find the else-if/else harder to reason about then the transactions block, which seems to have the same logic overall?

Would be nice if you could have transactions and spans use the same controlflow, whether it be this one or the one transactions already use.

relay-server/src/utils/rate_limits.rs
```diff
@@ -610,6 +632,34 @@ where
...
+                // Metrics were extracted and aren't rate limited. Check if there
+                // is a separate rate limit for indexed spans:
```
iker-barriocanal, line 651: Is it possible that metrics aren't rate-limited yet but metrics have been extracted? How does this use case impact the application of rate limits on the metrics rate limiter and span rate limits?

Example use case: rate limits were extracted in a downstream relay but not flushed yet, and this relay only received the span.

relay-server/src/utils/rate_limits.rs
```diff
@@ -610,6 +632,34 @@ where
             rate_limits.merge(checkin_limits);
         }
 
+        if summary.span_quantity > 0 {
+            // Check for rate limits on the main category but do not consume
+            // quota. Quota will be consumed by the metrics rate limiter instead.
```
iker-barriocanal, line 636: Not sure I get this -- metrics rate limiter consumes spans quota? Is this the planned change in the PR description to rate-limit metrics?

relay-server/src/services/outcome_aggregator.rs
jjbayer: Fair points! We do still have the log below so I figured one is enough, now that outcomes aggregation has been stable for a long time. https://github.com/getsentry/relay/blob/0adcea4438bc22adf517f54f17501324eb68a4b7/relay-server/src/services/outcome.rs#L815

relay-server/src/services/processor.rs
jjbayer: > Should we extract metrics from spans that are rate-limited?

If the rate limit is on `Span`, then no. If the rate limit is on `SpanIndexed` only, then yes. Just like transactions, most rate limited spans will be caught on the fast path -- this call makes sure that we count up in redis.

> Do span metrics only depend on span metrics rate limits?

Yes, they are not rate limited by `SpanIndexed`. They might get limited by the metrics bucket rate limiter or the cardinality limiter though.

relay-server/src/utils/rate_limits.rs
jjbayer: The `event_metrics_extracted` field above tracks the same thing for transactions.

relay-server/src/utils/rate_limits.rs
```diff
@@ -194,6 +205,7 @@ impl EnvelopeSummary {
             if let Some(source_quantities) = item.source_quantities() {
                 summary.secondary_transaction_quantity += source_quantities.transactions;
                 summary.profile_quantity += source_quantities.profiles;
+                // TODO: span metrics quantity
```
jjbayer: Resolve TODO.

relay-server/src/utils/rate_limits.rs
jjbayer: Yes. Calling `self.check` with a number > 0 would increment the redis counter. But we do not want to increment the redis counter for `DataCategory::Span` here, because it will be incremented by the bucket containing span metrics.

We do the same for transactions:

https://github.com/getsentry/relay/blob/e2c7729ef072c79a2623ad57f181a61687bccb09/relay-server/src/utils/rate_limits.rs#L523-L527

relay-server/src/utils/rate_limits.rs
jjbayer: Thanks, I aligned it with events now, and this actually helped find an inconsistency! I also copied some unit tests that already existed for spans.

relay-server/src/utils/rate_limits.rs
jjbayer: In that case, because there is a rate limit on the main category, the span payload will still be dropped.

relay-server/src/utils/rate_limits.rs
Dav1dde: Shouldn't we also use the same field for spans?

Dav1dde: Shouldn't we also use the same field for spans?

(I am fine if we want to use a new field to make it easier to understand the code, but then we should also update the transactions code)

Dav1dde (approved):
LGTM, but I only really compared logic to what already exists for transactions 👍

iker-barriocanal (approved):

PR #3244, opened 2024-03-08
test(mobile-score): Add test documenting mobile score profile

narsaynorath:
Adds a test documenting a mobile score profile we want to experiment
with. All fields will be optional for now until we figure out the
combinations of measurements that are sent together.

The next step will be to add a PR that tags these scores with a score
profile name that we can use for versioning while prototyping or making
changes to perf score profile weights.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -1109,6 +1109,98 @@ mod tests {
         RawUserAgentInfo,
     };
 
+    const IOS_MOBILE_EVENT: &str = r#"
+        {
+            "type": "transaction",
+            "sdk": {"name": "sentry.cocoa"},
+            "start_timestamp": "2021-04-26T07:59:01+0100",
+            "timestamp": "2021-04-26T08:00:00+0100",
+            "release": "1.2.3",
+            "transaction": "gEt /api/:version/users/",
+            "transaction_info": {"source": "custom"},
+            "platform": "cocoa",
+            "contexts": {
+                "trace": {
+                    "trace_id": "ff62a8b040f340bda5d830223def1d81",
+                    "span_id": "bd429c44b67a3eb4",
+                    "op": "ui.load"
+                },
+                "device": {
+                    "family": "iOS",
+                    "model": "iPhone14,7"
+                },
+                "app": {
+                    "app_identifier": "io.sentry.sample.iOS-Swift",
+                    "app_name": "iOS-Swift"
+                },
+                "os": {
+                    "name": "iOS",
+                    "version": "16.4"
+                }
+            },
+            "measurements": {
+                "app_start_warm": {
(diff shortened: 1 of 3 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -2650,6 +2872,457 @@ mod tests {
...
+                            }
+                        ]
+                    }
+                }
+            ]
+        }))
+        .unwrap();
+
+        normalize_performance_score(&mut event, Some(&performance_score));
+        println!("{:#?}", event);
```
iker-barriocanal:
```suggestion
```

relay-event-normalization/src/event.rs
```diff
@@ -2650,6 +2872,457 @@ mod tests {
...
```
iker-barriocanal:
```suggestion
```

iker-barriocanal (commented):
Thanks for adding tests! Are the events the minimal payloads to validate the mobile scores? If not, could we reduce them to make them minimal? A minimal payload eases the understanding of a test when it fails.

On a side note, could we update the type of commit in the PR, according to the [guidelines](https://develop.sentry.dev/commit-messages/#type)?

relay-event-normalization/src/event.rs
```diff
@@ -2650,6 +2872,457 @@ mod tests {
...
+
+        insta::assert_ron_snapshot!(SerializableAnnotated(&Annotated::new(event)), {}, @r###"
```
jjbayer: nit: This snapshot is so large that it might be worth moving it to a separate file (which `insta` can do for you automatically):

```suggestion
        insta::assert_ron_snapshot!(SerializableAnnotated(&Annotated::new(event)), {});
```

relay-event-normalization/src/event.rs
narsaynorath: Oops, thanks for catching this. It was leftover from some debugging

relay-event-normalization/src/event.rs
```diff
@@ -1109,6 +1109,80 @@ mod tests {
...
+                },
+                "time_to_initial_display": {
+                    "value": 8049.345970153808,
+                    "unit": "millisecond"
+                }
+            }
+        }
+        "#;
+
+    const ANDROID_MOBILE_EVENT: &str = r#"
```
iker-barriocanal, line 1137: The iOS event has 3 measurements. Does the minimal Android event need 9 measurements, or can we reduce it?

iker-barriocanal (approved):

relay-event-normalization/src/event.rs
narsaynorath: Good point, I removed a number of the measurements. I had them in because they were measurements from a real Sentry event, but they aren't necessary for the minimal test input.

PR #3245, opened 2024-03-08
feat(perf): Extract `span.status_code` tag for HTTP spans

gggritso:
We need this to do status code breakdowns for the HTTP module. Note: The HTTP module, as designed, doesn't need to store _the full code_, but rather just the first digit. We break down by 2xx, 4xx, and 5xx codes. That said, it seems simpler and more robust to just extract all the response codes, since in practice we don't expect more than half a dozen codes to actually come in per span.

I added a banal test, but this is my first PR in Relay so please save me from myself 🙏🏻 let me know what you need in terms of code changes and/or test coverage.


relay-dynamic-config/src/defaults.rs
```diff
@@ -260,6 +264,10 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
                 Tag::with_key("resource.render_blocking_status")
                     .from_field("span.sentry_tags.resource.render_blocking_status")
                     .when(is_resource.clone()),
+                // HTTP module:
+                Tag::with_key("span.status_code")
+                    .from_field("span.sentry_tags.status_code")
```
iker-barriocanal, line 269: Do we need the tag in `d:spans/exclusive_time_light@millisecond` too?

iker-barriocanal (approved):
LGTM, thanks for contributing! Could we add a changelog entry?

relay-dynamic-config/src/defaults.rs
```diff
@@ -209,6 +209,10 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
...
+                    .always(),
```
```suggestion
                    .when(is_http.clone()),
```

relay-dynamic-config/src/defaults.rs
jjbayer: Good point.

relay-dynamic-config/src/defaults.rs
```diff
@@ -260,6 +264,10 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
...
```
```suggestion
                    .when(is_http.clone()),
```

Dav1dde:
> That said, it seems simpler and more robust to just extract all the response codes, since in practice we don't expect more than half a dozen codes to actually come in per span.

Note this is worst case a multiplicator of 12x on the entire cardinality, it might seem low, but every additional tag is a multiplicator of the existing cardinality, so if there is no use-case for it, I would consider just going with the broader category of `1xx, 2xx, 3xx, 4xx, 5xx`

relay-dynamic-config/src/defaults.rs
gggritso: Unsure what you mean—from what I can see I added the tag to that metric, did I do something wrong?

relay-dynamic-config/src/defaults.rs
gggritso: Updated, thanks!

relay-dynamic-config/src/defaults.rs
gggritso: Done 👍🏻

gggritso:
@iker-barriocanal updated changelog 👍🏻 
@Dav1dde fair point thank you, and I'm aware of the cardinality increase but we're _already_ getting feature requests to filter by the exact response code, so I think falling back to broader category would be temporary at best

Dav1dde:
Fair enough!

relay-dynamic-config/src/defaults.rs
iker-barriocanal: The tag is added to two metrics: `d:spans/exclusive_time@millisecond` and `d:spans/exclusive_time_light@millisecond`. Do we need the tag in both metrics, or just one of the two metrics is good enough?

gggritso: Oh I see! I need it for both, since the `light` metric is only used on the landing page, and the regular one is used on the domain landing pages, where we have to break down by transaction

iker-barriocanal: Sounds good! :shipit:

jjbayer (approved):

PR #3248, opened 2024-03-11
feat(server): Scrub transactions before enforcing quotas

jjbayer:
We previously enforced quotas before PII scrubbing because rate limiting did not depend on having PII-scrubbed data.

Now that we want to rate limit spans that are extracted from transactions, which in turn depend on PII-scrubbing, we need to turn the order around.

Note that this should barely increase cost, because the vast majority of rate limited events is caught by cached rate limits on the fast path, not the consistent rate limiting step in the processor (which is mainly there for incrementing the redis counter).


tests/integration/test_spans.py
```diff
@@ -1101,6 +1101,70 @@ def summarize_outcomes():
...
+    relay.send_event(project_id, event)
+    spans = list(spans_consumer.get_spans(max_attempts=2, timeout=10))
+    assert len(spans) == 2  # one for the transaction, one for the contained span
+    assert summarize_outcomes() == {(16, 0): 2}  # SpanIndexed, Accepted
+
+    # Second send should be rejected immediately.
+    relay.send_event(project_id, event)
+    spans = list(spans_consumer.get_spans(max_attempts=1, timeout=2))
+    assert len(spans) == 0  # all rejected
+    assert summarize_outcomes() == {(16, 2): 2}  # SpanIndexed, RateLimited
```
jjbayer, line 1162: These assertions fail without the change in `process_transactions`.

Dav1dde (approved):

PR #3249, opened 2024-03-11
feat(mobile-score): Add profile version as tag

narsaynorath:
Add a `version` property to perf score profiles so we can filter and separate changes to the profiles. The `version` gets assigned as `sentry.score_profile_version` in a transaction's tags. This would be useful if we wanted to hard cut over to a new version instead of blending users in, or just for creating test bed profiles. In our case, we want to send profiles with `version: mobile.alpha` so we can iterate on profiles but avoid polluting the end user results.

This should have low cardinality because we would only ever expect a few types of profile versions to be in the while at a time.

This adds a new string to the metrics indexer. The corresponding PR in `sentry` is: https://github.com/getsentry/sentry/pull/66795


relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -27,11 +27,12 @@ const PLACEHOLDER_UNPARAMETERIZED: &str = "<< unparameterized >>";
 /// Tags we set on metrics for performance score measurements (e.g. `score.lcp.weight`).
 ///
 /// These are a subset of "universal" tags.
-const PERFORMANCE_SCORE_TAGS: [CommonTag; 6] = [
+const PERFORMANCE_SCORE_TAGS: [CommonTag; 7] = [
     CommonTag::BrowserName,
     CommonTag::Environment,
     CommonTag::GeoCountryCode,
     CommonTag::Release,
+    CommonTag::ScoreProfileVersion,
```
narsaynorath: I'm not entirely sure if this should be considered a `CommonTag`. It's not used for every transaction. What is the definition of `"universal" tags` being used here? Is there a better alternative to define this if I want to isolate this tag from the universal ones?

relay-event-normalization/src/event.rs
```diff
@@ -818,6 +818,16 @@ pub fn normalize_performance_score(
                     );
                 }
                 if should_add_total {
+                    if profile.version.is_some() {
```
jjbayer: With this, you don't need the `.unwrap()`:

```suggestion
                    if let Some(version) = profile.version {
```

relay-event-normalization/src/event.rs
```diff
@@ -818,6 +818,16 @@ pub fn normalize_performance_score(
...
+                        event
+                            .tags
+                            .value_mut()
+                            .get_or_insert_with(Tags::default)
+                            .push(Annotated::new(TagEntry(
+                                Annotated::new("sentry.score_profile_version".to_string()),
+                                Annotated::new(profile.version.as_ref().unwrap().to_string()),
```
```suggestion
                                Annotated::new(version.clone()),
```

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -725,6 +735,7 @@ mod tests {
                     "environment": "fake_environment",
                     "geo.country_code": "US",
                     "release": "1.2.3",
+                    "sentry.score_profile_version": "alpha",
```
jjbayer, line 739: Please make sure to register the new tag name [here](https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py) before merging this PR.

relay-event-normalization/src/event.rs
```diff
@@ -818,6 +818,16 @@ pub fn normalize_performance_score(
                     );
                 }
                 if should_add_total {
+                    if profile.version.is_some() {
+                        event
+                            .tags
```
jjbayer: Do you actually need this as a tag on the event payload in the product? Or would a context or an entry in `event.data` be sufficient?

relay-server/src/metrics_extraction/transactions/mod.rs
jjbayer: It would be slightly better to declare this as an optional field on `TransactionMeasurementTags`, just like `measurement_rating`. You then just need to make sure that it gets converted in https://github.com/getsentry/relay/blob/154516c08de16df0aef96e29c449e3f5eefb6d26/relay-server/src/metrics_extraction/transactions/types.rs#L183

jjbayer (commented):
Looks largely OK to me, see comments.

relay-server/src/metrics_extraction/transactions/mod.rs
narsaynorath: Added it here: https://github.com/getsentry/sentry/pull/66795

relay-event-normalization/src/event.rs
narsaynorath: I thought about the pros and cons and while I would prefer the tag not to appear to the user in the UI, we still want it to be usable as a filter which means that event context or `event.data` aren't viable options, correct me if I'm wrong.

I can see a use-case where we want to filter the sampled events to get only ones for a specific score version since it doesn't get recalculated on new profiles.

Are there any other large cons to consider when adding this as a tag? If this is something we can easily change in the future, then I'm okay with keeping it as a tag if you agree.

relay-server/src/metrics_extraction/transactions/mod.rs
narsaynorath: This is great, thanks! Added.

relay-server/src/metrics_extraction/transactions/mod.rs
iker-barriocanal: Could you link that PR from the description? Thanks beforehand!

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -322,6 +322,13 @@ impl TransactionExtractor<'_> {
                     } else {
                         tags.clone()
                     },
+                    score_profile_version: if is_performance_score {
+                        event
+                            .tag_value("sentry.score_profile_version")
+                            .map(|version| version.to_string())
+                    } else {
+                        None
+                    },
```
iker-barriocanal: Not sure if we can simplify this, but maybe something like this works:
```suggestion
                    score_profile_version: is_performance_score.then(|| event.tag_value("sentry.score_profile_version")).map(|version| version.to_string())
```

relay-event-normalization/src/event.rs
```diff
@@ -2852,6 +2862,97 @@ mod tests {
...
+    fn test_computes_performance_score_and_tags_with_profile_version() {
+        let json = r#"
+        {
+            "type": "transaction",
+            "timestamp": "2021-04-26T08:00:05+0100",
+            "start_timestamp": "2021-04-26T08:00:00+0100",
+            "measurements": {
+                "cls": {"value": 0.11},
+                "inp": {"value": 120.0}
+            }
```
iker-barriocanal: Does the minimal test require two measurements?

relay-event-normalization/src/event.rs
jjbayer: > I would prefer the tag not to appear to the user in the UI, we still want it to be usable as a filter

I think using tags for this purpose is discouraged, but I cannot find any docs on it right now. Could you please consult with the Search team on this? AFAIK contexts are indexed.

AbhiPrasad: contexts are not indexed by default, we instead promote certain contexts fields to be indexed/searchable or turn them into tags. For example with browser name/version from browser context.

if the product requires it to be searchable, then having it be a context that is promoted to be a tag makes most sense to me.

jjbayer: @AbhiPrasad thanks, could you point me to the code that makes context fields searchable, or any docs around it? I remember seeing it at some point but was not able to find it just now.

AbhiPrasad: We have https://www.notion.so/sentry/A-guide-on-how-to-make-something-searchable-in-Discover-2f74c50fbbbe47988e723c59733307b8 - but not great docs outside of that. I think we need to prioritize getting something written down for this.

relay-event-normalization/src/event.rs
narsaynorath: I've modified this to be a bit more minimal. Now there's only one profile and just a single measurement to drive a calculation result.

https://github.com/getsentry/relay/pull/3249/commits/834aa616c79c48e9e2744774c648f330ab9215d2

relay-server/src/metrics_extraction/transactions/mod.rs
narsaynorath: https://github.com/getsentry/relay/pull/3249/commits/451678d1412c22187bdd978450035a0226db4e46
I had to do something slightly different but your suggestion helps make this cleaner, thanks 👍

jjbayer (approved):

relay-server/src/metrics_extraction/transactions/mod.rs
```diff
@@ -322,13 +323,15 @@ impl TransactionExtractor<'_> {
...
-                    },
+                    score_profile_version: is_performance_score
+                        .then(|| {
+                            event
+                                .context::<PerformanceScoreContext>()
+                                .unwrap()
+                                .score_profile_version
+                                .value()
+                        })
+                        .and_then(|version| version.map(|s| s.to_string())),
```
iker-barriocanal: If there's no `PerformanceScoreContext` this panics. Something like the following should work:
```suggestion
                    score_profile_version: is_performance_score
                        .then(|| {
                            event.context::<PerformanceScoreContext>()
                        })
                        .map(|context| context.score_profile_version.value())
                        .and_then(|version| version.map(|s| s.to_string())),
```

jjbayer: @iker-barriocanal thanks for catching this!

jjbayer (changes requested):

relay-server/src/metrics_extraction/transactions/mod.rs
narsaynorath: @iker-barriocanal really appreciate you catching that! I see now that the use of `.unwrap()` is dangerous. When I put in your code snippet I get this error:

```rust
error[E0609]: no field `score_profile_version` on type `std::option::Option<&relay_event_schema::protocol::PerformanceScoreContext>`
   --> relay-server/src/metrics_extraction/transactions/mod.rs:328:48
    |
328 |                         .map(|context| context.score_profile_version.value())
    |                                                ^^^^^^^^^^^^^^^^^^^^^ unknown field
    |
help: one of the expressions' fields has a field of the same name
    |
328 |                         .map(|context| context.unwrap().score_profile_version.value())
    |                                                +++++++++
```

~In this case is it safe to use `unwrap` because it's in the context of `.and_then`?~ Edit: This isn't the case. I made a change to try to use `and_then`, but I'm still getting the error with the code snippet you provided.

narsaynorath: This code snippet seems to work, however I feel like the last line feels like it's unnecessary/could be improved. Thoughts on this?

```rust
score_profile_version: is_performance_score
    .then(|| event.context::<PerformanceScoreContext>())
    .and_then(|context| context.unwrap().score_profile_version.value())
    .and_then(|version| Some(version).cloned()),
```

Dav1dde:
```rs
score_profile_version: is_performance_score
    .then(|| event.context::<PerformanceScoreContext>())
    .and_then(|context| context.unwrap().score_profile_version.value())
    .map(|version| version.clone()),
```

Dav1dde: Untested without unwrap:

```rs
score_profile_version: is_performance_score
    .then(|| event.context::<PerformanceScoreContext>())
    .and_then(|context| context?.score_profile_version.value())
    .map(|version| version.clone()),
```

narsaynorath: Thanks @Dav1dde, one of the linting tools gave me a warning about using an explicit closure to clone the version, so I just moved the call into the `.and_then` argument. This looks like it works without warnings

```rust
score_profile_version: is_performance_score
    .then(|| event.context::<PerformanceScoreContext>())
    .and_then(|context| context?.score_profile_version.value().cloned()),
```

jjbayer (approved):

PR #3250, opened 2024-03-11
feat(ai): Add support for AI token span data

colin-sentry:
Upstream in SDKs (https://github.com/getsentry/sentry-python/pull/2791) we added strings like "ai.prompt_tokens.used"

This adds them as known data fields, otherwise they are removed as PII since most configurations consider any superstring of "token" to be a password.


relay-pii/src/regexes.rs
```diff
@@ -286,6 +286,6 @@ static US_SSN_REGEX: Lazy<Regex> = Lazy::new(|| {
 
 static PASSWORD_KEY_REGEX: Lazy<Regex> = Lazy::new(|| {
     Regex::new(
-        r"(?i)(password|secret|passwd|api_key|apikey|auth|credentials|mysql_pwd|privatekey|private_key|token)"
+        r"(?i)(password|secret|passwd|api_key|apikey|auth|credentials|mysql_pwd|privatekey|private_key|token([a-r,t-z]|\b))"
```
Dav1dde: I think this could be just `token[^s]`

Although this does not match nicely at all, "tokenfoo" would match "tokenf" instead of just "token"

olksdr (commented):
Also, please, add tests for expected behaviour.

iker-barriocanal (commented):
As a side note, could you reword the PR description to align with [Sentry's commit message guidelines](https://develop.sentry.dev/commit-messages/)? Thanks beforehand!

olksdr:
@colin-sentry  are there any plans to follow up on this ?

colin-sentry:
@olksdr I'm having a bit of trouble writing a test for this - the test included in this PR still passes if you revert the regex change.

phacops:
Do we need all the new fields on span data?

colin-sentry:
> Do we need all the new fields on span data?

Discussed, by adding them to span data we can avoid them being removed as PII

jjbayer (approved):

PR #3251, opened 2024-03-11
feat(inbound filters): Add iOS chrome translation filter

scttcper:
Filters out `TypeError: undefined is not an object (evaluating 'a.L')` errors that come from Chrome on iOS.

See this notion doc for more details https://www.notion.so/sentry/TypeError-undefined-is-not-an-object-evaluating-a-L-b7a96303a7aa4d2fbbfc18e0ddb1468b TLDR:
- Poorly grouped google translation related errors
- Very large number of these errors especially in projects that are static sites with mobile users using translation services
- The errors are not actionable

here's an example in our own projects https://sentry.sentry.io/issues/5053303600/

#skip-changelog

### DIFF
--- a/relay-filter/src/browser_extensions.rs
+++ b/relay-filter/src/browser_extensions.rs
@@ -51,6 +51,16 @@ static EXTENSION_EXC_VALUES: Lazy<Regex> = Lazy::new(|| {
     .expect("Invalid browser extensions filter (Exec Vals) Regex")
 });
 
+static IOS_EXC_VALUES: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(
+        r#"(?ix)
+        # Translation errors in Chrome on iOS
+        undefined\sis\snot\san\sobject\s\(evaluating\s'a.L'\)
+    "#,
+    )
+    .expect("Invalid browser extensions filter (iOS Vals) Regex")
+});
+
 static EXTENSION_EXC_SOURCES: Lazy<Regex> = Lazy::new(|| {
     Regex::new(
         r"(?ix)
@@ -82,6 +92,13 @@ pub fn matches(event: &Event) -> bool {
     if let Some(ex_val) = get_exception_value(event) {
         if EXTENSION_EXC_VALUES.is_match(ex_val) {
             return true;
+        }
+        // Filtering for iOS based chrome trasnlation errors
+        if let Some(os_name) = event.tag_value("os.name") {
+            // Using iOS instead of device.name because it can be iPad or iPhone
+            if os_name == "iOS" && IOS_EXC_VALUES.is_match(ex_val) {
+                return true;
+            }
         }
     }
     if let Some(ex_source) = get_exception_source(event) {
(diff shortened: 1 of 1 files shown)


scttcper:
I'm new to rust and relay, please review closely! no idea what i'm doing

relay-filter/src/browser_extensions.rs
```diff
@@ -83,6 +93,13 @@ pub fn matches(event: &Event) -> bool {
         if EXTENSION_EXC_VALUES.is_match(ex_val) {
             return true;
         }
+        // Filtering for iOS based chrome trasnlation errors
+        if let Some(os_name) = event.tag_value("os.name") {
```
Dav1dde:
```suggestion
        if event.tag_value("os.name") == Some("iOS") && IOS_EXC_VALUES.is_match(ex_val) {
```

Should do.

relay-filter/src/browser_extensions.rs
```diff
@@ -293,4 +324,19 @@ mod tests {
             )
         }
     }
+
+    #[test]
+    fn test_filter_ios_specific_extensions() {
+        let exceptions = ["undefined is not an object (evaluating 'a.L')"];
```
Dav1dde: The other tests have something like:

```suggestion
        let exceptions = [get_event_with_exception_value_and_os_name("undefined is not an object (evaluating 'a.L')", "iOS");];
```

Could do the same here.

relay-filter/src/browser_extensions.rs
```diff
@@ -293,4 +324,19 @@ mod tests {
...
+
+        for exc_value in &exceptions {
+            let event = get_event_with_exception_value_and_os_name(exc_value, "iOS");
```
Dav1dde: There is no test if the rule does *not* match when the os is not `iOS`

Dav1dde (commented):
I wonder if we actually need the condition on the os name, do you think this would ever match when it shouldn't?

relay-filter/src/browser_extensions.rs
```diff
@@ -83,6 +93,13 @@ pub fn matches(event: &Event) -> bool {
         if EXTENSION_EXC_VALUES.is_match(ex_val) {
             return true;
         }
+        // Filtering for iOS based chrome trasnlation errors
```
iker-barriocanal:
```suggestion
        // Filtering for iOS based chrome translation errors
```

relay-filter/src/browser_extensions.rs
```diff
@@ -189,6 +206,20 @@ mod tests {
...
+
+        let tags = vec![Annotated::new(TagEntry(
+            Annotated::new("os.name".to_string()),
+            Annotated::new(os_name.to_string()),
+        ))];
+
+        Event {
+            tags: Annotated::new(Tags(PairList::from(tags))),
+            ..event
+        }
```
iker-barriocanal: I think this is more clear:
```suggestion
        let mut event = get_event_with_exception_value(val);
        event.tags = vec![Annotated::new(TagEntry(
            Annotated::new("os.name".to_string()),
            Annotated::new(os_name.to_string()),
        ))];
        event
```
(maybe some additional changes are required to compile)

iker-barriocanal (commented):
> I wonder if we actually need the condition on the os name, do you think this would ever match when it shouldn't?

+1. Line 35 contains another iOS-specific filter. Unless we're certain we'd filter too much, I'd keep it simple and just match the message.

scttcper:
Alright lets keep it simple

iker-barriocanal (approved):

PR #3252, opened 2024-03-12
ref(bucket-encoding): Re-use a single buffer for encoded buckets

Dav1dde:
We can re-use a single buffer for encoding multiple buckets instead of creating a new allocation every time.

Attaching the buffer to the encoder isn't the best API, but it is in this case a lot easier to deal with than carrying the buffer through the method calls and then not forgetting to reset it multiple times.

#skip-changelog

### DIFF
--- a/relay-server/src/utils/bucket_encoding.rs
+++ b/relay-server/src/utils/bucket_encoding.rs
@@ -10,12 +10,16 @@ static BASE64: data_encoding::Encoding = data_encoding::BASE64;
 
 pub struct BucketEncoder<'a> {
     global_config: &'a GlobalConfig,
+    buffer: String,
 }
 
 impl<'a> BucketEncoder<'a> {
     /// Creates a new bucket encoder with the provided config.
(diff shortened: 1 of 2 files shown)


relay-server/src/utils/bucket_encoding.rs
```diff
@@ -44,34 +48,40 @@ impl<'a> BucketEncoder<'a> {
...
+        self.do_encode(enc, set)
     }
 
-    fn do_encode<T: Encodable>(enc: BucketEncoding, data: T) -> io::Result<ArrayEncoding<T>> {
+    fn do_encode<T: Encodable>(
+        &mut self,
+        enc: BucketEncoding,
+        data: T,
+    ) -> io::Result<ArrayEncoding<T>> {
+        self.buffer.clear();
```
olksdr, line 78: Maybe add docs  with note that internal string buffer cleared on each sequential call of the function.  Since it can be easy to miss if someone will want to refactor this part.

olksdr (approved):

PR #3254, opened 2024-03-12
feat(metrics): Add metadata for buckets

Dav1dde:
Adds `BucketMetadata` to the metrics `Bucket`. The metadata can be used to track meta information about the bucket, like amount of aggregations (total merges).

The metadata currently only tracks the amount of "merges".

Bucket metadata is a purely Relay internal detail, clients are not expected to send it and Relay will ignore and reset the metadata if it is not coming from a trusted/internal Relay. This makes it possible to to implement the [received timestamp for metrics](https://github.com/getsentry/team-ingest/issues/265) while at the same time forwarding the metadata correctly from PoPs to Processing Relays without any additional configurations.

This PR contains:
- Implementation of "merges" metadata
- Extraction of Relay Id from Envelope requests
- Resetting the metadata from unknown/non-internal sources

The metadata is currently unused, I will add all missing integration tests once it is possible to properly assert the metadata in an integration test. I have manually verified that the aggregation of metadata is working.

Change is smaller than it looks, most of it is just test snapshots.


relay-metrics/src/bucket.rs
```diff
@@ -735,6 +744,43 @@ impl CardinalityItem for Bucket {
...
+    pub fn new() -> Self {
+        Self {
+            merges: NonZeroU64::MIN,
+        }
+    }
+
+    /// Whether the metadata does not contain more information than the default.
+    pub fn is_default(&self) -> bool {
+        let Self { merges: volume } = self;
+        *volume == NonZeroU64::MIN
```
jjbayer:
```suggestion
        let Self { merges } = self;
        *merges == NonZeroU64::MIN
```

relay-server/src/extractors/request_meta.rs
```diff
@@ -433,18 +453,35 @@ impl PartialMeta {
...
+            .and_then(|h| h.to_str().ok())
+            .and_then(|h| h.parse::<RelayId>().ok());
+
+        if let Some(relay_id) = relay_id {
+            relay_log::configure_scope(|s| s.set_tag("relay_id", relay_id));
+
+            relay = state
+                .relay_cache()
+                .send(GetRelay { relay_id })
+                .await
```
jjbayer: Does this `.await` block on requesting information from the upstream?

relay-metrics/src/bucket.rs
```diff
@@ -735,6 +744,43 @@ impl CardinalityItem for Bucket {
     }
 }
 
+/// Relay internal metadata for a metric bucket.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct BucketMetadata {
+    /// How many times the bucket was merged.
+    ///
+    /// Merging two buckets sums the amount of merges.
+    pub merges: NonZeroU64,
```
jjbayer, line 757: nit: I don't have a better name, but it would be nice to clarify somehow that this is the number of merges + 1 for the creation of the bucket.

jjbayer (commented):
Looks good! I'm unsure about the `.await` in `from_request_parts`.

What happens when a bucket goes into the pre-aggregator because a project state has not been loaded? Will the accounting be correct?

relay-server/src/extractors/request_meta.rs
Dav1dde: Yes, the same way it does for the metrics endpoint.

For our own Relays this should always come from cache (they are configured in the Relay config).

jjbayer (approved):

PR #3255, opened 2024-03-12
feat(spans): Rate limit span metrics

### LINKED ISSUE #3260: Rate limit span metrics

jjbayer:
Like we do for transaction metrics, limit span metrics when there is a quota defined for `DataCategory::Spans`.

Fixes: https://github.com/getsentry/relay/issues/3260


relay-server/src/envelope.rs
```diff
@@ -588,12 +590,14 @@ impl AddAssign for SourceQuantities {
...
             transactions,
+            spans,
             profiles,
             buckets,
         } = self;
         *transactions += other.transactions;
+        *spans += other.spans;
         *profiles += other.profiles;
-        *buckets = other.buckets;
+        *buckets += other.buckets;
```
jjbayer, line 600: I messed this up in the PR that was supposed to fix it.

relay-server/src/services/processor.rs
```diff
@@ -1903,47 +1903,74 @@ impl EnvelopeProcessorService {
...
+                        quotas,
+                        ItemScoping {
+                            category: DataCategory::Transaction,
+                            scoping: &scoping,
+                            namespace: None,
+                        },
+                        count,
+                        over_accept_once,
+                    )
+                    .ok() // don't limit if there was a redis error
```
jjbayer: This is an important change: Previously, we would drop everything if there was a redis error. I changed this to accept everything instead, for two reasons:

1. It is more correct IMO (we should not drop customer data if the current budget is unknown).
2. It simplifies the code.

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -8,31 +9,22 @@ use relay_quotas::{DataCategory, ItemScoping, Quota, RateLimits, Scoping};
...
 use crate::envelope::SourceQuantities;
-use crate::services::outcome::{DiscardReason, Outcome, TrackOutcome};
+use crate::services::outcome::{Outcome, TrackOutcome};
 
 /// Contains all data necessary to rate limit metrics or metrics buckets.
 #[derive(Debug)]
 pub struct MetricsLimiter<Q: AsRef<Vec<Quota>> = Vec<Quota>> {
-    /// A list of aggregated metric buckets.
-    metrics: Vec<Bucket>,
+    /// TODO: docs
```
jjbayer:
```suggestion
    /// A list of aggregated metric buckets with some counters.
```

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,206 @@ pub enum ExtractionMode {
...
+}
+
+impl std::ops::Deref for SummarizedBucket {
+    type Target = Bucket;
+
+    fn deref(&self) -> &Self::Target {
+        &self.bucket
+    }
+}
+
```
jjbayer:
```suggestion
```

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,206 @@ pub enum ExtractionMode {
...
+                counts,
             })
         } else {
-            Err(buckets)
+            Err(buckets.into_iter().map(|s| s.bucket).collect())
         }
     }
 
-    #[allow(dead_code)]
+    // TODO: docs
```
jjbayer:
```suggestion
    /// Returns a reference to the scoping information.
```

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,206 @@ pub enum ExtractionMode {
...
     pub fn scoping(&self) -> &Scoping {
         &self.scoping
     }
 
-    #[allow(dead_code)]
+    // TODO: docs
```
jjbayer:
```suggestion
    /// Returns a reference to the list of quotas.
```

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,206 @@ pub enum ExtractionMode {
...
     pub fn quotas(&self) -> &[Quota] {
         self.quotas.as_ref()
     }
 
-    #[allow(dead_code)]
-    pub fn transaction_count(&self) -> usize {
-        self.transaction_count
+    // TODO: docs
```
jjbayer:
```suggestion
    /// Counts the number of transactions represented in this batch.
    ///
    /// Returns
    /// - `None` if the batch does not contain transaction metrics.
    /// - `Some(0)` if the batch contains transaction metrics, but no `usage` count.
    /// - `Some(n > 0)` if the batch contains a `usage` count.
```

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,206 @@ pub enum ExtractionMode {
...
+    pub fn transaction_count(&self) -> Option<usize> {
+        self.counts.transactions
+    }
+
+    // TODO: docs
```
jjbayer:
```suggestion
    /// Counts the number of spans represented in this batch.
    ///
    /// Returns
    /// - `None` if the batch does not contain span metrics.
    /// - `Some(0)` if the batch contains span metrics, but no `usage` count.
    /// - `Some(n > 0)` if the batch contains a `usage` count.
```

relay-server/src/utils/rate_limits.rs
```diff
@@ -204,7 +212,9 @@ impl EnvelopeSummary {
 
             if let Some(source_quantities) = item.source_quantities() {
                 summary.secondary_transaction_quantity += source_quantities.transactions;
+                summary.secondary_span_quantity += source_quantities.spans;
                 summary.profile_quantity += source_quantities.profiles;
+                // TODO: span metrics quantity
```
jjbayer:
```suggestion
```

relay-server/src/services/processor.rs
iker-barriocanal: nit: adding point 1 as a comment in the code is valuable (same for spans below).

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -8,31 +9,22 @@ use relay_quotas::{DataCategory, ItemScoping, Quota, RateLimits, Scoping};
...
-
-    /// Binary index of metrics/buckets that encode processed profiles.
-    profile_buckets: Vec<bool>,
-
     /// The number of transactions contributing to these metrics.
-    transaction_count: usize,
-
-    /// The number of profiles contained in these metrics.
-    profile_count: usize,
+    counts: TotalEntityCounts,
```
iker-barriocanal: `counts` also counts spans now, so something like:
```diff
-/// The number of transactions contributing to these metrics.
+/// The number of performance items (transactions, spans) contributing to these metrics.
```
(can't make a suggestion on a life outside of the diff)

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -45,30 +37,34 @@ const PROFILE_TAG: &str = "has_profile";
...
         Err(_) => {
             relay_log::error!("invalid MRI: {}", metric.name());
-            return None;
+            return BucketSummary::default();
         }
     };
 
-    if mri.namespace != MetricNamespace::Transactions {
-        return None;
+    if mri.namespace == MetricNamespace::Transactions {
```
iker-barriocanal: Could we make `match mri.namespace` instead?

iker-barriocanal (approved):
LGTM.

I think the code around rate limiting is more complex than it should and we should, but this is not actionable for this PR.

relay-server/src/services/processor.rs
Dav1dde: Should we `relay_log` the error? Latest Rust got `inspect_err` 👀

relay-server/src/services/processor.rs
```diff
@@ -1903,47 +1903,74 @@ impl EnvelopeProcessorService {
...
+            });
 
-            if was_enforced {
-                if let Ok(limits) = rate_limits {
+            let span_rate_limits = bucket_limiter.span_count().and_then(|count| {
```
Dav1dde: This is pretty much the same code as for transactions, I feel like a separate function or a loop would make this more readable. 

With an iterator you can `flatten()` the `Option<>` rate limits and then `reduce`.

relay-server/src/services/processor.rs
```diff
@@ -1903,47 +1903,74 @@ impl EnvelopeProcessorService {
...
+                            scoping: &scoping,
+                            namespace: None,
+                        },
+                        count,
+                        over_accept_once,
+                    )
+                    .ok() // don't limit if there was a redis error
+            });
+
+            let merged_rate_limits = match (transaction_rate_limits, span_rate_limits) {
```
Dav1dde: Generic merge of rate limits, which doesn't need to know about transactions/spans, make it a method on the result `fn merge(a: Option<Self>, B: Option<Self>) -> Option<Self>` or a function?

relay-server/src/utils/rate_limits.rs
```diff
@@ -196,6 +199,11 @@ impl EnvelopeSummary {
                 summary.span_metrics_extracted = true;
             }
 
+            if item.is_span() && item.metrics_extracted() {
```
Dav1dde: Duplicate code, the if above is 1:1 the same?

relay-server/src/utils/rate_limits.rs
```diff
@@ -153,6 +153,9 @@ pub struct EnvelopeSummary {
     /// marking the envelope data category a [`DataCategory::Transaction`].
     pub secondary_transaction_quantity: usize,
 
+    /// See `secondary_transaction_quantity`.
+    pub secondary_span_quantity: usize,
```
Dav1dde, line 157: There is never an outcome produced for this in `reject`?

Dav1dde (changes requested):
Request changes because I think there is a logic bug in the `MangedEnvelope::reject`

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,212 @@ pub enum ExtractionMode {
...
+
+#[derive(Debug)]
+struct SummarizedBucket {
+    bucket: Bucket,
+    summary: BucketSummary,
+}
+
+#[derive(Debug, Default, Clone)]
+struct TotalEntityCounts {
+    transactions: Option<usize>,
```
Dav1dde, line 183: Why is `profiles` `usize` but all the others `Option<usize>`, also technically `transactions: Some(0)` is allowed here, shouldn't this just be an `usize`?

I think that simplifies a bunch of code.

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -45,30 +37,34 @@ const PROFILE_TAG: &str = "has_profile";
 /// Additionally tracks whether the transactions also contained profiling information.
 ///
 /// Returns `None` if the metric was not extracted from transactions.
-fn count_metric_bucket(metric: BucketView<'_>, mode: ExtractionMode) -> Option<TransactionCount> {
+fn count_metric_bucket(metric: BucketView<'_>, mode: ExtractionMode) -> BucketSummary {
```
Dav1dde: Probably can get a better name now, something something summary?

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -45,30 +37,34 @@ const PROFILE_TAG: &str = "has_profile";
 /// Additionally tracks whether the transactions also contained profiling information.
```
Dav1dde, line 37: Docs could include spans now I guess

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,212 @@ pub enum ExtractionMode {
...
         buckets: Vec<Bucket>,
         quotas: Q,
         scoping: Scoping,
         mode: ExtractionMode,
     ) -> Result<Self, Vec<Bucket>> {
-        let counts: Vec<_> = buckets
-            .iter()
-            .map(|metric| count_metric_bucket(BucketView::new(metric), mode))
+        let buckets: Vec<_> = buckets
+            .into_iter()
```
Dav1dde, line 234: Simplifies code, but this results in an additional allocation, which is probably fine, just wanna bring it up.

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,212 @@ pub enum ExtractionMode {
...
+    #[cfg(feature = "processing")]
+    pub fn transaction_count(&self) -> Option<usize> {
+        self.counts.transactions
+    }
+
+    /// Counts the number of spans represented in this batch.
+    ///
+    /// Returns
+    /// - `None` if the batch does not contain span metrics.
+    /// - `Some(0)` if the batch contains span metrics, but no `usage` count.
```
Dav1dde: Is this information relevant/required -> can't we just return `usize`?

Dav1dde (commented):
Oops forgot a file

relay-server/src/services/processor.rs
jjbayer: Will do. I got held back a bit by [this](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#941ff156a7064236910ce1d1684cdb30) guide, but that actually does not legitimize code duplication.

relay-server/src/utils/metrics_rate_limits.rs
jjbayer: How about `BucketSummary::new`?

relay-server/src/utils/metrics_rate_limits.rs
jjbayer: They mean different things:

1. `None` means there are no transaction-related metrics at all, so we don't even have to check any rate limits.
2. `Some(0)` means there are transaction-related metrics in the batch that do not increase the redis counter, but still need to be removed when a quota is active.

I will add more docs to clarify.

relay-server/src/utils/metrics_rate_limits.rs
jjbayer: Does it? `into_iter()` takes ownership of the existing vector AFAIK.

relay-server/src/utils/metrics_rate_limits.rs
jjbayer: See earlier comment. I will extend this doc comment to justify.

relay-server/src/utils/rate_limits.rs
jjbayer: Whoops.

relay-server/src/utils/rate_limits.rs
jjbayer: Yes, thanks.

relay-server/src/utils/metrics_rate_limits.rs
Dav1dde: But do we actually need this, I only see you act on the Some/None part for rate limits?

relay-server/src/utils/metrics_rate_limits.rs
Dav1dde: `into_iter()` takes ownership, but each element of the vector changes size -> needs an allocation, later we extract the buckets again -> another allocation. Instead of just having 1 allocation for the extra data.

relay-server/src/utils/metrics_rate_limits.rs
Dav1dde: -> We need the distinction 👍

tests/integration/fixtures/mini_sentry.py
```diff
@@ -462,5 +462,5 @@ def reraise_test_failures():
         "maxCustomMeasurements": 10,
     },
     "filters": {"version": 1, "filters": []},
-    "options": {},
+    "options": {"relay.span-usage-metric": True},
```
Dav1dde, line 465: Do you need it for all tests? If it's just for a single test, you can just enable it for that test.

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -130,112 +131,229 @@ pub enum ExtractionMode {
...
+    /// - `Some(n > 0)` if the batch contains a `usage` count for spans.
+    ///
+    /// The distinction between `None` and `Some(0)` is needed to decide whether or not a rate limit
+    /// must be checked.
+    spans: Option<usize>,
+    /// The number of profiles represented in the current batch.
+    ///
+    /// We do not explicitly check the rate limiter for profiles, so there is no need to
+    /// distinguish between `None` and `Some(0)`.
+    profiles: usize,
```
Dav1dde, line 197: baby nit: Maybe have the profiles part of the transaction count `transactions: Option<TransactionCountWithProfiles>`?

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -246,91 +364,78 @@ impl<Q: AsRef<Vec<Quota>>> MetricsLimiter<Q> {
...
+            if let Some(limit) = active_rate_limits.longest() {
+                self.drop_with_outcome(
+                    category,
+                    Outcome::RateLimited(limit.reason_code.clone()),
+                    outcome_aggregator.clone(),
+                );
+
+                return true;
+            } else {
+                // Also check profiles:
```
Dav1dde, line 409: You're also doing this for spans now

relay-server/src/services/processor.rs
jjbayer: I ended up using `map_err` because of borrowing issues.

relay-server/src/utils/rate_limits.rs
jjbayer: Added a test for `ManagedEnvelope::reject` now.

relay-server/src/utils/metrics_rate_limits.rs
jjbayer: Renamed to `summarize_bucket` to keep the diff small.

tests/integration/fixtures/mini_sentry.py
jjbayer: It should be enabled by default, will remove the option in https://github.com/getsentry/relay/pull/3259.

relay-server/src/utils/metrics_rate_limits.rs
jjbayer: Yeah that would work

Dav1dde (approved):
:shipit:

PR #3257, opened 2024-03-12
fix(server): Include buckets in SourceQuantities addition

jjbayer:
#skip-changelog

### DIFF
--- a/relay-server/src/envelope.rs
+++ b/relay-server/src/envelope.rs
@@ -586,8 +586,14 @@ pub struct SourceQuantities {
 
 impl AddAssign for SourceQuantities {
     fn add_assign(&mut self, other: Self) {
-        self.transactions += other.transactions;
-        self.profiles += other.profiles;
+        let Self {
+            transactions,
+            profiles,
+            buckets,
+        } = self;
+        *transactions += other.transactions;
+        *profiles += other.profiles;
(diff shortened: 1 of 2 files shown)


tests/integration/test_spans.py
```diff
@@ -66,7 +66,7 @@ def test_span_extraction(
     relay.send_event(project_id, event)
 
     if discard_transaction:
-        transactions_consumer.poll(timeout=2.0) is None
+        assert transactions_consumer.poll(timeout=2.0) is None
```
jjbayer, line 69: This is another drive-by oopsie

relay-server/src/envelope.rs
```diff
@@ -586,8 +586,14 @@ pub struct SourceQuantities {
...
-        self.transactions += other.transactions;
-        self.profiles += other.profiles;
+        let Self {
+            transactions,
+            profiles,
+            buckets,
+        } = self;
+        *transactions += other.transactions;
+        *profiles += other.profiles;
+        *buckets = other.buckets;
```
jjbayer, line 596: No idea how to test this to be honest. But it seems right.

Dav1dde (approved):
pattern matching ftw

olksdr (approved):

PR #3259, opened 2024-03-13
ref(spans): Remove usage kill switch

jjbayer:
The option was intended only for the initial rollout and implemented in a hacky way. It can now be removed.

#skip-changelog


Dav1dde (approved):

iker-barriocanal (approved):
Thanks for following up!

olksdr (approved):

PR #3261, opened 2024-03-13
ref(compression): Disable compression for small responses

Dav1dde:
Unity SDKs have problems with the compression (bug) for envelope responses, we can disable the compression for small payloads which will cover the problem.

SDK fix: https://github.com/getsentry/sentry-unity/pull/1587

#skip-changelog

### DIFF
--- a/relay-server/src/services/server.rs
+++ b/relay-server/src/services/server.rs
@@ -46,7 +46,7 @@ const KEEPALIVE_RETRIES: u32 = 5;
 /// header within this time, the connection is closed.
 const CLIENT_HEADER_TIMEOUT: Duration = Duration::from_secs(5);
 
-/// Only compress responses above this configured size.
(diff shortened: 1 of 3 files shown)


olksdr (approved):
:shipit:

relay-server/src/services/server.rs
```diff
@@ -45,6 +46,13 @@ const KEEPALIVE_RETRIES: u32 = 5;
 /// header within this time, the connection is closed.
 const CLIENT_HEADER_TIMEOUT: Duration = Duration::from_secs(5);
 
+/// Only compress responses above this configured size.
```
iker-barriocanal: Could we add the unit to the docstrings? (assuming it's bytes)
```suggestion
/// Only compress responses above this configured size, in bytes.
```

iker-barriocanal (approved):

PR #3263, opened 2024-03-13
feat(starfish): enable metrics extraction for redis spans

DominikB2014:
Work for https://github.com/getsentry/sentry/issues/66875

Enables metrics for redis spans that will be used within the cache module. We will be adding any necessary tags in the future PRs once any SDK changes are in place.

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -9,13 +9,7 @@ use crate::project::ProjectConfig;
 use crate::Tag;
 
 /// A list of `span.op` patterns that indicate databases that should be skipped.
-const DISABLED_DATABASES: &[&str] = &[
-    "*clickhouse*",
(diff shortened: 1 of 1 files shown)


iker-barriocanal (commented):
Thanks for contributing! The change itself lgtm, but I have a few questions before moving forward:

- Do we know why Redis was disabled originally?
- If Redis was disabled for load-related reasons, do we have any estimations on how much load this currently represents? Could this be problematic for the infrastructure?
- Is there sufficient monitoring in downstream services to ensure new spans don't lead to unintended scenarios?
- Should the change be controlled by a feature flag?

DominikB2014:
@iker-barriocanal yeah good point, I thought we just disabled redis as its not supported in starfish but upon investigation there's a little more to this, i'm going to close this and were going to look into this some more before enabling metrics for redis spans.

PR #3265, opened 2024-03-13
ref(profiling): add segment_id to the profile

viglia:
Pass the `segment_id`, which is the `span_id` of the root span, down to the profile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 - Filter null values from metrics summary tags. ([#3204](https://github.com/getsentry/relay/pull/3204))
 - Emit a usage metric for every span seen. ([#3209](https://github.com/getsentry/relay/pull/3209))
 - Add namespace for profile metrics. ([#3229](https://github.com/getsentry/relay/pull/3229))
+- Add segment_id to the profile ([#3265](https://github.com/getsentry/relay/pull/3265))
 
 ## 24.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@
 - Filter null values from metrics summary tags. ([#3204](https://github.com/getsentry/relay/pull/3204))
 - Emit a usage metric for every span seen. ([#3209](https://github.com/getsentry/relay/pull/3209))
 - Add namespace for profile metrics. ([#3229](https://github.com/getsentry/relay/pull/3229))
-- Add segment_id to the profile ([#3265](https://github.com/getsentry/relay/pull/3265))
+- Add segment_id to the profile. ([#3265](https://github.com/getsentry/relay/pull/3265))
 
 ## 24.2.0
 
--- a/relay-event-schema/src/protocol/contexts/trace.rs
+++ b/relay-event-schema/src/protocol/contexts/trace.rs
@@ -1,6 +1,7 @@
 #[cfg(feature = "jsonschema")]
 use relay_jsonschema_derive::JsonSchema;
 use relay_protocol::{Annotated, Empty, Error, FromValue, IntoValue, Object, Value};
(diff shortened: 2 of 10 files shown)


relay-profiling/src/android.rs
```diff
@@ -217,6 +217,12 @@ pub fn parse_android_profile(
         profile.metadata.environment = environment.to_owned();
     }
 
+    if let Some(segment_id) = transaction_metadata.get("segment_id") {
+        if let Some(transaction_metadata) = profile.metadata.transaction.as_mut() {
+            transaction_metadata.segment_id = SpanId(segment_id.to_owned());
```
jjbayer, line 222: Not part of this PR, but why does `transaction_metadata` exist as both a typed object and an untyped map on `ProfileMetadata`? Could we consolidate this somehow and make `extract_transaction_metadata` return a struct?

jjbayer (commented):
@viglia could you please add a description to the PR?

relay-profiling/src/android.rs
viglia: @jjbayer quick answer is they're not exactly the same.

Here an example from a real profile:

```
    "transaction": {
        "active_thread_id": 134675695638272,
        "id": "89c91b9711e2479ab8ba56c88e2038f7",
        "name": "/api/0/organizations/{organization_slug}/events/",
        "trace_id": "84b8e259de564bc7b8622bb9f7fdb997"
    },
    "transaction_metadata": {
        "environment": "prod",
        "http.method": "GET",
        "release": "backend@03bd0a459082842c3a3963192123f75d91ab90d9",
        "transaction": "/api/0/organizations/{organization_slug}/events/",
        "transaction.end": "2024-03-05T14:42:39.082330942Z",
        "transaction.op": "http.server",
        "transaction.start": "2024-03-05T14:42:38.865197896Z",
        "transaction.status": "ok"
    },
```

Longer answer on why this was introduced to begin with, I think dates back to PR #1982

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 - Filter null values from metrics summary tags. ([#3204](https://github.com/getsentry/relay/pull/3204))
 - Emit a usage metric for every span seen. ([#3209](https://github.com/getsentry/relay/pull/3209))
 - Add namespace for profile metrics. ([#3229](https://github.com/getsentry/relay/pull/3229))
+- Add segment_id to the profile ([#3265](https://github.com/getsentry/relay/pull/3265))
```
```suggestion
- Add segment_id to the profile. ([#3265](https://github.com/getsentry/relay/pull/3265))
```

relay-event-schema/src/protocol/contexts/trace.rs
```diff
@@ -39,7 +40,19 @@ impl AsRef<str> for TraceId {
...
+    Default,
+    Eq,
+    Hash,
+    PartialEq,
+    Empty,
+    IntoValue,
+    ProcessValue,
+    Serialize,
+    Deserialize,
+)]
```
jjbayer: For consistency, could you please use `impl_str_serde!` instead?

https://github.com/getsentry/relay/blob/437fe72b2ece585d52652b9d0fe0311a0ff5b1d5/relay-event-schema/src/protocol/event.rs#L66

relay-profiling/src/android.rs
jjbayer: OK! It would still be nice to type out both objects, but it does not have to be in this PR.

jjbayer (approved):
See open comments.

relay-event-schema/src/protocol/contexts/trace.rs
viglia: sure thing. Done!

PR #3266, opened 2024-03-14
fix(cogs): Fix cogs measurements for batched metrics

Dav1dde:
The cogs were updated in a loop -> only the last iteration was actually counted.

#skip-changelog


iker-barriocanal (commented):
LGTM. Could we add a test?

Dav1dde:
> LGTM. Could we add a test?

I don't have time to start that rabbit hole.

iker-barriocanal (approved):

Issue #3147, opened 2024-02-20
[EPIC] Metric Stats


Dav1dde:
## Description

We need outcomes for DDM / custom metrics.

Outcomes for Metrics are generally a tricky problem for several reasons. Metrics are aggregated in multiple stages, SDKs, customer Relays, pop Relays and processing Relays, which makes giving an accurate number of volume hard. On top of that, volume is just a small scaling factor we have to consider, a much bigger factor is the cardinality of a metric.

Ideally outcomes can capture the volume and the cardinality of a metric.

Outcomes should tell us the volume of a single metric (defined by its MRI) and its cardinality per hour.

Our current outcomes cannot capture this information, we need a new mechanism to collect metric outcomes.

#### Requirements

- Indefinite (?) retention
- Fast enough to query for billing purposes
- Volume and Cardinality needs to be represented
- Sentry UI needs access to show the user
- Billing needs access for billing
-  Bizops needs access to process with their own pipelines (can they query snuba instead of reading the topic?)

#### Why not use Outcomes?

- Outcomes currently don't provide a way to group by metric name (incl. metric type and namespace).
- Cardinality needs to be `max()` aggregated not `sum()`'ed

#### Quantity / Volume

We want to determine the volume of metrics received by the first layer of our infrastructure (PoP-Relays). Client side aggregated metrics are counted with a quantity of `1`.

For example: If Relay receives 500 `statsd` items for a single metric per hour, this metric would be considered to have a volume/quantity of 500 metrics per hour.

#### Cardinality

We are interested in the cardinality of a single metric (MRI) per hour.

Cardinality can be queried from storage or collected through the Relay cardinality limiter.

#### Metric Stats Namespaces

Volume: `c:metric_stats/volume@none`
Tags:
- `mri`: Metric Name/MRI: `<type>:<namespace>/<name>[@<unit>]`
- `mri.type`: Metric Type
- `mri.namespace`: Metric namespace (extracted from the MRI)
- `outcome.id`: Outcome ID, metric outcomes share the same numeric outcome ID with regular outcomes.
- `outcome.reason`: Optional machine readable, free-form reason code.

Cardinality: `g:metric_stats/cardinality@none`
- `mri`: Metric Name/MRI: `<type>:<namespace>/<name>[@<unit>]`
- `mri.type`: Metric Type
- `mri.namespace`: Metric namespace (extracted from the MRI)
- `cardinality.limit`: The cardinality limit id which generated this report.
[...]

PR #3267, opened 2024-03-14
feat(metric-stats): Add metric stats namespace

Dav1dde:
Adds the namespace, but it is still disabled, so we can make a `librelay` release for Sentry.

Epic: https://github.com/getsentry/relay/issues/3147

#skip-changelog

### DIFF
--- a/relay-base-schema/src/metrics/mri.rs
+++ b/relay-base-schema/src/metrics/mri.rs
@@ -116,6 +116,10 @@ pub enum MetricNamespace {
     Profiles,
     /// User-defined metrics directly sent by SDKs and applications.
     Custom,
+    /// Metric stats.
+    ///
+    /// Metrics about metrics.
+    MetricStats,
     /// An unknown and unsupported metric.
     ///
     /// Metrics that Relay either doesn't know or recognize the namespace of will be dropped before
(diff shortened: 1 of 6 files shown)


relay-base-schema/src/metrics/mri.rs
```diff
@@ -116,6 +116,10 @@ pub enum MetricNamespace {
     Profiles,
     /// User-defined metrics directly sent by SDKs and applications.
     Custom,
+    /// Metric stats.
+    ///
+    /// Metrics about metrics.
+    MetricStats,
```
olksdr: How about have it only like:
```suggestion
    Stats,
```

Since the we already have `MertricNamespace`, so it should be easy to deduce that this is metrics about metrics. What do you think?

relay-cogs/src/lib.rs
```diff
@@ -141,6 +141,8 @@ pub enum AppFeature {
     MetricsSessions,
     /// Metrics in the custom namespace.
     MetricsCustom,
+    /// Metrics in the `metric_stats` namespace.
+    MetricsMetricStats,
```
olksdr: if we rename `MetricStats` -> `Stats` than this will be just clear and without duplication.

relay-base-schema/src/metrics/mri.rs
Dav1dde: It was the plan to name the namespace `metric_stats` in Snuba, since it contains stats of other metrics, would you also want to rename the namespace or just the enum?

relay-cogs/src/lib.rs
Dav1dde: It looks weird, but imo it makes sense, `Metrics` is the cogs classifier and `MetricStats` is the namespace.

relay-base-schema/src/metrics/mri.rs
olksdr: I would just rename the enum in Relay, but the string definition could be whatever we already are agreed on.
But I also see your point and I myself do not like when we have such inconsistency which could lead to longer debug sessions later on.

olksdr (approved):
LGTM
In the followup I would like to see an integration test.

Dav1dde:
> In the followup I would like to see an integration test.

Will do, impossible now because the namespace is disabled.

PR #3268, opened 2024-03-14
feat(spans): Collect exclusive time for all spans

phacops:
We would like to start collecting the exclusive time for all spans while limiting cardinality by only tagging with low cardinality values. This is for @Zylphrex 's work with the metrics explorer.


relay-dynamic-config/src/defaults.rs
```diff
@@ -270,6 +270,23 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
                     .when(is_http.clone()),
             ],
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "d:spans/exclusive_time@millisecond".into(),
```
iker-barriocanal: This metric is defined above on line 143 with a different condition and collection of tags. Is the intention to emit a new metric, or relax the requirements on the existing one?

phacops: The intention is to emit the same metric with less tags for all spans but still have the additional tags on the metric for certain conditions.

I suppose I'm expecting Relay to emit both metrics when conditions are met for both but then aggregate the 2 before it's sent.

I thought that was a way to keep it clear which tag applied for all spans vs certain spans only.

phacops: Actually, since it'll emit the count twice, that doesn't work, I need to do `condition` and `!condition`.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -1308,7 +1308,7 @@ mod tests {
     #[test]
     fn test_app_start_cold_outlier() {
         let metrics = extract_span_metrics_mobile("app.start.cold", 181000.0);
-        assert!(metrics.is_empty());
+        assert_eq!(metrics.len(), 1);
```
iker-barriocanal: Validating there's one metric without knowing which one is not very valuable, and we should assert on the expected metric instead (with e.g. a snapshot).

iker-barriocanal (approved):
LGTM.

To validate `exclusive_time_condition`, could we add a test to ensure we don't emit more than one exclusive time metric for each span?

relay-dynamic-config/src/defaults.rs
jjbayer: IMO it would be more declarative to update the condition on the existing metric to `None` (i.e. always extract), and then change the `.always()` calls on the tags to `.when(exclusive_time_condition.clone()`.

We might want to rename `exclusive_time_condition` to `known_modules` or something similar.

jjbayer (commented):
What about `exclusive_time_light`? Do we need the transaction tag for the metrics explorer? If yes, are there queries that would benefit from using `exclusive_time_light` (fewer rows to scan)?

Zylphrex:
> What about `exclusive_time_light`? Do we need the transaction tag for the metrics explorer? If yes, are there queries that would benefit from using `exclusive_time_light` (fewer rows to scan)?

@jjbayer So the metrics explorer wants to expose the ability to surface samples for metrics including this span exclusive time metric. One of the main things we want to be able to support is to query all spans hence the motivation to extract this metric for all spans instead of known module spans.

Now we want the transaction tag for the metrics explorer because it is likely one of the main ways for people to filter and narrow down the set of spans they want. To my understanding, the exclusive time light metric was specifically added as an optimization for the use cases where filtering on a transaction is not needed. There will be some queries that will benefit from using the exclusive time light metric so I will be raising this point to the team but we will still want this exclusive time metric for all spans.

jjbayer:
> There will be some queries that will benefit from using the exclusive time light metric

@Zylphrex alright, feel free to add it later in a follow-up PR!

relay-dynamic-config/src/defaults.rs
```diff
@@ -142,17 +146,18 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
             category: DataCategory::Span,
             mri: "d:spans/exclusive_time@millisecond".into(),
             field: Some("span.exclusive_time".into()),
-            condition: Some(
-                (is_db.clone() | is_resource.clone() | is_mobile.clone() | is_http.clone())
-                    & duration_condition.clone(),
-            ),
+            condition: None,
```
jjbayer, line 149: Should we still apply `duration_condition` here? Asking because even `span_per_op` and `spans_per_segment` apply the duration condition.

jjbayer (approved):
See open question about `duration_condition`.

relay-dynamic-config/src/defaults.rs
phacops: I don't think we'd want to apply the duration condition here. We want to track absolutely all spans here and the condition was mainly created for a mobile specific use case.

relay-server/src/metrics_extraction/event.rs
phacops: I changed the test to look for the specific metirc name.

PR #3269, opened 2024-03-14
fix(spans): Extracts environment from inp spans

edwardgou-sentry:
Updates standalone inp span tag extraction to also extract environment tags.

#skip-changelog


edwardgou-sentry:
moving this back to draft since @phacops has environment extraction included in a separate pr

phacops:
We should really refactor that to extract the same tags for standalone spans and spans from event as `environment` is one of the tag we extract by default.

edwardgou-sentry:
btw, adding `release` as a tag as well since I noticed it's also missing

jjbayer (commented):
PR looks OK to me, but please sync with @shruthilayaj on where we expect the release and environments to be on the span. [This doc](https://www.notion.so/sentry/Fields-Migration-0acf054d01fc47bb9c95daf616dea12f) lists them as top-level fields but I'm not sure if that makes sense.

edwardgou-sentry:
> PR looks OK to me, but please sync with @shruthilayaj on where we expect the release and environments to be on the span. [This doc](https://www.notion.so/sentry/Fields-Migration-0acf054d01fc47bb9c95daf616dea12f) lists them as top-level fields but I'm not sure if that makes sense.

👍 We're gunna go ahead with this because SDKs are already sending release and environment in the span data for INP spans, so this is needed. 
We'll continue the conversation with @shruthilayaj and will refactor sdks and/or relay if there are any changes from that

phacops (approved):

Dav1dde (approved):

PR #3270, opened 2024-03-14
feat(profiles): Add support for continuous profiling

phacops:
This aims to add support for continuous profiling.

If the continuous profiling option is enabled, the SDK will start a profiler at the same time it starts. The profiler will generate a random `profiler_id` and start profiling in chunks of a fixed duration (likely 10s) and send each chunk to Relay.

The chunks will be handled as a separate `ItemType` and have a separate `DataCategory`. The schema of the payload is our Sample format V2, resembling V1 but incompatible with it. Changes are:
- sample timestamps are now in milliseconds, represented as a `f64` instead of a `String` and renamed `timestamp_ms`
- OS/runtime/device metadata are removed
- queue metadata is merged with thread metadata
- new `profiler_id` field representing a profiling session
- no more transaction metadata

Each chunk will then be sent through the same topic to our consumers for further processing and storage. `Accepted` outcomes based on the duration of each chunk in milliseconds will be emitted there.


relay-profiling/src/sample_v2.rs
```diff
@@ -0,0 +1,221 @@
...
+    ///
+    /// Mutates the profile chunk. Removes invalid samples and threads.
+    /// Throws an error if the profile chunk is malformed.
+    /// Removes extra metadata that are not referenced in the samples.
+    ///
+    /// profile.normalize("cocoa", "arm64e")
+    pub fn normalize(&mut self, platform: &str) -> Result<(), ProfileError> {
+        // Clean samples before running the checks.
+        self.remove_idle_samples_at_the_edge();
+        self.remove_single_samples_per_thread();
```
Zylphrex: Does this still make sense in this continuous profiling world? We removed it because we couldn't render anything with a single sample but here we have the possibility to load additional chunks right?

Unless we're saying if a chunk only has a single sample for the thread, then the thread probably finished and there is no more data for it?

relay-profiling/src/sample_v2.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Ensures valid profile chunk or returns an error.
+    ///
+    /// Mutates the profile chunk. Removes invalid samples and threads.
+    /// Throws an error if the profile chunk is malformed.
+    /// Removes extra metadata that are not referenced in the samples.
+    ///
+    /// profile.normalize("cocoa", "arm64e")
+    pub fn normalize(&mut self, platform: &str) -> Result<(), ProfileError> {
+        // Clean samples before running the checks.
+        self.remove_idle_samples_at_the_edge();
```
Zylphrex: Also here, the edge no longer defines a transaction boundary so keeping the idle samples could make sense right? Since it could extend into the previous/next chunk.

relay-profiling/tests/fixtures/profiles/sample_v2/valid.json
```diff
@@ -0,0 +1,64 @@
...
+  "chunk_id": "0432a0a4c25f4697bf9f0a2fcbe6a814",
+  "profiler_id": "4d229f1d3807421ba62a5f8bc295d836",
+  "timestamp": "2022-09-01T09:45:00.000Z",
+  "release": "0.1 (199)",
+  "platform": "cocoa",
+  "profile": {
+    "samples": [
+      {
+        "stack_id": 0,
+        "thread_id": "0x0000000102adc700",
```
Zylphrex: Shouldn't thread ids be numbers and not hex?

phacops: That's the thing, it doesn't matter. For `cocoa`, we're focusing on queues and for the rest thread IDs. They're all `u64` anyway so need to be passed as strings. This is how we remove the `cocoa` specific `queue_metadata` field.

Having this as one string identifier simplifies things, you could have queue addresses or thread IDs in there.

relay-profiling/src/sample_v2.rs
phacops: Good point.

relay-profiling/src/sample_v2.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    debug_meta: Option<DebugMeta>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    environment: Option<String>,
+    platform: String,
+    release: String,
+
+    timestamp: DateTime<Utc>,
```
Zylphrex: Is this the start or end timestamp of the chunk? I assume start?

relay-profiling/src/sample_v2.rs
```diff
@@ -0,0 +1,221 @@
...
+use crate::error::ProfileError;
+use crate::measurements::Measurement;
+use crate::sample_v1::{DebugMeta, Frame, ThreadMetadata, Version};
+
+const MAX_PROFILE_CHUNK_DURATION_MS: f64 = 10000f64;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ProfileMetadata {
+    chunk_id: String,
+    profiler_id: String,
```
Zylphrex: Are these uuids?

relay-profiling/src/sample_v2.rs
```diff
@@ -0,0 +1,221 @@
...
+    platform: String,
+    release: String,
+
+    timestamp: DateTime<Utc>,
+    version: Version,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct Sample {
+    timestamp_ms: f64,
```
Zylphrex: Is this a relative or absolute timestamp?

relay-base-schema/src/data_category.rs
```diff
@@ -69,6 +69,16 @@ pub enum DataCategory {
...
     SpanIndexed = 16,
+    /// ProfileDuration
+    ///
+    /// This data category is used to count the number of milliseconds we have per profile chunk.
+    ProfileDuration = 17,
+    /// ProfileDurationIndexed
+    ///
+    /// This data category is used to count the number of milliseconds we have per indexed profile chunk.
+    /// We will then bill per second.
+    ProfileDurationIndexed = 18,
```
jjbayer: Please introduce these new categories in a separate Relay PR, and release librelay to ensure that the categories are known in downstream consumers before we start producing them.

relay-profiling/src/sample_v1.rs
```diff
@@ -16,7 +16,7 @@ use crate::MAX_PROFILE_DURATION;
 const MAX_PROFILE_DURATION_NS: u64 = MAX_PROFILE_DURATION.as_nanos() as u64;
 
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
-struct Frame {
+pub struct Frame {
```
jjbayer: If these types are used in both `sample_v1` and `sample_v2`, should we move them to `sample.rs` or `sample/common.rs`?

relay-profiling/src/sample_v2.rs
jjbayer: +1 on this and previous comments, let's add some doc comments to these fields.

jjbayer (commented):
Looks good overall! Could you please add an integration test to `test_store.py` to make sure the entire pipeline works as expected?

relay-base-schema/src/data_category.rs
phacops: Turns out, we don't even want a distinction between indexed/processed.

https://github.com/getsentry/relay/pull/3284

relay-profiling/src/sample_v2.rs
phacops: Unique UUIDs (so v4 or v8 depending on the implementation).

relay-profiling/src/sample_v2.rs
phacops: They are all unix timestamps, so absolute, in seconds and with millisecond precision.

relay-profiling/src/sample_v1.rs
phacops: I re-organize into a separate folder.

relay-base-schema/src/data_category.rs
jjbayer: So it can be removed then? Should we just rebase this PR on top of https://github.com/getsentry/relay/pull/3284?

relay-profiling/src/sample/v2.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// Random UUID for each profiler session
+    profiler_id: String,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    debug_meta: Option<DebugMeta>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    environment: Option<String>,
+    platform: String,
+    release: String,
```
jjbayer: Are these two fields always required?

relay-profiling/src/sample/v2.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// was captured.
+    timestamp: f64,
+    stack_id: usize,
+    thread_id: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct ProfileChunk {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    measurements: Option<BTreeMap<String, Measurement>>,
```
jjbayer: Do we need a semantic difference between `None` and an empty map? If not, I would go with `#[serde(default, skip_serializing_if = "BTreeMap::is_empty")]`.

relay-profiling/src/sample/v2.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct ProfileData {
+    samples: Vec<Sample>,
+    stacks: Vec<Vec<usize>>,
+    frames: Vec<Frame>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    thread_metadata: Option<BTreeMap<String, ThreadMetadata>>,
```
jjbayer: Same question as for `measurements`.

relay-server/src/services/processor/profile.rs
```diff
@@ -72,8 +72,18 @@ pub fn transfer_id(state: &mut ProcessEnvelopeState<TransactionGroup>) {
 #[cfg(feature = "processing")]
 pub fn process(state: &mut ProcessEnvelopeState<TransactionGroup>, config: &Config) {
     let profiling_enabled = state.project_state.has_feature(Feature::Profiling);
+    let continuous_profiling_enabled = state
+        .project_state
+        .has_feature(Feature::ContinuousProfiling);
```
jjbayer: Let's also add this check to `filter` so we can drop continuous profiles in PoPs.

relay-server/src/services/processor/profile.rs
```diff
@@ -126,6 +136,25 @@ pub fn expand_profile(
...
 
+#[cfg(feature = "processing")]
+pub fn expand_profile_chunk(item: &mut Item, config: &Config) -> ItemAction {
+    match relay_profiling::expand_profile_chunk(&item.payload()) {
+        Ok(payload) => {
+            if payload.len() <= config.max_profile_size() {
+                item.set_payload(ContentType::Json, payload);
+                ItemAction::Keep
+            } else {
+                ItemAction::Drop(Outcome::Invalid(DiscardReason::Profiling(
```
jjbayer: We should update `Item::quantity()` to avoid creating outcomes that default to `quantity=1`:

https://github.com/getsentry/relay/blob/b6211aaa94887f57f34505d0cd7c0f8c8f6d5f2f/relay-server/src/envelope.rs#L646-L651

relay-profiling/src/sample/v2.rs
phacops: Always.

relay-base-schema/src/data_category.rs
phacops: Yes. My plan was to merge #3284 to master then rebase though.

relay-profiling/src/sample/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    #[default]
+    Unknown,
+    #[serde(rename = "1")]
+    V1,
+    #[serde(rename = "2")]
+    V2,
+}
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct Frame {
```
Dav1dde, line 25: I know you just moved the files, but it would be nice if we took the opportunity and cleaned it up a bit:

- all fields should be public
- documentation on types and fields

relay-profiling/src/sample/v2.rs
```diff
@@ -0,0 +1,159 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::error::ProfileError;
+use crate::measurements::Measurement;
+use crate::sample::{DebugMeta, Frame, ThreadMetadata, Version};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ProfileMetadata {
+    /// Random UUID identifying a chunk
+    chunk_id: String,
```
Dav1dde: Fields should be pub

relay-profiling/src/sample/v2.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// Random UUID for each profiler session
+    profiler_id: String,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    debug_meta: Option<DebugMeta>,
```
Dav1dde: Docs on most fields, would be nice to also have them here.

relay-profiling/src/sample/v2.rs
```diff
@@ -0,0 +1,159 @@
...
+        serde_path_to_error::deserialize(d).map_err(ProfileError::InvalidJson)?;
+
+    profile
+        .profile
+        .normalize(profile.metadata.platform.as_str())?;
+
+    Ok(profile)
+}
+
+pub fn parse(payload: &[u8]) -> Result<Vec<u8>, ProfileError> {
```
Dav1dde: Docs would be good here, not sure why the function is that way in the first place, the caller could just use `parse_profile` and then serialize?

Is it supposed to normalize it -> better name?

Dav1dde (commented):
Only a quick check without checking logic.

relay-profiling/src/sample/v2.rs
phacops: It's coming from other formats: we needed to modify the profile after it's parsed and normalized for some reasons, before we returned the serialized payload. It was also useful to separate those 2 functions for tests as I wanted to check specific values of the field and getting back a serialized payload was unhelpful.

I'll see what I can do here, it's likely possible to compress into one function.

Dav1dde: Feel free to leave it and just give the function a more descriptive name, maybe just `parse_and_normalize` and a quick doc string?

relay-server/src/services/processor/profile.rs
phacops: I don't know if we have a good quantity to set here. Since quantity is now the duration of a profile, it means we need to first parse and process the profile properly, and since Relay won't emit `Accepted` outcomes for profile chunks and only invalid ones, how useful it is to try to process the profile to determine the appropriate duration we failed to ingest?

It will make some charts look a bit weird since it won't be the same unit for valid and invalid profile chunks. I could still try to get a duration but we would still have cases where the profile chunk is malformed or sending bogus data and we can't determine the duration of the profile chunk.

relay-profiling/src/sample/v2.rs
phacops: I refactored it the way you suggested.

relay-server/src/services/processor/profile.rs
jjbayer: After a short discussion with @Dav1dde, I believe the only correct solution is having two different outcomes: one for the number of profile chunks and one for the duration. Not necessarily a blocker for this PR, but I think it's a change we should make.

jjbayer (commented):
I would still add an integration test before we merge this, apart from that this looks good to me.

relay-server/src/services/processor/profile.rs
phacops: https://github.com/getsentry/relay/pull/3303

relay-profiling/src/sample/mod.rs
```diff
@@ -0,0 +1,83 @@
...
```
jjbayer, line 25: nit: A doc comment on the type itself would be nice.

relay-server/src/services/processor/profile_chunk.rs
```diff
@@ -0,0 +1,59 @@
...
+        .project_state
+        .has_feature(Feature::ContinuousProfiling);
+    state.managed_envelope.retain_items(|item| match item.ty() {
+        ItemType::ProfileChunk => {
+            if continuous_profiling_enabled {
+                ItemAction::Keep
+            } else {
+                ItemAction::DropSilently
+            }
+        }
```
jjbayer:
```suggestion
        ItemType::ProfileChunk if !continuous_profiling_enabled {
            ItemAction::DropSilently
        }
```

jjbayer (approved):

Dav1dde (approved):

PR #3271, opened 2024-03-14
feat: Remove sessions

lynnagara:
Sessions does not exist anymore


relay-kafka/src/config.rs
```diff
@@ -155,7 +149,6 @@ impl Default for TopicAssignments {
             transactions: "ingest-transactions".to_owned().into(),
             outcomes: "outcomes".to_owned().into(),
             outcomes_billing: None,
-            sessions: "ingest-sessions".to_owned().into(),
```
jjbayer, line 167: Does the ops repo still have configuration for this? We don't want Relay to crash because of an unknown topic parameter.

relay-server/src/envelope.rs
```diff
@@ -96,10 +96,6 @@ pub enum ItemType {
     UnrealReport,
     /// User feedback encoded as JSON.
     UserReport,
-    /// Session update data.
-    Session,
-    /// Aggregated session data.
-    Sessions,
```
jjbayer: Relay now converts session items from SDKs to metrics, but we still need to accept these incoming item types.

jjbayer (changes requested):
@lynnagara thank you for picking this up!

We cannot get rid of sessions entirely, because SDKs still send them. But we can remove the Kafka topic and hard-code the decision to drop sessions after converting them to metrics, essentially everything around:

https://github.com/getsentry/relay/blob/0dc102fcb784515994e65913d6834e86218e40b4/relay-dynamic-config/src/metrics.rs#L102-L105

relay-kafka/src/config.rs
lynnagara: Yes. https://github.com/getsentry/ops/blob/master/k8s/clusters/us/default.yaml#L3256-L3257. Though is this comment actually backwards and it needs to be removed from ops first?

lynnagara:
thanks for the feedback @jjbayer! i tried to pared this down to the minimum now -- removed the envelope changes and just taking out the part that produces to kafka. I'm not sure why the metrics extraction tests are failing now though - is it possible metrics are not being properly extracted anymore?

relay-dynamic-config/src/metrics.rs
```diff
@@ -99,9 +96,9 @@ impl SessionMetricsConfig {
         self.version >= EXTRACT_ABNORMAL_MECHANISM_VERSION
     }
 
-    /// Returns `true` if the session should be dropped after extracting metrics.
+    /// Session is dropped after extracting metrics.
     pub fn should_drop(&self) -> bool {
-        self.drop
+        true
```
jjbayer: Personally I would just remove this method.

relay-server/src/services/store.rs
```diff
@@ -242,13 +227,7 @@ impl StoreService {
...
                 }
                 ItemType::Session | ItemType::Sessions => {
-                    self.produce_sessions(
-                        scoping.organization_id,
-                        scoping.project_id,
-                        retention,
-                        client,
-                        item,
-                    )?;
+                    // Do nothing
```
jjbayer: There's a catch-all at the bottom of this match, so we can just remove this branch.

tests/integration/test_session.py
```diff
@@ -118,25 +100,8 @@ def test_session_with_processing_two_events(
...
-        "duration": None,
-        "status": "ok",
-        "errors": 0,
-        "release": "sentry-test@1.0.0",
-        "environment": "production",
-        "retention_days": 90,
-        "sdk": "raven-node/2.6.3",
-    }
+
+    sessions_consumer.assert_empty()
```
jjbayer: Shall we just remove the entire test case (same below)?

jjbayer (approved):
> Though is this comment actually backwards and it needs to be removed from ops first?

I think the comment is mostly about single tenant, are the kafka topics empty there as well? I just checked my local relay and it seems like it's fine with an unknown topic key in the config, so this should be safe to merge.

https://github.com/getsentry/ops/blob/b5a4fb9c0f0a2e2d2eaf50799d1907a593ad5f80/k8s/clusters/us/default.yaml#L3262-L3264

relay-dynamic-config/src/metrics.rs
lynnagara: done

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 - Implement volume metric stats. ([#3281](https://github.com/getsentry/relay/pull/3281))
 - Scrub transactions before enforcing quotas. ([#3248](https://github.com/getsentry/relay/pull/3248))
 - Kafka topic config supports default topic names as keys. ([#3282](https://github.com/getsentry/relay/pull/3282))
+- Stop producing to sessions topic, the feature is now fully migrated to metrics ([#3271](https://github.com/getsentry/relay/pull/3271))
+
```
```suggestion
- Stop producing to sessions topic, the feature is now fully migrated to metrics. ([#3271](https://github.com/getsentry/relay/pull/3271))
```

Dav1dde (approved):

PR #3272, opened 2024-03-15
ci(arm): Adds support for cross compilation and publishes Relay ARM docker images 

### LINKED ISSUE #1305: arm64 support for Docker images
Looks like the [Docker images](https://hub.docker.com/r/getsentry/relay) are built for amd64 only and don't support ARM architectures.

I was trying to run Sentry Relay on a AWS T4g instance, and  ran into this error:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

Looks like the base image while running is debian:buster, which supports arm64. Any chance arm64 could be supported?

Dav1dde:
Updates the build/ci to support building multi arch docker images.

- Removes the now-unused docker images and build scripts, they are no longer used and we don't want to maintain them, they will eventually break.
- Uses `regctl` to transfer multi arch images directly between registries
- Fixes `relay-debug.zip` and `relay.src.zip` inconsistency

Future improvements:
- Add `aarch64-apple-darwin` to the matrix build
- Release aarch64 binaries on release builds
- Sign images (?)

I tested all modified workflows (like binary) and they still work:
- [Release Binary](https://github.com/getsentry/relay/actions/runs/8753578569/job/24023564299)
- [DockerHub](https://hub.docker.com/layers/getsentry/relay/0f9eb4f0/images/sha256-f3b81bd3a30d2cfac9db4cf885ffa0d4e7f04fa6d7692dcd391ef2bd2a6c53dd?context=explore)

As a goodie, the docker image pipeline should be a bit faster now when re-running (~ 5 minutes), cutting down on that total CI time :)

Closes: #1305


.github/workflows/ci.yml
```diff
@@ -183,100 +183,178 @@ jobs:
...
+
+    env:
+      PLATFORMS: "linux/amd64,linux/arm64"
+      DOCKER_IMAGE: "ghcr.io/getsentry/${{ matrix.image_name }}"
+      REVISION: "${{ github.event.pull_request.head.sha || github.sha }}"
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: docker/setup-qemu-action@v3
```
olksdr, line 273: so it does use qemu for something?

.github/workflows/ci.yml
```diff
@@ -183,100 +183,178 @@ jobs:
...
+    if: ${{ (github.ref_name == 'master') }}
+
+    env:
+      GHCR_DOCKER_IMAGE: "ghcr.io/getsentry/${{ matrix.image_name }}"
+      DH_DOCKER_IMAGE: "getsentry/${{ matrix.image_name }}"
+      REVISION: "${{ github.event.pull_request.head.sha || github.sha }}"
+
+    steps:
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3.5.0
```
olksdr, line 312: Do we already sign images?

olksdr (approved):
Since in the description stays you tested the release workflows I have no other questions 

great! :shipit: 

Once thing is left - try to deploy newly built image after this PR is merged and see that everything is still functioning as expected 😄

.github/workflows/ci.yml
Dav1dde: Not 100% sure, I think buildx needs it to build the image, but the binary is compiled outside of the docker image, hence why it's almost instant.

.github/workflows/ci.yml
Dav1dde: This is just so the `regclient` action checks the signature of the regclient binary, they had that in their readme and thought it was a good idea.

PR #3273, opened 2024-03-15
fix(spans): Extract metrics from transaction spans

jjbayer:
Because span metrics extraction happens as part of transaction metrics extraction, at which point we did not convert the transaction into a span yet, span metrics were so far only extracted from child spans of transactions and standalone spans.

With this change, the transaction span gets its metrics extracted as well.

ref: https://github.com/getsentry/relay/issues/2494


relay-server/src/services/processor/span/processing.rs
```diff
@@ -246,24 +289,24 @@ pub fn maybe_discard_transaction(state: &mut ProcessEnvelopeState<TransactionGro
 #[derive(Clone, Debug)]
 struct NormalizeSpanConfig<'a> {
     /// The time at which the event was received in this Relay.
-    pub received_at: DateTime<Utc>,
+    received_at: DateTime<Utc>,
```
jjbayer, line 235: drive-by fix.

relay-server/src/services/processor.rs
```diff
@@ -577,6 +577,31 @@ struct ProcessEnvelopeState<'a, Group> {
     reservoir: ReservoirEvaluator<'a>,
 }
 
+#[cfg(test)]
+#[cfg(feature = "processing")]
+impl<'a, Group: TryFrom<ProcessingGroup>> ProcessEnvelopeState<'a, Group> {
+    fn simple(event_json: &str, group: ProcessingGroup, project_state: ProjectState) -> Self {
```
jjbayer: Just a helper method to create a state for testing.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -155,44 +159,12 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState<TransactionGroup>) {
         return;
     }
 
-    let mut add_span = |span: Annotated<Span>| {
```
jjbayer: I used the opportunity to upgrade this helper to a real function.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -230,7 +221,59 @@ pub fn extract_from_event(state: &mut ProcessEnvelopeState<TransactionGroup>) {
...
+            managed_envelope.track_outcome(
+                Outcome::Invalid(DiscardReason::InvalidSpan),
+                relay_quotas::DataCategory::SpanIndexed,
+                1,
+            );
+        }
+    }
+}
+
+/// Converts the span to an envelope item and extracts metrics for it.
```
```suggestion
/// Converts the span to an envelope item.
```

relay-server/src/services/processor.rs
```diff
@@ -577,6 +577,31 @@ struct ProcessEnvelopeState<'a, Group> {
     reservoir: ReservoirEvaluator<'a>,
 }
 
+#[cfg(test)]
+#[cfg(feature = "processing")]
```
```suggestion
#[cfg(all(test, feature = "processing"))]
```

relay-server/src/utils/managed_envelope.rs
```diff
@@ -208,6 +208,12 @@ impl ManagedEnvelope {
         envelope
     }
 
+    /// Returns an untracked envelope that does not report to outcomes or test store.
+    #[cfg(test)]
+    pub fn silent(envelope: Box<Envelope>, group: ProcessingGroup) -> Self {
```
Dav1dde: Du we need this? It's just a `Self::standalone(envelope, Addr::dummy(), Addr::dummy(), group)`?

relay-server/src/utils/managed_envelope.rs
```diff
@@ -208,6 +208,12 @@ impl ManagedEnvelope {
...
+        Self::standalone(envelope, Addr::custom().0, Addr::custom().0, group)
```
Dav1dde:
```suggestion
        Self::standalone(envelope, Addr::dummy(), Addr::dummy(), group)
```

relay-server/src/services/processor.rs
Dav1dde: I don't mind having it, but usually we just have free standing functions instead

jjbayer:
Requires change: Metrics extraction has to happen before dynamic sampling of the transaction.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -1440,4 +1444,38 @@ mod tests {
             assert!(metrics.iter().any(|b| &*b.name == mri));
         }
     }
+
+    #[test]
+    fn extracts_root_span_from_childless_transaction() {
```
iker-barriocanal: nit:
```suggestion
    fn extracts_metrics_from_root_span_from_childless_transaction() {
```

tests/integration/test_metrics.py
```diff
@@ -1540,13 +1541,18 @@ def test_span_metrics(
...
         assert headers == [("namespace", b"spans")]
         if metric["name"] in ("c:spans/count_per_op@none", "c:spans/usage@none"):
             continue
-        assert metric["tags"]["span.description"] == expected_description
-        assert metric["tags"]["span.group"] == expected_group
+
+        if metric["tags"]["span.op"] == "my-transaction-op":
+            # This is a transaction span
+            pass
+        else:
```
iker-barriocanal:
```suggestion
        # Ignore transaction spans
        if metric["tags"]["span.op"] != "my-transaction-op":
```

iker-barriocanal (approved):
LGTM.

Is it worth adding an integration test with multiple relays to ensure metrics from a transaction span are only extracted once (1x transaction, 1x span)? Since we're making changes around span<>transaction conversion, to ensure we don't break anything in the future. I'm not sure if that'd cover uncovered use cases, so your call.

relay-server/src/metrics_extraction/event.rs
jjbayer:
```suggestion
    fn extracts_span_metrics_from_transaction() {
```

jjbayer:
> Is it worth adding an integration test with multiple relays to ensure metrics from a transaction span are only extracted once (1x transaction, 1x span)?

Added span metrics to `test_transaction_metrics` now, which tests exactly that.

tests/integration/test_metrics.py
```diff
@@ -36,14 +36,18 @@ def _session_payload(timestamp: datetime, started: datetime):
     }
 
 
-def metrics_by_name(metrics_consumer, count, timeout=None):
+def metrics_by_name(metrics_consumer, count=None, timeout=None):
     metrics = {"headers": {}}
 
-    for _ in range(count):
-        metric, metric_headers = metrics_consumer.get_metric(timeout)
+    attempts = 99 if count is None else count
```
iker-barriocanal:
```suggestion
    attempts = count or 99
```

iker-barriocanal (commented):
Thanks for extending the test!

tests/integration/test_metrics.py
jjbayer: It's theoretically valid to call the function with `count=0` to assert that the topic is empty, so I prefer handling `None` explicitly.

Dav1dde (approved):

PR #3274, opened 2024-03-15
ref: Warn about dbg macro

jjbayer:
* Introduce a workspace lint to warn about `dbg!` usage.
* Enable workspace lints in all crates.

#skip-changelog


Dav1dde:
We should start moving towards [lints](https://doc.rust-lang.org/cargo/reference/manifest.html#the-lints-section).

iker-barriocanal (approved):

PR #3276, opened 2024-03-15
ref(profiling): fix tests as segment_id is not passed as part of the profile payload

viglia:
so the existing tests must work without segment_id being defined

#skip-changelog


jjbayer (approved):
Thanks for following up! Could you please make sure that there is still at least one test that _does_ set `segment_id` on the input?

PR #3277, opened 2024-03-15
feat(csp): Add support for Reporting API

### LINKED ISSUE #2645: Accept `csp-violation` reports?
We see a lot of security reports that fail parsing because they have `type:csp-violation`, which we do not support.

1. Is this a well-known type that we should support?
2. If so, extend the [security report types we support](https://github.com/getsentry/relay/blob/7e44fb03b58723cac120e1de1c7c6cecbea4532b/relay-event-schema/src/protocol/security_report.rs#L1056-L1062).

ref: [POP-RELAY-2TA](https://sentry.my.sentry.io/organizations/sentry/issues/395812/)
Ref: https://github.com/getsentry/sentry/issues/52794

olksdr:
This PR adds support for [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API) to [CSP security reports](https://developer.mozilla.org/en-US/docs/Web/API/CSPViolationReportBody). 

Reporting API has [slightly different format](https://csplite.com/csp66/#sample-violation-report) for the CSP report body, but the biggest difference that browser sends the CSP reports in batches. 

Since we also want to support old format, I've tried to minimize the changes and re-used as much code as I could. 

related: https://github.com/getsentry/sentry-docs/pull/9440
fix: https://github.com/getsentry/relay/issues/2645
related: https://github.com/getsentry/sentry/issues/38940


olksdr:
cc. @oioki once this is merged, we will have support for new CSP format

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -442,6 +481,26 @@ struct CspReportRaw {
...
 
+/// Defines CSP report sent through the [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API).
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct CspViolationRaw {
+    #[serde(rename = "type")]
+    ty: String,
+    body: CspRaw,
+
+    /// For forward compatibility.
```
Dav1dde: The field seems unused and the struct is also never serialized, seems like that does nothing and can be removed.

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -1069,7 +1135,9 @@ impl SecurityReportType {
     pub fn from_json(data: &[u8]) -> Result<Option<Self>, serde_json::Error> {
         #[derive(Deserialize)]
         #[serde(rename_all = "kebab-case")]
-        struct SecurityReport {
+        struct SecurityReport<'a> {
+            #[serde(rename = "type", borrow)]
+            ty: Option<&'a str>,
```
Dav1dde: This is going to fail deserialization when there is anything escaped in the type. Should use a `Cow<'a, str>`

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -442,6 +481,26 @@ struct CspReportRaw {
...
+    #[serde(flatten)]
+    other: BTreeMap<String, serde_json::Value>,
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(untagged)]
+enum CspVariant {
+    Csp(CspReportRaw),
+    CspViolation(CspViolationRaw),
+}
```
Dav1dde:
```suggestion
enum CspVariant {
    Csp { #[serde(rename = "csp-report")] csp_report: CspRaw },
    CspViolation { body: CspRaw }
}
```

All other types don't add value, the `CspViolationRaw` just contains additional unused fields and the `CspReportRaw` is also just a deserialization helper and never used.

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -491,9 +550,16 @@ pub struct Csp {
...
+        let variant = serde_json::from_slice::<CspVariant>(data)?;
+        match variant {
+            CspVariant::Csp(value) => Csp::simple_csp(event, value.csp_report)?,
+            CspVariant::CspViolation(value) => Csp::simple_csp(event, value.body)?,
+        }
 
+        Ok(())
+    }
+
+    fn simple_csp(event: &mut Event, raw_csp: CspRaw) -> Result<(), serde_json::Error> {
```
Dav1dde: Why is it simple?

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -442,6 +481,26 @@ struct CspReportRaw {
     csp_report: CspRaw,
 }
 
+/// Defines CSP report sent through the [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API).
```
Dav1dde: Are there any docs describing the JSON schema?

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -1069,7 +1131,9 @@ impl SecurityReportType {
     pub fn from_json(data: &[u8]) -> Result<Option<Self>, serde_json::Error> {
         #[derive(Deserialize)]
         #[serde(rename_all = "kebab-case")]
-        struct SecurityReport {
+        struct SecurityReport<'a> {
+            #[serde(rename = "type", borrow)]
+            ty: Option<Cow<'a, &'a str>>,
```
Dav1dde:
```suggestion
            ty: Option<Cow<'a, str>>,
```

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -1080,6 +1144,12 @@ impl SecurityReportType {
 
         Ok(if helper.csp_report.is_some() {
             Some(SecurityReportType::Csp)
+        } else if let Some(ty) = helper.ty {
+            if *ty == "csp-violation" {
```
Dav1dde: Nit, we could actually make the type an enum?

Dav1dde (approved):

relay-event-schema/src/protocol/security_report.rs
olksdr: I think the type will be always the same. But I'm also fine with creating an enum for this. And then also remove `Cow<...>` altogether.

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -497,6 +497,14 @@ enum CspVariant {
     CspViolation { body: CspRaw },
 }
 
+/// The type of the CSP report which comes through the Reporting API.
+#[derive(Clone, Debug, PartialEq, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+enum CspViolationType {
+    CspViolation,
+    Other,
```
Dav1dde, line 506:
```suggestion
    #[serde(other)]
    Other,
```

relay-event-schema/src/protocol/security_report.rs
```diff
@@ -1144,8 +1151,8 @@ impl SecurityReportType {
 
         Ok(if helper.csp_report.is_some() {
             Some(SecurityReportType::Csp)
-        } else if let Some(ty) = helper.ty {
-            if *ty == "csp-violation" {
+        } else if let Some(_ty) = helper.ty {
```
Dav1dde:
```suggestion
        } else if let Some(CspViolationType::CspViolation) = helper.ty {
```

Dav1dde (approved):

PR #3279, opened 2024-03-18
ref(metrics): Change bucket name to Arc<str>

Dav1dde:
The bucket name is immutable and in serveral places it is being cloned. Instead of using a String and pruning the capacity (which wasn't always done) we can instead use an immutable variant which allows for cheap clones instead.

In a follow-up we can replace metrics created by Relay with a single `Arc` instance instead of creating the mri everytime.
 
#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregator.rs
+++ b/relay-metrics/src/aggregator.rs
@@ -5,6 +5,7 @@ use std::collections::{BTreeMap, HashMap};
 use std::error::Error;
 use std::hash::Hasher;
 use std::iter::FromIterator;
+use std::sync::Arc;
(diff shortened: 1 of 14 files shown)


relay-metrics/src/aggregator.rs
```diff
@@ -663,16 +664,14 @@ impl Aggregator {
                     );
                 }
 
-                let mut metric_name = mri.to_string();
-                // do this so cost tracking still works accurately.
-                metric_name.shrink_to_fit();
```
jjbayer, line 668: Nice!

jjbayer (approved):

PR #3281, opened 2024-03-18
feat(metric-stats): Emit accepted volume

Dav1dde:
Emits positive metric stats for the custom metrics namespace after submission to Kafka.

Only implements the happy path for metric stats and only volume.


relay-dynamic-config/src/feature.rs
```diff
@@ -45,6 +45,11 @@ pub enum Feature {
     /// Serialized as `organizations:custom-metrics`.
     #[serde(rename = "organizations:custom-metrics")]
     CustomMetrics,
+    /// Enables metric stats.
+    ///
+    /// Serialized as `organizations:metric-stats`.
+    #[serde(rename = "organizations:metric-stats")]
+    MetricStats,
```
Dav1dde: Considering changing this to a global config option

relay-metrics/src/view.rs
```diff
@@ -376,6 +377,11 @@ impl<'a> BucketView<'a> {
         &self.inner.name
     }
 
+    /// Returns the name of the bucket.
```
jjbayer, line 380: nit: doc comment does not really distinguish it from the method above, maybe add something like "Caller holds shared ownership of the string"?

relay-server/src/services/metric_stats.rs
```diff
@@ -0,0 +1,132 @@
...
+
+impl Service for MetricStatsService {
+    type Interface = TrackMetric;
+
+    fn spawn_handler(self, mut rx: relay_system::Receiver<Self::Interface>) {
+        tokio::spawn(async move {
+            while let Some(message) = rx.recv().await {
+                self.handle_track_metric(message);
+            }
+            dbg!("END");
```
jjbayer:
```suggestion
            relay_log::info!("MetricStatsService has stopped.");
```

jjbayer (commented):
Looks good to me!

As discussed:

* Slight preference for a global config org rollout rate.
* Preference for a thread-less "service" / facade for `MetricStatsService`.

relay-server/src/metric_stats.rs
```diff
@@ -0,0 +1,249 @@
...
+/// Tracks stats about metrics.
+///
+/// Metric stats are similar to outcomes for envelopes, they record
+/// the final "fate" of a metric and its properties.
+///
+/// Unlike outcomes metric stats are tracked on a per MRI basis
+/// and contain additional metadata, like the cardinality of a metric.
+#[derive(Clone, Debug)]
+pub struct MetricStats {
+    config: Arc<Config>,
```
jjbayer, line 30: nit: Is `config` only used to check `processing_enabled`? If so, we could give MetricStats a noop variant instead, something like

```rust
enum MetricStats {
    Disabled,
    Enabled {
        global_config: GlobalConfigHandle,
        aggregator: Addr<Aggregator>,
    }
}
```

jjbayer (approved):

relay-server/src/metric_stats.rs
Dav1dde: That makes dealing with `self` awkward, since I have to pattern match it in every method.

PR #3282, opened 2024-03-18
ref: Kafka topic configuration

lynnagara:
As part of the kafka control plane project, Kafka config for all Sentry services should be automatically generated from a single source of truth in the ops/sentry-kafka-schemas repos.

The system knows about the default topic name + any override value per environment if relevant, but is not aware of the Relay specific values like "events" and "metrics".

This is aligned with the configuration format of Sentry, Snuba and super-big-consumers.


relay-kafka/src/config.rs
```diff
@@ -85,41 +85,52 @@ impl KafkaTopic {
 
 /// Configuration for topics.
 #[derive(Serialize, Deserialize, Debug)]
-#[serde(default)]
+#[serde(default, deny_unknown_fields)]
 pub struct TopicAssignments {
     /// Simple events topic name.
+    #[serde(alias = "ingest-events")]
```
lynnagara, line 91: alias is supposed to be temporary, while we cut over config in ops. Later this should change to rename

relay-kafka/src/config.rs
```diff
@@ -156,8 +167,7 @@ impl Default for TopicAssignments {
             outcomes: "outcomes".to_owned().into(),
             outcomes_billing: None,
             sessions: "ingest-sessions".to_owned().into(),
-            metrics: "ingest-metrics".to_owned().into(),
```
lynnagara, line 159: I removed this because I think it's confusing / error prone. There shouldn't be a metrics default. Only 2 completely separate topics for either release health metrics and generic metrics.

relay-kafka/src/config.rs
```diff
@@ -156,8 +167,7 @@ impl Default for TopicAssignments {
...
-            metrics_sessions: None,
+            metrics_sessions: "ingest-metrics".to_owned().into(),
```
lynnagara, line 168: Now there is duplication here between in default and the "alias" values, which I would like to avoid. Not sure the best way to tackle this.

relay-kafka/src/config.rs
```diff
@@ -85,41 +85,53 @@ impl KafkaTopic {
 
 /// Configuration for topics.
 #[derive(Serialize, Deserialize, Debug)]
-#[serde(default)]
+#[serde(default, deny_unknown_fields)]
```
Dav1dde: We should not validate for this. It makes preparing changes before rolling out Relay a lot easier and allows for rollbacks to older versions which do not have a certain field.

relay-kafka/src/config.rs
```diff
@@ -85,41 +85,53 @@ impl KafkaTopic {
...
+    #[serde(alias = "ingest-sessions")]
     pub sessions: TopicAssignment,
-    /// Default topic name for all aggregate metrics. Specialized topics for session-based and
-    /// generic metrics can be configured via `metrics_sessions` and `metrics_generic` each.
-    pub metrics: TopicAssignment,
-    /// Topic name for metrics extracted from sessions. Defaults to the assignment of `metrics`.
-    pub metrics_sessions: Option<TopicAssignment>,
+    /// Topic name for metrics extracted from sessions, aka release health.
+    #[serde(alias = "metrics")]
+    #[serde(alias = "ingest-metrics")]
```
Dav1dde:
```suggestion
    #[serde(alias = "ingest-metrics", alias = "metrics")]
```

relay-kafka/src/config.rs
```diff
@@ -85,41 +85,53 @@ impl KafkaTopic {
...
     /// Session health topic name.
+    #[serde(alias = "ingest-sessions")]
     pub sessions: TopicAssignment,
-    /// Default topic name for all aggregate metrics. Specialized topics for session-based and
-    /// generic metrics can be configured via `metrics_sessions` and `metrics_generic` each.
-    pub metrics: TopicAssignment,
-    /// Topic name for metrics extracted from sessions. Defaults to the assignment of `metrics`.
-    pub metrics_sessions: Option<TopicAssignment>,
+    /// Topic name for metrics extracted from sessions, aka release health.
+    #[serde(alias = "metrics")]
```
Dav1dde: This is tricky, if both `metrics` and `ingest-metrics` are defined the precedence isn't clear (it was before). That being sad, I do like the change.

relay-kafka/src/config.rs
Dav1dde: I think that's fine.

relay-kafka/src/config.rs
```diff
@@ -344,11 +355,11 @@ mod tests {
     #[test]
     fn test_kafka_config() {
         let yaml = r#"
-events: "ingest-events-kafka-topic"
+ingest-events: "ingest-events-kafka-topic"
```
Dav1dde, line 355: Can we have another test that verifies the old names still work (otherwise we break self hosted)

relay-kafka/src/config.rs
lynnagara: This is intended to be very temporary just while we cut things over in production. After the migration I think we should be very strict and only support the actual topic names

relay-kafka/src/config.rs
lynnagara: removed

relay-kafka/src/config.rs
lynnagara: Sounds good, added the test

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Kafka topic config supports default topic names as keys ([#3282](https://github.com/getsentry/relay/pull/3282))
```
```suggestion
- Kafka topic config supports default topic names as keys. ([#3282](https://github.com/getsentry/relay/pull/3282))
```

relay-kafka/src/config.rs
Dav1dde: I meant, that if someone currently has both topics configured the precedence has now changed, which is a breaking change and this might affect us as well.

relay-kafka/src/config.rs
```diff
@@ -88,38 +88,49 @@ impl KafkaTopic {
...
+    #[serde(alias = "ingest-events")]
     pub events: TopicAssignment,
     /// Events with attachments topic name.
+    #[serde(alias = "ingest-attachments")]
     pub attachments: TopicAssignment,
     /// Transaction events topic name.
+    #[serde(alias = "ingest-transactions")]
     pub transactions: TopicAssignment,
     /// Outcomes topic name.
+    #[serde(alias = "outcomes")]
```
Dav1dde: Field is already called outcomes.

relay-kafka/src/config.rs
```diff
@@ -88,38 +88,49 @@ impl KafkaTopic {
...
-    /// generic metrics can be configured via `metrics_sessions` and `metrics_generic` each.
-    pub metrics: TopicAssignment,
-    /// Topic name for metrics extracted from sessions. Defaults to the assignment of `metrics`.
-    pub metrics_sessions: Option<TopicAssignment>,
+    /// Topic name for metrics extracted from sessions, aka release health.
+    #[serde(alias = "metrics", alias = "ingest-metrics")]
+    pub metrics_sessions: TopicAssignment,
     /// Topic name for all other kinds of metrics. Defaults to the assignment of `metrics`.
     #[serde(alias = "metrics_transactions")]
+    #[serde(alias = "ingest-generic-metrics")]
```
Dav1dde:
```suggestion
    #[serde(alias = "metrics_transactions", alias = "ingest-generic-metrics")]
```

CHANGELOG.md
Dav1dde: Please move it to the `Internal` section.

Dav1dde (approved):

relay-kafka/src/config.rs
lynnagara: I double checked ST and prod -- we don't use both in any of the environments we control at least.

CHANGELOG.md
lynnagara: Out of curiosity, is this normally generated by hand?

Dav1dde: Yes, usually done by hand, but many PRs without impact have no changelog entry (this one should)

PR #3283, opened 2024-03-18
feat(spans): Enable Redis span metrics ingestion

DominikB2014:
Work for https://github.com/getsentry/sentry/issues/66875

Enables metrics for redis spans that will be used within the cache module. We will be adding any necessary tags in the future PRs once any SDK changes are in place. 

This was previously disabled because we didn't use it, but now we are looking into redis for the caching module.
We're also wildcarding the key here, so this should be fairly low cardinality.


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Internal**
```
phacops: Missing the semi-column.

phacops (approved):

jjbayer (approved):

Dav1dde (approved):

PR #3285, opened 2024-03-19
ref(statsd): Remove MetricStats

Dav1dde:
Removes metric stats, we used them for COGS evaluations, they are no longer being used.

Also reduces confusion with a new metric stats concept we're introducing.

#skip-changelog


iker-barriocanal (approved):
I love the red PRs 😄

PR #3287, opened 2024-03-19
feat(spans): Extract op/description while converting otel spans to sentry spans

sl0thentr0py:
We need the op/description for these cases for further tag/metric extraction and for stuff to show up in the product.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Features**
 
 - Add support for Reporting API for CSP reports ([#3277](https://github.com/getsentry/relay/pull/3277))
+- Extract op and description while converting opentelemetry spans to sentry spans ([#3287](https://github.com/getsentry/relay/pull/3287))
 
 **Internal**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 **Features**
 
 - Add support for Reporting API for CSP reports ([#3277](https://github.com/getsentry/relay/pull/3277))
-- Extract op and description while converting opentelemetry spans to sentry spans ([#3287](https://github.com/getsentry/relay/pull/3287))
+- Extract op and description while converting opentelemetry spans to sentry spans. ([#3287](https://github.com/getsentry/relay/pull/3287))
 
 **Internal**:
 
--- a/relay-spans/src/span.rs
+++ b/relay-spans/src/span.rs
@@ -114,7 +114,7 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
                         description = statement;
                     }
                 }
-            } else if key == "http.method" || key == "http.request.method" {
+            } else if key == "http.method" || key == "request.method" {
                 let http_op = match kind {
                     2 => "http.server",
                     3 => "http.client",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Features**
 
 - Add support for Reporting API for CSP reports ([#3277](https://github.com/getsentry/relay/pull/3277))
+- Extract op and description while converting opentelemetry spans to sentry spans ([#3287](https://github.com/getsentry/relay/pull/3287))
```
```suggestion
- Extract op and description while converting opentelemetry spans to sentry spans. ([#3287](https://github.com/getsentry/relay/pull/3287))
```

relay-spans/src/span.rs
```diff
@@ -148,10 +169,16 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
     // TODO: This is wrong, a segment could still have a parent in the trace.
     let is_segment = parent_span_id.is_empty().into();
 
+    if let Some(http_method) = http_method {
+        if let Some(http_route) = http_route {
+            description = format!("{} {}", http_method, http_route);
+        }
+    }
```
Dav1dde: You're updating the description for `db` in the loop, but http outside?

sl0thentr0py: because it relies on two different attributes

relay-spans/src/span.rs
```diff
@@ -103,6 +107,23 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
             };
             if key == "sentry.op" {
                 op = otel_value_to_string(value);
+            } else if key.starts_with("db") {
+                op = op.or(Some("db".to_string()));
```
phacops, line 111: Maybe we should remove the support for `sentry.op` since I don't think it will be used given we'll start auto-detecting the op in Relay and with this, it means we could potentially overwrite the op after detecting it.

cc @AbhiPrasad

relay-spans/src/span.rs
Dav1dde:
```suggestion
    if let (Some(http_method), Some(http_route)) = (http_route, http_method) {
        description = format!("{} {}", http_method, http_route);
    }
```

relay-spans/src/span.rs
```diff
@@ -103,6 +107,23 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
...
+            } else if key == "http.method" || key == "request.method" {
+                let http_op = match kind {
+                    2 => "http.server",
+                    3 => "http.client",
+                    _ => "http",
+                };
+                op = op.or(Some(http_op.to_string()));
+                http_method = otel_value_to_string(value);
+            } else if key == "http.route" || key == "url.path" {
+                http_route = otel_value_to_string(value);
```
Dav1dde, line 126: If you only use the `http_route` to build the description via `format`, you don't have to allocate here.

We'd only need a `Display` value here.

Dav1dde (approved):

relay-spans/src/span.rs
sl0thentr0py: yea this whole logic is a bit iffy because i stuck with one pass through the attributes array, it would be easier if this was a hashmap or we can also do multiple passes through the array to avoid the overwriting for cleaner logic.

but as a first version this is fine I think, we can iterate and clean up later.

relay-spans/src/span.rs
sl0thentr0py: i'm not sure what that implies in terms of what all i need to change, feel free to commit to the PR

Dav1dde: No worries, let's ignore it for now.

jjbayer (commented):
@phacops @AbhiPrasad do we track a mapping between OTel span fields and Sentry span fields somewhere, or is Relay the source of truth?

It would be nice to have these mappings listed in a declarative way, either in Relay itself or in dev docs.

phacops:
> @phacops @AbhiPrasad do we track a mapping between OTel span fields and Sentry span fields somewhere, or is Relay the source of truth?

@AbhiPrasad started https://github.com/getsentry/sentry-conventions to track those conventions. Right now, I'd say Relay is the source of truth.

relay-spans/src/span.rs
phacops: I don't think we can convert that as a hash map as it's the way the OTel schema is but multiple passes on the array would be cleaner indeed.

PR #3291, opened 2024-03-20
ref(store): Remove produce_metrics

jjbayer:
This PR

1. Removes code from the store service that produced metrics for envelope items. Metrics go through the separate `StoreMetrics` message now.
2. Logs an error for any unexpected envelope item that reaches the store services. This will prevent us from silently dropping items in the future. Long-term, we should use a different format than the envelope to communicate with the store service.

#skip-changelog

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -254,7 +254,6 @@ impl StoreService {
                         item,
                     )?;
                 }
-                ItemType::MetricBuckets => self.produce_metrics(scoping, item, retention)?,
                 ItemType::Profile => self.produce_profile(
                     scoping.organization_id,
                     scoping.project_id,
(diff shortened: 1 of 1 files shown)


Dav1dde (approved):

relay-server/src/services/store.rs
```diff
@@ -193,14 +193,14 @@ impl StoreService {
...
-        envelope: Box<Envelope>,
+        mut envelope: Box<Envelope>,
         start_time: Instant,
         scoping: Scoping,
     ) -> Result<(), StoreError> {
         let retention = envelope.retention();
         let client = envelope.meta().client();
         let event_id = envelope.event_id();
-        let event_item = envelope.get_item_by(|item| {
+        let event_item = envelope.as_mut().take_item_by(|item| {
```
Dav1dde, line 202: Why change this?

jjbayer: I want the event items to be gone from the envelope so I don't have to check it a second time in the match statement below.

PR #3292, opened 2024-03-20
ref(metrics): Implement MetricName newtype with optimized namespace access

Dav1dde:
Wrap the metric name in a newtype and allow easy and fast access to the namespace instead of always re-parsing with `MetricResourceIdentifier`.

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregator.rs
+++ b/relay-metrics/src/aggregator.rs
@@ -5,7 +5,6 @@ use std::collections::{BTreeMap, HashMap};
 use std::error::Error;
 use std::hash::Hasher;
 use std::iter::FromIterator;
-use std::sync::Arc;
 use std::time::Duration;
 use std::{fmt, mem};
 
@@ -19,7 +18,7 @@ use tokio::time::Instant;
 use crate::bucket::{Bucket, BucketValue};
 use crate::protocol::{self, MetricNamespace, MetricResourceIdentifier};
 use crate::statsd::{MetricCounters, MetricGauges, MetricHistograms, MetricSets, MetricTimers};
-use crate::BucketMetadata;
+use crate::{BucketMetadata, MetricName};
 
 /// Any error that may occur during aggregation.
 #[derive(Debug, Error, PartialEq)]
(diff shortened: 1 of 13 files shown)


relay-metrics/src/aggregator.rs
```diff
@@ -935,21 +930,23 @@ mod tests {
             .map(|(k, e)| (k, &e.value)) // skip flush times, they are different every time
             .collect();
 
-        insta::assert_debug_snapshot!(buckets, @r#"
+        insta::assert_debug_snapshot!(buckets, @r###"
```
jjbayer, line 933: Is your `cargo-insta` up-to-date? Some linters complain about superfluous "#" IIRC.

relay-metrics/src/bucket.rs
```diff
@@ -349,6 +349,131 @@ impl BucketValue {
     }
 }
 
+/// The name of a metric.
+///
+/// Optimized string represenation of a metric name, the contained name
+/// does not need to be valid MRI, but usually is.
```
jjbayer: nit

```suggestion
/// Optimized string representation of a metric name.
///
/// The contained name does not need to be valid MRI, but usually is.
```

relay-metrics/src/bucket.rs
```diff
@@ -349,6 +349,131 @@ impl BucketValue {
...
+}
+
+impl<'de> Deserialize<'de> for MetricName {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Deserialize::deserialize(deserializer).map(Self)
+    }
+}
```
jjbayer: Does this not derive automatically with serde feature `"rc"`? https://serde.rs/feature-flags.html#rc

jjbayer (approved):

relay-metrics/src/bucket.rs
Dav1dde: No it doesn't but you reminded me I can just use `#[serde(transparent)]` instead!

relay-metrics/src/aggregator.rs
Dav1dde: Just updated and it still changes it to `r###"`.

PR #3295, opened 2024-03-21
chore(csp): Filter out unsupported report types

olksdr:
In case if the default report group defined and used for CSP reports (with Reporting Api), browser also will use that group for sending `deprecation` [reports](https://wicg.github.io/deprecation-reporting/), which we do not support and should filter them out. 

followup for #3293 #3277 

#skip-changelog

### DIFF
--- a/relay-server/src/endpoints/security_report.rs
+++ b/relay-server/src/endpoints/security_report.rs
@@ -59,9 +59,11 @@ impl SecurityReportParams {
 
         if let Ok(items) = variant {
             for item in items {
-                let report_item =
-                    Self::create_security_item(&query, Bytes::from(item.to_owned().to_string()));
-                envelope.add_item(report_item);
+                let data = Bytes::from(item.to_owned().to_string());
+                if is_supported_type(&data) {
+                    let report_item = Self::create_security_item(&query, data);
+                    envelope.add_item(report_item);
+                }
             }
         } else {
             let report_item = Self::create_security_item(&query, body);
@@ -70,6 +72,21 @@ impl SecurityReportParams {
 
         Ok(envelope)
     }
+}
+
+/// Check if the report type is supported.
+///
+/// Note: use only for the reports sent through the Reporting Api.
+fn is_supported_type(data: &[u8]) -> bool {
+    #[derive(Deserialize)]
+    #[serde(rename_all = "kebab-case")]
+    #[serde(tag = "type")]
+    enum SecurityReportVariant {
+        CspViolation,
+    }
+
+    let helper: Option<SecurityReportVariant> = serde_json::from_slice(data).ok();
+    helper.is_some()
 }
 
 fn is_security_mime(mime: Mime) -> bool {
--- a/relay-server/src/endpoints/security_report.rs
+++ b/relay-server/src/endpoints/security_report.rs
@@ -59,11 +59,9 @@ impl SecurityReportParams {
 
         if let Ok(items) = variant {
             for item in items {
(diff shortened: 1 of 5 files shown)


relay-server/src/endpoints/security_report.rs
```diff
@@ -72,6 +74,21 @@ impl SecurityReportParams {
     }
 }
 
+/// Check if the report type is supported.
```
jjbayer:
```suggestion
/// Checks if the report type is supported.
```

relay-server/src/endpoints/security_report.rs
```diff
@@ -72,6 +74,21 @@ impl SecurityReportParams {
...
+///
+/// Note: use only for the reports sent through the Reporting Api.
```
jjbayer:
```suggestion
/// Note: use only for the reports sent through the Reporting API.
```

relay-server/src/endpoints/security_report.rs
```diff
@@ -59,9 +59,11 @@ impl SecurityReportParams {
 
         if let Ok(items) = variant {
             for item in items {
-                let report_item =
-                    Self::create_security_item(&query, Bytes::from(item.to_owned().to_string()));
-                envelope.add_item(report_item);
+                let data = Bytes::from(item.to_owned().to_string());
+                if is_supported_type(&data) {
```
jjbayer: Instead of parsing the data here, couldn't we filter out the unsupported type in the envelope processor (after or during the parsing step)?

olksdr: I first thought to do it in processor, but then , I've got an idea to discard those items as earlier as possible and later we won't do extra work.  But yeah, I will move this back to processor.

olksdr: done in [a0e3a4a](https://github.com/getsentry/relay/pull/3295/commits/a0e3a4aa596985b901e6c5980a605607c7e40d9e)

tests/integration/test_security_report.py
```diff
@@ -157,12 +157,86 @@ def test_csp_violation_reports_with_processing(
...
+
+        events = []
+        event, _ = events_consumer.get_event()
+        events.append(event)
+
+        # We will discard the unsupported "deprecation" report type.
+        assert len(mini_sentry.test_failures) > 0
+        assert {str(e) for _, e in mini_sentry.test_failures} == {
+            "Relay sent us event: failed to extract security report: event filtered with reason: InvalidCsp",
+            "Relay sent us event: dropped envelope: internal error",
```
Dav1dde: Why is this error okay, there shouldn't be internal errors because of correctly handled CSP reports?

olksdr: From what I see in the code, we let this particular error fall through and at the end of the processing we just drop the managed envelope, which produces this error.  This seems like expected  behaviour.

Dav1dde: But that means we will get Sentry errors with `dropped envelope: internal error` *and* outcomes with outcome `Invalid(Internal)` don't think this is good.

If that's always been happening, :shipit: - but we should change that asap imo.

olksdr: ok, revisit this part in [137abbd](https://github.com/getsentry/relay/pull/3295/commits/137abbd96a48788e620fd3eebb70e5d7c46a70ca)

olksdr: actually we should not produce any error in this case

Dav1dde (approved):

PR #3296, opened 2024-03-21
instr(server): topic name as tag

jjbayer:
This will allow us to match `ProcessingMessageProduced` and `ProcessingProduceError` by topic. Neither the `event_type` nor the `KafkaTopic` type match the actual topic name 1:1, because the `ShardedProducer` can have multiple topic names for a single `KafkaTopic`.

#skip-changelog

### DIFF
--- a/relay-kafka/src/producer/mod.rs
+++ b/relay-kafka/src/producer/mod.rs
@@ -158,12 +158,14 @@ impl KafkaClient {
     }
 
     /// Sends message to the provided kafka topic.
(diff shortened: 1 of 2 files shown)


Dav1dde (approved):
> Don't like having it as a return value, but I also can't think of something better.

The alternative would be to have a separate `.topic_name()` method that does additional map lookups. With the return value, we leverage the lookups that `.send()` already does.

PR #3301, opened 2024-03-21
ref(store): Include more info in store error

Dav1dde:
`FormData` is one of these items hitting the line, I want more information to understand this better.

#skip-changelog

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -299,13 +299,24 @@ impl StoreService {
                     self.produce_span(scoping, start_time, event_id, retention, item)?
                 }
                 other => {
-                    let event_type = event_item.as_ref().map(|item| item.ty());
-                    relay_log::error!(
-                        tags.project_key = %scoping.project_key,
-                        tags.event_type = ?event_type,
-                        envelope_items = envelope.items().len(),
-                        "StoreService received unexpected item type: {other}"
-                    );
+                    let event_type = event_item.as_ref().map(|item| item.ty().as_str());
+                    let item_types = envelope
+                        .items()
+                        .map(|item| item.ty().as_str())
+                        .collect::<Vec<_>>();
+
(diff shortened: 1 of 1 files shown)


relay-server/src/services/store.rs
```diff
@@ -299,7 +299,13 @@ impl StoreService {
                     self.produce_span(scoping, start_time, event_id, retention, item)?
                 }
                 other => {
-                    relay_log::error!("StoreService received unexpected item type: {other}");
+                    let event_type = event_item.as_ref().map(|item| item.ty());
+                    relay_log::error!(
+                        tags.project_key = %scoping.project_key,
+                        tags.event_type = ?event_type,
+                        envelope_items = envelope.items().len(),
```
iker-barriocanal: Since `envelope_items` is in the context, it's cheap to add the types of the items in the envelope, in case that's useful.

relay-server/src/services/store.rs
```diff
@@ -299,7 +299,13 @@ impl StoreService {
                     self.produce_span(scoping, start_time, event_id, retention, item)?
                 }
                 other => {
-                    relay_log::error!("StoreService received unexpected item type: {other}");
+                    let event_type = event_item.as_ref().map(|item| item.ty());
```
iker-barriocanal: Not sure how `event_type` is displayed, is it more readable with `.unwrap_or("null")`?

iker-barriocanal (approved):

jjbayer (approved):

relay-server/src/services/store.rs
Dav1dde: Not sure what you mean, add the envelop items to `extra`?

iker-barriocanal: Adding a list of the types of the envelope's items to the event's extra/context.

PR #3303, opened 2024-03-21
feat(profiles): Add a new category to count profile chunks

phacops:
We can't track invalid profile chunks' duration because we might not be able to parse them and it's expensive in case they need to be rate-limited but we still want to track invalid outcomes for them.

The plan is to add this data category and record outcomes by count on top of the duration.

### DIFF
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.8.52
 
-- Add a data category for continuous profiling. ([#3284](https://github.com/getsentry/relay/pull/3284))
+- Add a data categories for continuous profiling. ([#3284](https://github.com/getsentry/relay/pull/3284),[#3303](https://github.com/getsentry/relay/pull/3303))
 
 ## 0.8.50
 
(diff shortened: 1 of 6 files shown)


jjbayer (approved):

Dav1dde (changes requested):
`DataCategory` enum is not updated, the file modified is auto generated.

Also please update the python changelog.

py/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 0.8.52
 
-- Add a data category for continuous profiling. ([#3284](https://github.com/getsentry/relay/pull/3284))
+- Add a data categories for continuous profiling. ([#3284](https://github.com/getsentry/relay/pull/3284),[#3303](https://github.com/getsentry/relay/pull/3303))
```
Dav1dde: That version is already released, new entry please.

Checking your last PR, you also updated the Relay changelog, maybe should also update that with this PR just so it's consistent.

Dav1dde (approved):

PR #3304, opened 2024-03-21
fix(csp): More lenient parsing of csp reports

Dav1dde:
There are CSP reports which send the status code as a string, makes it possible for all numbers in the CSP report to be parsed from string.

#skip-changelog


olksdr (approved):
Thanks for looking into it! 
:shipit:

PR #3306, opened 2024-03-21
feat(spans): parameterize redis spans

DominikB2014:
Work towards [#cach](https://github.com/getsentry/sentry/issues/66875)

First pass to parameterize redis spans, there is still a lot of other commands to consider, and improvements to be made, but if we merge this we can audit the cardinality and real data.

Also in a future PR we will have to tag the scrubbed keys as well.


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -245,13 +251,90 @@ fn scrub_redis_keys(string: &str) -> Option<String> {
         .captures(string)
         .map(|caps| (caps.name("command"), caps.name("args")));
     let scrubbed = match parts {
-        Some((Some(command), Some(_args))) => command.as_str().to_owned() + " *",
+        Some((Some(command), Some(_args))) => {
```
```suggestion
        Some((Some(command), Some(keys_and_args))) => {
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -245,13 +251,90 @@ fn scrub_redis_keys(string: &str) -> Option<String> {
...
+            command.as_str().to_owned()
```
jjbayer: You can use `format!("{} {}", ...)` to prevent extra allocations.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -245,13 +251,90 @@ fn scrub_redis_keys(string: &str) -> Option<String> {
...
+        for arg in split_args {
+            scrubed_keys.push(parameterize_redis_key(arg.as_str()));
+        }
+    }
+    scrubed_keys
+}
+// Splits a string on whitespace but ignores spaces between single and double quotes.
+// There are edge cases with this implementation for example if the quote was escaped, or if there is a double quote within single quotes.
+// This should be fine for auditing purposes
+fn split_string_with_quotes(string: &str) -> Vec<String> {
```
jjbayer: Because correctly parsing string literals is hard, I have a suggestion for simplification: What if we always only parametrize the _first_ key, and make everything else `"*"`? With something like

```rust
let scrubbed_args = if let Some((first, rest)) = args.split_once(" ") {
    if first.starts_with('"') || first.starts_with('\'') { // It's a string literal
        "*".to_owned()
    } else {
        let mut result = parameterize_redis_key(first);
        result.push_str(" ");
        result.push_str(rest);
    }
}
```

you would not need `split_string_with_quotes` at all.

DominikB2014: I get the intention, but i'm hesitant in only parameterizing the first key. The plan is to show metrics for each key namespace, in the scenario where you often perform a command that touches many keys, the metrics would be incorrect.

I am open to trying this out first, as its much simpler and we can audit to see what the data really looks like!

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -30,6 +30,12 @@ const MAX_SEGMENT_LENGTH: usize = 25;
 /// Some bundlers attach characters to the end of a filename, try to catch those.
 const MAX_EXTENSION_LENGTH: usize = 10;
 
+const REDIS_COMMAND_SINGLE_KEY: [&str; 13] = [
+    "GET", "SET", "DEL", "EXISTS", "HGET", "HSET", "HDEL", "HEXISTS", "HLEN", "APPEND", "EXISTS",
```
phacops: There are a lot more commands than this in Redis. We should try to parse them in a more generic fashion.

relay-event-normalization/src/normalize/span/description/mod.rs
phacops: I'm with @jjbayer on this.

You should also not really distinguish between the single and multi argument commands. I'd parse the command then replace everything with a `*` and then we can re-evaluate and specialize based on commands perhaps.

Redis has a somewhat easy protocol to understand. If we need to extract more from this, we can look at the definitions of each command and parse it properly. I wouldn't be surprised if there weren't a library doing that already in Rust.

DominikB2014: For v0 we are really looking to get some level of key grouping here, so ideally we don't replace everything except the command with a *. I'm okay to just try @jjbayer suggestion and only parametrize the first key.

jjbayer:
@DominikB2014 is this PR still relevant? Or can we close it?

DominikB2014:
> @DominikB2014 is this PR still relevant? Or can we close it?

@jjbayer not needed, I'll close it!

PR #3307, opened 2024-03-21
fix(performance_scores): Fix config field that should be optional

phacops:
#skip-changelog


narsaynorath (approved):

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -339,6 +339,7 @@ pub struct PerformanceScoreProfile {
     /// See [`RuleCondition`] for all available options to specify and combine conditions.
     pub condition: Option<RuleCondition>,
     /// The version of the profile, used to isolate changes to score calculations.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
Dav1dde:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
```

Dav1dde (approved):

PR #3309, opened 2024-03-22
ref: Replace OnceCell with std OnceLock

Dav1dde:
We can't get rid of the `once_cell` crate yet, because we use `once_cell::sync::Lazy` all over the place and that hasn't been stabilized yet.

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3731,7 +3731,6 @@ dependencies = [
  "gloo-console 0.3.0",
  "gloo-net 0.4.0",
  "instant",
- "once_cell",
  "reqwasm",
  "tokio",
  "wasm-bindgen",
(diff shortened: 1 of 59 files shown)


olksdr (approved):
:shipit:

iker-barriocanal:
Do we need `OnceLock` in all these cases, or is [`std::cell::OnceCell`](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) also ok?

Dav1dde:
> Do we need OnceLock in all these cases, or is [std::cell::OnceCell](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) also ok?

All the `static` places, no. All the other places (globs, PII) maybe (depends whether we need the data structure to be Sync). I can look into that in a follow-up.

This is a 1:1 replacement, we've always used `once_cell::sync` never `once_cell::unsync`.

PR #3310, opened 2024-03-22
feat(spans): Set all span tags on transaction span

jjbayer:
Previously, only span tags extracted from the transaction itself were added to the transaction span. With this PR, tags that are derived from the transaction span are set as well.

ref: https://github.com/getsentry/relay/issues/2494

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -22,7 +22,7 @@ use relay_protocol::{Annotated, Empty, Error, ErrorKind, Meta, Object, Value};
 use smallvec::SmallVec;
 
 use crate::normalize::request;
-use crate::span::tag_extraction::{self, extract_span_tags};
+use crate::span::tag_extraction;
 use crate::utils::{self, get_event_user_tag, MAX_DURATION_MOBILE_MS};
 use crate::{
     breakdowns, legacy, mechanism, schema, span, stacktrace, transactions, trimming, user_agent,
@@ -244,12 +244,7 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) {
     }
 
     if config.enrich_spans {
-        extract_span_tags(
-            event,
-            &tag_extraction::Config {
-                max_tag_value_size: config.max_tag_value_length,
-            },
-        );
+        tag_extraction::extract_span_tags_from_event(event, config.max_tag_value_length);
     }
 }
 
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -22,7 +22,7 @@ use relay_protocol::{Annotated, Empty, Error, ErrorKind, Meta, Object, Value};
 use smallvec::SmallVec;
 
 use crate::normalize::request;
-use crate::span::tag_extraction;
+use crate::span::tag_extraction::{self, extract_span_tags_from_event};
 use crate::utils::{self, get_event_user_tag, MAX_DURATION_MOBILE_MS};
 use crate::{
     breakdowns, legacy, mechanism, schema, span, stacktrace, transactions, trimming, user_agent,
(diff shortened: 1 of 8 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -244,12 +244,7 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) {
     }
 
     if config.enrich_spans {
-        extract_span_tags(
-            event,
-            &tag_extraction::Config {
-                max_tag_value_size: config.max_tag_value_length,
-            },
-        );
```
jjbayer, line 252: I replaced the Config struct, which only had a single field, with the field itself.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -141,14 +141,23 @@ impl std::fmt::Display for RenderBlockingStatus {
...
 
-/// Configuration for span tag extraction.
-pub struct Config {
-    /// The maximum allowed size of tag values in bytes. Longer values will be cropped.
-    pub max_tag_value_size: usize,
+/// Wrapper for [`extract_span_tags`].
+///
+/// Tags longer than `max_tag_value_size` bytes will be truncated.
+pub(crate) fn extract_span_tags_from_event(event: &mut Event, max_tag_value_size: usize) {
+    let mut spans = std::mem::take(&mut event.spans);
```
jjbayer, line 149: Need to take ownership here, so I can pass a reference to the Event and a mutable reference to spans.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -185,7 +190,7 @@ pub(crate) fn extract_span_tags(event: &mut Event, config: &Config) {
 }
 
 /// Extracts tags shared by every span.
-pub fn extract_shared_tags(event: &Event) -> BTreeMap<SpanTagKey, String> {
+fn extract_shared_tags(event: &Event) -> BTreeMap<SpanTagKey, String> {
```
jjbayer, line 194: This function is now private. Users all go through `extract_span_tags`.

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics.snap
```diff
@@ -48,7 +48,7 @@ expression: metrics
             1.0,
         ),
         tags: {
-            "span.op": "mYOp",
+            "span.op": "myop",
```
jjbayer, line 51: This tag was previously transferred manually, so it did not go through regular tag normalization (i.e. lowercasing).

relay-server/src/services/processor/span.rs
```diff
@@ -31,21 +31,12 @@ pub fn filter(state: &mut ProcessEnvelopeState<SpanGroup>) {
 }
 
 /// Creates a span from the transaction and applies tag extraction on it.
-pub fn extract_transaction_span(event: &Event) -> Span {
-    let mut transaction_span: Span = event.into();
+///
+/// Returns `None` when [`tag_extraction::extract_span_tags`] clears the span, which it shouldn't.
+pub fn extract_transaction_span(event: &Event, max_tag_value_size: usize) -> Option<Span> {
```
jjbayer, line 36: Not too happy with the `Option` here, but there is no guarantee that `extract_span_tags` does not set the `Annotated<Span>` to empty.

relay-event-normalization/src/normalize/span/tag_extraction.rs
Dav1dde: That would be good as a code comment.

Dav1dde (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer, line 149:
```suggestion
    // Temporarily take ownership to pass both an event reference and a mutable span reference to `extract_span_tags`.
    let mut spans = std::mem::take(&mut event.spans);
```

PR #3311, opened 2024-03-22
chore(spans): Upgrade SQL parser

jjbayer:
We can move off our own fork now that https://github.com/sqlparser-rs/sqlparser-rs/pull/1140 has been merged & released.

#skip-changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,7 +149,7 @@ sha1 = "0.10.5"
 sha2 = "0.10.6"
 similar-asserts = "1.4.2"
 smallvec = { version = "1.11.2", features = ["serde"] }
-sqlparser = { version = "0.44.0" }
+sqlparser = "0.44.0"
 sqlx = { version = "0.7.3", default-features = false }
 statsdproxy = { version = "0.1.2", default-features = false }
 symbolic-common = { version = "12.1.2", default-features = false }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5053,8 +5053,9 @@ dependencies = [
 
 [[package]]
 name = "sqlparser"
-version = "0.43.1"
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -65,7 +65,10 @@ brotli = "3.3.4"
 bytecount = "0.6.0"
 bytes = "1.4.0"
 cadence = "0.29.0"
-chrono = { version = "0.4.31", default-features = false, features = ["std", "serde"] }
+chrono = { version = "0.4.31", default-features = false, features = [
+    "std",
+    "serde",
+] }
```
jjbayer, line 71: EvenBetterToml made some formatting changes in this file.

Cargo.toml
```diff
@@ -146,19 +149,19 @@ sha1 = "0.10.5"
 sha2 = "0.10.6"
 similar-asserts = "1.4.2"
 smallvec = { version = "1.11.2", features = ["serde"] }
-sqlparser = { git = "https://github.com/getsentry/sqlparser-rs", rev = "1bfe51c3db3e5b6bda4f52c56860af378d8ec2d3" }
+sqlparser = { version = "0.44.0" }
```
jjbayer: @Dav1dde when do we declare features in the workspace? I noticed `sqlparser` has no features listed here, but `smallvec` does.

olksdr: maybe 
```suggestion
sqlparser = "0.44.0"
```

Since there are no features are enabled.

olksdr (approved):

Cargo.toml
Dav1dde: We should basically never have features in the workspace, only reason you could put them there is it's a feature which always needs to be enabled. I haven't cleaned this up yet, I just moved stuff from the workspace crates to the workspace, but left all the stuff which was configured before.

Dav1dde (approved):

PR #3313, opened 2024-03-25
feat(cardinality): Implement name based cardinality limits

Dav1dde:
Implements metric name based cardinality limits, this does not yet use a Redis pipeline.


relay-cardinality/src/redis/state.rs
```diff
@@ -0,0 +1,160 @@
...
+    CardinalityLimit,
+};
+
+#[derive(Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+struct SubScopeKey {}
+
+// enum SubScope {
+//     One(QuotaScoping, Vec<RedisEntry>),
+//     Many(BTreeMap<QuotaScoping, Vec<RedisEntry>>),
+// }
```
iker-barriocanal: Do we need this?

relay-cardinality/src/redis/state.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// Internal state combining relevant entries for the respective quota.
+pub struct LimitState<'a> {
+    /// The limit of the quota.
+    pub limit: u64,
+
+    // /// Scoping of the quota.
```
iker-barriocanal: Do we need this comment?

relay-cardinality/src/redis/quota.rs
```diff
@@ -74,7 +98,37 @@ impl QuotaScoping {
...
+
+impl std::ops::Deref for QuotaScoping {
+    type Target = PartialQuotaScoping;
+
+    fn deref(&self) -> &Self::Target {
+        &self.parent
+    }
+}
+
+struct DisplayOptMinus<T>(Option<T>);
```
iker-barriocanal, line 128: Is the goal of this struct to display the inner object optionally? If yes, I find the extra complexity is not needed and I'd suggest moving the logic to line 101 above or defaulting to a fall-back value instead.

relay-cardinality/src/redis/limiter.rs
```diff
@@ -147,164 +145,32 @@ impl Limiter for RedisSetLimiter {
...
             }
-            drop(cache); // Give up the cache lock!
         }
 
         Ok(())
     }
 }
 
-/// Internal state combining relevant entries for the respective quota.
-struct LimitState<'a> {
```
iker-barriocanal: (I find moving this struct with the rest of the changes in the PR makes tracking actual changes quite hard)

relay-cardinality/src/redis/limiter.rs
```diff
@@ -58,12 +56,12 @@ impl RedisSetLimiter {
     fn check_limits(
         &self,
         con: &mut Connection,
-        state: &mut LimitState<'_>,
+        state: &LimitState<'_>,
+        scope: QuotaScoping,
```
iker-barriocanal, line 60: `LimitState` also contains a quota scoping too. Could we clarify we should use the one in the function's argument?

Ideally there'd be one, but I don't have suggestions on how to accomplish that.

iker-barriocanal (commented):
I find understanding the (sub)quotas a bit hard, so just leaving a shallow review with a few comments,

relay-cardinality/src/redis/state.rs
Dav1dde: Already gone :)

relay-cardinality/src/redis/quota.rs
Dav1dde: We have the same in `relay-quota` (just without the dash), we need it to stay compatible with the existing keys.

relay-cardinality/src/redis/limiter.rs
Dav1dde: `LimitState` contains the partial scope not the full scope. We could pass in the `id` and `limit` instead of the state if that makes it more clear.

relay-cardinality/src/redis/limiter.rs
jjbayer: Agree with this. Better to split PRs into "this only moves stuff around" and a separate one for functional changes.

relay-cardinality/src/redis/quota.rs
```diff
@@ -8,31 +10,34 @@ use crate::window::Slot;
...
     pub namespace: Option<MetricNamespace>,
     pub organization_id: Option<OrganizationId>,
     pub project_id: Option<ProjectId>,
     window: SlidingWindow,
+    scope: CardinalityScope,
 }
 
-impl QuotaScoping {
+impl PartialQuotaScoping {
     /// Creates a new [`QuotaScoping`] from a [`Scoping`] and [`CardinalityLimit`].
```
jjbayer:
```suggestion
    /// Creates a new [`PartialQuotaScoping`] from a [`Scoping`] and [`CardinalityLimit`].
```

relay-cardinality/src/redis/quota.rs
```diff
@@ -64,6 +69,25 @@ impl QuotaScoping {
...
+        QuotaScoping { parent: self, name }
+    }
+}
+
+/// A quota scoping extracted from a [`CardinalityLimit`], a [`Scoping`]
+/// and a [`CardinalityItem`](crate::CardinalityItem).
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct QuotaScoping {
+    parent: PartialQuotaScoping,
+    name: Option<u32>,
```
jjbayer, line 100: Haven't reviewed the PR fully, but what makes the `name` special, i.e. more special than the namespace? Why is the namespace part of `PartialQuotaScoping` but `name` is not?

relay-cardinality/src/redis/state.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// All invalid/unknown limits are skipped, see also [Self::new].
+    pub fn from_limits(scoping: Scoping, limits: &'a [CardinalityLimit]) -> Vec<Self> {
+        limits
+            .iter()
+            .filter_map(|limit| LimitState::new(scoping, limit))
+            .collect::<Vec<_>>()
+    }
+
+    /// Returns a [`QuotaScoping`] if the `entry` matches the limit contained in the state.
+    pub fn matches(&self, entry: Entry) -> Option<QuotaScoping> {
```
jjbayer: nit: 
```suggestion
    pub fn matching_scope(&self, entry: Entry) -> Option<QuotaScoping> {
```

relay-cardinality/src/redis/state.rs
```diff
@@ -0,0 +1,175 @@
...
+pub struct LimitState<'a> {
+    /// The limit of the quota.
+    pub limit: u64,
+
+    /// Scoping of the quota.
+    scope: PartialQuotaScoping,
+    /// Entries grouped by quota scoping.
+    ///
+    /// Contained scopes are a superset of `scope`.
+    sub_scopes: BTreeMap<QuotaScoping, Vec<RedisEntry>>,
```
jjbayer: Similar question as above: Why is there a single limit state holding multiple `name`s , instead of a separate limit state per name?

jjbayer (commented):
No objections, but I'd like to understand `QuotaScoping` vs `PartialQuotaScoping` a bit better. Happy to discuss this in person as well!

Dav1dde:
> No objections, but I'd like to understand QuotaScoping vs PartialQuotaScoping a bit better. Happy to discuss this in person as well!

Let's talk tomorrow and I'll see to improve the docs afterwards.

In the meantime I'll move the state out in a separate PR, which should address the diff concerns.

relay-cardinality/src/redis/quota.rs
Dav1dde: The `namespace` is a filter, it selects metrics which the limit applies to (in SQL: `where`).

The name is actually part of the quota, each metric name gets its own Redis set (in SQL: `group by`).

relay-cardinality/src/redis/state.rs
Dav1dde: There is a 1:1 relationship between cardinality limit and a limit state, which helps with a few things:

- Everything that is related to the cardinality limit is tracked in a single state (e.g. all metrics).
- There is known fixed amount of states independently of the amount of items passed in, which means there is no need to have 2 phases of "gather all states by looping through the items" and a 2nd phase of "adding items to their states".
- We can use the already setup `PartialQuotaScoping` to "filter" the metrics down and only then branch out and create more scopes when necessary.
- Additionally the `PartialQuotaScope` defines the Redis sharding key, having the "full" scopes already grouped by their common partial scopes makes batching very easy instead of having another phase which groups all the independent states together.

relay-cardinality/src/redis/limiter.rs
Dav1dde: Done in #3329

relay-cardinality/src/redis/quota.rs
Dav1dde: I really hate that rust analyzers rename ignores doc comments ... ty

relay-cardinality/src/redis/quota.rs
```diff
@@ -8,31 +10,37 @@ use crate::window::Slot;
 use crate::{CardinalityLimit, CardinalityScope, OrganizationId, Scoping, SlidingWindow};
 
 /// A quota scoping extracted from a [`CardinalityLimit`] and a [`Scoping`].
-#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
-pub struct QuotaScoping {
+///
+/// The partial quota scoping can be used to select/match on cardinality entries
+/// but it needs to be completed into a [`QuotaScoping`] by
```
jjbayer: missing doc

relay-cardinality/src/redis/quota.rs
```diff
@@ -64,6 +72,32 @@ impl QuotaScoping {
         UnixTimestamp::from_secs(timestamp.as_secs() + shift)
     }
 
+    /// Creates a [`QuotaScoping`] from the partial scoping and the passed [`Entry`].
+    ///
+    /// This unconditionally creates a quota scoping from the passed entry and
+    /// does not check whether the scoping even applies to the entry. The caller
+    /// needs to ensure this by calling [`Self::matches`] prior to calling `full`.
+    pub fn full(self, entry: Entry<'_>) -> QuotaScoping {
```
jjbayer:
```suggestion
    pub fn into_full(self, entry: Entry<'_>) -> QuotaScoping {
```

relay-cardinality/src/redis/quota.rs
```diff
@@ -64,6 +72,32 @@ impl QuotaScoping {
...
+    }
+}
+
+/// A quota scoping extracted from a [`CardinalityLimit`], a [`Scoping`]
+/// and completed with a [`CardinalityItem`](crate::CardinalityItem).
+///
+/// The scoping must be created using [`PartialQuotaScoping::full`].
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct QuotaScoping {
+    parent: PartialQuotaScoping,
```
iker-barriocanal, line 99: I don't like the name `parent` but I don't have better suggestions :D

relay-cardinality/src/redis/state.rs
```diff
@@ -12,13 +14,16 @@ use crate::{
...
 pub struct LimitState<'a> {
-    /// Entries which are relevant for the quota.
-    pub entries: Vec<RedisEntry>,
-    /// Scoping of the quota.
-    pub scope: QuotaScoping,
     /// The limit of the quota.
     pub limit: u64,
 
+    /// Scoping of the quota.
+    scope: PartialQuotaScoping,
```
jjbayer:
```suggestion
    partial_scoping: PartialQuotaScoping,
```

relay-cardinality/src/redis/state.rs
```diff
@@ -12,13 +14,16 @@ use crate::{
...
+    /// Entries grouped by quota scoping.
+    ///
+    /// Contained scopes are a superset of `scope`.
+    sub_scopes: BTreeMap<QuotaScoping, Vec<RedisEntry>>,
```
jjbayer:
```suggestion
    scopings: BTreeMap<QuotaScoping, Vec<RedisEntry>>,
```

relay-cardinality/src/redis/quota.rs
```diff
@@ -74,7 +108,37 @@ impl QuotaScoping {
...
+            write!(f, "{inner}-")
+        } else {
+            Ok(())
+        }
+    }
+}
+
+fn fnv32(s: &str) -> u32 {
+    use hash32::Hasher;
+    use std::hash::Hash;
```
iker-barriocanal: Can we move these imports to the top of the file?

jjbayer (approved):

iker-barriocanal (approved):

PR #3315, opened 2024-03-25
instr(kafka): Emit producer metrics

jjbayer:
Emit statistics collected by rdkafka as statsd metric. 

These metrics are disabled by default and need to be enabled by setting `statistics.interval.ms` in the Kafka config.

ref: https://github.com/getsentry/team-ingest/issues/310

#skip-changelog

### DIFF
--- a/relay-kafka/src/producer/utils.rs
+++ b/relay-kafka/src/producer/utils.rs
@@ -6,11 +6,11 @@ use relay_statsd::metric;
 
 use crate::statsd::{KafkaCounters, KafkaGauges};
 
-/// Kafka producer context that logs producer errors.
+/// Kafka client and producer context that logs statistics and producer errors.
 #[derive(Debug)]
-pub struct CaptureErrorContext;
+pub struct Context;
 
-impl ClientContext for CaptureErrorContext {
+impl ClientContext for Context {
     /// Report client statistics as statsd metrics.
     ///
     /// This method is only called if `statistics.interval.ms` is configured.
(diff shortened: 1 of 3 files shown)


relay-kafka/src/producer/utils.rs
```diff
@@ -4,20 +4,56 @@ use rdkafka::producer::{DeliveryResult, ProducerContext};
...
+}
+
+impl ProducerContext for Context {
     type DeliveryOpaque = ();
 
     /// This method is called after attempting to send a message to Kafka.
     /// It's called asynchronously for every message, so we want to handle errors explicitly here.
     fn delivery(&self, result: &DeliveryResult, _delivery_opaque: Self::DeliveryOpaque) {
+        // TODO: any `Accepted` outcomes (e.g. spans) should be logged here instead of on the caller side,
+        // such that we do not over-report in the error case.
```
Dav1dde, line 55: Do we have a ticket for this?

Dav1dde (approved):

Dav1dde:
I assume we need to also configure the mentioned interval in the ops config?

olksdr (approved):

relay-kafka/src/producer/utils.rs
jjbayer: We do now: https://github.com/getsentry/relay/issues/3318

jjbayer:
> I assume we need to also configure the mentioned interval in the ops config?

Yep, already working on that: https://github.com/getsentry/ops/pull/9796

PR #3316, opened 2024-03-25
chore(dev): Update black

jjbayer:
See https://github.com/getsentry/relay/security/dependabot/82.

#skip-changelog


olksdr (approved):

Dav1dde (approved):

asottile-sentry:
btw this vulnerability is only if you are exposing black to untrusted code so generally you can ignore it

PR #3319, opened 2024-03-25
feat(outcomes): Aggregate all outcomes

jjbayer:
https://github.com/getsentry/relay/pull/1134 introduced aggregation for most outcomes, but it did not include inbound filters.

Outcomes are no longer used for investigating individual events, so we might as well save resources by aggregating all outcomes unconditionally.

### DIFF
--- a/tests/integration/test_outcome.py
+++ b/tests/integration/test_outcome.py
@@ -977,10 +977,10 @@ def test_outcomes_aggregate_dynamic_sampling(relay, mini_sentry):
     assert outcome == expected_outcome
 
 
-def test_outcomes_do_not_aggregate(
+def test_outcomes_aggregate_inbound_filters(
     relay, relay_with_processing, mini_sentry, outcomes_consumer
 ):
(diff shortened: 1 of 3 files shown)


tests/integration/test_outcome.py
```diff
@@ -1001,42 +1001,27 @@ def test_outcomes_do_not_aggregate(
     outcomes_consumer = outcomes_consumer(timeout=1.2)
 
     # Send empty body twice
-    event_id1 = _send_event(relay)
-    event_id2 = _send_event(relay)
+    _send_event(relay)
```
Dav1dde, line 1004: The test is called `test_outcomes_do_not_aggregate`, maybe we should just delete it if it's no longer useful or at least rename it.

Dav1dde (approved):

tests/integration/test_outcome.py
jjbayer: 👍 updated the test name and comments now.

olksdr (approved):

PR #3321, opened 2024-03-25
ref(cardinality): Pipeline Redis script invocations

Dav1dde:
- Pipelines all script invocations which use the same parent quota
- Updates Redis to a fork which supports pipelining scripts.
- Updates the `PooledClientInner` enum to box the clients because of a clippy lint `large size difference between variants`
- Updates num-traits because of conflicts with the new Redis client version which has a higher min version

```
error: large size difference between variants
  --> relay-redis/src/real.rs:75:1
   |
75 | / enum PooledClientInner {
76 | |     Cluster(PooledConnection<redis::cluster::ClusterClient>),
   | |     -------------------------------------------------------- the largest variant contains at least 408 bytes
77 | |     Single(PooledConnection<redis::Client>),
   | |     --------------------------------------- the second-largest variant contains at least 200 bytes
78 | | }
   | |_^ the entire enum is at least 408 bytes
```

Requires: https://github.com/getsentry/relay/pull/3313

#skip-changelog

### DIFF
--- a/relay-cardinality/src/redis/limiter.rs
+++ b/relay-cardinality/src/redis/limiter.rs
@@ -8,10 +8,10 @@ use crate::{
     redis::{
         cache::{Cache, CacheOutcome},
         quota::QuotaScoping,
-        script::{CardinalityScript, Status},
+        script::{CardinalityScript, CardinalityScriptResult, Status},
         state::{LimitState, RedisEntry},
     },
-    statsd::{CardinalityLimiterHistograms, CardinalityLimiterTimers},
+    statsd::{CardinalityLimiterGauges, CardinalityLimiterTimers},
     CardinalityLimit, Result,
 };
 use relay_common::time::UnixTimestamp;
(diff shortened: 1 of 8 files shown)


relay-cardinality/src/redis/script.rs
```diff
@@ -39,6 +39,21 @@ pub struct CardinalityScriptResult {
     pub statuses: Vec<Status>,
 }
 
+impl CardinalityScriptResult {
+    pub fn validate(&self, num_hashes: usize) -> Result<()> {
```
iker-barriocanal, line 48: nit: should we type the error in the `Result`?

relay-cardinality/src/redis/script.rs
```diff
@@ -77,14 +92,29 @@ impl CardinalityScript {
         Self(Script::new(include_str!("cardinality.lua")))
     }
 
-    pub fn invoke(
+    pub fn pipe(&self) -> CardinalityScriptPipeline<'_> {
```
iker-barriocanal, line 107: Please, move the `CardinalityScript` definition to be right before the impl block so that it's easier to find.

relay-cardinality/src/redis/limiter.rs
```diff
@@ -172,10 +181,27 @@ impl Limiter for RedisSetLimiter {
...
     statuses: Vec<Status>,
 }
 
+impl CheckedLimits {
+    fn new(
+        scope: QuotaScoping,
+        entries: Vec<RedisEntry>,
+        result: CardinalityScriptResult,
+    ) -> Result<Self> {
+        result.validate(entries.len())?;
```
iker-barriocanal, line 197: Have we considered accepting the statuses here and checking the results in the caller?

relay-cardinality/src/redis/limiter.rs
```diff
@@ -55,39 +55,48 @@ impl RedisSetLimiter {
...
 
-        Ok(CheckedLimits {
-            entries,
-            statuses: result.statuses,
-        })
+        let results = pipeline.invoke(connection)?;
+
+        scopes
+            .into_iter()
+            .zip(results)
```
iker-barriocanal, line 92: Does the pipeline always return a response for each invocation?

iker-barriocanal (approved):

relay-cardinality/src/redis/script.rs
Dav1dde: It is typed, `crate::Result` -> `pub type Result<T, E = Error> = std::result::Result<T, E>;`. Same pattern is used for [std::io errors](https://doc.rust-lang.org/std/io/type.Result.html).

relay-cardinality/src/redis/limiter.rs
Dav1dde: Validating here makes the surrounding code much nicer (the iterator).

relay-cardinality/src/redis/limiter.rs
Dav1dde: Yes! I'll add some docs.

PR #3322, opened 2024-03-25
feat(spans): Collect duration for all spans

phacops:
The performance summary page and trace explorer will require we collect this metric for all spans.

As for `exclusive_time`, we'll limit cardinality by having a reduced number of tags.


jjbayer (commented):
Code looks good to me, but I think we should add a kill switch (automator option exposed through `RELAY_OPTIONS`). This change will again add a lot of metrics buckets, and we've already been hitting [internal limits](https://sentry.my.sentry.io/organizations/sentry/issues/725906) recently.

jjbayer (approved):
With https://github.com/getsentry/ops/pull/9815 in place soon, I think we can risk this. We can always roll back if the load is too high.

One integration test seems to need updating.

relay-dynamic-config/src/defaults.rs
```diff
@@ -400,37 +401,40 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
...
+                    .from_field("span.sentry_tags.transaction")
+                    .always(),
+                Tag::with_key("transaction.op")
+                    .from_field("span.sentry_tags.transaction.op")
+                    .always(),
+                // Mobile module:
                 Tag::with_key("span.description")
                     .from_field("span.sentry_tags.description")
-                    .always(), // already guarded by condition on metric
+                    .when(app_start_condition.clone()),
```
iker-barriocanal, line 422: Shouldn't this always be captured? (maybe the same for the span group)

phacops: Not for this metric. It's a driver of high-cardinality and since we're enabling it for all spans regardless if we can scrub the description, we're not adding those tags.

Issue #3278, opened 2024-03-18
Convert segment spans to transactions


jjbayer:
The current state of Performance is that

1. There are SDKs that emit transactions, and there are SDKs that emit standalone spans.
2. There are parts of the product that require transactions, and there are parts of the product that require spans.

We already extract spans from transactions to make span-dependent product features work for "transaction"-SDKs. Now we need to also extract transactions from spans to make transaction-dependent product features work for "span"-SDKs:

```mermaid
flowchart LR
    
    SDK -->|transaction| Relay
    SDK -->|span| Relay
    Relay -->|transaction| txc[Transaction Consumer] 
    Relay -->|span| spc[Span Consumer]
    Relay -->|span from transaction| spc
    Relay -->|transaction from span| txc
    
    linkStyle 5 color:green;
```

- In order to prevent cycles, every envelope item should be marked with `span_extracted` and `transaction_extracted` headers.
- While span extraction from transactions happens as late as possible, to make sure that all dynamic sampling, rate limiting etc. was already applied before duplicating the spans, it might be good to extract transactions from spans as _early_ as possible. The transaction processing pipeline is more mature than the span processing pipeline, so early extraction should give us a more stable result. On the other hand, we might want to limit extraction to processing Relays, just to make sure we can iterate on it fast.
```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/relay/pull/3469
- [x] Test with a single production project
```

PR #3324, opened 2024-03-25
fix(spans): Prevent duplicate extraction of score.total metric

jjbayer:
Since https://github.com/getsentry/relay/pull/3273, we accidentally emit the metric `d:transactions/measurements.score.total@ratio` twice for every transaction.

This happens because the metric is not only extracted from transactions, but also from spans: https://github.com/getsentry/relay/blob/0b8687c03b3eaf23db368733754d17c446bcb482/relay-dynamic-config/src/defaults.rs#L459-L461

As a quick fix, extract that metric only if the span is not a segment span extracted from a transaction. Long-term, we should stop extracting that _transaction_ metric from spans, and base performance score computation on `"d:spans/webvital.score.total@ratio` instead.

### DIFF
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -356,6 +356,7 @@ mod tests {
     use chrono::{TimeZone, Utc};
     use insta::assert_debug_snapshot;
     use relay_base_schema::metrics::{InformationUnit, MetricUnit};
+    use relay_protocol::RuleCondition;
     use similar_asserts::assert_eq;
 
     use super::*;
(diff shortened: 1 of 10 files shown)


tests/integration/test_spans.py
```diff
@@ -145,6 +146,60 @@ def test_span_extraction(
...
+        for item in envelope.items:
+            if item.type == "metric_buckets":
+                for metric in item.payload.json:
+                    if (
+                        metric["name"]
+                        == "d:transactions/measurements.score.total@ratio"
+                    ):
+                        score_total_seen += 1
+
+    assert score_total_seen == 1
```
jjbayer, line 200: Without the extra condition, this test fails with `2 == 1`.

relay-event-schema/src/protocol/span.rs
```diff
@@ -97,6 +97,10 @@ pub struct Span {
     #[metastructure(skip_serialization = "empty")]
     pub platform: Annotated<String>,
 
+    /// Whether the span is a segment span that was converted from a transaction.
+    #[metastructure(skip_serialization = "empty")]
+    pub was_transaction: Annotated<bool>,
```
Dav1dde, line 102: This effectively adds this field to the Span schema. Relays will accept this from other Relays (and SDKs) as well as forward it.

Is this the best way to deal with the problem and something we are okay with?

Dav1dde (approved):

edwardgou-sentry (approved):
Thank you!

relay-event-schema/src/protocol/span.rs
jjbayer: I think so, yes. I want an explicit flag to represent "this is a span extracted from a transaction", which gets serialized into the payload so the next Relay in the chain can check it as well. We will also need this for https://github.com/getsentry/relay/issues/3278. If an untrusted source sets this flag, it will only mess up their own data.

As for the schema, we already accept any field in the span payload, and the schema is not automatically documented, so adding a field in Relay does not really make a difference:
https://github.com/getsentry/relay/blob/f820884cdc4af39ecd49c3bc5fdf1bc436184283/relay-event-schema/src/protocol/span.rs#L104-L107
https://github.com/getsentry/relay/blob/f820884cdc4af39ecd49c3bc5fdf1bc436184283/relay-event-schema/src/protocol/event.rs#L467-L468

PR #3326, opened 2024-03-25
fix(perfscore): Adds span op tag to perf score totals

edwardgou-sentry:
Updates score total measurements to be tagged by `span.op`. We need to be able to filter scores based on `span.op` as we may have more standalone spans in the future with different score profiles (such as mobile).

#skip-changelog


jjbayer (approved):
@edwardgou-sentry could you please add test coverage for this change? For example by expanding the following test a bit. Currently it only looks at metric names, it could assert on tags as well:

https://github.com/getsentry/relay/blob/79472d3e71305d8db9731f7bb65be2a3be4639b8/relay-server/src/metrics_extraction/event.rs#L1418-L1419

PR #3328, opened 2024-03-26
feat(schema): Allow integers as username

### LINKED ISSUE #67601: Allow Django Integration Usernames to be Numbers
### Problem Statement

Hello!

 Today I am getting a warning when using the Django integration and the "default PII" setting.
 
 Our username fields are numbers, so I get this message on every alert:
 
Discarded invalid value (1)
user.username
expected a string
value: xxxx

I would love a way to tell Sentry that it's OK that our usernames are numbers and not have it discard them.

### Solution Brainstorm

An init param of username_format_validation=False, or username_format=int or something like that.

### Product Area

Ingestion and Filtering

jjbayer:
The django integration sends integers as usernames if the underlying `User` model returns an integer for `get_username()` (see issue).

Fixes https://github.com/getsentry/sentry/issues/67601.


relay-event-schema/src/protocol/user.rs
```diff
@@ -202,6 +202,19 @@ mod tests {
...
+    fn test_user_lenient_username() {
+        let input = r#"{"username":42}"#;
+        let output = r#"{"username":"42"}"#;
+        let user = Annotated::new(User {
+            username: Annotated::new("42".to_string().into()),
+            ..User::default()
+        });
+
+        assert_eq!(user, Annotated::from_json(input).unwrap());
+        assert_eq!(output, user.to_json().unwrap());
```
Dav1dde, line 215: imo the round trip is enough

```rs
assert_eq!( Annotated::from_json(input).unwrap().to_json().unwrap(), r#"{"username":"42"}"#)
```

Can also maybe just use insta here to assert the output?

Dav1dde (approved):

olksdr (approved):

relay-event-schema/src/protocol/user.rs
jjbayer: I copy-pasted the existing tests around this, would like to keep them as similar as possible.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Extract op and description while converting opentelemetry spans to sentry spans. ([#3287](https://github.com/getsentry/relay/pull/3287))
 - Drop `event_id` and `remote_addr` from all outcomes. ([#3319](https://github.com/getsentry/relay/pull/3319))
 - Support for AI token metrics ([#3250](https://github.com/getsentry/relay/pull/3250))
+- Accept integers in `event.user.username`. ([#3328](https://github.com/getsentry/relay/pull/3328))
```
iker-barriocanal, line 12: nit: floats are also accepted in `LenientString`
```suggestion
- Accept numbers in `event.user.username`. ([#3328](https://github.com/getsentry/relay/pull/3328))
```

iker-barriocanal (approved):

PR #3329, opened 2024-03-26
ref(cardinality): Moves limit state into its own module

Dav1dde:
Split from #3313 

Only takes the state and moves it to its own module and adds some documentation.

#skip-changelog

### DIFF
--- a/relay-cardinality/src/redis/state.rs
+++ b/relay-cardinality/src/redis/state.rs
@@ -1,0 +1,154 @@
+use relay_statsd::metric;
+
+use crate::{
+    limiter::{EntryId, Scoping},
+    redis::quota::QuotaScoping,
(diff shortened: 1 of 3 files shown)


relay-cardinality/src/redis/state.rs
```diff
@@ -0,0 +1,154 @@
+use relay_statsd::metric;
```
olksdr, line 1: Do you want to add module level docs as well?

jjbayer (approved):

olksdr (approved):

relay-cardinality/src/redis/state.rs
Dav1dde: I can but the module is private (like all submodules of the `redis` module) and don't show up on the built rust docs, I think all the relevant info is already on the individual structs and that's the ones you'd see in the IDE.

PR #3341, opened 2024-03-26
fix(statsd): Prefix kafka client metrics

jjbayer:
There was a naming collision for `kafka.message_size`, so I'm prefixing all client stats.

#skip-changelog

### DIFF
--- a/relay-kafka/src/statsd.rs
+++ b/relay-kafka/src/statsd.rs
@@ -86,14 +86,14 @@ impl GaugeMetric for KafkaGauges {
     fn name(&self) -> &'static str {
         match self {
             KafkaGauges::InFlightCount => "kafka.in_flight_count",
-            KafkaGauges::MessageCount => "kafka.message_count",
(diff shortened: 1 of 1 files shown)


Dav1dde (approved):

relay-kafka/src/statsd.rs
```diff
@@ -86,14 +86,14 @@ impl GaugeMetric for KafkaGauges {
     fn name(&self) -> &'static str {
         match self {
             KafkaGauges::InFlightCount => "kafka.in_flight_count",
```
Dav1dde, line 88: This one too?

jjbayer: This one does not technically come out of `statistics`, so I'm undecided.

Dav1dde: ah good point, then I agree it should not have the mid-fix

olksdr (approved):

PR #3342, opened 2024-03-26
feat(cardinality): Implement cardinality reporting

Dav1dde:
Makes the cardinality limiter able to report cardinality per "sub"-scope and return the collected information.

Cardinality Limits can now opt-in into recording/reporting cardinality per limit/generated scope. The reports will be accumulated and exposed through the `CardinalityResult`.

The collected reports are not yet used in Relay, it's a preparation to use the reports in a followup:

```rs
// Somewhere post cardinality limiter in the processor:
for (limit, report) in r.cardinality_reports() {
    self.metric_stats.track_cardinality(limit, report);
}
```

#skip-changelog


relay-cardinality/src/limiter.rs
```diff
@@ -19,10 +21,42 @@ pub struct Scoping {
     pub project_id: ProjectId,
 }
 
-/// Accumulator of all cardinality limiter rejections.
-pub trait Rejections<'a> {
+/// Cardinality report for a specific limit.
```
iker-barriocanal, line 24: What's reported if none of the options are available? (org id, project id, name)

relay-cardinality/src/limiter.rs
```diff
@@ -19,10 +21,42 @@ pub struct Scoping {
...
+
+/// Accumulator of all cardinality limiter decisions.
+pub trait Reporter<'a> {
     /// Called for ever [`Entry`] which was rejected from the [`Limiter`].
     fn reject(&mut self, limit: &'a CardinalityLimit, entry_id: EntryId);
+
+    /// Called for every individual limit applied.
+    ///
+    /// The callback can be called multiple times with different reports
+    /// for the same `limit` or not at all if there was no change in cardinality.
```
iker-barriocanal, line 58: > there was no change in cardinality.

What does that mean? Is this function called only when going to redis and not when using a cache?

relay-cardinality/src/limiter.rs
```diff
@@ -172,14 +215,16 @@ pub struct CardinalityLimits<'a, T> {
...
 impl<'a, T> CardinalityLimits<'a, T> {
-    fn new(source: Vec<T>, rejections: RejectionTracker<'a>) -> Self {
+    fn new(source: Vec<T>, reporter: DefaultReporter<'a>) -> Self {
         Self {
             source,
-            rejections: rejections.entries,
-            limits: rejections.limits,
+            rejections: reporter.entries,
+            limits: reporter.limits,
+            reports: reporter.reports,
```
iker-barriocanal: Should we make `reporter` an attribute of `CardinalityLimits`, instead of tracking these 3 separately?

relay-cardinality/src/redis/quota.rs
```diff
@@ -125,6 +126,34 @@ impl std::ops::Deref for QuotaScoping {
     }
 }
 
+// Required for hashbrown's `entry_ref`.
+impl From<&QuotaScoping> for QuotaScoping {
+    fn from(value: &QuotaScoping) -> Self {
+        value.clone()
+    }
+}
```
iker-barriocanal, line 134: Why not clone directly? Hiding this clone here may be harder to track if it becomes a problem in the future.

iker-barriocanal (approved):
Thanks for updating the docs!

relay-cardinality/src/limiter.rs
Dav1dde: All 3 options can be `None`, that would be a global limit which does not apply to an org, project and metric. I tried to explain that on the fields themselves, e.g.:

```rs
    /// Only available if the the limit was at least scoped to
    /// [`CardinalityScope::Organization`](crate::CardinalityScope::Organization).
```

Any ideas how I can make it more clear?

relay-cardinality/src/limiter.rs
Dav1dde: Exactly! Which comes down to, it is valid to only call the callback when the actual cardinality in the window increased.

relay-cardinality/src/redis/quota.rs
Dav1dde: See the comment above, this impl is required for `hashbrown`'s `entry_ref`. There is unfortunately no way around that if we want to use `entry_ref`.

relay-cardinality/src/limiter.rs
Dav1dde: Mh, I tried it just now, but it makes a bunch of lines break because the variables are now longer.

relay-cardinality/src/limiter.rs
```diff
@@ -172,14 +215,16 @@ pub struct CardinalityLimits<'a, T> {
     source: Vec<T>,
     rejections: HashSet<usize>,
     limits: HashSet<&'a CardinalityLimit>,
+    reports: BTreeMap<&'a CardinalityLimit, Vec<CardinalityReport>>,
```
jjbayer: Could this be a single map instead of a hash set and a btreemap, with an empty `Vec` if `limit.report` is false?

relay-cardinality/src/limiter.rs
```diff
@@ -19,10 +21,42 @@ pub struct Scoping {
...
     fn reject(&mut self, limit: &'a CardinalityLimit, entry_id: EntryId);
+
+    /// Called for every individual limit applied.
+    ///
+    /// The callback can be called multiple times with different reports
+    /// for the same `limit` or not at all if there was no change in cardinality.
+    ///
+    /// For example, with a name scoped limit can be called once for every
+    /// metric name matching the limit.
+    fn cardinality(&mut self, limit: &'a CardinalityLimit, report: CardinalityReport);
```
jjbayer, line 62:
```suggestion
    fn report_cardinality(&mut self, limit: &'a CardinalityLimit, report: CardinalityReport);
```

jjbayer (approved):

relay-cardinality/src/limiter.rs
iker-barriocanal: I'd add it to the struct's docstrings something like "if all the options are None, it's a global rate limit".

relay-cardinality/src/limiter.rs
Dav1dde: I just implemented it, but I it's actually wrong. There can be cardinality reports for limits which have not been exceeded.

I'll revert and rename the field to make it more clear!

PR #3344, opened 2024-03-26
feat(feedback): produce to ingest-feedback-events topic

aliu39:
Option PR:
depends on https://github.com/getsentry/sentry/pull/67839 (MERGED)

Feature flag PRs (outdated/unused):
~https://github.com/getsentry/getsentry/pull/13426 (REVERTED)~
~https://github.com/getsentry/sentry/pull/67747 (REVERTED)~
~https://github.com/getsentry/sentry-options-automator/pull/973 (CLOSED)~

Relates to https://github.com/getsentry/sentry/issues/66100


Dav1dde:
You're on a pretty old base for this branch, I recommend updating. The Kafka Topic config recently changed and your change would not work in SaaS.

.gitignore
```diff
@@ -14,6 +14,7 @@ target
 # Editors
 /.idea/
 .DS_Store
+.vscode/
```
Dav1dde: While I agree with this change, shouldn't do this in a feature PR.

You can locally in your own git config ignore files or even ignore changes to files instead.

relay-dynamic-config/src/global.rs
```diff
@@ -156,6 +156,17 @@ pub struct Options {
     )]
     pub metric_stats_rollout_rate: f32,
 
+    /// Rollout rate for producing to ingest-feedback-topic.
+    ///
+    /// Rate needs to be between `0.0` and `1.0`.
+    /// If set to `1.0` all organizations will ingest to the feedback topic
```
Dav1dde:
```suggestion
    /// If set to `1.0` all organizations will ingest to the feedback topic.
```

relay-kafka/src/config.rs
```diff
@@ -124,6 +127,8 @@ pub struct TopicAssignments {
     /// COGS measurements.
     #[serde(alias = "shared-resources-usage")]
     pub cogs: TopicAssignment,
+    /// Feedback events topic name.
+    pub feedback: TopicAssignment,
```
Dav1dde: Probably missing a `#[serde(alias = "ingest-feedback-events")]`, assuming that's what the topic is called in the ops repo/Kafka schema.

Please double check this!

relay-server/src/services/store.rs
```diff
@@ -184,6 +184,14 @@ impl StoreService {
     ) -> Result<(), StoreError> {
         let retention = envelope.retention();
         let event_id = envelope.event_id();
+
+        let use_ingest_feedback_topic = self
+            .global_config
+            .current()
+            .options
+            .ingest_topic_rollout_rate
+            > 0.0;
```
Dav1dde: That's not a rollout rate, it's a boolean. There is `crate::utils::is_rolled_out` if you want a rollout rate, otherwise please change everything to a boolean enabled/disabled option.

aliu39: Thanks! I wasn't sure how to use the rollout rate yesterday

relay-dynamic-config/src/global.rs
```diff
@@ -156,6 +156,17 @@ pub struct Options {
...
+    /// Rollout rate for producing to ingest-feedback-topic.
+    ///
+    /// Rate needs to be between `0.0` and `1.0`.
+    /// If set to `1.0` all organizations will ingest to the feedback topic.
+    #[serde(
+        rename = "feedback.ingest-topic.rollout-rate",
+        deserialize_with = "default_on_error",
+        skip_serializing_if = "is_default"
+    )]
+    pub ingest_topic_rollout_rate: f32,
```
cmanallen: Can we namespace this field?  `feedback_ingest...`

cmanallen (changes requested):

relay-server/src/services/store.rs
```diff
@@ -184,6 +186,15 @@ impl StoreService {
...
         let retention = envelope.retention();
         let event_id = envelope.event_id();
+
+        let feedback_ingest_topic_rollout_rate = self
+            .global_config
+            .current()
+            .options
+            .feedback_ingest_topic_rollout_rate;
+        let use_feedback_topic =
+            is_rolled_out(scoping.organization_id, feedback_ingest_topic_rollout_rate);
```
cmanallen: This will cause the rollout load volume to be unbalanced.  Large orgs will send significantly more messages than small orgs.  This might be okay.  Be aware going from 10% -> 20% does not necessarily indicate a 2x increase in volume.

tests/integration/fixtures/mini_sentry.py
```diff
@@ -182,6 +182,12 @@ def add_full_project_config(self, project_id, dsn_public_key=None, extra=None):
         self.project_configs[project_id] = ret_val
         return ret_val
 
+    def set_global_config_option(self, option_name, value):
+        # must be called before initializing relay fixture
+        old_val = self.global_config["options"].get(option_name)
+        self.global_config["options"][option_name] = value
+        return old_val
```
cmanallen: Why return the old value?

cmanallen (approved):

relay-server/src/services/store.rs
aliu39: 10 -> 20 will equal a 2x increase in volume from any _given org_, right? But not overall.

tests/integration/fixtures/mini_sentry.py
aliu39: Just some extra functionality for anyone using the method in the future

tests/integration/test_feedback.py
```diff
@@ -42,24 +43,33 @@ def generate_feedback_sdk_event():
...
     )
 
-    _events_consumer = events_consumer(timeout=20)
-    feedback = generate_feedback_sdk_event()
+    if use_feedback_topic:
+        mini_sentry.set_global_config_option("feedback.ingest-topic.rollout-rate", 1.0)
+        consumer = feedback_consumer(timeout=20)
+    else:
+        mini_sentry.set_global_config_option("feedback.ingest-topic.rollout-rate", 0.0)
+        consumer = events_consumer(timeout=20)
```
jjbayer, line 64: Might be worth also assigning `other_consumer` and calling `.assert_empty()` on it at the end of the test.

relay-server/src/services/store.rs
```diff
@@ -199,6 +210,10 @@ impl StoreService {
             KafkaTopic::Attachments
         } else if event_item.as_ref().map(|x| x.ty()) == Some(&ItemType::Transaction) {
             KafkaTopic::Transactions
+        } else if use_feedback_topic
```
jjbayer: nit: `is_rolled_out` is called for every envelope, even if it does not contain a UserReportV2. So I would turn this condition around and evaluate `is_rolled_out` only when the event type matches.

jjbayer (approved):
LGTM. Please make sure the consumer side is ready before enabling the feature flag.

tests/integration/fixtures/mini_sentry.py
aliu39: Actually removing it, don't see why anyone would use this haha

jjbayer (approved):

relay-server/src/services/store.rs
```diff
@@ -199,6 +202,17 @@ impl StoreService {
             KafkaTopic::Attachments
         } else if event_item.as_ref().map(|x| x.ty()) == Some(&ItemType::Transaction) {
             KafkaTopic::Transactions
+        } else if event_item.as_ref().map(|x| x.ty()) == Some(&ItemType::UserReportV2) {
+            let feedback_ingest_topic_rollout_rate = self
+                .global_config
+                .current()
+                .options
+                .feedback_ingest_topic_rollout_rate;
+            if is_rolled_out(scoping.organization_id, feedback_ingest_topic_rollout_rate) {
```
iker-barriocanal, line 211: This is checking the rollout rate with the org, not the event -- let's please roll out slowly.

iker-barriocanal (approved):

relay-server/src/services/store.rs
Dav1dde: I think it's just used as a boolean on/off in practice, so should be fine 👍

relay-server/src/services/store.rs
```diff
@@ -199,6 +202,17 @@ impl StoreService {
             KafkaTopic::Attachments
         } else if event_item.as_ref().map(|x| x.ty()) == Some(&ItemType::Transaction) {
             KafkaTopic::Transactions
+        } else if event_item.as_ref().map(|x| x.ty()) == Some(&ItemType::UserReportV2) {
+            let feedback_ingest_topic_rollout_rate = self
+                .global_config
+                .current()
+                .options
+                .feedback_ingest_topic_rollout_rate;
```
Dav1dde, line 210: tiny nit:

```suggestion
            let global_config = self.global_config.current();
            let topic_rollout_rate = global_config.options.feedback_ingest_topic_rollout_rate;
```

I like splitting this, so it takes up less lines and becomes a bit more readable. But need to check if rustfmt aggrees with me here ...

Dav1dde (approved):

relay-server/src/services/store.rs
iker-barriocanal: Not sure what the feedback event volume is, but suddenly going from 0 to N may cause a backlog until consumers scale up.

aliu39: Is there a function I can use to do it on the event level?

aliu39: We will probably keep it at the org level since feedback volumes aren't that high, and we'd like to test it on all feedback in s4s. Will comment here if that changes. Holding off merging until https://getsentry.atlassian.net/browse/OPS-5543 is done.

iker-barriocanal: > We will probably keep it at the org level since feedback volumes aren't that high

That's totally fine, my comment was just a note to roll out slowly to avoid any surprises 😄

PR #3347, opened 2024-03-27
feat(server): Expose rate limits for metrics

jan-auer:
Adds a new rate limit component to quota limits in the `x-sentry-rate-limits` header that communicates the namespace(s) that a metrics rate limit should apply to. The data category of limits that have this component is always `metric_bucket`, in other quotas this component will not be set. 

Clients will receive rate limit entries in the response of every envelope that contains a metric item, regardless of whether the contents match the specific namespace. This is a performance optimization so Relay can avoid to parse item contents in the fast path while the request is still open.

**Limitations**:
- As soon as a single metric rate limit is active, *all* envelope requests containing a metrics item will result in a 429 status code. It is the client's responsibility to parse the header and ignore irrelevant rate limits.
- These rate limits are not propagated on the batch metrics endpoint. External Relays do not use this endpoint and send envelopes instead, so backoff will work as expected. However, if external Relays would start to use that endpoint, additional changes would be necessary as they also do not receive quotas.

Documentation is added via https://github.com/getsentry/develop/pull/1205

Epic: https://github.com/getsentry/team-ingest/issues/303


relay-server/src/services/project.rs
```diff
@@ -1102,11 +1102,19 @@ impl Project {
...
             envelope_limiter.enforce(envelope.envelope_mut(), &scoping)?;
         enforcement.track_outcomes(envelope.envelope(), &scoping, outcome_aggregator);
         envelope.update();
 
+        // Special case: Expose active rate limits for all metric namespaces if there is at least
+        // one metrics item in the Envelope to communicate backoff to SDKs. This is necessary
+        // because `EnvelopeLimiter` cannot not check metrics without parsing item contents.
+        if envelope.envelope().items().any(|i| i.ty().is_metrics()) {
+            let metrics_scoping = scoping.item(DataCategory::MetricBucket);
+            rate_limits.merge(self.rate_limits.check_with_quotas(quotas, metrics_scoping));
```
jan-auer, line 1116: This does not work yet. The scoping does not have a namespace, hence no quotas with namespaces match.

jan-auer: Updated. We now check with `MetricNamespaceScoping::Any`, which will resolve all metrics. This also makes the API a bit more explicit by adding specific doc comments for the semantics of `None`.

relay-server/src/services/processor.rs
```diff
@@ -2010,44 +2012,31 @@ impl EnvelopeProcessorService {
             // We set over_accept_once such that the limit is actually reached, which allows subsequent
             // calls with quantity=0 to be rate limited.
             let over_accept_once = true;
+            let mut rate_limits = RateLimits::new();
+
+            for category in [DataCategory::Transaction, DataCategory::Span] {
```
jan-auer, line 2017: Drive-by rewrite for readability.

relay-server/src/services/project.rs
```diff
@@ -1178,15 +1187,10 @@ impl Project {
...
-            scoping: &scoping,
-            namespace: None,
-        };
-
-        let limits = self
-            .rate_limits()
-            .check_with_quotas(project_state.get_quotas(), item_scoping);
+        let limits = self.rate_limits().check_with_quotas(
+            project_state.get_quotas(),
+            scoping.item(DataCategory::MetricBucket),
```
jan-auer, line 1192: This might be a bug, filed in #3354

relay-quotas/src/rate_limit.rs
```diff
@@ -504,7 +511,57 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_rate_limit_matches_namespaces() {
+        let rate_limit = RateLimit {
+            categories: smallvec![],
+            scope: RateLimitScope::Organization(42),
+            reason_code: None,
+            retry_after: RetryAfter::from_secs(1),
+            namespace: smallvec![MetricNamespace::Custom],
```
jjbayer: Should we run the same test against a RateLimit with empty namespaces?

relay-quotas/src/rate_limit.rs
```diff
@@ -156,8 +157,11 @@ pub struct RateLimit {
     /// A marker when this rate limit expires.
     pub retry_after: RetryAfter,
 
-    /// The namespace of this rate limit.
-    pub namespace: Option<MetricNamespace>,
+    /// The metric namespace of this rate limit.
+    ///
+    /// Ignored on all data categories except for `MetricBucket`. If empty, this rate limit applies
+    /// to metrics of all namespaces.
+    pub namespace: SmallVec<[MetricNamespace; 1]>,
```
jjbayer: This field is not part of any protocol, right? So we could rename it.
```suggestion
    pub namespaces: SmallVec<[MetricNamespace; 1]>,
```

relay-quotas/src/rate_limit.rs
```diff
@@ -679,7 +736,59 @@ mod tests {
...
+
+        rate_limits.add(RateLimit {
+            categories: smallvec![DataCategory::MetricBucket],
+            scope: RateLimitScope::Organization(42),
+            reason_code: None,
+            retry_after: RetryAfter::from_secs(1),
+            namespace: smallvec![MetricNamespace::Custom],
+        });
+
+        // Same scope but different categories
```
jjbayer:
```suggestion
        // Same category but different namespace
```

relay-server/src/services/processor.rs
jjbayer: Thanks!

relay-server/src/utils/rate_limits.rs
```diff
@@ -35,6 +35,13 @@ pub fn format_rate_limits(rate_limits: &RateLimits) -> String {
 
         if let Some(ref reason_code) = rate_limit.reason_code {
             write!(header, ":{reason_code}").ok();
+        } else if !rate_limit.namespace.is_empty() {
+            write!(header, ":").ok();
```
jjbayer: It took me a while to understand this code.

```suggestion
            write!(header, ":").ok(); // delimits the empty reason code
```

relay-server/src/utils/rate_limits.rs
```diff
@@ -842,6 +889,32 @@ mod tests {
...
+    #[test]
+    fn test_parse_rate_limits_namespace() {
+        let scoping = Scoping {
+            organization_id: 42,
+            project_id: ProjectId::new(21),
+            project_key: ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap(),
+            key_id: Some(17),
+        };
+
+        // contains "foobar", an unknown scope that should be mapped to Unknown
```
jjbayer: Is this a comment from a different test?

relay-server/src/utils/rate_limits.rs
```diff
@@ -842,6 +889,32 @@ mod tests {
...
+        let formatted = "42:metric_bucket:organization::custom;spans";
```
jjbayer, line 901: Should we test `:reason_code:` as well?

jjbayer (approved):
Just a couple of nits.

relay-server/src/utils/rate_limits.rs
jan-auer: Thanks, the comment helps. I was hoping to rewrite this in a better way, but everything else I could come up with ended up much more complex.

relay-quotas/src/rate_limit.rs
jan-auer: There are tests above that test with empty list on different data categories -- a rate limit on MetricBuckets without namespace would actually be semantically incorrect. Though it will work as the other tests ensure.

PR #3349, opened 2024-03-27
instr(kafka): More broker stats

jjbayer:
Follow-up to #3315: Also emit `int_latency` and `outbuf_latency` metrics. 

#skip-changelog

### DIFF
--- a/relay-kafka/src/producer/utils.rs
+++ b/relay-kafka/src/producer/utils.rs
@@ -20,7 +20,7 @@ impl ClientContext for Context {
         relay_statsd::metric!(gauge(KafkaGauges::MessageSize) = statistics.msg_size);
         relay_statsd::metric!(gauge(KafkaGauges::MessageSizeMax) = statistics.msg_size_max);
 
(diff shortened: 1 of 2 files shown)


relay-kafka/src/producer/utils.rs
```diff
@@ -20,7 +20,7 @@ impl ClientContext for Context {
         relay_statsd::metric!(gauge(KafkaGauges::MessageSize) = statistics.msg_size);
         relay_statsd::metric!(gauge(KafkaGauges::MessageSizeMax) = statistics.msg_size_max);
 
-        for broker in statistics.brokers.values() {
+        for (_, broker) in statistics.brokers {
```
Dav1dde, line 23: Why this change?

Dav1dde (approved):

relay-kafka/src/producer/utils.rs
jjbayer: It was either this or `if let Some(int_latency) = &broker.int_latency {`, and I thought this function might as well consume the object.

PR #3350, opened 2024-03-27
fix: Fix kafka topic default

lynnagara:
should be ingest-performance-metrics not ingest-generic metrics

INC-696


olksdr (approved):

relay-kafka/src/config.rs
```diff
@@ -102,7 +102,7 @@ pub struct TopicAssignments {
     #[serde(alias = "metrics", alias = "ingest-metrics")]
     pub metrics_sessions: TopicAssignment,
     /// Topic name for all other kinds of metrics. Defaults to the assignment of `metrics`.
-    #[serde(alias = "metrics_transactions", alias = "ingest-generic-metrics")]
+    #[serde(alias = "metrics_transactions", alias = "ingest-performance-metrics")]
```
Dav1dde, line 105: Are we sure it's not both, e.g. single tenant?

In the incident channel you said it worked on single tenant, if I search for `ingest-generic-metrics` I do find some results in the ops repo.

Dav1dde (changes requested):

relay-kafka/src/config.rs
lynnagara: Yeah, the key here is the logical or default topic, what can get overridden to in each environment (such as different single tenants) via the `TopicAssignment` value.

lynnagara: I need to find a way to validate these against what we have registered in https://github.com/getsentry/sentry-kafka-schemas/blob/main/topics/ingest-performance-metrics.yaml to avoid this happening in the future.

Dav1dde (approved):

relay-kafka/src/config.rs
Dav1dde: We're already pulling in `sentry-kafka-schemas`, can probably use it to validate the topics in a test?

lynnagara: Sounds good, I need to spend some time to figure how to make this work + also try to remove the duplication of default topics in the file -- will follow up with something shortly.

PR #3355, opened 2024-03-28
ref(normalization): Remove duplicated normalization

iker-barriocanal:
When light normalization was introduced, the normalization of `platform` and `level` was duplicated from the store normalization. Now that there's a single normalization step, this PR removes that duplication.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -1515,7 +1515,7 @@ mod tests {
 
         assert_eq!(event.level.value().unwrap().to_string(), "info");
         assert_eq!(event.ty.value().unwrap().to_string(), "transaction");
-        assert_eq!(event.platform.value().unwrap().to_owned(), "other");
+        assert_eq!(event.platform.as_str().unwrap(), "other");
     }
 
     #[test]
--- a/relay-event-normalization/src/event.rs
+++ b/relay-event-normalization/src/event.rs
@@ -278,7 +278,6 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) {
     normalize_breadcrumbs(event);
     normalize_release_dist(event); // dist is a tag extracted along with other metrics from transactions
     normalize_event_tags(event); // Tags are added to every metric
-    normalize_platform_and_level(event);
 
     // TODO: Consider moving to store normalization
     if config.device_class_synthesis_config {
@@ -575,18 +574,6 @@ fn normalize_dist(distribution: &mut Annotated<String>) {
             *dist = trimmed.to_string();
         }
         Ok(())
-    });
-}
-
-/// Defaults the `platform` and `level` required attributes.
-fn normalize_platform_and_level(event: &mut Event) {
-    // The defaulting behavior, was inherited from `StoreNormalizeProcessor` and it's put here since only light
-    // normalization happens before metrics extraction and we want the metrics extraction pipeline to already work
-    // on some normalized data.
-    event.platform.get_or_insert_with(|| "other".to_string());
-    event.level.get_or_insert_with(|| match event.ty.value() {
-        Some(EventType::Transaction) => Level::Info,
-        _ => Level::Error,
     });
 }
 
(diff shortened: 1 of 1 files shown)


relay-event-normalization/src/event.rs
```diff
@@ -578,18 +577,6 @@ fn normalize_dist(distribution: &mut Annotated<String>) {
     });
 }
 
-/// Defaults the `platform` and `level` required attributes.
-fn normalize_platform_and_level(event: &mut Event) {
```
iker-barriocanal, line 582: These are normalized in
https://github.com/getsentry/relay/blob/9bf4cbe9a5a2be8ac3d436a666724fd0586d1b17/relay-event-normalization/src/event.rs#L1030-L1033

relay-event-normalization/src/event.rs
```diff
@@ -1519,38 +1506,39 @@ mod tests {
...
 
     #[test]
     fn test_normalize_platform_and_level_with_error_event() {
         let json = r#"
         {
-            "type": "error"
+            "type": "error",
+            "exception": {
+                "values": [{"type": "ValueError", "value": "Should not happen"}]
+            }
```
iker-barriocanal, line 1528: `type` is not enough for an event to be considered an error, see 
https://github.com/getsentry/relay/blob/9bf4cbe9a5a2be8ac3d436a666724fd0586d1b17/relay-event-normalization/src/event.rs#L1057

relay-event-normalization/src/event.rs
```diff
@@ -1519,38 +1506,39 @@ mod tests {
...
-          "platform": "other",
-        }
-        "###);
+        assert_eq!(event.level.value().unwrap().to_string(), "info");
+        assert_eq!(event.ty.value().unwrap().to_string(), "transaction");
+        assert_eq!(event.platform.value().unwrap().to_owned(), "other");
     }
 
     #[test]
     fn test_normalize_platform_and_level_with_error_event() {
```
Dav1dde, line 1522: Test names are no longer correct.

relay-event-normalization/src/event.rs
```diff
@@ -278,7 +278,6 @@ fn normalize(event: &mut Event, meta: &mut Meta, config: &NormalizationConfig) {
     normalize_breadcrumbs(event);
     normalize_release_dist(event); // dist is a tag extracted along with other metrics from transactions
     normalize_event_tags(event); // Tags are added to every metric
-    normalize_platform_and_level(event);
```
Dav1dde, line 281: Is order a concern here, there are 8 normalize functions in between this and `normalize_default_attributes`, do any depend on the result from here?

relay-event-normalization/src/event.rs
```diff
@@ -1519,38 +1506,39 @@ mod tests {
...
+        normalize_default_attributes(&mut event, &mut meta, &NormalizationConfig::default());
 
-        insta::assert_ron_snapshot!(SerializableAnnotated(&Annotated::new(event)), {}, @r###"
-        {
-          "level": "info",
-          "type": "transaction",
-          "platform": "other",
-        }
-        "###);
+        assert_eq!(event.level.value().unwrap().to_string(), "info");
```
Dav1dde, line 1516:
```suggestion
        assert_eq!(event.level.as_str(), Some("info"));
```

Looking at the `Annotated` signatures, this should work.

relay-event-normalization/src/event.rs
iker-barriocanal: Nope, no dependencies.

relay-event-normalization/src/event.rs
iker-barriocanal: What's no longer correct? Tests are still normalizing the platform and the level given an error/transaction event.

iker-barriocanal: Is this about updating the snapshot to assert specific items? If yes, I made that change to bring tests a bit closer to [Relay guidelines](https://www.notion.so/sentry/Relay-Code-and-Architecture-4fac481cf27741aba54d39594f1cae58?pvs=4#fd14645580f848d9969d124a59cc3f7d):

> Keep unit tests constrained to a single concern. Do not test multiple things in one test

relay-event-normalization/src/event.rs
Dav1dde: Thanks for checking!

relay-event-normalization/src/event.rs
Dav1dde: The function `normalize_platform_and_level` doesn't exist anymore, fine with keeping it if you think that's the best name.

Sorry worded that badly.

Dav1dde (approved):

relay-event-normalization/src/event.rs
iker-barriocanal: Updated the last entry. It's not possible as the others are not `String`s but enums and don't implement `AsRef`. `to_string` goes through the `Display` implementation, where we have the string mapping defined.

PR #3357, opened 2024-03-28
ref(cardinality): Don't call Redis when its not necessary

Dav1dde:
There is no need to call Redis without any data. Without the change we would still call into redis-rs even when the cache already handled everything.

Makes sure to skip without adding a `CardinalityLimiterTimers::Redis` timing.

#skip-changelog

### DIFF
--- a/relay-cardinality/src/redis/state.rs
+++ b/relay-cardinality/src/redis/state.rs
@@ -82,6 +82,15 @@ impl<'a> LimitState<'a> {
         self.scopes.entry(scope).or_default().push(entry)
     }
 
+    /// Returns `true` if this state does not contain any entries.
+    ///
+    /// The state can be empty if:
+    /// - there are no entries matching this limit.
+    /// - all entries matching this limit were already handled by the cache.
+    pub fn is_empty(&mut self) -> bool {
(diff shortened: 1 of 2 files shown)


relay-cardinality/src/redis/state.rs
```diff
@@ -82,6 +82,11 @@ impl<'a> LimitState<'a> {
         self.scopes.entry(scope).or_default().push(entry)
     }
 
+    /// Returns `true` if this state does not contain any entries.
+    pub fn is_empty(&mut self) -> bool {
+        self.scopes.is_empty()
```
jjbayer, line 91: Under what circumstances is `self.scopes` empty?

jjbayer (approved):
Whe

relay-cardinality/src/redis/state.rs
Dav1dde: We build the state for all limits, it is empty:

1) if there are no buckets that match a limit (e.g. due to the namespace filter), 
2) if all buckets were already handled by the cache

I'll add it to the doc string!

iker-barriocanal (approved):

PR #3358, opened 2024-03-28
feat(metrics): Unicode escapes for tag values in statsd

jan-auer:
**Disclaimer**: See [the epic](https://github.com/getsentry/team-ingest/issues/304) for more context.

Adds support for escape sequences based on best practices recommended in [RFC 5137](https://www.rfc-editor.org/rfc/rfc5137).

Tag values allow all printable unicode characters. Control characters have to be stripped from output. There is a list of restricted characters that have to be escaped according to the following mapping:
 - Tab is escaped as `\t`.
 - Carriage return is escaped as `\r`.
 - Line feed is escaped as `\n`.
 - Backslash is escaped as `\\`.
 - Commas and pipes are given unicode escapes in the form `\u{2c}` and `\u{7c}`, respectively.

**Note**: Documentation previously allowed all `\s` in tag values, which
included newlines. These are not legal and would lead to invalid statsd
payloads. Newlines and all whitespace characters other than a plain space now
require an escape sequence and will be removed as control characters.

Epic: https://github.com/getsentry/team-ingest/issues/304


relay-metrics/src/aggregator.rs
```diff
@@ -95,7 +95,7 @@ impl BucketKey {
 /// Note that this does not necessarily match the exact memory footprint of the tags,
 /// because data structures or their serialization have overheads.
 pub fn tags_cost(tags: &BTreeMap<String, String>) -> usize {
-    tags.iter().map(|(k, v)| k.capacity() + v.capacity()).sum()
+    tags.iter().map(|(k, v)| k.len() + v.len()).sum()
```
jan-auer, line 98: This is an unrelated bugfix that showed up in one of the tests: By using capacity, we were overestimating the size of the buckets. Due to the introduction of `unescaper::unescape`, the allocated tag strings were larger than their contents, which changed test behavior.

relay-metrics/src/bucket.rs
```diff
@@ -538,7 +537,7 @@ pub struct Bucket {
     ///
     /// Namespaces and units must consist of ASCII characters and match the regular expression
     /// `/\w+/`. The name component of MRIs consist of unicode characters and must match the
-    /// regular expression `/\w[\w\d_-.]+/`. Note that the name must begin with a letter.
+    /// regular expression `/\w[\w\-.]*/`. Note that the name must begin with a letter.
```
jan-auer, line 540: The validation/normalization implementation in Relay differs from the rules for legal metric names:
- We do not yet enforce the first letter. This will come in as a separate refactor.
- Dashes are intentionally replaced with underscores. Right now, the product wants just underscores, but we still permit SDKs to send dashes for future proofing.

relay-metrics/src/aggregator.rs
Dav1dde: Mh I wonder if this wasn't deliberate, because capacity is the cost of the metric in memory (just doesn't work for serialization).

jan-auer: It was deliberate, I originally used `capacity()` since we used this to control memory consumption. More than that, we now rely on this for serialization now and Joris has since changed the corresponding check on metric names to use `.len()`. So it should be fine to use `len`. 

If we want robust memory measurements, I'm afraid we'll have to explore different approaches like arena allocators.

jjbayer: > Joris has since changed the corresponding check on metric names to use .len()

@jan-auer are you referring to this line? The reason why it uses `.len()` is because metric names are now represented by `Arc<str>`, not `String` (see https://github.com/getsentry/relay/pull/3279).

https://github.com/getsentry/relay/blob/e614e9127c6909bd0a4acca464705dcebdc101e4/relay-metrics/src/aggregator.rs#L84

No heavy objections against this change, but the cleanest solution would be to have two different estimation functions for memory footprint and serialization cost.

relay-metrics/src/bucket.rs
jjbayer: > Dashes are intentionally replaced with underscores

Should we add this information to the doc comment here?

jjbayer, line 540: Should the `.` be escaped? We mean literal dot, not any character, right?

relay-base-schema/src/metrics/mod.rs
```diff
@@ -25,7 +25,7 @@ pub fn try_normalize_metric_name(name: &str) -> Option<Cow<'_, str>> {
         return None;
     }
 
-    let normalize_re = NORMALIZE_RE.get_or_init(|| Regex::new("[^a-zA-Z0-9_/.]+").unwrap());
+    let normalize_re = NORMALIZE_RE.get_or_init(|| Regex::new("[^a-zA-Z0-9_.]+").unwrap());
```
jjbayer, line 30: Why is `/` not replaced anymore?

Also, should the `.` be escaped? Or is it not interpreted as a wildcard when it occurs within `[...]`?

relay-system/Cargo.toml
```diff
@@ -17,7 +17,7 @@ futures = { workspace = true }
 once_cell = { workspace = true }
 relay-log = { workspace = true }
 relay-statsd = { workspace = true }
-tokio = { workspace = true, features = ["rt", "signal"] }
+tokio = { workspace = true, features = ["rt", "signal", "macros"] }
```
jjbayer, line 20: What is this for?

jjbayer (approved):

relay-metrics/src/bucket.rs
```diff
@@ -413,10 +413,9 @@ fn parse_tags(string: &str) -> Option<BTreeMap<String, String>> {
             continue;
         }
 
-        let mut value = name_value.next().unwrap_or_default().to_owned();
-        protocol::validate_tag_value(&mut value);
-
-        map.insert(name.to_owned(), value);
+        if let Ok(value) = protocol::unescape_tag_value(name_value.next().unwrap_or_default()) {
```
Dav1dde, line 416: I think we should return `None` if the unescaping fails.

```suggestion
        let value = protocol::unescape_tag_value(name_value.next().unwrap_or_default()).ok()?;
```

Seems to be what we do elsewhere as well.

relay-base-schema/src/metrics/mod.rs
Dav1dde: > Why is / not replaced anymore?

👀 

> Or is it not interpreted as a wildcard when it occurs within [...]?

It's a literal `.` in a match group.

relay-metrics/src/protocol.rs
```diff
@@ -54,3 +97,59 @@ pub(crate) fn hash_set_value(string: &str) -> u32 {
     hasher.write(string.as_bytes());
     hasher.finish32()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_unescape_tag_value() {
```
Dav1dde, line 107: We should should have some tests:

- for characters outside of ascii (chinese etc.)
- for other control characters, that they are replaced correctly (the function still calls into the old `validate`)

relay-metrics/src/bucket.rs
```diff
@@ -1132,6 +1138,18 @@ mod tests {
             "#);
     }
 
+    #[test]
+    fn test_parse_tags_escaped() {
```
Dav1dde, line 1142: A test which parses tags outside of ascii would be nice

relay-metrics/src/bucket.rs
jan-auer: Inside of character groups, dot is literal.

relay-base-schema/src/metrics/mod.rs
jan-auer: `/` was exempt from replacing and is now replaced. The original spec allowed slashes, but for a long while SDK guidelines have excluded it. We're now aligning with the latest character set described in the epic and on develop docs.

relay-system/Cargo.toml
jan-auer: It fixes a compilation issue when the relay-system crate is compiled alone (e.g. to run tests). The crate actually depends on macros, we just didn't notice before as another workspace member enables this feature usually.

relay-metrics/src/bucket.rs
jan-auer: Actually, I just found that the first point is wrong. We do validate that the first character is a letter.

I've described this more clearly in the docs of `try_normalize_metric_name` now. To readers of the public Rust docs (i.e. client implementors) this information ideally shouldn't matter.

relay-metrics/src/bucket.rs
jan-auer: Just a few lines above, we skip individual tags with invalid keys. Are you sure we should break out? That would fail parsing the entire metric.

jan-auer:
Thanks for the reviews. I've addressed most comments. Going to merge once we reach consensus on https://github.com/getsentry/relay/pull/3358#discussion_r1553097585

relay-metrics/src/bucket.rs
Dav1dde: Sorry I missed that, let's keep the behaviour!

PR #3360, opened 2024-03-28
feat(metric-stats): Report cardinality to metric stats

Dav1dde:
Emits collected cardinality reports as metric stats.

Changed cardinality limit to a `u32`, this is still way too big for us ever to reach it, but it makes the conversion to f64 easier.

### DIFF
--- a/relay-server/src/metric_stats.rs
+++ b/relay-server/src/metric_stats.rs
@@ -58,7 +58,7 @@ impl MetricStats {
     }
 
     /// Tracks the metric volume and outcome for the bucket.
-    pub fn track(&self, scoping: Scoping, bucket: Bucket, outcome: Outcome) {
+    pub fn track_metric(&self, scoping: Scoping, bucket: Bucket, outcome: Outcome) {
         if !self.is_enabled(scoping) {
             return;
         }
@@ -78,7 +78,7 @@ impl MetricStats {
     }
 
     /// Tracks the cardinality of a metric.
-    pub fn cardinality(
+    pub fn track_cardinality(
         &self,
         scoping: Scoping,
         limit: &CardinalityLimit,
@@ -239,10 +239,10 @@ mod tests {
         let scoping = scoping();
         let mut bucket = Bucket::parse(b"rt@millisecond:57|d", UnixTimestamp::now()).unwrap();
 
-        ms.track(scoping, bucket.clone(), Outcome::Accepted);
+        ms.track_metric(scoping, bucket.clone(), Outcome::Accepted);
 
         bucket.metadata.merges = bucket.metadata.merges.saturating_add(41);
-        ms.track(
+        ms.track_metric(
             scoping,
             bucket,
             Outcome::RateLimited(Some(ReasonCode::new("foobar"))),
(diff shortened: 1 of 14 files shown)


relay-server/src/metric_stats.rs
```diff
@@ -66,6 +77,35 @@ impl MetricStats {
             .send(MergeBuckets::new(scoping.project_key, vec![volume]));
     }
 
+    /// Tracks the cardinality of a metric.
+    pub fn cardinality(
```
iker-barriocanal: Wdyt about `track_cardinality`? I find it more predictable from the API's user pov, even if it's similar to the `track` above.

relay-server/src/metric_stats.rs
```diff
@@ -18,6 +21,14 @@ fn volume_metric_mri() -> MetricName {
         .clone()
 }
 
+fn cardinality_metric_mri() -> MetricName {
+    static CARDINALITY_METRIC_MRI: OnceLock<MetricName> = OnceLock::new();
+
+    CARDINALITY_METRIC_MRI
+        .get_or_init(|| "g:metric_stats/cardinality@none".into())
+        .clone()
+}
```
iker-barriocanal, line 30: Could we make this function a `const`? Resolving the return value at compile time allows removing the `OnceLock`.

relay-server/src/metric_stats.rs
```diff
@@ -109,6 +149,42 @@ impl MetricStats {
...
+                "cardinality.window".to_owned(),
+                limit.window.window_seconds.to_string(),
+            ),
+        ]);
+
+        Some(Bucket {
+            timestamp: UnixTimestamp::now(),
+            width: 0,
+            name: cardinality_metric_mri(),
+            value: BucketValue::Gauge(GaugeValue::single(cardinality.into())),
```
iker-barriocanal, line 183: Why is this a gauge?

relay-server/src/metric_stats.rs
jjbayer: `Arc` does not have a `const` constructor.

relay-server/src/metric_stats.rs
jjbayer: +1, I left a similar comment here: https://github.com/getsentry/relay/pull/3342/files#r1542984095

relay-server/src/metric_stats.rs
jjbayer: Multiple Relay instances will report cardinality for the same window and tag combination. The only useful way to aggregate these multiple reports is by taking the `max`, for which a gauge is good enough.

jjbayer (approved):

relay-server/src/metric_stats.rs
iker-barriocanal: How is that different from a counter emitted every x seconds? I assume the gauges have different tag values for the pod names, so they don't aggregate together.

Dav1dde: The counter adds values, we specifically need the max per hour. So we might be emitting the values `1, 2, 3, 4, 5` and we're interested in just `5` not the sum of `15`.

relay-server/src/metric_stats.rs
Dav1dde: @jjbayer sorry, I remember renaming it not sure what happened to it :(

relay-server/src/metric_stats.rs
iker-barriocanal: > need the max per hour

I assumed a lower granularity so the usage of a counter was maaaybe worth it

PR #3361, opened 2024-03-29
feat: Remove default topic duplication and add kafka topic test

lynnagara:
Fixes related to INC-696, which accidentally had `ingest-generic-metrics` as the default topic name alias instead of `ingest-performance-metrics`.

This PR attempts to do 2 things
- Move defining the default topic name to a single place rather than 2
- Check the default topic names against those registered against sentry-kafka-schemas to ensure they are aligned with what the other codebases have registered as defaults

Needs https://github.com/getsentry/sentry-kafka-schemas/pull/242

Note that `outcomes-billing` no longer defaults to `outcomes` and `ingest-performance-metrics` no longer defaults to `ingest-metrics`, as there is no scenario where they should share topics anymore.


lynnagara:
a bit stumped by the failing integration test, any ideas how to fix it?

relay-kafka/src/config.rs
```diff
@@ -80,93 +80,62 @@ impl KafkaTopic {
...
+        metrics_generic: (KafkaTopic::MetricsGeneric, "ingest-performance-metrics"),
+        profiles: (KafkaTopic::Profiles, "profiles"),
+        replay_events: (KafkaTopic::ReplayEvents, "ingest-replay-events"),
+        replay_recordings: (KafkaTopic::ReplayRecordings, "ingest-replay-recordings"),
+        monitors: (KafkaTopic::Monitors, "ingest-monitors"),
+        spans: (KafkaTopic::Spans, "snuba-spans"),
+        metrics_summaries: (KafkaTopic::MetricsSummaries, "snuba-metrics-summaries"),
+        cogs: (KafkaTopic::Cogs, "shared-resources-usage"),
     }
 }
```
jjbayer: Nice!

relay-kafka/src/config.rs
```diff
@@ -80,93 +80,62 @@ impl KafkaTopic {
...
-}
 
-impl Default for TopicAssignments {
-    fn default() -> Self {
-        Self {
-            events: "ingest-events".to_owned().into(),
-            attachments: "ingest-attachments".to_owned().into(),
-            transactions: "ingest-transactions".to_owned().into(),
-            outcomes: "outcomes".to_owned().into(),
-            outcomes_billing: None,
```
jjbayer, line 166: How are we handling this special case? I assume this is the reason why integration tests are failing, see

https://github.com/getsentry/relay/blob/94cb3df759108f483b7cf3c2f41e0c6262eefb06/relay-kafka/src/config.rs#L138

lynnagara: I'm not handling it at all. This is no longer treated as a special case in any of our other services (see https://github.com/getsentry/sentry/blob/d414afaab6fcd22e8609d0a885e25cad0752b83b/src/sentry/conf/server.py#L3477 and https://github.com/getsentry/sentry-kafka-schemas/blob/main/topics/outcomes-billing.yaml for example). If someone reaalllly wanted to share the topic technically still they could accomplish it via overriding `outcomes-billing` back to `outcomes` in relay config (but that's not really something we'd do anymore). It only ever existed for dev and dev was updated to just match prod on this recently.

jjbayer:
> a bit stumped by the failing integration test, any ideas how to fix it?

I picked a few of the failing tests randomly and they all passed on my machine, must have something to with the topic configuration. I'll take another look.

relay-kafka/src/config.rs
```diff
@@ -80,93 +80,62 @@ impl KafkaTopic {
...
+                    )*
+
+                }
+            }
         }
+    };
+}
+
+define_topic_assignments! {
+    TopicAssignments {
```
Dav1dde, line 127: Can we get the documentation back? It should be possible to include it in the macro and just generate it on the fields using `#[doc = $foo]`.

relay-kafka/src/config.rs
Dav1dde: Does it break self hosted?

relay-kafka/src/config.rs
```diff
@@ -447,4 +416,28 @@ transactions: "ingest-transactions-kafka-topic"
         assert_eq!(shards, 65000);
         assert_eq!(3, mapping.len());
     }
+
+    #[test]
+    fn test_default_topic_is_valid() {
+        let topic_assignments = TopicAssignments::default();
+
+        // A few topics are not defined currently, remove this once added
```
Dav1dde:
```suggestion
        // A few topics are not defined currently, remove this once added to `sentry-kafka-schemas`.
```

Dav1dde (approved):
Love the changes!

Dav1dde:
> a bit stumped by the failing integration test, any ideas how to fix it?

Possibly to do with how it auto generates topic names, maybe the mapping is broken?

jjbayer:
> a bit stumped by the failing integration test, any ideas how to fix it?

I pushed an update to the test topics config, let's see if that makes tests pass 🤞

Dav1dde:
> I pushed an update to the test topics config, let's see if that makes tests pass 🤞

I am a bit worried, this is now officially a breaking change. If we have it configured that way, somebody else will probably too (self hosted?).

relay-kafka/src/config.rs
lynnagara: It doesn't. Self hosted runs a billing consumer too. https://github.com/getsentry/self-hosted/blob/c4ae491929170c075bea608941b227a8db81eec2/docker-compose.yml#L268-L270.

Getsentry already produces to that topic and it gets written to clickhouse similarly to before.

relay-kafka/src/config.rs
```diff
@@ -80,93 +80,62 @@ impl KafkaTopic {
...
+
+define_topic_assignments! {
+    TopicAssignments {
+        events: (KafkaTopic::Events, "ingest-events"),
+        attachments: (KafkaTopic::Attachments, "ingest-attachments"),
+        transactions: (KafkaTopic::Transactions, "ingest-transactions"),
+        outcomes: (KafkaTopic::Outcomes, "outcomes"),
+        outcomes_billing: (KafkaTopic::OutcomesBilling, "outcomes-billing"),
+        metrics_sessions: (KafkaTopic::MetricsSessions, "ingest-metrics"),
+        metrics_generic: (KafkaTopic::MetricsGeneric, "ingest-performance-metrics"),
```
jjbayer: I think we need to bring back the `metrics` and `metrics_transactions` aliases to ensure backward compatibility. Maybe we can tweak the macro to allow multiple aliases.

jjbayer (changes requested):
> I am a bit worried, this is now officially a breaking change. If we have it configured that way, somebody else will probably too (self hosted?).

Agreed, we should restore the existing topic aliases (see comment).

relay-kafka/src/config.rs
lynnagara: What do we need the backward compatibility for? Do we think that self-hosted users would have used these aliases? And is it the intention to support that forever? (it is not part of the default self-hosted configuration)

jjbayer: I guess to not cause a breaking change if it is unnecessary? We can also call it out as a breaking change in the changelog, the number of users who configure topics manually is probably close to zero.

relay-kafka/src/config.rs
lynnagara: Sounds good, added back the original descriptions

relay-kafka/src/config.rs
lynnagara: I updated the changelog in this PR. Did you want one calling it out in self-hosted too?

My personal feelings on this are that making this more complicated in order to maintain compatibility with something that probably no one has used is not really worth it. I'm not sure how this is generally handled in Relay, but in sentry and snuba codebases at least, breaking compatibility is generally fine as long as we call it out (and we did already break compatibility when this same change was shipped there)

jjbayer (approved):
I like the changelog entry, it's very clear. If we are sure that this does not break default self-hosted or any other deployments maintained by us, ship it!

PR #3365, opened 2024-04-02
feat(perf): Allow loopback IP addresses in metrics `domain` tag

gggritso:
`http.client` spans with descriptions like `GET http://127.0.0.1:10000` do not get an extracted `"domain"` tag at all because the `.domain()` method of the `url::Url` struct _only_ returns a value for string domains like `myhost.com`, and returns `None` for IP addresses!

Switching to `host_str` extracts the tag for IP addresses too, which we want! In our internal data a missing domain is _by far_ the most prevalent domain tag, and calls to Snuba and Flagr are completely missing.

Note: This does _not_ affect non-loopback IPs like `8.8.8.8` because they are not allowlists, and get scrubbed out of span descriptions, so they don't get extracted into a `domain` tag at all.


jjbayer (approved):
LGTM. @gggritso please keep an eye on the cardinality of this tag before and after this is deployed. I.e. this query:

```sql
SELECT org_id, project_id,
uniq(tags.raw_value [indexOf(tags.key, 9223372036854775808 + 248)]) AS `span.domain`
FROM generic_metric_distributions_aggregated_dist
WHERE (granularity = 2)
AND (timestamp >= now() - INTERVAL 1 HOUR)
AND (metric_id IN [9223372036854776214])
GROUP BY org_id, project_id
ORDER BY span.domain DESC
LIMIT 100
```

iker-barriocanal (approved):
Do we have any estimations on expected cardinality? Should we run this behind a feature flag?

gggritso:
@jjbayer @iker-barriocanal fun plot twist, but I took a second look at the code and it turned out that this change doesn't affect IPs aside from `127.0.0.1`. I will tackle those separately. Because of that, I don't expect a big cardinality change overall. A lot of blank `domain` tags will become a string, but its values should match the `description` tag anyway

PR #3366, opened 2024-04-02
feat(spans): Macro for transaction <-> span conversion

jjbayer:
We currently convert transactions to segment spans, and in the near future will convert segment spans to transactions as well (see https://github.com/getsentry/relay/issues/3278).

The goal of this PR is to have a declarative implementation of the conversion in both directions, such that it is easy to extend and to make sure that the two implementations (`From<&Event> for Span` and `From<&Span> for Event`) do not diverge over time.

Future work (not in this PR): Actually conversion of incoming segment spans to transactions.

#skip-changelog

### DIFF
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -95,7 +95,7 @@ pub struct Span {
 
     /// Platform identifier.
     ///
-    /// See [`Event::platform`].
+    /// See [`Event::platform`](`crate::protocol::Event::platform`).
     #[metastructure(skip_serialization = "empty")]
     pub platform: Annotated<String>,
 
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -1,3 +1,5 @@
+mod conversion;
+
 #[cfg(feature = "jsonschema")]
 use relay_jsonschema_derive::JsonSchema;
 use relay_protocol::{Annotated, Empty, FromValue, Getter, IntoValue, Object, Val, Value};
@@ -107,38 +109,38 @@ pub struct Span {
     pub other: Object<Value>,
 }
 
-impl From<&Event> for Span {
-    fn from(event: &Event) -> Self {
-        let mut span = Self {
-            _metrics_summary: event._metrics_summary.clone(),
-            description: event.transaction.clone(),
-            is_segment: Some(true).into(),
-            received: event.received.clone(),
-            start_timestamp: event.start_timestamp.clone(),
-            timestamp: event.timestamp.clone(),
-            measurements: event.measurements.clone(),
-            platform: event.platform.clone(),
-            was_transaction: true.into(),
-            ..Default::default()
-        };
+// impl From<&Event> for Span {
+//     fn from(event: &Event) -> Self {
+//         let mut span = Self {
+//             _metrics_summary: event._metrics_summary.clone(),
+//             description: event.transaction.clone(),
+//             is_segment: Some(true).into(),
+//             received: event.received.clone(),
+//             start_timestamp: event.start_timestamp.clone(),
+//             timestamp: event.timestamp.clone(),
+//             measurements: event.measurements.clone(),
+//             platform: event.platform.clone(),
+//             was_transaction: true.into(),
+//             ..Default::default()
+//         };
 
-        if let Some(trace_context) = event.context::<TraceContext>().cloned() {
-            span.exclusive_time = trace_context.exclusive_time;
-            span.op = trace_context.op;
-            span.parent_span_id = trace_context.parent_span_id;
-            span.segment_id = trace_context.span_id.clone(); // a transaction is a segment
-            span.span_id = trace_context.span_id;
-            span.status = trace_context.status;
-            span.trace_id = trace_context.trace_id;
-        }
+//         if let Some(trace_context) = event.context::<TraceContext>().cloned() {
+//             span.exclusive_time = trace_context.exclusive_time;
+//             span.op = trace_context.op;
+//             span.parent_span_id = trace_context.parent_span_id;
+//             span.segment_id = trace_context.span_id; // a transaction is a segment
+//             span.span_id = trace_context.span_id;
+//             span.status = trace_context.status;
+//             span.trace_id = trace_context.trace_id;
+//         }
 
-        if let Some(profile_context) = event.context::<ProfileContext>() {
-            span.profile_id = profile_context.profile_id.clone();
-        }
+//         if let Some(profile_context) = event.context::<ProfileContext>() {
+//             span.profile_id = profile_context.profile_id.clone();
+//         }
 
-        span
-    }
-}
+//         span
+//     }
+// }
 
 impl Getter for Span {
     fn get_value(&self, path: &str) -> Option<Val<'_>> {
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -109,39 +109,6 @@ pub struct Span {
     pub other: Object<Value>,
 }
 
-// impl From<&Event> for Span {
-//     fn from(event: &Event) -> Self {
-//         let mut span = Self {
-//             _metrics_summary: event._metrics_summary.clone(),
-//             description: event.transaction.clone(),
-//             is_segment: Some(true).into(),
-//             received: event.received.clone(),
-//             start_timestamp: event.start_timestamp.clone(),
-//             timestamp: event.timestamp.clone(),
-//             measurements: event.measurements.clone(),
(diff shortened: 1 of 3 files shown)


relay-event-schema/src/protocol/span.rs
```diff
@@ -486,87 +454,6 @@ mod tests {
         assert!(!RuleCondition::eq("span.was_transaction", false).matches(&span));
     }
 
-    #[test]
-    fn span_from_event() {
```
jjbayer, line 490: I moved this test to the new module, and expanded it with more fields.

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -0,0 +1,231 @@
...
+                                )*
+                            ]),
+                        )
+                    ),
+                    ..Default::default()
+                }
+            }
+        }
+    };
+}
```
jjbayer, line 109: I'm considering moving the macro definition itself to a different module, just so users of the macro don't have to scroll down every time the want to update the macro invocation.

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -0,0 +1,254 @@
...
+///             span.trace_id: context.trace_id
+///         ;
+///     ;
+///     fixed_for_span:
+///         // ...
+///     ;
+///     fixed_for_event:
+///         // ...
+///     ;
+/// );
```
jjbayer, line 31: I would have preferred a flat mapping like this:

```rust
map_fields!(
  span.received <=> event.received,
  span.trace_id <=> event.contexts.trace.trace_id,
  span.is_segment <= Some(true)
);
```

but I could not make it work with a simple macro, and a procedural macro seemed overkill for this purpose.

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -0,0 +1,254 @@
...
+use relay_protocol::Annotated;
+use std::collections::BTreeMap;
+
+/// Implements the conversion between transaction events and segment spans.
+///
+/// Invoking this macro implements both `From<&Event> for Span` and `From<&Span> for Event`.
+///
+/// Example:
+///
+/// ```skip
```
Dav1dde:
```suggestion
/// ```ignore
```

See: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes - then we should get better syntax highlighting

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -0,0 +1,254 @@
...
+/// map_fields!(
+///     top-level:
+///         span.received: event.received
+///     contexts:
+///         TraceContext:
+///             span.trace_id: context.trace_id
+///         ;
+///     ;
```
Dav1dde: I think I prefer 

```rust
contexts: [ 
    TraceContext {
    }
]
```

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -0,0 +1,254 @@
...
+///     fixed_for_span:
+///         // ...
+///     ;
```
Dav1dde: And for these:

```
fixed_for_span: {
    // ...
}

File: relay-event-schema/src/protocol/span/convert.rs (hunk 3 above)
Comment (Dav1dde) [reply relay-event-schema/src/protocol/span/convert.rs:31]: I think we can make that work if it is just about `<=>`, `=>`, `<=`, do you want me take a look?

File: relay-event-schema/src/protocol/span/convert.rs [hunk 7]
```diff
@@ -0,0 +1,254 @@
...
+        span.start_timestamp <=> event.start_timestamp,
+        span.timestamp <=> event.timestamp
+    ;
+    contexts:
+        TraceContext:
+            span.exclusive_time <=> context.exclusive_time,
+            span.op <=> context.op,
+            span.parent_span_id <=> context.parent_span_id,
+            // A transaction corresponds to a segment span, so span_id and segment_id have the same value:
+            span.span_id, span.segment_id <=> context.span_id,
```
Dav1dde, line 129: that is interesting syntax, should we just split it into two different lines?

Dav1dde (approved):
Just bikeshedding on syntax we can also revisit later.

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -0,0 +1,254 @@
...
+///
+/// ```ignore
+/// map_fields!(
+///     top-level:
+///         span.received: event.received
+///     contexts:
+///         TraceContext:
+///             span.trace_id: context.trace_id
+///         ;
+///     ;
```
iker-barriocanal: Which fields do this semicolon map to? Does `top-level` also need them?

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -0,0 +1,254 @@
...
+/// ```
+macro_rules! map_fields {
+    (
+        top-level:
+            $(span.$span_field:ident <=> event.$event_field:ident), *
+        ;
+        contexts:
+        $(
+            $ContextType:ident:
+                $(span.$foo:ident $(, $(span.$span_field_from_context:ident),+)? <=> context.$context_field:ident), *
```
iker-barriocanal:
```suggestion
                $(span.$context_field:ident $(, $(span.$span_field_from_context:ident),+)? <=> context.$context_field:ident), *
```

relay-event-schema/src/protocol/span/convert.rs
iker-barriocanal: What's the value in `event.context.span_id` if the `span.span_id` and `span.segment_id` are different?

relay-event-schema/src/protocol/span/convert.rs
jjbayer: There's two problems:

1. You need to cast a context to a specific type to access its fields.
2. The compiler will complain about parsing ambiguity on the LHS identifier when there's no separators between the `<=>` and `<=`  blocks.

But I'm very open to ideas, there's probably possibilities I haven't discovered yet!

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -0,0 +1,254 @@
...
+            // A transaction corresponds to a segment span, so span_id and segment_id have the same value:
+            span.span_id, span.segment_id <=> context.span_id,
+            span.status <=> context.status,
+            span.trace_id <=> context.trace_id
+        ;
+        ProfileContext:
+            span.profile_id <=> context.profile_id
+        ;
+    ;
+    fixed_for_span:
```
iker-barriocanal: Do we need these sections? Would it be easier to read if we have a symbol as a span/event variable? Example:
```txt
@span._metrics_summary <=> @event._metrics_summary,  // picks the relative span and event fields
@span.is_segment <= Some(true), // picks the relative span field, and assigns the fixed value
```

relay-event-schema/src/protocol/span/convert.rs
Dav1dde: Yeah contexts would need to stay separate (or with special syntax), but maybe it's possible to combine `top-level` `fixed-for-span` and `fixed-for-event`.

iker-barriocanal (approved):

relay-event-schema/src/protocol/span/convert.rs
jjbayer: Thanks!

relay-event-schema/src/protocol/span/convert.rs
jjbayer: The first one. I improved the variable names to clarify:

https://github.com/getsentry/relay/blob/1809a0f7e8cb5fc15febb6268879a995bf6cb3d2/relay-event-schema/src/protocol/span/convert.rs#L94-L97

relay-event-schema/src/protocol/span/convert.rs
jjbayer: I like the `{}` braces, so I introduced them everywhere now. Wasn't sure about `:`, `[` and `,`, so I fell back to a minimal syntax with braces.

relay-event-schema/src/protocol/span/convert.rs
jjbayer: Replaced `:` and `;` with `{` and `}` for readability.

relay-event-schema/src/protocol/span/convert.rs
jjbayer: The parser would already fail on the left-hand side. What would work is a different symbol, e.g.

```
@span._metrics_summary <=> event._metrics_summary,  // picks the relative span and event fields
span.is_segment <= Some(true), // picks the relative span field, and assigns the fixed value
```

PR #3370, opened 2024-04-03
feat(spans): Tag duration by group for all ops supporting description scrubbing

phacops:
In order to make comparison between spans of the same group, we need to start tagging the duration metric by group as well. That way, we'll be able to indicate, for a given op + description, if the span was faster or slower.

image

This will increase cardinality for this metric, putting it on part with what we have for exclusive time.


relay-dynamic-config/src/defaults.rs
```diff
@@ -416,13 +416,13 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
                 Tag::with_key("transaction.op")
                     .from_field("span.sentry_tags.transaction.op")
                     .always(),
+                Tag::with_key("span.group")
+                    .from_field("span.sentry_tags.group")
+                    .always(),
```
Zylphrex: Won't this extract the span group for every span? The exclusive time metric only extracts it for `know_modules_condition`

jjbayer: +1

phacops: I was thinking the group is calculated only for known modules, otherwise left `None`, and we wouldn't tag a `None` value.

In the end, we may calculate the group for more than what we support and then have another layer of conditions so, using the `known_modules_condition` is safer indeed.

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics.snap
```diff
@@ -3349,7 +3370,6 @@ expression: metrics
             "environment": "fake_environment",
             "release": "1.2.3",
             "span.description": "Cold Start",
-            "span.group": "2d675185edfeb30c",
```
Zylphrex: This looks like you need to also extract the group for app starts in addition to the known groups. Or make app start spans a known group.

jjbayer (approved):
Let's keep an eye on cardinality limits and bucket rate limits when we deploy this.

PR #3371, opened 2024-04-03
feat(spans): extract cache.hit and cache.item_size metrics

DominikB2014:
Work for #https://github.com/getsentry/sentry/issues/68118

1. Creates a new metric for `cache.item_size`
2. Adds `cache.hit` tag to various applicable metrics 
3. Ensure cache spans collect exclusive time metrics (this might not be needed in the final cache module v0, so we can look into removing after that's been determined).

#skip-changelog 

^ there seems to be a bug, where it says i didn't modify, but i did?


relay-dynamic-config/src/defaults.rs
```diff
@@ -122,12 +131,16 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
...
         & duration_condition.clone();
 
-    let know_modules_condition =
-        (is_db.clone() | is_resource.clone() | is_mobile.clone() | is_http.clone())
-            & duration_condition.clone();
+    let know_modules_condition = (is_db.clone()
+        | is_resource.clone()
+        | is_mobile.clone()
+        | is_http.clone()
+        | is_cache.clone())
```
jjbayer: By adding `is_cache` to `known_modules_condition`, we get multiple tags that we might not need (see `.when(know_modules_condition.clone()),` below). Could we start by only adding the `cache.hit` tag instead?

relay-dynamic-config/src/defaults.rs
```diff
@@ -263,6 +280,30 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
...
                     .when(is_http.clone()),
+                // Cache module
+                Tag::with_key("cache.hit")
+                    .from_field("span.sentry_tags.cache_hit")
+                    .when(is_cache.clone()),
+            ],
+        },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "d:spans/cache.item_size@byte".into(),
```
jjbayer, line 287: @Dav1dde @phacops what's the status on [strings.py](https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py)? Do we still want to add metric names there?

relay-dynamic-config/src/defaults.rs
DominikB2014: Yeah we can, works for me!

relay-dynamic-config/src/defaults.rs
Dav1dde: I think that has been "deprecated" and it isn't supposed to be changed anymore, there is a Python test preventing modifications using a hashsum of the file.

Although I think it's a bit unfortunate because it makes querying for specific metrics in clickhouse directly impossible :(

relay-dynamic-config/src/defaults.rs
DominikB2014: done!

relay-dynamic-config/src/defaults.rs
```diff
@@ -30,6 +30,13 @@ const MONGODB_QUERIES: &[&str] = &["*\"$*", "{*", "*({*", "*[{*"];
 /// A list of patterns for resource span ops we'd like to ingest.
 const RESOURCE_SPAN_OPS: &[&str] = &["resource.script", "resource.css", "resource.img"];
 
+const CACHE_SPAN_OPS: &[&str] = &[
+    "cache.get_item",
+    "cache.save",
+    "cache.clear",
+    "cache.delete_item",
+];
```
iker-barriocanal, line 38: Does `cache.*` cover all operations without false positives?

relay-dynamic-config/src/defaults.rs
```diff
@@ -263,6 +276,30 @@ fn span_metrics() -> impl IntoIterator<Item = MetricSpec> {
...
+            tags: vec![
+                Tag::with_key("environment")
+                    .from_field("span.sentry_tags.environment")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("span.op")
+                    .from_field("span.sentry_tags.op")
+                    .always(), // already guarded by condition on metric
+                Tag::with_key("transaction")
+                    .from_field("span.sentry_tags.transaction")
+                    .always(), // already guarded by condition on metric
```
iker-barriocanal, line 299: IIUC the cache operation goes into either of these, like a redis `GET` command. Are we confident that the current normalization for this metric does not explode in cardinality? Can storage handle the new load?

relay-dynamic-config/src/defaults.rs
DominikB2014: Yes it should!

relay-dynamic-config/src/defaults.rs
jjbayer: Can't be worse than on exclusive time, right?

relay-dynamic-config/src/defaults.rs
jjbayer: IMO we should still allow MRIs to be added to that file. The chance of those appearing in existing customer data is extremely low, and it makes running cardinality queries much easier, like you said.

jjbayer (approved):

relay-dynamic-config/src/defaults.rs
iker-barriocanal: Should we simplify the patterns then?

iker-barriocanal (approved):

relay-dynamic-config/src/defaults.rs
DominikB2014: I don't mind being more explicit here since there is only 4 operations, it feels a bit more robust versus a wildcard on spans ops starting with cache.

PR #3372, opened 2024-04-04
instr(cardinality): Include amount of scopes in statsd metric

Dav1dde:
Just another tag for the metric.

#skip-changelog


relay-cardinality/src/redis/limiter.rs
```diff
@@ -233,6 +236,24 @@ impl IntoIterator for CheckedLimits {
...
+/// Buckets the amount of scopes contained in a scope into a metric tag.
+fn num_scopes_tag(state: &LimitState<'_>) -> &'static str {
+    match state.scopes().len() {
+        0 => "0",
+        1 => "1",
+        2 => "2",
+        3 => "3",
+        4 => "4",
+        5 => "5",
+        ..=10 => "10",
```
iker-barriocanal: Should we add a prefix to the numbers?
```suggestion
        ..=10 => "<=10",
```

iker-barriocanal (approved):

relay-cardinality/src/redis/limiter.rs
Dav1dde: Not a fan, I think it generates more noise in the metric (to read and query for) and `<= 10` is still technically incorrect it would need to be `5 < x <= 10` which becomes really long and unreadable.

PR #3375, opened 2024-04-04
feat(spans): Extract transaction from segment span

jjbayer:

The current state of Performance is that

1. There are SDKs that emit transactions, and there are SDKs that emit standalone spans.
2. There are parts of the product that require transactions, and there are parts of the product that require spans.

We already extract spans from transactions to make span-dependent product features work for "transaction"-SDKs. Now we need to also extract transactions from spans to make transaction-dependent product features work for "span"-SDKs:

```mermaid
flowchart LR
    
    SDK -->|transaction| Relay
    SDK -->|span| Relay
    Relay -->|span| spc[Span Consumer]
    Relay -->|transaction| txc[Transaction Consumer] 

    Relay -->|span from transaction| spc
    Relay -->|transaction from span| txc
    
    linkStyle 5 color:green;
```

## Point of conversion

When to actually convert the segment span to transaction? I've listed two options below. At the moment, I believe only Option 1 is feasible because the span processing pipeline does not have all features implemented yet.

### Option 1 - At the start of envelope processing [Selected (see Option 1a)]

```mermaid
flowchart TD
SDK -->|span| extract
extract-->|span| process_standalone_span
extract-->|transaction| process_transaction
```

Pros:

- No need to duplicate code for e.g. transaction metrics extraction. An extracted transaction can pass through the same processing pipeline as an "organic" transaction sent from the SDK, with the same normalization, PII scrubbing, etc.
- The transaction processing pipeline is more mature and better tested than the span processing pipeline.

Cons:

- Some duplicate work: Normalization and PII scrubbing will run for both the original segment span and the transaction extracted from it.
- Inconsistent with how we extract spans from transactions (this is done after processing).
- Extraction will occur in edge Relays (not just processing Relays), so any updates to the conversion would take months to propagate to external Relays.

### Option 1a - At the start of span processing in processing Relays [Selected]

Like Option 1, but only done in processing relays:

Pros:

- Gets rid of one of the Cons of Option 1.
- Spans are currently only parsed in processing Relays. No need to refactor that.

Cons:

- Need to make sure that transactions are normalized, even if normalization is disabled in processing Relays.

### Option 2 - At the end of envelope processing (in processing Relays) [Discarded]

```mermaid
flowchart TD

process_span["process span (normalize, filter, metrics, sample)"]

SDK -->|span| process_span
process_span -->|span| extract_transaction
extract_transaction -->|span| enforce_quotas
extract_transaction -->|transaction| extract_metrics_tx
extract_metrics_tx -->|transaction| enforce_quotas
```

Pros:

- Assuming that span processing already filters, normalizes, samples and scrubs spans correctly, there would be no duplicate work done for the extracted transaction. All that's left would be transaction metrics extraction and rate limiting
- Consistent with how we extract spans from transactions.

Cons:

- Cannot leverage the fully mature transaction processing pipeline.
  - **BLOCKING:** Inbound filters and dynamic sampling for spans are not ready yet.  
- Needs some duplicate code to extract transaction metrics from extracted transactions.

## Prevent duplicate data

We already cross the spans/transactions in two places:

1. For every transaction, Relay extracts one standalone span for the transaction's child spans, _and_ a standalone segment span for the transaction itself.
2. For compatibility of performance scores, there is one transaction metric that is also extracted from standalone spans: `"d:transactions/measurements.score.total@ratio"`.

To prevent circular conversion of data, I suggest to introduce two new item headers:

1. `"transaction_extracted"` for span items, which will be checked before converting a span to a transaction. For segment spans extracted from transactions, this flag will be `true` from the start.
2. `"spans_extracted"` for transaction items, which will be checked before extracting spans or span metrics from a transaction. For transactions extracted from spans, this flag will be `true` from the start.

In addition, we will stop extracting `"d:transactions/measurements.score.total@ratio"` from spans.

## TODO

- [x] Ensure transaction gets normalized even if normalization is disabled
- [ ] Test with a local dev setup to make sure transactions appear in the product without breaking consumers.
- [x] Modify test input to make `is_segment` false for some.
- [x] Add a test with `score.total` to ensure that it is never extracted more than once.


relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -78,31 +79,58 @@ macro_rules! map_fields {
...
+                                        context.$context_field = span.$primary_span_field.clone();
+                                        has_fields = true;
+                                    }
                                 )*
-                            ]),
-                        )
+                                if has_fields {
+                                    let context_key = <$ContextType as DefaultContext>::default_key().into();
+                                    contexts.insert(context_key, ContextInner(context.into_context()).into());
+                                }
```
jjbayer, line 114: This prevents an empty `ProfileContext` from appearing in the event.

relay-event-normalization/src/normalize/span/exclusive_time.rs
```diff
@@ -42,6 +42,11 @@ fn set_event_exclusive_time(
         return;
     };
 
+    if trace_context.exclusive_time.value().is_some() {
+        // Exclusive time already set, respect.
+        return;
+    }
```
jjbayer: This is a necessary change in behavior: A transaction derived from a standalone span does not have child spans, so we need to keep the `exclusive_time` set on the transaction, otherwise `exclusive_time` will always equal the full duration.

relay-sampling/src/evaluation.rs
```diff
@@ -65,6 +65,11 @@ impl<'a> ReservoirEvaluator<'a> {
         }
     }
 
+    /// Returns a shared reference to the reservoir counters.
+    pub fn counters(&self) -> ReservoirCounters {
+        self.counters.clone()
+    }
```
jjbayer, line 71: Needed to split off a `ProcessEnvelope`.

relay-server/src/envelope.rs
```diff
@@ -554,6 +554,14 @@ pub struct ItemHeaders {
     #[serde(default, skip_serializing_if = "is_false")]
     metrics_extracted: bool,
 
+    /// Whether or not a transaction has been extracted from a segment span.
+    #[serde(default, skip_serializing_if = "is_false")]
+    transaction_extracted: bool,
+
+    /// Whether or not spans have been extracted from a transaction.
+    #[serde(default, skip_serializing_if = "is_false")]
+    spans_extracted: bool,
```
jjbayer, line 569: With these two flags, we can prevent conversion between events and spans from going in circles.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -46,26 +46,48 @@ impl Extractable for Span {
...
     event: &Event,
+    spans_extracted: bool,
     config: &MetricExtractionConfig,
     max_tag_value_size: usize,
 ) -> Vec<Bucket> {
     let mut metrics = generic::extract_metrics(event, config);
 
+    // If spans were already extracted for an event,
+    // we rely on span processing to extract metrics.
+    if !spans_extracted {
```
jjbayer: Note to self: Double-check if semantics make sense.

relay-server/src/services/processor/event.rs
```diff
@@ -383,6 +384,9 @@ pub fn serialize<G: EventProcessing>(
     // If transaction metrics were extracted, set the corresponding item header
     event_item.set_metrics_extracted(state.event_metrics_extracted);
 
+    // TODO: The state should simply maintain & update an `ItemHeaders` object.
```
jjbayer, line 387: I will follow up on this in a different PR.

jjbayer:
Still requires some more test coverage, but opening for review to get feedback.

relay-server/src/envelope.rs
```diff
@@ -1060,6 +1090,21 @@ impl Envelope {
         Box::new(Self { items, headers })
     }
 
+    /// Creates an envelope from an event, plus headers.
```
```suggestion
    /// Creates an envelope from headers and an envelope.
```

relay-server/src/services/processor/span/processing.rs
```diff
@@ -91,10 +98,18 @@ pub fn process(
             return ItemAction::Drop(Outcome::Invalid(DiscardReason::Internal));
         };
 
+        if should_extract_transactions && !item.transaction_extracted() {
+            if let Some(transaction) = convert_to_transaction(&annotated_span) {
```
iker-barriocanal: Span and transaction normalization may have different requirements, so I'd extract transactions before normalizing spans (line 91 above).

relay-server/src/services/processor/span/processing.rs
```diff
@@ -141,6 +156,37 @@ pub fn process(
...
+                        addrs.test_store.clone(),
+                        ProcessingGroup::Transaction,
+                    ),
+                    project_state: state.project_state.clone(),
+                    sampling_project_state: state.sampling_project_state.clone(),
+                    reservoir_counters: state.reservoir.counters().clone(),
+                });
+            }
+            Err(e) => {
+                relay_log::error!("Failed to create event envelope: {e}");
```
iker-barriocanal: Is the project ID/counter useful to add to the message?

relay-server/src/services/processor/span/processing.rs
```diff
@@ -141,6 +156,37 @@ pub fn process(
...
+                addrs.envelope_processor.send(ProcessEnvelope {
+                    envelope: ManagedEnvelope::standalone(
+                        envelope,
+                        addrs.outcome_aggregator.clone(),
+                        addrs.test_store.clone(),
+                        ProcessingGroup::Transaction,
+                    ),
+                    project_state: state.project_state.clone(),
+                    sampling_project_state: state.sampling_project_state.clone(),
+                    reservoir_counters: state.reservoir.counters().clone(),
```
iker-barriocanal: Are we cloning reservoir counters twice? `counters()` clones once and there's an explicit clone here. `counters` should return a reference and we should clone here instead.

relay-sampling/src/evaluation.rs
```diff
@@ -65,6 +65,11 @@ impl<'a> ReservoirEvaluator<'a> {
         }
     }
 
+    /// Returns a shared reference to the reservoir counters.
+    pub fn counters(&self) -> ReservoirCounters {
+        self.counters.clone()
```
iker-barriocanal: The function clones but the docs say it returns a reference. Related to my comment below: this function should return the reference without cloning.

tests/integration/test_spans.py
```diff
@@ -454,7 +475,9 @@ def test_span_ingestion(
         headers={"Content-Type": "application/x-protobuf"},
     )
 
+    print("Waiting for spans..")
     spans = list(spans_consumer.get_spans(timeout=10.0, max_attempts=6))
+    print("Done waiting for spans.")
```
iker-barriocanal:
```suggestion
    spans = list(spans_consumer.get_spans(timeout=10.0, max_attempts=6))
```
Please also delete the `print`s below (these should be cough by [automation](https://github.com/getsentry/relay/issues/3398) 😞)

tests/integration/test_spans.py
```diff
@@ -384,16 +394,26 @@ def test_span_ingestion(
...
 
     # 1 - Send OTel span and sentry span via envelope
     envelope = envelope_with_spans(start, end)
-    relay.send_envelope(project_id, envelope)
+    relay.send_envelope(
+        project_id,
+        envelope,
+        headers={
+            "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/111.0.0.0 Safari/537.36"
+        },
```
iker-barriocanal: Are these headers only for `browser.name`? If yes, I don't see any benefits of having it and it can distract from what the test tries to validate.

tests/integration/test_spans.py
```diff
@@ -556,20 +580,46 @@ def test_span_ingestion(
             "description": "my 3rd protobuf OTel span",
             "duration_ms": 500,
             "exclusive_time_ms": 500.0,
-            "is_segment": True,
+            "is_segment": False,
```
iker-barriocanal, line 585: Why is this span no longer a segment?

iker-barriocanal (commented):
Overall lgtm. Leaving a few comments and questions:
- How are we planning to measure COGS for transactions coming from segments?
- A single message with an envelope with N spans will result in N new messages in the processor, and we'd drop messages if the processor's queue is full. Solving that problem is out of scope of this PR, but what do you think about adding observability for that? This queue can grow quite fast without accepting new requests.

relay-server/src/services/processor/span/processing.rs
jjbayer: The project ID should already be set further up the call stack: https://github.com/getsentry/relay/blob/3f124cb016071ee5c86418704624b85ca3831518/relay-server/src/services/processor.rs#L1686-L1697

We can improve instrumentation if this error ever triggers.

relay-server/src/services/processor/span/processing.rs
jjbayer: Thanks!

relay-server/src/services/processor/span/processing.rs
jjbayer: Good point, refactored now. `convert_to_transaction` relies on `is_segment` normalization, so I moved that part out of normalization to still run before.

tests/integration/test_spans.py
jjbayer: It's needed to check deduplication of `d:transactions/measurements.score.total@ratio`. I added a comment to clarify.

tests/integration/test_spans.py
jjbayer: I deliberately gave it a `parent_span_id` so I could verify that transactions are _not_ extracted from regular spans, only from segment spans.

jjbayer (commented):
@iker-barriocanal 

> How are we planning to measure COGS for transactions coming from segments?

The extraction part will be accounted as `AppFeature::Spans`, the processing as `AppFeature::Transactions`. I think this is fine for now, as both are part of Performance cost.

> A single message with an envelope with N spans will result in N new messages in the processor, and we'd drop messages if the processor's queue is full. Solving that problem is out of scope of this PR, but what do you think about adding observability for that? This queue can grow quite fast without accepting new requests.

Good point, I added two metrics now so we can observe the number of spin-off transactions per envelope.

relay-event-normalization/src/normalize/span/exclusive_time.rs
```diff
@@ -42,6 +42,11 @@ fn set_event_exclusive_time(
         return;
     };
 
+    if trace_context.exclusive_time.value().is_some() {
+        // Exclusive time already set, respect.
```
Dav1dde: I respect that, bro 🤣 

```suggestion
        // Exclusive time already set, respect it.
```

relay-sampling/src/evaluation.rs
```diff
@@ -65,6 +65,11 @@ impl<'a> ReservoirEvaluator<'a> {
         }
     }
 
+    /// Gets shared ownership of the reference counters.
```
Dav1dde:
```suggestion
    /// Gets shared ownership of the reservoir counters.
```

Is there any invariants to uphold if someone gets access to the counters (if so please add them to the docs)?

E.g. counters need to only be update when X and then you also need to keep Y in sync.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -47,31 +47,41 @@ impl Extractable for Span {
...
     max_tag_value_size: usize,
     span_extraction_sample_rate: Option<f32>,
 ) -> Vec<Bucket> {
     let mut metrics = generic::extract_metrics(event, config);
 
-    if !sample(span_extraction_sample_rate.unwrap_or(1.0)) {
-        return metrics;
+    // If spans were already extracted for an event,
+    // we rely on span processing to extract metrics.
+    if !spans_extracted && sample(span_extraction_sample_rate.unwrap_or(1.0)) {
```
Dav1dde, line 59: Unrelated thought: Should we change the signature of sample to: `fn sample(rate: impl Into<Option<f32>>) -> bool` instead?

relay-server/src/services/processor/span/processing.rs
```diff
@@ -87,6 +95,15 @@ pub fn process(
             _ => return ItemAction::Keep,
         };
 
+        set_segment_attributes(&mut annotated_span);
+
+        if should_extract_transactions && !item.transaction_extracted() {
+            if let Some(transaction) = convert_to_transaction(&annotated_span) {
+                extracted_transactions.push(transaction);
+                item.set_transaction_extracted(true);
```
Dav1dde, line 104: I stumbled over this, see comments on `transaction_extracted`.

relay-server/src/statsd.rs
```diff
@@ -382,6 +382,9 @@ pub enum RelayTimers {
     /// This metric is tagged with:
     ///  - `type`: The type of the health check, `liveness` or `readiness`.
     HealthCheckDuration,
+
+    /// Measurees how many transactions were created from segment spans in a single envelope.
```
iker-barriocanal:
```suggestion
    /// Measures how many transactions were created from segment spans in a single envelope.
```

relay-server/src/envelope.rs
```diff
@@ -554,6 +554,17 @@ pub struct ItemHeaders {
...
+    /// Whether or not a transaction has been extracted from a segment span.
+    #[serde(default, skip_serializing_if = "is_false")]
+    transaction_extracted: bool,
+
+    /// Whether or not spans and span metrics have been extracted from a transaction.
+    ///
+    /// This header is set to `true` after both span extraction and span metrics extraction,
+    /// and can be used to skip extraction.
+    #[serde(default, skip_serializing_if = "is_false")]
+    spans_extracted: bool,
```
Dav1dde: I stumbled over the usage of these fields a bit and I am wondering if it makes sense to change the behaviour a bit. Please bear with me and correct me if I misunderstood something.

If there is a transaction extracted from a span:
- The span gets `transaction_extracted = true`
- The new transaction gets `span_extracted = true` (that isn't actually true, we just need it so we skip span extraction, but this transaction never had a span extracted)

Would it make more sense to have the opposite behaviour where:
- The span gets nothing (except we need this info for some other reason later down the road)
- The transaction gets `extracted_from_span = True` (or `extracted = True / Span`)
- A span extracted from a transaction would get `extracted_from_transaction = True` (or `extracted = True / Transaction`)

We can use an `enum` for `extracted` which means we should be nicely future proof and it is quite readable (imo) and we indicate where something came instead of the opposite.

relay-server/src/statsd.rs
```diff
@@ -382,6 +382,9 @@ pub enum RelayTimers {
...
+    TransactionsFromSpansPerEnvelope,
```
iker-barriocanal: Should we move the variant to `RelayHistograms`?

iker-barriocanal (approved):

relay-server/src/services/processor/span/processing.rs
```diff
@@ -163,6 +181,47 @@ pub fn process(
...
+
+                transaction_count += 1;
+
+                addrs.envelope_processor.send(ProcessEnvelope {
+                    envelope: ManagedEnvelope::standalone(
+                        envelope,
+                        addrs.outcome_aggregator.clone(),
+                        addrs.test_store.clone(),
+                        ProcessingGroup::Transaction,
+                    ),
```
Dav1dde: I think this should be `ManagedEnvelope::new`, not sure if we should bypass the semaphore here?

relay-server/src/services/processor/span/processing.rs
```diff
@@ -327,6 +390,19 @@ fn get_normalize_span_config<'a>(
     }
 }
 
+fn set_segment_attributes(span: &mut Annotated<Span>) {
+    let Some(span) = span.value_mut() else { return };
+
+    // TODO: A span might be a segment span even if the parent_id is not empty
+    // (parent within a trace). I.e. do not overwrite here.
+    let is_segment = span.parent_span_id.is_empty();
```
Dav1dde, line 419: Leftover todo, can't we already implement this?

relay-sampling/src/evaluation.rs
Dav1dde: I see now you're just using it clone the envelope into a new one, maybe we should have a constructor which allows you to "fork off" an envelope from an existing one instead instead of exposing all the internals one by one.

relay-server/src/services/processor/span/processing.rs
jjbayer: It's kind off orthogonal to this PR (I only moved this code here). I will follow-up with @phacops though in a separate PR.

relay-server/src/envelope.rs
Dav1dde: We discussed this offline and came to the conclusion we should keep the current naming.

> I'm trying your suggestion [here](https://github.com/getsentry/relay/pull/3375/files#r1560777429) and it works, but I have two problems:
I still want to mark spans as transaction_extracted even if it happens only in processing relays, just to be robust against refactors
An extracted_from_span header is more truthful, but a spans_extracted header seems more robust against refactors (same argument as above, if we move any extraction code to edge relays it might be nice to prevent double extractions)
I.e. I believe we should have had a spans_extracted flag from the start. In theory we could have both an extracted_from_span and a spans_extracted header, but the mere amount of flags would make me dizzy

The alternative of using `skip_*_extraction` came up, but we may want to stick with the state rather then the behaviour so we can use that information (state) later on for more actions that just skipping extraction.

relay-server/src/metrics_extraction/event.rs
jjbayer: I kinda like the simple signature of `sample`, to be honest.

relay-sampling/src/evaluation.rs
jjbayer: > maybe we should have a constructor which allows you to "fork off" an envelope

I tried this (as a method on `ProcessEnvelopeState`) but it did not really improve semantics, because the fork requires some properties from the envelope, some from the state, and some from the processor itself. Maybe this will become more straight-forward when we clean up / merge `ManagedEnvelope` with `TypedEnvelope` and `ProcessEnvelopeState`.

relay-server/src/services/processor/span/processing.rs
jjbayer: Good call, I passed in the buffer guard now.

Dav1dde (approved):

jjbayer:
Update: verified that a transaction extracted from spans ends up in the product without errors: 
image

PR #3376, opened 2024-04-04
instr(metrics): Include namespace in store message variant

Dav1dde:
Includes the namespace in the variant name, this is an easy way to have all metrics tagged with the namespace, including the ones emitted from `relay-kafka`.

This will be very interesting to see for the metric bucket encoding change.

#skip-changelog

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -16,8 +16,8 @@ use relay_event_schema::protocol::{EventId, SessionStatus, VALID_PLATFORMS};
 
 use relay_kafka::{ClientError, KafkaClient, KafkaTopic, Message};
 use relay_metrics::{
-    Bucket, BucketView, BucketViewValue, BucketsView, FiniteF64, GaugeValue, MetricNamespace,
-    SetView,
+    Bucket, BucketView, BucketViewValue, BucketsView, FiniteF64, GaugeValue, MetricName,
+    MetricNamespace, SetView,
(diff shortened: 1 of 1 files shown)


jjbayer (commented):
To be consistent with

https://github.com/getsentry/relay/blob/d907bfda853f9d694f258a666d4b4c896a289150/relay-server/src/statsd.rs#L551-L552, 

should we introduce a namespace tag instead?

Dav1dde:
> should we introduce a namespace tag instead?

It's a bit annoying, the size metric is emitted from deep within `relay-kafka`, the `variant` is already one of these parameters that is passed in which doesn't really make sense and only exists for the metric.

I could add an `Option<MetricNamespace>` argument, and then conditionally include it in the metric but that is another (even more targeted) argument which doesn't make sense.

The other choice is moving the metric to a higher level where the message (not just bytes) still exists, but it makes sense that is in `relay-kafka`, it makes it impossible to *not* log the metric on accident.

Then there is adding another redundant metric ...

I ended up with changing the variant as the most consistent and easiest change.

iker-barriocanal (approved):

iker-barriocanal:
I think the change is fine -- it also makes comparing spans vs span metrics easier in the graphs.

jjbayer (approved):
OK! Might be worth checking existing dashboards, not sure if any of them filter on `"metric"`.

PR #3377, opened 2024-04-04
feat(normalization): Optionally skip normalization

iker-barriocanal:
This PR adds `processing.normalize_events` static configuration for Relay to:

- `disabled` to disable normalization for envelopes coming from internal relays.
- `default` to run normalization, except the steps that break future compatibility.
- `full` to run full normalization.

This change doesn't impact existing and new external relays or Self-hosted instances.

## Background

Events are normalized once in each relay. Sentry's infrastructure has two layers of relays, PoPs and processing relays, effectively running normalization twice. Since Sentry controls its infrastructure, we can optimize the status quo by only running normalization once. Events must be normalized before PoP events process them (e.g., extracting metrics), so the PoP relay layer runs it.

## Caveats

- Most of the time, traffic goes to PoPs and then processing relays. This may not be the case in some scenarios, where envelopes go directly to processing relays. Events should be normalized in both cases.
- Normalization runs a few steps that break the future compatibility but are required to run for the post-relay processing pipeline. These steps should only run in the last layer of internal relays.
- Self-hosted instances may only have one (processing) relay.

## When does normalization run

- Relay doesn't run normalization if `normalize_events=disabled` _and_ the envelope is forwarded from another internal relay (assuming the previous one previously run it). Normalization is run in every other case.
- When relay runs normalization, the full normalization is run when `normalize_events=full`, or `normalize_events=default` and it's a processing relay.

## Rollout

1. Deploy this PR.
2. Enable full normalization in PoPs and disable normalization in processing relays.
3. Deploy PoPs.
4. Deploy processing relays.

Notes: 
- Steps 2-4 will go to S4S before SaaS.
- PoPs (step 3) must be deployed before processing (step 4) relays, to avoid forwarding events without normalization. As a result, events will temporarily be fully normalized twice. I've introduced a temporary test to ensure full normalization is idempotent, which I'll remove after the rollout.


relay-config/src/config.rs
```diff
@@ -1014,6 +1014,18 @@ pub struct Processing {
     /// Maximum rate limit to report to clients.
     #[serde(default = "default_max_rate_limit")]
     pub max_rate_limit: Option<u32>,
+    /// True if normalization shouldn't run for events coming from internal relays. Defaults to `false`.
+    #[serde(default)]
+    pub disable_normalization: bool,
```
jjbayer: Should we flip this to `enable_normalization` with default `true`?

relay-config/src/config.rs
```diff
@@ -1014,6 +1014,18 @@ pub struct Processing {
...
+    /// Force relay to run full normalization. Defaults to `false`.
+    ///
+    /// Full normalization includes additional steps that break future
+    /// compatibility and should only run in the last layer of relays.
+    /// Processing relays run it by default, even if the flag is disabled.
+    ///
+    /// This config has no effect if normalization doesn't run.
+    #[serde(default)]
+    pub full_normalize: bool,
```
jjbayer: Should we have a single enum instead of two boolean flags? With values `Default`, `Full`, and `Skip`. Because the combination `(True, True)` does not make sense.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -1390,6 +1390,82 @@ mod tests {
         assert_eq!(&second, &third, "idempotency check failed");
     }
 
+    /// Validate full normalization is idempotent.
+    ///
+    /// Both PoPs and processing relays will temporarily run full normalization
+    /// in events, during the rollout of running normalization once in internal
+    /// relays.
+    // TODO(iker): remove this test after the rollout is done.
```
Dav1dde, line 1398: We should keep that, normalization should always be idempotent.

relay-server/src/services/processor.rs
```diff
@@ -1242,13 +1242,18 @@ impl EnvelopeProcessorService {
                 is_validated: false,
             };
 
-            let is_last_relay_normalize = self.inner.config.processing_enabled();
+            // Under normal operation, envelopes should go through PoP relays
+            // and then processing relays. However, during incidents, envelopes
+            // may go straight to processing without going through PoPs. Relays
+            // should run full normalization in both cases.
```
Dav1dde: This assumes a certain architecture, but we want to have a code which is architecture independent. 

What we want to support is ingestion through a chain of "trusted" Relays and only one of the Relays in the chain needs to do the work. That chain can consistent of a single Relay. The comment should reflect that. You can think of this way, will an outside reader understand that comment without knowing the infrastructure we're hosting Relay in?

relay-config/src/config.rs
Dav1dde: I agree we should have an enum here or a better naming, I know what the code is supposed to do and I am still confused.

relay-server/src/services/processor.rs
```diff
@@ -1368,6 +1375,18 @@ impl EnvelopeProcessorService {
         Ok(())
     }
 
+    /// Whether the event in the envelope should be normalized.
+    ///
+    /// Under normal operation, envelopes should go through PoP relays and then
+    /// processing relays. However, during incidents, envelopes may go straight
+    /// to processing without going through PoPs. Events should be normalized in
+    /// both cases, but should not be normalized twice (for efficiency).
```
Dav1dde: Again, we should describe an abstract architecture we want to support independent of what we already have. A chain of trusted Relays in which certain things only need to happen once (at the first layer). Which is independent of whether something is a processing Relay or not.

relay-config/src/config.rs
iker-barriocanal: Merge two configs into an enum in [83894b6](https://github.com/getsentry/relay/pull/3377/commits/83894b69eefca3bb456cb850fe2ac84decff41e1).

relay-config/src/config.rs
iker-barriocanal: Done in [83894b6](https://github.com/getsentry/relay/pull/3377/commits/83894b69eefca3bb456cb850fe2ac84decff41e1).

relay-event-normalization/src/normalize/mod.rs
iker-barriocanal: Default normalization, yes; but I'm not so sure about non-future compatible steps too. The test above this one verifies default normalization is idempotent.

relay-server/src/services/processor.rs
iker-barriocanal: With the config change in [83894b6](https://github.com/getsentry/relay/pull/3377/commits/83894b69eefca3bb456cb850fe2ac84decff41e1) this comment no longer exists. I think it's a good idea to generalize the code, but it's non-trivial to do at this point and it's harder to reason about the config values set (FWIW, I'd like to have something simpler than what this PR presents).

So I'd like to move forward with the "is this a processing relay?" check, and reevaluate the code if/when we make bigger architectural changes.

relay-server/src/services/processor.rs
iker-barriocanal: Replied to the comment above in https://github.com/getsentry/relay/pull/3377#discussion_r1555914322, resolving this one to keep the convo focused on the previous comment.

relay-config/src/config.rs
```diff
@@ -1014,6 +1014,9 @@ pub struct Processing {
     /// Maximum rate limit to report to clients.
     #[serde(default = "default_max_rate_limit")]
     pub max_rate_limit: Option<u32>,
+    /// Granularity of normalization this event should do.
```
```suggestion
    /// How much normalization this Relay should apply to incoming data.
```

relay-config/src/config.rs
```diff
@@ -1014,6 +1014,9 @@ pub struct Processing {
...
+    #[serde(default)]
+    pub normalize_events: NormalizeEvents,
```
jjbayer: Should we strip the `_events` suffix here? We are normalizing more and more non-event data items, and I guess the logic should (eventually) apply regardless of the data type?

```suggestion
    pub normalize: Normalize,
```

relay-config/src/config.rs
```diff
@@ -1032,10 +1035,29 @@ impl Default for Processing {
...
+            normalize_events: NormalizeEvents::Default,
         }
     }
 }
 
+/// Configuration for the level of normalization this Relay should do.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum NormalizeEvents {
+    /// Disables normalization for events coming from internal relays.
```
jjbayer: etc.

```suggestion
    /// Disables normalization for events coming from internal Relays.
```

relay-server/src/services/processor.rs
```diff
@@ -1212,6 +1212,25 @@ impl EnvelopeProcessorService {
...
+                // routed differently, skipping some relays and reaching another
+                // internal one. Events should be fully normalized in all cases,
+                // so we force normalization even if it's disabled.
+                if self.inner.config.processing_enabled()
+                    && !state.envelope().meta().is_from_internal_relay()
+                {
+                    true
+                } else {
+                    return Ok(());
+                }
```
jjbayer: Sorry for going in circles here, but I'm in favor of removing the extra check and simply skipping normalization unconditionally, as discussed with @jan-auer. We just need to make sure that we track down any playbooks that suggest sending data straight to processing Relays before we enable this config in production.

relay-config/src/config.rs
iker-barriocanal: Done in [d936ff4](https://github.com/getsentry/relay/pull/3377/commits/d936ff4d29e3814fdc916f42dd955d03cbb8dfa1).

relay-config/src/config.rs
iker-barriocanal: Done in [d936ff4](https://github.com/getsentry/relay/pull/3377/commits/d936ff4d29e3814fdc916f42dd955d03cbb8dfa1).

relay-config/src/config.rs
iker-barriocanal: Done in [d936ff4](https://github.com/getsentry/relay/pull/3377/commits/d936ff4d29e3814fdc916f42dd955d03cbb8dfa1).

relay-server/src/services/processor.rs
iker-barriocanal: After discussing this, we're going forward with this approach -- mostly due to increased reliability, minimal complexity increase, and it can be updated in the future with minimal breaking changes*.

*Breaking changes would only impact Self-Hosted instances with at least two layers of relays and where the last layer of relay also gets traffic (not just the outer ones).

relay-server/src/services/processor.rs
```diff
@@ -1210,6 +1210,25 @@ impl EnvelopeProcessorService {
         &self,
         state: &mut ProcessEnvelopeState<G>,
     ) -> Result<(), ProcessingError> {
+        let full_normalization = match self.inner.config.normalization() {
+            Normalize::Disabled => {
+                // We assume envelopes coming from an internal relay have
+                // already been normalized.  During incidents, like a PoP region
+                // not being available, envelopes can go to other PoP regions or
+                // directly to processing relays.  Events should be fully
+                // normalized, independently of the ingestion path.
```
```suggestion
                // We assume envelopes coming from an internal relay have
                // already been normalized. During incidents, like a PoP region
                // not being available, envelopes can go to other PoP regions or
                // directly to processing relays. Events should be fully
                // normalized, independently of the ingestion path.
```

jjbayer (approved):

PR #3378, opened 2024-04-04
fix(spool): Fix performance regression when enqueueing envelopes

Dav1dde:
For every `Enqueue` message we execute an insert and then afterwards run the `current_size` query on the database, which becomes slower with every additional byte inserted into the table. After 500mb inserted each query already takes more than 250ms!

The new query may slightly be under estimating the actual size of the database but should be close enough for our needs.

In the future we should re-consider this current mechanism and keep track of the database size in memory or other means, I didn't want to refactor the spooling code and potentially introduce new bugs now, this should fix our immediate problem. We can revisit the architecture of the spool at a later point.

In my very simple tests it showed that inserting in a batch by constructing multiple statements via the query builder vs multiple separate insert statements made no noticeable difference.

### DIFF
--- a/relay-server/src/services/spooler/sql.rs
+++ b/relay-server/src/services/spooler/sql.rs
@@ -70,11 +70,11 @@ pub fn delete<'a>(key: QueueKey) -> Query<'a, Sqlite, SqliteArguments<'a>> {
         .bind(key.sampling_key.to_string())
 }
 
-/// Creates a query which fetches the `envelopes` table size.
+/// Creates a query which estimates the `envelopes` table size.
 ///
-/// This info used to calculate the current allocated database size.
-pub fn current_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
-    sqlx::query(r#"SELECT SUM(pgsize - unused) FROM dbstat WHERE name="envelopes""#)
+/// This info used to estimate the current allocated database size.
+pub fn estimate_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
+    sqlx::query(r#"SELECT SUM(length(envelope)) FROM envelopes"#)
 }
 
 /// Creates the query to select only 1 record's `received_at` from the database.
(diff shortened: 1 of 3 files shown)


relay-server/src/services/spooler/sql.rs
```diff
@@ -70,11 +70,11 @@ pub fn delete<'a>(key: QueueKey) -> Query<'a, Sqlite, SqliteArguments<'a>> {
...
 
-/// Creates a query which fetches the `envelopes` table size.
+/// Creates a query which estimates the `envelopes` table size.
 ///
-/// This info used to calculate the current allocated database size.
-pub fn current_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
-    sqlx::query(r#"SELECT SUM(pgsize - unused) FROM dbstat WHERE name="envelopes""#)
+/// This info used to estimate the current allocated database size.
+pub fn estimate_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
+    sqlx::query(r#"SELECT SUM(length(envelope)) FROM envelopes"#)
```
olksdr, line 77: This way we still need to iterate the table ,
what if we actually revert to the initial query 
```rust
sqlx::query(
        "SELECT page_count * page_size as size FROM pragma_page_count(), pragma_page_size();",
    )
```

which should be much faster but also not exact number , but estimate

Dav1dde: Sounds good, the `sum(length())` is pretty fast, but the page count/size is still a bit faster, the values are very similar:

```
dbstat: 492855289
sum(length(envelope)): 491520000
page_size * page_count: 507478016
```

Dav1dde: This breaks for small spools (in tests):

```
dbstat:                 3308
sum(length(envelope)):  1600
page_size * page_count: 16384
```

Dav1dde: Also are the pages reclaimed when shrinking the spool, it doesn't help us if the DB reserved a bunch of data/memory and it is actually unused.

Dav1dde:
Identified the problem, the "fix" has its own problems. It scales better with fewer big items, but worse with many smaller items.

PR #3379, opened 2024-04-05
fix(spool): Revert to page counts for size estimate

jjbayer:
INC-703 showed that `estimate_spool_size` via  `dbstat` is too slow. This PR replaces the estimate with the number of non-free pages multiplied by the page size. Though inaccurate, benchmarks show that this is fast regardless of the db size.

Fixes https://github.com/getsentry/team-ingest/issues/307.

### DIFF
--- a/relay-server/src/services/spooler/mod.rs
+++ b/relay-server/src/services/spooler/mod.rs
@@ -689,7 +689,7 @@ impl OnDisk {
         }
     }
 
-    /// Estimates the db size by multiplying `page_count * page_size`.
+    /// Estimates the db size by multiplying `used_page_count * page_size`.
     async fn estimate_spool_size(&self) -> Result<i64, BufferError> {
         let size: i64 = sql::estimate_size()
             .fetch_one(&self.db)
--- a/relay-server/src/services/spooler/mod.rs
+++ b/relay-server/src/services/spooler/mod.rs
@@ -691,7 +691,7 @@ impl OnDisk {
 
     /// Estimates the db size by multiplying `page_count * page_size`.
     async fn estimate_spool_size(&self) -> Result<i64, BufferError> {
-        let size: i64 = sql::current_size()
+        let size: i64 = sql::estimate_size()
             .fetch_one(&self.db)
             .await
             .and_then(|r| r.try_get(0))
--- a/relay-server/src/services/spooler/mod.rs
+++ b/relay-server/src/services/spooler/mod.rs
@@ -1268,7 +1268,7 @@ impl Drop for BufferService {
 mod tests {
     use std::str::FromStr;
     use std::sync::Mutex;
-    use std::time::Duration;
+    use std::time::{Duration, Instant};
 
     use insta::assert_debug_snapshot;
     use rand::Rng;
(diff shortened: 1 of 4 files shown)


relay-server/src/services/spooler/mod.rs
```diff
@@ -1569,26 +1569,25 @@ mod tests {
             "buffer.writes:1|c",
             "buffer.envelopes_written:3|c",
             "buffer.envelopes_disk_count:3|g",
-            "buffer.disk_size:1031|h",
+            "buffer.disk_size:24576|h",
```
jjbayer, line 1572: This is pretty far off. But the relative error should be lower for larger disk buffers.

relay-server/src/services/spooler/sql.rs
```diff
@@ -74,7 +74,9 @@ pub fn delete<'a>(key: QueueKey) -> Query<'a, Sqlite, SqliteArguments<'a>> {
 ///
 /// This info used to calculate the current allocated database size.
 pub fn current_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
```
Dav1dde: Can you update the docs for this function and the calling function?

Dav1dde:
```suggestion
pub fn estimate_size<'a>() -> Query<'a, Sqlite, SqliteArguments<'a>> {
```

relay-server/src/services/spooler/mod.rs
```diff
@@ -1879,4 +1878,79 @@ mod tests {
...
+            r#"SELECT (page_count - freelist_count) * page_size as size FROM pragma_page_count(), pragma_freelist_count(), pragma_page_size();"#,
+        );
+        let pragma: i64 = q.fetch_one(&db).await.unwrap().get(0);
+        println!("pragma: {} {:?}", pragma, t.elapsed());
+
+        // Result:
+        // pgsize =      2'408'464'463 t.elapsed() = 7.007533833s
+        // pgsize_agg =  2'408'464'463 t.elapsed() = 5.010104791s
+        // brute_force = 1'750'000'000 t.elapsed() = 7.893590875s
+        // pragma =      3'036'307'456 t.elapsed() = 213.417µs
```
Dav1dde, line 1954: Nice! But it over estimates by quite a bit > 1 GiB?

Edit: To clarify I think that's okay, just surprising.

relay-server/src/services/spooler/sql.rs
jjbayer: 👍 fun fact: The doc on the calling function was already correct because we never updated it.

Dav1dde: Not quite, it didn't subtract the free pages.

Dav1dde (approved):
I think this is changelog worthy, in my PR I had:

```
**Bug Fixes**:

- Fix performance regression in the spooler. ([#3378](https://github.com/getsentry/relay/pull/3378))
```

olksdr (approved):

relay-server/src/services/spooler/mod.rs
jjbayer: Yep the estimates are pretty far off, this is something we will have to keep in mind.

jjbayer:
As discussed with @olksdr: Because the spool size estimation is also used to decide whether or not to transition back to memory mode, and the new computation over-estimates the spool size, this PR might require a follow-up if we notice that Relay gets stuck in disk mode.

With the performance regression fixed, getting stuck in disk mode should not be a big problem performance-wise. We could even consider operating _always_ in disk mode to simplify the state machine.

PR #3380, opened 2024-04-05
feat(spans): Remove count_per_segment & count_per_op metric

shruthilayaj:
We didn't end up doing the work to use this and pivoted to other features.
We also collect the information we need with the the exclusive_time metric
now, so this can be removed.

No references to this metric in the sentry backend.


phacops (commented):
Can you kill `count_per_op` as well? This was an internal metric we don't use anymore.

phacops (approved):

Dav1dde (approved):
> No references to this metric in the sentry backend.

They are referenced in [strings.py](https://github.com/getsentry/sentry/blob/15d52c5fd679a57561cf506cddd17570c1722d32/src/sentry/sentry_metrics/indexer/strings.py#L221-L222) but that should probably stay.

cc: @jjbayer jfyi

phacops:
> They are referenced in [strings.py](https://github.com/getsentry/sentry/blob/15d52c5fd679a57561cf506cddd17570c1722d32/src/sentry/sentry_metrics/indexer/strings.py#L221-L222) but that should probably stay.

Indeed since this file is frozen and we do not want to modify it.

PR #3383, opened 2024-04-05
feat(perf): Extract scrubbed IP addresses into the `span.domain` tag

gggritso:
tl;dr URLs with IP domains like `http://8.8.8.8/data` get the `span.domain` tag of `*.*.*.*`. Also a small refactor to support this.

Right now, if a span's description looks like `GET http://8.8.8.8/data` the scrubbed `span.description` is `GET http://*.*.8.8/` and `span.domain` is `None`. This is inconsistent, and also makes for a bad user experience because people cannot find requests to certain IPs by the domain.

There are two causes:

1. IP addresses in descriptions are ran through domain scrubbing, even though they're not domains. This turns `8.8.8.8` into `*.*.8.8` because the scrubber thinks the last two 8s are a TLD
2. The scrubbed descriptions are parsed by `URL` and are invalid. The parsing fails, and no `domain` tag is produced.

This PR makes some refactors, and turns off scrubbing of IP host strings.

## Changes

- `normalize_domain` --> `scrub_host`. Instead of a function that takes a domain string and a port and does the asterisking, the function accepts a `Url::Host` and does different things based on whether it's an IP or a qualified domain. This is nicer because the name is clearer and the function does fewer things
- `scrub_domain_name` is a new function that _only_ replaces a string with asterisks. This makes it possible to call this function independently of other scrubbing logic (e.g., to forgo calling it for IP addresses)
- `concatenate_host_and_port` is a simple concatenator, since I saw this logic in a few places
- `scrub_ipv4` and `scrub_ipv6` that fully scrub out IP addresses

## Questions

This will increase the cardinality by the number of ports used with IP addresses in the wild.

I think we'll probably want to do some Big Query research to see how many unique IPs are in the dataset, and maybe only _partially_ scrub the IP (e.g., `*.*.*.67`).

Otherwise, there are code-level questions:

~1. Should I separate this PR out into the refactor and a separate PR to disable IP scrubbing?~
2. I had a hard time deciding when to return/accept `&str` and `String`. I ended up returning `String` in most places, since I think the functions that get the object back to should ownership?
3. Overall, I'm not confident about Rust style here, you tell me!


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +194,78 @@ fn scrub_file(description: &str) -> Option<String> {
...
+/// Scrub a `Url::Host` object. For domain names, run the string through a scrubber. For IP addresses, return the IP address as-is.
+///
+/// # Arguments
+///
+/// * `host` - `Url::Host`
+///
+/// # Returns
+///
+/// A host `String`, or `None` if scrubbing fails.
+pub fn scrub_host(host: Option<Host<&str>>) -> Option<String> {
```
jjbayer: This function maps `Some` to `Some` and `None` to `None`, so should we change the signature to `scrub_host(host: Host<&str>) -> String` and call `maybe_host.map(scrub_host)` on the option?

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +194,78 @@ fn scrub_file(description: &str) -> Option<String> {
...
-        let with_port = port.map_or_else(|| (*domain).to_owned(), |p| format!("{}:{}", domain, p));
-        return Some(with_port);
+    if port.is_none() {
+        return host.to_string();
     }
-    None
+
+    let port = port.unwrap();
+
+    return format!("{host}:{port}");
```
jjbayer: > I had a hard time decising when to return/accept &str and String.

`&str` is a reference and `String` is owned data, so a `&str` is preferable if possible. When a function may return either owned or borrowed data, `Cow<str>` is the way to go.

2. `match` expressions are preferable over `is_none` + `unwrap`:

```rust
match (host, port) {
    (None, _) => Cow::Borrowed(""),
    (host, None) => Cow::Borrowed(host),
    (Some(host), Some(port)) => Cow::Owned(format!("{host}:{port}"))
}
```

jjbayer:
@gggritso thank you for picking this up and for improving separation of concerns!

> I think we'll probably want to do some Big Query research to see how many unique IPs are in the dataset, but maybe it's safe enough to try and see?

That would be OK with me! As long as we follow up with some cardinality analysis.

> 1. Should I separate this PR out into the refactor and a separate PR to disable IP scrubbing?

That would be ideal, but IMO the PR is still small enough to keep them both in one.

> 2. I had a hard time decising when to return/accept `&str` and `String`. I ended up returning `String` in most places, since I think the functions that get the object back to should ownership?

See PR comment.

relay-event-normalization/src/normalize/span/description/mod.rs
gggritso: @jjbayer I made the change to return `Cow` from this method, but it caused a bit of gross code in another part, where I had to do some somersaults:

```rust
                    Some(String::from(concatenate_host_and_port(
                        Some(domain.as_str()),
                        port,
                    )))
```

Admittedly, I don't know what I'm doing, I only have a very basic understanding of `Cow`. Should I do something different, or clean this up in another way, or is this fine?

relay-event-normalization/src/normalize/span/description/mod.rs
gggritso: Done!

gggritso:
@jjbayer I made a few major changes:

1. No more actual extraction of IP addresses, all IPs are _fully scrubbed out_
2. Tried to address your feedback, but I'm not good at Rust, so I might have done a bad job.

Thanks for taking an early look!

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +195,106 @@ fn scrub_file(description: &str) -> Option<String> {
...
+/// Scrubbed string
+pub fn scrub_ipv4(ip: Ipv4Addr) -> String {
+    if IPV4_ALLOW_LIST.contains(&ip.to_string().as_str()) {
+        return ip.to_string();
+    }
+
+    String::from("*.*.*.*")
+}
+
+const IPV4_ALLOW_LIST: [&str; 1] = ["127.0.0.1"];
```
jjbayer: nit: We [usually](https://develop.sentry.dev/rust/#declaration-order) define constants before any other definitions, even if they are not public.

jjbayer: Let's make this an array of `Ipv4Addr`, so the check does not have to allocate a `String` (same for IPv6).

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +195,106 @@ fn scrub_file(description: &str) -> Option<String> {
...
+///
+/// # Returns
+///
+/// Scrubbed string
+pub fn scrub_ipv4(ip: Ipv4Addr) -> String {
+    if IPV4_ALLOW_LIST.contains(&ip.to_string().as_str()) {
+        return ip.to_string();
+    }
+
+    String::from("*.*.*.*")
```
jjbayer: This could be a `Cow::Borrowed("*.*.*.*")` (string literals have static lifetime, so can always be borrowed from).

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +195,106 @@ fn scrub_file(description: &str) -> Option<String> {
...
+/// * `host` - Hostname string
+/// * `port` - Port number
+///
+/// # Returns
+///
+/// Either an empty string or concatenated host and port
+pub fn concatenate_host_and_port(host: Option<&str>, port: Option<u16>) -> Cow<str> {
+    match (host, port) {
+        (None, _) => Cow::Borrowed(""),
+        (host, None) => Cow::Borrowed(host.unwrap()),
```
```suggestion
        (Some(host), None) => Cow::Borrowed(host),
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -293,10 +350,9 @@ fn scrub_resource(resource_type: &str, string: &str) -> Option<String> {
             return Some("browser-extension://*".to_owned());
         }
         scheme => {
-            let domain = url
-                .domain()
-                .and_then(|d| normalize_domain(d, url.port()))
-                .unwrap_or("".into());
+            let scrubbed_host = url.host().map(scrub_host).unwrap_or(String::from(""));
```
jjbayer: With the comments above, if `scrub_ipv4` becomes a `Cow`, `scrub_host` returns a `Cow` => can also borrow here.
```suggestion
            let scrubbed_host = url.host().map(scrub_host).unwrap_or(Cow::Borrowed(""));
```

gggritso:
@jjbayer made another round of changes! Should be pretty close now 🙌🏻

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -30,6 +31,13 @@ const MAX_SEGMENT_LENGTH: usize = 25;
 /// Some bundlers attach characters to the end of a filename, try to catch those.
 const MAX_EXTENSION_LENGTH: usize = 10;
 
+/// IP addresses that are preserved during scrubbing
+const IPV4_ALLOW_LIST: [Ipv4Addr; 1] = [Ipv4Addr::LOCALHOST];
+const IPV6_ALLOW_LIST: [Ipv6Addr; 1] = [Ipv6Addr::LOCALHOST];
+
+/// Domain names that are preserved during scrubbing
+const DOMAIN_ALLOW_LIST: [&str; 1] = ["localhost"];
```
Dav1dde:
```suggestion
const IPV4_ALLOW_LIST: &[Ipv4Addr] = &[Ipv4Addr::LOCALHOST];
const IPV6_ALLOW_LIST: &[Ipv6Addr] = [&Ipv6Addr::LOCALHOST];

/// Domain names that are preserved during scrubbing
const DOMAIN_ALLOW_LIST: &[&str] = &["localhost"];
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -165,9 +173,9 @@ fn scrub_http(string: &str) -> Option<String> {
 
     let scrubbed = match Url::parse(url) {
         Ok(url) => {
-            let host = url.host().map(|h| h.to_string())?;
-            let domain = normalize_domain(host.as_str(), url.port())?;
             let scheme = url.scheme();
+            let scrubbed_host = url.host().map(scrub_host).unwrap_or(Cow::Borrowed(""));
+            let domain = concatenate_host_and_port(Some(scrubbed_host.as_ref()), url.port());
```
Dav1dde:
```suggestion
            let scrubbed_host = url.host().map(scrub_host);
            let domain = concatenate_host_and_port(scrubbed_host.as_deref(), url.port());
```

Why do the `unwrap_or` if the function already takes an option?

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +201,99 @@ fn scrub_file(description: &str) -> Option<String> {
...
-/// Normalize the domain and port of a URL.
+/// Scrub a `Url::Host` object. For domain names, run the string through a scrubber. For IP addresses, return the IP address as-is.
+///
+/// # Arguments
+///
+/// * `host` - `Url::Host`
+///
+/// # Returns
+///
+/// A host `String`, or `None` if scrubbing fails.
```
Dav1dde: We don't usually document arguments like here, but I wouldn't mind if it provided actual additional information which can't already be gained from the type system.

You'd gain more from adding an example instead (which would also double down as a unittest).

I'd use something like this:

```suggestion
/// Scrub a [`Host`] object. 
///
/// Domain names are run through a scrubber. All IP addresses except well known ones are replaced with a scrubbed variant.
/// Returns the scrubbed value as a string.
///
/// # Examples
///
/// ```
/// assert_eq!(&scrub_host(Host::Domain("foo.bar.baz"), "expected value"));
/// assert_eq!(&scrub_host(Host::Ipv4(Ipv4Addr::LOCALHOST), "127.0.0.1"));
/// ...
/// ```
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +201,99 @@ fn scrub_file(description: &str) -> Option<String> {
     }
 }
 
-/// Normalize the domain and port of a URL.
+/// Scrub a `Url::Host` object. For domain names, run the string through a scrubber. For IP addresses, return the IP address as-is.
```
Dav1dde: Note there are multiple errors in these docs!

In this one:
- IP addresses are scrubbed.
- Return value is not an option.
- Return value is not a `String`.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +201,99 @@ fn scrub_file(description: &str) -> Option<String> {
...
+///
+/// # Arguments
+///
+/// * `host` - `Url::Host`
+///
+/// # Returns
+///
+/// A host `String`, or `None` if scrubbing fails.
+pub fn scrub_host(host: Host<&str>) -> Cow<'static, str> {
```
Dav1dde:
```suggestion
pub fn scrub_host(host: Host<&str>) -> Cow<'_, str> {
```

No need to confine this to `'static`. You can now return the host name contained in `Host` without cloning.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +201,99 @@ fn scrub_file(description: &str) -> Option<String> {
...
+/// # Arguments
+///
+/// * `ip` - IP to be scrubbed
+///
+/// # Returns
+///
+/// Scrubbed string
+pub fn scrub_ipv4(ip: Ipv4Addr) -> Cow<'static, str> {
+    if IPV4_ALLOW_LIST.contains(&ip) {
+        return Cow::Owned(ip.to_string());
```
Dav1dde: It's possible to never allocate here and let the function also return `&'static str` without the `Cow` by storing the ip addresses as `&[(Ipv4Addr, &'static str)]`, e.g. `&[(Ipv4Addr::LOCALHOST, "127.0.0.1")]`.

Something like this:
```rs
IPV4_ALLOW_LIST.into_iter().find_map(|(s, repr)| (s == ip).then_some(repr)).unwrap_or("*.*.*.*")
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +201,99 @@ fn scrub_file(description: &str) -> Option<String> {
...
+/// * `domain` - Domain name (e.g., "my.domain.com")
 ///
 /// # Returns
 ///
-/// The normalized domain and port as a `String`, or `None` if normalization fails.
-pub fn normalize_domain(domain: &str, port: Option<u16>) -> Option<String> {
-    if let Some(allow_listed) = normalized_domain_from_allowlist(domain, port) {
-        return Some(allow_listed);
+/// The scrubbed domain as a `String`, or `None` if normalization fails.
+pub fn scrub_domain_name(domain: String) -> String {
```
Dav1dde: There is no need for this to always take a `String`.

```suggestion
pub fn scrub_domain_name(domain: &str) -> Cow<'_, str> {
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -193,51 +201,99 @@ fn scrub_file(description: &str) -> Option<String> {
...
     let tld = tokens.next();
     let domain = tokens.next();
     let prefix = tokens.next().map(|_| "*");
 
-    let mut replaced = prefix
+    return prefix
         .iter()
         .chain(domain.iter())
         .chain(tld.iter())
         .join(".");
```
Dav1dde:
```suggestion
    prefix
        .iter()
        .chain(domain.iter())
        .chain(tld.iter())
        .join(".")
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -15,11 +17,13 @@ use sqlparser::ast::Visit;
 use sqlparser::ast::{ObjectName, Visitor};
 use url::Url;
 
-use crate::span::description::{normalize_domain, scrub_span_description};
+use crate::span::description::scrub_span_description;
 use crate::utils::{
     extract_transaction_op, http_status_code_from_span, MAIN_THREAD_NAME, MOBILE_SDKS,
 };
 
+use super::description::{concatenate_host_and_port, scrub_domain_name};
```
Dav1dde: We use crate imports, as they already exist above.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -293,10 +349,9 @@ fn scrub_resource(resource_type: &str, string: &str) -> Option<String> {
             return Some("browser-extension://*".to_owned());
         }
         scheme => {
-            let domain = url
-                .domain()
-                .and_then(|d| normalize_domain(d, url.port()))
-                .unwrap_or("".into());
+            let scrubbed_host = url.host().map(scrub_host).unwrap_or(Cow::Borrowed(""));
+            let domain = concatenate_host_and_port(Some(scrubbed_host.as_ref()), url.port());
```
Dav1dde: Same as above:
You unwrap the option just to create another option


```suggestion
            let scrubbed_host = url.host().map(scrub_host);
            let domain = concatenate_host_and_port(scrubbed_host.as_deref(), url.port());
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -358,16 +362,25 @@ pub fn extract_tags(
...
+
+                    // According to OTel semantic conventions the server port should be in a separate property, called `server.port`, but incoming data sometimes disagrees
+                    let (domain, port) = match lowercase_address.split_once(':') {
                         Some((domain, port)) => (domain, port.parse::<u16>().ok()),
-                        None => (server_host, None),
+                        None => (server_address, None),
+                    };
+
+                    // Leave IP addresses alone. Scrub qualified domain names
+                    let domain = if let Ok(address) = IpAddr::from_str(domain) {
```
Dav1dde: You shouldn't call `from_str` directly, it is invoked through `str::parse`:

```suggestion
                    let domain = if let Ok(address) = domain.parse::<IpAddr>() {
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -2,7 +2,9 @@
 //! These are then used for metrics extraction.
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Write;
+use std::net::IpAddr;
 use std::ops::ControlFlow;
+use std::str::FromStr;
```
Dav1dde: Use `FromStr` through `str::parse` and you don't need the import.

Dav1dde (changes requested):

relay-event-normalization/src/normalize/span/description/mod.rs
jjbayer: I actually considered this, but thought it might get a bit convoluted. Though since both allowlists only contain a single value, we could probably also hard code here, i.e. `if ip == Ipv4Addr::LOCALHOST { return "127.0.0.1" }`.

relay-event-normalization/src/normalize/span/description/mod.rs
jjbayer: Nice!

relay-event-normalization/src/normalize/span/description/mod.rs
Dav1dde: Or just have a big match instead of the static list?

relay-event-normalization/src/normalize/span/description/mod.rs
gggritso: Surprised the linter didn't catch this!

Dav1dde: Same :(

gggritso:
@Dav1dde @jjbayer thanks for bearing with me, and for the thorough review 🙏🏻 I tried to implement all your suggestions, but I might have missing something, please let me know! The major changes:

- `Cow` in a lot more places
- no more allowlists for IPs, using a `match` instead
- better function docs, including unit tests
- removed unnecessary casting

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -197,51 +201,114 @@ fn scrub_file(description: &str) -> Option<String> {
...
+/// Returns the scrubbed value as a string.
+///
+/// # Examples
+///
+/// ```
+/// use std::net::{Ipv4Addr};
+/// use relay_event_normalization::span::description::{scrub_ipv4};
+///
+/// assert_eq!(scrub_ipv4(Ipv4Addr::LOCALHOST), "127.0.0.1");
+/// assert_eq!(scrub_ipv4(Ipv4Addr::new(8, 8, 8, 8)), String::from("*.*.*.*"));
```
Dav1dde:
```suggestion
/// assert_eq!(scrub_ipv4(Ipv4Addr::new(8, 8, 8, 8)), "*.*.*.*");
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -197,51 +201,114 @@ fn scrub_file(description: &str) -> Option<String> {
...
+}
+///
+/// # Examples
+///
+/// ```
+/// use std::net::{Ipv6Addr};
+/// use relay_event_normalization::span::description::{scrub_ipv6};
+///
+/// assert_eq!(scrub_ipv6(Ipv6Addr::LOCALHOST), "::1");
+/// assert_eq!(scrub_ipv6(Ipv6Addr::new(8, 8, 8, 8, 8, 8, 8, 8)), String::from("*:*:*:*:*:*:*:*"));
```
Dav1dde:
```suggestion
/// assert_eq!(scrub_ipv6(Ipv6Addr::new(8, 8, 8, 8, 8, 8, 8, 8)), "*:*:*:*:*:*:*:*");
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -197,51 +201,114 @@ fn scrub_file(description: &str) -> Option<String> {
...
+    match ip {
+        Ipv4Addr::LOCALHOST => "127.0.0.1",
+        _ => "*.*.*.*",
+    }
+}
+///
+/// # Examples
+///
+/// ```
+/// use std::net::{Ipv6Addr};
```
Dav1dde:
```suggestion
/// use std::net::Ipv6Addr;
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -197,51 +201,114 @@ fn scrub_file(description: &str) -> Option<String> {
...
+
+/// Scrub an IPv4 address.
+///
+/// Allow well-known IPs like loopback, and fully scrub out all other IPs.
+/// Returns the scrubbed value as a string.
+///
+/// # Examples
+///
+/// ```
+/// use std::net::{Ipv4Addr};
```
Dav1dde:
```suggestion
/// use std::net::Ipv4Addr;
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -197,51 +201,114 @@ fn scrub_file(description: &str) -> Option<String> {
...
+/// assert_eq!(scrub_ipv4(Ipv4Addr::LOCALHOST), "127.0.0.1");
+/// assert_eq!(scrub_ipv4(Ipv4Addr::new(8, 8, 8, 8)), String::from("*.*.*.*"));
+/// ```
+pub fn scrub_ipv4(ip: Ipv4Addr) -> &'static str {
+    match ip {
+        Ipv4Addr::LOCALHOST => "127.0.0.1",
+        _ => "*.*.*.*",
+    }
+}
+///
```
Dav1dde, line 249: Missing the function description.

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -197,51 +201,114 @@ fn scrub_file(description: &str) -> Option<String> {
...
+/// Returns a string. In cases where the string is not domain-like, returns the original string.
+///
+/// # Examples
 ///
-/// The normalized domain and port as a `String`, or `None` if normalization fails.
-pub fn normalize_domain(domain: &str, port: Option<u16>) -> Option<String> {
-    if let Some(allow_listed) = normalized_domain_from_allowlist(domain, port) {
-        return Some(allow_listed);
+/// ```
+/// use relay_event_normalization::span::description::{scrub_domain_name};
```
Dav1dde:
```suggestion
/// use relay_event_normalization::span::description::scrub_domain_name;
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -197,51 +201,114 @@ fn scrub_file(description: &str) -> Option<String> {
...
 ///
-/// * `domain` - The domain of the URL.
-/// * `port` - The port of the URL.
+/// # Examples
 ///
-/// # Returns
+/// ```
+/// use url::{Host, Url};
+/// use std::net::{Ipv4Addr, Ipv6Addr};
+/// use relay_event_normalization::span::description::{scrub_host};
```
Dav1dde:
```suggestion
/// use relay_event_normalization::span::description::scrub_host;
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -197,51 +201,114 @@ fn scrub_file(description: &str) -> Option<String> {
...
-fn normalized_domain_from_allowlist(domain: &str, port: Option<u16>) -> Option<String> {
-    if let Some(domain) = DOMAIN_ALLOW_LIST.iter().find(|allowed| **allowed == domain) {
-        let with_port = port.map_or_else(|| (*domain).to_owned(), |p| format!("{}:{}", domain, p));
-        return Some(with_port);
+/// Concatenate an optional host and port
+///
+/// # Arguments
+///
+/// * `host` - Hostname string
+/// * `port` - Port number
```
Dav1dde: Doesn't really add info, imo just rewrite it like the other docs.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -378,10 +391,14 @@ pub fn extract_tags(
...
                         span_tags.insert(
                             SpanTagKey::RawDomain,
-                            format!("{url_scheme}://{lowercase_host}"),
+                            format!("{url_scheme}://{lowercase_address}"),
                         );
                     }
-                    normalize_domain(domain, port)
+
+                    Some(String::from(concatenate_host_and_port(
+                        Some(domain.to_string().as_str()),
```
```suggestion
                        Some(domain.as_ref()),
```

If you just need a `&str` from a `Cow` you don't have to own it first. If you want an owned version of a `Cow` and don't need it later anymore, you should use `into_owned` instead, which does not allocate if you already had a `Cow::Owned`.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -358,16 +362,25 @@ pub fn extract_tags(
...
+                    // According to OTel semantic conventions the server port should be in a separate property, called `server.port`, but incoming data sometimes disagrees
+                    let (domain, port) = match lowercase_address.split_once(':') {
                         Some((domain, port)) => (domain, port.parse::<u16>().ok()),
-                        None => (server_host, None),
+                        None => (server_address, None),
+                    };
+
+                    // Leave IP addresses alone. Scrub qualified domain names
+                    let domain = if let Ok(address) = domain.parse::<IpAddr>() {
+                        Cow::Owned(address.to_string())
```
Dav1dde: Can't this be:

```suggestion
                    let domain = if domain.parse::<IpAddr>().is_ok() {
                        Cow::Borrowed(domain)
```

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -378,10 +391,14 @@ pub fn extract_tags(
                     {
                         span_tags.insert(
                             SpanTagKey::RawDomain,
-                            format!("{url_scheme}://{lowercase_host}"),
+                            format!("{url_scheme}://{lowercase_address}"),
                         );
                     }
-                    normalize_domain(domain, port)
+
+                    Some(String::from(concatenate_host_and_port(
```
Dav1dde: `concatenate_host_and_port(..).into_owned()` instead of `String::from`.

Dav1dde (commented):
A bunch of docs nits and I don't know why they weren't caught by CI :(

gggritso:
@Dav1dde updated! Added/fixed docstrings and committed your `Cow` usage suggestions as-is, thanks!

Dav1dde (approved):
:shipit:

PR #3386, opened 2024-04-05
feat(spans): Copy transaction tags to segment

phacops:
For the trace explorer, filtering on user tags is important so we'll start copying them to the tags.

### DIFF
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -349,8 +349,6 @@ impl Getter for SpanData {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::BTreeMap;
-
     use crate::protocol::Measurement;
     use chrono::{TimeZone, Utc};
     use relay_base_schema::metrics::{InformationUnit, MetricUnit};
@@ -376,10 +374,6 @@ mod tests {
       "value": 9001.0,
       "unit": "byte"
     }
-  },
-  "tags": {
-    "foo": "bar",
-    "baz": 1
   }
 }"#;
         let mut measurements = Object::new();
@@ -403,14 +397,6 @@ mod tests {
             status: Annotated::new(SpanStatus::Ok),
             origin: Annotated::new("auto.http".to_owned()),
             measurements: Annotated::new(Measurements(measurements)),
-            tags: BTreeMap::from([
-                (
-                    "foo".to_string(),
-                    JsonLenientString("bar".to_string()).into(),
-                ),
-                ("baz".to_string(), JsonLenientString("1".to_string()).into()),
-            ])
-            .into(),
             ..Default::default()
         });
         assert_eq!(json, span.to_json_pretty().unwrap());
--- a/relay-event-schema/src/protocol/span.rs
+++ b/relay-event-schema/src/protocol/span.rs
@@ -6,8 +6,8 @@ use relay_protocol::{Annotated, Empty, FromValue, Getter, IntoValue, Object, Val
 
 use crate::processor::ProcessValue;
 use crate::protocol::{
-    EventId, Measurements, MetricsSummary, OperationType, OriginType, SpanId, SpanStatus, Tags,
-    Timestamp, TraceId,
+    EventId, JsonLenientString, Measurements, MetricsSummary, OperationType, OriginType, SpanId,
+    SpanStatus, Timestamp, TraceId,
 };
 
 #[derive(Clone, Debug, Default, PartialEq, Empty, FromValue, IntoValue, ProcessValue)]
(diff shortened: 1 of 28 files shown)


relay-server/src/services/store.rs
```diff
@@ -1385,8 +1388,8 @@ struct SpanKafkaMessage<'a> {
     span_id: &'a str,
     #[serde(default)]
     start_timestamp_ms: u64,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    tags: Option<&'a RawValue>,
+    #[serde(skip_deserializing, skip_serializing_if = "BTreeMap::is_empty")]
+    tags: BTreeMap<String, String>,
```
iker-barriocanal: According to [docs](https://serde.rs/field-attrs.html#skip_serializing):

> Skip this field when serializing, but not when deserializing.

Is this intentional?

relay-server/src/services/store.rs
```diff
@@ -1350,6 +1351,8 @@ struct SpanKafkaMessage<'a> {
     start_timestamp: f64,
     #[serde(rename(deserialize = "timestamp"), skip_serializing)]
     end_timestamp: f64,
+    #[serde(default, rename(deserialize = "tags"), skip_serializing)]
+    annotated_tags: Vec<(String, String)>,
```
iker-barriocanal: > `rename(deserialize = "tags")`

Does this field conflict with `tags` below? Which of the two is produced to Kafka?

relay-event-schema/src/protocol/span.rs
```diff
@@ -59,7 +59,7 @@ pub struct Span {
 
     /// Arbitrary tags on a span, like on the top-level event.
     #[metastructure(pii = "maybe")]
-    pub tags: Annotated<Object<JsonLenientString>>,
+    pub tags: Annotated<Tags>,
```
iker-barriocanal: The data structure changes from a map to an array. Is this change supported downstream (e.g. consumers)?

iker-barriocanal: Also -- should we skip serialization if there are no tags? `skip_serialization = "empty"`.

tests/integration/test_spans.py
```diff
@@ -1287,3 +1287,89 @@ def summarize_outcomes():
...
+            "sdk.version": "2.6.3",
+            "transaction": "hi",
+            "transaction.op": "hi",
+        },
+        "span_id": "968cff94913ebb07",
+        "start_timestamp_ms": int(
+            start_timestamp.replace(tzinfo=timezone.utc).timestamp() * 1e3
+        ),
+        "trace_id": "a0fa8803753e40fd8124b21eeb2986b5",
+        "tags": expected_tags,
```
iker-barriocanal: The test seems to only validate the tags. We should only assert that field and ignore those that aren't important for the test.

iker-barriocanal (commented):
The code LGTM, please have a look at the comments below.

This PR introduces the Map -> Array, which implicitly also changes the `LenientString`. Existing tests, including the kafka schema checks, pass; so I think it should be good. However, I'm a bit worried this could break something so I'd suggest deploying in S4S first.

relay-server/src/services/store.rs
phacops: You linked the documentation to `skip_serializing`, which I'm not using.

My intention is to not deserialize this field at all and only serialize it if it's not empty, and that's what this annotation will do I believe.

relay-server/src/services/store.rs
phacops: This is what I want to deserialize as `tags` indeed, and that's why I added the `skip_deserializing` above.

The annotated protocol represents the `Tags` struct as pairs of strings and that's why I need to deserialize like this and then repack it as a map. It's a little bit of mental gymnastic to set the serde annotations correctly but I think they're right.

relay-server/src/services/store.rs
iker-barriocanal: Oh right, misread :D

relay-event-schema/src/protocol/span.rs
phacops: This field is not set by current SDKs. It was added for a span-only world where we'd set tags directly on spans. Now, this is not propagated directly downstream as it's transformed to the `SpanKafkaMessage` first I modified in this PR. So, yes, the changes needed to support this type change are done.

@jjbayer do you know of any other use of that field outside of standalone spans?

We can indeed skip the serialization if it's empty.

jjbayer: `span.tags` has been around for ~4 years and as far as I can remember it is used by some SDKs.

I pushed a change that solves the problem by making `map_fields!` call `.into()` on everything, plus some missing `From` and `Into` implementations. Let me know what you think!

iker-barriocanal: Looks great, thank you!

iker-barriocanal (approved):
LGTM! Please update the scope of the integration test before merging

PR #3387, opened 2024-04-07
feat(spans): Route spans according to trace_id

phacops:
We're in need of buffering spans later in the pipeline and it helps to have them on the same partition. We only need them routed per `segment_id` but since it's not a `Uuid`, using `trace_id` is simpler. An alternative would be to properly transform the `segment_id` into a `u64` and then call `Uuid::from_u64_pair(segment_id, 0)` but that seems unnecessary.

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -1031,7 +1031,7 @@ impl StoreService {
                         project_id,
                         received,
                         retention_days,
-                        segment_id,
+                        segment_id: segment_id.unwrap_or_default(),
                         span_id,
                         sum,
                         tags,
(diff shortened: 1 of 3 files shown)


relay-server/src/services/store.rs
```diff
@@ -1378,16 +1378,15 @@ struct SpanKafkaMessage<'a> {
     /// Number of days until these data should be deleted.
     #[serde(default)]
     retention_days: u16,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    segment_id: Option<&'a str>,
+    segment_id: &'a str,
```
jjbayer: We deserialize a `SpanKafkaMessage` from a `Span`, right? Can we be sure that it has a `segment_id`?

phacops: Reverted since we're using `trace_id` and not `segment_id` for routing.

jjbayer (approved):

iker-barriocanal (approved):
Thanks for the PR description!

PR #3389, opened 2024-04-08
feat(metric-stats): Add cardinality limited outcome id

Dav1dde:
Adds the new outcome id `6` `cardinality_limited` to the outcomes enum.

The outcome will already be emitted for the `METRIC_BUCKET` data category with this change, but the main purpose is to use the outcome for [metric stats](https://github.com/getsentry/relay/issues/3147).

Systems checked for support:
- [X] Snuba
- [x] Getsentry
- [x] Sentry


jjbayer (commented):
Does this need to be  a separate outcome? Could it be just a different `reason` instead?

Dav1dde:
That's what we have already and it doesn't fit very well.

- `filtered`: from what I gathered is more of a customer opted into some filtering.
- `rate_limited`: what we're using right now but this is mainly used for quotas and rate limits, if we emit a rate limit now which is independent of volume there is a good chance we're actually messing something up in the quota code. It also means we have to pack even more information into the reason, especially if we want to include the cardinality limit which was breached (similar to rate limits) into the code.
- `invalid`: nothing invalid about this
- `abuse`: also not an abuse case
- `client_discard`: wasn't the client

Overall I think a separate category just leaves less room for errors and misinterpretations long term, with a lot more flexibility in the reason.

jjbayer (approved):

iker-barriocanal (approved):
> The outcome will already be emitted for the METRIC_BUCKET data category with this change

I assume the new ID is supported in sentry.

PR #3390, opened 2024-04-08
ref(metrics): Validate metric units

jan-auer:
Validates that metric units consist of ASCII word characters.
Previously, more lenient validation was applied. Since Relay is now the
only place to normalize MRIs, we're enforcing the rules more strictly.

Note that the spec for what's considered a valid unit has not changed.

In a future change, Relay will start to emit errors or meta data for
invalid MRIs and tags so that users receive some feedback and can fix
potential issues.

Epic: https://github.com/getsentry/team-ingest/issues/304
#skip-changelog

### DIFF
--- a/relay-base-schema/src/metrics/units.rs
+++ b/relay-base-schema/src/metrics/units.rs
@@ -17,7 +17,7 @@ pub enum MetricUnit {
     Information(InformationUnit),
     /// Fractions such as percentages, defaulting to `"ratio"`.
     Fraction(FractionUnit),
-    /// user-defined units without builtin conversion or default.
+    /// User-defined units without built-in conversion or default.
     Custom(CustomUnit),
     /// Untyped value without a unit (`""`).
     #[default]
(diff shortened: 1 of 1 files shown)


Dav1dde (commented):
Should we "normalize" the units like we do metric names and replace all non-matching characters with an `_`?

jan-auer:
I'm torn about this. For tags we ended up deciding that we won't normalize since this could lead to unintentionally overwriting tags etc. For units, I think this is less of a concern, but it can still have severe effects in the product since the MRI changes. So we would have to be _extremely_ careful with any change to that normalization logic in the future. 

For the time being, it's probably easiest to drop the metrics, which means we can become more lenient in the future.

Dav1dde:
> For the time being, it's probably easiest to drop the metrics, which means we can become more lenient in the future.

The problem with that is, there is absolutely 0 feedback that a metric has been dropped because of an invalid unit.

Dav1dde (approved):

jan-auer:
Yes, this is something I should've made clear earlier. This is all based on the premise that we'll be adding proper error messages across the board. cc @jjbayer

PR #3394, opened 2024-04-08
feat(spans): Sample extraction

jjbayer:
Add a global option to sample span extraction, such that the feature can be ramped up slowly even in deployments with only a few dominant projects (e.g. S4S).

#skip-changelog

### DIFF
--- a/relay-dynamic-config/src/global.rs
+++ b/relay-dynamic-config/src/global.rs
@@ -166,10 +166,9 @@ pub struct Options {
     #[serde(
         rename = "relay.span-extraction.sample-rate",
         deserialize_with = "default_on_error",
-        default = "one",
-        skip_serializing_if = "is_one"
+        skip_serializing_if = "is_default"
     )]
-    pub span_extraction_sample_rate: f32,
+    pub span_extraction_sample_rate: Option<f32>,
 
     /// All other unknown options.
     #[serde(flatten)]
@@ -300,14 +299,6 @@ pub enum BucketEncoding {
 /// Returns `true` if this value is equal to `Default::default()`.
 fn is_default<T: Default + PartialEq>(t: &T) -> bool {
     *t == T::default()
-}
-
-fn one() -> f32 {
-    1.0
-}
-
-fn is_one(value: &f32) -> bool {
-    *value == 1.0
 }
 
 fn default_on_error<'de, D, T>(deserializer: D) -> Result<T, D::Error>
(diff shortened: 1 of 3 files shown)


mwarkentin:
@jjbayer thanks! Let me know if there's anything you need from the SRE side for this.

relay-dynamic-config/src/global.rs
```diff
@@ -156,6 +156,20 @@ pub struct Options {
     )]
     pub metric_stats_rollout_rate: f32,
 
+    /// Overall sampling of span extraction.
+    ///
+    /// This number represents the fraction of transactions for which
+    /// spans are extracted. It applies on top of [`crate::Feature::ExtractSpansAndSpanMetricsFromEvent`],
+    /// so both feature flag and sample rate need to be enabled to get any spans extracted.
+    ///
+    /// Any value below 1.0 will cause the product to break, so use with caution.
```
olksdr:
```suggestion
    /// Note: Any value below 1.0 will cause the product to break, so use with caution.
```

In my opinion it's a little bit more visible, harder to miss

olksdr (approved):

relay-dynamic-config/src/global.rs
```diff
@@ -156,6 +156,20 @@ pub struct Options {
     )]
     pub metric_stats_rollout_rate: f32,
 
+    /// Overall sampling of span extraction.
+    ///
+    /// This number represents the fraction of transactions for which
+    /// spans are extracted. It applies on top of [`crate::Feature::ExtractSpansAndSpanMetricsFromEvent`],
+    /// so both feature flag and sample rate need to be enabled to get any spans extracted.
+    ///
+    /// Note: Any value below 1.0 will cause the product to break, so use with caution.
```
iker-barriocanal, line 167: We should mention that if the feature is not present, Relay assumes `1.0`.

iker-barriocanal (approved):

relay-dynamic-config/src/global.rs
```diff
@@ -156,6 +156,20 @@ pub struct Options {
     )]
     pub metric_stats_rollout_rate: f32,
 
+    /// Overall sampling of span extraction.
+    ///
+    /// This number represents the fraction of transactions for which
+    /// spans are extracted. It applies on top of [`crate::Feature::ExtractSpansAndSpanMetricsFromEvent`],
+    /// so both feature flag and sample rate need to be enabled to get any spans extracted.
+    ///
```
jjbayer, line 164:
```suggestion
    ///
    /// `None` is the default and interpreted as a value of 1.0 (extract everything).
    ///
```

Issue #3317, opened 2024-03-25
Apply inbound filters to spans (using abstraction layer)


jjbayer:
```[tasklist]
### Tasks
- [ ] Implement trait introduced in https://github.com/getsentry/relay/pull/3397 for type `Span`.
- [ ] https://github.com/getsentry/relay/issues/3550
- [ ] https://github.com/getsentry/relay/issues/3551
```

jjbayer:
Set Apr 9 as due date, because Replays depend on the abstraction layer as well.

jjbayer:
New due date is Apr 30, because Replays are already unbocked by https://github.com/getsentry/relay/pull/3397.

PR #3397, opened 2024-04-09
ref(filters): Access fields through trait

jjbayer:
Currently, inbound filters only work for event types, that is error events and transactions. In the near future we will want inbound filters to apply to other data types as well, such as standalone spans and session replays. To prepare for this generalization, make filters act on a trait that supplies the necessary field instead of the `Event` itself:

```rust
// `Filterable` for legacy inbound filters,
// `Getter` for generic inbound filters.
pub fn should_filter<F: Filterable + Getter>(item: &F, ...); 
```

ref: #3317 

#skip-changelog


relay-filter/src/transaction_name.rs
```diff
@@ -3,44 +3,37 @@
...
 use relay_common::glob3::GlobPatterns;
-use relay_event_schema::protocol::{Event, EventType};
 
-use crate::{FilterStatKey, IgnoreTransactionsFilterConfig};
+use crate::{FilterStatKey, Filterable, IgnoreTransactionsFilterConfig};
 
-fn matches(event: &Event, patterns: &GlobPatterns) -> bool {
-    if event.ty.value() != Some(&EventType::Transaction) {
-        return false;
-    }
```
jjbayer, line 13: This check moved to the trait implementation.

relay-filter/src/interface.rs
```diff
@@ -0,0 +1,74 @@
+//! This module contains the trait for items that can be filtered by Inbound Filters, plus
+//! the implementation for [`Event`].
+use url::Url;
+
+use relay_event_schema::protocol::{Csp, Event, EventType, Exception, LogEntry, Values};
+
+/// A data item to which filters can be applied.
+pub trait Filterable {
```
jjbayer, line 8: I'm open for ideas on how to name this trait.

relay-filter/src/legacy_browsers.rs
```diff
@@ -2,82 +2,71 @@
...
 use relay_ua::UserAgent;
 
-use crate::{FilterStatKey, LegacyBrowser, LegacyBrowsersFilterConfig};
+use crate::{FilterStatKey, Filterable, LegacyBrowser, LegacyBrowsersFilterConfig};
 
 /// Checks if the event originates from legacy browsers.
-pub fn matches(event: &Event, browsers: &BTreeSet<LegacyBrowser>) -> bool {
-    if let Some(user_agent_string) = event.user_agent() {
-        let user_agent = relay_ua::parse_user_agent(user_agent_string);
+fn matches(user_agent: &str, browsers: &BTreeSet<LegacyBrowser>) -> bool {
```
jjbayer, line 10: I made the `matches` functions private everywhere, they do not need to be public.

relay-filter/src/interface.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// The logentry message. Only for error events.
+    fn logentry(&self) -> Option<&LogEntry>;
+
+    /// The release string of the data item.
+    fn release(&self) -> Option<&str>;
+
+    /// The transaction name. Only for transaction events.
+    fn transaction(&self) -> Option<&str>;
+
+    /// The URL from which the data originates. Used for localhost filtering.
```
iker-barriocanal:
```suggestion
    /// The URL from which the request originates. Used for localhost filtering.
```
?

relay-filter/src/interface.rs
```diff
@@ -0,0 +1,74 @@
...
+        self.csp.value()
+    }
+
+    fn exceptions(&self) -> Option<&Values<Exception>> {
+        self.exceptions.value()
+    }
+
+    fn ip_addr(&self) -> Option<&str> {
+        let user = self.user.value()?;
+        Some(user.ip_address.value()?.as_ref())
```
iker-barriocanal, line 48: There's a getter implementation for this, and we already import the trait in the crate, so this should be possible?
```suggestion
        get_value!(self.user.ip_address!).as_ref()
```

relay-filter/src/localhost.rs
```diff
@@ -34,29 +32,22 @@ pub fn matches(event: &Event) -> bool {
 }
 
 /// Filters events originating from the local host.
-pub fn should_filter(event: &Event, config: &FilterConfig) -> Result<(), FilterStatKey> {
+pub fn should_filter<F>(item: &F, config: &FilterConfig) -> Result<(), FilterStatKey>
+where
+    F: Filterable,
+{
```
iker-barriocanal: nit: should we inline the `Filterable` impl, since there's only one?
```suggestion
pub fn should_filter<F: Filterable>(item: &F, config: &FilterConfig) -> Result<(), FilterStatKey> {
```

iker-barriocanal (approved):
nit: we should align all `matches` implementations for consistency -- sometimes they take `item` (convenience) and some others just the value to filter on.

cmanallen (approved):
Perfect.  This will be easy to implement for replays.

relay-filter/src/interface.rs
jjbayer: Good idea, but `get_value!` only works if the root object is `Annotated` AFAIK. We could change that, but it seems out of scope. Side note: `ip_address!` unwraps the value, the exclamation mark in `get_value!` paths should only be used in tests.

relay-filter/src/localhost.rs
jjbayer: 👍 I inlined it everywhere now, except where it would break the signature into multiple lines.

jjbayer (commented):
> nit: we should align all matches implementations for consistency

@iker-barriocanal They are all private functions, so IMO they should do what makes the most sense in local context. I pass the value directly where the function needs only that value, and the full item in cases where the function needs more than one value. Let me know if that makes sense.

PR #3399, opened 2024-04-09
ref(server): Put processor `Addr`s in separate struct

jjbayer:
Non-functional change to get rid of `#[allow(clippy::too_many_arguments)]` on `EnvelopeProcessorService::new`. This PR was originally part of https://github.com/getsentry/relay/pull/3375, which adds yet another `Addr`, but I decided to make a separate PR for reviewability.

#skip-changelog

### DIFF
--- a/relay-server/src/service.rs
+++ b/relay-server/src/service.rs
@@ -182,6 +182,7 @@ impl ServiceState {
                 #[cfg(feature = "processing")]
                 store_forwarder: store.clone(),
             },
+            #[cfg(feature = "processing")]
             metric_stats,
         )
         .spawn_handler(processor_rx);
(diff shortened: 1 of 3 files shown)


iker-barriocanal (approved):

Dav1dde:
Should we not just pass the entire [`Registry`](https://github.com/getsentry/relay/blob/master/relay-server/src/service.rs#L49-L61) instead?

jjbayer:
> Should we not just pass the entire [`Registry`](https://github.com/getsentry/relay/blob/master/relay-server/src/service.rs#L49-L61) instead?

I considered it, but IMO it still makes sense to give the service only the addresses that it needs.

PR #3400, opened 2024-04-09
fix(spans): Also sample metrics when sample rate is < 1.0

jjbayer:
Follow-up to #3394: Span metrics are extracted in a different code path than span payloads, so apply the same sampling rate to metrics extraction.

#skip-changelog


olksdr (approved):
lgtm

PR #3403, opened 2024-04-09
feat(replay): separate feedback from attachments in the same envelope

aliu39:
The idea is to restructure the code so that UserReportV2's don't end up in the catch-all extract_kafka_messages at the bottom of `store_envelope`, with attachments

Relates to https://github.com/getsentry/team-replay/issues/393


relay-server/src/services/store.rs
```diff
@@ -244,6 +231,23 @@ impl StoreService {
...
+                    let remote_addr = envelope.meta().client_addr().map(|addr| addr.to_string());
+
+                    self.produce_user_report_v2(
+                        event_id.ok_or(StoreError::NoEventId)?,
+                        scoping.organization_id,
+                        scoping.project_id,
+                        start_time,
+                        item,
+                        remote_addr,
+                        is_rolled_out(scoping.organization_id, feedback_ingest_topic_rollout_rate),
```
Dav1dde: It's better if you pass in the topic instead or even better, let the produce function figure out the topic.

relay-server/src/services/store.rs
```diff
@@ -1245,6 +1277,24 @@ struct UserReportKafkaMessage {
...
+    start_time: u64,
+    /// The event id.
+    event_id: EventId,
+    /// The project id for the current event.
+    project_id: ProjectId,
+    /// The client ip address.
+    remote_addr: Option<String>, //TODO: is this necessary or expected by anything downstream?
+    /// Attachments that are potentially relevant for processing.
+    attachments: Vec<ChunkedAttachment>, //TODO: is this necessary?
+}
```
Dav1dde: Unused with a bunch of todos?

relay-server/src/services/store.rs
```diff
@@ -661,6 +665,34 @@ impl StoreService {
...
+        &self,
+        event_id: EventId,
+        organization_id: u64,
+        project_id: ProjectId,
+        start_time: Instant,
+        item: &Item,
+        remote_addr: Option<String>,
+        use_feedback_topic: bool,
+    ) -> Result<(), StoreError> {
+        // TODO: now that feedback is on its own topic, make it its own type of KafkaMessage
```
Dav1dde: That's something which is better tracked in a ticket than code.

Dav1dde (commented):
Please see if you can add some tests or modify existing tests (if there are any, why didn't they break with that change?).

Also before moving on with this to SaaS it'd be best if you figured out the wire format for the Feedback topic, it's better to avoid supporting multiple formats on the other "end".

aliu39:
> Also before moving on with this to SaaS it'd be best if you figured out the wire format for the Feedback topic, it's better to avoid supporting multiple formats on the other "end".

Thanks for the review! Could you elaborate what you mean by "wire format"? Do you mean the `UserReportV2KafkaMessage` type?

You're right, I'll make a ticket for adding that message type and remove it from this PR, it's something I forgot to clean that up. After adding coverage I could deploy this with a FF enabled for s4s only.

Dav1dde:
> Thanks for the review! Could you elaborate what you mean by "wire format"? Do you mean the UserReportV2KafkaMessage type?

Yes, you're using the `Event` right now and it looks like you want to move to `UserReportV2KafkaMessage`, it's better if you do it early, it's gonna be a lot harder to do later if you have to support both versions (we can't instantly switch Relays to a new version so for a certain period of time all downstream consumers have to support both formats).

aliu39:
> Yes, you're using the `Event` right now and it looks like you want to move to `UserReportV2KafkaMessage`, it's better if you do it early, it's gonna be a lot harder to do later if you have to support both versions (we can't instantly switch Relays to a new version so for a certain period of time all downstream consumers have to support both formats).

Got it, I'll definitely look into it and give it more thought before finalizing this PR. Thanks for the suggestion!

aliu3ntry:
Removing commented code in a sec!

aliu39:
> Also before moving on with this to SaaS it'd be best if you figured out the wire format for the Feedback topic, it's better to avoid supporting multiple formats on the other "end".

I've decided to hold off on making a UserReportV2KafkaMessage type until the new topic/infra has been deployed to all regions in prod. It adds an extra layer of complexity and testing that will delay this feature, which the SDK team needs to GA user feedback in ~1 week.

It probably shouldn't be done until we write a separate ingest strategy for the feedback consumer (right now it shares this with events).

relay-server/src/services/store.rs
```diff
@@ -236,6 +238,7 @@ impl StoreService {
                     attachments.push(attachment);
                 }
                 ItemType::UserReport => {
+                    // Crash report modal and older SDKs and feedback endpoints
```
JoshFerge: i think these comments could be better placed in the ItemType definition?

relay-server/src/services/store.rs
```diff
@@ -189,13 +189,15 @@ impl StoreService {
         let retention = envelope.retention();
         let event_id = envelope.event_id();
 
+        let use_feedback_ingest_v2 = self.global_config.current().options.feedback_ingest_v2;
```
JoshFerge: not sure what this naming implies, maybe just `separate_attachments_from_feedback`?

JoshFerge (commented):
other than the small nits / naming, lgtm

relay-server/src/services/store.rs
```diff
@@ -673,6 +689,36 @@ impl StoreService {
...
+    fn produce_user_report_v2(
+        &self,
+        event_id: EventId,
+        project_id: ProjectId,
+        organization_id: u64,
+        start_time: Instant,
+        item: &Item,
+        remote_addr: Option<String>,
+    ) -> Result<(), StoreError> {
+        // check rollout rate option (effectively a FF) to determine whether to produce to new infra
```
aliu39, line 702: See https://github.com/getsentry/relay/pull/3344

relay-server/src/services/store.rs
aliu39: Decided to name it this because it also results in some code restructuring / different code path for producing feedback in StoreService.

I agree the naming's a little funky though, will sleep on it + see what Vienna thinks?

jjbayer: I would also prefer aligning the variable name (and the field name `feedback_ingest_v2`) with either the feature name ("inline attachments") or renaming it to describe what it actually enables.

relay-server/src/services/store.rs
```diff
@@ -244,6 +246,20 @@ impl StoreService {
                         item,
                     )?;
                 }
+                ItemType::UserReportV2 => {
+                    if use_feedback_ingest_v2 {
```
jjbayer: With this, you can make sure that there's an `error` log if there's an unexpected feedback item in the envelope.

```suggestion
                ItemType::UserReportV2 if use_feedback_ingest_v2 => {
```

relay-server/src/services/store.rs
```diff
@@ -673,6 +689,36 @@ impl StoreService {
...
+    ) -> Result<(), StoreError> {
+        // check rollout rate option (effectively a FF) to determine whether to produce to new infra
+        let global_config = self.global_config.current();
+        let feedback_ingest_topic_rollout_rate =
+            global_config.options.feedback_ingest_topic_rollout_rate;
+        let topic = if is_rolled_out(organization_id, feedback_ingest_topic_rollout_rate) {
+            KafkaTopic::Feedback
+        } else {
+            KafkaTopic::Events
+        };
```
jjbayer, line 710: Out of scope for this PR, but should we hard-code this to `Feedback` soon? Or are some installations not ready for that yet? (self-hosted, devenv, ...)

tests/.idea/workspace.xml
jjbayer, line 1: Did you check this in on purpose?

tests/integration/test_feedback.py
```diff
@@ -139,3 +155,89 @@ def test_feedback_events_without_processing(
...
+    mini_sentry.set_global_config_option(
+        "feedback.ingest-topic.rollout-rate", 1.0 if use_feedback_topic else 0.0
+    )
+    # Test will only pass with this option set
+    mini_sentry.set_global_config_option("feedback.ingest-inline-attachments", True)
+
+    if use_feedback_topic:
+        mini_sentry.set_global_config_option("feedback.ingest-topic.rollout-rate", 1.0)
+        feedback_consumer_ = feedback_consumer(timeout=20)
+        other_consumer_ = events_consumer(timeout=20)
```
iker-barriocanal: Why the `_` suffix?

relay-server/src/services/store.rs
iker-barriocanal: The flag's docs say it's temporary. Is it only used for testing in development or should we update the docs?

aliu39: It's used like a feature flag - we'll set it to true in test region(s), test the SDK, and remove this logic + non-v2 code path once satisfied. The flag is just to quickly "revert" in case things start to break

aliu39: What did you have in mind for updating the docs?

relay-server/src/services/store.rs
aliu39: Yes this has to do with ops deployment of the feedback topic -- right now, its only in s4s and not other regions. This check and rollout rate will eventually be removed too

tests/.idea/workspace.xml
aliu39: No 😅

iker-barriocanal, line 1: Please, remove this file.

relay-dynamic-config/src/global.rs
```diff
@@ -167,6 +167,18 @@ pub struct Options {
     )]
     pub feedback_ingest_topic_rollout_rate: f32,
 
+    /// Flag for handling feedback and attachments in the same envelope. Temporary FF for fast-revert
+    /// (will remove after user feedback GA release).
```
iker-barriocanal, line 171: Without context, this comment is not clear to me what the flag should do. What do you think about something like the following?

```suggestion
    /// Kill switch to stop producing feedback items separately into Kafka. The topic is controlled by `feedback_ingest_topic_rollout_rate`.
```

tests/integration/test_feedback.py
```diff
@@ -139,3 +155,89 @@ def test_feedback_events_without_processing(
...
+    events_consumer,
+    attachments_consumer,
+    use_feedback_topic,
+):
+    mini_sentry.add_basic_project_config(
+        42, extra={"config": {"features": ["organizations:user-feedback-ingest"]}}
+    )
+    mini_sentry.set_global_config_option(
+        "feedback.ingest-topic.rollout-rate", 1.0 if use_feedback_topic else 0.0
+    )
```
iker-barriocanal, line 174: This option is overwritten on either line 179 or 183, let's remove this one.

tests/integration/test_feedback.py
```diff
@@ -139,3 +155,89 @@ def test_feedback_events_without_processing(
...
+
+    if use_feedback_topic:
+        mini_sentry.set_global_config_option("feedback.ingest-topic.rollout-rate", 1.0)
+        feedback_consumer_ = feedback_consumer(timeout=20)
+        other_consumer = events_consumer(timeout=20)
+    else:
+        mini_sentry.set_global_config_option("feedback.ingest-topic.rollout-rate", 0.0)
+        feedback_consumer_ = events_consumer(timeout=20)
+        other_consumer = feedback_consumer(timeout=20)
+    attachments_consumer_ = attachments_consumer(timeout=20)
```
iker-barriocanal, line 186: Unless there's a reason to have it:
```suggestion
        feedback_consumer = events_consumer(timeout=20)
        other_consumer = feedback_consumer(timeout=20)
    attachments_consumer = attachments_consumer(timeout=20)
```
(same with the `if` block)

iker-barriocanal (approved):
LGTM! Please see comments below before merging.

iker-barriocanal:
@aliu3ntry please address the suggestions above, I didn't realize the PR had auto-merging enabled.

tests/integration/test_feedback.py
aliu3ntry: Sounds good, I didn't want to shadow/override the fixtures with the same name, but should be fine for this test!

aliu39: Sounds good, I didn't want to shadow/override the fixtures with the same name, but can make this change!

tests/integration/test_feedback.py
aliu39: good catch!

aliu39:
Oops, sorry! Reopened at https://github.com/getsentry/relay/pull/3486 @iker-barriocanal

PR #3405, opened 2024-04-10
feat(clock-drift): Apply clock drift normalization only if `sent_at` is defined

### LINKED ISSUE #1625: Clock drift correction is applied to envelopes without sent_at
### Environment

SaaS (https://sentry.io/)

### Version

_No response_

### Steps to Reproduce

set your device time to may 30,  and send a transaction and an event. You'll notice that events generated by  the device will have a similar warning
image
Where transactions will have a different error on the UI
image

### Expected Result

Transaction not having the time adjusted.

### Actual Result

Turns out the Backend adjust the timestamp from the transactions and its breadcrumbs,  resulting in events that can be separated by a huge gap of time on Discover (Example below)
https://sentry.io/organizations/sentry-sdks/discover/results/?end=2022-06-03T00%3A23%3A43&field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=5522756&query=id%3A192b99ee99334f73a97950a71f7e4161++or+id%3A8c645ff84eca4ccd9d47b1eff49fde02&sort=-timestamp&start=2022-05-30T17%3A20%3A15&userModified=true&yAxis=count%28%29
image

I believe this time adjustment may be erroneous, it makes it harder to understand on discover when filtering by the user,
=>https://sentry.io/organizations/sentry-sdks/discover/results/?end=2022-06-03T00%3A23%3A43.000&field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=5522756&query=user.display%3A0bf59fdf-6052-4aa1-84cd-adb9d2b49487+timestamp%3A%3C2022-06-02T18%3A12%3A46%2B00%3A00++timestamp%3A%3E2022-05-30T22%3A54%3A48%2B00%3A00&sort=-timestamp&start=2022-05-30T17%3A20%3A15.000&userModified=true&yAxis=count%28%29

Additionally, it may lead to folks thinking something is slow today where in reality, the slowness actually happened days ago.

iambriccardo:
This PR applies clock drift normalization only when `sent_at` is defined in the envelope headers.

The previous implementation was falling back to `event.timestamp` if the incoming event was a transaction. That was expected behavior for old SDKs which didn't send `sent_at` however we do not have anymore SDKs using the `/store` endpoint, which means that we can assume that `sent_at` is sent for each envelope. If not sent, we won't perform clock drift normalization instead of using `event.timestamp` like before.

Closes: https://github.com/getsentry/relay/issues/1625

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ## Unreleased
 
+**Bug fixes:**
+
+- Fix performance regression in disk spooling by using page counts to estimate the spool size. ([#3379](https://github.com/getsentry/relay/pull/3379))
+- Perform clock drift normalization only when `sent_at` is set in the `Envelope` headers. ([#3405](https://github.com/getsentry/relay/pull/3405))
+
 **Features**:
+
 - **Breaking change:** Kafka topic configuration keys now support the default topic name. The previous aliases `metrics` and `metrics_transactions` are no longer supported if configuring topics manually. Use `ingest-metrics` or `metrics_sessions` instead of `metrics`, and `ingest-performance-metrics` or `metrics_generic` instead of `metrics_transactions`. ([#3361](https://github.com/getsentry/relay/pull/3361))
 - Add support for continuous profiling. ([#3270](https://github.com/getsentry/relay/pull/3270))
 - Add support for Reporting API for CSP reports ([#3277](https://github.com/getsentry/relay/pull/3277))
@@ -16,12 +22,6 @@
 - Support the full unicode character set via UTF-8 encoding for metric tags submitted via the statsd format. Certain restricted characters require escape sequences, see [docs](https://develop.sentry.dev/sdk/metrics/#normalization) for the precise rules. ([#3358](https://github.com/getsentry/relay/pull/3358))
 - Stop extracting count_per_segment and count_per_op metrics. ([#3380](https://github.com/getsentry/relay/pull/3380))
 - Extract `cache.item_size` and `cache.hit` metrics. ([#3371]https://github.com/getsentry/relay/pull/3371)
-
-**Bug Fixes**:
-
-- Fix performance regression in disk spooling by using page counts to estimate the spool size. ([#3379](https://github.com/getsentry/relay/pull/3379))
-- Allow spans with no `exclusive_time`. ([#3385](https://github.com/getsentry/relay/pull/3385))
-- Perform clock drift normalization only when `sent_at` is set in the `Envelope` headers. ([#3405](https://github.com/getsentry/relay/pull/3405))
 
 **Internal**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## Unreleased
 
-**Bug fixes:**
-
-- Fix performance regression in disk spooling by using page counts to estimate the spool size. ([#3379](https://github.com/getsentry/relay/pull/3379))
-
 **Features**:
 
 - Add support for continuous profiling. ([#3270](https://github.com/getsentry/relay/pull/3270))
@@ -23,7 +19,9 @@
 
 **Bug Fixes**:
 
+- Fix performance regression in disk spooling by using page counts to estimate the spool size. ([#3379](https://github.com/getsentry/relay/pull/3379))
 - Allow spans with no `exclusive_time`. ([#3385](https://github.com/getsentry/relay/pull/3385))
+- Perform clock drift normalization only when `sent_at` is set in the `Envelope` headers. ([#3405](https://github.com/getsentry/relay/pull/3405))
 
 **Internal**:
 
--- a/tests/integration/test_clock_drift.py
+++ b/tests/integration/test_clock_drift.py
@@ -1,0 +1,85 @@
+from datetime import timedelta, datetime
+
+from sentry_sdk.envelope import Envelope
+
+
+def mock_transaction(timestamp):
+    return {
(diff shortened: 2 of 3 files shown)


relay-event-normalization/src/validation.rs
```diff
@@ -308,30 +314,30 @@ fn normalize_timestamps(
 ) {
     let received_at = received_at.unwrap_or_else(Utc::now);
 
-    let mut sent_at = None;
+    let mut event_timestamp = None;
```
iambriccardo: Renamed because I didn't want confusion with the new `has_sent_at` property.

iker-barriocanal:
>That was expected behavior for old SDKs but now it should not be needed anymore.

What are the SDK versions where this is fixed? Do we have a big event influx from these outdated versions? Bonus points if this is documented in develop docs!

Dav1dde:
```rs
    // TODO: Temporary workaround before processing. Experimental SDKs relied on a buggy
    // clock drift correction that assumes the event timestamp is the sent_at time. This
    // should be removed as soon as legacy ingestion has been removed.
```

Curious what this has to do with the legacy ingestion (I assume directly to Sentry).

iambriccardo:
> ```rust
>     // TODO: Temporary workaround before processing. Experimental SDKs relied on a buggy
>     // clock drift correction that assumes the event timestamp is the sent_at time. This
>     // should be removed as soon as legacy ingestion has been removed.
> ```
> 
> Curious what this has to do with the legacy ingestion (I assume directly to Sentry).

@jan-auer mentioned this in one of our private discussions. From what I understood, the `store` endpoint is used by old SDKs that send envelopes without the `sent_at` header assuming that the event `timestamp` is the same as the time the envelope is sent. New SDKs, are sending envelopes with the `sent_at` which differs from event `timestamp`.

CHANGELOG.md
```diff
@@ -23,7 +19,9 @@
 
 **Bug Fixes**:
 
+- Fix performance regression in disk spooling by using page counts to estimate the spool size. ([#3379](https://github.com/getsentry/relay/pull/3379))
 - Allow spans with no `exclusive_time`. ([#3385](https://github.com/getsentry/relay/pull/3385))
+- Perform clock drift normalization only when `sent_at` is set in the `Envelope` headers. ([#3405](https://github.com/getsentry/relay/pull/3405))
```
jjbayer: Please move Bug Fixes to above Features, it's the first thing people should read IMO.

tests/integration/test_clock_drift.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+    envelope.add_transaction(mock_transaction(one_month_ago))
+
+    relay.send_envelope(project_id, envelope, headers={"Accept-Encoding": "gzip"})
+
+    transaction_event = mini_sentry.captured_events.get(
+        timeout=1
+    ).get_transaction_event()
+    error_name, error_metadata = transaction_event["_meta"]["timestamp"][""]["err"][0]
+    assert error_name == "clock_drift"
```
jjbayer, line 45: Should we also assert on the actual timestamps in the event? (Same for tests below).

CHANGELOG.md
iambriccardo: Yep, will do.

jjbayer:
@iambriccardo in the PR description, could you elaborate a bit more on why this is now OK? (Transactions no longer sent to `/store` endpoint...).

tests/integration/test_clock_drift.py
iambriccardo: I was considering it, my only problem was that `received_at` seems to be inferred by Relay on receive time, so I might have to either mock things or freeze time but it seems like this is not possible in an integration test.

iambriccardo: What I know is that `sdk_time` will be `one_month_ago` and `server_time` will be `received_at` and the new timestamp is + the difference of the above.

iambriccardo: I opted for a middle ground, I do test for `>` in the timestamps which should be normalized, and `=` when the timestamps must remain the same.

Dav1dde:
Re store endpoint, but these old (broken?) SDKs are still sending data, why can we change it now?

jan-auer:
This was used just for beta SDKs while Performance monitoring was experimental. As we launched the first stable version of Performance Monitoring to general availability, we required an SDK upgrade. Hence this is safe to remove.

Dav1dde:
Nice, thanks!

tests/integration/test_clock_drift.py
Dav1dde: The assert could be a much smaller window though, right? `transaction_event["timestamp"] > now.timestamp()`

Dav1dde (approved):

tests/integration/test_clock_drift.py
iambriccardo: If there was a way to freeze or hardcode the `received_at` time, I could compute the exact test expectations. 

If we can't, to make the test more precise, we could assert `transaction_event["timestamp"] == now` within some bounds but I am not super confident to define them since they are dependent on the time it takes for the integration test to run and I would not want to introduce a flaky test.

Dav1dde: I wouldn't exactly assert, but if you compare to the `now` value taken at the beginning of the test, that should be good, additionally you could also take another "now" value at the end and check whether the expected value is between those. I wouldn't try to freeze time etc. that will just break at some point. Being within a few seconds should be good enough.

PR #3407, opened 2024-04-10
feat(spans): Extract trace context status for indexed spans

edwardgou-sentry:
Extract trace context `status` for indexed spans. To be used in the queues module so we can see the status of tasks.

#skip-changelog


relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_mobile.snap
```diff
@@ -39,6 +39,7 @@ expression: "(&event.value().unwrap().spans, metrics)"
                 "release": "1.2.3",
                 "sdk.name": "sentry.javascript.react-native",
                 "sdk.version": "unknown",
+                "trace.status": "unknown",
```
Dav1dde, line 42: Where does this `unknown` come from and is that something we want?

jjbayer (approved):

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_mobile.snap
iker-barriocanal: This ^. The code LGTM, but according to [dev docs](https://develop.sentry.dev/sdk/event-payloads/contexts/#trace-context), `unknown` means:

> An unknown error raised by APIs that don't return enough error information

And `trace.status`:

> _Optional_. Whether the trace failed or succeeded.

iker-barriocanal (approved):

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_mobile.snap
edwardgou-sentry: We coalesce this tag to `unknown` if theres no trace status supplied. I think we do want this since `unknown` is a valid value for `trace.status` and it matches the intended usage to me.

I'll use this for now and we can update this later if needed.

PR #3409, opened 2024-04-11
instr(spans): Attach invalid span on error

jjbayer:
Logging the invalid span as `?span` resulted in `[Filtered]` in practice. Use an attachment instead.

ref: [RELAY-2NT9](https://sentry.my.sentry.io/organizations/sentry/issues/791734/)

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor/span/processing.rs
+++ b/relay-server/src/services/processor/span/processing.rs
@@ -15,6 +15,7 @@ use relay_event_normalization::{
 };
 use relay_event_schema::processor::{process_value, ProcessingState};
 use relay_event_schema::protocol::{BrowserContext, Contexts, Event, Span, SpanData};
+use relay_log::protocol::{Attachment, AttachmentType};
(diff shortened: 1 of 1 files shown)


relay-server/src/services/processor/span/processing.rs
```diff
@@ -125,11 +126,22 @@ pub fn process(
...
             Err(err) => {
-                relay_log::error!(
-                    error = &err as &dyn Error,
-                    span = ?annotated_span,
-                    source = "standalone",
-                    "invalid span"
+                relay_log::with_scope(
+                    |scope| {
+                        scope.add_attachment(Attachment {
+                            buffer: annotated_span.to_json().unwrap_or_default().into(),
```
Dav1dde, line 132: Any PII or size concerns?

Dav1dde (approved):

relay-server/src/services/processor/span/processing.rs
jjbayer: PII scrubbing ran at this point, spans are _supposedly_ small in size.

PR #3411, opened 2024-04-11
fix(spans): Do not overwrite `is_segment: true` if set by SDK

jjbayer:
We previously ignored `span.is_segment` if it was already set by the SDK, and set it to `false` for every span that has a `parent_id`. This is wrong because segments can have a `parent_id` in the trace.


phacops:
This PR makes sense since a segment can have a `parent_span_id`. Trusting the `is_segment` flag from the SDK is the right move (or setting it if we don't see a `parent_span_id`).

phacops (approved):

iker-barriocanal (approved):

relay-server/src/services/processor/span/processing.rs
```diff
@@ -414,12 +414,15 @@ fn get_normalize_span_config<'a>(
...
-    if is_segment {
+    // If `is_segment` was not set by the SDK and the span has no parent, assume it is a segment:
+    if is_root && span.is_segment.value().is_none() {
+        span.is_segment = true.into();
+    }
+
+    // If the span is a segment, always set to segment_id to the current span_id:
+    if span.is_segment.value() == Some(&true) {
         span.segment_id = span.span_id.clone();
     }
```
Dav1dde, line 431: Shouldn't we also check for this, if the SDK does that already we have to set `is_segment` to true as well.

jjbayer: You mean check the opposite direction? I.e. set `is_segment` if `segment_id == span_id`?

Dav1dde: Yes I think so, basically, expand `is_root` to:

```rs
let is_root = span.parent_span_id.is_empty() || span.segment_id == span.span_id;
```

jjbayer: Added this now. I also changed the logic to _not_ accept `is_segment: false` from the SDK if the condition above is met.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -414,12 +414,15 @@ fn get_normalize_span_config<'a>(
 fn set_segment_attributes(span: &mut Annotated<Span>) {
     let Some(span) = span.value_mut() else { return };
 
-    // TODO: A span might be a segment span even if the parent_id is not empty
-    // (parent within a trace). I.e. do not overwrite here.
-    let is_segment = span.parent_span_id.is_empty();
+    let is_definitely_segment = span.parent_span_id.is_empty() || span.segment_id == span.span_id;
```
Dav1dde: Hehe nice name

Dav1dde (approved):

relay-server/src/services/processor/span/processing.rs
jjbayer: Hmm the second condition is also true when both are `None`... I'll add some more test coverage to not mess this up.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -413,13 +413,20 @@ fn get_normalize_span_config<'a>(
...
+        // The span is a segment if and only if the segment_id matches the span_id.
+        span.is_segment = (segment_id == span_id).into();
+    } else if span.parent_span_id.is_empty() {
+        // If the span has no parent, it is automatically a segment:
+        span.is_segment = true.into();
+    }
 
-    span.is_segment = Annotated::new(is_segment);
-    if is_segment {
+    // If the span is a segment, always set to segment_id to the current span_id:
```
```suggestion
    // If the span is a segment, always set the segment_id to the current span_id:
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fix performance regression in disk spooling by using page counts to estimate the spool size. ([#3379](https://github.com/getsentry/relay/pull/3379))
 - Perform clock drift normalization only when `sent_at` is set in the `Envelope` headers. ([#3405](https://github.com/getsentry/relay/pull/3405))
+- Do not overwrite `span.is_segment` if already set by SDK. ([#3411](https://github.com/getsentry/relay/pull/3411))
```
```suggestion
- Do not overwrite `span.is_segment: true` if already set by SDK. ([#3411](https://github.com/getsentry/relay/pull/3411))
```

tests/integration/test_spans.py
```diff
@@ -308,8 +308,8 @@ def envelope_with_spans(
                     {
                         "description": "https://example.com/p/blah.js",
                         "op": "resource.script",
-                        "span_id": "bd429c44b67a3eb1",
-                        "segment_id": "968cff94913ebb07",
+                        "span_id": "b0429c44b67a3eb1",
+                        "segment_id": "b0429c44b67a3eb1",
```
jjbayer, line 312: I gave this span a new span_id and the same segment_id, to make sure it is classified as `is_segment`.

tests/integration/test_spans.py
```diff
@@ -526,19 +526,19 @@ def test_span_ingestion(
...
+            "span_id": "b0429c44b67a3eb1",
             "start_timestamp_ms": int(start.timestamp() * 1e3),
             "trace_id": "ff62a8b040f340bda5d830223def1d81",
         },
         {
             "description": r"test \" with \" escaped \" chars",
             "duration_ms": 1500,
             "exclusive_time_ms": 345.0,
-            "is_segment": True,
+            "is_segment": False,
```
jjbayer, line 537: This span has a different segment_id than span_id, so it's not a segment.

tests/integration/test_spans.py
```diff
@@ -598,7 +598,7 @@ def test_span_ingestion(
 
     # If transaction extraction is enabled, expect transactions:
     if extract_transaction:
-        expected_transactions = 5
+        expected_transactions = 3
```
jjbayer, line 601: Two spans that were previously mis-classified as segment are now non-segments.

tests/integration/test_spans.py
```diff
@@ -1230,8 +1225,8 @@ def test_span_ingestion_with_performance_scores(
                         },
                         "profile_id": "3d9428087fda4ba0936788b70a7587d0",
                         "op": "ui.interaction.click",
-                        "span_id": "bd429c44b67a3eb1",
-                        "segment_id": "968cff94913ebb07",
+                        "span_id": "cd429c44b67a3eb1",
+                        "segment_id": "cd429c44b67a3eb1",
```
jjbayer, line 1229: I checked some real `ui.interaction.click` spans in the product and they all have `span_id == segment_id`.

tests/integration/test_spans.py
Dav1dde: Can you put this as a comment in the code?

tests/integration/test_spans.py
Dav1dde: Also please put that as a comment in the code

Dav1dde: Ah already merged, dw about it. I'll throw up a PR.

Dav1dde: #3453

PR #3413, opened 2024-04-12
fix(spans): Scrub extensions in description and tags

phacops:
The cardinality report was flagging some projects with file spans with high cardinality which turned out to be file extensions being numbers.

This will apply the same scrubbing we do on file extensions for resource spans to all extensions.

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -398,7 +398,7 @@ fn scrub_resource_segment(segment: &str) -> Cow<str> {
     segment
 }
 
-pub(crate) fn scrub_resource_file_extension(mut extension: &str) -> &str {
+fn scrub_resource_file_extension(mut extension: &str) -> &str {
     // Only accept short, clean file extensions.
     let mut digits = 0;
     for (i, byte) in extension.bytes().enumerate() {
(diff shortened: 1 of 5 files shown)


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -780,6 +784,13 @@ mod tests {
         "*"
     );
 
+    span_description_test!(
+        span_description_file_extension_with_numbers_only,
+        "backup.2024041101",
```
Dav1dde, line 789: The scrubbing fails on `backup.2024041101.tar.gz` right? Or the simpler `backup-2024041101` 

Seems like we'd need a better approach since the date is just accidentally (?) the last part.

jjbayer: I believe `backup.2024041101.tar.gz` would become `*.gz`, because everything before the top-level extension is removed anyway.

jjbayer (approved):

relay-event-normalization/src/normalize/span/description/mod.rs
Dav1dde: Ah right, I was practically staring at that code!

Dav1dde (approved):

PR #3414, opened 2024-04-12
ref(server): Maintain copy of event item headers

jjbayer:
#3375  added a third event item header that needs to be extracted from the incoming item, updated and then re-serialized back into the event item manually:

```mermaid
flowchart LR
Item --> payload
Item --> headers
headers --> metrics_extracted
headers --> sample_rates
headers --> spans_extracted
payload --> Event
Event --> payload2
metrics_extracted --> headers2
sample_rates --> headers2
spans_extracted --> headers2
payload2 --> Item2
headers2 --> Item2
```

By maintaining an instance of `ItemHeaders` instead of a bunch of individual fields on `ProcessEnvelopeState`, simplify to:

```mermaid
flowchart LR
Item --> payload
Item --> headers
payload --> Event
Event --> payload2
payload2 --> Item2
headers ---> Item2
```

Eventually, we could remove both `event` and `event_headers` from the stat, and make them return values of `event::extract` instead.

#skip-changelog


relay-server/src/envelope.rs
```diff
@@ -481,34 +481,34 @@ fn is_false(val: &bool) -> bool {
 pub struct ItemHeaders {
     /// The type of the item.
     #[serde(rename = "type")]
-    ty: ItemType,
+    pub ty: ItemType,
```
jjbayer, line 484: `Item` and `ItemHeaders` were some of the few serializable data types on which we restricted field access. Making them public seems more aligned with other data types, and is necessary to access its fields during processing.

relay-server/src/envelope.rs
```diff
@@ -566,15 +566,23 @@ pub struct ItemHeaders {
...
-    sampled: bool,
+    pub sampled: bool,
+
+    /// The release, used by security reports.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sentry_release: Option<String>,
+
+    /// The environment, used by security reports.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sentry_environment: Option<String>,
```
jjbayer, line 585: These were previously set and read using string access.

relay-server/src/envelope.rs
```diff
@@ -1003,6 +1013,11 @@ impl Item {
             ItemType::Unknown(_) => false,
         }
     }
+
+    pub fn into_parts(self) -> (ItemHeaders, Bytes) {
```
Dav1dde: Missing docs, isn't there a clippy lint 🤔

relay-server/src/services/processor.rs
```diff
@@ -571,26 +571,17 @@ struct ProcessEnvelopeState<'a, Group> {
...
 
-    /// Track whether transaction metrics were already extracted.
-    event_metrics_extracted: bool,
-
-    /// Track whether spans and span metrics were already extracted.
+    /// Item headers of the extracted event.
     ///
-    /// Only applies to envelopes with a transaction item.
-    spans_extracted: bool,
+    /// `None` if the envelope does not contain an event item.
```
Dav1dde, line 576: Should we just use a default here?

relay-server/src/services/processor/span/processing.rs
```diff
@@ -346,7 +350,9 @@ pub fn extract_from_event(
 
     add_span(transaction_span.into());
 
-    state.spans_extracted = true;
+    if let Some(headers) = state.event_headers.as_mut() {
```
Dav1dde: What if it is not `None`, we should still set it to true?

Dav1dde (commented):
I think we should just use a default if we don't have headers, then set all the things we need to set and when serializing we can rely on `skip_serializing_if`.

relay-server/src/services/processor/span/processing.rs
jjbayer: I realized I can keep the reference from the beginning of the function alive and make it mutable.

Semantically, `state.event_headers` cannot be `None` if `state.event` is not empty, so my first attempt at this PR was to put them in an `Option` together. But that made event handling more complicated because of the two option levels (`Option(Annotated(Option(Event), _))`). Long term, I believe we should remove both from the state and pass them as plain old variables, as you suggested.

relay-server/src/services/processor.rs
jjbayer: `ItemHeaders` does not implement `Default` atm, it would require a default `ItemType` which sounds dangerous. In the end I think we should remove this from the state (see other comment).

relay-server/src/envelope.rs
jjbayer: Removed in favor of destructuring.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -253,7 +253,11 @@ pub fn extract_from_event(
         return;
     };
 
-    if state.spans_extracted {
+    let Some(event_headers) = state.event_headers.as_mut() else {
+        return;
+    };
```
Dav1dde, line 258: Should we log an error here if we don't have any headers? This seems very error prone and is going back to the problem we tried to solve (implicit optional state).

relay-server/src/services/processor.rs
Dav1dde: Should we use something like item headers instead which can represent the "default" case and also can be constructed from `ItemHeaders` as well as be turned into `Option<ItemHeaders>` again?

Dav1dde (approved):
This feels like a step backward and forward at the same time. It makes a bunch of stuff nicer but at the same time adds more of the implicit state we wanted to get rid of to the `ProcessEnvelopeState`.

There is maybe a way to make the `event_headers` not optional (see previous comments), but I can't tell if that's worth the effort.

Overall I think it makes it better and it's just a temporary "bandaid" until we will further improve the processor.

jjbayer:
> Should we use something like item headers instead which can represent the "default" case and also can be constructed from ItemHeaders as well as be turned into Option<ItemHeaders> again?

Thanks, this made me realize there is a bug in the PR (which no test caught): `Item::set_payload` and `Item::new` modify the headers, so we should not overwrite the entire header struct afterwards. This is a good reason to keep `Item::headers` a private field.

Let's skip this change and work toward moving these intermediary values off the state instead.

PR #3415, opened 2024-04-12
ref(kafka): Remove ShardedProducer

### LINKED ISSUE #3339: Remove ShardedProducer
The sharded kafka producer, which does application-level partitioning, is not used in production. We could remove it to simplify the code (needs "breaking change" entry in the changelog).

iambriccardo:
This PR removes the `ShardedProducer` and it's related configuration params since it's not used anymore.

Closes: https://github.com/getsentry/relay/issues/3339


relay-kafka/src/producer/mod.rs
```diff
@@ -345,17 +263,10 @@ impl Producer {
...
-            last_report: Instant::now().into(),
-            inner: ProducerInner::Sharded(ShardedProducer { shards, producers }),
-        }
-    }
-
     /// Sends the payload to the correct producer for the current topic.
     fn send(
         &self,
-        organization_id: u64,
+        _organization_id: u64,
```
iambriccardo: This parameter is not needed anymore but it's piped through so much code that I was wondering if we should keep it for future use. If not, we can remove it.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 - Route spans according to trace_id. ([#3387](https://github.com/getsentry/relay/pull/3387))
 - Log span when encountering a validation error. ([#3401](https://github.com/getsentry/relay/pull/3401))
 - Optionally skip normalization. ([#3377](https://github.com/getsentry/relay/pull/3377))
+- Remove `ShardedProducer`. ([#3415](https://github.com/getsentry/relay/pull/3415))
```
jjbayer: Can we mark this as a breaking change and put it under the other breaking change in `**Features**`?

relay-kafka/src/producer/mod.rs
jjbayer: Let's remove it.

iambriccardo:
I made sure that no active configs in prod are using the sharded configuration.

CHANGELOG.md
Dav1dde: @iambriccardo can you also mark it as a breaking change?

iambriccardo: How do we mark it as a breaking change?

iambriccardo: Nvm just saw

Dav1dde: And now move it below the other breaking change :p.

Yeah it's not very fancy, we should probably at some point rework our changelog a bit.

iambriccardo: I feel like we should just create a set of rules that one should follow since right now it's a bit copy pasty.

jjbayer (approved):

Dav1dde (approved):

Issue #3359, opened 2024-03-28
Form Data events are dropped


Dav1dde:
Form Data events are dropped in the store.

https://github.com/getsentry/relay/blob/3bd026610b483e3c1120238b7e1aa5dd0feecc4d/relay-server/src/services/store.rs#L283-L302

[S4S Issue](https://sentry.my.sentry.io/organizations/sentry/issues/729445/)

jjbayer:
I added some more metadata to the error, but I don't see a pattern. We might need to attach the entire incoming envelope to the error to see if this is a bug or just custom SDK setups sending invalid envelopes.

Dav1dde:
Option: Just drop it in the store.

iambriccardo:
This test replicates the error:
```python
def test_attachments_with_form_data(
    mini_sentry, relay_with_processing, attachments_consumer, outcomes_consumer
):
    project_id = 42
    event_id = "515539018c9b4260a6f999572f1661ee"

    mini_sentry.add_full_project_config(project_id)
    relay = relay_with_processing()
    attachments_consumer = attachments_consumer()
    outcomes_consumer = outcomes_consumer()

    # Create attachments with both file content and form data
    attachments = [
        ("att_1", "foo.txt", b"file content"),
        ("form_key", None, b"form value")  # Form data entry
    ]

    relay.send_attachments(project_id, event_id, attachments)
```

iambriccardo:
The problem was found, we are discussing offline what is the expected handling of form data. If it should be converted into its own event or produced as attachment.

PR #3419, opened 2024-04-12
instr(store): More metadata for dropped form_data

jjbayer:
ref: https://github.com/getsentry/relay/issues/3359

#skip-changelog

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -34,7 +34,7 @@ use crate::services::outcome::{DiscardReason, Outcome, TrackOutcome};
 use crate::services::processor::Processed;
 use crate::statsd::RelayCounters;
 use crate::utils::{
-    self, is_rolled_out, ArrayEncoding, BucketEncoder, ExtractionMode, TypedEnvelope,
+    self, is_rolled_out, ArrayEncoding, BucketEncoder, ExtractionMode, FormDataIter, TypedEnvelope,
 };
 
 /// Fallback name used for attachment items without a `filename` header.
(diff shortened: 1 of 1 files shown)


relay-server/src/services/store.rs
```diff
@@ -302,10 +302,26 @@ impl StoreService {
                         .items()
                         .map(|item| item.ty().as_str())
                         .collect::<Vec<_>>();
+                    let attachment_types = envelope
+                        .items()
+                        .map(|item| {
```
Dav1dde, line 307: This looks like a `filter_map`?

Dav1dde (approved):

relay-server/src/services/store.rs
```diff
@@ -302,10 +302,26 @@ impl StoreService {
...
+                        })
+                        .collect::<Vec<_>>();
 
                     relay_log::with_scope(
                         |scope| {
                             scope.set_extra("item_types", item_types.into());
+                            scope.set_extra("attachment_types", attachment_types.into());
+                            if other == &ItemType::FormData {
+                                let payload = item.payload();
+                                let form_data_keys = FormDataIter::new(&payload)
```
jjbayer, line 320: This parses the payload inside the store service, which is not ideal but it only concerns a very small portion of traffic. We can remove it once the issue is resolved.

relay-server/src/services/store.rs
jjbayer: I want an entry for every envelope item, such that attachment_types and item_types have the same length. Ideally they would be in the same list, but hey, quick and dirty.

PR #3420, opened 2024-04-12
feat(replays): Add inbound filters for Annotated<Replay> types.

cmanallen:
Adds inbound-filters for replay.

Related: https://github.com/getsentry/sentry/issues/44700


cmanallen:
Edit: Oh I see it immediately now.  Its implemented for `Replay` not `Annotated<Replay>`.

@getsentry/ingest Any idea why I'm getting these compile errors?

```Compiling relay-server v24.3.0 (/Users/colton/code/work/relay/relay-server)
error[E0277]: the trait bound `relay_protocol::Annotated<relay_event_schema::protocol::Replay>: Filterable` is not satisfied
   --> relay-server/src/services/processor/replay.rs:132:17
    |
131 |             relay_filter::should_filter(
    |             --------------------------- required by a bound introduced by this call
132 |                 &replay,
    |                 ^^^^^^^ the trait `Filterable` is not implemented for `relay_protocol::Annotated<relay_event_schema::protocol::Replay>`
    |
    = help: the following other types implement trait `Filterable`:
              relay_event_schema::protocol::Event
              relay_event_schema::protocol::Replay
note: required by a bound in `relay_filter::should_filter`
   --> /Users/colton/code/work/relay/relay-filter/src/lib.rs:47:25
    |
47  | pub fn should_filter<F: Filterable + Getter>(
    |                         ^^^^^^^^^^ required by this bound in `should_filter`

error[E0277]: the trait bound `relay_protocol::Annotated<relay_event_schema::protocol::Replay>: Getter` is not satisfied
   --> relay-server/src/services/processor/replay.rs:132:17
    |
131 |             relay_filter::should_filter(
    |             --------------------------- required by a bound introduced by this call
132 |                 &replay,
    |                 ^^^^^^^ the trait `Getter` is not implemented for `relay_protocol::Annotated<relay_event_schema::protocol::Replay>`
    |
    = help: the following other types implement trait `Getter`:
              DynamicSamplingContext
              relay_event_schema::protocol::Event
              relay_event_schema::protocol::Replay
              relay_event_schema::protocol::Span
              SpanData
note: required by a bound in `relay_filter::should_filter`
   --> /Users/colton/code/work/relay/relay-filter/src/lib.rs:47:38
    |
47  | pub fn should_filter<F: Filterable + Getter>(
    |                                      ^^^^^^ required by this bound in `should_filter`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `relay-server` (lib) due to 2 previous errors
```

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 
 **Features**:
 
+- Add inbound filters for Annotated<Replay> types. ([#3420](https://github.com/getsentry/relay/pull/3420))
+
```
Dav1dde: Needs to be updated, after the 24.4.0 release

Dav1dde (commented):
LGTM @jjbayer can you take a look too?

tests/integration/test_replay_events.py
```diff
@@ -136,3 +136,22 @@ def test_replay_events_without_processing(mini_sentry, relay_chain):
...
+    )
+    filter_settings = project_config["config"]["filterSettings"]
+    filter_settings["localhost"] = {"isEnabled": True}
+
+    replay_events_consumer = replay_events_consumer(timeout=10)
+    replay = generate_replay_sdk_event()
+    replay["request"]["url"] = "http://localhost:1200"
+
+    relay.send_replay_event(42, replay)
+
```
jjbayer, line 160: Could you also assert on the `outcomes_consumer` here? There should be a `Filtered` outcome reported to it.

relay-server/src/services/processor.rs
```diff
@@ -503,6 +506,7 @@ impl ProcessingError {
             // These outcomes are emitted at the source.
             Self::MissingProjectId => None,
             Self::EventFiltered(_) => None,
+            Self::ReplayFiltered(_) => None,
```
jjbayer: I think this should map to `Outcome::Filtered(FilterStatKey)`. Then `ManagedEnvelope::reject` will automatically emit outcomes for the envelope's `replay_quantity` (can be verified in the integration test).

relay-server/src/services/processor/replay.rs
```diff
@@ -114,32 +111,55 @@ fn handle_replay_event_item(
...
-            Ok(json) => Ok(json.into_bytes().into()),
-            Err(error) => {
-                relay_log::error!(
-                    error = &error as &dyn Error,
-                    ?event_id,
-                    "failed to serialize replay"
-                );
-                Ok(payload)
+        Ok(replay) => {
+            if let Some(ref replay_type) = replay.0 {
```
iker-barriocanal: nit:
```suggestion
            if let Some(ref replay_type) = replay.value() {
```

relay-server/src/services/processor/replay.rs
```diff
@@ -114,32 +111,55 @@ fn handle_replay_event_item(
...
+
+            match replay.to_json() {
+                Ok(json) => Ok(json.into_bytes().into()),
+                Err(error) => {
+                    relay_log::error!(
+                        error = &error as &dyn Error,
+                        ?event_id,
+                        "failed to serialize replay"
+                    );
+                    Ok(payload)
```
iker-barriocanal, line 146: Not related to this PR, but why does this return `Ok(payload)` if we could not serialize the replay?

tests/integration/test_replay_events.py
```diff
@@ -136,3 +136,22 @@ def test_replay_events_without_processing(mini_sentry, relay_chain):
 
     replay_event = envelope.items[0]
     assert replay_event.type == "replay_event"
+
+
+def test_replay_event_with_processing_other(
```
iker-barriocanal: Why `_other`? Wdyt about a test name to reflect the expectation of filtering out the replay?

iker-barriocanal (commented):
Other than the outcome Joris mentioned, LGTM.

relay-server/src/services/processor.rs
cmanallen: I've added this but I don't see any outcomes being emitted.

tests/integration/test_replay_events.py
cmanallen: I copied the test above and renamed it so they wouldn't conflict.  And then forgot to actually name it.  My bad!

relay-server/src/services/processor.rs
jjbayer: The integration tests are tricky, sometimes the default consumer timeout is not enough for relay to produce all the data. I verified locally that you get an outcome in the integration test with `outcomes_consumer.get_outcome(timeout=5)`.

cmanallen: My mistake.  I wasn't aware.  Thank you!!

relay-server/src/services/processor/replay.rs
cmanallen: I think it was to be permissive in the early days.  We could remove it in another PR.

tests/integration/test_replay_events.py
```diff
@@ -136,3 +136,34 @@ def test_replay_events_without_processing(mini_sentry, relay_chain):
...
+
+    replay_events_consumer = replay_events_consumer(timeout=10)
+    replay = generate_replay_sdk_event()
+    replay["request"]["url"] = "http://localhost:1200"
+
+    relay.send_replay_event(42, replay)
+
+    replay_events_consumer.assert_empty()
+
+    outcome = outcomes_consumer.get_outcome(timeout=5)
```
Dav1dde: I'd switch the order, first wait for the outcome, then assert empty. `assert_empty` just does the best effort thing, wait a tiny bit then assert there are no messages -> if we're already waiting longer we can use that and get better results in the tests.

Dav1dde (approved):

tests/integration/test_replay_events.py
jjbayer: +1, and then call call `.assert_empty()` on the `outcomes_consumer` as well, just for good measure.

jjbayer (approved):

PR #3422, opened 2024-04-15
ref(schema): Split BagSize into max_depth and max_bytes

iker-barriocanal:
This PR splits the `BagSize` into the separate `max_depth` and `max_bytes` attributes.

Following up https://github.com/getsentry/relay/pull/2887, string values are confusing and may lead to investigating where the mappings are happening and applied. Further, updating `BagSize` values impact multiple fields and may have unintended effects.

### Tests

Using the attributes independently is not tested, and we'll do them once we have that use case.

Since the two attributes are now split, setting them up individually is possible. Although this should work, it's not tested since a dummy struct deriving `ProcessValue` [needs to be defined](https://github.com/getsentry/relay/blob/5e319709dbbc51433b3ec9d908b149d3ad405ce6/relay-event-derive/src/lib.rs#L3-L4) in `relay-event-schema` and the test with the `TrimmingProcessor` in `relay-event-normalization`, requiring exposing the dummy struct which is not great. So, I've added a comment to test this in the future. Current use cases are supported as expected.

#skip-changelog

### DIFF
--- a/relay-event-derive/src/lib.rs
+++ b/relay-event-derive/src/lib.rs
@@ -321,8 +321,8 @@ struct FieldAttrs {
     characters: Option<TokenStream>,
     max_chars: Option<TokenStream>,
     max_chars_allowance: Option<TokenStream>,
-    max_struct_depth: Option<TokenStream>,
-    max_struct_bytes: Option<TokenStream>,
+    max_depth: Option<TokenStream>,
+    max_bytes: Option<TokenStream>,
 }
 
 impl FieldAttrs {
@@ -384,18 +384,18 @@ impl FieldAttrs {
             quote!(0)
         };
 
-        let max_struct_depth = if let Some(ref max_struct_depth) = self.max_struct_depth {
-            quote!(Some(#max_struct_depth))
+        let max_depth = if let Some(ref max_depth) = self.max_depth {
+            quote!(Some(#max_depth))
         } else if let Some(ref parent_attrs) = inherit_from_field_attrs {
-            quote!(#parent_attrs.max_struct_depth)
+            quote!(#parent_attrs.max_depth)
         } else {
             quote!(None)
         };
 
-        let max_struct_bytes = if let Some(ref max_struct_bytes) = self.max_struct_bytes {
-            quote!(Some(#max_struct_bytes))
+        let max_bytes = if let Some(ref max_bytes) = self.max_bytes {
+            quote!(Some(#max_bytes))
         } else if let Some(ref parent_attrs) = inherit_from_field_attrs {
-            quote!(#parent_attrs.max_struct_bytes)
+            quote!(#parent_attrs.max_bytes)
         } else {
             quote!(None)
         };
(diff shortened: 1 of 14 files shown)


iker-barriocanal:
@jjbayer I mentioned:
- Some rust-analyzer errors that didn't happen when checking/building Relay. These don't happen anymore (can't remember the errors, don't know why).
- The `Frame` tests I mentioned don't run on the `TrimmingProcessor` and were unrelated.
- The requirement trait requiring `Annotated<Event>` was not correct. The difficulties I was running into were coming from defining the dummy struct for tests in `relay-event-normalization`, instead of `relay-event-schema`.

relay-event-derive/src/lib.rs
```diff
@@ -332,7 +321,8 @@ struct FieldAttrs {
     characters: Option<TokenStream>,
     max_chars: Option<TokenStream>,
     max_chars_allowance: Option<TokenStream>,
-    bag_size: Option<TokenStream>,
+    max_struct_depth: Option<TokenStream>,
+    max_struct_bytes: Option<TokenStream>,
```
jjbayer: Can we call these `max_depth` and `max_bytes`? IMO the `_struct_` infix does not really provide information.

jjbayer (commented):
See comment about naming, rest LGTM. Are there already tests that assert that these attributes work correctly?

relay-event-normalization/src/trimming.rs
```diff
@@ -61,22 +63,22 @@ impl Processor for TrimmingProcessor {
...
         state: &ProcessingState<'_>,
     ) -> ProcessingResult {
-        // If we encounter a bag size attribute it resets the depth and size
-        // that is permitted below it.
-        if let Some(bag_size) = state.attrs().bag_size {
-            self.bag_size_state.push(BagSizeState {
-                size_remaining: bag_size.max_size(),
+        // If we encounter a max_struct_bytes or max_struct_depth attribute it
+        // resets the size and depth that is permitted below it.
+        // XXX(iker): test setting only one of the two attributes.
```
Dav1dde, line 68: 👀

iker-barriocanal:
@jjbayer @Dav1dde re: tests -- this is what I don't like in this PR. There are a bunch of tests in `relay-event-normalization/src/trimming.rs` (e.g. `test_custom_context_trimming`) to validate these attributes work, but no tests that validate them individually. Doing the latter isn't trivial without exposing random stuff (see PR description), so I decided to add the `todo` comment instead and assume that there will be a test when attributes are used individually. I'd love to hear your thoughts or any other suggestions you may have.

Dav1dde:
I think a single test that just makes sure either attribute used independently works is fine, the test can define its own datastructure and use the proc macro on it. I assume that'd be straight forward, but I am also fine with skipping it if you say it's too much effort (for now).

Dav1dde (approved):
As discussed, offline, we'll do tests when we actually use the attributes independently.

PR #3425, opened 2024-04-15
fix(config): Use default values for processing config

iker-barriocanal:
Initializing Relay with the `processing` config fails if `enabled` and `kafka_config` aren't provided. This PR defaults these values to the default implementation.

I don't understand why this doesn't fail in existing integration tests, but I see the failure when spinning up relay locally.

This change also implies that setting a `processing` key without `enabled` and `kafka_config` is now acceptable, which was not before.

#skip-changelog


jan-auer:
Can we leave `enabled` as required? It would be an error to configure processing config but not enable it, so better fail loud and early.

This was done very explicitly to avoid configuration errors.

iker-barriocanal:
I'll move the config to a separate place

PR #3426, opened 2024-04-15
ref(health): Add timeout to health checks

Dav1dde:
Configures 900ms, that should be plenty of time to return the status within 100ms to the caller, so we stay within 1 second for health check responses.

Tried to add an integration test, but it is fast enough that even a timeout of 0ms never fails in tests.

#skip-changelog


gi0baro (approved):
LGTM

jjbayer (approved):

PR #3428, opened 2024-04-15
fix(config): Move normalization config to top level

iker-barriocanal:
This PR unblocks the roll out of enabling normalization only in PoP Relays.

Normalization config was inside the `processing` config, which requires specific config for processing Relays. This is convenient as it raises errors during Relay startup when core features are not properly configured. However, normalization config lives inside the processing one and it's not possible to enable it in non-processing relays. This PR moves it to a separate top-level config to make that possible.

The last Relay release, on March 15, did not include the broken config, so it's ok to assume the config isn't used.

#skip-changelog


tests/integration/test_store.py
```diff
@@ -1320,30 +1320,30 @@ def test_kafka_ssl(relay_with_processing):
...
     relay,
-    normalize,
+    normalization_level,
 ):
     project_id = 42
     mini_sentry.add_basic_project_config(project_id)
     relay = relay(
         upstream=mini_sentry,
-        options={"processing": {"normalize": normalize}},
+        options={"normalization": {"level": normalization_level}},
```
iker-barriocanal, line 1335: The test should have failed with invalid config -- https://github.com/getsentry/relay/issues/3429.

iker-barriocanal (commented):
Open to better suggestions for config names 😄

Dav1dde (approved):
> The last Relay release, on March 15, did not include the broken config, so it's ok to assume the config isn't used.

The last release was yesterday (unlucky), but still fine to assume that.

relay-config/src/config.rs
```diff
@@ -1035,23 +1032,31 @@ impl Default for Processing {
...
+#[derive(Debug, Default, Serialize, Deserialize)]
+#[serde(default)]
+pub struct Normalization {
+    /// Level of normalization for Relay to apply to incoming data.
+    #[serde(default)]
+    pub level: NormalizationLevel,
+}
+
 /// Configuration for the level of normalization this Relay should do.
 #[derive(Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq)]
```
Dav1dde:
```suggestion
#[derive(Copy, Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq)]
```

relay-config/src/config.rs
```diff
@@ -2188,9 +2195,9 @@ impl Config {
         self.values.processing.enabled
     }
 
-    /// How much normalization this Relay should apply to incoming data.
-    pub fn normalization(&self) -> &Normalize {
-        &self.values.processing.normalize
+    /// Level of normalization for Relay to apply to incoming data.
+    pub fn normalization_level(&self) -> &NormalizationLevel {
+        &self.values.normalization.level
```
Dav1dde:
```suggestion
    pub fn normalization_level(&self) -> NormalizationLevel {
        self.values.normalization.level
```

Dav1dde (commented):
Okay, 2 mini nits.

PR #3430, opened 2024-04-15
fix(event): Fix remaining snapshot to fix CI in master

iker-barriocanal:
CI in `master` is currently red because of a stale snapshot.

#skip-changelog


Dav1dde:
Fixed with https://github.com/getsentry/relay/pull/3431

PR #3431, opened 2024-04-15
fix(spans): Fixes a broken spans snapshot

edwardgou-sentry:
Fixes a broken snapshot test, from https://github.com/getsentry/relay/pull/3407, due to merging to master without pulling latest snapshot changes.

#skip-changelog


chadwhitacre:
Overriding and merging as this is blocking https://github.com/getsentry/publish/issues/3696.

PR #3437, opened 2024-04-16
feat(metric-stats): Implement per type cardinality limits

Dav1dde:
Implements by metric type cardinality limits, these limits don't necessarily make sense to be enforced, but can be used for cardinality reports if you want to weigh cardinality for each type differently.

#skip-changelog

### DIFF
--- a/relay-base-schema/src/metrics/name.rs
+++ b/relay-base-schema/src/metrics/name.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 
 use serde::{Deserialize, Serialize};
 
-use crate::metrics::MetricNamespace;
+use crate::metrics::{MetricNamespace, MetricType};
 
 /// Optimized string represenation of a metric name.
 ///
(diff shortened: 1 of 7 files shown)


relay-base-schema/src/metrics/name.rs
```diff
@@ -15,6 +15,36 @@ use crate::metrics::MetricNamespace;
...
+    /// assert_eq!(name.try_type(), Some(MetricType::Counter));
+    /// let name = MetricName::from("d:custom/foo@none");
+    /// assert_eq!(name.try_type(), Some(MetricType::Distribution));
+    /// let name = MetricName::from("s:custom/foo@none");
+    /// assert_eq!(name.try_type(), Some(MetricType::Set));
+    /// let name = MetricName::from("g:custom/foo@none");
+    /// assert_eq!(name.try_type(), Some(MetricType::Gauge));
+    /// ```
+    pub fn try_type(&self) -> Option<MetricType> {
+        match self.0.as_bytes().get(..2) {
```
jjbayer, line 39: Why `as_bytes()`? Is it just to skip the boundary check in [`String::get`](https://doc.rust-lang.org/std/string/struct.String.html#examples-52)?

relay-cardinality/src/config.rs
```diff
@@ -53,6 +53,12 @@ pub enum CardinalityScope {
     /// The limit will be enforced for a specific project.
     Project,
 
+    /// A per metric type cardinality limit.
+    ///
+    /// This limit usually doesn't make sense to enforce,
+    /// but is useful for cardinality reports.
+    Type,
```
jjbayer, line 66: Is this per org, project _and_ type? Might be worth documenting (i.e. not a per-type limit across projects of an org).

relay-cardinality/src/redis/quota.rs
```diff
@@ -112,16 +122,19 @@ impl QuotaScoping {
...
 
         // Use a pre-allocated buffer instead of `format!()`, benchmarks have shown
         // this does have quite a big impact when cardinality limiting a high amount
         // of different metric names.
         let mut result = String::with_capacity(200);
         write!(
             &mut result,
-            "{KEY_PREFIX}:{KEY_VERSION}:scope-{{{organization_id}-{project_id}-{namespace}}}-{name}{slot}"
-        ).expect("formatting into a string never fails");
+            "{KEY_PREFIX}:{KEY_VERSION}:scope-{{{organization_id}-{project_id}-{namespace}}}-{metric_type}{metric_name}{slot}"
```
jjbayer, line 134: Do we need to bump the KEY_VERSION for this?

relay-base-schema/src/metrics/name.rs
Dav1dde: Yes, no need to check for the boundary, we know the first 2 characters must be ascii.

relay-cardinality/src/redis/quota.rs
Dav1dde: No, it's compatible, `type` and `name` are mutually exclusive and there is no overlap between name and type.

jjbayer: Should it be an enum in `QuotaScoping` then?

jjbayer (approved):
Still two open comments, but nothing blocking.

relay-cardinality/src/config.rs
Dav1dde: Thanks, tried to improve the docs, let me know what you think.

relay-cardinality/src/redis/quota.rs
Dav1dde: Yes it should be. But not only `name` and `type` should be in that enum it should be something like:

```rs
enum QuotaBlaScope {
    Organization {
        organization_id: OrganizationId,
    },
    Project {
        organization_id: OrganizationId,
        project_id: ProjectId,
    },
    Type {
        organization_id: OrganizationId,
        project_id: ProjectId,
        metric_type: MetricType,
    },
    Organization {
        organization_id: OrganizationId,
        project_id: ProjectId,
        metric_name: MetricName,
    },
}
```

And we should re-use this in the `CardinalityReport` and throughout the module.

I decided to just not do any enum for now since the code is very close together (QuotaScoping and its parent), but maybe we should.

jjbayer: Up to you! Maybe it's something that can be handled in a follow-up PR?

PR #3438, opened 2024-04-16
feat(metric-stats): Add additional tags to metric stats metrics

Dav1dde:
Reports additional tags with the metric stats metrics.

#skip-changelog

### DIFF
--- a/relay-server/src/metric_stats.rs
+++ b/relay-server/src/metric_stats.rs
@@ -129,6 +129,7 @@ impl MetricStats {
 
         let mut tags = BTreeMap::from([
             ("mri".to_owned(), bucket.name.to_string()),
+            ("mri.type".to_owned(), bucket.value.ty().to_string()),
             ("mri.namespace".to_owned(), namespace.to_string()),
             (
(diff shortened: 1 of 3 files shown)


relay-server/src/metric_stats.rs
```diff
@@ -160,22 +161,30 @@ impl MetricStats {
             return None;
         }
 
-        let name = report.metric_name.as_ref()?;
-        let namespace = name.namespace();
-
-        if !namespace.has_metric_stats() {
-            return None;
-        }
```
jjbayer, line 168: Isn't this important?

Dav1dde: Yes and no, we can now track cardinality limits which are not associated with a metric (`Type`). The on/off is now only controlled through `report: True` on the cardinality limit, which I think does make sense.

I'll kill the `has_metric_stats` in a follow-up, Jan didn't seem to like it anyways, currently it's still needed for the volume metric 👍 .

jjbayer (approved):

Issue #3404, opened 2024-04-10
[INC-696] Validate Kafka configuration on startup


jjbayer:
When Relay is configured to publish to a non-existing topic, it currently starts normally and only starts logging errors when traffic for that topic comes in.

Relay should fail startup when a Kafka topic does not exist.

Note: This issue is a narrower version of https://github.com/getsentry/relay/issues/2766.

Dav1dde:
> Relay should fail startup when a Kafka topic does not exist.

This is possible but breaks with a cluster that relies on topics to be auto created, which is the [default setting](https://kafka.apache.org/documentation/#brokerconfigs_auto.create.topics.enable).

Probably an issue for self hosted and our own tests also rely on this behaviour.

lynnagara:
Self hosted doesn't rely on topic auto-creation (it uses snuba bootstrap + https://github.com/getsentry/self-hosted/blob/67382fd2abc0357e6d39634c4051cae0240da9c2/install/create-kafka-topics.sh for the rest IIRC)

Dav1dde:
Ah okay, I saw the list and though it was a bit short, if the rest is created elsewhere that's good to know.

Our current plan would be to have config flag which toggles a strict mode in which Relay validates topics, question is whether it should be on by default.

PR #3441, opened 2024-04-17
ref(config): Validate some Kafka configuration on startup

Dav1dde:
Checks for unused topic configurations but does not fail, this is useful when preparing to rollout a new topic or rolling back to an older Relay version.

Validates that configuration for every topic can be built.

related to https://github.com/getsentry/relay/issues/3404

#skip-changelog

### DIFF
--- a/relay-kafka/src/config.rs
+++ b/relay-kafka/src/config.rs
@@ -82,18 +82,20 @@ impl KafkaTopic {
 }
 
 macro_rules! define_topic_assignments {
-    ($struct_name:ident {
-        $($field_name:ident : ($kafka_topic:path, $default_topic: literal, $doc: literal)),* $(,)? }) => {
-
+    ($($field_name:ident : ($kafka_topic:path, $default_topic:literal, $doc:literal)),* $(,)?) => {
         /// Configuration for topics.
         #[derive(Serialize, Deserialize, Debug)]
         #[serde(default)]
         pub struct TopicAssignments {
             $(
                 #[serde(alias = $default_topic)]
-                #[doc=$doc]
+                #[doc = $doc]
                 pub $field_name: TopicAssignment,
             )*
+
+            /// Additional topic assignments configured but currently unused by this Relay instance.
+            #[serde(flatten)]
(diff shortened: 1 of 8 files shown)


olksdr (commented):
Is it possible to add the test for this? Maybe integration test?

relay-kafka/src/config.rs
```diff
@@ -82,18 +82,20 @@ impl KafkaTopic {
...
             $(
                 #[serde(alias = $default_topic)]
-                #[doc=$doc]
+                #[doc = $doc]
                 pub $field_name: TopicAssignment,
             )*
+
+            /// Additional topic assignments configured but currently unused by this Relay instance.
+            #[serde(flatten)]
+            pub unused: BTreeMap<String, TopicAssignment>,
```
jjbayer, line 98: Should we deny unknown fields instead? https://serde.rs/container-attrs.html#deny_unknown_fields

Dav1dde: We could, see the PR description, deny unknown adds some behaviour we maybe don't want.

jjbayer: Thanks, I should have read the description. So the error is more like a warning, maybe we should start reporting warnings to sentry.

We could theoretically log the error in a `deserialize_with` function on `unused` (not sure if that's better, just brainstorming here).

jjbayer (approved):

iker-barriocanal (approved):
A test wuld be great

relay-kafka/src/config.rs
Dav1dde: Yeah, initially I planned to do a warning, but I really would like it to be reported to Sentry!

> We could theoretically log the error in a deserialize_with function on unused (not sure if that's better, just brainstorming here).

Open to changing it, but with the current solution the thing I don't like is the `unused_topics` accessor. We'd get rid of that with that.

Maybe we should move the validation to `Config` and then we can just call `config.validate()`, potentially even having a `ValidatedConfig(Config)`.

PR #3442, opened 2024-04-17
ref(browser-extensions): Add new regex for permission denied to inbound filter

HazAT:
There are some errors in Sentry that throw this:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Property_access_denied

Often these errors are caused by extensions that do things that are not allowed.
The samples we found in Sentry are thrown in minified React bundles.

ref: https://github.com/getsentry/sentry-javascript/issues/10882#issuecomment-2059220860

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Internal:**
 
 - Emit negative outcomes in metric stats for metrics. ([#3436](https://github.com/getsentry/relay/pull/3436))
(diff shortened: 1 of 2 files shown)


Dav1dde (approved):

Dav1dde:
@HazAT feel free to update the title (= commit message) again, I think I butchered it a bit

HazAT:
@Dav1dde I like yours more, thx

lobsterkatie (approved):

scttcper (approved):

PR #3443, opened 2024-04-17
feat(protocol): Add Linux distributions to os context

supervacuus:
Users of the Native SDK would love to know and search for the Linux distributions from which their events are coming: https://github.com/getsentry/sentry-native/issues/943

After a basic implementation on the SDK side to parse `/etc/os-release` (https://github.com/getsentry/sentry-native/pull/963), we also need to ensure the data is indexable. This PR is one step towards that goal.

There is a "develop" docs [PR](https://github.com/getsentry/develop/pull/1227) where we should discuss naming, structure, and content.

Cc: @kahest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Add Linux distributions to os context. ([#3443](https://github.com/getsentry/relay/pull/3443))
+
 **Internal:**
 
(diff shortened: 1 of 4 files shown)


supervacuus:
~I am not sure how this change affects [test_duplicate_performance_score]( https://github.com/getsentry/relay/actions/runs/8720528271/job/23922247663?pr=3443#step:9:1251). Maybe there is some assumption in `mini_sentry` that the change fails to address, but I could not find it.~

Dav1dde (approved):

iker-barriocanal (approved):

PR #3444, opened 2024-04-17
feat(spans): Convert `span.data` fields

jjbayer:
Map `span.data.release` to `event.release` and `span.data.environment` to `event.environment`.

This PR also contains a refactor of the `map_fields!` macro to get a flatter declaration.

#skip-changelog


relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -1,124 +1,133 @@
...
+    ($event:expr, $path_root:ident $(. $path_segment:ident)*) => {
+        {
+            let value = ($event).$path_root.value();
+            $(
+                let value = value.and_then(|value|&value.$path_segment.value());
+            )*
+            value
+        }
+    };
+);
```
jjbayer, line 81: I added these helper functions to make the macro itself more modular.

relay-event-schema/src/protocol/span.rs
```diff
@@ -201,11 +201,11 @@ pub struct SpanData {
 
     /// The sentry environment.
     #[metastructure(field = "environment")]
-    pub environment: Annotated<Value>,
+    pub environment: Annotated<String>,
 
     /// The release version of the project.
     #[metastructure(field = "release")]
-    pub release: Annotated<Value>,
+    pub release: Annotated<LenientString>,
```
jjbayer, line 208: I typed out these fields so they could be more easily converted to and from their transaction counterpart. They were previously `Value` because we did not want to fail span deserialization of the entire span if the field contains something else. `span.data` used to be an untyped data bag, but is becoming more and more narrowly defined as we adopt Otel conventions (see https://develop.sentry.dev/sdk/performance/span-data-conventions/).

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -129,38 +135,30 @@ macro_rules! map_fields {
...
+    span.status <=> event.contexts.trace.status,
+    span.trace_id <=> event.contexts.trace.trace_id,
+    span.profile_id <=> event.contexts.profile.profile_id,
+    span.data.release <=> event.release,
+    span.data.environment <=> event.environment
+    ;
+    span.is_segment <= true,
+    span.was_transaction <= true
+    ;
+    EventType::Transaction => event.ty
```
jjbayer, line 161: `map_fields!` is now called with a flat list of field correspondences, as I intended it initially.

relay-spans/src/span.rs
```diff
@@ -83,12 +83,15 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
 
     let span_id = hex::encode(span_id);
     let trace_id = hex::encode(trace_id);
-    let parent_span_id = hex::encode(parent_span_id);
+    let parent_span_id = match parent_span_id.as_slice() {
+        &[] => None,
+        _ => Some(hex::encode(parent_span_id)),
+    };
```
jjbayer, line 89: Needed this fix to prevent errors in transaction normalization (we validate that the `parent_span_id` has a certain format). An empty Otel span ID corresponds to a `null` / missing span ID in sentry:

>    /// The `span_id` of this span's parent span. If this is a root span, then this
>    /// field must be empty. The ID is an 8-byte array.

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-proto/src/proto/tonic/opentelemetry.proto.trace.v1.rs#L117-L118

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -129,38 +135,30 @@ macro_rules! map_fields {
...
+    span.op <=> event.contexts.trace.op,
+    span.parent_span_id <=> event.contexts.trace.parent_span_id,
+    // A transaction corresponds to a segment span, so span_id and segment_id have the same value:
+    span.span_id <=> event.contexts.trace.span_id,
+    span.segment_id <=> event.contexts.trace.span_id,
+    span.status <=> event.contexts.trace.status,
+    span.trace_id <=> event.contexts.trace.trace_id,
+    span.profile_id <=> event.contexts.profile.profile_id,
+    span.data.release <=> event.release,
+    span.data.environment <=> event.environment
```
jjbayer, line 156: These are the new field mappings that this PR adds.

relay-event-schema/src/protocol/span/convert.rs
```diff
@@ -238,7 +238,41 @@ mod tests {
             profile_id: EventId(
                 a0aaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaab,
             ),
-            data: ~,
+            data: SpanData {
```
Dav1dde, line 241: Is this change expected?

relay-event-schema/src/protocol/span/convert.rs
Dav1dde: 🧑‍🍳

Dav1dde (approved):
Macro 🔥

relay-event-schema/src/protocol/span/convert.rs
jjbayer: Yep, we know map `event.release` to `span.data.release` (same for `environment`).

Dav1dde: And that causes span data to exist?

jjbayer: Yes, there was no field that was mapped to `span.data` before so `span.data` was an empty `Annotated` (`data: ~`).

PR #3447, opened 2024-04-17
fix(server): Check rate limits with namespace in fast-path

### LINKED ISSUE #3354: Potential bug: metric rate limits checked without namespace
`Project::check_buckets` and the inner helper `Project::check_buckets_inner` passes an `ItemScoping` without namespace to the rate limit check. This means rate limits and quotas with namespaces will not be checked.

jan-auer:
Checks rate limits for metrics with their namespace in the fast-path.
Previously, the namespace was not passed to the check function, which caused all
namespace-specific rate limits to be skipped during the check.

This also simplifies `check_buckets` and removes an indirection to an inner
method. The signature of the method could no longer be fulfilled, as now there
can be partial rate limiting inside of the function. Originally, this function
was introduced to test that no error log is emitted for one of the failure
branches. However, the test did not actually assert that and so we remove this
test and simplify instead.

#skip-changelog
Fixes https://github.com/getsentry/relay/issues/3354

### DIFF
--- a/relay-server/src/services/project.rs
+++ b/relay-server/src/services/project.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeSet;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


relay-server/src/services/project.rs
```diff
@@ -1309,18 +1287,6 @@ mod tests {
...
-    fn no_error_on_disabled_project() {
-        let project_key = ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap();
-        let mut project = Project::new(project_key, Arc::new(Config::default()));
-        let mut project_state = ProjectState::allowed();
-        project_state.disabled = true;
-        project.state = State::Cached(project_state.into());
-
-        let result = project.check_buckets_inner(Addr::custom().0, vec![]);
-        assert!(matches!(result, CheckedBuckets::ProjectDisabled(0)));
-    }
```
jjbayer, line 1322: I agree that the test did not _actually_ verify that no error was emitted, but reducing test coverage here increases vulnerability against regressions in future refactors. If there is an easy way to capture the logs in the test, as you suggested in the PR description, I would prefer that.

jan-auer: Do we really want to test logging here? This is not part of the functional behavior of this function, it seems wrong to me that we're making assumptions about a specific log level.

relay-server/src/services/project.rs
```diff
@@ -1137,102 +1138,79 @@ impl Project {
...
-        let len = buckets.len();
-
         let Some(project_state) = self.valid_state() else {
-            return CheckedBuckets::ProjectExpired(len);
+            relay_log::error!(
+                tags.project_key = self.project_key.as_str(),
+                "there is no project state: dropping {} buckets",
+                buckets.len(),
+            );
+            return None;
```
Dav1dde, line 1149: We have to call `reject_metrics` here as well, right? I know we didn't before either, but at least something we need to fix in the future.

I'll put that onto the metric stats todo

relay-server/src/services/project.rs
```diff
@@ -1137,102 +1138,79 @@ impl Project {
...
         };
 
-        let limits = self.rate_limits().check_with_quotas(
-            project_state.get_quotas(),
-            scoping.item(DataCategory::MetricBucket),
-        );
+        let namespaces: BTreeSet<MetricNamespace> = buckets
+            .iter()
+            .filter_map(|bucket| bucket.name.try_namespace())
```
Dav1dde, line 1168: This does ignore "invalid" buckets silently, we could use `namespace()` and collect them in `Namespace::Unknown`.

relay-server/src/services/project.rs
```diff
@@ -1137,102 +1138,79 @@ impl Project {
...
-            let reason_code = limits.longest().and_then(|limit| limit.reason_code.clone());
-            utils::reject_metrics::<Vec<Bucket>>(
-                &outcome_aggregator,
-                utils::extract_metric_quantities(&buckets, mode),
-                scoping,
-                Outcome::RateLimited(reason_code),
-                None,
-                None,
+        let mode = project_state.get_extraction_mode();
+        for namespace in namespaces {
```
Dav1dde, line 1172: I think we can check the limits for all namespaces, collect the namespaces which are limited then extract quantities and drop buckets only once.

relay-server/src/services/project.rs
jan-auer: I'll look into this in a separate PR and then try to bring this test back. We'll have to introduce our own Test subscriber for this, which I think is going to be a useful utility once we port more unit tests over from the integration test suite. This one test I'm still torn about, as it asserts behavior that is not part of any contract or signature, but it's not a critical concern for me.

relay-server/src/services/project.rs
jan-auer: Without a valid project state, I think we do not have sufficient information to build the scoping required to emit outcomes. Though we have at least partial scoping, so we could refactor this function to first obtain partial scoping and then use it until the full one is available.

relay-server/src/services/project.rs
jan-auer: Not sure I follow, this will still correctly rate limit such buckets as it will assume an unknown namespace, which will never match any quota with specific namespace. Unknown namespaces will be handled by the component that is responsible for them at another place -- the project cache doesn't make assumptions about which namespace can be ingested.

relay-server/src/services/project.rs
jan-auer: Right. I had that at first and discarded it as it caused much more code and required yet another map with quantities along with refactors to SourceQuantities to allow increments. That added code didn't seem worth the minor performance improvement (if any).

Dav1dde: Ah you're right for a second I thought we didn't need to know which limit was triggered just that a limit for that namespace triggered. But we do need it for the reason code otherwise we could just collect a single source quantities.

relay-server/src/services/project.rs
Dav1dde: Worded this badly, we never end up checking the quota for `Unknown` with the `filter_map` and somebody could have configured one for `Unknown`.

All invalid metrics or metrics in the unknown namespace would not be checked, not even against a global quota without a namespace (all namespaces).

Not sure if this matters though.

Dav1dde (approved):

relay-server/src/services/project.rs
jan-auer: I'll add a check for that in the project config sanitation code, since those would clearly be invalid limits with unintended consequences.

jan-auer: f7e547cee

PR #3449, opened 2024-04-17
feat(ai): Extract AI category

colin-sentry:
We want to separate the 'ai' and 'ai.pipeline' categories to be able to query them independently in pages.


phacops:
You really want to separate `ai.pipeline` as a different category? It seems like it belongs to the `ai` category.

relay-dynamic-config/src/defaults.rs
```diff
@@ -440,6 +440,9 @@ fn span_metrics(transaction_extraction_enabled: bool) -> impl IntoIterator<Item
                 Tag::with_key("span.group")
                     .from_field("span.sentry_tags.group")
                     .when(know_modules_condition.clone() | app_start_condition.clone()),
+                Tag::with_key("span.category")
```
colin-sentry: @Zylphrex says this should be on `know_modules_condition` - but I need it for AI at least

relay-dynamic-config/src/defaults.rs
```diff
@@ -153,7 +153,7 @@ fn span_metrics(transaction_extraction_enabled: bool) -> impl IntoIterator<Item
         & duration_condition.clone();
 
     let know_modules_condition =
-        (is_db.clone() | is_resource.clone() | is_mobile.clone() | is_http.clone())
+        (is_ai.clone() | is_db.clone() | is_resource.clone() | is_mobile.clone() | is_http.clone())
```
jjbayer: Adding `ai` to `know_modules_condition` also activates some tags on the `exclusive_time` metric for this module. Is that on purpose?

colin-sentry: AI doesn't use exclusive time for anything, so the other side effects are probably benign.

jjbayer: Does it use neither `exclusive_time` nor `exclusive_time_light`? In that case, we should not even extract these metrics for `ai` (i.e. add a `!is_ai` to the metric condition fields).

colin-sentry: Done.

colin-sentry: Actually it does use `exclusive_time_light` for SPM and some other functions

jjbayer (dismissed):
Let's not add `is_ai` to `known_modules_condition`, and add it to the necessary tag conditions explicitly instead. After that, I think this PR is good to go!

I will follow-up with a PR that removes `known_modules_condition` entirely -- It makes it hard to understand what modules actually need specific tags.

colin-sentry:
> Let's not add `is_ai` to `known_modules_condition`, and add it to the necessary tag conditions explicitly instead. After that, I think this PR is good to go!
> 
> I will follow-up with a PR that removes `known_modules_condition` entirely -- It makes it hard to understand what modules actually need specific tags.

Done

relay-dynamic-config/src/defaults.rs
iker-barriocanal: Does this mean that `span.category` should be extracted for all known module conditions, besides AI spans? The tag is also extracted for a bunch of HTTP and DB spans.

relay-dynamic-config/src/defaults.rs
```diff
@@ -476,6 +483,9 @@ fn span_metrics(transaction_extraction_enabled: bool) -> impl IntoIterator<Item
                 Tag::with_key("span.origin")
                     .from_field("span.origin")
                     .always(),
+                Tag::with_key("span.category")
+                    .from_field("span.sentry_tags.category")
+                    .always(), // already guarded by condition on metric
```
iker-barriocanal, line 492: The category and op are very similar (see snapshots). Do we need both?

tests/integration/test_spans.py
```diff
@@ -653,30 +653,33 @@ def test_span_ingestion(
             "org_id": 1,
             "project_id": 42,
             "retention_days": 90,
-            "tags": {
-                "span.op": "resource.script",
-                "span.group": "8a97a9e43588e2bd",
-            },
```
iker-barriocanal: The tag `span.group` no longer exists, is this intended?

relay-dynamic-config/src/defaults.rs
colin-sentry: I need it for AI at least - `category = ai.pipeline` is how we identify the top level spans

tests/integration/test_spans.py
colin-sentry: this diff is confusing, but the resource.script is moved further down with span.group - this is the span for db.query now

relay-dynamic-config/src/defaults.rs
colin-sentry: I suppose so, but I'll just make it AI only for now until someone else wants it

relay-server/src/metrics_extraction/event.rs
```diff
@@ -1053,6 +1053,24 @@ mod tests {
                         "ai_total_tokens_used": {
                             "value": 20
                         }
+                    },
+                    "data": {
+                        "ai.pipeline.name": "Autofix Pipeline"
```
colin-sentry, line 1058: This is the name of the parent span which represents a specific AI pipeline.

This gets converted into the tag `span.ai.pipeline.group` which can then be queried in the frontend to find how many tokens were used in total for a single span

Span -> span group -> "foreign key join" on ai.pipeline.group -> sum (counter)

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -47,6 +47,8 @@ pub enum SpanTagKey {
 
     // Specific to spans
     Action,
+    /// The group of the ancestral span with op ai.pipeline.*
+    AIPipelineGroup,
```
iker-barriocanal, line 51: Can we add a test to ensure this tag is used as expected?

relay-dynamic-config/src/defaults.rs
iker-barriocanal: Does looking at the prefix of `op` work? When `category=ai.pipeline`, `op=ai.pipeline.whatever`. Cardinality should be the same, just wondering whether we need the extra tag.

iker-barriocanal (approved):
Let's add a test case I mention below, other than that LGTM.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -462,6 +465,19 @@ pub fn extract_tags(
             span_tags.insert(SpanTagKey::Description, truncated);
         }
 
+        if category == Some("ai") {
+            if let Some(ai_pipeline_name) = span
+                .data
+                .value()
+                .and_then(|data| data.get_value("ai\\.pipeline\\.name"))
```
jjbayer: Please declare this as a field on the `SpanData` type for documentation purposes.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -462,6 +465,19 @@ pub fn extract_tags(
...
+                .and_then(|val| val.as_str())
+            {
+                let mut ai_pipeline_group = format!("{:?}", md5::compute(ai_pipeline_name));
+                ai_pipeline_group.truncate(16);
```
jjbayer, line 476: Is it actually worth hashing this value? The ai pipeline name looks pretty short, so why not set that as a tag directly, instead of hashing it?

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics.snap
```diff
@@ -11572,8 +11576,10 @@ expression: metrics
         tags: {
             "environment": "fake_environment",
             "release": "1.2.3",
-            "span.description": "Autofix Pipeline",
-            "span.group": "86148ae2d6c09430",
+            "span.ai.pipeline.group": "86148ae2d6c09430",
```
colin-sentry, line 11579: @iker-barriocanal there is a snapshot test for it here

relay-dynamic-config/src/defaults.rs
colin-sentry: You can't do prefix matches like that in snuba from what I've read

relay-event-normalization/src/normalize/span/tag_extraction.rs
colin-sentry: It's hashed the same way description is because it needs to be "joined" on the group ID

relay-dynamic-config/src/defaults.rs
jjbayer: IMO having this tag is fine, it does not increase cardinality because it is derived from `op`.

relay-event-normalization/src/normalize/span/tag_extraction.rs
jjbayer: Makes sense, I did not get that it's the span group of the parent before.

Is the AI pipeline always a segment span? In that case, we could alternatively use the segment_name field (see `SpanData`) to store the parent name.

jjbayer (approved):

PR #3450, opened 2024-04-17
log(feedback): Add logging to verify produce to ingest-feedback-events

aliu39:
Relates to https://github.com/getsentry/sentry/issues/66100
Temporary debugging for s4s. Right now the consumer is deployed and rollout rate is set only in s4s. We're having trouble finding metrics in datadog and want to log as a sanity check.

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -599,6 +599,15 @@ impl StoreService {
                     event_type = message.variant(),
                     topic = topic_name,
                 );
+                if topic_name == "ingest-feedback-events" {
+                    if let KafkaMessage::Event(event_msg) = message {
+                        relay_log::info!("Producing to ingest-feedback-events.\n start_time={}\n event_id={}\n project_id={}",
+                            event_msg.start_time,
+                            event_msg.event_id,
+                            event_msg.project_id
+                        );
+                    }
+                }
             }
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 **Internal:**
 
 - Emit negative outcomes in metric stats for metrics. ([#3436](https://github.com/getsentry/relay/pull/3436))
+- Temporarily log when messages are produced to ingest-feedback-events (low volume, s4s only). ([#3450](https://github.com/getsentry/relay/pull/3450))
 
 ## 24.4.0
 


relay-server/src/services/store.rs
```diff
@@ -599,6 +599,15 @@ impl StoreService {
                     event_type = message.variant(),
                     topic = topic_name,
                 );
+                if topic_name == "ingest-feedback-events" {
+                    if let KafkaMessage::Event(event_msg) = message {
+                        relay_log::info!("Producing to ingest-feedback-events.\n start_time={}\n event_id={}\n project_id={}",
```
aliu39, line 604: Is there a way to pass a kv dict of "extras" like in python?

aliu39:
These are the datadog metrics I tried:

[relay](https://app.datadoghq.com/metric/explorer?fromUser=false&fullscreen_end_ts%5B0%5D=1713304879%5B%E2%80%A6%5D218479170&fullscreen_end_ts%5B1%5D=1713304879170&fullscreen_paused=false&fullscreen_refresh_mode=sliding&fullscreen_start_ts=1713218479170&start=1713301852146&end=1713305452146&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYAEwGIFCeGE4a+NoAxkrIIOoIulmWgWZxykaYWAB08to4ORoaojDVcABucCIItfVQeDlQwFrdcU5GMJIiyBoIALQjhVgALItxPAAEAEZYG8NdS+Nwk1IUCDjyEFk81Rga455wABQAlCA8ALpUru54mKHhH6qP4YGJlBLvL4gWZYNC5UDyDBwhAIXLJHAwJxZf7NLJJNCiOBOOSKZQZAlQfGEpz0JjKERuDxod78CD2KrEhRogkiJSfHh8aHyAkIADCUmEMBQIj+aB4QA) -- corresponds to [this code](https://github.com/getsentry/relay/blob/56c3eb8a4f6d72d86d18ee34d01de15d5c42acea/relay-server/src/services/store.rs#L597)

[consumer group (see "consumer offset rate")](https://app.datadoghq.com/dashboard/qkh-eey-uxf/sentry-consumers?fromUser=false&refresh_mode=sliding&tpl_var_consumer_group%5B0%5D=ingest-feedback-events-consumers&tpl_var_sentry_region%5B0%5D=st-sentry4sentry&view=spans&from_ts=1713391616377&to_ts=1713392516377&live=true)

relay-server/src/services/store.rs
Dav1dde: Why are our existing metrics not enough?

We shouldn't log in potentially high throughput code (on info level).

Dav1dde: Ha, nvm, I just read your description. I'll take a look at the metrics tomorrow.

aliu39: Sounds good! This log won't be high throughput. Since the topic, consumer, and rollout rate only exist/are set in s4s. And there's little to no feedback sent in s4s, except employees who use the feedback widget on a few sentry pages (ex: performance).

aliu39:
I also tried sending feedbacks locally as a sanity check.
- set the option to 1.0 with sentry shell
- Local demo app -> local relay (with println logging) -> kafka devservices container -> devserver
- same thing w/option = 0.0

At each step of this pipeline, I found logs to verify it's working as expected. I also checked #feed-options-automator to verify the option was set to 1.0 in s4s. So this is my last idea to debug before passing it to ops

aliu39:
I should also mention I made a project in s4s and successfully sent feedback to it. We just have no proof it's going through the new infra

aliu39:
Resolved -- s4s metrics now visible in the dashboards above

PR #3451, opened 2024-04-17
feat(perf): Extract `span.domain` tag from the `server.address` span attribute

gggritso:
If the `server.address` span attribute is present, prefer using it for the `span.domain` tag, instead of parsing the URL.

1. `server.address`, if present, should be the authoritative answer to what address the `http.client` span request went to. If it's present, we should obey it
2. Relative URLs cannot be parsed! Instead of attempting to parse a URL we know might fail, let's try the `server.address` property instead


gggritso:
This is completely not necessary!

PR #3452, opened 2024-04-18
 feat(ddm): Add negative outcomes to metric stats in pops

iambriccardo:
This PR includes the following changes:
* Adds negative outcomes to the project service.
* Removes a series of `#[cfg(feature = "processing")]` since now we are also using metric stats in pops.
* Metric stats now accepts types that can be converted into `BucketView` for more flexibility.
* Adds integration test for making sure cardinality limited outcomes are emitted.

### DIFF
--- a/relay-metrics/src/view.rs
+++ b/relay-metrics/src/view.rs
@@ -6,6 +6,7 @@ use crate::{
     aggregator, BucketMetadata, CounterType, DistributionType, GaugeValue, MetricName, SetType,
     SetValue,
 };
+use relay_base_schema::metrics::MetricType;
 use std::collections::BTreeMap;
 use std::fmt;
 use std::ops::Range;
@@ -385,6 +386,11 @@ impl<'a> BucketView<'a> {
             BucketValue::Set(s) => BucketViewValue::Set(SetView::new(s, self.range.clone())),
             BucketValue::Gauge(g) => BucketViewValue::Gauge(*g),
         }
+    }
+
+    /// Type of the value of the bucket view.
+    pub fn value_ty(&self) -> MetricType {
+        self.inner.value.ty()
     }
 
     /// Name of the bucket.
--- a/relay-metrics/src/view.rs
+++ b/relay-metrics/src/view.rs
@@ -389,8 +389,13 @@ impl<'a> BucketView<'a> {
     }
 
     /// Type of the value of the bucket view.
-    pub fn value_ty(&self) -> MetricType {
-        self.inner.value.ty()
+    pub fn ty(&self) -> MetricType {
+        match &self.inner.value {
+            BucketValue::Counter(_) => MetricType::Counter,
+            BucketValue::Distribution(_) => MetricType::Distribution,
+            BucketValue::Set(_) => MetricType::Set,
+            BucketValue::Gauge(_) => MetricType::Gauge,
+        }
     }
 
     /// Name of the bucket.
--- a/relay-server/src/metric_stats.rs
+++ b/relay-server/src/metric_stats.rs
@@ -6,7 +6,9 @@ use relay_cardinality::{CardinalityLimit, CardinalityReport};
 use relay_config::Config;
 #[cfg(feature = "processing")]
 use relay_metrics::GaugeValue;
-use relay_metrics::{Aggregator, Bucket, BucketValue, MergeBuckets, MetricName, UnixTimestamp};
+use relay_metrics::{
+    Aggregator, Bucket, BucketValue, BucketView, MergeBuckets, MetricName, UnixTimestamp,
+};
 use relay_quotas::Scoping;
 use relay_system::Addr;
 
@@ -61,20 +63,23 @@ impl MetricStats {
     }
 
     /// Tracks the metric volume and outcome for the bucket.
-    pub fn track_metric(&self, scoping: Scoping, bucket: Bucket, outcome: Outcome) {
-        // TODO: we want to use a reference of bucket view.
+    pub fn track_metric<'a, V>(&self, scoping: Scoping, bucket: V, outcome: Outcome)
+    where
+        V: Into<BucketView<'a>>,
+    {
         if !self.is_enabled(scoping) {
             return;
         }
 
-        let Some(volume) = self.to_volume_metric(&bucket, &outcome) else {
+        let bucket_view = bucket.into();
+        let Some(volume) = self.to_volume_metric(&bucket_view, &outcome) else {
             return;
         };
 
         relay_log::trace!(
             "Tracking volume of {} for mri '{}': {}",
-            bucket.metadata.merges.get(),
-            bucket.name,
+            bucket_view.metadata().merges.get(),
+            bucket_view.name(),
             outcome
         );
         self.aggregator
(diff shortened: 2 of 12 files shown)


relay-server/src/services/project_cache.rs
```diff
@@ -480,6 +482,7 @@ impl Services {
         test_store: Addr<TestStore>,
         upstream_relay: Addr<UpstreamRelay>,
         global_config: Addr<GlobalConfigManager>,
+        metric_stats: MetricStats,
```
iambriccardo: Not really happy about this, since metric stats it's not accessed as a service. I could put the param top level.

relay-server/src/metric_stats.rs
```diff
@@ -61,27 +59,30 @@ impl MetricStats {
     }
 
     /// Tracks the metric volume and outcome for the bucket.
-    pub fn track_metric(&self, scoping: Scoping, bucket: Bucket, outcome: Outcome) {
+    pub fn track_metric<'a, V>(&self, scoping: Scoping, bucket: V, outcome: Outcome)
```
iambriccardo: Decided to switch the implementation to take any type that can be converted into a `BucketView` since we ended up thinking that the semantics of owning the supplied `Bucket` introduce complications in places where this code is called.

relay-server/src/metric_stats.rs
```diff
@@ -201,11 +206,13 @@ impl MetricStats {
     }
 }
 
-#[cfg(all(test, feature = "processing"))]
```
iambriccardo, line 207: We want to enable most of the tests in any case, but right now we track cardinality only if processing is enabled so those tests are still kept only for processing, to avoid unused variable warnings.

relay-metrics/src/view.rs
```diff
@@ -387,6 +388,11 @@ impl<'a> BucketView<'a> {
         }
     }
 
+    /// Type of the value of the bucket view.
+    pub fn value_ty(&self) -> MetricType {
+        self.inner.value.ty()
+    }
+
```
Dav1dde: I think we should implement `ty()` on `BucketViewValue` instead.

relay-server/src/metric_stats.rs
```diff
@@ -61,19 +62,23 @@ impl MetricStats {
     }
 
     /// Tracks the metric volume and outcome for the bucket.
-    pub fn track_metric(&self, scoping: Scoping, bucket: Bucket, outcome: Outcome) {
+    pub fn track_metric<'a, V>(&self, scoping: Scoping, bucket: V, outcome: Outcome)
+    where
+        V: Into<BucketView<'a>>,
```
Dav1dde:
```suggestion
    pub fn track_metric<'a, T>(&self, scoping: Scoping, bucket: T, outcome: Outcome)
    where
        T: Into<BucketView<'a>>,
```

relay-server/src/metric_stats.rs
```diff
@@ -61,19 +62,23 @@ impl MetricStats {
...
+    {
         if !self.is_enabled(scoping) {
             return;
         }
 
-        let Some(volume) = self.to_volume_metric(&bucket, &outcome) else {
+        let bucket_view = bucket.into();
```
Dav1dde: I'd just shadow bucket here:

```suggestion
        let bucket = bucket.into();
```

relay-server/src/metric_stats.rs
```diff
@@ -120,20 +125,20 @@ impl MetricStats {
         is_rolled_out(organization_id, rate)
     }
 
-    fn to_volume_metric(&self, bucket: &Bucket, outcome: &Outcome) -> Option<Bucket> {
-        let volume = bucket.metadata.merges.get();
+    fn to_volume_metric(&self, bucket_view: &BucketView, outcome: &Outcome) -> Option<Bucket> {
```
Dav1dde:
```suggestion
    fn to_volume_metric(&self, bucket: &BucketView<'_>, outcome: &Outcome) -> Option<Bucket> {
```

- `bucket` is good enough, that's all we care about, the view is already in the type
- Personal preference: I like keeping the lifetime around even if the compiler infers it.

relay-server/src/metric_stats.rs
Dav1dde: Nice 🙏

relay-server/src/utils/metrics_rate_limits.rs
```diff
@@ -94,25 +94,31 @@ where
...
 }
 
-pub fn reject_metrics<I: IntoIterator<Item = Bucket>>(
+pub fn reject_metrics<'a, I, V>(
     addr: &Addr<TrackOutcome>,
     quantities: SourceQuantities,
     scoping: Scoping,
     outcome: Outcome,
     metric_stats: Option<&MetricStats>,
     buckets: Option<I>,
```
Dav1dde:
```suggestion
pub fn reject_metrics<'a, T: Into<BucketView<'a>>>(
    addr: &Addr<TrackOutcome>,
    quantities: SourceQuantities,
    scoping: Scoping,
    outcome: Outcome,
    metric_stats: Option<&MetricStats>,
    buckets: impl IntoIterator<Item = T>,
```

This should work and means you only have to supply a single generic. The trick is that `None` implements `IntoIterator`.

relay-server/src/services/processor.rs
```diff
@@ -2886,7 +2886,7 @@ impl UpstreamRequest for SendMetricsRequest {
                 // Request did not arrive, we are responsible for outcomes.
                 Err(error) if !error.is_received() => {
                     for (scoping, quantities) in self.quantities {
-                        utils::reject_metrics::<Vec<Bucket>>(
+                        utils::reject_metrics::<Vec<&Bucket>, &Bucket>(
```
Dav1dde: I can fix ~~her~~ this. Maybe ... In a future refactor ...

relay-server/src/services/project.rs
```diff
@@ -1317,7 +1321,14 @@ mod tests {
         project_state.disabled = true;
         project.state = State::Cached(project_state.into());
 
-        let result = project.check_buckets_inner(Addr::custom().0, vec![]);
+        let (addr, _) = Addr::custom();
+        let metric_stats = MetricStats::new(
+            Arc::new(Config::default()),
+            GlobalConfigHandle::fixed(GlobalConfig::default()),
+            addr,
+        );
```
Dav1dde: Worth adding a `MetricStats::test()` constructor?

relay-server/src/services/project_cache.rs
Dav1dde: Not a fan either, maybe should just pass it as a separate arg like `config` to be consistent?

iambriccardo: Yes, lets do that.

relay-server/src/utils/metrics_rate_limits.rs
iambriccardo: The problem here is that the iterator will return one element or zero of type `T` which in our case would become `Vec<Bucket>` which doesn't implement `From` for `BucketView`.

Dav1dde: I am not sure I understand, I was thinking of something like this:

```rs

fn foo<T: Into<u32>>(_: impl IntoIterator<Item = T>) {}

fn main() {
    foo::<u16>(None);
    foo(vec![1u16]);
}
```

Dav1dde (approved):
:shipit:

relay-server/src/services/processor.rs
```diff
@@ -2894,13 +2895,13 @@ impl UpstreamRequest for SendMetricsRequest {
                 // Request did not arrive, we are responsible for outcomes.
                 Err(error) if !error.is_received() => {
                     for (scoping, quantities) in self.quantities {
-                        utils::reject_metrics::<Vec<Bucket>>(
+                        utils::reject_metrics::<&Bucket>(
                             &self.outcome_aggregator,
+                            &self.metric_stats,
```
iambriccardo, line 2900: Right now we propagate the `metric_stats` but it's not needed since we pass no buckets. This is done because in a follow up ticket we will have to see how we want to propagate upstream the volume of rejected `BucketView` values. The reason for why this doesn't work right now can be seen in a comment inside `reject_metrics`.

PR #3454, opened 2024-04-18
fix(runtime): Bump service runtimes with tokio::spawn from 1 to 2 threads

Dav1dde:
Stats are showing a backlog in the router service and some health checks timing out. The router service makes use of `tokio::spawn` it is possible that the backlog is made worse by the runtime being single threaded.

The backlog leads to health checks timing out.

This change did have very positive effects on the project cache. If it doesn't seem to improve the situation we can revert it again.

#skip-changelog

### DIFF
--- a/relay-server/src/service.rs
+++ b/relay-server/src/service.rs
@@ -322,10 +322,10 @@ impl Runtimes {
     #[allow(unused_variables)]
     pub fn new(config: &Config) -> Self {
         Self {
-            upstream: create_runtime("upstream-rt", 1),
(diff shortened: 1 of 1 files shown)


iambriccardo (approved):

jjbayer (approved):
Should we just do this for every service that implements a health check?

Dav1dde:
> Should we just do this for every service that implements a health check?

Only upstream left 🤣

Dav1dde:
Bumped the other runtimes as well which contain a `tokio::spawn`. `Upstream` is also called from a health check.

PR #3456, opened 2024-04-18
feat(spans): Add more fields to span <-> transaction conversion

jjbayer:
Add some more fields to the span-transaction bidirectional mapping:

```rust
  span.data.segment_name <=> event.transaction, // transaction gets mapped to both segment_name and description
  span.data.browser_name <=> event.contexts.browser.name,
  span.data.sdk_name <=> event.client_sdk.name
```

This PR also fixes some problems with the `map_fields!` macro.

ref: #3278

#skip-changelog

PR #3457, opened 2024-04-18
feat(spans): Align otel attributes with sentry span

jjbayer:
To simplify conversion in the future, align attribute names in Otel's span attributes with keys in Sentry's `span.data`. The old keys in `span.data` remain as legacy alias.

See also https://github.com/getsentry/develop/pull/1230 for the span schema / attribute convention.


relay-event-schema/src/protocol/span.rs
```diff
@@ -290,10 +290,10 @@ pub struct SpanData {
     pub user: Annotated<Value>,
 
     /// Replay ID
-    #[metastructure(field = "replay_id")]
+    #[metastructure(field = "sentry.replay.id", legacy_alias = "replay_id")]
     pub replay_id: Annotated<Value>,
 
-    #[metastructure(field = "sdk.name")]
+    #[metastructure(field = "sentry.sdk.name")]
```
jjbayer: No legacy alias need here, this field was only introduced in https://github.com/getsentry/relay/pull/3456.

relay-server/src/services/store.rs
```diff
@@ -1357,6 +1357,7 @@ struct SpanKafkaMessage<'a> {
     event_id: Option<EventId>,
     #[serde(rename(deserialize = "exclusive_time"))]
     exclusive_time_ms: f64,
+    #[serde(default)]
     is_segment: bool,
```
jjbayer, line 1363: When we cannot derive `is_segment`, is should be set to `false`.

relay-spans/src/otel_to_sentry_tags.rs
```diff
@@ -1,34 +0,0 @@
...
-        ("sentry.source", "source"),
-        ("sentry.user.email", "user.email"),
-        ("sentry.user.geo.city", "user.geo.city"),
-        ("sentry.user.geo.country_code", "user.geo.country_code"),
-        ("sentry.user.geo.region", "user.geo.region"),
-        ("sentry.user.ip_address", "user.ip_address"),
-        ("sentry.user.segment", "user.segment"),
-        ("sentry.user.username", "user.username"),
-        ("url.full", "request.url"),
-        ("url.query_string", "request.query_string"),
```
jjbayer, line 32: @phacops do you know if any of these mappings was already used for product features? If not, I'd rather remove the mapping and start fresh.

relay-event-schema/src/protocol/span.rs
```diff
@@ -200,11 +200,11 @@ pub struct SpanData {
     pub db_system: Annotated<Value>,
 
     /// The sentry environment.
-    #[metastructure(field = "environment")]
+    #[metastructure(field = "sentry.environment", legacy_alias = "environment")]
```
jjbayer, line 203: The disadvantage of this rename is that the new key will now be serialized into the payload. But `span.data` is not (yet) forwarded to sentry directly on standalone spans (we only extract some of it into `sentry_tags`, but that happens in relay). Spans as part of transactions are materialized into nodestore, but they don't rely on these particular `span.data` items AFAIK.

relay-spans/src/span.rs
```diff
@@ -169,11 +185,8 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
             (otel_span.end_time_unix_nano - otel_span.start_time_unix_nano) as f64 / 1e6f64;
     }
 
-    // TODO: This is wrong, a segment could still have a parent in the trace.
-    let is_segment = parent_span_id.is_none().into();
-
```
jjbayer, line 174: Relying on `set_segment_attributes` instead.

relay-spans/src/span.rs
```diff
@@ -88,79 +94,89 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
...
-            };
-            if key == "sentry.op" {
-                op = otel_value_to_string(value);
-            } else if key.starts_with("db") {
-                op = op.or(Some("db".to_string()));
-                if key == "db.statement" {
-                    if let Some(statement) = otel_value_to_string(value) {
-                        description = statement;
-                    }
+            match attribute.key.as_str() {
```
jjbayer, line 108: Sorry for the large diff, I converted this `if` to a `match` expression.

relay-spans/src/span.rs
```diff
@@ -391,4 +408,153 @@ mod tests {
...
+                    "value": {
+                        "stringValue": "my 1st transaction"
+                    }
+                }
+            ]
+        }"#;
+        let otel_span: OtelSpan = serde_json::from_str(json).unwrap();
+        let span_from_otel = otel_to_sentry_span(otel_span);
+
+        // TODO: measurements, metrics_summary
```
jjbayer, line 482: This will need to be handled in a follow-up.

relay-spans/src/span.rs
Dav1dde: Large diffs that clean up code make me happy.

relay-spans/src/span.rs
```diff
@@ -169,11 +185,8 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
             (otel_span.end_time_unix_nano - otel_span.start_time_unix_nano) as f64 / 1e6f64;
     }
 
-    // TODO: This is wrong, a segment could still have a parent in the trace.
-    let is_segment = parent_span_id.is_none().into();
-
     if let (Some(http_method), Some(http_route)) = (http_method, http_route) {
-        description = format!("{} {}", http_method, http_route);
+        description = Some(format!("{} {}", http_method, http_route));
```
Dav1dde:
```suggestion
        description = Some(format!("{http_method} {http_route}"));
```

Dav1dde (approved):

relay-event-schema/src/protocol/span.rs
```diff
@@ -292,11 +292,11 @@ pub struct SpanData {
     pub user: Annotated<Value>,
 
     /// Replay ID
-    #[metastructure(field = "replay_id")]
+    #[metastructure(field = "sentry.replay.id", legacy_alias = "replay_id")]
     pub replay_id: Annotated<Value>,
 
     /// The sentry SDK (see [`crate::protocol::ClientSdkInfo`]).
-    #[metastructure(field = "sdk.name")]
+    #[metastructure(field = "sentry.sdk.name")]
```
jjbayer, line 299: No need for an alias here, this field was only introduced in https://github.com/getsentry/relay/pull/3456.

PR #3458, opened 2024-04-18
feat(spans): Respect country code TLDs when scrubbing span tags

gggritso:
Domains like `application.co.uk` get scrubbed out to `*.co.uk` which is not useful! `.co.uk` is a ccTLD, and should be left intact. I'm using the `psl` crate, which is a pretty popular one, and it powers some popular but higher-level libraries like `addr`. The new scrubber knows about ccTLDs and leaves them alone.

The unit test coverage is good, but my knowledge of Rust is not. Would love some help on cleaning up the code line-wise!


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add inbound filters for Annotated<Replay> types. ([#3420](https://github.com/getsentry/relay/pull/3420))
 - Add Linux distributions to os context. ([#3443](https://github.com/getsentry/relay/pull/3443))
+- Parse ccTLDs. ([#3458](https://github.com/getsentry/relay/pull/3458))
```
jjbayer: I think we could even move this to the `**Bug Fixes**` section.

```suggestion
- Respect country code TLDs when scrubbing span tags. ([#3458](https://github.com/getsentry/relay/pull/3458))
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -279,25 +278,51 @@ pub fn scrub_ipv6(ip: Ipv6Addr) -> &'static str {
...
-            .join("."),
-    )
+    let parsed_domain = psl::domain(domain.as_bytes());
+
+    if parsed_domain.is_none() {
+        // If parsing fails, return the original string
+        return Cow::Borrowed(domain);
+    }
+
+    let parsed_domain = parsed_domain.unwrap();
```
```suggestion
    let Some(parsed_domain) = parsed_domain else {
        // If parsing fails, return the original string
        return Cow::Borrowed(domain);
    };
```

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -279,25 +278,51 @@ pub fn scrub_ipv6(ip: Ipv6Addr) -> &'static str {
...
-            .iter()
-            .chain(domain.iter())
-            .chain(tld.iter())
-            .join("."),
-    )
+    let parsed_domain = psl::domain(domain.as_bytes());
+
+    if parsed_domain.is_none() {
+        // If parsing fails, return the original string
+        return Cow::Borrowed(domain);
```
jjbayer, line 293: Do we have a test case for this branch?

relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -279,25 +278,51 @@ pub fn scrub_ipv6(ip: Ipv6Addr) -> &'static str {
...
+                Some("") => "",
+                _ => "*.",
+            };
+
+            format!("{subdomain}{second_level_domain}")
+        }
+        Err(_) => String::from("*"),
+    };
+
+    return Cow::Owned(format!("{scrubbed_domain}{scrubbed_suffix}"));
```
jjbayer: I gave this a shot, not sure if it's more readable but it uses fewer string allocations:

```suggestion
    let suffix = parsed_domain.suffix().as_bytes();
    let Some(second_level_domain) = parsed_domain.as_bytes().strip_suffix(suffix) else {
        return Cow::Borrowed(domain);
    };

    let subdomain = domain
        .as_bytes()
        .strip_suffix(suffix)
        .and_then(|s| s.strip_suffix(second_level_domain));

    match subdomain {
        None | Some(b"") => Cow::Borrowed(domain),
        Some(_subdomain) => {
            let scrubbed = [b"*.", second_level_domain, suffix].concat();
            match String::from_utf8(scrubbed) {
                Ok(s) => Cow::Owned(s),
                Err(_) => Cow::Borrowed("*"),
            }
        }
    }
```

relay-event-normalization/src/normalize/span/description/mod.rs
Dav1dde: ~~If you change it to a let-else you should also use the bound value, I assume Joris just forgot to change it in the diff.~~

jjbayer: `parsed_domain` is used on line 298.

Dav1dde: Eh good morning, I am just blind.

relay-event-normalization/Cargo.toml
```diff
@@ -20,6 +20,7 @@ itertools = { workspace = true }
 maxminddb = { workspace = true }
 md5 = { workspace = true }
 once_cell = { workspace = true }
+psl = "2.1.33"
```
Dav1dde: Everything else is in the workspace, please put it to the rest.

I assume we need this, but this is a gigantic dependency in terms of size, is there no alternative?

jjbayer: We could source the [Public Suffix List](https://publicsuffix.org/list/public_suffix_list.dat) manually, but I assume you'd end up with a similar size. We could theoretically limit ourselves to the 100 most popular suffixes or something like that.

Dav1dde: Why is that link ~300kb but the list in the [psl crate](https://github.com/addr-rs/psl/blob/main/src/list.rs) 2.2mib?

jjbayer: That's allegedly the same list compiled into rust code, but I cannot find the script that compiles it.

gggritso: It looks like the list is compiled using [psl_codegen](https://docs.rs/psl-codegen/latest/psl_codegen/fn.compile_psl.html) Maybe we can provide a shorter list of suffixes? That said, is the 2.2MiB size important? [crates.io](https://crates.io/crates/psl) says it's only 187KiB which seems alright, no?

Dav1dde: The size is the compressed size. If we need the crate (it's the best option) it's fine, but if there is a chance to not have an even bigger binary, then yeah I think it matters.

gggritso: That's fair!

I chose `psl` because it's the most popular on crates.io, and it's recommended by the Public Suffix List website for being fast (which, apparently, is at the cost of binary size). My thinking is that if someone's gone through the trouble of curating a full list of suffixes, we might as well use it!

I think the alternative would be to use the crates that `psl` itself relies on, and supply our own list of suffixes. This would (I assume) reduce the binary size (unclear how much) but then we'd have to maintain the suffix list ourselves, and would add some complexity to the repo.

I'll defer to your judgement on this, but it seems like using `psl`, if nothing else, has reasonable tradeoffs

Dav1dde: 👍 thanks for the research, sounds good to me!

relay-event-normalization/src/normalize/span/description/mod.rs
gggritso: There's a docstring test for this case 👍🏻

Dav1dde (approved):

jjbayer (approved):

PR #3462, opened 2024-04-19
ref(kafka): Simplify producer types

jjbayer:
Remove `SingleProducer` and `ProducerInner` types.

Follow-up to https://github.com/getsentry/relay/pull/3415.

#skip-changelog

### DIFF
--- a/relay-kafka/src/producer/mod.rs
+++ b/relay-kafka/src/producer/mod.rs
@@ -92,7 +92,7 @@ struct Producer {
 impl Producer {
     fn new(topic_name: String, producer: Arc<ThreadedProducer>) -> Self {
         Self {
-            last_report: Instant::now().into(),
+            last_report: Cell::new(Instant::now()),
             topic_name,
             producer,
         }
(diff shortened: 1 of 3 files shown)


relay-kafka/src/producer/mod.rs
```diff
@@ -80,16 +80,29 @@ pub trait Message {
...
     topic_name: String,
     /// Real kafka producer.
     producer: Arc<ThreadedProducer>,
 }
 
-impl fmt::Debug for SingleProducer {
+impl Producer {
+    fn new(topic_name: String, producer: Arc<ThreadedProducer>) -> Self {
+        Self {
+            last_report: Instant::now().into(),
```
Dav1dde:
```suggestion
            last_report: Cell::new(Instant::now()),
```

Personal opinion, I'd use `into` only if the type isn't in scope.

Dav1dde (approved):
Nice!

olksdr (approved):

relay-kafka/src/producer/mod.rs
jjbayer: Agree, explicit is better than implicit.

PR #3466, opened 2024-04-19
feat(spans): collect exclusive_time_light for cache spans

DominikB2014:
1. Collect exclusive time light metrics for cache.* spans
- We have cases where we don't need to filter by transaction in the UI and without the light metric, discover will not return any data. Discover always attemps to query for the light metric if the transaction field is not passed into it in the query.

2. Fix some typo's, we were trying to extract `cache_hit` instead of `cache.hit` as defined here
https://github.com/getsentry/relay/blob/72918c1be88e4a034fa807ca03b86f57e23097c2/relay-event-normalization/src/normalize/span/tag_extraction.rs#L108


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
```
DominikB2014, line 3: @getsentry/ingest is there anything wrong with my changelog entry here?

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+
+**Features**:
+
+**Internal**:
+
+- Collect exclusive_time_light metrics for cache.\* spans ([#3466]https://github.com/getsentry/relay/pull/3466)
```
phacops:
```suggestion
- Collect exclusive_time_light metrics for `cache.*` spans. ([#3466](https://github.com/getsentry/relay/pull/3466))
```

CHANGELOG.md
phacops: Yes, see below for my suggestion.

CHANGELOG.md
colin-sentry: the format is incorrect, should be

`([#3466](https://github.com/getsentry/relay/pull/3466))`

phacops (approved):

CHANGELOG.md
DominikB2014: Thanks!

jjbayer (approved):

iker-barriocanal (approved):

PR #3468, opened 2024-04-19
feat(crons): Add `owner` to `MonitorConfig`

wedamija:
We recently added the ability to assign an owner to a Monitor. We want to support setting this via upsert.

### DIFF
--- a/relay-monitors/src/lib.rs
+++ b/relay-monitors/src/lib.rs
@@ -115,7 +115,10 @@ pub struct MonitorConfig {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     recovery_threshold: Option<u64>,
 
-    /// Who the owner of the monitor should be.
+    /// Who the owner of the monitor should be. Uses the ActorTuple [0] 
+    /// identifier format.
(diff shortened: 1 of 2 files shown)


relay-monitors/src/lib.rs
```diff
@@ -114,6 +114,10 @@ pub struct MonitorConfig {
     /// How many consecutive OK check-ins it takes to resolve an issue.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     recovery_threshold: Option<u64>,
+
+    /// Who the owner of the monitor should be.
```
evanpurkhiser:
```suggestion
    /// Who the owner of the monitor should be. Uses the ActorTuple [0] 
    /// identifier format.
    ///
    /// [0]: https://github.com/getsentry/sentry/blob/3644f5c4f2a99073bf925181b5237a6e05c1d6c2/src/sentry/utils/actor.py#L17
```

evanpurkhiser (approved):

iker-barriocanal (approved):

PR #3469, opened 2024-04-22
feat(spans): Get metric summaries from OTel spans

jjbayer:
Populate `span._metrics_summary` from `otel_span.attributes[sentry.metrics_summary]`.

#skip-changelog


relay-spans/src/span.rs
```diff
@@ -473,14 +564,81 @@ mod tests {
...
+                    "value": {
+                        "doubleValue": 9001.0
+                    }
+                },
+                {
+                    "key": "sentry.measurements.memory.unit",
+                    "value": {
+                        "stringValue": "byte"
+                    }
+                },
```
jjbayer: I made `.value` and `.unit` separate attributes to reduce nesting. But I'm open to the alternative: making `sentry.measurements.memory` an attribute with a `kvlistValue` inside (like the metrics summary below).

relay-spans/src/span.rs
```diff
@@ -473,14 +564,81 @@ mod tests {
...
+                {
+                    "key": "sentry.metrics_summary.some_metric",
```
jjbayer, line 534: Every metric name in the metrics summary becomes its own attribute to reduce nesting. We cannot flatten this any further because the value of this attribute is an array.

relay-spans/src/span.rs
jjbayer: @cleptric @phacops looking for opinions here.

cleptric: We'll discuss the future of `measurements` this week, with a tendency to make them metrics. We can hold off on this until we made a decision.

relay-spans/src/span.rs
cleptric: This looks good to me. https://opentelemetry.io/docs/specs/otel/common/attribute-type-mapping/ is marked as experimental, though I don't see why any 3rd party vendors would emit Sentry specific attributes 🙂

iker-barriocanal (approved):

PR #3470, opened 2024-04-22
ref(gocd): Add combined trigger for relay pipelines

### LINKED ISSUE #3264: Add an additional deploy that triggers both pipelines
This was requested by the ingest team in order to make regular deploys easier, especially for those not on the ingest team.

IanWoodard:
resolves #3264
Adding a new `deploy-relay` pipeline that triggers both the `deploy-relay-processing` and `deploy-relay-pop` pipelines.

Some key features include:
* **Schedulability Checks:** Before triggering the deploys, the `deploy-relay` pipeline first checks that both `deploy-relay-processing` and `deploy-relay-pop` are schedulable. This means they are not paused or currently running.
* **Sequential Triggering:** While the intention is to trigger both pipelines simultaneously, they are technically triggered sequentially. The `deploy-relay-processing` pipeline is triggered first, followed by `deploy-relay-pop`. However, testing has shown no perceptible difference between the scheduling times between the two pipelines.

The main source of deviation is the duration of the pipelines themselves and GoCD's scheduling of agents behind the scenes to cary out the tasks which are both practically unavoidable.

This change aims to streamline relay deploys by enabling one-click deployments of both processing and pops while still enabling the individual triggering of each pipeline separately.
#skip-changelog

### DIFF
--- a/gocd/pipelines/relay.yaml
+++ b/gocd/pipelines/relay.yaml
@@ -1,0 +1,53 @@
+# More information on gocd-flavor YAML can be found here:
+# - https://github.com/tomzo/gocd-yaml-config-plugin#pipeline
+# - https://www.notion.so/sentry/GoCD-New-Service-Quickstart-6d8db7a6964049b3b0e78b8a4b52e25d
+format_version: 10
+pipelines:
+  deploy-relay:
+    group: relay
+    lock_behavior: unlockWhenFinished
+
+    materials:
+      relay_repo:
+        git: git@github.com:getsentry/relay.git
+        shallow_clone: true
+        branch: master
+        destination: relay
+
+    stages:
+      - checks:
+          approval:
+            type: manual
+          fetch_materials: true
+
+          jobs:
+            checks-pipeline-status:
+              environment_variables:
+                # Required for checkruns.
+                GOCD_ACCESS_TOKEN: "{{SECRET:[devinfra][gocd_access_token]}}"
+              timeout: 300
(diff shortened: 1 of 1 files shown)


olksdr (approved):
let's see how this looks and works 👍

iker-barriocanal (approved):
> Before triggering the deploys, the deploy-relay pipeline first checks that both deploy-relay-processing and deploy-relay-pop are schedulable. This means they are not paused or currently running.

IIUC this can lead to multiple simultaneous deployments of the same pipeline. Is there anything specific we should keep in mind for rollbacks?

IanWoodard:
> > Before triggering the deploys, the deploy-relay pipeline first checks that both deploy-relay-processing and deploy-relay-pop are schedulable. This means they are not paused or currently running.
> 
> IIUC this can lead to multiple simultaneous deployments of the same pipeline. Is there anything specific we should keep in mind for rollbacks?

Multiple pipelines in a group can run simultaneously where `relay-processing` or `relay-pop` are pipeline groups, but only one instance of a given pipeline can run at any given time. This means we can be deploying to both `deploy-relay-processing-s4s` and `deploy-relay-processing-us` at the same time, but not two different instances of `deploy-relay-processing-s4s` at the same time. Moreover, `relay-processing` and `relay-pop` operate independently so they can both run at the same time. None of this impacts our ability to rollback as rolling back cancels all ongoing deploys for the pipeline group before initiating the rollback.

PR #3472, opened 2024-04-23
feat(mobile-ui): Collect frame delay metric

narsaynorath:
Adds a new metric for collecting the `frames.delay` property from a span's `data` object in the Starfish Mobile UI module. 

I've added the tags that I think are necessary to query and filter for this metric, and they are common filters for other metrics in this config. Also noted is that the data is in the `data` object, similar to #3473 and thus I modified measurement extraction to copy this value over to span measurements.


relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics_mobile.snap
```diff
@@ -74,7 +74,44 @@ expression: "(&event.value().unwrap().spans, metrics)"
...
+                ui_component_name: ~,
+                url_scheme: ~,
+                user: ~,
+                replay_id: ~,
+                sdk_name: ~,
+                other: {
+                    "frames.delay": I64(
+                        123,
+                    ),
+                },
```
narsaynorath: Am I doing something incorrect to have this field appear in the `other` property compared to the other keys?

iker-barriocanal (commented):
Similar to https://github.com/getsentry/relay/pull/3473 -- is this doable through measurements?

relay-dynamic-config/src/defaults.rs
```diff
@@ -622,6 +622,38 @@ fn span_metrics(
                     .always(), // already guarded by condition on metric
             ],
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "d:spans/mobile.frames_delay@millisecond".into(),
```
markushi: not sure if relevant, but as per develop spec, the SDKs send the frame delay [in seconds](https://develop.sentry.dev/sdk/performance/span-data-conventions/#mobile) right now

narsaynorath: Thanks for catching that @markushi!

narsaynorath:
Drafting until #3473 is settled, this PR will require changes similarly to that one.

relay-dynamic-config/src/defaults.rs
```diff
@@ -727,6 +727,38 @@ fn span_metrics(
                     .always(),
             ],
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "d:spans/mobile.frames_delay@second".into(),
```
```suggestion
            mri: "g:spans/mobile.frames_delay@second".into(),
```

jjbayer (commented):
Should this be a gauge as well? Rest looks good to me!

relay-dynamic-config/src/defaults.rs
iker-barriocanal: Do gauges work, or do we need distributions?

narsaynorath: Updated to use a gauge :+1:

narsaynorath: Gauges work, I forgot to make that change here. Thanks!

jjbayer (approved):

PR #3473, opened 2024-04-23
feat(mobile-ui): Add slow, frozen, and total frames metrics

narsaynorath:
Adds gauge metrics for slow, frozen, and total frames metrics for the Mobile UI Starfish module. The tags I've chosen are common and already applied to other mobile conditions.

The data source for these metrics come from the `data` property on spans, which I've extracted into `measurements` on spans.


relay-dynamic-config/src/defaults.rs
```diff
@@ -622,6 +622,102 @@ fn span_metrics(
                     .always(), // already guarded by condition on metric
             ],
         },
+        MetricSpec {
+            category: DataCategory::Span,
+            mri: "d:spans/mobile.slow_frames@none".into(),
```
jjbayer: Unless you need percentiles, a gauge metric should be enough here. We're transitioning to gauges for the other metrics as well (see https://github.com/getsentry/relay/pull/3448).
```suggestion
            mri: "g:spans/mobile.slow_frames@none".into(),
```

iker-barriocanal (commented):
Transactions have these values are [measurements](https://github.com/getsentry/relay/blob/03643c31bd113ff1ab52107828abeae75de2b469/relay-event-schema/src/protocol/event.rs#L476) (see [test](https://github.com/getsentry/relay/blob/03643c31bd113ff1ab52107828abeae75de2b469/relay-event-normalization/src/event.rs#L1545)). Spans also have [measurements](https://github.com/getsentry/relay/blob/03643c31bd113ff1ab52107828abeae75de2b469/relay-event-schema/src/protocol/span.rs#L87), are these values available in there? If not, can we reuse them?

relay-event-schema/src/protocol/span.rs
```diff
@@ -299,6 +299,18 @@ pub struct SpanData {
     #[metastructure(field = "sentry.sdk.name")]
     pub sdk_name: Annotated<String>,
 
+    /// Slow Frames
+    #[metastructure(field = "frames.slow")]
+    pub frames_slow: Annotated<Value>,
```
jjbayer, line 308: Do SDKs already send this field? We're trying to [align](https://github.com/getsentry/relay/pull/3457) field names in `span.data` with attribute names in OTel spans, so we should think about what the canonical name of this field should be. That is, If there's an [OTel convention](https://opentelemetry.io/docs/specs/semconv/) for it, use it, otherwise prefix with `sentry.`.

jjbayer (changes requested):

relay-event-schema/src/protocol/span.rs
narsaynorath: SDKs [already send this field](https://develop.sentry.dev/sdk/performance/span-data-conventions/#mobile) and I don't believe there's a corresponding OTel convention so I will prefix it with `sentry.`

narsaynorath:
> Transactions have these values are [measurements](https://github.com/getsentry/relay/blob/03643c31bd113ff1ab52107828abeae75de2b469/relay-event-schema/src/protocol/event.rs#L476) (see [test](https://github.com/getsentry/relay/blob/03643c31bd113ff1ab52107828abeae75de2b469/relay-event-normalization/src/event.rs#L1545)). Spans also have [measurements](https://github.com/getsentry/relay/blob/03643c31bd113ff1ab52107828abeae75de2b469/relay-event-schema/src/protocol/span.rs#L87), are these values available in there? If not, can we reuse them?

@iker-barriocanal We do have them there, but the problem is that these values are associated with specific spans and we can't really propagate them from the transaction down. And in the current implementation they're sent in the span `data` property and not `measurements`.

I see there's a method for extracting values from span data to span measurements [here](https://github.com/getsentry/relay/blob/master/relay-event-normalization/src/normalize/span/tag_extraction.rs#L600), would it be preferable if I pulled these values into that data structure?

jjbayer:
> Spans also have [measurements](https://github.com/getsentry/relay/blob/03643c31bd113ff1ab52107828abeae75de2b469/relay-event-schema/src/protocol/span.rs#L87), are these values available in there? If not, can we reuse them?

> I see there's a method for extracting values from span data to span measurements [here](https://github.com/getsentry/relay/blob/master/relay-event-normalization/src/normalize/span/tag_extraction.rs#L600), would it be preferable if I pulled these values into that data structure?

The difference between transaction measurements and span.measurements is that span measurements are not converted to metrics automatically, so even as a measurement we would need to define a metric in `defaults.rs`. However, I do believe populating the measurements data structure makes sense, because it will put the numbers into the indexed spans dataset as well.

TL;DR yes, I would populate span.measurements as we do for `http.decoded_response_content_length`, etc.

jjbayer (approved):
@narsaynorath could you please update the PR description before merging? Distributions -> gauges, plus maybe mention measurements. Thanks!

PR #3474, opened 2024-04-23
ref(sampling): Remove sampling mode 'total'

jjbayer:
Stop supporting dynamic sampling mode `"total"`, which adjusted for the client sample rate.

* Old downstream Relays will default to `"received"` now that the field is missing (it was not serialized in this case anyway).
* Sentry has already stopped sending this field: https://github.com/getsentry/sentry/blob/1f2bc3e56e111e42b5fa1bb50ac5ae9f393f1a0b/src/sentry/relay/config/__init__.py#L346


relay-server/src/services/processor/dynamic_sampling.rs
```diff
@@ -41,7 +41,7 @@ use crate::utils::{self, sample, ItemAction, SamplingResult};
 /// the main project state.
 ///
 /// If there is no transaction event in the envelope, this function will do nothing.
-pub fn normalize(state: &mut ProcessEnvelopeState<TransactionGroup>) {
+pub fn ensure_dsc(state: &mut ProcessEnvelopeState<TransactionGroup>) {
```
iker-barriocanal, line 44: 👍

iker-barriocanal (approved):
Thanks for the cleanup and fixing the changelog!

CHANGELOG.md
```diff
@@ -2,13 +2,13 @@
...
 ## Unreleased
 
-
 **Bug fixes:**
 
 - Respect country code TLDs when scrubbing span tags. ([#3458](https://github.com/getsentry/relay/pull/3458))
 
 **Features**:
 
+- **Breaking change:** Stop supporting dynamic sampling mode `"total"`, which adjusted for the client sample rate. ([#3474](https://github.com/getsentry/relay/pull/3474))
```
olksdr: Maybe it's better to add additional section for this? I think we did it in the past

relay-server/src/utils/dynamic_sampling.rs
```diff
@@ -76,13 +76,8 @@ pub fn is_trace_fully_sampled(
...
     }
 
-    let adjustment_rate = match root_project_config.mode {
-        SamplingMode::Total => dsc.sample_rate,
-        _ => None,
-    };
-
     // TODO(tor): pass correct now timestamp
-    let evaluator = SamplingEvaluator::new(Utc::now()).adjust_client_sample_rate(adjustment_rate);
+    let evaluator = SamplingEvaluator::new(Utc::now());
```
olksdr, line 81: the TODO on top of this line, can it be removed? Or what does the TODO even mean?

olksdr (approved):

relay-server/src/utils/dynamic_sampling.rs
jjbayer: I noticed it as well, I will look into it.

CHANGELOG.md
jjbayer: I copied this from the 24.4.0 section, but I can also make a separate section if you prefer.

olksdr: In [`23.4.0`](https://github.com/getsentry/relay/blob/master/CHANGELOG.md#2340) it was separate section, and I find it a bit better and easier to spot for out users. What do you think?

jjbayer: Alright, added a new section and updated 24.4.0 section as well.

PR #3476, opened 2024-04-23
feat(spans): Trace-based sampling for standalone spans

jjbayer:
Apply trace-based dynamic sampling rules to spans. This either drops or retains span envelopes based on the dynamic sampling context (DSC) defined in the envelope header.

*Not* in this PR:

- Compute sampling biases from spans on sentry-side (all transaction-based for now)
- Split envelopes by DSC (when spans in a single envelope have different DSCs)
- Allow DSC per item, i.e. in the item header.


relay-sampling/src/evaluation.rs
```diff
@@ -155,21 +155,15 @@ pub struct SamplingEvaluator<'a> {
...
-            reservoir: None,
+            reservoir,
         }
     }
 
-    /// Sets a [`ReservoirEvaluator`].
-    pub fn set_reservoir(mut self, reservoir: &'a ReservoirEvaluator) -> Self {
-        self.reservoir = Some(reservoir);
-        self
-    }
```
jjbayer, line 171: Drive-by refactor: make the reservoir state known at creation time.

relay-server/src/services/processor.rs
```diff
@@ -671,6 +677,13 @@ impl<'a, Group> ProcessEnvelopeState<'a, Group> {
         self.remove_event();
         self.managed_envelope.reject_event(outcome);
     }
+
+    fn sampling_reservoir(&self) -> Option<&ReservoirEvaluator<'a>> {
+        match self.managed_envelope.group() {
+            ProcessingGroup::Transaction => Some(&self.reservoir),
+            _ => None,
+        }
+    }
```
jjbayer: Ideally `reservoir` would be a field on the processing group, and only exist for the transactions group.

relay-server/src/services/processor.rs
```diff
@@ -1300,12 +1317,6 @@ impl EnvelopeProcessorService {
             }
         };
 
-        if let Some(sampling_state) = state.sampling_project_state.clone() {
-            state
-                .envelope_mut()
-                .parametrize_dsc_transaction(&sampling_state.config.tx_name_rules);
-        }
```
jjbayer, line 1310: I moved this to `process_envelope`, because it should be done for any DSC, not just events.

relay-sampling/src/dsc.rs
```diff
@@ -37,7 +37,7 @@ pub struct DynamicSamplingContext {
     /// transaction.
     ///
     /// Set on transaction start, or via `scope.transaction`.
-    #[serde(default)]
+    #[serde(default, alias = "segment_name")]
```
Dav1dde, line 40: The docs also need updating.

Maybe we should the alias and the field name or we have a better field name which describes on a higher level what we store here?

relay-server/src/services/processor/dynamic_sampling.rs
```diff
@@ -62,43 +62,40 @@ pub fn ensure_dsc(state: &mut ProcessEnvelopeState<TransactionGroup>) {
...
-            let sampling_config = match state.project_state.config.sampling {
-                Some(ErrorBoundary::Ok(ref config)) if !config.unsupported() => Some(config),
-                _ => None,
-            };
+    // For transactions, we require transaction metrics to be enabled before sampling.
+    let transaction_metrics_enabled = matches!(&state.project_state.config.transaction_metrics, Some(ErrorBoundary::Ok(c)) if c.is_enabled());
+    let group = state.managed_envelope.group();
+    if matches!(group, ProcessingGroup::Transaction) && !transaction_metrics_enabled {
+        return SamplingResult::Pending;
+    }
```
Dav1dde: We have a sampling trait now, this would be something the trait should encapsulate.

It's now generic what you can pass in here, but we go back from generic to specific again in the implementation. Maybe not having a trait in this case is better, then it won't give the illusion that the function does the right thing if you just implement the trait.

relay-server/src/services/processor.rs
Dav1dde: You can already expose it just via the `Sampling` trait or implement it only for a specific group `impl ProcessEnvelopeState<'_, T> where T: Sampling`.

relay-server/src/services/processor/dynamic_sampling.rs
```diff
@@ -140,10 +137,10 @@ pub fn sample_envelope_items(
     }
 }
 
-/// Computes the sampling decision on the incoming transaction.
+/// Computes the sampling decision on the incoming envelope.
 fn compute_sampling_decision(
     processing_enabled: bool,
-    reservoir: &ReservoirEvaluator,
+    reservoir: Option<&ReservoirEvaluator>,
```
Dav1dde, line 138: I really don't like this `Option<&ReservoirEvaluator>` everywhere, can't we have a `ReservoirEvaluator::disabled()` instead or make it a trait and `Option<ReservoirEvaluator>` implements the trait.

Maybe a future refactor.

relay-server/src/services/processor.rs
Dav1dde: Please add such info as a code comment not just PR comments.

```
// Parameterize DSC for all enevelopes because ...
```

relay-server/src/services/processor/dynamic_sampling.rs
```diff
@@ -62,43 +62,40 @@ pub fn ensure_dsc(state: &mut ProcessEnvelopeState<TransactionGroup>) {
...
+    G: Sampling,
+{
     // Running dynamic sampling involves either:
     // - Tagging whether an incoming error has a sampled trace connected to it.
     // - Computing the actual sampling decision on an incoming transaction.
-    match state.event_type().unwrap_or_default() {
-        EventType::Default | EventType::Error => {
-            tag_error_with_sampling_decision(state, config);
-            SamplingResult::Pending
-        }
```
Dav1dde, line 73: Where did this branch go?

relay-server/src/services/processor/span/processing.rs
```diff
@@ -135,14 +144,18 @@ pub fn process(
...
         }
 
-        // TODO: dynamic sampling
+        if let Some(sampling_outcome) = &sampling_outcome {
+            relay_log::trace!(
+                "Dropping span because of sampling rule {}",
+                sampling_outcome
+            );
+            return ItemAction::Drop(sampling_outcome.clone());
+        }
```
Dav1dde, line 156: Untested:

```suggestion
        if let Some(sampling_outcome) = sampling_outcome {
            relay_log::trace!(
                "Dropping span because of sampling rule {}",
                sampling_outcome
            );
            return ItemAction::Drop(sampling_outcome);
        }
```

You leave the branch here, which means you should be able to take ownership of the value.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -135,14 +144,18 @@ pub fn process(
...
 
         if let Err(e) = scrub(&mut annotated_span, &state.project_state.config) {
             relay_log::error!("failed to scrub span: {e}");
         }
 
-        // TODO: rate limiting
```
Dav1dde, line 147: I assume because we already have rate limiting?

jjbayer: Yes! Forgot to remove this comment in a previous PR.

relay-server/src/services/processor/dynamic_sampling.rs
jjbayer: This is a leftover from when errors and transactions shared the same code path. Nowadays, `process_errors` calls `tag_error_with_sampling_decision` [directly](https://github.com/getsentry/relay/blob/23a48f2c7142ab940d5e6e3d3289c087eba2fcf2/relay-server/src/services/processor.rs#L1447).

relay-server/src/services/processor/dynamic_sampling.rs
jjbayer: I kept the `Option` for now, but refactored the constructors to make them a bit more ergonomic.

relay-server/src/services/processor/span/processing.rs
jjbayer: This returns from `retain_items`, but the sampling outcome is defined outside the loop.

relay-server/src/services/processor/dynamic_sampling.rs
jjbayer: > We have a sampling trait now, this would be something the trait should encapsulate.

Great idea, I moved the `supports_sampling` and `supports_reservoir_sampling` logic to the trait now.

relay-server/src/services/processor.rs
```diff
@@ -167,9 +167,29 @@ macro_rules! processing_group {
 /// Should be used only with groups which are responsible for processing envelopes with events.
 pub trait EventProcessing {}
 
+/// A trait for processing groups that can be dynamically sampled.
+pub trait Sampling {
+    /// Whether dynamic sampling should run under the given project's conditions.
+    fn supports_sampling(project_state: &ProjectState) -> bool;
+
+    fn supports_reservoir_sampling() -> bool;
```
Dav1dde, line 176: Doc comment, could also be an associated const, not sure if that makes sense `Sampling::SUPPORTS_RESERVOIR_SAMPLING`.

Dav1dde (approved):

relay-server/src/services/processor.rs
jjbayer: I kinda like that `supports_sampling` and `supports_reservoir_sampling` are both methods, even though one of them is effectively constant.

PR #3478, opened 2024-04-24
feat(protocol): Add features to client sdks

philipphofmann:
This PR adds features to the ClientSDKInfo. Related develop docs PR: https://github.com/getsentry/develop/pull/1244 and team SDK GH issue: https://github.com/getsentry/team-sdks/issues/83.


philipphofmann:
`Install libcurl-dev` is failing for all CI jobs. Can someone please help me to fix this?


It seems like something is wrong with the signing of these packages.
```
E: Failed to fetch https://packages.microsoft.com/ubuntu/22.04/prod/dists/jammy/InRelease  Clearsigned file isn't valid, got 'NOSPLIT' (does the network require authentication?)
E: The repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease' is no longer signed.
```

iker-barriocanal (commented):
A list can only track enabled features, would we be interested in tracking some other initialization values for future compatibility? For example, one can't tell the difference between a `sample_rate=0.7` vs `sample_rate=1` with just a list. This is solved with an object, in which we forward unknown attributes and thus leave the door open for future extension.

(sorry for getting a bit late here)

philipphofmann:
> A list can only track enabled features, would we be interested in tracking some other initialization values for future compatibility? For example, one can't tell the difference between a `sample_rate=0.7` vs `sample_rate=1` with just a list. This is solved with an object, in which we forward unknown attributes and thus leave the door open for future extension.

The purpose is only to track enabled features. If we want to track something like `sample_rate=0.7` vs `sample_rate=1`, I think we should add another object. This sounds more like SDK config or something. We also plan on indexing this data in Looker, so a list with only strings has a lower cardinality.

iker-barriocanal (approved):

iker-barriocanal:
Sgtm then 👍 

For CI failures, rerunning jobs should be good enough. Seems like something was off in GitHub jobs

PR #3479, opened 2024-04-24
fix(normalization): Don't reset error and default event type on deserialization

iker-barriocanal:
Resolves https://github.com/getsentry/team-ingest/issues/328.

This PR:
- Stops resetting event's `type` during deserialization for error and default events. Normalization still runs and overwrites it, and the field is not used before normalization (`finalize`).
- Adds integration tests to assert the same normalization behavior on the following scenarios: single POP (full normalization configured), single processing (full normalization running by default), and a chain of a POP (full normalization configured) and processing (normalization disabled).
	- I considered adding these tests in `relay-server`, but running the `EnvelopeProcessorService` (to cover all steps) with proper mocked values is a bit of a PITA to write and maintain

The test payloads are adapted (shortened + values changed) from actual events.

### Background

Relay infers the event type in normalization and updates `type` in the payload. However, it resets the field during deserialization as it assumes normalization always runs after it.
Normalizing payloads only in POPs results in setting `type` but processing relays removing it, resulting in ingesting wrongly normalized payloads.

#skip-changelog


relay-server/src/services/processor/event.rs
```diff
@@ -457,7 +457,9 @@ fn event_from_json_payload(
         .map_err(ProcessingError::InvalidJson)?;
 
     if let Some(event_value) = event.value_mut() {
-        event_value.ty.set_value(event_type);
+        if event_type.is_some() {
+            event_value.ty.set_value(event_type);
+        }
```
iker-barriocanal, line 468: `event_type=None` doesn't remove `event.type` anymore. As a result, the function doesn't offer any alternatives to remove `event.type`. This can easily be extended with an enum of three variants; but only two would be used at this point, which makes it look like `Option`, so IMO the extra enum is not worth it for now.

iker-barriocanal, line 468: Removing these lines results in a test failure of `tests/integration/test_normalization.py::test_relay_chain`.

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,89 @@
+import json
+import os
+
+
+def get_test_data(name):
```
iker-barriocanal, line 5: The `-input` and `-output` format aligns with the rest of the fixtures in that directory.

Dav1dde (approved):

jjbayer (changes requested):
Should we check the normalization config to decide whether or not to overwrite the event type? That seems like a more conservative change. I believe we still want to fix the error type for SDKs that send the wrong one, instead of trusting the event type completely.

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,89 @@
...
+    project_id = 42
+    mini_sentry.add_basic_project_config(project_id)
+    relay = relay(
+        upstream=mini_sentry,
+        options={"normalization": {"level": "full"}},
+    )
+
+    relay.send_event(project_id, input)
+    ingested = mini_sentry.captured_events.get(timeout=10).get_event()
+    assert_ingested_event(expected, ingested)
```
jjbayer: mini nit: Having the assertion in the test case itself makes failures easier to debug IMO, e.g. `assert drop_props(expected) == drop_props(ingested)`

relay-server/src/services/processor/event.rs
jjbayer, line 468: Does this comment need updating?

```
// Event items can never include transactions, so retain the event type and let
// inference deal with this during store normalization.
```

relay-server/src/services/processor/event.rs
```diff
@@ -72,9 +72,15 @@ pub fn extract<G: EventProcessing>(
...
+            let (mut annotated_event, len) = event_from_json_payload(item, None)?;
             // Event items can never include transactions, so retain the event type and let
-            // inference deal with this during store normalization.
-            event_from_json_payload(item, None)?
+            // inference deal with this during normalization.
+            if let Some(event) = annotated_event.value_mut() {
+                if matches!(event.ty.value(), Some(EventType::Transaction)) {
+                    event.ty.set_value(None);
+                }
+            }
```
iker-barriocanal: Order of things happening:

1. POPs deserialization.
2. POPs normalization.
3. Processing deserialization.
4. ~Processing normalization~ (we're getting rid of this).

During deserialization:
- If we always reset, step 3 above removes the type of error events (bad).
- If we never reset, step 2 will consider the payload as a transaction and reject it because it's invalid.
- If we only reset for transactions, the first layer of relays fixes the payloads and the subsequent layers receive fixed traffic directly.

jjbayer (approved):
Thanks, LGTM if tests pass!

PR #3480, opened 2024-04-24
perf: move json calls to parameters

anonrig:
Moves `json.loads` and `json.dumps` calls to parameters for experimentation on replacing them with `orjson`.

Ref: https://github.com/getsentry/sentry/issues/68903

### DIFF
--- a/py/CHANGELOG.md
+++ b/py/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add ability to change the json parser. ([#3480](https://github.com/getsentry/relay/pull/3480))
+
 ## 0.8.57
 
 - Add a data category for metirc hours.  [#3384](https://github.com/getsentry/relay/pull/3384)
--- a/py/sentry_relay/processing.py
+++ b/py/sentry_relay/processing.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+from typing import Callable, Any
 
 from sentry_relay._lowlevel import lib, ffi
 from sentry_relay.utils import (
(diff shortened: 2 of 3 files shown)


py/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Move json calls to parameters. ([#3480](https://github.com/getsentry/relay/pull/3480))
```
jan-auer: _nit_: Without reading the code, it's hard to understand what the change here actually is or how I'd use the new parameters.

jan-auer (approved):

py/CHANGELOG.md
anonrig: You're absolutely right. Changed the Changelog.

py/sentry_relay/processing.py
```diff
@@ -292,11 +322,14 @@ def normalize_global_config(config):
...
     """
-    serialized = json.dumps(config)
+    serialized = json_dumps(config)
     normalized = rustcall(lib.normalize_global_config, encode_str(serialized))
     rv = decode_str(normalized, free=True)
     try:
-        return json.loads(rv)
-    except json.JSONDecodeError:
+        return json_loads(rv)
+    except:  # noqa
```
asottile-sentry: this should be `except Exception:` if the intent is to catch all user-space exceptions

as written this catches `KeyboardInterrupt`, `SystemExit`, `SystemError`, etc. which _should_ terminate the process

anonrig: Python is a weird language. Thanks for the review.

PR #3481, opened 2024-04-25
feat(spans): Updates exclusive_time metric to extract known module tags when span is a queue span

edwardgou-sentry:
Updates exclusive_time metric to extract known module tags when span is a queue span.
Also expands queue span detection to pick up all `"queue.task.*","queue.submit.*","queue.task", "queue.submit", "queue.publish", "queue.process",` ops, instead of just celery ops.

#skip-changelog


edwardgou-sentry:
Need this for Queues module since I need to breakdown the `exclusive_time` metric using `transaction`

jjbayer (approved):

iker-barriocanal (approved):

PR #3484, opened 2024-04-25
feat(pii): Consider `Bearer` a password key

iker-barriocanal:
Strings in the format `Bearer <token>` are not PII-scrubbed. This PR considers `Bearer` (case-insensitive) a password to scrub these tokens.

The use case that triggered this is an array of two values, where the first one has been filtered out by a PII password filter but the second one isn't. Although "tuple-like" arrays with a K-V format are a different use case, tokens forwarded individually should also be scrubbed.


jjbayer (approved):
I think "bearer" is unique enough to not cause any false positives.

PR #3487, opened 2024-04-25
feat(spans): Adds message latency to metrics and message size + retry count to indexed spans for queues

edwardgou-sentry:
~Requires: https://github.com/getsentry/relay/pull/3481~ merged

Adds `messaging.message.receive.latency` as a gauge metric for queue spans, tagged by `environment`, `span.op`, and `transaction`.
Adds `messaging.message.body.size` and `messaging.message.retry.count` as measurements to indexed spans.

#skip-changelog


relay-event-schema/src/protocol/span.rs
```diff
@@ -315,6 +315,18 @@ pub struct SpanData {
...
+    #[metastructure(field = "sentry.messaging.message.retry.count")]
+    pub messaging_message_retry_count: Annotated<Value>,
+
+    /// Message Receive Latency
+    #[metastructure(field = "sentry.messaging.message.receive.latency")]
+    pub messaging_message_receive_latency: Annotated<Value>,
+
+    /// Message Body Size
+    #[metastructure(field = "sentry.messaging.message.body.size")]
+    pub messaging_message_body_size: Annotated<Value>,
```
jjbayer: Do SDKs already send these fields in `span.data`? If not, there might be better ways to transmit them: Either as part of `span.measurements` (then you would not require glue code in tag extraction), or directly as metrics. See comment [here](https://github.com/getsentry/relay/pull/3469#discussion_r1574636848).

If SDKs already send these fields, we need to make sure the field names match the actual data. In the test input it's `messaging.message.receive.latency` (without the `sentry.` prefix). There has been some back and forth about the `sentry.` prefix but I think the consensus is we will only use it for very sentry-specific fields (see Slack convo [here](https://sentry.slack.com/archives/C019637C760/p1714054206063119?thread_ts=1714052567.842039&cid=C019637C760).

edwardgou-sentry: These are new fields being added to the SDKs! I'll double check with the SDK team to see what they think

antonpirker: About sending it as metrics: Should the SDK start sending metrics with an (maybe auto enabled) integration? I think this could make users mad because metrics have to be paid separately. What is our consensus about this?

Another question: should we in general add "string" data in spans as `span.data` and "number" data as `measurements`?

edwardgou-sentry: > About sending it as metrics: Should the SDK start sending metrics with an (maybe auto enabled) integration? I think this could make users mad because metrics have to be paid separately. What is our consensus about this?

We don't want to launch the queues module with an added performance cost at the moment, so I think we'll have to rule out sending metrics if extra cost is involved. (cc @bcoe)

relay-dynamic-config/src/defaults.rs
```diff
@@ -766,6 +766,57 @@ fn span_metrics(transaction_extraction_enabled: bool) -> impl IntoIterator<Item
...
+            tags: vec![
+                Tag::with_key("environment")
+                    .from_field("span.sentry_tags.environment")
+                    .always(),
+                Tag::with_key("span.op")
+                    .from_field("span.sentry_tags.op")
+                    .always(),
+                Tag::with_key("transaction")
+                    .from_field("span.sentry_tags.transaction")
+                    .always(),
```
jjbayer: Just asking for good measure: Are we sure we need the transaction tag on all three of these metrics?

jjbayer (approved):

relay-dynamic-config/src/defaults.rs
edwardgou-sentry: `messaging.message.receive.latency` requires `transaction` since we will break down by transaction in the product workflow.

I revisited the mocks, and realized that `messaging.message.body.size` and `messaging.message.retry.count` are actually not used in any aggregate views, so we probably don't need them as metrics. I've removed them from the metrics extraction, but I am keeping them in the tag extraction since we still need them for indexed spans.

PR #3488, opened 2024-04-26
feat(metrics): Add received_at timestamp in bucket metadata

iambriccardo:
This PR adds a new `received_at` field to the `BucketMetadata` struct. The goal of this field is to eventually propagate it to Kafka to measure the end-to-end latency of our metrics pipeline. For this reason, the `received_at` measurement must happen in the outermost internal Relay. 

The field `received_at` will be set/overridden by the outermost Relay via the option `keep_metadata` which when set to `false` will tell Relay to override the incoming metadata.

The merge operation on two `received_at` timestamps is defined as `min(r1, r2)` if both timestamps are non-null, otherwise the first non-null value will be taken. If both values are null, the merge result will still yield null.

Closes: https://github.com/getsentry/team-ingest/issues/265


relay-config/src/config.rs
```diff
@@ -524,13 +524,17 @@ struct SentryMetrics {
...
+    /// receive time of the instance.
+    pub override_received_at_metadata: bool,
 }
 
 impl Default for SentryMetrics {
     fn default() -> Self {
         Self {
             meta_locations_expiry: 15 * 24 * 60 * 60,
             meta_locations_max: 5,
+            override_received_at_metadata: false,
```
iambriccardo: By default we do not want to override the metadata to avoid unexpected behaviors due to the missing config param.

relay-config/src/config.rs
```diff
@@ -524,13 +524,17 @@ struct SentryMetrics {
     ///
     /// Defaults to 5.
     pub meta_locations_max: usize,
+    /// Whether to override the [`received_at`] field in the [`BucketMetadata`] with the current
+    /// receive time of the instance.
+    pub override_received_at_metadata: bool,
```
iambriccardo: In production, we would like to set this option to true only for pop relays, since we want to have the `received_at` metadata field set to the outermost relay timestamp that receives a given bucket.

relay-server/src/services/store.rs
```diff
@@ -507,6 +507,7 @@ impl StoreService {
             BucketViewValue::Gauge(g) => MetricValue::Gauge(g),
         };
 
+        // TODO: propagate the `received_at` field upstream.
```
iambriccardo, line 510: In a follow up PR we want to propagate to kafka the message with the received at field.

relay-server/src/metrics_extraction/generic.rs
```diff
@@ -56,13 +56,20 @@ where
             continue;
         };
 
+        // For extracted metrics we assume the `received_at` timestamp is equivalent to the time
+        // in which the metric is extracted.
+        #[cfg(not(test))]
```
iambriccardo: Instead of mocking time, we generate it differently if we are in test, in order to make the unit tests deterministic.

relay-server/src/services/processor.rs
```diff
@@ -1852,7 +1859,11 @@ impl EnvelopeProcessorService {
         for bucket in &mut buckets {
             clock_drift_processor.process_timestamp(&mut bucket.timestamp);
             if !keep_metadata {
-                bucket.metadata = BucketMetadata::new();
+                bucket.metadata = BucketMetadata::default();
+            }
+
+            if override_received_at_metadata {
```
iambriccardo: We purposefully override the metadata after the decision to keep it has been made.

iambriccardo: This piece of code could be optimized to avoid overwriting a field but I felt it's more comprehensible like this.

relay-server/src/services/processor.rs
```diff
@@ -3366,4 +3387,158 @@ mod tests {
...
+                assert_eq!(buckets.len(), 1);
+                assert_eq!(buckets[0].metadata.received_at, expected_received_at);
+            } else {
+                panic!();
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_process_batched_metrics_bucket_metadata() {
```
iambriccardo, line 3471: Will add integration tests in a follow up PR where we are also going to write the data to kafka.

relay-config/src/config.rs
Dav1dde: We already have that handling in the processor as `keep_metadata` for example in `handle_process_metrics` which should already do the right thing.

relay-metrics/src/bucket.rs
```diff
@@ -748,33 +749,55 @@ pub struct BucketMetadata {
...
     pub fn merge(&mut self, other: Self) {
         self.merges = self.merges.saturating_add(other.merges.get());
+        self.received_at = match (self.received_at, other.received_at) {
+            (Some(received_at), None) => Some(received_at),
+            (None, Some(received_at)) => Some(received_at),
+            (Some(left_received_at), Some(right_received_at)) => {
+                Some(min(left_received_at, right_received_at))
+            }
+            (None, None) => None,
+        };
```
Dav1dde: This could be `a.or(b).min(b.or(a))` not sure if that's better ...

relay-server/src/metrics_extraction/generic.rs
```diff
@@ -56,13 +56,20 @@ where
...
+        let received_at = UnixTimestamp::now();
+        #[cfg(test)]
+        let received_at = UnixTimestamp::from_secs(0);
```
Dav1dde:
```suggestion
        let received_at = if cfg!(not(test)) { UnixTimestamp::now() } else { UnixTimestamp::from_secs(0) };
```

relay-server/src/metrics_extraction/generic.rs
```diff
@@ -56,13 +56,20 @@ where
             continue;
         };
 
+        // For extracted metrics we assume the `received_at` timestamp is equivalent to the time
+        // in which the metric is extracted.
+        #[cfg(not(test))]
+        let received_at = UnixTimestamp::now();
```
Dav1dde: We should do this outside of the loop at the beginning of the function.

relay-metrics/src/bucket.rs
```diff
@@ -748,33 +749,55 @@ pub struct BucketMetadata {
...
 
 impl Default for BucketMetadata {
     fn default() -> Self {
-        Self::new()
+        Self {
+            merges: NonZeroU32::MIN,
+            received_at: None,
+        }
     }
 }
```
Dav1dde, line 798: I don't think we need default here, let's move it to a separate constructor, maybe empty if we need it somewhere. But generally there shouldn't be any instances where we create a bucket without timestamp.

relay-server/src/metrics_extraction/sessions/types.rs
```diff
@@ -119,13 +119,20 @@ impl IntoMetric for SessionMetric {
             unit: MetricUnit::None,
         };
 
+        // For extracted metrics we assume the `received_at` timestamp is equivalent to the time
+        // in which the metric is extracted.
+        #[cfg(not(test))]
+        let received_at = UnixTimestamp::now();
+        #[cfg(test)]
+        let received_at = UnixTimestamp::from_secs(0);
```
Dav1dde: Wonder if we should move this into the metadata instead.

relay-server/src/services/processor.rs
```diff
@@ -1852,7 +1859,11 @@ impl EnvelopeProcessorService {
         for bucket in &mut buckets {
             clock_drift_processor.process_timestamp(&mut bucket.timestamp);
             if !keep_metadata {
-                bucket.metadata = BucketMetadata::new();
+                bucket.metadata = BucketMetadata::default();
```
Dav1dde: `keep_metadata` is already the signal you need instead of `override_received_at_metadata`.

relay-server/src/services/processor.rs
```diff
@@ -1885,6 +1899,7 @@ impl EnvelopeProcessorService {
         let buckets = match serde_json::from_slice(&payload) {
             Ok(Wrapper { buckets }) => buckets,
             Err(error) => {
+                println!("{:?}", error);
```
Dav1dde: 👀

relay-server/src/services/store.rs
Dav1dde: Let's put the gh issue here

relay-server/src/services/processor.rs
```diff
@@ -3366,4 +3387,158 @@ mod tests {
...
+            processor.handle_process_metrics(&mut token, message);
+
+            let value = project_cache_rx.recv().await.unwrap();
+            if let ProjectCache::MergeBuckets(merge_buckets) = value {
+                let buckets = merge_buckets.buckets();
+                assert_eq!(buckets.len(), 1);
+                assert_eq!(buckets[0].metadata.received_at, expected_received_at);
+            } else {
+                panic!();
+            }
```
```suggestion
           let ProjectCache::MergeBuckets(merge_buckets) = value else { panic!() };
           let buckets = merge_buckets.buckets();
           assert_eq!(buckets.len(), 1);
           assert_eq!(buckets[0].metadata.received_at, expected_received_at);
```

relay-server/src/metrics_extraction/generic.rs
iambriccardo: Good catch!

relay-server/src/metrics_extraction/generic.rs
iambriccardo: Good catch!

relay-metrics/src/bucket.rs
```diff
@@ -1275,7 +1300,9 @@ mod tests {
           }
         ]"#;
 
-        let buckets = serde_json::from_str::<Vec<Bucket>>(json).unwrap();
+        let mut buckets = serde_json::from_str::<Vec<Bucket>>(json).unwrap();
+        buckets[0].metadata = BucketMetadata::new(UnixTimestamp::from_secs(1615889440));
```
Dav1dde: You can just put this into the json, right?

relay-server/src/services/processor.rs
```diff
@@ -1874,6 +1890,8 @@ impl EnvelopeProcessorService {
         } = message;
 
         let received = relay_common::time::instant_to_date_time(start_time);
+        let received_timestamp = UnixTimestamp::from_secs(received.timestamp() as u64);
```
```suggestion
        let received_timestamp = UnixTimestamp::from_instant(start_time);
```

There is another usage I cannot comment on (thanks github)

relay-server/src/services/processor.rs
```diff
@@ -3366,4 +3384,164 @@ mod tests {
...
+        let mut token = Cogs::noop().timed(ResourceId::Relay, AppFeature::Unattributed);
+        let project_key = ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap();
+        let start_time = Instant::now();
+        let config = {
+            let config_json = serde_json::json!({
+                "processing": {
+                    "enabled": true,
+                    "kafka_config": [],
+                }
+            });
```
Dav1dde: Is `Config::default()` not good enough here?

relay-server/src/services/processor.rs
```diff
@@ -3366,4 +3384,164 @@ mod tests {
...
+        let mut token = Cogs::noop().timed(ResourceId::Relay, AppFeature::Unattributed);
+        let start_time = Instant::now();
+        let config = {
+            let config_json = serde_json::json!({
+                "processing": {
+                    "enabled": true,
+                    "kafka_config": [],
+                }
+            });
+            Config::from_json_value(config_json).unwrap()
```
Dav1dde: Same here I think this is not needed

tests/integration/test_metrics.py
```diff
@@ -807,12 +820,7 @@ def test_transaction_metrics(
         transactions_consumer.assert_empty()
     else:
         event, _ = transactions_consumer.get_event()
-        if with_external_relay:
-            # there is some rounding error while serializing/deserializing
-            # timestamps... haven't investigated too closely
-            span_time = 9.910107
-        else:
-            span_time = 9.910106
+        span_time = 9.910106
```
iambriccardo: Tests seem to succeed with the actual value `9.910106`.

tests/integration/test_metrics.py
```diff
@@ -43,13 +43,24 @@ def metrics_by_name(metrics_consumer, count, timeout=None):
...
         metric, metric_headers = metrics_consumer.get_metric(timeout)
+        metric = metrics_without_keys([metric], keys={"metadata"})[0]
         metrics[metric["name"]] = metric
         metrics["headers"][metric["name"]] = metric_headers
 
     metrics_consumer.assert_empty()
     return metrics
 
 
+def metrics_without_keys(received_metrics, keys):
```
iambriccardo: Decided to opt for a generic solution that can remove any key.

relay-server/src/testutils.rs
```diff
@@ -155,6 +155,34 @@ pub fn create_test_processor(config: Config) -> EnvelopeProcessorService {
     )
 }
 
+pub fn create_test_processor_with_addrs(
```
Dav1dde, line 158: `create_test_processor` should forward to this

Dav1dde (approved):

PR #3489, opened 2024-04-26
feat(relay-cabi): Expose cardinality limit validation

ArthurKnaus:
Expose `relay_validate_cardinality_limit_config` to enable validation in `sentry` repo.

Relates to https://github.com/getsentry/sentry/issues/69750

#skip-changelog


py/tests/test_processing.py
```diff
@@ -325,6 +325,22 @@ def test_validate_project_config():
...
+        "limit": 1000,
+        "namespace": "custom",
+        "scope": "name",
+    }
+    # Does not raise:
+    sentry_relay.validate_cardinality_limit_config(json.dumps(config), strict=True)
+    config["foobar"] = "foo"
+    with pytest.raises(ValueError) as e:
+        sentry_relay.validate_cardinality_limit_config(json.dumps(config), strict=True)
+    assert str(e.value) == 'json atom at path ".foobar" is missing from rhs'
```
iker-barriocanal: Validating the config in the C-ABI and returning that error in the librelay produces harder-to-debug messages: "what is `rhs`?" "What was the value my config was compared against?". Instead, letting the C-ABI handle the conversion/normalization and comparing values in Python produces errors with actual diffs, making it easier to debug.

In short, instead of following the project config validation approach, we should follow the global config one. The code is very similar, and it's just below the diff in the PR in all updated files.

py/tests/test_processing.py
```diff
@@ -325,6 +325,22 @@ def test_validate_project_config():
     assert str(e.value) == 'json atom at path ".foobar" is missing from rhs'
 
 
+def test_validate_cardinality_limit_config():
```
iker-barriocanal: This test validates a valid and an invalid config. Let's keep them in separate unit tests.

iker-barriocanal (commented):
Thanks for contributing!

py/tests/test_processing.py
ArthurKnaus: Yeah, the whole things was copy paste from `validate_project_config` 🙈

ArthurKnaus:
@iker-barriocanal thx for the review, I changed the implementation to the normalization approach 👍 
btw how do I resolve the Changelog check? I already added an entry in `py/CHANGELOG.md` 🤔

py/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Expose cardinality limit validation ([3489](https://github.com/getsentry/relay/pull/3489))
```
iker-barriocanal:
```suggestion
- Expose cardinality limit config normalization. ([3489](https://github.com/getsentry/relay/pull/3489))
```

relay-cabi/src/processing.rs
```diff
@@ -487,6 +488,19 @@ pub unsafe extern "C" fn relay_validate_project_config(
     }
 }
 
+/// Validate cardinality limit config.
+///
+/// If `strict` is true, checks for unknown fields in the input.
```
iker-barriocanal:
```suggestion
/// Normalize a cardinality limit config.
```

iker-barriocanal (approved):
LGTM, please see comments before merging.

A `#skip-changelog` in the PR description skips the changelog checks. This is ok as there's already an entry in py/changelog and that's all we need.

PR #3490, opened 2024-04-26
feat(metrics): Global extraction config

jjbayer:
Both metrics conditional tagging and span metrics extraction duplicate a lot of configuration into every single project config, leading to unnecessary memory consumption.

Add a metrics extraction section to global config to deduplicate. To make it possible for projects to opt in or out of certain groups of metrics (e.g. span metrics), divide the section into multiple groups of metrics, e.g.

```js
{
  "metricExtraction": {
    "templates": {
      "spans": {
        "isEnabled": false, // not enabled by default
        "metrics": [],
        "tags": []
      },
      "histogram_outliers": {
        "isEnabled": true, // enabled by default
        "metrics": [],
        "tags": []
      }
    }
  }
}
```

## Rollout plan

1. Allow metrics extraction to be defined globally (this PR) and bump max. supported extraction version.
1. In sentry, bump version to stop extraction in old relays.
1. In Relay, move static span metrics from project config to global config (e.g. section `spans_hardcoded`) and leave only the `isEnabled` flags in project config.

ref: https://github.com/getsentry/team-ingest/issues/331


relay-dynamic-config/src/global.rs
```diff
@@ -38,6 +38,16 @@ pub struct GlobalConfig {
         skip_serializing_if = "is_default"
     )]
     pub options: Options,
+
+    /// Configuration for global metrics extraction rules.
+    ///
+    /// These are merged with rules in project configs before
+    /// applying.
+    #[serde(
+        deserialize_with = "default_on_error",
```
jjbayer: This will log an error when serialization fails.

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +170,112 @@ impl TransactionMetricsConfig {
     }
 }
 
+/// Combined view of global and project-specific metrics extraction configs.
+#[derive(Debug, Clone)]
+pub struct CombinedMetricExtractionConfig<'a> {
+    global: &'a MetricExtractionTemplates,
+    project: &'a MetricExtractionConfig,
+}
```
jjbayer: This is the object we now pass into `extract_metrics`.

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +170,112 @@ impl TransactionMetricsConfig {
...
+
+    /// Returns `true` if the continaed templates are empty.
+    pub fn is_empty(&self) -> bool {
+        self.templates.is_empty()
+    }
+}
+
+/// Set of metrics & tags that can be enabled or disabled as a group.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MetricExtractionTemplate {
```
jjbayer: Not sure if we should call this "templates", "sets", or something else. Bikeshedding welcome!

relay-server/src/metrics_extraction/generic.rs
```diff
@@ -23,22 +23,28 @@ pub trait Extractable: Getter {
...
-    for metric_spec in &config.metrics {
+    for metric_spec in config.metrics() {
         if metric_spec.category != instance.category() {
             continue;
         }
 
+        let is_new = mris_seen.insert(metric_spec.mri.as_str());
+        if !is_new {
+            continue;
+        }
```
jjbayer: I added this logic because with multiple sources of MRI definitions, I expect accidental duplicates to become more probable. @jan-auer let me know if this goes against the original design.

jjbayer: We could move this check to right before `metrics.push(Bucket` to allow multiple non-overlapping definitions (first match wins, like with tags).

relay-dynamic-config/src/metrics.rs
```diff
@@ -114,7 +114,8 @@ pub struct CustomMeasurementConfig {
 ///      - Emit a `usage` metric and use it for rate limiting.
 ///      - Delay metrics extraction for indexed transactions.
 ///  - 4: Adds support for `RuleConfigs` with string comparisons.
-const TRANSACTION_EXTRACT_VERSION: u16 = 4;
+///  - 5: No change, bumped together with [`MetricExtractionConfig::MAX_SUPPORTED_VERSION`].
+const TRANSACTION_EXTRACT_VERSION: u16 = 5;
```
jjbayer: If we do not bump this version as well, external Relays will still perform metrics extraction from transactions, but skip the span metrics. See https://github.com/getsentry/relay/blob/89de2d1ee6819f26fee76df204b638159e5b42a3/relay-server/src/services/processor.rs#L1159-L1166

relay-dynamic-config/src/metrics.rs
```diff
@@ -214,24 +314,25 @@ impl MetricExtractionConfig {
     /// The latest version for this config struct.
     ///
     /// This is the maximum version supported by this Relay instance.
-    pub const VERSION: u16 = 2;
+    pub const MAX_SUPPORTED_VERSION: u16 = 3;
```
jjbayer, line 327: I renamed this constant because I always get confused by the `self.version <= Self::VERSION` check.

relay-server/src/metrics_extraction/generic.rs
```diff
@@ -67,16 +67,19 @@ where
     }
 
     // TODO: Inline this again once transaction metric extraction has been moved to generic metrics.
-    tmp_apply_tags(&mut metrics, instance, &config.tags);
+    tmp_apply_tags(&mut metrics, instance, config.tags());
```
iker-barriocanal, line 73: Out of scope of this PR, but we should promote this temporary function 😄

relay-server/src/metrics_extraction/generic.rs
```diff
@@ -23,7 +23,7 @@ pub trait Extractable: Getter {
 /// The instance must have a valid timestamp; if the timestamp is missing or invalid, no metrics are
 /// extracted. Timestamp and clock drift correction should occur before metrics extraction to ensure
 /// valid timestamps.
-pub fn extract_metrics<T>(instance: &T, config: &MetricExtractionConfig) -> Vec<Bucket>
+pub fn extract_metrics<T>(instance: &T, config: CombinedMetricExtractionConfig) -> Vec<Bucket>
```
iker-barriocanal: Does taking ownership of `CombinedMetricExtractionConfig` make the code easier, vs the reference? It doesn't matter today, but it may in the future if we extend the object and make clones expensive.

relay-dynamic-config/src/global.rs
iker-barriocanal: What should Relay do in that case? Falling back to default will result in not extracting metrics. Should we forward items and extract items in the next Relay?

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,113 @@ impl TransactionMetricsConfig {
...
+            let is_enabled = is_enabled_by_override.unwrap_or(template.is_enabled);
+
+            is_enabled.then_some(template)
+        })
+    }
+}
+
+impl<'a> From<&'a MetricExtractionConfig> for CombinedMetricExtractionConfig<'a> {
+    fn from(value: &'a MetricExtractionConfig) -> Self {
+        Self::new(MetricExtractionTemplates::EMPTY, value)
```
iker-barriocanal: Should this be documented to note only the project is set? Reading a from implementation may be confusing.

relay-dynamic-config/src/metrics.rs
iker-barriocanal: I don't like "templates" because it kind of implies it needs something else to be "complete" and it's not useful on its own, which is not the case. I can't think of a name I really like, but "sets", "specs" or "definitions" sound better (or "metric tagging config"?).

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,113 @@ impl TransactionMetricsConfig {
...
+    /// A list of metric specifications to extract.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub metrics: Vec<MetricSpec>,
+
+    /// A list of tags to add to previously extracted metrics.
+    ///
+    /// These tags add further tags to a range of metrics. If some metrics already have a matching
+    /// tag extracted, the existing tag is left unchanged.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub tags: Vec<TagMapping>,
```
iker-barriocanal, line 273: `MetricSpec` already has a `tags` property, can we reuse that one?

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,113 @@ impl TransactionMetricsConfig {
...
+#[serde(rename_all = "camelCase")]
+pub struct MetricExtractionTemplate {
+    /// Whether the set is enabled by default.
+    ///
+    /// Project configs can overwrite this flag to opt-in or out of a set.
+    pub is_enabled: bool,
+
+    /// A list of metric specifications to extract.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub metrics: Vec<MetricSpec>,
```
iker-barriocanal, line 266: Why do we need `category` here? Should it be only applied to the given category?

relay-dynamic-config/src/metrics.rs
```diff
@@ -545,4 +657,99 @@ mod tests {
...
+                        }
+                    ]
+                }
+            }
+        }))
+        .unwrap()
+    }
+
+    #[test]
+    fn metrics_extraction_global_defaults() {
```
iker-barriocanal: Can we add tests like these for merging tags?

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,113 @@ impl TransactionMetricsConfig {
...
+            .flat_map(|template| template.tags.iter());
+
+        project.chain(enabled_global)
+    }
+
+    fn enabled_templates(&self) -> impl Iterator<Item = &MetricExtractionTemplate> {
+        self.global.templates.iter().filter_map(|(key, template)| {
+            let is_enabled_by_override =
+                self.project.global_templates.get(key).map(|c| c.is_enabled);
+            let is_enabled = is_enabled_by_override.unwrap_or(template.is_enabled);
```
iker-barriocanal, line 218: What should Relay do when two metrics have different tags in different templates? For example, the project config sets `key1:value1` in `template1`, but the global config sets `key2:value2` in `template2`.

relay-server/src/metrics_extraction/generic.rs
jjbayer: Solved, we need to support multiple occurrences (e.g. on-demand metrics).

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,113 @@ impl TransactionMetricsConfig {
     }
 }
 
+/// Combined view of global and project-specific metrics extraction configs.
+#[derive(Debug, Clone)]
+pub struct CombinedMetricExtractionConfig<'a> {
```
Dav1dde, line 176: Technically, for historic reasons we would call this `DynamicMetricExtractionConfig`, although this is now the 3rd time where an independent person used `Combined` instead of `Dynamic` (me, Iker, you) and I think it's finally time to retire `Dynamic` and stick to `Combined`.

It started with `DynamicMeasurementsConfig`, then came `DynamicQuotas` and afterwards `DynamicGenericFilters*`.

relay-dynamic-config/src/metrics.rs
Dav1dde: I like `Template` but `group` could also work. You can enable/disable certain groups. I think we also use the groups naming already throughout Relay.

relay-dynamic-config/src/metrics.rs
Dav1dde: Should we also rename this to include `max supported` somewhere?

```suggestion
const TRANSACTION_EXTRACTION_MAX_SUPPORTED_VERSION: u16 = 5;
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,113 @@ impl TransactionMetricsConfig {
...
+    }
+
+    /// Returns an iterator of metric specs.
+    pub fn metrics(&self) -> impl Iterator<Item = &MetricSpec> {
+        let project = self.project.metrics.iter();
+        let enabled_global = self
+            .enabled_templates()
+            .flat_map(|template| template.metrics.iter());
+
+        project.chain(enabled_global)
```
Dav1dde, line 202: Is here a problem when global and project configure the same `MetricSpec`, e.g. we'd be extracting the same metric twice?

relay-server/src/metrics_extraction/event.rs
```diff
@@ -48,11 +48,11 @@ impl Extractable for Span {
 pub fn extract_metrics(
     event: &Event,
     spans_extracted: bool,
-    config: &MetricExtractionConfig,
+    config: CombinedMetricExtractionConfig,
```
Dav1dde: Baby nit: I like these "generic" lifetimes to be explicit.

```suggestion
    config: CombinedMetricExtractionConfig<'_>,
```

relay-server/src/metrics_extraction/event.rs
```diff
@@ -65,19 +65,19 @@ pub fn extract_metrics(
 
 fn extract_span_metrics_for_event(
     event: &Event,
-    config: &MetricExtractionConfig,
+    config: CombinedMetricExtractionConfig,
```
Dav1dde:
```suggestion
    config: CombinedMetricExtractionConfig<'_>,
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -242,6 +344,16 @@ impl MetricExtractionConfig {
...
 
+/// Configures metrics extraction templates.
+///
+/// Project configs can enable or disable globally defined templates.
+#[derive(Clone, Default, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MetricExtractionTemplateOverride {
+    /// `true` if a template should be enabled.
+    pub is_enabled: bool,
+}
```
Dav1dde: Do we want to support more complicated overriding in the future?

While I generally like the idea it feels like we're building custom override logic for every sub-feature independently (compare generic inbound filters).

Apart from this, even if we only have it for enabled/disabled now, it's a good idea to already have the object in place and it works nicely with default enabled/disabled templates 👍 .

relay-server/src/metrics_extraction/event.rs
```diff
@@ -48,11 +48,11 @@ impl Extractable for Span {
 pub fn extract_metrics(
     event: &Event,
     spans_extracted: bool,
-    config: &MetricExtractionConfig,
+    config: CombinedMetricExtractionConfig,
     max_tag_value_size: usize,
     span_extraction_sample_rate: Option<f32>,
 ) -> Vec<Bucket> {
-    let mut metrics = generic::extract_metrics(event, config);
+    let mut metrics = generic::extract_metrics(event, config.clone());
```
Dav1dde:
```suggestion
    let mut metrics = generic::extract_metrics(event, config);
```

`config` should be Copy

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,113 @@ impl TransactionMetricsConfig {
     }
 }
 
+/// Combined view of global and project-specific metrics extraction configs.
+#[derive(Debug, Clone)]
```
Dav1dde, line 175:
```suggestion
#[derive(Debug, Clone, Copy)]
```

Dav1dde (commented):
I assume we can use the global ones for global stuff and need the project specific extraction for project specific alerts etc.?

relay-dynamic-config/src/metrics.rs
jjbayer: `group` it is! Against `template` you could argue that it's not incomplete, against `set` you could argue that it's not a set in the mathematical sense. "group" is neutral (even though it's not a group in the mathematical sense either :D).

relay-dynamic-config/src/metrics.rs
jjbayer: I initially added a check for this but removed it because we actually want to be able to extract MRIs twice. If you have an idea for a smarter check (e.g. equality check), let me know.

https://github.com/getsentry/relay/pull/3490#discussion_r1581090210

relay-dynamic-config/src/metrics.rs
jjbayer: `extract_metrics` only sets a tag value once. If there's already a value for a key, the tag is skipped.

relay-dynamic-config/src/metrics.rs
jjbayer: I made this `#[cfg(test)]` now.

relay-dynamic-config/src/metrics.rs
jjbayer: I initially had a separate struct with `metrics` and `tags` which I flattened into `MetricExtractionTemplate` and `MetricExtractionConfig`, but I reverted it because it made the code less readable.

relay-dynamic-config/src/metrics.rs
jjbayer: > Do we want to support more complicated overriding in the future?

We might!

> While I generally like the idea it feels like we're building custom override logic for every sub-feature independently (compare generic inbound filters).

We might be able to make that code generic (anything that implements an `is_enabled()` trait method can be merged), but generic inbound filters have a bit more merging logic IIRC.

relay-dynamic-config/src/global.rs
Dav1dde: Mh maybe we should do an explicit error boundary and skip extraction in an error case, like we do for inbound filters. This should prevent and futureproof all possible managed/self hosted Relay issues.

relay-server/src/metrics_extraction/event.rs
jjbayer: I agree that is should be either a reference or `Copy`. Went for a reference now.

relay-dynamic-config/src/metrics.rs
jjbayer: Never mind, making it `#[cfg(test)]` does not work because it's used in a different crate. But I added some docs to clarify.

relay-dynamic-config/src/metrics.rs
jjbayer: IMO it would not add much value, the group can be either enabled or disabled, so you either get all metrics and all tags, or none.

relay-dynamic-config/src/metrics.rs
iker-barriocanal: (`category` is a property of `MetricSpec`)

relay-dynamic-config/src/metrics.rs
iker-barriocanal: IMO this is worth documenting, it's easy to assume that tags are incremental.

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,114 @@ impl TransactionMetricsConfig {
...
+}
+
+impl<'a> From<&'a MetricExtractionConfig> for CombinedMetricExtractionConfig<'a> {
+    /// Creates a combined config with an empty global component. Used in tests.
+    fn from(value: &'a MetricExtractionConfig) -> Self {
+        Self::new(MetricExtractionGroups::EMPTY, value)
+    }
+}
+
+/// global groups for metric extraction.
```
iker-barriocanal:
```suggestion
/// Global groups for metric extraction.
```

relay-dynamic-config/src/metrics.rs
```diff
@@ -170,13 +171,114 @@ impl TransactionMetricsConfig {
...
+    pub groups: BTreeMap<String, MetricExtractionGroup>,
+}
+
+impl MetricExtractionGroups {
+    /// Empty config, used in tests.
+    const EMPTY: &'static Self = &Self {
+        groups: BTreeMap::new(),
+    };
+
+    /// Returns `true` if the continaed templates are empty.
```
iker-barriocanal:
```suggestion
    /// Returns `true` if the contained groups are empty.
```

relay-dynamic-config/src/metrics.rs
iker-barriocanal: I guess we could write custom (de)serializers and save some memory, but probably not worth the memory benefit.

iker-barriocanal (commented):
Overall LGTM, leaving some non-blocking questions:
- Should internal relays log an error when an invalid config is received from upstream?
- Should processing relays always attempt to extract metrics, with whatever config is available?

On a side note, I'd leave the `DynamicQuota` -> `CombinedQuota` and similar changes to a separate PR.

relay-dynamic-config/src/metrics.rs
jjbayer: Yes, see https://github.com/getsentry/relay/blob/e827dc7b8f9f9b98d3b1e031ee20bc0a1e41b14f/relay-server/src/metrics_extraction/generic.rs#L37-L39

relay-dynamic-config/src/metrics.rs
jjbayer: I went ahead and renamed `DynamicMeasurementsConfig` to `CombinedMeasurementsConfig`, and `DynamicQuotas` to `CombinedQuotas`. IMO it's a more descriptive name.

relay-dynamic-config/src/metrics.rs
jjbayer: 👍, I added a line to the documentation of `extract_metrics`.

jjbayer (commented):
@iker-barriocanal I special-cased processing relays now.

relay-dynamic-config/src/metrics.rs
Dav1dde: It is, thanks!

relay-dynamic-config/src/metrics.rs
Dav1dde: @jjbayer you don't like the type to be Copy?

relay-server/src/metrics_extraction/event.rs
```diff
@@ -1119,7 +1119,8 @@ mod tests {
         project.sanitize();
 
         let config = project.metric_extraction.ok().unwrap();
-        let metrics = extract_metrics(event.value().unwrap(), false, &config, 200, None);
+        let combined_config = (&config).into();
```
Dav1dde, line 1122: Not sure if that's better:

```suggestion
        let combined_config = CombinedMetricExtractionConfig::from(&config);
```

Dav1dde (approved):

relay-dynamic-config/src/metrics.rs
jjbayer: Depends whether we consider it a reference type or a value type. I don't have a strong opinion, assuming the compiler will optimize both versions to the same thing.

relay-server/src/metrics_extraction/event.rs
jjbayer: I'll leave this for the next PR, where I refactored these tests a bit: https://github.com/getsentry/relay/pull/3504

PR #3491, opened 2024-04-26
fix(spans): Extract http status code from span data when provided as int

narsaynorath:
Some SDKs send this field as an integer, which gets converted to `None` by the `v.as_str()` call. 

I added code to catch if the value is an integer and convert it to a string. In terms of testing, I tested the function which extracts the status, but also added sample spans to the mobile event to generate a snapshot validating both string or integers are extracted.

The issue was identified as a difference between Android and iOS events.


narsaynorath:
I can't say whether there are SDKs that send the response code in span `tags` as integers, so I haven't applied the change there. Let me know if you'd like me to do so.

relay-event-normalization/src/normalize/utils.rs
```diff
@@ -31,10 +32,13 @@ pub fn http_status_code_from_span(span: &Span) -> Option<String> {
         .data
         .value()
         .and_then(|data| data.http_response_status_code.value())
-        .and_then(|v| v.as_str())
-        .map(|v| v.to_string())
+        .map(|v| match v {
+            Value::String(s) => Some(s.as_str().to_string()),
+            Value::I64(i) => Some(i.to_string()),
```
Dav1dde, line 37:
```suggestion
            Value::I64(i) | Value::U64(i) => Some(i.to_string()),
```

relay-event-normalization/src/normalize/utils.rs
```diff
@@ -31,10 +32,13 @@ pub fn http_status_code_from_span(span: &Span) -> Option<String> {
         .data
         .value()
         .and_then(|data| data.http_response_status_code.value())
-        .and_then(|v| v.as_str())
-        .map(|v| v.to_string())
+        .map(|v| match v {
+            Value::String(s) => Some(s.as_str().to_string()),
```
Dav1dde:
```suggestion
            Value::String(s) => Some(s.as_str().to_owned()),
```

Dav1dde (approved):

PR #3494, opened 2024-04-26
chore(self-hosted): Add self-hosted e2e test action

hubertdeng123:
We've done a lot of work this quarter to improve the self-hosted e2e test suite, and want to turn this back on in relay.

More here: https://www.notion.so/sentry/Getting-self-hosted-e2e-tests-back-on-pre-commit-8a3e91cfe7fa475689f343c4d41eca31?pvs=4

Tests should no longer be as flaky as before, which is why it was removed in https://github.com/getsentry/relay/pull/2446

#skip-changelog


iker-barriocanal (commented):
Could we make it optional for some time to ensure no flakiness? After we've collected some data and ensured they work as expected, we can make them required.

Dav1dde:
Is it possible to instead run these E2E integration tests outside of Relay and they just listen on changes to Relay master? This wouldn't introduce another source of flakiness to Relay and save drastically on CI (and dev) time since it does only run once per PR instead of being triggered constantly.

iker-barriocanal:
Random idea: do we need to run tests for every PR, or a nightly test run is good enough?

azaslavsky (approved):

hubertdeng123:
@iker-barriocanal @Dav1dde Definitely open to making this optional for a week or so to see how they perform, before making them required as we are introducing another potential source of flakiness and we want this to go as smoothly as possible. It does look like the self-hosted-end-to-end action currently takes less time than the Sentry-Relay Integration Tests, so we hope that this doesn't drastically increase CI time. We've also set it to a timeout of 25 minutes, which is lower than the timeout of 30 minutes given for the Sentry-Relay Integration tests. 

To give more context on this change, we'd strongly prefer to get these tests on PR checks for a couple of reasons.

1. We'd like a tighter feedback loop on breaking changes for self-hosted. Many times these changes fall on us to initially investigate, before routing to the teams responsible. As a result, it takes pretty long for these fixes to get in. This in the past has caused stress during the time of the month we release self-hosted. 
2. Ideally, teams should own their features when they roll out to self-hosted. That includes making sure the appropriate tests are written in order to catch breakage. Getting these PR checks in are the first step towards this future, as it's difficult for the team working on self-hosted to write sufficient coverage for all new features that are being released.

Does this address the concerns you all have? Hope this makes sense and happy to discuss further!

iker-barriocanal (approved):
Not opposed to moving forward with the current approach!

Dav1dde (approved):

.github/workflows/ci.yml
```diff
@@ -552,3 +552,61 @@ jobs:
           echo "Testing against ${RELAY_TEST_IMAGE}"
           make test-relay-integration
 
+  self-hosted-end-to-end:
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    # temporary, remove once we are confident the action is working
+    continue-on-error: true
+    needs: build
```
Dav1dde: You need to depend on build-docker now.

.github/workflows/ci.yml
```diff
@@ -552,3 +552,61 @@ jobs:
...
+        # the image from ghcr
+        if: "!github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'"
+        uses: getsentry/action-self-hosted-e2e-tests@267fab2add9e173029c1e09ecfd82e04688d7c44
+        with:
+          project_name: relay
+          image_url: ghcr.io/getsentry/relay:${{ github.event.pull_request.head.sha || github.sha }}
+          docker_repo: getsentry/relay
+          docker_password: ${{ secrets.DOCKER_HUB_RW_TOKEN }}
+
+  publish-to-dockerhub:
```
Dav1dde: Introduced from a faulty merge!

Dav1dde (changes requested):

.github/workflows/ci.yml
```diff
@@ -552,3 +552,35 @@ jobs:
...
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Download Docker Image
+        uses: actions/download-artifact@v3
+        with:
+          name: relay-docker-image
+
+      - name: Import Docker Image
+        run: docker load -i relay-docker-image.tgz
```
Dav1dde: Oh and that doesn't exist anymore

.github/workflows/ci.yml
hubertdeng123: done, thanks!

.github/workflows/ci.yml
hubertdeng123: removed

.github/workflows/ci.yml
Dav1dde: I think the entire download and import is not necessary, do you even need the checkout of the repo?

Dav1dde (approved):

.github/workflows/ci.yml
hubertdeng123: I don't think I'll need the checkout of the repo, will remove and verify

PR #3496, opened 2024-04-29
feat(server): Extract forwarded ip from `CF-Connecting-IP` header

### LINKED ISSUE #3493: Cloudflare IP visible in Sentry for NextJs tunnel instead of real user IP
Hello, I am having trouble with getting real user IP to save in my tunneled Sentry errors/messages while using Cloudflare as a proxy for my NextJS application.

The problem is that Cloudflare replaces the "X-Forwarded-For" header with its own server IP and the real user IP is under "CF-Connecting-IP" header. I have tried using a middleware for my Sentry tunnel endpoint to replace the request's and response's "X-Forwarded-For" and "X-Vercel-Forwarder-For" as well as "X-Sentry-Forwarded-For" with the value of real user IP, however Sentry still keeps storing the cloudflare IP instead. 

Would it be possible to add this "[CF-Connecting-IP](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip)" as a priority header so that Sentry can automatically grab its value instead of the others if it exists? I will keep trying to replace it somehow in the middleware in the meantime, but it would be much easier if Sentry did it automatically.

Dav1dde:
Extracts the IP from the [`CF-Connecting-IP`](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip) header.

Also some related docs and code improvements.

Closes: #3493

### DIFF
--- a/relay-server/src/extractors/forwarded_for.rs
+++ b/relay-server/src/extractors/forwarded_for.rs
@@ -20,7 +20,7 @@ impl ForwardedFor {
     ///
     /// The Sentry SaaS infrastructure sets this header.
     const SENTRY_FORWARDED_HEADER: &'static str = "X-Sentry-Forwarded-For";
-    /// Cloudflare forwards the client ip in its own [`X-Vercel-Forwarded-For`] header.
+    /// Vercel forwards the client ip in its own [`X-Vercel-Forwarded-For`] header.
     ///
     /// [`X-Vercel-Forwarded-For`](https://vercel.com/docs/concepts/edge-network/headers#x-vercel-forwarded-for)
     const VERCEL_FORWARDED_HEADER: &'static str = "X-Vercel-Forwarded-For";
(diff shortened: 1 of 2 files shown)


relay-server/src/extractors/forwarded_for.rs
```diff
@@ -9,26 +9,51 @@ use axum::http::HeaderMap;
...
     const FORWARDED_HEADER: &'static str = "X-Forwarded-For";
-    const VERCEL_FORWARDED_HEADER: &'static str = "X-Vercel-Forwarded-For";
+    /// Sentry's custom forwarded for header.
+    ///
+    /// Clients or proxies can use `X-Sentry-Forwarded-For` when they cannot control the value of
+    /// the [`X-Forwarded-For`](Self::FORWARDED_HEADER) header.
+    ///
+    /// The Sentry SaaS infrastructure sets this header.
     const SENTRY_FORWARDED_HEADER: &'static str = "X-Sentry-Forwarded-For";
+    /// Cloudflare forwards the client ip in its own [`X-Vercel-Forwarded-For`] header.
```
iker-barriocanal:
```suggestion
    /// Vercel forwards the client ip in its own [`X-Vercel-Forwarded-For`] header.
```

iker-barriocanal (approved):
Thanks for updating the docs and improving the code!

PR #3501, opened 2024-04-29
feat(spans): Tag queue op spans with "messaging.destination.name"

edwardgou-sentry:
Tags indexed queue op spans with "messaging.destination.name".
Tags queue op exclusive time metrics with "messaging.destination.name".

"messaging.destination.name" is the name of the queue/topic that a publisher/consumer span is operating on, which should be fairly low cardinality for most users.

#skip-changelog


jjbayer (approved):
Looks OK to me as long as the tags are actually low cardinality. Looks like there's some merge conflicts.

PR #3504, opened 2024-04-30
ref(spans): Move metrics extraction config to global config

jjbayer:
Move the bulk of span metrics extraction config (`defaults.rs`) to the newly created global metrics extraction config. Only the feature flag checks to enable/disable groups of metrics remains in the project config. This should save a lot of memory.

ref: https://github.com/getsentry/team-ingest/issues/331


relay-server/src/metrics_extraction/event.rs
```diff
@@ -1559,8 +1526,8 @@ mod tests {
                 "ttfd": "ttfd"
             }
         }"#;
-        let span = Annotated::from_json(span).unwrap().into_value().unwrap();
-        let metrics = extract_span_metrics(&span);
+        let span: Span = Annotated::from_json(span).unwrap().into_value().unwrap();
```
iker-barriocanal: nit: was it a best practice to use the turbofish operator? I can't find any references:
```suggestion
        let span = Annotated::<Span>::from_json(span).unwrap().into_value().unwrap();
```

iker-barriocanal (approved):

relay-server/src/metrics_extraction/event.rs
jjbayer: 👍 Maybe not a best practice, but it's more explicit and it makes this call consistent with the one below.

relay-dynamic-config/src/global.rs
```diff
@@ -70,6 +71,25 @@ impl GlobalConfig {
...
+                if let Entry::Vacant(entry) = config.groups.entry(group_name) {
+                    entry.insert(MetricExtractionGroup {
+                        is_enabled: false, // must be enabled via project config
+                        metrics,
+                        tags: Default::default(),
+                    });
+                }
+            }
+        }
+    }
```
Dav1dde, line 92: Not a fan of these `normalize` etc methods which need to be called and modify the state of the object. Can we make this a separate type which just adds the defaults? Or just directly move everthing to Sentry.

It's just so easy to forget or miss.

jjbayer: Will follow-up on this.

PR #3506, opened 2024-04-30
feat(normalization): Add feature flags for normalization rollout

iker-barriocanal:
The previous normalization rollout failed and required a rollback, requiring a long time to complete. This PR adds feature flags to apply changes instantly (<30s).

#skip-changelog


relay-dynamic-config/src/global.rs
```diff
@@ -196,6 +196,29 @@ pub struct Options {
     )]
     pub span_extraction_sample_rate: Option<f32>,
 
+    /// If enabled, runs full normalization in non-processing Relays.
+    ///
+    /// `None` and `Some(false)` are equivalent and NOOPs. Doesn't apply to processing Relays.
+    ///  Outdated relays with a stale protocol/normalization receiving this flag
+    ///  will not forward unknown fields. Disabling the flag solves this
+    ///  behavior.
```
iker-barriocanal: This is an interesting edge case: setting up a Relay that supports this flag (post-release on May 15) and updating upstream Relays results in running full normalization in the external Relay. This edge case doesn't exist if we complete the rollout before May 15 as I intend, and disabling/removing the flag fixes this behavior.

relay-server/src/services/processor.rs
```diff
@@ -1227,24 +1227,37 @@ impl EnvelopeProcessorService {
...
+            && matches!(options.processing_disable_normalization, Some(true))
+            && (state.envelope().meta().is_from_internal_relay())
+        {
+            return Ok(());
+        } else if !self.inner.config.processing_enabled()
+            && matches!(options.force_full_normalization, Some(true))
+        {
+            true
+        } else {
+            match self.inner.config.normalization_level() {
```
iker-barriocanal, line 1265: This `match` block remains unchanged, only feature flags have been prioritized.

relay-dynamic-config/src/global.rs
```diff
@@ -203,6 +203,29 @@ pub struct Options {
...
+    #[serde(
+        rename = "relay.force_full_normalization",
+        deserialize_with = "default_on_error",
+        skip_serializing_if = "is_default"
+    )]
+    pub force_full_normalization: Option<bool>,
+
+    /// If enabled, disables normalization in processing Relays.
+    ///
+    /// `None` and `Some(false)` are equivalent and NOOPs.
```
Dav1dde: Then we can just use `serde(default)` and a `bool`

relay-dynamic-config/src/global.rs
```diff
@@ -203,6 +203,29 @@ pub struct Options {
     )]
     pub span_extraction_sample_rate: Option<f32>,
 
+    /// If enabled, runs full normalization in non-processing Relays.
+    ///
+    /// `None` and `Some(false)` are equivalent and NOOPs. Doesn't apply to processing Relays.
```
Dav1dde: dito, just use a boolean then

Dav1dde (approved):
> This is an interesting edge case: setting up a Relay that supports this flag (post-release on May 15) and updating upstream Relays results in running full normalization in the external Relay.

I don't like this ticking timebomb, we should remove this as soon as possible again from the codebase.

PR #3508, opened 2024-04-30
feat(metric-stats): Emit negative outcome for denied metrics

### LINKED ISSUE #3482: Emit Negative Outcomes: Denied Metrics
Emit metric stats and outcomes for metrics that are dropped as a result of the [`config.metrics.denied_names`](https://github.com/getsentry/relay/blob/bf2441dfbd594bd02a3581faef6c886bf2666c42/relay-dynamic-config/src/metrics.rs#L23) config in [`Project::filter_metrics`](https://github.com/getsentry/relay/blob/bf2441dfbd594bd02a3581faef6c886bf2666c42/relay-server/src/services/project.rs#L615-L618)

## Acceptance Criteria

 - Outcomes and metric stats should be emitted in a batched fashion wherever possible.
 - Outcomes should use [`DataCategory::MetricBucket`](https://github.com/getsentry/relay/blob/bf2441dfbd594bd02a3581faef6c886bf2666c42/relay-base-schema/src/data_category.rs#L65-L66) and `Outcome::Filtered` with reason code `"denied_name"`.
 - The volume metric stats (`c:metric_stats/volume@none`) should be emitted with the correct tagging and with `OutcomeId::FILTERED`.

## Validation

- [Organization stats](https://sentry.sentry.io/stats/?dataCategory=metrics&project=1&statsPeriod=24h&transform=periodic) show the number of "Filtered Metrics" after sending metrics that have a denied name.

iambriccardo:
This PR emits negative outcomes for denied metrics. In addition, it improves the code that handles rate limiting of metric buckets and improves tests.

Closes: https://github.com/getsentry/relay/issues/3482


relay-server/src/services/project.rs
```diff
@@ -598,39 +593,71 @@ impl Project {
...
+    ) -> Vec<Bucket> {
+        let mut denied_buckets = vec![];
+
+        let accepted_buckets = buckets
+            .into_iter()
+            .filter_map(|mut bucket| {
+                let Ok(mri) = MetricResourceIdentifier::parse(&bucket.name) else {
+                    relay_log::trace!(mri = &*bucket.name, "dropping metrics with invalid MRI");
+                    return None;
+                };
```
Dav1dde: We can just use `bucket.name.namespace()` instead.

relay-server/src/services/project.rs
```diff
@@ -598,39 +593,71 @@ impl Project {
...
+
+                if !is_metric_namespace_valid(state, &mri.namespace) {
+                    relay_log::trace!(mri = &*bucket.name, "dropping metric in disabled namespace");
+                    return None;
+                };
```
Dav1dde: I think we should also count these as dropped and count them.

relay-server/src/services/project.rs
```diff
@@ -598,39 +593,71 @@ impl Project {
...
-                relay_log::trace!(mri = &*bucket.name, "dropping metrics with invalid MRI");
-                return false;
-            };
-
-            if !is_metric_namespace_valid(state, &mri.namespace) {
-                relay_log::trace!(mri = &*bucket.name, "dropping metric in disabled namespace");
-                return false;
-            };
+    /// Removes metric buckets that are not allowed to be ingested and reports denied buckets.
+    fn filter_buckets(
```
Dav1dde: It also modifies buckets :(

Maybe something like `pre_process_buckets`?

relay-server/src/services/project.rs
```diff
@@ -702,8 +721,18 @@ impl Project {
...
+                    outcome_aggregator.clone(),
+                    metric_stats.clone(),
+                    mode,
+                    &state,
+                    buckets,
+                );
 
-                let buckets = self.rate_limit_metrics(&state, buckets, outcome_aggregator.clone());
+                let buckets =
+                    self.rate_limit_buckets(outcome_aggregator.clone(), mode, &state, buckets);
```
Dav1dde: Please double check me here:

It seems we only call `rate_limit_buckets` when we also call `rate_limit_and_merge_buckets` right after, latter also checks the cached limits but for processing Relays also forwards to the consistent rate limiter -> we can remove `rate_limit_buckets` as it is not needed.

This also allows us to only `filter_buckets` when merging into the aggregator in `rate_limit_and_merge_buckets` which we should rename to something better.

relay-server/src/services/project.rs
iambriccardo: Will figure out a better name

relay-server/src/services/project.rs
```diff
@@ -422,7 +422,7 @@ enum State {
 impl State {
     fn state_value(&self) -> Option<Arc<ProjectState>> {
         match self {
-            State::Cached(state) => Some(Arc::clone(state)),
```
Dav1dde, line 425: Can keep this syntax, it makes it explicit you're cloning the smart pointer not its contents.

relay-server/src/services/project.rs
```diff
@@ -633,13 +633,25 @@ impl Project {
...
+
+        // Pre-process buckets to make sure they are valid.
+        let buckets = self.pre_process_buckets(
+            outcome_aggregator.clone(),
+            metric_stats,
+            extraction_mode,
+            &state,
+            buckets,
+        );
+
```
Dav1dde, line 653: Exit early if buckets is empty

Dav1dde (approved):

PR #3509, opened 2024-04-30
feat(bucket-encoding): Switch base64 to base64 without padding

Dav1dde:
Switches the base64 encoding in the bucket encoding to an encoding without padding to save some bytes on small payloads.

#skip-changelog


iker-barriocanal (approved):
TIL that was padding and removing it was a thing.

PR #3510, opened 2024-04-30
fix(spans): cache item size not being collected

DominikB2014:
1. cache.item_size was not being collected at all, looks like this is because it was not being retrieved correctly as it comes from measurements.
2. the enviornment tag was not being added


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+
+**Bug fixes:**
+
+- Extract `cache.item_size` from measurements instead of data. ([#3510](https://github.com/getsentry/relay/pull/3510))
+- Collect `enviornment` tag as part of exclusive_time_light for cache spans. ([#3510](https://github.com/getsentry/relay/pull/3510))
+
+**Features**:
+
+**Internal**:
+
```
iker-barriocanal: nit: no need to add empty sections:
```suggestion
```

iker-barriocanal (approved):

PR #3511, opened 2024-04-30
feat(spans): Adds more queue span tags to metrics

edwardgou-sentry:
Adds `trace.status` as a tag to `exclusive_time` metrics, just for queue spans. `trace.status` is low cardinality, see: https://develop.sentry.dev/sdk/event-payloads/contexts/#trace-context

Adds `messaging.destination.name` as a tag to `messaging.message.receive.latency` metric. `messaging.destination.name` should also be low cardinality since it the queue/topic name that a consumer/producer span operates on.

#skip-changelog

### DIFF
--- a/relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics.snap
+++ b/relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics.snap
@@ -5817,6 +5817,7 @@ expression: metrics
             "messaging.destination.name": "default",
             "span.category": "queue",
             "span.op": "queue.task.celery",
+            "trace.status": "ok",
             "transaction": "gEt /api/:version/users/",
             "transaction.op": "myop",
         },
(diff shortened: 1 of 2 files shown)


edwardgou-sentry:
I _think_ this should be the last remaining change to tags and metrics required to get the queues module running.

relay-server/src/metrics_extraction/snapshots/relay_server__metrics_extraction__event__tests__extract_span_metrics.snap
```diff
@@ -5882,6 +5884,7 @@ expression: metrics
         ),
         tags: {
             "environment": "fake_environment",
+            "messaging.destination.name": "default",
```
iker-barriocanal, line 5887: Are the `default` values placeholders? Can we add a test with an expected payload to verify the tag value is correct?

edwardgou-sentry: Hey @iker-barriocanal , these values are not placeholders, they come from the tests here:
https://github.com/getsentry/relay/blob/master/relay-server/src/metrics_extraction/event.rs#L1060
Thanks!

iker-barriocanal (approved):

Issue #3513, opened 2024-05-02
Update Thread distribution between Relay runtimes


Dav1dde:
Rework the the Relay runtimes to only have a single global runtime using `max_thread_count` threads and have a separate thread pool (or runtime) managed with a semaphore for the processor service.

In the initial implementation we just want to assign `max_thread_count` to the global runtime and `max_parallelism - 2` (needs to be adjusted for low values of `max_thread_count`) for the processor.

In the future the processor should dynamically adjust it's parallelism based on the available resources and take up all resources the global runtime does *not* need. Pretty much a priority system where the global runtime has the highest priority and the processor has a lower priority.

COS:
- [ ] Remove all additional runtimes
- [ ] Update thread/parallelism counts

PR #3516, opened 2024-05-02
feat(server): Use a single tokio runtime, adjust processor task count

Dav1dde:
As described in #3513.

- Removes all additional secondary runtimes for individual services.
- Adjusts the thread count for the processor to not use all resources given.

Will be followed by an ops PR to adjust configuration settings.

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -586,7 +586,7 @@ struct Limits {
     max_replay_message_size: ByteSize,
     /// The maximum number of threads to spawn for CPU and web work, each.
     ///
-    /// The total number of threads spawned will roughly be `2 * max_thread_count + 1`. Defaults to
+    /// The total number of threads spawned will roughly be `2 * max_thread_count`. Defaults to
     /// the number of logical CPU cores on the host.
     max_thread_count: usize,
     /// The maximum number of seconds a query is allowed to take across retries. Individual requests
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -2630,9 +2630,8 @@ impl Service for EnvelopeProcessorService {
         // Adjust thread count for small cpu counts to not have too many idle cores
         // and distribute workload better.
         let thread_count = match self.inner.config.cpu_concurrency() {
-            0 | 1 => 1,
-            2 | 3 => 2,
-            4 => 3,
+            conc @ 0..=2 => conc.max(1),
+            conc @ 3..=4 => conc - 1,
             conc => conc - 2,
         };
         relay_log::info!("starting {thread_count} envelope processing workers");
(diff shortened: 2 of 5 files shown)


jjbayer (commented):
Code looks good to me but I'm nervous about shipping this. Should we run it on the load tester first? Or let it sit on Canary for a week before we deploy it to all pods?

Dav1dde:
We can, but I don't see this as risky and would have moved on from canary and S4S after a workday.

relay-config/src/config.rs
```diff
@@ -586,7 +586,7 @@ struct Limits {
     max_replay_message_size: ByteSize,
     /// The maximum number of threads to spawn for CPU and web work, each.
     ///
-    /// The total number of threads spawned will roughly be `2 * max_thread_count + 1`. Defaults to
+    /// The total number of threads spawned will roughly be `2 * max_thread_count`. Defaults to
```
jan-auer, line 589: Can we add to this comment that Relay will use at most `max_thread_count` CPU?

relay-server/src/services/processor.rs
```diff
@@ -2627,7 +2627,14 @@ impl Service for EnvelopeProcessorService {
...
 
     fn spawn_handler(self, mut rx: relay_system::Receiver<Self::Interface>) {
-        let thread_count = self.inner.config.cpu_concurrency();
+        // Adjust thread count for small cpu counts to not have too many idle cores
+        // and distribute workload better.
+        let thread_count = match self.inner.config.cpu_concurrency() {
+            0 | 1 => 1,
+            2 | 3 => 2,
+            4 => 3,
+            conc => conc - 2,
```
jan-auer: Have you run load tests with this? It would be good to know the actual needs here, if this is easy to do.

jan-auer: Took me a while to understand the assignment, would this be easier to undestand?

```suggestion
            0 | 1 | 2 @ conc => conc.max(1),
            3 | 4 @ conc => conc - 1,
            conc => conc - 2,
```

(could also be assigned to a variable before match)

jan-auer (approved):

relay-server/src/services/processor.rs
Dav1dde: No, this was simply a best effort guess for small amounts of cores going off our discussed `-2` starting point. 

Having a small amount of cores would reserve a much bigger % to the runtime than if we have a large amount of cores. E.g. for 2 cores allocating 1 thread to the processor would allocate 50% to the runtime, while having 12 cores only allocates 17% to the runtime.

relay-config/src/config.rs
Dav1dde: With the current setup this isn't gurantueed right? If the runtime needs more processing power than expected it can go over.

PR #3521, opened 2024-05-02
ref(spans): Rename exclusive time attribute name

phacops:
This aligns more with how OTel named their timestamp fields, with the unit spelled out more.

This field is not in use yet but will be with https://github.com/getsentry/sentry-javascript/pull/11876.

#skip-changelog

### DIFF
--- a/relay-spans/src/span.rs
+++ b/relay-spans/src/span.rs
@@ -169,7 +169,7 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
                 "http.route" | "url.path" => {
                     http_route = otel_value_to_string(value);
                 }
-                key if key.contains("exclusive_time_ns") => {
+                key if key.contains("exclusive_time_nano") => {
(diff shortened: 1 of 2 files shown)


relay-spans/src/span.rs
```diff
@@ -169,7 +169,7 @@ pub fn otel_to_sentry_span(otel_span: OtelSpan) -> EventSpan {
                 "http.route" | "url.path" => {
                     http_route = otel_value_to_string(value);
                 }
-                key if key.contains("exclusive_time_ns") => {
+                key if key.contains("exclusive_time_nano") => {
```
Dav1dde, line 172: Do they actually suffix stuff with `_nano` and not `_nanos`?

jjbayer: Yes, for example `StartTimeUnixNano` https://opentelemetry.io/docs/specs/otel/metrics/data-model/.

jjbayer (approved):

PR #3523, opened 2024-05-02
feat(spans): Forward span.data on the Kafka message

phacops:
We'd like to use more data from the span and eventually store it. First step is to add it to the span Kafka message so downstream can use it.

We understand there might be some duplication between `sentry_tags` and `data` and we accept it for now to not have to parse the object.

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -1390,6 +1390,8 @@ struct SpanKafkaMessage<'a> {
     #[serde(default)]
     is_segment: bool,
 
+    #[serde(default, skip_serializing_if = "none_or_empty_object")]
+    data: Option<&'a RawValue>,
     #[serde(borrow, default, skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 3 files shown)


relay-server/src/services/store.rs
```diff
@@ -1390,6 +1390,8 @@ struct SpanKafkaMessage<'a> {
     #[serde(default)]
     is_segment: bool,
 
+    #[serde(default, skip_serializing_if = "none_or_empty_object")]
+    data: Option<&'a RawValue>,
```
Dav1dde, line 1396: I was confused where this field is set, then saw that it's because of the useless serialization roundtrip 🤣

Dav1dde (approved):

relay-server/src/services/store.rs
jjbayer: Yeah we're basically trying to make the roundtrip as cheap as possible.

jjbayer (approved):
If the Kafka broker and consumers can take the increase in message size, ship it!

Might be worth doing a Canary deploy (`deploy-relay-experimental`) first to see if the maximum message size or consumer processing time increases.

PR #3524, opened 2024-05-03
ref(spans): Tag duration like exclusive time

phacops:
We are going towards a world where exclusive time will be optional as some SDKs might not be able to send it. The whole OTel world do not use exclusive time.

We're currently using in the product the exclusive time metric in most places, even in places where duration would be appropriate because we're showing up a leaf span (a span without children so exclusive time equals duration).

This is preliminary work so we have both metrics tagged the same way so we can transition to the duration where appropriate.

In the future, when we allow spans without exclusive time, the metric won't be recorded and it'll be fine since we rely on duration most of the time.

As for the light version of the metric, it's not clear if we need to emit it yet. The biggest difference in that metric is the absence of the `transaction` tag. Perhaps the speed advantage we had at the beginning disappeared as we added more tags and we can stop emitting the light metrics.


jjbayer (approved):
The added tags seem to be either low-cardinality or derived from a tag that was already added (e.g. `description` and `group`), so this should be fine. It can't hurt to keep an eye on overall cardinality and kafka message counts after this change though.

PR #3525, opened 2024-05-03
fix(ci): Skip flaky error checks

jjbayer:
These checks fail repeatedly with the following exception, preventing full Relay deploys.

```bash
urllib.error.URLError: <urlopen error [Errno 110] Connection timed out>
```

#skip-changelog


Dav1dde (approved):
Not sure if we break gocd if we remove tasks?

Issue #3483, opened 2024-04-25
Configure Relay aggregator for `custom` to 60s granularity


jan-auer:
DDM will launch without 10s granularity, which means the aggregator in Relay can be configured to the higher 60s bucketing interval.

See https://github.com/getsentry/sentry/issues/69648
```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/relay/pull/3527
- [ ] https://github.com/getsentry/ops/pull/10350
```

jan-auer:
Applied in all regions.

PR #3527, opened 2024-05-03
ref(metrics): Support complex conditions to share aggregators

jan-auer:
Introduces `and`, `or`, and `not` conditions to offer more flexibility in
assigning metrics aggregators to namespaces. Particularly, this change allows to
apply an aggregator override to several namespaces at once:

```yaml
secondary_aggregators:
  - name: minute
    condition:
      op: or
      inner:
        - op: eq
          field: namespace
          value: spans
        - op: eq
          field: namespace
          value: custom
    config:
      bucket_interval: 60
```

Task: https://github.com/getsentry/relay/issues/3483
#skip-changelog


relay-config/src/config.rs
```diff
@@ -2397,9 +2395,8 @@ impl Config {
     /// Returns aggregator config for a given metrics namespace.
     pub fn aggregator_config_for(&self, namespace: MetricNamespace) -> &AggregatorServiceConfig {
         for entry in &self.values.secondary_aggregators {
-            match entry.condition {
-                Condition::Eq(Field::Namespace(ns)) if ns == namespace => return &entry.config,
-                _ => (),
```
jan-auer, line 2409: This was actually incredibly dangerous -- there is no warning that we silently fall back to a wrong aggregator config. We should not add such blanked match arms, especially not if the enum is not exhaustive.

jjbayer (approved):

relay-config/src/config.rs
jjbayer: 👍 agreed.

PR #3531, opened 2024-05-03
fix(spans): Normalize segment span ID

jjbayer:
`SpanId` is represented as a string internally, so if we construct it without going through `Annotated::from_value`, lowercase span IDs and uppercase span IDs will be treated as different. This is problematic in `set_segment_attributes`, which compares the span's ID to its segment ID.

We should probably represent `SpanId` as `[u8; 8]`, or at least go through `Annotated::from_value` for all Otel types, but this PR should do as a quick fix.

#skip-changelog

### DIFF
--- a/relay-spans/src/span.rs
+++ b/relay-spans/src/span.rs
@@ -65,11 +65,12 @@ fn otel_value_to_string(value: OtelValue) -> Option<String> {
 }
 
 fn otel_value_to_span_id(value: OtelValue) -> Option<String> {
-    match value {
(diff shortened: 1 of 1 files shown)


Dav1dde (approved):
+1 for `[u8; 8]`

jjbayer:
> +1 for `[u8; 8]`

I will follow-up in https://github.com/getsentry/relay/pull/3535.

iker-barriocanal (approved):

PR #3533, opened 2024-05-03
feat(spans): Use zero-padded span ID as event ID

jjbayer:
For events created from segment spans, deterministically map the span ID to an event ID.

Options:

- [chosen] Pad span ID with zeros. This still leaves `2^64 == 18,446,744,073,709,551,616` combinations so should be good enough to prevent collisions within a project.
- XOR span ID with the trace ID. This might be a better alternative if collisions are an issue, but makes it harder to spot a span-ID-turned-event-ID with the naked eye.
- [rejected] Create a UUID v5 from a namespace (as we do [here](https://github.com/getsentry/relay/blob/c81f37f4d55e45c94cc344d145b151cced609e5f/relay-monitors/src/lib.rs#L212-L215)). Discarded because it is not reversible.

#skip-changelog

### DIFF
--- a/relay-event-schema/src/protocol/event.rs
+++ b/relay-event-schema/src/protocol/event.rs
@@ -15,8 +15,8 @@ use crate::protocol::{
     AppContext, Breadcrumb, Breakdowns, BrowserContext, ClientSdkInfo, Contexts, Csp, DebugMeta,
     DefaultContext, DeviceContext, EventType, Exception, ExpectCt, ExpectStaple, Fingerprint, Hpkp,
     LenientString, Level, LogEntry, Measurements, Metrics, MetricsSummary, OsContext,
-    ProfileContext, RelayInfo, Request, ResponseContext, Span, Stacktrace, Tags, TemplateInfo,
-    Thread, Timestamp, TraceContext, TransactionInfo, User, Values,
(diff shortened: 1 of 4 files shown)


mjq-sentry (approved):

relay-event-schema/src/protocol/event.rs
```diff
@@ -65,6 +65,16 @@ impl FromStr for EventId {
 
 relay_common::impl_str_serde!(EventId, "an event identifier");
 
+impl TryFrom<&SpanId> for EventId {
+    type Error = ();
```
Dav1dde: We already have `thiserror` could make a quick proper error for this:

```rs
#[error("cannot convert span id to event id")]
pub struct TryFromSpanIdError;
```

Could also include the source error `pub struct TryFromSpanIdError(FooError);`

Dav1dde (approved):
An integration test could make sense here.

relay-event-schema/src/protocol/event.rs
jjbayer: I decided to make this consistent with the `FromStr for EventId` implementation above, but fleshed out the error type in `TryFrom<&Span>` a bit.

jjbayer:
> An integration test could make sense here.

👍 Added some assertions to the existing integration test now.

PR #3534, opened 2024-05-03
fix(normalization): Consider feature flag for event deserialization

iker-barriocanal:
https://github.com/getsentry/relay/commit/b74053c53983aeaa3a5c12dd75e25f85742e39e1 added feature flags to control the normalization of Relay. However, resetting the event.type during deserialization wasn't updated and only depended on the static config, ignoring the feature flag that overrides it. This PR fixes that by prioritizing the feature flag during deserialization.

#skip-changelog


iambriccardo (approved):
LGTM

PR #3539, opened 2024-05-03
fix(ai-monitoring): Make sure spans with op ai.pipeline also get tag

colin-sentry:
We send spans with the op `ai.pipeline` and `ai.pipeline.langchain` (for example) - this makes the former also get span group, which is expected in the frontend.


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -90,7 +90,12 @@ pub(crate) fn scrub_span_description(
                 }
             }
             ("resource", ty) => scrub_resource(ty, description),
-            ("ai", sub) if sub.starts_with("run.") || sub.starts_with("pipeline.") => {
+            ("ai", sub)
```
jjbayer: Would do something like

```rust
("ai", sub) => {
                match sub.split_once('.').unwrap_or((sub, "")) {
                    ("run" | "pipeline", _) => {
                        // ai.run.* and ai.pipeline.* are low cardinality (<100 per org) and describe
                        // the names of nodes of an AI pipeline.
                        Some(description.to_owned())
                    }
                    _ => None,
                }
```

jjbayer (approved):

PR #3544, opened 2024-05-06
feat(metrics): Emit negative outcomes when namespaces are disabled

### LINKED ISSUE #3532: Emit negative outcomes when metrics are dropped because of invalid namespace
We would like to emit negative outcomes when metrics are dropped because of invalid namespace, which happens here: https://github.com/getsentry/relay/blob/6fed2a76832bb13d84ff6b9e1800e344cb960186/relay-server/src/services/project.rs#L592-L592

iambriccardo:
This PR implements negative outcomes for metrics that are rejected due to disabled namespace.

Closes: https://github.com/getsentry/relay/issues/3532

### DIFF
--- a/relay-server/src/services/project.rs
+++ b/relay-server/src/services/project.rs
@@ -421,7 +421,7 @@ enum State {
 impl State {
     fn state_value(&self) -> Option<Arc<ProjectState>> {
         match self {
-            State::Cached(state) => Some(state.clone()),
+            State::Cached(state) => Some(Arc::clone(state)),
             State::Pending(_) => None,
         }
     }
@@ -578,6 +578,7 @@ impl Project {
         state: &ProjectState,
         buckets: Vec<Bucket>,
     ) -> Vec<Bucket> {
+        let mut disabled_namespace_buckets = vec![];
         let mut denied_buckets = vec![];
 
         let accepted_buckets = buckets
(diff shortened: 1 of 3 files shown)


relay-server/src/services/project.rs
```diff
@@ -710,7 +725,7 @@ impl Project {
 
         match &mut self.state {
             State::Cached(state) => {
-                let state = state.clone();
+                let state = Arc::clone(state);
```
iambriccardo, line 728: Change due to a suggestion in a previous related PR: https://github.com/getsentry/relay/pull/3508/files

relay-server/src/services/project.rs
```diff
@@ -610,9 +612,19 @@ impl Project {
             })
             .collect();
 
-        if !denied_buckets.is_empty() {
-            let scoping = self.scoping();
-            if let Some(scoping) = scoping {
+        if let Some(scoping) = self.scoping() {
+            if !disabled_namespace_buckets.is_empty() {
+                utils::reject_metrics(
```
iambriccardo, line 617: Here i was torn between allowing multiple outcomes to be supported in `reject_metrics` but I feel like this implementation is a bit more explicit even though it's more verbose.

Dav1dde (approved):

PR #3545, opened 2024-05-06
ref(integration-tests): Remove generators from consumer integration tests

### LINKED ISSUE #3546: Remove generators from consumers in integration tests

iambriccardo:
This PR removes the use of generators in integration tests utilities, since it's very easy to fall in the mistake of asserting the result of a generator which would lead to a true value since a function that has `yield` is guaranteed to always return a generator instance.

Closes: https://github.com/getsentry/relay/issues/3546

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Extract `cache.item_size` from measurements instead of data. ([#3510](https://github.com/getsentry/relay/pull/3510))
 - Collect `enviornment` tag as part of exclusive_time_light for cache spans. ([#3510](https://github.com/getsentry/relay/pull/3510))
 - Emit negative outcomes for denied metrics. ([#3508](https://github.com/getsentry/relay/pull/3508))
+- Remove generators from consumers in integration tests ([#3545](https://github.com/getsentry/relay/pull/3545))
 
 ## 24.4.2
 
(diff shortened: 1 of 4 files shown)


tests/integration/fixtures/processing.py
```diff
@@ -355,14 +355,18 @@ def get_metric(self, timeout=None):
...
 
             if message is None:
-                return
+                break
             else:
                 assert message.error() is None
-                yield json.loads(message.value()), message.headers()
+                metrics.append((json.loads(message.value()), message.headers()))
+
+        return metrics
```
jjbayer, line 369: Alternatively you could have kept the generator as `_get_metrics`, and make `get_metrics` return `list(self._get_metrics(...))`. But this works too 👍

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Extract `cache.item_size` from measurements instead of data. ([#3510](https://github.com/getsentry/relay/pull/3510))
 - Collect `enviornment` tag as part of exclusive_time_light for cache spans. ([#3510](https://github.com/getsentry/relay/pull/3510))
 - Emit negative outcomes for denied metrics. ([#3508](https://github.com/getsentry/relay/pull/3508))
+- Remove generators from consumers in integration tests ([#3545](https://github.com/getsentry/relay/pull/3545))
```
jjbayer: For something like this, you can also use `#skip-changelog` in the PR description.

jjbayer (approved):

CHANGELOG.md
iambriccardo: Good to know! Will update!

tests/integration/fixtures/processing.py
iambriccardo: I wanted to save some work for the compiler 😆

PR #3549, opened 2024-05-06
fix(spans): Scrub UUID in fallback regex

jjbayer:
When we cannot parse SQL, we should detect UUIDs.

#skip-changelog


gggritso (approved):
👍🏻

Dav1dde (approved):

PR #3552, opened 2024-05-06
feat(spans): Add inbound filters for spans

### LINKED ISSUE #3550: Call inbound filter logic in in `span::process`.

iambriccardo:
This PR adds support for inbound filters on spans which as of now should happen only on processing relays.

Closes: https://github.com/getsentry/relay/issues/3550


relay-server/src/services/processor/span/processing.rs
```diff
@@ -98,6 +102,19 @@ pub fn process(
             _ => return ItemAction::Keep,
         };
 
+        if let Some(span) = annotated_span.value() {
```
iambriccardo: After some different implementations, I settled on just putting the filtering here, even though I don't know if it should be put into the `process` function. The only thing I know is that separating into two functions would require quite a bit of refactoring.

relay-event-schema/src/protocol/span.rs
```diff
@@ -375,6 +375,7 @@ impl Getter for SpanData {
             "ui\\.component_name" => self.ui_component_name.value()?.into(),
             "url\\.scheme" => self.url_scheme.value()?.into(),
             "transaction" => self.segment_name.as_str()?.into(),
+            "release" => self.release.as_str()?.into(),
```
iambriccardo, line 401: Added this field since it was not there.

relay-filter/src/interface.rs
```diff
@@ -108,3 +110,41 @@ impl Filterable for Replay {
         self.user_agent()
     }
 }
+
+impl Filterable for Span {
```
iambriccardo, line 114: We might want to extend the impl of the methods returning `None` in case we find a way to access those fields on the `Span`.

relay-server/src/services/processor.rs
```diff
@@ -3475,10 +3475,10 @@ mod tests {
             processor.handle_process_metrics(&mut token, message);
 
             let value = project_cache_rx.recv().await.unwrap();
-            let ProjectCache::MergeBuckets(merge_buckets) = value else {
+            let ProjectCache::AddMetricBuckets(add_metric_buckets) = value else {
```
iambriccardo: Had to fix these tests for the PR to pass CI. The failures are related to: https://github.com/getsentry/relay/commit/fc1c2aa9ef2be4562f1062ec512e4c9b96e51118

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Properly handle AI metrics from the Python SDK's `@ai_track` decorator ([#3539](https://github.com/getsentry/relay/pull/3539))
 
+**Features**:
+
+- Add support for inbound filters on standalone spans. ([#3552](https://github.com/getsentry/relay/pull/3552))
```
jjbayer: mini nit: We will actually start filtering right away.

```suggestion
- Apply legacy inbound filters to standalone spans. ([#3552](https://github.com/getsentry/relay/pull/3552))
```

relay-filter/src/interface.rs
```diff
@@ -108,3 +110,41 @@ impl Filterable for Replay {
...
+    }
+
+    fn release(&self) -> Option<&str> {
+        self.data.value().and_then(|d| d.release.as_str())
+    }
+
+    fn transaction(&self) -> Option<&str> {
+        if self.is_segment.value().map_or(false, |&s| s) {
+            return self.description.as_str();
+        }
```
jjbayer: Let's use `span.data.segment_name` here. `span.description` will only match the transaction name if the span is a segment itself.

relay-filter/src/interface.rs
```diff
@@ -108,3 +110,41 @@ impl Filterable for Replay {
...
+
+        None
+    }
+
+    fn url(&self) -> Option<Url> {
+        None
+    }
```
jjbayer: This we should be able to get from `span.data`, though I'm not 100% sure which field. Check https://opentelemetry.io/docs/specs/semconv/http/http-spans/ and https://develop.sentry.dev/sdk/performance/span-data-conventions/.

relay-server/src/services/processor.rs
jjbayer: I think you can just merge or rebase on latest master, see https://github.com/getsentry/relay/pull/3560.

relay-server/src/services/processor/span/processing.rs
```diff
@@ -117,11 +121,28 @@ pub fn process(
             return ItemAction::Drop(Outcome::Invalid(DiscardReason::Internal));
         };
 
+        if let Some(span) = annotated_span.value() {
+            metric!(timer(RelayTimers::EventProcessingFiltering), {
```
jjbayer: Let's remove this metric for now, we don't want to mix span processing and event processing metrics.

tests/integration/test_spans.py
```diff
@@ -1589,3 +1589,101 @@ def test_span_extraction_with_tags(
...
+                    },
+                ).encode()
+            ),
+        )
+    )
+
+    relay.send_envelope(project_id, envelope)
+
+    spans = spans_consumer.get_spans(timeout=10.0, max_attempts=6)
+    assert len(spans) == 0
```
jjbayer: Let's explicitly check the `outcomes_consumer` here. See for example https://github.com/getsentry/relay/blob/ec501ca5196735e9c4a1db6de1cbb79048fc0761/tests/integration/test_outcome.py#L1518-L1536.

Then at the end, you can check that both consumers are empty with

```
    spans_consumer.assert_empty()
    outcomes_consumer.assert_empty()
```

relay-filter/src/interface.rs
```diff
@@ -108,3 +110,41 @@ impl Filterable for Replay {
...
+
+    fn exceptions(&self) -> Option<&Values<Exception>> {
+        None
+    }
+
+    fn ip_addr(&self) -> Option<&str> {
+        None
+    }
```
jjbayer: For events, this is the user IP address. As for `url()`, we should pick a field in span.data and get the value from there (we can sync on this because it's not obvious which field to pick).

relay-filter/src/interface.rs
```diff
@@ -108,3 +110,41 @@ impl Filterable for Replay {
...
+
+    fn url(&self) -> Option<Url> {
+        None
+    }
+
+    fn user_agent(&self) -> Option<&str> {
+        None
+    }
```
jjbayer: Let's add a field `user_agent.original` to `SpanData` and get the value from there: https://opentelemetry.io/docs/specs/semconv/attributes-registry/user-agent/

relay-filter/src/interface.rs
```diff
@@ -108,3 +110,45 @@ impl Filterable for Replay {
...
+
+    fn release(&self) -> Option<&str> {
+        self.data.value()?.release.as_str()
+    }
+
+    fn transaction(&self) -> Option<&str> {
+        if self.is_segment.value().map_or(false, |&s| s) {
+            return self.data.value()?.segment_name.as_str();
+        }
```
jjbayer: I actually think we can simplify this to 
```suggestion
        self.data.value()?.segment_name.as_str()
```

When we drop a transaction event because of this filter, we also drop its child spans. So dropping every span that can be attributes to this segment name is consistent.

tests/integration/test_spans.py
```diff
@@ -1594,3 +1594,125 @@ def test_span_extraction_with_tags(
...
+    def summarize_outcomes():
+        counter = Counter()
+        for outcome in outcomes_consumer.get_outcomes():
+            counter[(outcome["category"], outcome["outcome"])] += outcome["quantity"]
+        return counter
+
+    assert summarize_outcomes() == {(12, 1): 1}
+
+    spans_consumer.assert_empty()
+    outcomes_consumer.assert_empty()
```
jjbayer: nit: Because integration tests take a long time to run, I suggest keeping only one of these two integration tests, and cover other filters with unit tests. See for example https://github.com/getsentry/relay/blob/b5fa1ff0b90e04934c0ac1f09e6410bd70ce661b/relay-filter/src/csp.rs#L313-L326

jjbayer (approved):
Looks good! Please see comments before merging.

tests/integration/test_spans.py
```diff
@@ -1594,3 +1594,71 @@ def test_span_extraction_with_tags(
     assert transaction_span["tags"] == expected_tags
 
     spans_consumer.assert_empty()
+
+
+def test_span_filtering_with_generic_inbound_filter(
```
iambriccardo, line 1693: Created just one integration test for the end-to-end behavior since i tested other filters directly in unit tests.

PR #3554, opened 2024-05-06
feat(ai-monitoring): Calculate an AI model cost metric per-span

colin-sentry:
AI models have different prices, for example OpenAI's implementation of GPT-4 costs 30$ per million tokens of input, and 60$ per million dollars of output.

We'd like to be able to make graphs of how much money you've spent on your AI projects, grouped by, e.g., AI pipeline name.

It's messy to put this logic in SDKs, as we can't update those once launched. If we put the logic in Sentry, then any price changes would retroactively apply.


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -637,6 +638,45 @@ pub fn extract_measurements(span: &mut Span, is_mobile: bool) {
...
+                completion_tokens_used,
+                total_tokens_used,
+            ) {
+                span.measurements
+                    .get_or_insert_with(Default::default)
+                    .insert(
+                        "ai_total_cost".to_owned(),
+                        Measurement {
+                            value: total_cost.into(),
+                            unit: MetricUnit::None.into(),
```
iker-barriocanal: Docs of `calculate_ai_model_cost` say it's dollars. Should the unit include it?

relay-event-normalization/src/normalize/ai_model_costs.rs
```diff
@@ -0,0 +1,176 @@
+//! This file maps AI model IDs to their respective costs.
+//! Adapted from <https://github.com/langchain-ai/langchain/blob/5b6d1a907df8a7af0f47c5ffd394b3673d9539e5/libs/community/langchain_community/callbacks/openai_info.py#L4>
+
+/// Cost is in US dollars per thousand tokens
+fn calculate_cost_per_1k_tokens(model_id: &str, for_completion: bool) -> Option<f64> {
```
iker-barriocanal: Do non-US regions pay with different currencies?

iker-barriocanal: What does "for_completion" mean? Is it input/output?

iker-barriocanal: Are there plans to keep these costs up-to-date in case the price changes?

relay-event-normalization/src/normalize/ai_model_costs.rs
```diff
@@ -0,0 +1,176 @@
...
+    match (model_id, for_completion) {
+        // GPT-4 input
+        ("gpt-4", false) => Some(0.03),
```
iker-barriocanal: Are all values smaller than 1? If yes, have we considered forwarding integers and updating the unit to cents, to avoid floats?

iker-barriocanal (commented):
Could you expand the PR description to include some context around this PR?

I wonder if, instead of tracking the cost that is susceptible to change, we could track a more stable number like the tokens count. Estimating the cost from there is straightforward and more resilient for billing changes.

Dav1dde:
This looks like a function over the amount of tokens and model, why do we need this is a dedicated metric, why can we not do this at query time, which would work even if values are changing.

relay-event-normalization/src/normalize/span/tag_extraction.rs
colin-sentry: We don't have any money units - I was unsure what the policy was for creating net new units

relay-event-normalization/src/normalize/ai_model_costs.rs
colin-sentry: Everyone pays USD, or their local equivalent converted

colin-sentry: Some models charge more for "completion tokens" than "prompt tokens", i.e., the model responding is more expensive than asking a question.

colin-sentry: Generally prices stay pretty much the same and new models have different prices, the UI mentions that we don't collect prices for all models.

relay-event-normalization/src/normalize/ai_model_costs.rs
colin-sentry: Since it's a cost per thousand tokens, you end up with floating point math at multiplication time anyways - since prompts won't be even multiples of a thousand

colin-sentry:
> This looks like a function over the amount of tokens and model, why do we need this is a dedicated metric, why can we not do this at query time, which would work even if values are changing.

 It could potentially be done at query time, but then all of this complexity goes into the monolith - it didn't really feel any cleaner, and would definitely be slower.

colin-sentry:
> Could you expand the PR description to include some context around this PR?
> 
> I wonder if, instead of tracking the cost that is susceptible to change, we could track a more stable number like the tokens count. Estimating the cost from there is straightforward and more resilient for billing changes.

Done.

Dav1dde:
> It could potentially be done at query time, but then all of this complexity goes into the monolith - it didn't really feel any cleaner, and would definitely be slower.

And in Relay it's better, where it is distributed (forever) to customers running their own Relays (meaning it can't even be updated if it is wrong)?

With additional storage costs for the metrics on our end.

colin-sentry:
> > It could potentially be done at query time, but then all of this complexity goes into the monolith - it didn't really feel any cleaner, and would definitely be slower.
> 
> And in Relay it's better, where it is distributed (forever) to customers running their own Relays?
> 
> With additional storage costs for the metrics on our end.

I thought end users didn't usually do processing, isn't this usually done within our infra, even if users have their own POPs?

Dav1dde:
Metrics extraction runs on managed Relays.

Apart from that, I don't think so much hardcoded information which will need continuous updates and possibly adjustments should be in Relay. Either we pass it down from Sentry or you can just do it at query time.

@jjbayer wdyt?

jjbayer (commented):
> but then all of this complexity goes into the monolith

@colin-sentry as you pointed out, the complexity will exist somewhere, be it in Relay or in Sentry. The one reason to not put it in relay is that relay, just like SDKs, run on customers' premises and we have no control over when they update.

There are several options to keep relay flexible:

1. Only emit very simple metrics from Relay and do all arithmetics at query time.
2. Define the formulas in Relay, but use a config from Sentry to parametrize it. This is [what we do for performance scores](https://github.com/getsentry/relay/blob/111f6c823f09d253b44ff02b70af847c1a25b921/relay-dynamic-config/src/project.rs#L56-L58). E.g. the model-to-cost mapping could become part of [global config](https://github.com/getsentry/relay/blob/111f6c823f09d253b44ff02b70af847c1a25b921/relay-dynamic-config/src/global.rs#L23).
3. Try to extend the existing metrics extraction configuration to support arithmetic, so all the math can be defined as configuration.

Given that the math for calculating the cost is pretty straight forward, I would personally go for Option 1. I'm also happy to sync on this in a meeting if it makes things easier.

colin-sentry:
> > but then all of this complexity goes into the monolith
> 
> @colin-sentry as you pointed out, the complexity will exist somewhere, be it in Relay or in Sentry. The one reason to not put it in relay is that relay, just like SDKs, run on customers' premises and we have no control over when they update.
> 
> There are several options to keep relay flexible:
> 
> 1. Only emit very simple metrics from Relay and do all arithmetics at query time.
> 2. Define the formulas in Relay, but use a config from Sentry to parametrize it. This is [what we do for performance scores](https://github.com/getsentry/relay/blob/111f6c823f09d253b44ff02b70af847c1a25b921/relay-dynamic-config/src/project.rs#L56-L58). E.g. the model-to-cost mapping could become part of [global config](https://github.com/getsentry/relay/blob/111f6c823f09d253b44ff02b70af847c1a25b921/relay-dynamic-config/src/global.rs#L23).
> 3. Try to extend the existing metrics extraction configuration to support arithmetic, so all the math can be defined as configuration.
> 
> Given that the math for calculating the cost is pretty straight forward, I would personally go for Option 1. I'm also happy to sync on this in a meeting if it makes things easier.

Here are a few thoughts:

- If we do this outside of relay, we'll need to tag every tokens_used counter with the model it came from, which adds cardinality
- SneQL itself doesn't have an easy way to do a match statement (as far as I know), so this processing will need to be done in python for every page load, which seems suboptimal
- A middle-of-the-road solution might be to extract the model ID as a tag for the counter, then in the "last" relay (the ones in our infra), remove that tag and convert it into a derived counter - this would avoid adding the cardinality while also not requiring customers to upgrade things to get new costs. I don't know enough about relay to architect that, though.

relay-dynamic-config/src/ai.rs
```diff
@@ -1,73 +0,0 @@
-//! Configuration for measurements generated from AI model instrumentation.
```
colin-sentry, line 1: This had to be moved into normalization to avoid a circular import

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - Emit negative outcomes when metrics are rejected because of a disabled namespace. ([#3544](https://github.com/getsentry/relay/pull/3544))
 - Add AI model costs to global config. ([#3579](https://github.com/getsentry/relay/pull/3579))
 - Add support for `event.` in the `Span` `Getter` implementation. ([#3577](https://github.com/getsentry/relay/pull/3577))
+- Add a calculated measurement based on the AI model and the tokens used ([#3554](https://github.com/getsentry/relay/pull/3554))
```
```suggestion
- Add a calculated measurement based on the AI model and the tokens used. ([#3554](https://github.com/getsentry/relay/pull/3554))
```

relay-event-normalization/src/event.rs
```diff
@@ -2087,6 +2095,111 @@ mod tests {
...
+        assert_eq!(
+            spans
+                .first()
+                .unwrap()
+                .value()
+                .unwrap()
+                .measurements
+                .value()
+                .unwrap()
+                .get_value("ai_total_cost"),
```
jjbayer, line 2186: You can use the `get_value!` macro here.

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -244,6 +290,28 @@ mod tests {
...
+                    model_id: LazyGlob("babbage-002.ft-*"),
+                    for_completion: false,
+                    cost_per_1k_tokens: 0.0016,
+                },
+            ],
+        }
+        "###);
+
+        let serialized = serde_json::to_string(&deserialized).unwrap();
+        // Patch floating point
```
jjbayer: I think I forgot to remove this comment in the previous version.

```suggestion
```

relay-event-normalization/src/normalize/mod.rs
```diff
@@ -244,6 +290,28 @@ mod tests {
...
+        std::assert_eq!(&serialized, original);
```
jjbayer:
```suggestion
        assert_eq!(&serialized, original);
```

relay-event-normalization/src/normalize/span/ai.rs
```diff
@@ -0,0 +1,97 @@
...
+) -> Option<f64> {
+    if let Some(prompt_tokens) = prompt_tokens_used {
+        if let Some(completion_tokens) = completion_tokens_used {
+            let mut result = 0.0;
+            if let Some(cost_per_1k) = ai_model_costs.cost_per_1k_tokens(model_id, false) {
+                result += cost_per_1k * (prompt_tokens / 1000.0)
+            }
+            if let Some(cost_per_1k) = ai_model_costs.cost_per_1k_tokens(model_id, true) {
+                result += cost_per_1k * (completion_tokens / 1000.0)
+            }
```
jjbayer, line 23: nit: There's an opportunity to call `cost_per_1k_tokens` only once (and iterate the list only once) if we make it return a tuple `(/*prompt tokens*/, Option<f32>, Option<f32> /*completion tokens*/)`. But I'm not sure if it increases readability.

relay-event-normalization/src/normalize/span/ai.rs
```diff
@@ -0,0 +1,97 @@
...
+/// which would cause a circular dependency.
+pub fn extract_ai_measurements(span: &mut Span, ai_model_costs: &ModelCosts) {
+    let Some(span_op) = span.op.value() else {
+        return;
+    };
+
+    if !span_op.starts_with("ai.") {
+        return;
+    }
+    let total_tokens_used = span
```
jjbayer: nit: 

```rust
let Some(measurements) = span.measurements.value() else { return };
let total_tokens_used = measurements.get_value("ai_total_tokens_used");
// ...
```

relay-event-normalization/src/normalize/span/ai.rs
```diff
@@ -0,0 +1,97 @@
...
+            .cost_per_1k_tokens(model_id, false)
+            .map(|cost| cost * (total_tokens / 1000.0))
+    } else {
+        None
+    }
+}
+
+/// Extract the ai_total_cost measurement into the span.
+/// Note that this can't go in relay_event_normalization because it uses relay_dynamic_config
+/// which would cause a circular dependency.
```
jjbayer:
```suggestion
```

relay-event-normalization/src/event.rs
```diff
@@ -2087,6 +2095,111 @@ mod tests {
         "###);
     }
 
+    #[test]
+    fn test_ai_metrics() {
```
jjbayer: This test does not cover extracted metrics, right?

```suggestion
    fn test_ai_measurements() {
```

jjbayer (approved):

relay-event-normalization/src/event.rs
colin-sentry: I can never get it to work, will just leave it this way since it's in a test

relay-event-normalization/src/normalize/span/ai.rs
colin-sentry: I don't think that's really better, will leave it as-is

relay-event-normalization/src/normalize/span/ai.rs
colin-sentry: 👍

PR #3556, opened 2024-05-06
feat(spans): Extracts messaging.message.id for queue spans

edwardgou-sentry:
Extracts messaging.message.id for queue spans

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/tag_extraction.rs
+++ b/relay-event-normalization/src/normalize/span/tag_extraction.rs
@@ -688,6 +688,7 @@ pub fn extract_measurements(span: &mut Span, is_mobile: bool) {
                     &data.messaging_message_body_size,
                     "messaging.message.body.size",
                 ),
+                (&data.messaging_message_id, "messaging.message.id"),
             ] {
                 if let Some(value) = match field.value() {
                     Some(Value::F64(f)) => Some(*f),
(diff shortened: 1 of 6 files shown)


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -688,6 +688,7 @@ pub fn extract_measurements(span: &mut Span, is_mobile: bool) {
                     &data.messaging_message_body_size,
                     "messaging.message.body.size",
                 ),
+                (&data.messaging_message_id, "messaging.message.id"),
```
jjbayer: This will try to convert `data.messaging_message_id` to a floating point number and insert it as `measurement.id`, which will always fail.

jjbayer (changes requested):
I assume you want `message_id` to show up in Span storage, but neither as a `measurement` nor as a `sentry_tag` (i.e. indexable)? @phacops is working on getting `span.data` into storage. With that, this PR would become unnecessary.

If you require `message_id` to be indexed, you can add it to `sentry_tags` instead.

relay-event-normalization/src/normalize/span/tag_extraction.rs
edwardgou-sentry: sorry, i meant to extract this in `sentry_tags` 😅 . Removed this and added `message_id` logic under `extract_tags`. Thanks for catching this!

jjbayer (approved):

PR #3557, opened 2024-05-06
chore(self-hosted): Mark e2e test check as required

hubertdeng123:
It's been in CI for a week and it looks like it's working as intended with no flakes.
#skip-changelog


iker-barriocanal (approved):
I haven't seen any flakes 👍

PR #3561, opened 2024-05-07
feat(metrics): Forward received_at field to Kafka

### LINKED ISSUE #3515: Propagate `received_at` field of `BucketMetadata` to Kafka and consume it

iambriccardo:
This PR forwards the `received_at` field of `BucketMetadata` to Kafka. An important thing to notice is that the current Kafka schema allows additional fields: https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/ingest-metrics.v1.schema.json meaning that this PR could be deployed without any side effects on the schemas, however, it would be best to mark the field as optional in the Kafka schema itself.

The PR also introduces new facilities to relax timestamp assertions which are generally flaky.

Closes: https://github.com/getsentry/relay/issues/3515


tests/integration/test_store.py
```diff
@@ -725,6 +725,7 @@ def send_buckets(buckets):
     produced_buckets.sort(key=lambda b: (b["name"], b["value"]))
     for bucket in produced_buckets:
         del bucket["timestamp"]
+        del bucket["received_at"]
```
iambriccardo, line 730: This test was very tricky to update, we might want to follow up with a PR to fix timestamps across all of our tests.

jjbayer (approved):

relay-server/src/services/store.rs
```diff
@@ -1290,6 +1289,8 @@ struct MetricKafkaMessage<'a> {
     timestamp: UnixTimestamp,
     tags: &'a BTreeMap<String, String>,
     retention_days: u16,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
Dav1dde:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
```

Default is implied for options

tests/integration/asserts/time.py
```diff
@@ -0,0 +1,35 @@
+from datetime import timedelta, datetime, timezone
+
+
+class _WithinBounds:
+
```
Dav1dde, line 5: Surprised the formatter doesn't complain about this newline

tests/integration/asserts/time.py
```diff
@@ -0,0 +1,35 @@
...
+    def __eq__(self, other):
+        assert isinstance(other, int)
+        return self._lower_bound <= other <= self._upper_bound
+
+    def __str__(self):
+        return f"{self._lower_bound} <= x <= {self._upper_bound}"
+
+
+def time_after(lower_bound):
```
Dav1dde, line 18: This doesn't assert after, but after until now. I think making the upper bound on `time_within` optional and defaulting to now is fine.

tests/integration/test_metrics.py
```diff
@@ -1873,3 +1890,48 @@ def test_metrics_with_denied_names(
     assert volume_metric["tags"]["outcome.reason"] == "denied-name"
 
     assert "d:custom/memory_usage@byte" in metrics
+
+
+@pytest.mark.parametrize("mode", ["default", "chain"])
```
Dav1dde, line 1897: Chain only is good enough imo

tests/integration/test_store.py
Dav1dde: Can we not do this better now?

tests/integration/asserts/time.py
iambriccardo: Yeah, I was indeed thinking about a better name.

PR #3565, opened 2024-05-07
ref(metrics): Split project metric transforms into separate steps

Dav1dde:
The project cache always had two phases processing metrics, depending on whether the project was loaded:

No project state initially available:
```
-> pre aggregate --[project state available]--> filter metrics -> rate limit metrics -> push to aggregator
```

With project state:
```
-> filter metrics -> rate limit metrics -> push to aggregator
```

Now we need a separate filtering step independent from the project state but whether the received metrics are coming from a trusted internal source or not.

```
-> pre filter -> pre aggregate --[project state available]--> filter metrics -> rate limit metrics -> push to aggregator
-> pre filter -> filter metrics -> rate limit metrics -> push to aggregator
```

Since we're dealing with a lot of `Vec<Bucket>` flying around I split the two filtering steps out of the project into a `pre` and `post` step, which can now be tested, independently of the project. Allowing for better and easier tests. To make sure steps aren't missed `Vec<Buckets>` are no longer passed around but a wrapper type which contains compile time state information which transforms have been applied to the buckets already.

Long term I want to use this as an experiment to bring the `Buckets` structure to the rest of Relay. Have a properly managed structure for dealing with buckets, make it impossible to throw a way buckets without outcomes for them.

Contains changes of: #3566

#skip-changelog


relay-server/src/services/project/metrics/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+mod post;
+mod pre;
+
+pub use self::post::PostProjectTransform;
+pub use self::pre::PreProjectTransform;
+
+/// Transforms [`Buckets`] into different [`Buckets`] applying pre-defined transforms.
+pub trait Transformer<S> {
+    /// The new state the buckets will be transformed into.
+    type NewState;
```
jjbayer: Should we rename `S` and `NewState` to something like `StateIn` `StateOut`, or `InputState` `OutputState`?

relay-server/src/services/project/metrics/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    type IntoIter = std::vec::IntoIter<Bucket>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.buckets.into_iter()
+    }
+}
+
+// State in a private module, to make sure no one outside the module has access to it.
+mod state {
+    pub struct PreProject;
```
jjbayer: Why not just make this a private struct?

relay-server/src/services/project/metrics/post.rs
```diff
@@ -0,0 +1,283 @@
...
+
+use crate::metric_stats::MetricStats;
+use crate::services::outcome::{Outcome, TrackOutcome};
+use crate::services::project::ProjectState;
+use crate::utils;
+
+use super::{state, Transformer};
+
+/// Applies an initial filter to incoming metrics using the project state.
+pub struct PostProjectTransform<'a> {
```
jjbayer: Would be nice to give this a descriptive name (i.e. what it does, not when it's called).

relay-server/src/services/project/mod.rs
```diff
@@ -680,8 +594,21 @@ impl Project {
             return;
         }
 
+        let buckets = buckets.transform(metrics::PostProjectTransform {
```
jjbayer: IMO it would be slightly easier to understand if this was a plain old standalone function, i.e.

`fn post_project_transform(buckets: Buckets<PreProject>) -> Buckets<PostProject>`

But there might be benefits to the transformer pattern that I'm not yet seeing.

relay-metrics/src/aggregator.rs
```diff
@@ -459,6 +460,27 @@ impl fmt::Debug for CostTracker {
     }
 }
 
+/// The minimum trait requirements an [`Aggregator`] can work with.
+///
+/// The aggregator accepts buckets with this type and also returns
+/// aggregated with the same type again.
```
```suggestion
/// aggregated buckets with the same type again.
```

relay-server/src/services/project/metrics.rs
```diff
@@ -0,0 +1,333 @@
...
+use crate::services::project::ProjectState;
+use crate::services::project_cache::BucketSource;
+use crate::utils;
+
+pub struct Filtered;
+pub struct WithProjectState;
+
+/// Container for a vector of buckets.
+#[derive(Debug)]
+pub struct Buckets<State = ()> {
```
jjbayer: Does a default for this type parameter even make sense? I would expicitly require an init state instead, e.g. `Buckets<Unfiltered>`.

relay-server/src/services/project/metrics.rs
```diff
@@ -0,0 +1,333 @@
...
+        self.migrate()
+    }
+}
+
+impl Buckets<Filtered> {
+    pub fn apply_project_state(
+        mut self,
+        outcome_aggregator: &Addr<TrackOutcome>,
+        metric_stats: &MetricStats,
+        state: &ProjectState,
```
jjbayer: nit: 
```suggestion
        project_state: &ProjectState,
```

relay-server/src/services/project/metrics.rs
```diff
@@ -0,0 +1,333 @@
...
+                outcome_aggregator,
+                metric_stats,
+                utils::extract_metric_quantities(&denied_buckets, mode),
+                scoping,
+                Outcome::Filtered(FilterStatKey::DeniedName),
+                &denied_buckets,
+            );
+        }
+
+        self.migrate()
```
jjbayer: nit: Even though it's not required, I would inlude the type parameter here, just to make extra clear that this is a state transition:

```suggestion
        self.migrate_to::<WithProjectState>()
```

relay-metrics/src/aggregator.rs
```diff
@@ -459,6 +460,27 @@ impl fmt::Debug for CostTracker {
...
+///
+/// The aggregator being generic over the contained buckets allows
+/// type safety when aggregating buckets in different states.
+pub trait AggregatorBuckets: Sized + IntoIterator<Item = Bucket> {
```
jjbayer: Could the Aggregator, instead of having this trait, simply be marked by the bucket state? IIUC the aggregator always outputs the same bucket type as it gets as input, so for example an `Aggregator<Filtered>` would be an aggregator that aggregates buckets that have already been filtered.

jjbayer (commented):
Looks good, only two more concerns (see comments):

1. explicitness of states
2. potentially getting rid of `AggregatorBuckets`

relay-server/src/services/project/metrics.rs
Dav1dde: No state is the init state and the default at the same time, you can pass around `Buckets` if you don't care about state until you do. Having something explicit didn't really make it better imo.

relay-server/src/services/project/metrics.rs
Dav1dde: I think I'll just remove the method then.

relay-metrics/src/aggregator.rs
Dav1dde: Yes, but that requires changing not just the pre aggregation but all aggregation and moving `Buckets` into `relay-metrics`. I don't mind doing it, I didn't to keep the scope smaller.

jjbayer (approved):

PR #3567, opened 2024-05-07
ref(server): Make health_check fully async

jan-auer:
Moves actual health checks into a continuous background process so that
calls to the health check service always succeed instantly. If the
background status has not been updated for longer than a threshold, the
health check is assumed unhealthy.

This PR keeps the individual timeouts around the four probes (where only
three of them can actually time out). In a follow-up, we can consider to
further split down each of the probes and keep separate states.


relay-server/src/services/health_check.rs
```diff
@@ -182,127 +198,73 @@ impl HealthCheckService {
...
 
-        if self.is_shutting_down.load(Ordering::Relaxed) {
+        if shutdown.get().is_some() {
             return Status::Unhealthy;
         }
 
-        let (sys_mem, auth, agg, proj) = tokio::join!(
-            self.probe("system memory", self.system_memory_probe()),
+        // TODO(ja): log like probe
+        let sys_mem = self.system_memory_probe();
```
jan-auer: I'll address this TODO after an initial round of feedback. 

The issue here is that the system memory probe requires mutable access to the `System` instance, which means we cannot join it with other futures. However, since it is entirely sync we don't need to do that and thus we can avoid a redundant Mutex.

relay-server/src/endpoints/mod.rs
```diff
@@ -94,7 +94,7 @@ where
         .route("/api/:project_id/minidump/", minidump::route(config))
         .route("/api/:project_id/events/:event_id/attachments/", attachments::route(config))
         .route("/api/:project_id/unreal/:sentry_key/", unreal::route(config))
-        .route("/api/:project_id/spans/", spans::route(config))
```
Dav1dde: 👀

relay-server/src/services/health_check.rs
```diff
@@ -182,127 +196,74 @@ impl HealthCheckService {
...
 }
 
 impl Service for HealthCheckService {
     type Interface = HealthCheck;
 
-    fn spawn_handler(self, mut rx: relay_system::Receiver<Self::Interface>) {
-        let service = Arc::new(self);
+    fn spawn_handler(mut self, mut rx: relay_system::Receiver<Self::Interface>) {
+        let (status_tx, status_rx) = watch::channel(Statuses::healthy());
+        let check_interval = self.config.health_sys_info_refresh_interval();
```
Dav1dde: Don't think we should attach this to the config option, the debounce was deliberate before but now we can just get rid of it and have a single delay and not tie the refresh time to the `health_sys` option.

Maybe rename the option or delete it alltogether.

relay-server/src/services/health_check.rs
```diff
@@ -182,127 +196,74 @@ impl HealthCheckService {
...
-                    metric!(gauge(RelayGauges::NetworkOutage) = u64::from(is_outage));
-                }
-            });
-        }
-
-        if matches!(message, IsHealthy::Liveness) {
-            return Status::Healthy;
+            if let Ok(is_outage) = self.upstream_relay.send(IsNetworkOutage).await {
+                metric!(gauge(RelayGauges::NetworkOutage) = u64::from(is_outage));
+            }
```
Dav1dde: I just realized that this doesn't actually have an influence on the status.

relay-server/src/services/health_check.rs
```diff
@@ -182,127 +196,74 @@ impl HealthCheckService {
...
-        let service = Arc::new(self);
+    fn spawn_handler(mut self, mut rx: relay_system::Receiver<Self::Interface>) {
+        let (status_tx, status_rx) = watch::channel(Statuses::healthy());
+        let check_interval = self.config.health_sys_info_refresh_interval();
+        // Add 10% buffer to the internal timeouts to avoid race conditions.
+        let status_timeout = (check_interval + self.config.health_probe_timeout()).mul_f64(1.1);
 
         tokio::spawn(async move {
-            let mut shutdown = Controller::shutdown_handle();
+            let shutdown = Controller::shutdown_handle();
```
Dav1dde, line 233: I believe this can just be a field on the struct, and then we don't have to awkwardly pass it to `check_readiness`.

Dav1dde (commented):
Since the status is now just read from a `RwLock`/`watch`, we can do the same we do for global configs and give the endpoint a `HealthCheckHandle` which can read the status directly from the `watch`, which makes checking the health check actually sync without any message queues.

relay-server/src/services/health_check.rs
jan-auer: Agreed, I personally preferred the form of passing it in, but if you prefer a field I'm happy to change.

Dav1dde: I'd prefer the field, but I am fine with it either way!

relay-server/src/services/health_check.rs
Dav1dde: Mh since we wait for the response now, should we also give it a timeout (wrap it in `probe()` and always return `Healthy`)?

relay-server/src/endpoints/mod.rs
jan-auer: Thanks, no idea how that happened...

relay-server/src/services/health_check.rs
jan-auer: That's a good point. We still debounce memory checks, but now we debounce everything else with it. Per-se, I'd like to keep two separate configs (for us to tweak if necessary, not for customers to change), one for a check interval and one for a delay.

What if we rename + shim the option?

relay-server/src/services/health_check.rs
jan-auer: As I moved this code around, I asked myself the question why we put this in here in the first place. We could equally move this into the upstream service's outage monitor and log changes at the source. I *think* this was done so that we keep updating the value even if nothing changes. Not sure what the right placement is.

Agreed, another probe is nice because it parallelizes.

relay-server/src/services/health_check.rs
jan-auer: I've reviewed the logic now and noticed that we can also use shutdown to break the monitoring task. If we do that, we just need to set the last update to unhealthy -- once it times out we simply remain unhealthy. Also, liveness is _always_ healthy regardless of whether we're shutting down or not. 

I ended up removing the parameter now, check shutdown exclusively outside, and restructured the decision tree a bit more naturally.

relay-server/src/services/health_check.rs
```diff
@@ -182,127 +208,61 @@ impl HealthCheckService {
...
-        if let Some(cgroup) = self.system.cgroup_limits() {
-            Memory {
-                used: cgroup.total_memory.saturating_sub(cgroup.free_memory),
-                total: cgroup.total_memory,
-            }
-        } else {
-            Memory {
-                used: self.system.used_memory(),
-                total: self.system.total_memory(),
+                tokio::time::sleep(check_interval).await;
```
jan-auer, line 242: Note that in many other places we instead create a `tokio::time::interval` and then use `tokio::select!` with the shutdown handle and the interval. I opted for a different solution here as we don't need to stick to strict intervals and this version ended up less code. The alternative would look like this:

```rust
let mut ticker = tokio::time::interval(check_interval);
let mut shutdown = Controller::shutdown_handle();

loop {
    let _ = update_tx.send(StatusUpdate::new(relay_statsd::metric!(
        timer(RelayTimers::HealthCheckDuration),
        type = "readiness",
        { self.check_readiness().await }
    )));

    tokio::select! {
        biased;
        _ = ticker.tick() => (),
        _ = shutdown.notified() => break,
    }
}
```

relay-server/src/services/health_check.rs
```diff
@@ -182,127 +208,61 @@ impl HealthCheckService {
...
-    }
 }
 
 impl Service for HealthCheckService {
     type Interface = HealthCheck;
 
-    fn spawn_handler(self, mut rx: relay_system::Receiver<Self::Interface>) {
-        let service = Arc::new(self);
+    fn spawn_handler(mut self, mut rx: relay_system::Receiver<Self::Interface>) {
+        let (update_tx, update_rx) = watch::channel(StatusUpdate::healthy());
```
jan-auer: Should we initialize with unhealthy?

Dav1dde: Ideally yes, kubernetes should only send us traffic once we're known ready, that is the purpose of the readiness probe. Although this will show up on our graphs as unhealthy then.

relay-config/src/config.rs
```diff
@@ -1341,10 +1341,11 @@ impl Default for CardinalityLimiter {
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(default)]
 pub struct Health {
-    /// Interval in which Relay will refresh system information, like current memory usage.
+    /// Interval in which Relay will refresh internal health checks.
     ///
     /// Defaults to 3 seconds.
-    pub sys_info_refresh_interval_secs: u64,
+    #[serde(alias = "sys_info_refresh_interval_secs")]
```
Dav1dde: I think it's fine to drop this, stuff like this stays around for too long.

relay-config/src/config.rs
```diff
@@ -2292,8 +2293,8 @@ impl Config {
     }
 
     /// Interval to refresh system information.
```
Dav1dde: This is now outdated.

relay-server/src/services/health_check.rs
```diff
@@ -182,127 +208,61 @@ impl HealthCheckService {
...
             }
-        }
-    }
 
-    fn refresh(&mut self) {
-        if self.last_refresh.elapsed() >= self.refresh_interval {
-            self.system
-                .refresh_memory_specifics(MemoryRefreshKind::new().with_ram());
+            // Shutdown marks readiness health check as unhealthy.
+            update_tx.send(StatusUpdate::new(Status::Unhealthy)).ok();
```
Dav1dde, line 246:
```suggestion
            let _ = update_tx.send(StatusUpdate::new(Status::Unhealthy));
```

Converting to an option just to get rid of the `#[must_use]` is a smell imo, just ignore the error explicitly.

Dav1dde (approved):

relay-server/src/services/health_check.rs
jjbayer: That would make sense to me. We'll have to make sure that our health checks are configured in a way that they allow a start up container to be unhealthy for up to 3 seconds.

relay-server/src/services/health_check.rs
```diff
@@ -182,127 +208,61 @@ impl HealthCheckService {
...
-        if let Some(cgroup) = self.system.cgroup_limits() {
-            Memory {
-                used: cgroup.total_memory.saturating_sub(cgroup.free_memory),
-                total: cgroup.total_memory,
-            }
-        } else {
-            Memory {
-                used: self.system.used_memory(),
-                total: self.system.total_memory(),
+                tokio::time::sleep(check_interval).await;
```
jjbayer, line 242: With the `join!`, if one health check times out, other unrelated health checks will be delayed as well IIUC. Unless the timeout is always lower than the check interval.

We could make this a `sleep_until(next_check_time)` to be a little more robust against timeouts spilling over the check interval.

jjbayer (approved):

relay-config/src/config.rs
jan-auer: 👍  Checked now that it's not used, removing

relay-server/src/services/health_check.rs
jan-auer: First health check runs immediately

relay-server/src/services/health_check.rs
jan-auer: We use both extensively in the code base, `.ok()` is definitely a widely used pattern

relay-server/src/services/health_check.rs
jan-auer: We can use the pattern from other places if you prefer, but please check https://github.com/getsentry/relay/pull/3567#discussion_r1592825978

The validity timeout is the interval + poll timeout, so we'd never timeout prematurely.

PR #3571, opened 2024-05-08
ref(normalization): Emit sentry errors on normalization decisions and make feature flags and configs NOOP

iker-barriocanal:
This PR does the following:
- Run normalization unconditionally on all relays. Full normalization is only run in processing relays.
- Remove integration tests on feature flags and static config. I'll reintroduce these once the behavior is reintroduced.
- Computes normalization decisions only for observability purposes (see point below).
- Emits a sentry error on processing error/default events in both PoPs and Relays. A global config controls the sampling rate at which the errors are emitted. These errors contain tags and context that will be used to determine the correctness of decisions, and these are added to the Sentry SDK scope throughout the envelope processing.

Context: there's some unexpected behavior, that may be caused by an inaccurate normalization run decision, and results in error events with `type=default`. Source is unknown, and adding observability should provide enough data to confirm and scope down the issues.

**Note**: this PR should land master _before_ the next Relay release on May 15. Currently, normalization is controlled by feature flags that get propagated from internal relays and force them to run full event normalization, including steps that break backward compatibility. As a result, updates to the protocol or normalization may not be effective. Merging this PR solves this issue.

#skip-changelog


relay-server/src/services/processor.rs
```diff
@@ -1292,17 +1293,36 @@ impl EnvelopeProcessorService {
...
+            if let Some(event) = state.event.value() {
+                scope.set_extra(
+                    "platform",
+                    event.platform.as_str().unwrap_or("<not set>").into(),
+                );
+                scope.set_extra("sdk", event.sdk_name().into());
+            }
+        });
+
+        let full_normalization = self.inner.config.processing_enabled();
```
Dav1dde, line 1324: This used to be:

```
self.inner.config.processing_enabled()
            && options.processing_disable_normalization
            && (state.envelope().meta().is_from_internal_relay())
```

Just to make sure, this is the intended behaviour?

Dav1dde:
I think we should consider attaching the normalization status to the item instead (what Joris proposed). This would improve consistency and is easier to debug.

iker-barriocanal:
I'll update the implementation to use item headers instead. I'll follow up with a new PR.

PR #3574, opened 2024-05-10
build(py): Python 3.11 and Python 3.12 compatibility

Dav1dde:
A bunch of stuff is outdated and with Py 3.12 a lot of it was removed/deprecated, fixes all the deprecations I saw. Flake also had issues in the version we used and mis-reported some stuff, which is also fixed with the update.

#skip-changelog


jjbayer (approved):
Thanks!

Dav1dde:
@jan-auer can you mark the 3.10 check no longer as required?

jan-auer (approved):

PR #3579, opened 2024-05-10
feat(spans): Global config for AI model costs

jjbayer:
See https://github.com/getsentry/relay/pull/3554.

#skip-changelog


relay-dynamic-config/src/global.rs
```diff
@@ -46,6 +46,9 @@ pub struct GlobalConfig {
     /// applying.
     #[serde(skip_serializing_if = "is_ok_and_empty")]
     pub metric_extraction: ErrorBoundary<MetricExtractionGroups>,
+
+    /// Configuration for AI span measurements.
+    pub ai_model_costs: ErrorBoundary<ai::ModelCosts>,
```
jjbayer, line 53: Will add a `skip_serializing_if` here.

relay-dynamic-config/src/ai.rs
```diff
@@ -0,0 +1,78 @@
...
+            .find(|cost| cost.matches(model_id, for_completion))
+            .map(|c| c.cost_per_1k_tokens)
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ModelCost {
+    model_id: LazyGlob,
+    for_completion: bool,
```
Dav1dde, line 35: Maybe this makes sense as an enum, but not familiar enough with AI terminology:

```rs
enum Usage {
    Completion,
    NotCompletionButTheDefaultUsecase,
    // MaybeSomeFutureUsecase,
}
```

@colin-sentry wdyt?

relay-dynamic-config/src/global.rs
Dav1dde: 👍

Dav1dde (approved):

relay-dynamic-config/src/ai.rs
colin-sentry: the SDKs only send the static `prompt_tokens`, `completion_tokens`, and `total_tokens`, so it's probably fine to keep as a boolean

PR #3582, opened 2024-05-13
ref(metric-stats): Rate limit before cardinality limiting

Dav1dde:
For enforced quotas we do not want to track cardinality.

#skip-changelog


iambriccardo (approved):
LGTM

PR #3587, opened 2024-05-13
feat(ai-monitoring): Extract AI metrics

colin-sentry:
Read some AI-model to cost mappings from upstream global config, use those to produce a "total AI cost" metric


relay-server/src/services/processor/span/processing.rs
```diff
@@ -464,12 +473,14 @@ fn normalize(
...
         performance_score,
         measurements,
+        ai_model_costs,
         max_name_and_unit_len,
     } = config;
 
     // This follows the steps of `NormalizeProcessor::process_event`.
     // Ideally, `NormalizeProcessor` would execute these steps generically, i.e. also when calling
     // `process` on it.
+    println!("In normalize function");
```
colin-sentry, line 481: @phacops this line never gets printed

PR #3588, opened 2024-05-13
ref(profiles): Ensure UUIDs for chunk and profiler and sort samples

phacops:


relay-profiling/src/sample/v2.rs
```diff
@@ -95,6 +96,9 @@ impl ProfileData {
             return Err(ProfileError::NotEnoughSamples);
         }
 
+        // Remove NaN as a side-effect
+        self.sort_samples_by_timestamp();
```
Dav1dde: You can use `FiniteF64` for the timestamp instead which means you don't have `NaN`'s in the first place.

relay-profiling/src/sample/v2.rs
```diff
@@ -145,6 +149,12 @@ impl ProfileData {
         self.thread_metadata
             .retain(|thread_id, _| thread_ids.contains(thread_id));
     }
+
+    fn sort_samples_by_timestamp(&mut self) {
+        self.samples.retain(|s| !s.timestamp.is_nan());
+        self.samples
+            .sort_by(|a, b| a.timestamp.partial_cmp(&b.timestamp).unwrap());
```
Dav1dde:
```suggestion
            .sort_by(|a, b| a.timestamp.total_cmp(&b.timestamp));
```

relay-profiling/src/sample/v2.rs
```diff
@@ -95,6 +98,8 @@ impl ProfileData {
             return Err(ProfileError::NotEnoughSamples);
         }
 
+        self.sort_samples_by_timestamp();
```
Dav1dde: I'd just do this inline now.

relay-profiling/src/sample/v2.rs
```diff
@@ -165,4 +176,35 @@ mod tests {
...
+                Sample {
+                    stack_id: 0,
+                    thread_id: "1".to_string(),
+                    timestamp: FiniteF64::new(1000.0).unwrap(),
+                },
+            ],
+            ..Default::default()
+        };
+
+        chunk.sort_samples_by_timestamp();
```
Dav1dde: This should call `normalize`, the public interface not the implementation detail (e.g. the test doesn't make sure normalize calls into the sort function)

Dav1dde (approved):

PR #3592, opened 2024-05-14
ref(metrics): Log metric outcomes via dedicated struct

Dav1dde:
Replaces `utils::reject_metrics` with a dedicated struct responsible to log outcomes for metrics. Also updates the rate limiter to use the new struct, which surfaced a bug where we did not have outcomes for spans in `reject_metrics`.

Follow ups:
- Always take ownership of buckets when tracking outcomes, this should be the last step in the lifecycle of a bucket
- Implement proper outcomes in the batch metric buckets callback
- Fix the metadata situation, metadata of views should (for volume) needs to be handled differently

#skip-changelog

### DIFF
--- a/relay-server/src/metrics.rs
+++ b/relay-server/src/metrics.rs
@@ -127,7 +127,7 @@ impl MetricOutcomes {
 
     /// Tracks the cardinality of a metric.
     #[cfg(feature = "processing")]
-    pub fn track_cardinality(
+    pub fn cardinality(
         &self,
         scoping: Scoping,
         limit: &CardinalityLimit,
@@ -168,7 +168,7 @@ pub trait TrackableBucket {
     fn summary(&self, mode: ExtractionMode) -> BucketSummary;
 
     /// Metric bucket metadata.
-    fn metadata(&self) -> BucketMetadata;
+    fn metadata(&self) -> &BucketMetadata;
 }
 
 impl<T: TrackableBucket> TrackableBucket for &T {
@@ -184,7 +184,7 @@ impl<T: TrackableBucket> TrackableBucket for &T {
         (**self).summary(mode)
     }
 
-    fn metadata(&self) -> BucketMetadata {
+    fn metadata(&self) -> &BucketMetadata {
         (**self).metadata()
     }
 }
@@ -202,8 +202,8 @@ impl TrackableBucket for Bucket {
         BucketView::new(self).summary(mode)
     }
 
-    fn metadata(&self) -> BucketMetadata {
-        self.metadata
+    fn metadata(&self) -> &BucketMetadata {
+        &self.metadata
     }
 }
 
@@ -247,8 +247,8 @@ impl TrackableBucket for BucketView<'_> {
         }
     }
 
-    fn metadata(&self) -> BucketMetadata {
-        *self.metadata()
+    fn metadata(&self) -> &BucketMetadata {
+        self.metadata()
     }
 }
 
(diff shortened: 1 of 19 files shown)


relay-server/src/metrics.rs
```diff
@@ -0,0 +1,261 @@
...
+use crate::metric_stats::MetricStats;
+use crate::services::outcome::{Outcome, TrackOutcome};
+use crate::utils::ExtractionMode;
+#[cfg(feature = "processing")]
+use relay_cardinality::{CardinalityLimit, CardinalityReport};
+
+pub const PROFILE_TAG: &str = "has_profile";
+
+pub enum Quantities {
+    Extract(ExtractionMode),
```
jjbayer:
```suggestion
/// Indicates from where quantities should be taken.
pub enum Quantities {
    ExtractFromBuckets(ExtractionMode),
```

relay-server/src/services/processor.rs
```diff
@@ -2967,21 +2962,22 @@ impl UpstreamRequest for SendMetricsRequest {
...
-                            );
-                        }
+                    if error.is_received() {
+                        return;
+                    }
+
+                    for (scoping, quantities) in self.quantities {
+                        self.metric_outcomes.track(
+                            scoping,
+                            &[] as &[Bucket], // TODO(dav1d): provide buckets here
```
jjbayer, line 2974: Is this for a follow-up?

relay-server/src/services/project.rs
```diff
@@ -1126,22 +1128,12 @@ impl Project {
 
             if limits.is_limited() {
                 // NB: Until Vec::extract_if is stable, we have to iterate twice.
-                let matching_buckets = buckets
+                let _matching_buckets = buckets
```
jjbayer: Since iterators are lazy, isn't this a noop?

relay-server/src/utils/split_off.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[test]
+    fn test_split_off() {
+        let data = vec!["apple", "apple", "orange"];
+
+        let (apples, oranges) = split_off(data, |item| {
+            if item != "orange" {
+                Either::Left(item)
+            } else {
+                Either::Right(item)
+            }
```
jjbayer, line 54: Personally I would find this easier to use if the predicate returned `true` or `false`.

jjbayer (approved):

relay-server/src/utils/split_off.rs
Dav1dde: A boolean doesn't allow for a different type (partition + map)

relay-server/src/services/processor.rs
Dav1dde: Yes! We had the same problem before, it was the `None` in the `reject_metrics` call. I decided not to also add these functional changes to the PR, will do in a follow up.

relay-server/src/services/project.rs
Dav1dde: Good catch thanks, I think I messed up the merge here and I ran `clippy --fix` to fix a bunch of unused imports, lesson learned.

relay-server/src/utils/split_off.rs
jjbayer: Right. Should we introduce a dedicated enum then? Something like `SplitOff` vs. `Keep`? Not a blocker, just bikeshedding here.

Dav1dde: Introduced a `split_off` which takes a boolean and delegates to the `split_off_map` now. I did like the `Either` since it matches the return value `(left, right)`.

PR #3597, opened 2024-05-15
fix(spans): Restrict usage of otel endpoint

jjbayer:
Span ingestion was limited in general by the `organizations:standalone-span-ingestion` feature, but was GA'd for INP spans. However, we still want to restrict access to the experimental OTel `/spans/` endpoint, which is why this PR introduces a new feature `projects:relay-otel-endpoint`.

#skip-changelog

### DIFF
--- a/relay-server/src/envelope.rs
+++ b/relay-server/src/envelope.rs
@@ -40,7 +40,7 @@ use uuid::Uuid;
 
 use bytes::Bytes;
 use chrono::{DateTime, Utc};
-use relay_dynamic_config::ErrorBoundary;
+use relay_dynamic_config::{ErrorBoundary, Feature};
 use relay_event_normalization::{normalize_transaction_name, TransactionNameRule};
 use relay_event_schema::protocol::{Event, EventId, EventType};
 use relay_protocol::{Annotated, Value};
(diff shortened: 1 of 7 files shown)


relay-server/src/envelope.rs
```diff
@@ -1039,6 +1039,13 @@ pub struct EnvelopeHeaders<M = RequestMeta> {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     trace: Option<ErrorBoundary<DynamicSamplingContext>>,
 
+    /// A list of features required to process this envelope.
+    ///
+    /// This is an internal field that should only be set by Relay.
+    /// It is a serializable header such that it persists when spooling.
+    #[serde(default, skip_serializing_if = "SmallVec::is_empty")]
```
Dav1dde, line 1046: Should we serialize and deserialize this at all?

```suggestion
    #[serde(default, skip)]
```

Dav1dde (approved):

relay-server/src/envelope.rs
jjbayer: I wanted to make it `skip` initially, but this header is checked after disk spooling (see doc comment). This might actually be a problem for other `skip` (item) headers as well.

Dav1dde: Oops, didn't read the docs. 👍

PR #3599, opened 2024-05-15
feat(spans): Extract additional user fields for spans

Zylphrex:
Trying to bring the user fields on spans closer to transactions. User alone is not ergonomic and we'd like to be able to search by user email, username etc. Transactions also has the user's ip address but the spans table currently does not have the ipv4 and ipv6 columns for it so going to hold off on it as the other fields are more useful.

### DIFF
--- a/relay-event-normalization/src/normalize/span/tag_extraction.rs
+++ b/relay-event-normalization/src/normalize/span/tag_extraction.rs
@@ -1976,12 +1976,12 @@ LIMIT 1
         normalize_event(
             &mut event,
             &NormalizationConfig {
+                enrich_spans: true,
                 ..Default::default()
             },
         );
 
-        let mut event = event.into_value().unwrap();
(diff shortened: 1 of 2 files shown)


phacops (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -1921,4 +1938,67 @@ LIMIT 1
...
+            },
+        );
+
+        let event = event.into_value().unwrap();
+
+        let span = &event.spans.value().unwrap()[0];
+
+        let tags = span.value().unwrap().sentry_tags.value().unwrap();
+
+        assert_eq!(tags.get("user"), Some(&Annotated::new("id:1".to_string())));
```
```suggestion
        let tags = span.value().unwrap().sentry_tags;

        assert_eq!(get_value!(tags["user"]!).as_str(), "id:1");
```

jjbayer (approved):

PR #3600, opened 2024-05-15
feat(spans): disable redis span metrics extraction

DominikB2014:
We originally re-enabled db.redis in order to collect metrics for the cache module. However, since then we decided that the cache module will only look at `cache.*` span ops.

Therefore we can stop collecting `db.redis` span metrics.


Dav1dde (commented):
Should we also remove some of the redis spans from the `test_extract_span_metrics` test?

DominikB2014:
> test_extract_span_metrics

Yes!

DominikB2014:
> Should we also remove some of the redis spans from the `test_extract_span_metrics` test?

yes!

DominikB2014:
>Should we also remove some of the redis spans from the test_extract_span_metrics test?

@Dav1dde I'm indifferent to this, I could see us needing the db.redis spans in the future, so I'm okay with keeping these spans in the tests, if we re-add them back we'll have something to test against. Also having them there does verify that the metrics are not indeed being extracted.

Dav1dde:
Sounds good to me, I don't have an issue with keeping them around 👍

Dav1dde (approved):

PR #3603, opened 2024-05-15
metrics(processor): emit counter for feedback screenshots

aliu39:
Follow up from https://github.com/getsentry/relay/pull/3403

A useful metric for our product team to gauge the success of [user feedback screenshots](https://docs.sentry.io/product/user-feedback/#user-feedback-widget), after the new release of Javascript SDK v8. Unfortunately we have no way to track this in the Sentry backend, since the two event types split into separate pipelines after Relay. The only alternative is to add a field to the msg payload, which we'd prefer not to do.

**6/3 update:** Bruno spoke w/SDK team and they thought it made more sense to emit the metric here. We want to monitor the num of attachments sent in the same envelope as feedback (aka user report v2). 

#skip-changelog


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 **Internal**:
 
+- Add a counter metric for attachments processed in the same envelope as user feedback. ([#3603](https://github.com/getsentry/relay/pull/3603))
```
Dav1dde: Doesn't require a changelog entry.

relay-server/src/services/store.rs
```diff
@@ -346,6 +348,14 @@ impl StoreService {
             }
         }
 
+        if feedback_ingest_same_envelope_attachments && has_feedback && !attachments.is_empty() {
```
Dav1dde: Redundant check for `feedback_ingest_same_envelope_attachments`.

relay-server/src/services/store.rs
```diff
@@ -346,6 +348,14 @@ impl StoreService {
...
+            metric!(
+                counter(RelayCounters::FeedbackAttachments) += attachments.len() as i64,
+                organization_id = &scoping.organization_id.to_string(),
+                project_id = &scoping.project_id.to_string(),
```
Dav1dde: This is way too much cardinality.

Dav1dde (changes requested):
The metric should probably live in the processor not in the store, in a feedback specific code path which has access to the entire envelope.

The information you want to be tracking here is way to high cardinality.

Note: as an informative metric adding this to Relay is fine, but building processes etc. on top is a bad idea.

aliu39:
> but building processes etc. on top is a bad idea.

Wdym by this?

Dav1dde:
Informational yes, automatically including it in reports and building other processes ontop of the metric.

relay-server/src/statsd.rs
```diff
@@ -653,6 +653,14 @@ pub enum RelayCounters {
     /// Counts how many transactions were created from segment spans.
     #[cfg(feature = "processing")]
     TransactionsFromSpans,
+
+    /// The number of attachments processed in the same envelope as a user_report_v2 event.
+    ///
+    /// This metric is tagged with:
+    ///  - `organization_id`: Organization the feedback is sent to
+    ///  - `project_id`: Project the feedback is sent to
```
```suggestion
```

relay-server/src/services/store.rs
```diff
@@ -347,6 +349,10 @@ impl StoreService {
             }
         }
 
+        if has_feedback && !attachments.is_empty() {
+            metric!(counter(RelayCounters::FeedbackAttachments) += attachments.len() as i64);
```
jjbayer: What exactly do we want to count here, the number of feedback envelopes that have an attachment, or the total number of attachments? Can a single feedback envelope contain multiple screenshots?

jjbayer (commented):
> The only alternative is to add a field to the msg payload, which we'd prefer not to do.

1. Why not actually?

2. Do you want to measure this number before or after quotas are applied? If before, moving this metric to the envelope processor would make sense.

3. It probably makes sense to also count feedback items without attachments, i.e. "measure baseline". Could be a single metric tagged by `has_attachment: true/false`.

relay-server/src/services/store.rs
aliu39: Thanks so much for the considerations! I'm thinking it might be better to add a metadata field to the payload instead. Will close and come back to this if not

aliu39: > What exactly do we want to count here, the number of feedback envelopes that have an attachment, or the total number of attachments? Can a single feedback envelope contain multiple screenshots?

We'd just be counting the total number of attachments for now. Currently we only support 1 ss/attachment per feedback, don't expect this to change for a while

> It probably makes sense to also count feedback items without attachments, i.e. "measure baseline". Could be a single metric tagged by has_attachment: true/false.

Atm we're using the event.produced counter, filter by topic, for this

bruno-garcia:
> Informational yes, automatically including it in reports and building other processes ontop of the metric.

What processes are we building on top of it?

For context: We'd like to monitor usage of it

relay-server/src/statsd.rs
```diff
@@ -720,6 +723,8 @@ impl CounterMetric for RelayCounters {
             #[cfg(feature = "processing")]
             RelayCounters::TransactionsFromSpans => "transactions_from_spans",
             RelayCounters::MissingDynamicSamplingContext => "missing_dynamic_sampling_context",
+            #[cfg(feature = "processing")]
+            RelayCounters::FeedbackAttachments => "processing.feedback_attachments",
```
aliu39, line 725: open to other names

relay-server/src/services/processor/event.rs
```diff
@@ -420,6 +420,26 @@ pub fn has_unprintable_fields(event: &Annotated<Event>) -> bool {
     }
 }
 
+//TODO: comment
+//TODO: add test: all incr/no incr cases, envelope unchanged
+pub fn emit_feedback_metrics(envelope: &Envelope) -> i64 {
```
aliu39: open to other names

aliu39: let me know if this is the right place to put this fx

aliu3ntry:
Will add test coverage after we're aligned on whether to emit the metric here, placement of the helper fx, etc

relay-server/src/services/processor/event.rs
```diff
@@ -420,6 +420,26 @@ pub fn has_unprintable_fields(event: &Annotated<Event>) -> bool {
     }
 }
 
+//TODO: comment
+//TODO: add test: all incr/no incr cases, envelope unchanged
```
aliu39: Will add test coverage after we're aligned on whether to emit the metric here, placement of the helper fx, etc

relay-server/src/services/processor/event.rs
jjbayer: Why does this function return the count? To make it testable?

jjbayer (commented):
Looks OK to me.

relay-server/src/services/processor/event.rs
Dav1dde: I am fine with not having tests for metrics. A comment would be nice though

Dav1dde (approved):

relay-server/src/services/processor/event.rs
aliu39: Removed the return value since we're ok with no tests

aliu39:
Not sure why this lint keeps failing. `make lint && make style` is successful for me locally, and I added the cfg(feature) attribute to the emit_feedback_metrics fx call. @Dav1dde @jjbayer any ideas why?

Dav1dde:
@aliu3ntry you only made the call to the function conditional on the feature, but not the function itself (it still exists so all it references also must exist). 

I pushed a commit which removes all of the conditionals, don't think this is necessary here.

aliu39:
Ok - thanks!

PR #3604, opened 2024-05-15
ref(store): remove references to feedback topic option after rollout

aliu39:
Removing references to this option now that we have successfully rolled out and tested the feedback topic in all regions. 

https://github.com/getsentry/sentry/issues/66100
[Datadog](https://app.datadoghq.com/dashboard/z3p-ajn-5ma/user-feedback-infra?fromUser=false&refresh_mode=sliding&view=spans&from_ts=1715721120214&to_ts=1715807520214&live=true)

#skip-changelog


Dav1dde (approved):
👍

I assume this is okay for self hosted as well?

aliu39:
> I assume this is okay for self hosted as well?

Good catch, it's not... have to let the team know and might be some time before we add the topic/consumer there. So I'm gonna hold off on this, but might remove the inline-attachments flag

aliu39:
Had some complications with testing the topic+option in self-hosted - plan to come back to this later

PR #3605, opened 2024-05-15
ci(ci): Disable skipping publish for release, again

azaslavsky:
#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,6 +184,7 @@ jobs:
     env:
       FULL_CI: "${{
         github.ref == 'refs/heads/master'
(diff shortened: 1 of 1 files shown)


hubertdeng123 (approved):

Dav1dde (commented):
What's the intent behind this change, do you want to enable the publish job for the release?

It should be enough to add the condition to here instead https://github.com/getsentry/relay/blob/609d4f2bc40c5cfe65830d703f082dae24069d72/.github/workflows/ci.yml#L184-L188

azaslavsky:
@Dav1dde done. Can you please approve? This is currently blocking self-hosted release, which is now late from yesterday.

Dav1dde (approved):

PR #3610, opened 2024-05-16
ref: size-optimize the relay python library

asottile-sentry:
on linux this reduces the installed library size from 240MB to 14MB (might actually be smaller but I'm impatient -- that number was with `debug = true` still accidentally on)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,12 @@ debug = false
 # panic and error stack traces to Sentry.
 debug = true
 
+[profile.release-cabi]
+inherits = "release"
+lto = true
+debug = false
+strip = true
+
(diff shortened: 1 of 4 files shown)


asottile-sentry:
in my tests `debug` and `strip` made not terribly much difference -- a handful of MB at most.  lto made most of the improvement (almost certainly dead code elimination)

asottile-sentry:
the downside is this makes the build take about 2x -- but I suspect that can be won back by switching the macos builders to the macos arm runners

asottile-sentry:
hooray after switching over the macos builders it is no longer slower 🎉

jan-auer (approved):

PR #3611, opened 2024-05-16
ref: use macos arm builders to speed up library build

asottile-sentry:

### DIFF
--- a/.github/workflows/build_library.yml
+++ b/.github/workflows/build_library.yml
@@ -47,12 +47,12 @@ jobs:
       matrix:
         include:
           - target: x86_64-apple-darwin
-            py-platform: macosx-10_15_x86_64
+            py-platform: macosx-14_0_x86_64
           - target: aarch64-apple-darwin
-            py-platform: macosx-11_0_arm64
+            py-platform: macosx-14_0_arm64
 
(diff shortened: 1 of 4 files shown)


.github/workflows/build_library.yml
```diff
@@ -47,12 +47,12 @@ jobs:
       matrix:
         include:
           - target: x86_64-apple-darwin
-            py-platform: macosx-10_15_x86_64
+            py-platform: macosx-14_0_x86_64
```
jan-auer, line 50: Do we require >= macOS 14 now for all of sentry development? If not, I'd at least keep the intel build where it was. 

Technically, there shouldn't be any problems, I noticed that even built on newer versions binaries require a lower version number of the system libraries, but still...

jan-auer (commented):
We can also update the workflows in `ci.yml` and `build_binary.yml`

.github/workflows/build_library.yml
asottile-sentry: the value here ~generally needs to match the platform it is built from -- it's already wrong on master :)

but yes we already require macos>=14

jan-auer (approved):
:shipit:

PR #3613, opened 2024-05-16
feat(spans): Send microsecond precision timestamps

phacops:
As we're receiving standalone spans, we need to be able to order sibling spans and we currently store spans only with millisecond precision and that's not enough. This will add timestamps with up to a microsecond precision.

We'll still display and use millisecond in the product.

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -1347,10 +1347,6 @@ struct SpanMeasurement {
 
 #[derive(Debug, Deserialize, Serialize)]
 struct SpanKafkaMessage<'a> {
-    #[serde(skip_serializing)]
-    start_timestamp: f64,
-    #[serde(rename(deserialize = "timestamp"), skip_serializing)]
-    end_timestamp: f64,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     description: Option<&'a RawValue>,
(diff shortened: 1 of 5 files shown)


iambriccardo (approved):
The changes look good to me, did you change the Kafka schema?

relay-server/src/services/store.rs
```diff
@@ -1046,7 +1046,7 @@ impl StoreService {
                     KafkaMessage::MetricsSummary(MetricsSummaryKafkaMessage {
                         count,
                         duration_ms,
-                        end_timestamp,
+                        end_timestamp: end_timestamp_micro,
```
Dav1dde: Should this also be renamed in the `MetricsSummary`?

Dav1dde (commented):
Seems like the Kafka Schema is drifting apart from the envelope, that seems dangerous.

Any integration tests we can update? 👀

relay-server/src/services/store.rs
phacops: No, `MetricsSummary` already has this as a `f64`. I don't think we need to change anything there.

The field didn't exist on spans and that's why I added it. We were sending integers in milliseconds instead.

Dav1dde (approved):

PR #3615, opened 2024-05-16
fix(metrics): Apply global tags to transaction metrics

jjbayer:
Prerequisite to https://github.com/getsentry/sentry/pull/71004: Global tag mappings were not yet being applied to transaction metrics.

Beside the bug fix, this PR also bumps the metrics extraction version(s), and attempts to simplify the extraction function for transactions.


relay-dynamic-config/src/metrics.rs
```diff
@@ -115,7 +115,8 @@ pub struct CustomMeasurementConfig {
 ///      - Delay metrics extraction for indexed transactions.
 ///  - 4: Adds support for `RuleConfigs` with string comparisons.
 ///  - 5: No change, bumped together with [`MetricExtractionConfig::MAX_SUPPORTED_VERSION`].
-const TRANSACTION_EXTRACT_MAX_SUPPORTED_VERSION: u16 = 5;
+///  - 6: Bugfix to make transaction metrics extraction apply globally defined tag mappings.
+const TRANSACTION_EXTRACT_MAX_SUPPORTED_VERSION: u16 = 6;
```
jjbayer, line 119: We need to bump this version because old Relays will not apply global tags to legacy transaction metrics.

relay-dynamic-config/src/metrics.rs
```diff
@@ -324,7 +325,7 @@ impl MetricExtractionConfig {
     /// The latest version for this config struct.
     ///
     /// This is the maximum version supported by this Relay instance.
-    pub const MAX_SUPPORTED_VERSION: u16 = 3;
+    pub const MAX_SUPPORTED_VERSION: u16 = 4;
```
jjbayer, line 328: Unfortunately both versions need to be bumped in tandem: If I only bump `TRANSACTION_EXTRACT_MAX_SUPPORTED_VERSION`, old Relays will still set the `metrics_extracted` flag to true when they extract on-demand metrics, which means the crucial legacy transaction metrics (`duration` etc.) will not be extracted.

relay-server/src/services/processor.rs
```diff
@@ -1207,96 +1207,90 @@ impl EnvelopeProcessorService {
...
     }
 
-    /// Extract metrics from all envelope items.
-    ///
-    /// Caveats:
-    ///  - This functionality is incomplete. At this point, extraction is implemented only for
-    ///    transaction events.
-    fn extract_metrics(
+    /// Extract transaction metrics.
+    fn extract_transaction_metrics(
```
jjbayer, line 1210: Apologies for the large diff. The rewrite of this function is trying to achieve:

1. Fix the bug referenced in the PR title.
2. Make tandem version bumps unnecessary in the future.
3. Simplify: This function was originally written as a general purpose function, but since it's only called from `process_transactions` we can introduce some early returns.

relay-server/src/services/processor.rs
```diff
@@ -1207,96 +1207,90 @@ impl EnvelopeProcessorService {
...
-                    // If there's an error with global metrics extraction, it is safe to assume that this
-                    // Relay instance is not up-to-date, and we should skip extraction.
-                    relay_log::debug!("Failed to parse global extraction config {e}");
-                    return Ok(());
-                })
-            }
+        let generic_config = {
+            let config = match &state.project_state.config.metric_extraction {
+                ErrorBoundary::Ok(ref config) if config.is_supported() => config,
+                _ => return Ok(()),
```
jjbayer, line 1229: Generic project config is no longer optional. `event_metrics_extraced` will only be set if both configs are present, allowing us to bump the two extraction versions individually in the future.

relay-server/src/services/processor.rs
```diff
@@ -1207,96 +1207,90 @@ impl EnvelopeProcessorService {
...
-            }
+        let transaction_from_dsc = state
+            .managed_envelope
+            .envelope()
+            .dsc()
+            .and_then(|dsc| dsc.transaction.as_deref());
+
+        let extractor = TransactionExtractor {
+            config: tx_config,
+            generic_config: Some(&generic_config),
```
jjbayer: This is the actual bug fix: Tag definitions are applied from both global and project config, not just project config.

tests/integration/test_metrics.py
```diff
@@ -1873,3 +1875,70 @@ def test_metrics_with_denied_names(
...
+                buckets = item.payload.json
+                for bucket in buckets:
+                    if outlier := bucket.get("tags", {}).get("histogram_outlier"):
+                        tags[bucket["name"]] = outlier
+
+    assert tags == {
+        "d:transactions/measurements.fcp@millisecond": "outlier",
+        "d:transactions/duration@millisecond": "inlier",
+        "d:transactions/measurements.lcp@millisecond": "inlier",
+    }
```
jjbayer, line 1944: Without the bug fix, this dict would be empty.

tests/integration/fixtures/histogram-outliers.yml
jjbayer, line 1: I copied this from a snapshot in sentry, so the integration test also verifies that the config that sentry provides is applied correctly.

relay-server/Cargo.toml
```diff
@@ -141,5 +141,6 @@ insta = { workspace = true }
 relay-event-schema = { workspace = true, features = ["jsonschema"] }
 relay-protocol = { workspace = true, features = ["test"] }
 relay-test = { workspace = true }
+serde_yaml = { workspace = true }
```
Dav1dde: Doesn't seem to be used?

relay-server/src/services/processor.rs
```diff
@@ -1207,96 +1207,90 @@ impl EnvelopeProcessorService {
...
-                    relay_log::error!("Failed to parse global extraction config {e}");
-                    MetricExtractionGroups::EMPTY
-                } else {
-                    // If there's an error with global metrics extraction, it is safe to assume that this
-                    // Relay instance is not up-to-date, and we should skip extraction.
-                    relay_log::debug!("Failed to parse global extraction config {e}");
-                    return Ok(());
-                })
-            }
+        let generic_config = {
```
```suggestion
        let combined_config = {
```

relay-server/src/services/processor.rs
```diff
@@ -1207,96 +1207,90 @@ impl EnvelopeProcessorService {
...
+
+        if !tx_config.is_enabled() || state.project_state.has_feature(Feature::DiscardTransaction) {
+            return Ok(());
+        }
 
-            if let Some(config) = config {
-                let combined_config = CombinedMetricExtractionConfig::new(global_config, config);
+        let Some(event) = state.event.value() else {
+            return Ok(());
+        };
```
Dav1dde: Can we move this to the top after `state.event_metrics_extracted`?

Dav1dde (approved):

Dav1dde:
@jjbayer watch out we had a self hosted release and the changelog is messed up

PR #3617, opened 2024-05-16
ref(processing/store): remove references to feedback attachments option

aliu39:
Follow up after https://github.com/getsentry/team-replay/issues/393. We've set the option to true for all regions and tested that it works, so this is a dead code path we can remove

Relates to https://github.com/getsentry/sentry-javascript/pull/11428

#skip-changelog


jjbayer (commented):
Same question as in https://github.com/getsentry/relay/pull/3604: Is this safe for self-hosted?

aliu39:
> Same question as in #3604: Is this safe for self-hosted?

Unlike 3604, this is the same for self-hosted! The change is backwards-compatible with older SDK versions, as we'll still process feedback and attachments in separate envelopes, even with the flag enabled

Dav1dde (approved):
Should also remove `feedback.ingest-inline-attachments` from Sentry now.

PR #3618, opened 2024-05-17
ref(metric-meta): Allow null values in contexts

Dav1dde:
Some SDKs are sending null values in the contexts, see included test.

#skip-changelog

### DIFF
--- a/relay-metrics/src/meta/protocol.rs
+++ b/relay-metrics/src/meta/protocol.rs
@@ -50,13 +50,13 @@ pub struct Location {
     pub lineno: Option<u64>,
     /// Source code leading up to `lineno`.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub pre_context: Vec<String>,
+    pub pre_context: Vec<Option<String>>,
     /// Source code of the current line (`lineno`).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub context_line: Option<String>,
     /// Source code of the lines after `lineno`.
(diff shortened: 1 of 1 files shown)


relay-metrics/src/meta/protocol.rs
```diff
@@ -50,13 +50,13 @@ pub struct Location {
...
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub pre_context: Vec<String>,
+    pub pre_context: Vec<Option<String>>,
     /// Source code of the current line (`lineno`).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub context_line: Option<String>,
     /// Source code of the lines after `lineno`.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub post_context: Vec<String>,
+    pub post_context: Vec<Option<String>>,
```
jjbayer, line 59: Should we leave this as `Vec<String>` and filter out `None` values on deserialization? Or do the `None` values actually have significance / help the user?

jjbayer (approved):

relay-metrics/src/meta/protocol.rs
Dav1dde: Filtering a null from the middle `["foo", null, "bar"]` now means an empty line is skipped, I was thinking about deserializing null as `""` but the option seemed like the better approach since it matches what we do for `post_context` in errors (it uses `Annotated` there).

PR #3619, opened 2024-05-17
ref: dual-publish docker image to google's artifact registry

asottile-sentry:
gcr is being phased out within the year

#skip-changelog


Dav1dde (approved):
urgh regional registries

olksdr (approved):

PR #3621, opened 2024-05-17
ref: use artifact registry for relay deploys

asottile-sentry:
#skip-changelog


asottile-sentry:
don't merge until a deploy has happened with dual-writing enabled (such that rolling back is easier)

Dav1dde (approved):

PR #3622, opened 2024-05-17
fix(cogs): Only attribute rate limit buckets to spans and transaction metrics

Dav1dde:
Only spans and transactions are actually rate limited in that stage.

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -2671,7 +2671,13 @@ impl EnvelopeProcessorService {
             EnvelopeProcessor::SubmitClientReports(_) => AppFeature::ClientReports.into(),
             #[cfg(feature = "processing")]
             EnvelopeProcessor::RateLimitBuckets(v) => {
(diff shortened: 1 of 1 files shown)


relay-server/src/services/processor.rs
```diff
@@ -2671,7 +2671,13 @@ impl EnvelopeProcessorService {
             EnvelopeProcessor::SubmitClientReports(_) => AppFeature::ClientReports.into(),
             #[cfg(feature = "processing")]
             EnvelopeProcessor::RateLimitBuckets(v) => {
-                relay_metrics::cogs::ByCount(v.bucket_limiter.buckets()).into()
+                relay_metrics::cogs::ByCount(v.bucket_limiter.buckets().filter(|b| {
+                    matches!(
```
jjbayer, line 2677:
```suggestion
                    // Only spans and transactions are actually rate limited at this point.
                    // Other metrics do not cause costs.
                    matches!(
```

jjbayer (approved):

PR #3623, opened 2024-05-17
ref(ds): Low cardinality outcome reason

jjbayer:
Reduce noise in outcome reasons by mapping ranges of dynamic sampling rule IDs together, and ordering them. That is

```python
"123,1004,1500,1403,1404,1000"
# becomes
"1000,1004,1400,1500,?"
```


relay-server/src/services/outcome.rs
```diff
@@ -251,6 +252,79 @@ impl fmt::Display for Outcome {
...
+impl From<RuleId> for RuleCategory {
+    fn from(value: RuleId) -> Self {
+        match value.0 {
+            1000 => Self::BoostLowVolumeProjects,
+            1001 => Self::BoostEnvironments,
+            1002 => Self::IgnoreHealthChecks,
+            1003 => Self::BoostKeyTransactions,
+            1004 => Self::Recalibration,
+            1005 => Self::BoostReplayId,
+            n if (1400..1500).contains(&n) => Self::BoostLowVolumeTransactions,
```
matejminar: Is this 1400 including and 1500 excluding? If yes, then we're good 👍
Same question for BoostLatestReleases and Custom.

matejminar (approved):

relay-server/src/services/outcome.rs
jjbayer: Yes! I had a more explicit `n >= 1400 && n < 1500` condition here first, but the rust linter is very opinionated.

> The range `start..end` contains all values with `start <= x < end`

([docs](https://doc.rust-lang.org/std/ops/struct.Range.html))

Dav1dde:
```suggestion
            1400..=1499 => Self::BoostLowVolumeTransactions,
```

And for the following lines the same.

relay-server/src/services/outcome.rs
```diff
@@ -251,6 +252,79 @@ impl fmt::Display for Outcome {
...
+    Recalibration,
+    BoostReplayId,
+    BoostLowVolumeTransactions,
+    BoostLatestReleases,
+    Custom,
+    Other,
+}
+
+impl RuleCategory {
+    fn to_str(self) -> &'static str {
```
Dav1dde:
```suggestion
    fn as_str(self) -> &'static str {
```

We have `as_str` throughout Relay for enums

Dav1dde (approved):

PR #3626, opened 2024-05-21
fix(metric-stats): Use cardinality timestamp for reports

Dav1dde:
Aligns the timestamps of the cardinality reports with the actual cardinality limiting.

Without aligning the timestamps increased cardinality could spill over to new "metric minutes" which in general isn't a big problem but using the same timestamp is correct.

#skip-changelog


jjbayer (approved):
Makes sense, could you add a one-liner to the PR description about what can happen if these timestamps are not aligned?

PR #3628, opened 2024-05-21
feat(metrics): Limit metric name to 150 chars

### LINKED ISSUE #3614: Enforce limit of 200 chars per MRI
The indexer can support strings up to 200 chars long. In case a metric with an MRI over 200 chars is received by the consumer, it will be dropped. We want to avoid this behavior, thus we want to perform validation in Relay.

Either we drop the metric or we cut off the MRI, maybe in a smart way, by trying to cut off first the unit and then the name. If with both of them cut the length is still > 200, we will drop it.

iambriccardo:
This PR limits the metric name to 150 characters, which is added to the limit of 15 characters in the metric unit. Since we have a total of 200 characters supported by the indexer, we now have 35 chars for separators, type and namespace of the MRI.

Closes: https://github.com/getsentry/relay/issues/3614

### DIFF
--- a/relay-base-schema/src/metrics/mod.rs
+++ b/relay-base-schema/src/metrics/mod.rs
@@ -39,8 +39,13 @@ pub fn try_normalize_metric_name(name: &str) -> Option<Cow<'_, str>> {
         return Some(normalized_name);
     }
 
-    // We limit the string to a fixed size. Here we are taking slices assuming that we have a single
-    // character per index since we are normalizing the name above.
+    // We limit the string to a fixed size.
+    //
+    // Here we are taking slices, assuming that we have a single character per index since we are
+    // normalizing the name above.
+    //
+    // If we allow characters that take more than 1 byte per character, we will need to change this
+    // function. Otherwise, it will panic because the index might cut a character in half.
     Some(match normalized_name {
         Cow::Borrowed(value) => Cow::Borrowed(&value[..CUSTOM_METRIC_NAME_MAX_SIZE]),
         Cow::Owned(mut value) => {
--- a/relay-base-schema/src/metrics/mod.rs
+++ b/relay-base-schema/src/metrics/mod.rs
@@ -31,11 +31,11 @@ pub fn try_normalize_metric_name(name: &str) -> Option<Cow<'_, str>> {
 
     // Note: `-` intentionally missing from this list.
     let normalize_re = NORMALIZE_RE.get_or_init(|| Regex::new("[^a-zA-Z0-9_.]+").unwrap());
-    let normalized = normalize_re.replace_all(name, "_");
+    let normalized_name = normalize_re.replace_all(name, "_");
 
     // We limit the string to a fixed size. Here we are taking slices assuming that we have a single
     // character per index since we are normalizing the name above.
-    Some(match normalized {
+    Some(match normalized_name {
         Cow::Borrowed(value) => {
             Cow::Borrowed(&value[..min(value.len(), CUSTOM_METRIC_NAME_MAX_SIZE)])
         }
--- a/relay-base-schema/src/metrics/mod.rs
+++ b/relay-base-schema/src/metrics/mod.rs
@@ -9,6 +9,7 @@ pub use self::name::*;
 pub use self::units::*;
 
 use regex::Regex;
+use std::cmp::min;
 use std::{borrow::Cow, sync::OnceLock};
 
 const CUSTOM_METRIC_NAME_MAX_SIZE: usize = 150;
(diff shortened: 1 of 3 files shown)


relay-base-schema/src/metrics/mod.rs
```diff
@@ -26,9 +28,18 @@ pub fn try_normalize_metric_name(name: &str) -> Option<Cow<'_, str>> {
         return None;
     }
 
+    // Take a slice of the first n characters.
+    let end_index = name
```
iambriccardo, line 32: This index is non-inclusive, so it refers to the `n + 1`th element.

relay-base-schema/src/metrics/mod.rs
```diff
@@ -9,8 +9,11 @@ pub use self::name::*;
 pub use self::units::*;
 
 use regex::Regex;
+use std::cmp::min;
 use std::{borrow::Cow, sync::OnceLock};
 
+const CUSTOM_METRIC_NAME_MAX_SIZE: usize = 150;
```
Dav1dde, line 17: A short comment explaining the limit would be good, also can mention the technical limit in the indexer and how much space that leaves for units and namespace names.

Also we need to not forget to update the sentry dev docs.

relay-base-schema/src/metrics/mod.rs
```diff
@@ -28,7 +31,18 @@ pub fn try_normalize_metric_name(name: &str) -> Option<Cow<'_, str>> {
...
+    let normalized_name = normalize_re.replace_all(name, "_");
+
+    // We limit the string to a fixed size. Here we are taking slices assuming that we have a single
+    // character per index since we are normalizing the name above.
+    Some(match normalized_name {
+        Cow::Borrowed(value) => {
+            Cow::Borrowed(&value[..min(value.len(), CUSTOM_METRIC_NAME_MAX_SIZE)])
+        }
+        Cow::Owned(value) => {
+            Cow::Owned(value[..min(value.len(), CUSTOM_METRIC_NAME_MAX_SIZE)].to_string())
```
Dav1dde: We can avoid the allocation here using truncate

relay-base-schema/src/metrics/mod.rs
```diff
@@ -28,7 +31,18 @@ pub fn try_normalize_metric_name(name: &str) -> Option<Cow<'_, str>> {
...
     // Note: `-` intentionally missing from this list.
     let normalize_re = NORMALIZE_RE.get_or_init(|| Regex::new("[^a-zA-Z0-9_.]+").unwrap());
-    Some(normalize_re.replace_all(name, "_"))
+    let normalized_name = normalize_re.replace_all(name, "_");
+
+    // We limit the string to a fixed size. Here we are taking slices assuming that we have a single
+    // character per index since we are normalizing the name above.
+    Some(match normalized_name {
+        Cow::Borrowed(value) => {
+            Cow::Borrowed(&value[..min(value.len(), CUSTOM_METRIC_NAME_MAX_SIZE)])
```
Dav1dde: Instead of doing the min check twice we can also just do it outside of the match: `if foo.len() <= LIMIT { return foo; } ...`

relay-base-schema/src/metrics/mri.rs
```diff
@@ -464,6 +464,23 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_normalize_name_length() {
+        let long_mri = "c:custom/ThisIsACharacterLongStringForTestingPurposesToEnsureThatWeHaveEnoughCharactersToWorkWithAndToCheckIfOurFunctionProperlyHandlesSlicingAndNormalizationWithoutErrors";
```
Dav1dde, line 469: makes sense to also have a test on how it behaves when we have replacement characters in the name which get replaced with a single underscore

relay-base-schema/src/metrics/mod.rs
```diff
@@ -28,7 +33,21 @@ pub fn try_normalize_metric_name(name: &str) -> Option<Cow<'_, str>> {
...
+    if normalized_name.len() <= CUSTOM_METRIC_NAME_MAX_SIZE {
+        return Some(normalized_name);
+    }
+
+    // We limit the string to a fixed size. Here we are taking slices assuming that we have a single
+    // character per index since we are normalizing the name above.
+    Some(match normalized_name {
+        Cow::Borrowed(value) => Cow::Borrowed(&value[..CUSTOM_METRIC_NAME_MAX_SIZE]),
+        Cow::Owned(mut value) => {
+            value.truncate(CUSTOM_METRIC_NAME_MAX_SIZE);
```
jjbayer, line 52: Note that both `&value[..CUSTOM_METRIC_NAME_MAX_SIZE]` and `String::truncate` would panic if it wasn't for `normalize_re` and the .`len()` check. I don't have a better idea right now, but it's something to keep in mind if we ever refactor this function.

Maybe we can turn the early return of the length check into a `if normalized.len() > CUSTOM_METRIC_NAME_MAX_SIZE { ... }` block to clarify that these belong together, and add a `debug_assert!(value.is_ascii());`

jjbayer (approved):

relay-base-schema/src/metrics/mri.rs
```diff
@@ -474,6 +474,14 @@ mod tests {
             "ThisIsACharacterLongStringForTestingPurposesToEnsureThatWeHaveEnoughCharactersToWorkWithAndToCheckIfOurFunctionProperlyHandlesSlicingAndNormalizationW"
         );
 
+        let long_mri_with_replacement = "c:custom/ThisIsÄCharacterLongStringForŤestingPurposesToEnsureThatWeHaveEnoughCharactersToWorkWithAndToCheckIfOurFunctionProperlyHandlesSlicingAndNormalizationWithoutErrors";
```
Dav1dde: The test is supposed to show what happens if **multiple** characters get replaced with a **single** underscore. Also possibly makes sense to have a separate test case for exactly this behaviour.

relay-base-schema/src/metrics/mod.rs
iambriccardo: Yeah, this was considered, in my opinion our best bet is to put enough good tests so that a malicious refactor will be noticed.

PR #3629, opened 2024-05-21
ref(metric-stats): Make metric stats collection configurable

Dav1dde:
To make it possible to collect metric stats from our pop layer without enabling it for managed customer Relays. This follows a similar configuration pattern we have for outcomes.

#skip-changelog

### DIFF
--- a/relay-server/src/metric_stats.rs
+++ b/relay-server/src/metric_stats.rs
@@ -107,7 +107,8 @@ impl MetricStats {
     }
 
     fn is_enabled(&self, scoping: Scoping) -> bool {
-        self.config.processing_enabled() && self.is_rolled_out(scoping.organization_id)
+        (self.config.processing_enabled() || self.config.metric_stats_enabled())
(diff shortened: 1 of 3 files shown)


relay-server/src/metric_stats.rs
```diff
@@ -107,7 +107,8 @@ impl MetricStats {
     }
 
     fn is_enabled(&self, scoping: Scoping) -> bool {
-        self.config.processing_enabled() && self.is_rolled_out(scoping.organization_id)
+        (self.config.processing_enabled() || self.config.metric_stats_enabled())
```
jjbayer: nit: Should `metric_stats_enabled()` check `self.config.processing_enabled()` (that's where the behavior is documented)?

jjbayer (approved):

relay-server/src/metric_stats.rs
Dav1dde: Good idea, will update the code and adjust docs!

PR #3630, opened 2024-05-21
chore(llm-monitoring): Extract AI data from data instead of measurements

colin-sentry:
We need to start porting things over to data, since `measurements` is being deprecated and will be removed from the APIs.

For backcompat, this goes data -> measurement -> metric

### DIFF
--- a/relay-event-normalization/src/normalize/span/tag_extraction.rs
+++ b/relay-event-normalization/src/normalize/span/tag_extraction.rs
@@ -643,20 +643,46 @@ pub fn extract_tags(
     span_tags
 }
 
+fn value_to_f64(val: Option<&Value>) -> Option<f64> {
+    match val {
+        Some(Value::F64(f)) => Some(*f),
+        Some(Value::I64(i)) => Some(*i as f64),
+        Some(Value::U64(u)) => Some(*u as f64),
+        _ => None,
+    }
+}
+
 /// Copies specific numeric values from span data to span measurements.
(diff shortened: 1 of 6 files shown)


phacops:
Long term, we only want `data` to stay. So, I would change the metric to read from `data` fields and extract measurements to those `data` fields if need be, otherwise the SDK will start sending them directly and no extraction needed.

phacops (commented):
Long term, we only want `data` to stay. So, I would change the metric to read from `data` fields and extract measurements to those `data` fields if need be, otherwise the SDK will start sending them directly and no extraction needed.

iambriccardo (approved):
LGTM

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -643,20 +643,46 @@ pub fn extract_tags(
...
 }
 
+fn value_to_f64(val: Option<&Value>) -> Option<f64> {
+    match val {
+        Some(Value::F64(f)) => Some(*f),
+        Some(Value::I64(i)) => Some(*i as f64),
+        Some(Value::U64(u)) => Some(*u as f64),
+        _ => None,
+    }
+}
```
jjbayer, line 653: Nice!

jjbayer (approved):
@colin-sentry could you please expand the PR description to include the motivation for this change? Can be a one-liner.

PR #3631, opened 2024-05-21
feat(spans): extract cache key in indexed spans

KevinL10:
Adds a cache key field to the indexed spans.

We currently derive the key from the second word of the `span.description` as a temporary hack. Once the SDKs start streaming the cache key through `span.data`, we'll update how we derive the key.


relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -465,6 +467,15 @@ pub fn extract_tags(
                 let tag_value = if *cache_hit { "true" } else { "false" };
                 span_tags.insert(SpanTagKey::CacheHit, tag_value.to_owned());
             }
+            // Read the cache key from the 2nd word of the span description
+            // TODO: derive cache key from span.data instead when SDKs are ready
+            if let Some(cache_key) = span
+                .description
+                .value()
+                .and_then(|description| description.split(' ').nth(1))
```
DominikB2014: I think we should update this to grab it from span.data right away. Note that cache.key is a array 
https://develop.sentry.dev/sdk/performance/modules/caches/

So we'll have to convert that to a string, correct me if I'm wrong @jjbayer ?


1. The SDKs should have this support this soon as part of the cache module v0 launch, we'll inform customers that they must update to a minimum version.
2. Cache.key is what we'll be documenting for manually instrumenting.
3. This would break if there are multiple keys (think MGET), if the key has a space in it, or if someone puts a custom span description (should be common if manually instrumented). I'd rather not store a bunch of data that could have issues.

Just to add, the reason why it was okay on the frontend as a hack, is because there was no storage implications, and it allowed us to POC something quickly. Recently we decided to just display the whole span description anyways.

KevinL10: Once we serialize the cache keys as a string, at what point (if any) do we need to deserialize the cache.key so that we store it as an Array type in Clickhouse?

DominikB2014: So that's the unfortunate thing, will need to first confirm with @jjbayer, but I believe currently we only have the ability to store this data as a string in clickhouse through relay. So the only place to deserialize would be from the frontend. This also means, if we want to search by key, we'll probably have to do an awkward string search from the frontend side when querying for the searlized array.

@narsaynorath i believe u ran into something similar with mobile mobiles, maybe u have something to add here too?

narsaynorath: No, unfortunately we didn't have any similar needs for array fields in clickhouse.

From what I understand of the description, this is meant to tie us over until the SDKs are able to send some kind of `cache.key` value independently? @DominikB2014  Do we know the timeline on that and if it's soon, can we punt filtering on this field until that gets captured properly? Otherwise, yeah, it could be a headache if we generate bad data for customers.

Do they have a method sketched out for properly handling multiple keys? 🤔 maybe we could port similar functionality here, or if we really need to get something in, we can do a few defensive checks here and only derive cache keys for values that we know are well-formed. I think justification for my last suggestion requires understanding how often multiple keys actually occurs

DominikB2014: > @DominikB2014  Do we know the timeline on that and if it's soon, can we punt filtering on this field until that gets captured properly? L

SDKs have cache.key working already, I expect it to be done this week

KevinL10: updated to read `cache.key` directly from `span.data`. This should not affect anything if `cache.key` doesn't exist, and will store the serialized `cache.key` when it exists.

KevinL10: we could leave the string search logic on the serialized arrays to the frontend for now, and add the ability to store Array types later on

DominikB2014: Awesome, I agree, we can think about storing at as an array later on, as I imagine that would require more work!

DominikB2014 (commented):
Awesome stuff, I agree, let's worry about storing it as an array later on, as I imagine that will be more work.

Changes look good to me, but let's get some approvals from @getsentry/ingest before proceeding!

iambriccardo (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -465,6 +467,12 @@ pub fn extract_tags(
                 let tag_value = if *cache_hit { "true" } else { "false" };
                 span_tags.insert(SpanTagKey::CacheHit, tag_value.to_owned());
             }
+            if let Some(cache_keys) = span.data.value().and_then(|data| data.cache_key.value()) {
+                span_tags.insert(
+                    SpanTagKey::CacheKey,
+                    serde_json::to_string(cache_keys).unwrap(),
```
jjbayer: Let's not unwrap here, even if the risk of panicking is low (e.g. `if let Ok(cache_keys) = cache_keys`).

jjbayer (changes requested):

jjbayer (approved):

PR #3635, opened 2024-05-22
build: Remove unused dependencies

anonrig:
I've given @boshen 's cargo-shear a try: https://github.com/Boshen/cargo-shear

#skip-changelog

### DIFF
--- a/relay-dynamic-config/Cargo.toml
+++ b/relay-dynamic-config/Cargo.toml
@@ -30,9 +30,6 @@ relay-quotas = { workspace = true }
 relay-sampling = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
-smallvec = { workspace = true }
 
 [dev-dependencies]
-insta = { workspace = true }
 similar-asserts = { workspace = true }
-indexmap = { workspace = true }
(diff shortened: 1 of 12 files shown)


relay-dynamic-config/Cargo.toml
```diff
@@ -30,9 +30,6 @@ relay-quotas = { workspace = true }
 relay-sampling = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
-smallvec = { workspace = true }
 
 [dev-dependencies]
-insta = { workspace = true }
```
colin-sentry, line 36: this is probably required

colin-sentry: for `cargo insta review`

anonrig: I couldn't find any usages inside `relay-dynamic-config`. Is this line still required to run `cargo insta review` even if the workspace has `insta` as a dependency?

Dav1dde (approved):

relay-dynamic-config/Cargo.toml
jan-auer: `cargo insta` commands require the crate installed on the host, so it doesn't have to be in dependencies.

PR #3636, opened 2024-05-22
feat(remote-config): Add configuration endpoint

cmanallen:
Temporary addition to proxy Sentry.

- [ ] How do I get the relay-id?
- [ ] How do I check the feature?
- [ ] Is any of this right? Do I need to perform conditional checks for self-hosted relays? Do only processing relays reach out to Sentry?

Ref: https://github.com/getsentry/sentry/issues/70942


jjbayer (commented):
As discussed on Slack, you might be able to leverage the existing `forward` endpoint for this, and not need any Relay changes at all initially: https://github.com/getsentry/relay/blob/14febaaf33f349c7488bdea3f12526b395f70ae1/relay-server/src/endpoints/forward.rs#L256-L276

cmanallen:
Forwarding works as expected!

PR #3639, opened 2024-05-23
feat(pii): Scrub arrays of 2 elements as key-value pairs

### LINKED ISSUE #2567: Scrub auth tokens anywhere
Currently we do not scrub bearer tokens everywhere, e.g.

```json
{
    "vars": {
        "headers": [
            ["authorization", "Bearer abc123"]
        ]
    }
}
```

We do not treat `["authorization", "Bearer abc123"]` as a key-value pair, so we scrub the _key_, but not the value (see [code](https://github.com/getsentry/relay/blob/c4c6fc6104be17b7f753134fea92c0354dc9164d/relay-pii/src/processor.rs#L326-L328)).

Possible solutions:

1. Redact `Bearer <hex>` anywhere in a string (as we do for credit card numbers).
2. Build a heuristic that treats list of pairs as key-value mappings.

iambriccardo:
This PR adds support for treating an array of arrays of length two as an object. This was done since headers and other pieces of data are commonly encoded in `[["key", "value"], ["key", "value"]]` and our previous logic was scrubbing each string value individually.

The algorithm works by figuring out if given an array, it is composed of multiple arrays of length two where each first component is a string and then fakes entering the first element of the array as the key of an object.

Closes: https://github.com/getsentry/relay/issues/2567


relay-pii/src/processor.rs
```diff
@@ -1598,4 +1630,113 @@ mod tests {
...
+                ]
+              }
+            })
+            .into(),
+        );
+
+        let mut processor = PiiProcessor::new(config.compiled());
+        process_value(&mut event, &mut processor, ProcessingState::root()).unwrap();
+
+        let vars = event
```
iambriccardo: Is there a better way?

relay-pii/src/processor.rs
```diff
@@ -101,6 +101,47 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+                            key_name.as_str(),
+                            state.inner_attrs(),
+                            inner_value.value_type(),
+                        );
+
+                        let previous_meta = value[1].meta().clone();
+                        process_value(&mut value[1], self, &entered)?;
+                        // We check whether meta has changed from empty to non-empty in order to
+                        // understand if some rules matched downstream.
+                        if previous_meta.is_empty() && !value[1].meta().is_empty() {
```
iambriccardo: A better impl here would check for structural equivalence but we might gain some performance if we assume that metadata won't be there on an incoming annotated value. Maybe this assumption is totally wrong.

relay-pii/src/processor.rs
```diff
@@ -1598,4 +1639,223 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn test_tuple_array_scrubbed_with_string_selector_and_password_matcher() {
+        let config = serde_json::from_str::<PiiConfig>(
+            r##"
+                {
+                    "applications": {
+                        "$string": ["@password:remove"]
```
iambriccardo, line 1842: This rule `@password:remove` works by matching a key and then scrubbing the value, this is why it was put here.

jjbayer (commented):
We should try to get rid of the `.clone().into_value()` calls as they are potentially expensive. We should aim for a structure like this:

```rust
        match try_as_pairlist(value) {
            Ok(map) => self.process_map(value, meta, state),
            Err(_) => value.process_child_values(self, state)
        }
```

where `try_as_pairlist` would invoke another visitor i.e. processor that attempts to assemble a borrowing map view, and `process_map` is a generalization of the existing `process_object` that can handle any kind of map-like structure.

relay-pii/src/processor.rs
```diff
@@ -111,33 +111,39 @@ impl<'a> Processor for PiiProcessor<'a> {
...
-                        // `process_child_values`.
-                        let entered =
-                            state.enter_borrowed(key_name, state.inner_attrs(), value_type);
-
-                        processor::process_value(&mut value, self, &entered)?;
+                        // We enter the index of the array.
+                        let index_state = state.enter_index(index, state.inner_attrs(), value_type);
+                        // We enter the key of the first element of the array, since we treat it
+                        // as a pair.
+                        let key_state = index_state.enter_borrowed(
```
jjbayer: Why do we need to enter both by index _and_ by key? That seems semantically incorrect, because we want the same semantics for `[["a", 1], ["b": 2]]` as for `{"a": 1, "b": 2}`. With the current code, we treat the list like `[{"a": 1}, {"b": 2}]`.

See also https://github.com/getsentry/relay/blob/d812b633292b47950906de2df205c5a9324e4262/relay-pii/src/utils.rs#L28-L29

iambriccardo: Oh yeah, that's actually a good point. I mistakenly implemented it before. Good catch

relay-pii/src/processor.rs
```diff
@@ -101,6 +101,53 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+    fn process_array<T>(
+        &mut self,
+        array: &mut Array<T>,
+        _meta: &mut Meta,
+        state: &ProcessingState<'_>,
+    ) -> ProcessingResult
+    where
+        T: ProcessValue,
+    {
+        if is_pairlist(array) {
```
Dav1dde, line 113: I prefer a design where the check already returns you a prepared datastructure to visit:

```rs
if let Some(pairlist) = Pairlist::try_from(array) {
    pairlist.visit(self)
} else {
   visit_children
}
```

relay-pii/src/processor.rs
```diff
@@ -489,6 +597,30 @@ mod tests {
         rv
     }
 
+    fn extract_vars(event: Option<&Event>) -> &FrameVars {
```
Dav1dde: Thought I've seen a helper macro for this already somewhere 🤔

relay-pii/src/processor.rs
```diff
@@ -1598,4 +1730,209 @@ mod tests {
...
+            (
+                r#"[["foo", "bar"], ["foo", {"nested": {"something": 1}}]]"#,
+                true,
+            ),
+        ] {
+            let v = Annotated::<Value>::from_json(case).unwrap();
+            let Annotated(Some(Value::Array(mut a)), _) = v else {
+                panic!()
+            };
+            assert_eq!(is_pairlist(&mut a), expected, "{}", case);
```
Dav1dde:
```suggestion
            assert_eq!(is_pairlist(&mut a), expected, "{case}");
```

relay-pii/src/processor.rs
iambriccardo: We would just pipe through the array since this check is not doing anything with the data structure itself.

jjbayer: Same here, this was the original design, but I didn't manage to create a lightweight (i.e. borrowing) pairlist from the visitor because the visitor (`process_string`) gets a string reference without a lifetime parameter. I'm still open to going that route though, either by changing the `Processor` framework to allow such a thing, or by simply accepting the cost and creating an owned `PairList` tentatively.

relay-pii/src/processor.rs
jjbayer: `get_value!`

iambriccardo: Amazing

relay-pii/src/snapshots/relay_pii__convert__tests__bearer_tokens_scrubbed.snap
```diff
@@ -13,7 +13,7 @@ expression: data
                 "request": {
                   "headers": [
                     [
-                      "[Filtered]",
+                      "AuthToken",
```
iambriccardo, line 16: The first element is now kept since we treat is as a key.

relay-pii/src/convert.rs
```diff
@@ -1656,4 +1656,45 @@ THd+9FBxiHLGXNKhG/FRSyREXEt+NyYIf/0cyByc9tNksat794ddUqnLOg0vwSkv
...
+                                            "request": {
+                                                "headers": [
+                                                    [
+                                                        "secret",
+                                                        "A1BBC234QWERTY0987MNBV012765HJKL"
+                                                    ],
+                                                    [
+                                                        "passwd",
+                                                        "my_password"
+                                                    ]
```
jjbayer: Can we split this up to test both the key-detection and the value detection?

```suggestion
                                                    [
                                                        "passwd",
                                                        "asdf123"
                                                    ],
                                                    [
                                                        "something_else",
                                                        "my_password"
                                                    ]
```

relay-pii/src/processor.rs
```diff
@@ -202,6 +244,69 @@ impl<'a> Processor for PiiProcessor<'a> {
...
+        value: &mut Array<T>,
+        _meta: &mut Meta,
+        state: &ProcessingState<'_>,
+    ) -> ProcessingResult
+    where
+        T: ProcessValue,
+    {
+        self.is_pair = state.depth() == 0 && value.len() == 2;
+        if self.is_pair {
+            let value_type = ValueType::for_field(&value[0]);
```
jjbayer:
```suggestion
            let key_type = ValueType::for_field(&value[0]);
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Internal**:
+
+- Allow treating arrays of two elements as key-value pairs during PII scrubbing. ([#3639](https://github.com/getsentry/relay/pull/3639))
```
jjbayer: I would move this to features because it's something users are interested in.

```suggestion
- Treat arrays of pairs as key-value mappings during PII scrubbing. ([#3639](https://github.com/getsentry/relay/pull/3639))
```

relay-pii/src/processor.rs
```diff
@@ -202,6 +244,69 @@ impl<'a> Processor for PiiProcessor<'a> {
     }
 }
 
+#[derive(Default)]
+struct PairListProcessor {
```
jjbayer, line 248:
```suggestion
/// Checks whether an array is a pair with a string key.
struct PairCheck {
```

jjbayer (approved):

PR #3641, opened 2024-05-23
feat(spans): Also write inp metrics to transactions name space

edwardgou-sentry:
Right now INP metrics are span metric distributions with percentiles. We rely on these percentiles in the product UI.

There are plans to disable storing percentiles for distributions under the `spans` namespace, so we must move INP metrics to the `transactions` namespace as a workaround to keep INP percentiles. The plan is to extract these `transaction` namespace INP metrics for 90 days until we have enough data to switch over and deprecate the `spans` namespace INP metrics.

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -553,6 +553,101 @@ pub fn hardcoded_span_metrics() -> Vec<(String, Vec<MetricSpec>)> {
                 },
                 MetricSpec {
                     category: DataCategory::Span,
+                    mri: "d:transactions/webvital.score.total@ratio".into(),
+                    field: Some("span.measurements.score.total.value".into()),
+                    condition: Some(is_allowed_browser.clone()),
+                    tags: vec![
+                        Tag::with_key("span.op")
+                            .from_field("span.sentry_tags.op")
+                            .always(),
+                        Tag::with_key("transaction.op")
+                            .from_field("span.sentry_tags.transaction.op")
+                            .always(),
+                        Tag::with_key("transaction")
+                            .from_field("span.sentry_tags.transaction")
+                            .always(),
+                        Tag::with_key("environment")
+                            .from_field("span.sentry_tags.environment")
+                            .always(),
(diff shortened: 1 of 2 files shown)


relay-dynamic-config/src/defaults.rs
```diff
@@ -551,6 +551,101 @@ pub fn hardcoded_span_metrics() -> Vec<(String, Vec<MetricSpec>)> {
                             .always(), // already guarded by condition on metric
                     ],
                 },
+                MetricSpec {
+                    category: DataCategory::Span,
+                    mri: "d:transactions/webvital.score.total@ratio".into(),
+                    field: Some("span.measurements.score.total.value".into()),
```
jjbayer, line 557: This is already extracted as `d:transactions/measurements.score.total@ratio` ([PR](https://github.com/getsentry/relay/pull/3154)):

https://github.com/getsentry/relay/blob/5dff9661d2bd37d77290a01581c27b563d5b632a/relay-dynamic-config/src/defaults.rs#L815-L816

@edwardgou-sentry can we get rid of `d:transactions/measurements.score.total@ratio` in favor of the new `d:transactions/webvital.score.total@ratio`? Or is the one with `measurements.` still used by the product? Asking because the `measurements.` one has caused [problems](https://github.com/getsentry/relay/pull/3324/) before.

Either way we should keep only one of the two metrics.

jjbayer (changes requested):

iambriccardo (approved):
Changes LGTM, will leave it to @phacops for the final approval.

relay-dynamic-config/src/defaults.rs
edwardgou-sentry: Yeah I think you're right, we probably don't need two of these

edwardgou-sentry:
I've heard that the team is now planning to make an exception to keep percentiles for the inp metric (`d:spans/webvital.inp@millisecond`). So we actually do not need this change anymore 😅 

Closing this pr for now, but we can reopen and discuss if it becomes needed again

PR #3645, opened 2024-05-24
ref(spans): Limit spans on transactions to 800kib

Dav1dde:
Limits the maximum size of spans to 800kib to stay within the 1mb limit for transactions.

Due to normalization Relay may add more information to spans which can push the event over the 1mib limit causing upstream Relays to reject event, which was initially accepted. Limits the spans to 800kib as a best effort approach to not go past the limit.

Technically it is still possible to go over the limit, but reducing the maximum size for spans even further may compromise existing experiences.

Relates to: https://github.com/getsentry/sentry/issues/68331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Bug fixes**:
 
 - Apply globally defined metric tags to legacy transaction metrics. ([#3615](https://github.com/getsentry/relay/pull/3615))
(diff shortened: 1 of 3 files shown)


jjbayer (approved):
Could you please add a test case (e.g. programmatically constructed) to see what the trimmed event payload would look like?

PR #3646, opened 2024-05-24
test(spans): Balance timeouts to reduce flakes

### LINKED ISSUE #3607: [Flaky CI]: test_dynamic_sampling
### Flakiness Type

Assertion failure

### Name of Test

tests/integration/test_spans.py::test_dynamic_sampling[1.0]

### Link to Test Run

https://github.com/getsentry/relay/actions/runs/9103390226/job/25025141854?pr=3604

### Details

_No response_

Dav1dde:
Fixes: #3607

#skip-changelog

### DIFF
--- a/tests/integration/test_spans.py
+++ b/tests/integration/test_spans.py
@@ -1862,10 +1862,10 @@ def test_dynamic_sampling(
     if sample_rate == 1.0:
         spans = list(spans_consumer.get_spans(timeout=10, max_attempts=4))
         assert len(spans) == 4
-        outcomes = outcomes_consumer.get_outcomes(timeout=0.1)
+        outcomes = outcomes_consumer.get_outcomes(timeout=3)
         assert summarize_outcomes(outcomes) == {(16, 0): 4}  # SpanIndexed, Accepted
     else:
-        outcomes = outcomes_consumer.get_outcomes(timeout=10)
+        outcomes = outcomes_consumer.get_outcomes(timeout=3)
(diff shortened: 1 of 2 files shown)


tests/integration/test_spans.py
```diff
@@ -1862,10 +1862,10 @@ def summarize_outcomes(outcomes):
     if sample_rate == 1.0:
         spans = list(spans_consumer.get_spans(timeout=10, max_attempts=4))
         assert len(spans) == 4
-        outcomes = outcomes_consumer.get_outcomes(timeout=0.1)
+        outcomes = outcomes_consumer.get_outcomes(timeout=3)
         assert summarize_outcomes(outcomes) == {(16, 0): 4}  # SpanIndexed, Accepted
     else:
-        outcomes = outcomes_consumer.get_outcomes(timeout=10)
+        outcomes = outcomes_consumer.get_outcomes(timeout=3)
```
jjbayer: The reason the `else` timeout was longer is that in the `if` branch, we're already waiting for 4 spans, so more time has passed. This is also why the test fails now. Maybe `(1, 10)` would already work.

jjbayer (approved):
Pre-approving for whatever combination of timeout works.

tests/integration/test_spans.py
Dav1dde: This is so annoying, they are not flaky at all locally. I didn't wanna set a too high timeout just because that actually increases the runtime of the tests.

PR #3650, opened 2024-05-24
feat(spans): Extracts Queue tags and measurements from transaction trace context 

edwardgou-sentry:
Updates the "create span from transaction" code path to extract:
- `messaging.destination.name` and `messaging.message.id` as tags from `contexts.trace.data`
- `messaging.message.receive.latency`, `messaging.message.body.size`, and `messaging.message.retry.count` from `contexts.trace.data`

only for Queue transactions.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/normalize/span/tag_extraction.rs
+++ b/relay-event-normalization/src/normalize/span/tag_extraction.rs
@@ -187,7 +187,8 @@ pub fn extract_span_tags(event: &Event, spans: &mut [Annotated<Span>], max_tag_v
     // TODO: To prevent differences between metrics and payloads, we should not extract tags here
     // when they have already been extracted by a downstream relay.
     let shared_tags = extract_shared_tags(event);
-    let shared_measurements = extract_shared_measurements(event);
+    let segment_tags = extract_segment_tags(event);
+    let segment_measurements = extract_segment_measurements(event);
     let is_mobile = shared_tags
         .get(&SpanTagKey::Mobile)
         .is_some_and(|v| v.as_str() == "true");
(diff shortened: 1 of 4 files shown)


iambriccardo (approved):
LGTM

phacops (commented):
I thought we wanted to tag only the segment span with some new values, this will tag all spans extracted from the transaction. Is this what we want?

relay-event-schema/src/protocol/contexts/trace.rs
```diff
@@ -161,6 +161,21 @@ pub struct Data {
...
+    pub messaging_message_id: Annotated<String>,
+
+    #[metastructure(field = "messaging.message.receive.latency")]
+    pub messaging_message_receive_latency: Annotated<Value>,
+
+    #[metastructure(field = "messaging.message.retry.count")]
+    pub messaging_message_retry_count: Annotated<Value>,
+
+    #[metastructure(field = "messaging.message.body.size")]
+    pub messaging_message_body_size: Annotated<Value>,
```
jjbayer: Have you discussed this schema with SDK teams? It seems like a separate `messaging` context could make more sense.

jjbayer: Please add doc comments for these fields.

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -211,6 +212,17 @@ pub fn extract_span_tags(event: &Event, spans: &mut [Annotated<Span>], max_tag_v
...
         );
 
+        if !shared_measurements.is_empty() {
+            span.measurements
+                .get_or_insert_with(Default::default)
+                .extend(
+                    shared_measurements
+                        .clone()
+                        .into_iter()
+                        .map(|(k, v)| (k.to_owned(), Annotated::new(v))),
```
```suggestion
                        .iter()
                        .map(|(k, v)| (k.clone(), Annotated::new(v.clone()))),
```

relay-event-schema/src/protocol/contexts/trace.rs
edwardgou-sentry: 👍 added!

edwardgou-sentry:
> I thought we wanted to tag only the segment span with some new values, this will tag all spans extracted from the transaction. Is this what we want?

Good point. I've updated the tags and measurements to only get extracted and applied to segment spans:
https://github.com/getsentry/relay/pull/3650/files#diff-313f39c52849630c10c4f6da270c2ed0f99d7deac8894a838bd166dddc94b19dR216

Also added a new test to make sure that these tags/measurements don't get propagated to child spans.

relay-event-schema/src/protocol/contexts/trace.rs
edwardgou-sentry: We discussed some other options, but haven't thought about a new `messaging` context yet.

@cleptric What do you think?

cleptric: @jjbayer we use `span.data` for this module. As a transaction has no data attribute, it ends up on the trace context. I would like to avoid adding a one of context to the SDK if possible but also get the point about it in Relay 😬

jjbayer: OK, if `contexts.trace.data` is the transaction field corresponding to `span.data` then this makes sense.

jjbayer (approved):

PR #3653, opened 2024-05-27
chore(gocd): update pops targets canaries

gi0baro:
SSIA

#skip-changelog

### Legal Boilerplate


jjbayer (approved):

Dav1dde (approved):
Can drop all the legal stuff from the PR description, we don't need that in the commit message.

PR #3661, opened 2024-05-29
feat(metrics): Add support for partitioning in the aggregator

iambriccardo:
This PR implements an improved aggregator for metrics that is aware of partitioning and dispatches buckets with or without partitions.

The goal of making the aggregator partition aware is to maximize the number of buckets of the same partition at every flush cycle. Before, this was not possible since flushing was happening without considering partitions. As opposed to `None` shifting, now the partitions are spread out over time, which avoids request spikes.

### DIFF
--- a/relay-config/src/config.rs
+++ b/relay-config/src/config.rs
@@ -14,7 +14,7 @@ use relay_kafka::{
     ConfigError as KafkaConfigError, KafkaConfigParam, KafkaParams, KafkaTopic, TopicAssignment,
     TopicAssignments,
 };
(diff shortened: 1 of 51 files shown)


relay-config/src/config.rs
```diff
@@ -2398,7 +2393,8 @@ impl Config {
             max_project_key_bucket_bytes,
             initial_delay: 30,
             debounce_delay: 10,
-            shift_key: ShiftKey::Project,
+            flush_partitions: None,
+            flush_batching: FlushBatching::Project,
```
iambriccardo, line 2397: Maybe we can find a better name.

jan-auer (approved):

PR #3663, opened 2024-05-29
feat(spans): Forward origin on the Kafka message

phacops:
Data team would like to store it on BigQuery so we need to add it to the message.

#skip-changelog

### DIFF
--- a/relay-server/src/services/store.rs
+++ b/relay-server/src/services/store.rs
@@ -1361,6 +1361,8 @@ struct SpanKafkaMessage<'a> {
     #[serde(default)]
     organization_id: u64,
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    origin: Option<&'a str>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     parent_span_id: Option<&'a str>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     profile_id: Option<&'a str>,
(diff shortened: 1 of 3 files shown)


relay-server/src/services/store.rs
```diff
@@ -1361,6 +1361,8 @@ struct SpanKafkaMessage<'a> {
     #[serde(default)]
     organization_id: u64,
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    origin: Option<&'a str>,
```
jjbayer: Is the `origin` guaranteed to be ASCII? If not, we should probably use `#[serde(borrow)] Cow<'a, str>` to not fail deserialization.

https://serde.rs/lifetimes.html#borrowing-data-in-a-derived-impl

jjbayer (approved):

relay-server/src/services/store.rs
Dav1dde: Good catch!

Didn't we also have [`RawValue`](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) for this?

Dav1dde (approved):

relay-server/src/services/store.rs
phacops: We could go either way, although, it'd rather at least enforce it's a string.

PR #3667, opened 2024-05-30
fix(spans): Accept a UNIX timestamp for measurements

phacops:
Sample format V2 works with UNIX timestamps as opposed to relative timestamps and since we're sharing the measurements struct, we need to make it compatible.

#skip-changelog

### DIFF
--- a/relay-profiling/src/measurements.rs
+++ b/relay-profiling/src/measurements.rs
@@ -1,6 +1,6 @@
 use serde::{Deserialize, Serialize};
 
-use crate::utils::deserialize_number_from_string;
+use crate::utils::{deserialize_number_from_string, is_zero};
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct Measurement {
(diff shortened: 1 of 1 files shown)


JonasBa (approved):

relay-profiling/src/measurements.rs
```diff
@@ -11,12 +11,20 @@ pub struct Measurement {
...
+    )]
     elapsed_since_start_ns: u64,
 
     // Android 6.8.0 sends a string instead of a float64 so we need to accept both
     #[serde(deserialize_with = "deserialize_number_from_string")]
     value: f64,
+
+    // Sample Format V2 accepts a timestamp in seconds as a float
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    timestamp: Option<f64>,
```
jjbayer: Will a measurement have either a timetamp or `elapsed_since_start_ns`? Because then an enum might be the better representation here.

jjbayer (approved):
Could you please copy-paste one of the unit tests at the bottom of this file to test the new format?

PR #3669, opened 2024-05-31
ref(metrics): Switches set hashing algo to crc32

Dav1dde:
The SDKs already use CRC32, to stay compatible with the SDKs we should use crc32 as well. This may break some sets for usecases where the SDK did *not* hash. But overall it delivers a more consistent behaviour.

All existing usecases with metrics extraction keep using fnv32 to not break the product.


Dav1dde:
Not necessary anymore. Metrics will change in the future in a away where there is no longer any hashing on the SDK side, which means we can keep using fnv32 for now in Realy.

PR #3670, opened 2024-05-31
fix(spans): Do not trim essential fields

jjbayer:
Since https://github.com/getsentry/relay/pull/3645, we're seeing incomplete or missing trace IDs on spans (see linked sentry issues). What happens is that after visiting a long `span.description`, there is no budget left for essential fields and the trimming processor drops or trims them.

* Introduce `trim = "false"` field attribute to disallow trimming of specific fields.
* Apply the new field to `span.trace_id` and other essential fields to prevent them from being partially scrubbed or dropped entirely.

Current limitations:

- Not trimming these fields leads to over-accepting bytes. This should not matter as long as the fields that have `trim = "false"` are limited in size (which they are currently not, see https://github.com/getsentry/relay/pull/3535 as follow-up).
  - The more correct solution would be to subtract the length of `trim = "false"` fields _before_ trimming any other fields.
  - Alternatively, we could decide to never over-accept, and drop items after visiting them if the budget is negative. I implemented this in [1136f76](https://github.com/getsentry/relay/pull/3670/commits/1136f76d983422fa2f4ae240197d7e43ef2b6435) but it resulted in more failing tests, because the current trimming processor already over-accepts a little bit (key lengths), and I did not want to touch the default behavior. cc @Dav1dde.

Fixes [RELAY-2NY7](https://sentry.my.sentry.io/organizations/sentry/issues/1005495/), [RELAY-2NY8](https://sentry.my.sentry.io/organizations/sentry/issues/1005496/), [SNUBA-552](https://sentry.sentry.io/issues/5429321769/).


relay-event-derive/src/lib.rs
```diff
@@ -595,6 +605,17 @@ fn parse_field_attributes(
                                         panic!("Got non string literal for retain");
                                     }
                                 }
+                            } else if ident == "trim" {
+                                match name_value.lit {
+                                    Lit::Str(litstr) => match litstr.value().as_str() {
```
Dav1dde, line 610: Could make it a 'real' boolean, `trim = true` and use `Lit::Bool` here.

relay-event-derive/src/lib.rs
```diff
@@ -595,6 +605,17 @@ fn parse_field_attributes(
...
+                                        "true" => rv.trim = None,
+                                        "false" => rv.trim = Some(false),
+                                        other => panic!("Unknown value {other}"),
+                                    },
+                                    _ => {
+                                        panic!("Got non string literal for retain");
```
Dav1dde:
```suggestion
                                        panic!("Got non string literal for trim");
```

Dav1dde (commented):
a

relay-event-derive/src/lib.rs
jjbayer: That would be better, but I want to stay consistent with `required = "true"` and `pii = "true"`.

Dav1dde (approved):

PR #3678, opened 2024-06-03
ci(ci): Enable self hosted checks for forks

dav1d:
#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -645,7 +645,7 @@ jobs:
         run: docker load -i relay-docker-image
 
       - name: Run Sentry self-hosted e2e CI
-        uses: getsentry/action-self-hosted-e2e-tests@main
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
hubertdeng123, line 1: The `Build and push to ghcr.io` in `build-docker` job is being skipped due to being a fork, We'll need to change this too if we 'd like to enable self-hosted checks for forks

Dav1dde: The job now imports the image from a build artifact if access to ghcr is not available.

Dav1dde:
Mh still doesn't work because the install script also tries to pull https://github.com/getsentry/self-hosted/blob/c40b1530d104e9481d172e06dc6810bea0795d5c/install/ensure-relay-credentials.sh#L27

Dav1dde:
We don't want to update the self hosted install scripts to accommodate for this.

PR #3680, opened 2024-06-03
ref: stop writing relay image to gcr

asottile-sentry:
#skip-changelog


iambriccardo (approved):

Dav1dde (approved):
Should probably also rename the job name 

```
- name: Publish Relay to GCR
+ name: Publish Relay to AR
```

PR #3693, opened 2024-06-06
feat(metrics): Add metric to track how many times a partition overflows

iambriccardo:
This PR adds a metric to track how many times a partition overflows. It will be helpful to inform a decision on whether we should increase the size of the request.

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -2541,7 +2541,7 @@ impl EnvelopeProcessorService {
 
         let batch_size = self.inner.config.metrics_max_batch_size_bytes();
         let mut partition = Partition::new(batch_size);
-        let mut partition_overflowed = false;
+        let mut partition_splits = 0;
 
         for (scoping, message) in &scopes {
             let ProjectMetrics { buckets, .. } = message;
@@ -2556,14 +2556,14 @@ impl EnvelopeProcessorService {
                         // always result in a request, otherwise we would enter an endless loop.
                         self.send_global_partition(partition_key, &mut partition);
                         remaining = Some(next);
-                        partition_overflowed = true;
+                        partition_splits += 1;
                     }
                 }
             }
         }
 
-        if partition_overflowed {
-            metric!(counter(RelayCounters::PartitionOverflow) += 1);
+        if partition_splits > 0 {
+            metric!(histogram(RelayHistograms::PartitionSplits) = partition_splits);
         }
 
         self.send_global_partition(partition_key, &mut partition);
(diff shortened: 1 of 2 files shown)


relay-server/src/services/processor.rs
```diff
@@ -2550,6 +2550,7 @@ impl EnvelopeProcessorService {
 
                 while let Some(bucket) = remaining.take() {
                     if let Some(next) = partition.insert(bucket, *scoping) {
+                        metric!(counter(RelayCounters::PartitionOverflow) += 1);
```
iker-barriocanal: Please, collect the count outside the loop and emit a single metric. (I'm not sure if the statsd proxy solves this indirectly, but it doesn't harm doing it)

The metric itself LGTM, but note that this metric may generate a very high volume and Relay already generates too many metrics. Once the decision is made, we should remove this metric.

iker-barriocanal (approved):

relay-server/src/services/processor.rs
iker-barriocanal: Also -- I haven't checked if we have a metric for this use case already

iambriccardo: Good point, will do!

relay-server/src/services/processor.rs
```diff
@@ -2555,11 +2556,16 @@ impl EnvelopeProcessorService {
...
                         remaining = Some(next);
+                        partition_overflowed = true;
                     }
                 }
             }
         }
 
+        if partition_overflowed {
+            metric!(counter(RelayCounters::PartitionOverflow) += 1);
+        }
```
Dav1dde: I think a distribution telling us into how many parts one flush was split would be interesting. This would also not be skewed by how many times we flushed.

Dav1dde (commented):
I think we should tag the byte size metric in the upstream by the route/request instead.

relay-server/src/services/processor.rs
iambriccardo: Yeah I could add this instead

relay-server/src/services/processor.rs
Dav1dde: I think we should keep this metric and add it to our dashboards. This is one of the things which is heavily influenced by the amount of metrics Relay creates and receives and may need re-tuning in the future.

Dav1dde: But +1 in keeping the count outside, doing a boolean is not the same as keeping the count.

Dav1dde (approved):

PR #3694, opened 2024-06-06
fix(profiles): Fix unsampled profiles being dropped in processing Relays

Dav1dde:
Currently if transactions with profiles are unsampled in a processing Relay they are immediately dropped right after (integration test in `default` mode fails without these changes).

The changes hopefully also simplify the profile logic a bit and pull more logical decisions directly into processing function instead of it being hidden away in a bunch of implicit state.

Also:
- adds a few more feature flag checks, standalone profiles were never checked for whether the feature flags are active. 
- with the unsampled profiles flag, the feature flag for profiles was never actually checked.

#skip-changelog

### DIFF
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -1510,12 +1510,17 @@ impl EnvelopeProcessorService {
         &self,
         state: &mut ProcessEnvelopeState<TransactionGroup>,
     ) -> Result<(), ProcessingError> {
-        profile::filter(state);
+        let global_config = self.inner.global_config.current();
+
+        dynamic_sampling::ensure_dsc(state);
+
         event::extract(
             state,
             &self.inner.config,
             &self.inner.global_config.current(),
         )?;
+
+        profile::filter(state);
         profile::transfer_id(state);
 
         if_processing!(self.inner.config, {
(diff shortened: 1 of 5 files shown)


relay-server/src/services/processor.rs
```diff
@@ -1524,57 +1529,70 @@ impl EnvelopeProcessorService {
...
+        // Always run dynamic sampling on processing Relays,
+        // but delay decision until inbound filters have been fully processed.
+        let sampling_result =
+            if self.inner.config.processing_enabled() || matches!(filter_run, FiltersStatus::Ok) {
+                dynamic_sampling::run(state, &self.inner.config)
+            } else {
+                SamplingResult::NoMatch
+            };
 
-            dynamic_sampling::sample_envelope_items(
```
Dav1dde, line 1541: Removes the event from the envelope (and expands profiles).

relay-server/src/services/processor.rs
```diff
@@ -1524,57 +1529,70 @@ impl EnvelopeProcessorService {
...
-                &self.inner.global_config.current(),
-            );
+        // We avoid extracting metrics if we are not sampling the event while in non-processing
+        // Relays, in order to synchronize rate limits on indexed and processed transactions.
+        if self.inner.config.processing_enabled() || sampling_result.should_drop() {
+            self.extract_transaction_metrics(state, &sampling_result)?;
         }
 
-        if_processing!(self.inner.config, {
-            profile::process(state, &self.inner.config);
```
Dav1dde, line 1550: This removes all profiles if there is no event -> The Bug.

relay-server/src/services/processor/profile.rs
```diff
@@ -16,18 +15,23 @@ use crate::utils::ItemAction;
...
 pub fn filter<G>(state: &mut ProcessEnvelopeState<G>) {
-    let transaction_count: usize = state
-        .managed_envelope
-        .envelope()
-        .items()
-        .filter(|item| item.ty() == &ItemType::Transaction)
-        .count();
+    let profiling_enabled = state.project_state.has_feature(Feature::Profiling);
+    let has_transaction = state.event_type() == Some(EventType::Transaction);
+    let unsampled_profiles = state
```
jjbayer:
```suggestion
    let keep_unsampled_profiles = state
```

relay-server/src/services/processor/profile.rs
```diff
@@ -78,7 +81,9 @@ pub fn process(state: &mut ProcessEnvelopeState<TransactionGroup>, config: &Conf
             if !profiling_enabled {
                 return ItemAction::DropSilently;
             }
-            // If we don't have an event at this stage, we need to drop the profile.
+
+            // There should always be an event/transaction available at this stage.
+            // It is required to expand the profle. If it's missing, drop the item.
```
jjbayer:
```suggestion
            // It is required to expand the profile. If it's missing, drop the item.
```

jjbayer (approved):

PR #3695, opened 2024-06-06
ref(profiles): Remove profile_id from state, cleanup transfer logic

Dav1dde:
Cleanup the `profile_id` handling.

No need to do split that logic into two parts, also removes the profile_id from the state and instead explicitly handles it as a returned argument.

Requires: #3694 

#skip-changelog


jjbayer (approved):
Nice, the lost art of maintaining state in local variables.

iker-barriocanal (approved):

PR #3697, opened 2024-06-06
feat(spans): Group function spans

phacops:
We don't calculate a group hash for function spans and this means the span summary shows one group only for this op. We'd like to show each function individually.

This will reduce cardinality in function names by removing UUIDs or hexadecimal strings equal or longer than 5 characters and calculate the group hash, enabling us to store span metrics for each function separately.

### DIFF
--- a/relay-event-normalization/src/normalize/span/description/mod.rs
+++ b/relay-event-normalization/src/normalize/span/description/mod.rs
@@ -13,8 +13,8 @@ use std::path::Path;
 use url::{Host, Url};
 
 use crate::regexes::{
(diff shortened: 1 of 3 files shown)


relay-event-normalization/src/normalize/span/description/mod.rs
```diff
@@ -1072,6 +1077,20 @@ mod tests {
...
     span_description_test!(db_prisma, "User find", "db.sql.prisma", "User find");
 
+    span_description_test!(
+        function_python,
+        "sentry.event_manager.assign_event_to_group",
+        "function",
+        "sentry.event_manager.assign_event_to_group"
+    );
+
+    span_description_test!(
```
iker-barriocanal, line 1087: Can we add a test to validate the expected scrubbing of the regex?

iker-barriocanal (approved):

PR #3699, opened 2024-06-07
ci(ci): Switch kafka with redpanda

Dav1dde:


Dav1dde:
Doesn't seem to be faster.

PR #3700, opened 2024-06-07
test(kafka): Streamline polling of multiple messages

Dav1dde:
#skip-changelog

### DIFF
--- a/tests/integration/fixtures/processing.py
+++ b/tests/integration/fixtures/processing.py
@@ -187,10 +187,16 @@ class ConsumerBase:
                 self.assert_empty()
                 break
 
+            # Wait the full timeout duration if we're polling for an exact number of
+            # of messages, otherwise use a shorter timeout to keep tests faster.
+            # The rational being that once an item arrives on the topic the others
+            # are quick to follow.
             message = self.poll(min(2, timeout) if n is None else timeout)
 
         if n is not None:
-            assert n == messages, f"Expected {n} messages, only got {messages}"
+            assert (
+                n == messages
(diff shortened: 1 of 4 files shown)


tests/integration/fixtures/processing.py
```diff
@@ -167,6 +167,31 @@ def poll(self, timeout=None):
...
+        message = self.poll(timeout)
+        while message is not None:
+            yield message
+            messages += 1
+
+            if messages == n:
+                self.assert_empty()
+                break
+
+            message = self.poll(min(2, timeout) if n is None else timeout)
```
jjbayer, line 194: Can't we use `min(2, timeout)` even if `n is not None`?

jjbayer, line 194: Should the 2 be configurable? E.g. `def poll_many(self, initial_timeout=None, subsequent_timeout=2, n=None)`?

jjbayer (commented):
Nice!

tests/integration/fixtures/processing.py
Dav1dde: It makes everything a bit more fragile then, if you expect a certain amount of items and they are not there we can afford to wait a little bit longer (should never happen). But if for some reason something actually is a bit slower we won't fail with the longer timeout.

Dav1dde: Yes, I haven't done that yet because I couldn't decide on good names.

jjbayer: Makes sense.

jjbayer: nit: Can you put that reasoning in a code comment?

jjbayer (approved):

PR #3702, opened 2024-06-07
test(kafka): Assert consumers empty at the end of the test

Dav1dde:
#skip-changelog

### DIFF
--- a/tests/integration/fixtures/processing.py
+++ b/tests/integration/fixtures/processing.py
@@ -1,9 +1,7 @@
 import json
-from typing import Optional
 import msgpack
 import uuid
 
-import functools
 import pytest
 import os
 import confluent_kafka as kafka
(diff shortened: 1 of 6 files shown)


tests/integration/fixtures/processing.py
```diff
@@ -270,89 +272,123 @@ def assert_rate_limited(
             assert count == quantity
 
 
+def consumer_fixture(f):
+    consumer: list[Optional[ConsumerBase]] = [None]
```
jjbayer: I think you can achieve this with `nonlocal`.

tests/integration/fixtures/processing.py
```diff
@@ -270,89 +272,123 @@ def assert_rate_limited(
...
+    if consumer[0] is not None:
+        consumer[0].assert_empty()
+
+
 @pytest.fixture
 def events_consumer(kafka_consumer):
-    return lambda timeout=None: EventsConsumer(
-        timeout=timeout, *kafka_consumer("events")
+    yield from consumer_fixture(
+        lambda timeout=None: EventsConsumer(timeout=timeout, *kafka_consumer("events"))
```
jjbayer: Can we make this `yield from consumer_fixture(EventsConsumer, "events")`? I.e. move the lambda and the call to `kafka_consumer` into`consumer_fixture`.

jjbayer (approved):

PR #3706, opened 2024-06-07
fix(replays): Run TrimmingProcessor on replays

iker-barriocanal:
Although some fields in replays have `max_chars` defined, the trimming processor isn't run and thus the limit is not enforced. This PR runs the processor and enforces that limit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Bug fixes**:
 
-- Trim some fields in replays to the maximum expected length. ([#3706](https://github.com/getsentry/relay/pull/3706))
+- Trim fields in replays to the maximum expected length. ([#3706](https://github.com/getsentry/relay/pull/3706))
 
 ## 24.5.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Bug fixes**:
+
+- Trim some fields in replays to the maximum expected length. ([#3706](https://github.com/getsentry/relay/pull/3706))
+
 ## 24.5.1
 
 **Bug fixes**:
--- a/relay-event-normalization/src/replay.rs
+++ b/relay-event-normalization/src/replay.rs
@@ -2,10 +2,12 @@
 
 use std::net::IpAddr as StdIpAddr;
 
+use relay_event_schema::processor::{self, ProcessingState, Processor};
 use relay_event_schema::protocol::{Contexts, IpAddr, Replay};
 use relay_protocol::Annotated;
 
 use crate::normalize::user_agent;
+use crate::trimming;
 use crate::user_agent::RawUserAgentInfo;
 
 /// Replay validation or normalization error.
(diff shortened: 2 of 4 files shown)


relay-server/src/services/processor/replay.rs
```diff
@@ -175,12 +175,14 @@ fn process_replay_event(
     let mut replay =
         Annotated::<Replay>::from_json_bytes(payload).map_err(ReplayError::CouldNotParse)?;
 
+    if replay.value().is_none() {
+        return Err(ReplayError::NoContent);
+    }
+
     if let Some(replay_value) = replay.value_mut() {
```
jjbayer: This should work, `replay_value` goes out of scope before `replay` is used again.

```suggestion
    let Some(replay_value) = replay.value_mut() else {
        return Err(ReplayError::NoContent);
    }
    replay::validate(replay_value)?;
    replay::normalize(&mut replay, client_ip, user_agent);
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Bug fixes**:
+
+- Trim some fields in replays to the maximum expected length. ([#3706](https://github.com/getsentry/relay/pull/3706))
```
jjbayer: It applies to all fields that have a `max_size`, right?

```suggestion
- Trim fields in replays to the maximum expected length. ([#3706](https://github.com/getsentry/relay/pull/3706))
```

jjbayer (approved):

CHANGELOG.md
iker-barriocanal: > It applies to all fields that have a max_size, right?

Yes, but not all fields have a max_size. I'm open to change it but not sure if it's better.

jjbayer: `some` sounds arbitrary to me, that's why I would remove it. Maybe 
```suggestion
- Trim fields in replays to their defined maximum length. ([#3706](https://github.com/getsentry/relay/pull/3706))
```

relay-event-normalization/src/replay.rs
```diff
@@ -462,4 +469,32 @@ mod tests {
         let validation_result = validate(replay.value_mut().as_mut().unwrap());
         assert!(validation_result.is_err());
     }
+
+    #[test]
+    fn test_maxchars_trimming() {
+        let json = format!(r#"{{"dist": "{}"}}"#, "0".repeat(100));
+        let mut replay = Annotated::<Replay>::from_json(json.as_str()).unwrap();
```
bruno-garcia, line 476: test only stresses dist, should it test the other fields too?

cmanallen (approved):
As Bruno said more test coverage is appreciated if necessary.

The max length is defined by the annotation on the fields correct?  Does this mean string trimming wasn't working before?

iker-barriocanal:
> The max length is defined by the annotation on the fields correct?

Yep, that's correct.

> Does this mean string trimming wasn't working before?

Yes, let me know if I should clarify further the PR description.

> test only stresses dist, should it test the other fields too?

The test validates that the trimming processor runs on replays, and there are separate tests to validate that the trimming processor trims fields with `max_chars`, so it's ok to move forward with the current state. We can add more tests if there are known edge cases.

PR #3710, opened 2024-06-10
test(spans): Rate limits work for extracted spans

jjbayer:
Preparing to enable spike protection for spans.

#skip-changelog


tests/integration/test_spans.py
```diff
@@ -1588,10 +1595,37 @@ def summarize_outcomes():
     # one for the transaction, one for the contained span
     assert len(spans) == 2
     assert summarize_outcomes() == {(16, 0): 2}  # SpanIndexed, Accepted
+    # A limit only for span_indexed does not affect extracted metrics
+    metrics = metrics_consumer.get_metrics(timeout=1, n=10)
```
Dav1dde:
```suggestion
    metrics = metrics_consumer.get_metrics(n=10)
```

I think we shouldn't specify a timeout if not necessary for gets with a specific `n`.

Dav1dde (approved):

PR #3711, opened 2024-06-10
Remove rollout option for inline attachments

Swatinem:
We haven’t seen any problems since rolling this out, so make it the default.

#skip-changelog
☝🏻 unless someone has a different opinion?


Dav1dde (approved):

Dav1dde:
Can also remove the option from Sentry afterwards 🎉

Swatinem:
> Can also remove the option from Sentry afterwards 🎉

will do, yes

PR #3712, opened 2024-06-10
ref(rate-limits): Always enforce transactions on the indexed category

Dav1dde:
Always use the same code path for rate limits no matter if metrics extraction for transactions is enabled or not.

- effectively already the case because it is always enabled
- in the case of a disabled transaction metrics config (broken, outdated or too new) enforcing a limit will still drop the transaction but only emit an outcome for the 'base' and not the 'indexed' category. Compare this to enforcing cached rate limits with a valid config, it would emit an outcome for both categories. These limits are independent to having metrics extraction or not and should behave equally. In the case after metrics extraction there is already additional logic which takes care of emitting the correct categories.

#skip-changelog

### DIFF
--- a/relay-server/src/utils/rate_limits.rs
+++ b/relay-server/src/utils/rate_limits.rs
@@ -521,26 +521,6 @@ where
         Ok((enforcement, rate_limits))
     }
 
-    /// Returns a dedicated data category for indexing if metrics are to be extracted.
-    ///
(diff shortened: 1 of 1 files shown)


iker-barriocanal (approved):

Dav1dde:
Doesn't work because in the other branch the quota is not actually consumed, it is checked and consumed with the metric.

PR #3713, opened 2024-06-10
feat(spans): Correctly rate limit spans after extracting from transaction

### LINKED ISSUE #3578: Extracted spans should be rate limited with transaction
When a transaction is rate limited, spans extracted from that span should be rate limited as well.

This works for cached rate limits, but consistent (redis) quota enforcement happens _after_ span extraction. This has only been the case since https://github.com/getsentry/relay/pull/3248, so we might need to revisit the logic of that PR.

Example case: [Spans in discover](https://sentry.sentry.io/discover/homepage/?dataset=spansIndexed&end=2024-05-08T18%3A19%3A52&field=trace&field=transaction.id&field=segment.id&field=span.op&field=span.description&field=timestamp&name=All+Events&project=1&query=transaction.id%3A5f158696052f46609ff7ad0c946072f5&sort=-timestamp&start=2024-05-08T17%3A47%3A49&yAxis=count%28%29), matching [Project Abuse Limit in stats](https://sentry.io/_admin/customers/sentry/projects/sentry/?dataType=indexed_transactions&end=2024-05-08T18%3A06%3A58&start=2024-05-08T17%3A55%3A00&utc=true).

iambriccardo:
This PR correctly applies enforcements for rate limiting on child spans whenever the transaction is rate limited.

Closes: https://github.com/getsentry/relay/issues/3578

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 **Bug fixes**:
 
 - Trim fields in replays to their defined maximum length. ([#3706](https://github.com/getsentry/relay/pull/3706))
-- Apply rate limit on child spans when the transaction is rate limited. ([#3713](https://github.com/getsentry/relay/pull/3713))
+- Apply rate limit on extracted spans when the transaction is rate limited. ([#3713](https://github.com/getsentry/relay/pull/3713))
 
 **Internal**:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Bug fixes**:
 
 - Trim fields in replays to their defined maximum length. ([#3706](https://github.com/getsentry/relay/pull/3706))
+- Apply rate limit on child spans when the transaction is rate limited. ([#3713](https://github.com/getsentry/relay/pull/3713))
 
 **Internal**:
 
--- a/relay-server/src/utils/rate_limits.rs
+++ b/relay-server/src/utils/rate_limits.rs
@@ -682,7 +682,7 @@ where
         }
 
         // We want to process spans rate limits only if they were not already applied because a
-        // rate limited transaction had child spans that were also rate limited.
+        // rate limited transaction has also rate limited its child spans.
         if !enforcement.event.is_active() && summary.span_quantity > 0 {
             // Check for rate limits on the main category but do not consume
             // quota. Quota will be consumed by the metrics rate limiter instead.
--- a/relay-server/src/utils/rate_limits.rs
+++ b/relay-server/src/utils/rate_limits.rs
@@ -1,3 +1,4 @@
+use serde_json::error::Category::Data;
 use std::fmt::{self, Write};
 
 use relay_dynamic_config::{ErrorBoundary, ProjectConfig};
@@ -563,6 +564,8 @@ where
                 // Otherwise, the outcome is logged at a different place.
                 if !summary.transaction_metrics_extracted {
                     enforcement.event_metrics = CategoryLimit::new(category, 1, longest);
+                    enforcement.span_metrics =
+                        CategoryLimit::new(DataCategory::Span, summary.span_quantity, longest);
                 }
 
                 // If the main category is rate limited, we drop both the event and metrics. If
@@ -573,10 +576,14 @@ where
                 }
 
                 enforcement.event = CategoryLimit::new(index_category, 1, longest);
+                enforcement.spans =
+                    CategoryLimit::new(DataCategory::SpanIndexed, summary.span_quantity, longest);
             } else {
                 event_limits = (self.check)(scoping.item(category), 1)?;
                 longest = event_limits.longest();
                 enforcement.event = CategoryLimit::new(category, 1, longest);
+                enforcement.spans =
+                    CategoryLimit::new(DataCategory::Span, summary.span_quantity, longest);
             }
 
             // Record the same reason for attachments, if there are any.
--- a/relay-server/src/utils/rate_limits.rs
+++ b/relay-server/src/utils/rate_limits.rs
@@ -568,10 +568,8 @@ where
                 // Otherwise, the outcome is logged at a different place.
                 if !summary.transaction_metrics_extracted {
                     enforcement.event_metrics = CategoryLimit::new(category, 1, longest);
-                    if summary.span_quantity > 0 {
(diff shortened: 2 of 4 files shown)


relay-server/src/utils/rate_limits.rs
```diff
@@ -666,7 +677,12 @@ where
             rate_limits.merge(checkin_limits);
         }
 
-        if summary.span_quantity > 0 {
+        // We want to process spans rate limits only if they were not already applied because the
+        // transaction had child spans.
+        if summary.span_quantity > 0
+            && enforcement.span_metrics.is_default()
+            && enforcement.spans.is_default()
```
iambriccardo: The rationale behind this condition is that, if we have already applied span enforcements, we don't want to try and fetch quotas because:
1. We will waste compute
2. We will override the enforcement with default in case the `check` doesn't succeed

relay-server/src/utils/rate_limits.rs
```diff
@@ -573,10 +580,14 @@ where
...
 
                 enforcement.event = CategoryLimit::new(index_category, 1, longest);
+                enforcement.spans =
+                    CategoryLimit::new(DataCategory::SpanIndexed, summary.span_quantity, longest);
             } else {
                 event_limits = (self.check)(scoping.item(category), 1)?;
                 longest = event_limits.longest();
                 enforcement.event = CategoryLimit::new(category, 1, longest);
+                enforcement.spans =
+                    CategoryLimit::new(DataCategory::Span, summary.span_quantity, longest);
```
jjbayer: nit: I would prefer if we put all the new logic in the `if summary.span_quantity > 0` block so that it's easy to find when we look for span-related logic.

relay-server/src/utils/rate_limits.rs
```diff
@@ -563,6 +568,8 @@ where
                 // Otherwise, the outcome is logged at a different place.
                 if !summary.transaction_metrics_extracted {
                     enforcement.event_metrics = CategoryLimit::new(category, 1, longest);
+                    enforcement.span_metrics =
+                        CategoryLimit::new(DataCategory::Span, summary.span_quantity, longest);
```
jjbayer: `longest` can be none, should we set the enforcement on spans / span_metrics in that case?

relay-server/src/utils/rate_limits.rs
iambriccardo: I did do that in my first iteration but removed because of verbosity and because a category is inactive if it's count is 0 but I can add it back.

relay-server/src/utils/rate_limits.rs
iambriccardo: This was a question I also asked myself and given my lack of context of the product here, I don't know what we expect.

Right now the logic, in case the transaction check fails, it creates a default `span_metrics` so it will actually enter in the span rate limiting below.

Dav1dde (commented):
This is very similar to what we do for profiles and attachments, the code should mirror that.

Dav1dde:
Looked into it a bit with @iambriccardo, it's a bit trickier than profiles and attachments. With my upcoming changes to the rate limiting this will become a bit easier, so I think it's fine to keep it the way it is.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Bug fixes**:
 
 - Trim fields in replays to their defined maximum length. ([#3706](https://github.com/getsentry/relay/pull/3706))
+- Apply rate limit on child spans when the transaction is rate limited. ([#3713](https://github.com/getsentry/relay/pull/3713))
```
jjbayer: We also rate limit the extracted span that represents the transaction itself, which is not a child span.

```suggestion
- Apply rate limit on extracted spans when the transaction is rate limited. ([#3713](https://github.com/getsentry/relay/pull/3713))
```

tests/integration/test_spans.py
```diff
@@ -1662,6 +1703,71 @@ def summarize_outcomes():
...
+    # We have one nested span and the transaction itself becomes a span
+    spans = spans_consumer.get_spans(n=2, timeout=10)
+    assert len(spans) == 2
+    assert summarize_outcomes() == {(16, 0): 2}  # SpanIndexed, Accepted
+
+    # Second batch is limited
+    relay.send_envelope(project_id, envelope)
+    assert summarize_outcomes() == {
+        (2, 2): 1,  # Transaction, RateLimited
+        (12, 2): 2,  # SpanIndexed, RateLimited
```
jjbayer:
```suggestion
        (12, 2): 2,  # Span, RateLimited
```

tests/integration/test_spans.py
```diff
@@ -1662,6 +1703,71 @@ def summarize_outcomes():
...
+    project_id = 42
+    project_config = mini_sentry.add_full_project_config(project_id)
+    project_config["config"]["features"] = [
+        "projects:span-metrics-extraction",
+        "organizations:standalone-span-ingestion",
+        "organizations:indexed-spans-extraction",
+    ]
+    project_config["config"]["quotas"] = [
+        {
+            "categories": ["transaction"],
```
jjbayer: Could you parametrize this with `@pytest.mark.parametrize("category", ["transaction", "transaction_indexed"])`.

jjbayer (approved):

tests/integration/test_spans.py
jjbayer: Might need to fake the `metrics_extracted` flag on the envelope item to see a difference.

tests/integration/test_spans.py
```diff
@@ -1782,10 +1782,10 @@ def test_rate_limit_is_consistent_between_transaction_and_spans(
     This test does not cover consistent enforcement of total spans.
     """
     # TODO: add outcomes check when https://github.com/getsentry/relay/issues/3705 is done.
-    # Ignore 429s
-    mini_sentry.fail_on_relay_error = False
-
     relay = relay_with_processing(options=TEST_CONFIG)
+    # We want to avoid failing on 429s.
+    relay.fail_on_envelope_error = False
```
Dav1dde: We already have a few tests that assert 429s. We should keep this and assert the 429 instead of ignoring it. This change also effectively ignores any other error than 429.

iambriccardo: Yes, I was aware of that, then I will try to handle it.

PR #3714, opened 2024-06-10
feat(profiles): Support profiler_id in context

phacops:
As part of continuous profiling, we need to store a `profiler_id` on the profile context of a transaction.


Dav1dde (commented):
This won't work, the context is recreated in `transfer_id`:
https://github.com/getsentry/relay/blob/bdb89f2318a3f28d9212d4faf03b0634c329b8c9/relay-server/src/services/processor/profile.rs#L66-L87

relay-event-schema/src/protocol/contexts/profile.rs
```diff
@@ -10,8 +10,12 @@ use crate::protocol::EventId;
 #[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
 pub struct ProfileContext {
     /// The profile ID.
-    #[metastructure(required = "true")]
+    #[metastructure(required = "false")]
     pub profile_id: Annotated<EventId>,
+
+    /// The profiler ID.
+    #[metastructure(required = "false")]
+    pub profiler_id: Annotated<EventId>,
```
jjbayer, line 16: Would we transfer this `profiler_id` from the profile itself, just like we do for `profile_id`?

phacops: `profiler_id` wouldn't be transferred. Either it's there or it's not.

relay-event-schema/src/protocol/contexts/profile.rs
```diff
@@ -10,8 +10,12 @@ use crate::protocol::EventId;
 #[cfg_attr(feature = "jsonschema", derive(JsonSchema))]
 pub struct ProfileContext {
     /// The profile ID.
-    #[metastructure(required = "true")]
+    #[metastructure(required = "false")]
```
Dav1dde: Isn't required = false the default and we should just omit the entire thing?

relay-server/src/services/processor/profile.rs
```diff
@@ -76,13 +76,10 @@ pub fn transfer_id(
...
             contexts.add(ProfileContext {
                 profile_id: Annotated::new(profile_id),
+                ..Default::default()
             });
         }
-        None => {
-            if let Some(contexts) = event.contexts.value_mut() {
-                contexts.remove::<ProfileContext>();
-            }
-        }
```
Dav1dde: Now it's possible that we emit a profile id without a corresponding profile, the code was introduced to exactly prevent that.

relay-server/src/services/processor/profile.rs
```diff
@@ -76,13 +76,10 @@ pub fn transfer_id(
...
+        _ => (),
```
Dav1dde: If that is the intended behaviour `if let Some() = profile_id` is the thing to use here.

Dav1dde (approved):

relay-server/src/services/processor/profile.rs
```diff
@@ -76,11 +76,14 @@ pub fn transfer_id(
...
                 profile_id: Annotated::new(profile_id),
+                ..ProfileContext::default()
             });
         }
         None => {
             if let Some(contexts) = event.contexts.value_mut() {
-                contexts.remove::<ProfileContext>();
+                if let Some(profile_context) = contexts.get_mut::<ProfileContext>() {
+                    profile_context.profile_id = Annotated::empty();
+                }
```
Zylphrex, line 86: A test for this would be nice that only the `profile_id` (not `profiler_id`) gets removed.

Zylphrex (approved):
LGTM

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -132,6 +134,8 @@ impl SpanTagKey {
             SpanTagKey::TraceStatus => "trace.status",
             SpanTagKey::MessagingDestinationName => "messaging.destination.name",
             SpanTagKey::MessagingMessageId => "messaging.message.id",
+            SpanTagKey::ThreadName => "thread.name",
+            SpanTagKey::ThreadId => "thread.name",
```
```suggestion
            SpanTagKey::ThreadId => "thread.id",
```

Dav1dde (commented):
Please move the thread id/name changes to a separate PR

PR #3633, opened 2024-05-21
feat(spans): Add separate feature flags for add-ons span metrics and indexed spans

phacops:
With this PR

* `projects:span-metrics-extraction` will only control the set of "common" metrics. It no longer controls indexed span extraction.
* `projects:span-metrics-extraction-addons` in addition to `projects:span-metrics-extraction` will extract additional metrics for the `ai`, `queue` and `cache` modules.
* `organizations:indexed-spans-extraction` controls whether spans are extracted from transaction payloads. This flag is now independent of the metrics-related flags.

PR #3719, opened 2024-06-12
fix(spans): Always count span usage

jjbayer:
https://github.com/getsentry/relay/pull/3633 introduced separate feature flags for span extraction and span metrics, but this meant that there was no usage metric for spans that were extracted without common span metrics. We need the usage metric for accounting purposes.

### DIFF
--- a/relay-dynamic-config/src/defaults.rs
+++ b/relay-dynamic-config/src/defaults.rs
@@ -70,6 +70,20 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
 
     let features = &project_config.features;
 
+    // If there are any spans in the system, extract the usage metric for them:
+    let any_spans = features.has(Feature::ExtractSpansFromEvent)
+        || features.has(Feature::StandaloneSpanIngestion)
+        || features.has(Feature::ExtractCommonSpanMetricsFromEvent);
+    if any_spans {
+        config.metrics.push(MetricSpec {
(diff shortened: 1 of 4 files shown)


relay-dynamic-config/src/defaults.rs
```diff
@@ -70,6 +70,20 @@ pub fn add_span_metrics(project_config: &mut ProjectConfig) {
 
     let features = &project_config.features;
 
+    // If there are any spans in the system, extract the usage metric for them:
+    let any_spans = features.has(Feature::ExtractSpansFromEvent)
+        || features.has(Feature::StandaloneSpanIngestion)
+        || features.has(Feature::ExtractCommonSpanMetricsFromEvent);
```
phacops, line 76: You could add `Feature::ExtractAddonsSpanMetricsFromEvent` as well even though it's meant to be enabled only if common is enabled.

phacops (approved):

relay-protocol/src/value.rs
```diff
@@ -425,6 +425,7 @@ impl<'a> Val<'a> {
     pub fn as_str(&self) -> Option<&'a str> {
         match self {
             Self::String(value) => Some(value),
+
```
iker-barriocanal, line 428:
```suggestion
```

iker-barriocanal (approved):

PR #3720, opened 2024-06-12
feat(metrics): Track the number of partitions flushed

iambriccardo:
This PR adds a new metric to track the number of partitions in each flush cycle.

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregatorservice.rs
+++ b/relay-metrics/src/aggregatorservice.rs
@@ -294,10 +294,10 @@ impl AggregatorService {
             return;
         }
 
-        let partitions_number = partitions.len();
-        relay_log::trace!("flushing {} partitions to receiver", partitions_number);
+        let partitions_count = partitions.len() as u64;
+        relay_log::trace!("flushing {} partitions to receiver", partitions_count);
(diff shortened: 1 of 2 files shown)


relay-metrics/src/aggregatorservice.rs
```diff
@@ -294,7 +294,12 @@ impl AggregatorService {
             return;
         }
 
-        relay_log::trace!("flushing {} partitions to receiver", partitions.len());
+        let partitions_count = partitions.len() as u64;
+        relay_log::trace!("flushing {} partitions to receiver", partitions_count);
+        relay_statsd::metric!(
+            histogram(MetricHistograms::PartitionsFlushed) = partitions_count,
+            aggregator = self.aggregator.name(),
+        );
```
Dav1dde, line 302:
```suggestion
        relay_log::trace!("flushing {} partitions to receiver", partitions_count);
        relay_statsd::metric!(
            histogram(MetricHistograms::PartitionsFlushed) = partitions.len() as u64,
            aggregator = self.aggregator.name(),
        );
```

Dav1dde (approved):
Feel free to ignore

PR #3725, opened 2024-06-13
ref(ci): Only build relay for tests once

Dav1dde:
The makefile already builds Relay.

#skip-changelog


iker-barriocanal (approved):
Nice! 🚀

PR #3726, opened 2024-06-13
fix(metrics): Improve flush time calculation in metrics aggregator

iambriccardo:
This PR tries to improve the metrics aggregator flush time calculation by trying to align the previously misaligned intervals due to a different precision in how time differences were calculated.

### DIFF
--- a/relay-metrics/src/aggregator.rs
+++ b/relay-metrics/src/aggregator.rs
@@ -1577,4 +1577,38 @@ mod tests {
         ]
         "###);
     }
+
+    #[test]
+    fn test_get_flush_time_with_backdated_bucket() {
+        let mut config = test_config();
+        config.bucket_interval = 3600;
+        config.flush_partitions = Some(10);
+        config.flush_batching = FlushBatching::Partition;
+
+        let now_s =
+            (UnixTimestamp::now().as_secs() / config.bucket_interval) * config.bucket_interval;
+
+        // First bucket has a timestamp two hours ago.
+        let timestamp = UnixTimestamp::from_secs(now_s - 7200);
+        let bucket_key_1 = BucketKey {
+            project_key: ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap(),
+            timestamp,
+            metric_name: "c:transactions/foo".into(),
+            tags: BTreeMap::new(),
+        };
+
+        // Second bucket has a timestamp in this hour.
+        let timestamp = UnixTimestamp::from_secs(now_s);
+        let bucket_key_2 = BucketKey {
+            project_key: ProjectKey::parse("a94ae32be2584e0bbd7a4cbb95971fee").unwrap(),
+            timestamp,
+            metric_name: "c:transactions/foo".into(),
+            tags: BTreeMap::new(),
+        };
+
+        let flush_time_1 = config.get_flush_time(&bucket_key_1);
+        let flush_time_2 = config.get_flush_time(&bucket_key_2);
+
+        assert_eq!(flush_time_1.elapsed(), flush_time_2.elapsed());
+    }
 }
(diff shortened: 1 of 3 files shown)


relay-metrics/src/aggregator.rs
```diff
@@ -510,6 +474,56 @@ impl fmt::Debug for CostTracker {
...
+
+    let instant = if flush_timestamp > now {
+        Instant::now().checked_add(flush_timestamp - now)
+    } else {
+        Instant::now().checked_sub(now - flush_timestamp)
+    }
+    .unwrap_or_else(Instant::now);
+
+    // Since `Instant` doesn't allow to get directly how many seconds elapsed, we leverage the
+    // diffing to get a duration and round it to the smallest second.
```
Dav1dde:
```suggestion
    // diffing to get a duration and round it to the smallest second to get consistent times.
```

relay-metrics/src/aggregator.rs
```diff
@@ -679,6 +695,7 @@ impl Aggregator {
         timestamp: UnixTimestamp,
         bucket_width: u64,
     ) -> Result<UnixTimestamp, AggregateMetricsError> {
+        println!("BUCKET WIDTH {:?}", bucket_width);
```
Dav1dde: 👀

relay-metrics/src/aggregator.rs
```diff
@@ -535,6 +549,7 @@ pub struct Aggregator {
     config: AggregatorConfig,
     buckets: HashMap<BucketKey, QueuedBucket>,
     cost_tracker: CostTracker,
+    base_instant: Instant,
```
Dav1dde:
```suggestion
    reference_time: Instant,
```

Not sure about the naming here, I am not a fan of `base_instant`, the type already tells us its an instant, and base doesn't fit quite right.

relay-metrics/src/aggregator.rs
```suggestion
```

relay-metrics/src/aggregator.rs
```diff
@@ -1571,4 +1588,42 @@ mod tests {
         ]
         "###);
     }
+
+    #[test]
+    fn test_get_flush_time_with_backdated_bucket() {
```
Dav1dde, line 1592: I'd prefer if the test would test the aggregator, we should be able to fast forward time with [`tokio::time::advance`](https://docs.rs/tokio/latest/tokio/time/fn.advance.html).

relay-metrics/src/aggregator.rs
```diff
@@ -510,6 +474,56 @@ impl fmt::Debug for CostTracker {
...
+    } else {
+        Instant::now().checked_sub(now - flush_timestamp)
+    }
+    .unwrap_or_else(Instant::now);
+
+    // Since `Instant` doesn't allow to get directly how many seconds elapsed, we leverage the
+    // diffing to get a duration and round it to the smallest second.
+    let instant = base_instant + Duration::from_secs((instant - base_instant).as_secs());
+    instant + config.flush_time_shift(bucket_key)
+}
```
jjbayer, line 525: Is there a good reason for moving this function? It makes it harder to spot what has changed.

Dav1dde: With the introduction of the reference time, I don't consider this 'config' anymore (was questionable before). I suggested to move it and make it a private detail of the aggregator.

iambriccardo: @Dav1dde suggested moving it out of the config because he didn't want the config as self and the base instant supplied inside. He might motivate his choice better.

relay-metrics/src/aggregator.rs
Dav1dde: Talked about it, doesn't seem feasible. Let's keep the test as is.

Dav1dde (approved):

relay-metrics/src/aggregator.rs
jjbayer: I missed that it's a standalone function now. Thank you, all clear!

PR #3730, opened 2024-06-14
ref(normalization): Only run normalization once, with item headers

iker-barriocanal:
This PR updates the approach to decide whether to run normalization on an event.

Relay looks at the headers of items in the envelope to decide whether to run normalization, instead of relying on assumptions on whether the previous relay run normalization. Since external systems can rewrite the header, only envelopes coming from relays in the internal infrastructure are trusted.

Relay resets the `fully_normalized` flag after updating the event when processing envelopes, in order to ensure it runs normalization. It follows the same approach for events created in the processing pipeline.

### Events from attachments

Relay disables the normalization flag for events created from attachments, and it will later normalize them.

This step is required as only processing relays deal with attachments. It could be simpler if we moved the attachment processing out of processing relays, but this is out of scope of this PR. 

## Background

Currently, PoP relays run some normalization, and processing relays run full normalization. Since both layers are internal and their deployments are controlled by Sentry, running full normalization in PoP relays saves the CPU in processing relays (estimated around 10%) while ensuring it always runs once.

## Previous attempt

The previous attempt forced full normalization in PoPs and assumed processing relays would always receive the fully normalized events. However, it didn't consider events created _in_ processing relays (from minidumps and various crash reporting mechanisms), resulting in ingesting events without running normalization.

## Rollout and rollback

1. S4S with feature flags.
2. SaaS with feature flags.
3. S4S with config.
4. SaaS with config.

Feature flags will first enable full normalization in PoPs and then disable normalization in processing. The integration test `test_relay_chain_normalizes_events` validates this flow. Disabling feature flags will roll back this behavior.


relay-dynamic-config/src/global.rs
iker-barriocanal, line 1: There have been no protocol/normalization changes, so it's safe to reuse the same feature flags.

tests/fixtures/extended-event-input.json
```diff
@@ -161,5 +161,5 @@
         "username": "username"
     },
     "version": "7",
-    "location": null
+    "location": "not-empty-so-that-it-isnt-dropped"
```
iker-barriocanal, line 164: Extra fields with null values are not serialized. Setting this value to anything else ensures we test future-compatibility.

tests/integration/test_store.py
iker-barriocanal, line 1: These use cases are covered in new tests in `tests/integration/test_normalization.py`.

relay-dynamic-config/src/global.rs
```diff
@@ -227,7 +227,8 @@ pub struct Options {
     )]
     pub force_full_normalization: bool,
 
-    /// If enabled, disables normalization in processing Relays.
+    /// If enabled, disables normalization in processing Relays for events
```
jjbayer:
```suggestion
    /// If true, disables normalization in processing Relays for events
```

relay-server/src/services/processor.rs
```diff
@@ -663,6 +663,12 @@ struct ProcessEnvelopeState<'a, Group> {
 
     /// Reservoir evaluator that we use for dynamic sampling.
     reservoir: ReservoirEvaluator<'a>,
+
+    /// Track whether the event has already been fully normalized.
+    ///
+    /// If the processing pipeline applies changes to the event, it should
+    /// disable this flag to ensure the event is always normalized.
+    fully_normalized: bool,
```
jjbayer: See `event_metrics_extracted` above.

```suggestion
    event_fully_normalized: bool,
```

relay-server/src/services/processor.rs
```diff
@@ -1314,37 +1325,62 @@ impl EnvelopeProcessorService {
...
-                    {
-                        true
-                    } else {
-                        relay_log::trace!("Skipping event normalization");
-                        return Ok(());
-                    }
+        }
+
+        let options = &self.inner.global_config.current().options;
+        let from_internal = state.envelope().meta().is_from_internal_relay();
```
jjbayer: I would move the `from_internal` check to the place where we set `state.event_fully_normalized`.

relay-server/src/services/processor.rs
```diff
@@ -1314,37 +1325,62 @@ impl EnvelopeProcessorService {
...
+
+        let full_normalization = match self.inner.config.normalization_level() {
```
jjbayer, line 1429: With the new item header, do we still need the `processing` check? Could we now simplify this to (with the suggestion above):

```rust
let full_normalization = match self.inner.config.normalization_level() => {
    NormalizationLevel::Full => true,
    NormalizationLevel::Default => ! state.event_fully_normalized.
};
```

Second question: Do we still need the global config options? If so, can we create a helper function that returns a normalization level depending on static + global config? E.g. `self.inner.config.normalization_level(options)`.

iker-barriocanal: > Do we still need the global config options?

Yes, I'd like to reduce the risk of the rollout with feature flags.

> can we create a helper function that returns a normalization level depending on static + global config

I find it harder to read. There are 3 states: full normalization, partial normalization, and no normalization (the `return`), so a boolean is not enough and we'd need an enum. I'm happy to implement it if you think it's easier to read.

> With the new item header, do we still need the processing check?

Yes. The processing checks are mostly used for feature flags (there are different flags for pops/processing), but also to not require config changes to external relays and not break Self-Hosted.

After the rollout, we can remove the feature flags and have something like the following:

```rs
        let full_normalization = match self.inner.config.normalization_level() {
            NormalizationLevel::Full => true,
            NormalizationLevel::Default => {
                if from_internal && state.fully_normalized {
                    return Ok(());
                } else {
                    // still required for self-hosted
                    self.inner.config.processing_enabled()
                }
            }
        };
```

relay-server/src/services/processor.rs
iker-barriocanal: Not sure if I understand it -- do you mean adding `from_internal` to the state? I find that redundant as that info is in the envelope.

jjbayer: I mean defining `event_fully_normalized` as `event_item.fully_normalized && is_from_internal_relay` .. that has the advantage that any piece of code that reads from `event_fully_normalized` can rely on the fact that the event was normalized by a trusted relay.

relay-server/src/services/processor.rs
jjbayer: What do we need `options.force_full_normalization` for?

I think the expression in the `Default` branch can be simplified to 

```rust
if self.inner.config.processing_enabled()
      && options.processing_disable_normalization
      && from_internal
      && state.event_fully_normalized
  {
      // ...
  } else {
      self.inner.config.processing_enabled() || options.force_full_normalization
  }
  ```

File: relay-server/src/services/processor.rs (hunk 4 above)
Comment (iker-barriocanal) [reply relay-server/src/services/processor.rs]: Done! Thanks, this is good feedback.

File: relay-server/src/services/processor.rs (hunk 5 above)
Comment (iker-barriocanal) [reply relay-server/src/services/processor.rs:1429]: There are two feature flags, and `options.force_full_normalization` is for PoP relays. Simplified the branch too!

File: relay-server/src/envelope.rs [hunk 6]
```diff
@@ -568,6 +568,13 @@ pub struct ItemHeaders {
     #[serde(default, skip_serializing_if = "is_false")]
     spans_extracted: bool,
 
+    /// Whether the event has been _fully_ normalized.
+    ///
+    /// If the event has been partially normalized, this flag is false. By
+    /// default, all Relays run some normalization.
```
jjbayer, line 574: nit

```suggestion
    /// Whether the item has been _fully_ normalized.
    ///
    /// If the item has been partially normalized, this flag is false. By
    /// default, all Relays run some normalization.
    ///
    /// Currently only used for events.
```

relay-server/src/services/processor.rs
```diff
@@ -1391,37 +1404,58 @@ impl EnvelopeProcessorService {
         &self,
         state: &mut ProcessEnvelopeState<G>,
     ) -> Result<(), ProcessingError> {
-        let options = &self.inner.global_config.current().options;
+        let attachment_type = state
+            .envelope()
+            .get_item_by(|item| item.attachment_type().is_some())
+            .and_then(|item| item.attachment_type())
+            .map(|ty| ty.to_string())
+            .unwrap_or("none".to_string());
```
jjbayer: Let's not allocate in the default case.
```suggestion
            ;
```

relay-server/src/services/processor/event.rs
```diff
@@ -68,12 +69,9 @@ pub fn extract<G: EventProcessing>(
         return Err(ProcessingError::DuplicateItem(duplicate.ty().clone()));
     }
 
-    let is_normalization_enabled =
-        if config.processing_enabled() && global_config.options.processing_disable_normalization {
-            false
-        } else {
-            config.normalization_level().is_enabled()
-        };
+    let normalization_skipped = config.processing_enabled()
```
jjbayer: nit: The skipping has not happened yet at this point.

```suggestion
    let skip_normalization = config.processing_enabled()
```

jjbayer (approved):

PR #3732, opened 2024-06-14
fix(unreal): Stop overwriting a user supplied error level

### LINKED ISSUE #3727: Can't change the error level from an Unreal crash reporter event
Hi!
I'd like to change our crash event from Unreal's crashreporter from "error" to "fatal"
If I'm correct, relay only changes the error level when a crash has the "is asset" or "is ensure" tag. Else it's always using the "error" level
( see https://github.com/getsentry/relay/blob/7b8316f6aecdcc000bc53879ff17338225a5415a/relay-server/src/utils/unreal.rs#L249 ) 
Something like:
```rust
if let Some(explicit_level) = context.engine_data.get("level") {
    event.level = Level::from_str(explicit_level).unwrap();
} else {
    if runtime_props.is_assert.unwrap_or(false) {
        event.level = Annotated::new(Level::Error)
    }
    if runtime_props.is_ensure.unwrap_or(false) {
        event.level = Annotated::new(Level::Warning)
    }
}
```

Dav1dde:
Fixes: #3727

Also updates the test dump with config:
- `IsAssert` is set to `true`
- `__sentry` is extended to contain a `warning` level


jjbayer (approved):
LGTM, maybe @bitsandfoxes can confirm that this is the behavior we want.

tests/integration/fixtures/native/unreal_crash_with_config
iker-barriocanal, line 1: What does this file look like?

iker-barriocanal (approved):

bitsandfoxes:
> LGTM, maybe @bitsandfoxes can confirm that this is the behavior we want.

This is exactly the behaviour we're looking for! 🙏

bitsandfoxes (approved):
Thanks!

PR #3736, opened 2024-06-17
ref(cogs): Emit cogs as statsd metrics

Dav1dde:
Update usage accountant to use a PR/commit which makes the `Producer` emit a `Message` instead of `Vec<u8>`: https://github.com/getsentry/rust-usage-accountant/pull/11

Emits all the COGS data also as statsd, this allows us to track COGS in PoP Relays but also help us track down why COGS data is slightly off for some categories.

#skip-changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,7 +141,7 @@ sentry-core = "0.32.2"
 sentry-kafka-schemas = { version = "0.1.86", default-features = false }
 sentry-release-parser = { version = "1.3.2", default-features = false }
 sentry-types = "0.32.2"
-sentry_usage_accountant = { version = "0.1.0", default-features = false }
+sentry_usage_accountant = { git = "https://github.com/getsentry/rust-usage-accountant.git", rev = "48df2ef", default-features = false }
 semver = "1.0.23"
 serde = { version = "1.0.159", features = ["derive", "rc"] }
 serde-transcode = "1.1.1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -141,7 +141,7 @@ sentry-core = "0.32.2"
 sentry-kafka-schemas = { version = "0.1.86", default-features = false }
 sentry-release-parser = { version = "1.3.2", default-features = false }
 sentry-types = "0.32.2"
-sentry_usage_accountant = { version = "0.1.0", default-features = false }
+sentry_usage_accountant = { git = "https://github.com/getsentry/rust-usage-accountant.git", rev = "48df2ef", default-features = false }
```
jjbayer, line 144: What's the reason for this upgrade? Can we ask someone to release the usage accountant crate?

jjbayer (approved):
Please add a PR description and / or link a ticket.

nit: Maybe put the word "also" in the PR title to clarify that we're not replacing the old schema.

Cargo.toml
Dav1dde: It's still a PR which changes the callback to emit a `Message` instead of `Vec<u8>`

iker-barriocanal (approved):
> also help us track down why COGS data is slightly off for some categories

Why does emitting the (same?) data to statsd help debugging?

Dav1dde:
> Why does emitting the (same?) data to statsd help debugging?

That's the question, it's unclear if it is the same.

PR #3740, opened 2024-06-18
fix(cogs): Fix cogs calculations

Dav1dde:
Fixes the calculations for real, please don't look to closely why they were wrong ...

#skip-changelog


olksdr (approved):

iker-barriocanal (approved):
👍

PR #3743, opened 2024-06-19
build(Makefile): Add `--locked` flag to cargo build,test,doc

iker-barriocanal:
> Assert that `Cargo.lock` will remain unchanged

It's documented for cargo `build`, `test` and `doc`.

#skip-changelog


Dav1dde (commented):
We should only do this in CI, local development is fine.

Dav1dde (approved):
As discussed should rename it to `RELAY_CARGO_ARGS`.

PR #3744, opened 2024-06-19
ref(metrics): Remove debounce_delay aggregator option

iambriccardo:
This PR removes the `debounce_delay` aggregator option since this was not used anymore because of https://github.com/getsentry/relay/pull/3661 and https://github.com/getsentry/relay/pull/3726.

#skip-changelog

### DIFF
--- a/relay-metrics/src/aggregator.rs
+++ b/relay-metrics/src/aggregator.rs
@@ -162,21 +162,10 @@ pub struct AggregatorConfig {
     /// The initial delay in seconds to wait before flushing a bucket.
     ///
     /// Defaults to `30` seconds. Before sending an aggregated bucket, this is the time Relay waits
-    /// for buckets that are being reported in real time. This should be higher than the
-    /// `debounce_delay`.
+    /// for buckets that are being reported in real time.
     ///
     /// Relay applies up to a full `bucket_interval` of additional jitter after the initial delay to spread out flushing real time buckets.
     pub initial_delay: u64,
-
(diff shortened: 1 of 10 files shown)


relay-metrics/src/aggregatorservice.rs
```diff
@@ -33,23 +33,9 @@ pub struct AggregatorServiceConfig {
 
     /// The initial delay in seconds to wait before flushing a bucket.
     ///
-    /// Defaults to `30` seconds. Before sending an aggregated bucket, this is the time Relay waits
-    /// for buckets that are being reported in real time. This should be higher than the
-    /// `debounce_delay`.
-    ///
```
jjbayer, line 39: Why did you remove this comment?

jjbayer (approved):

Dav1dde (approved):

relay-metrics/src/aggregator.rs
```diff
@@ -506,7 +488,7 @@ fn get_flush_time(
         let floored_timestamp = (now.as_secs() / config.bucket_interval) * config.bucket_interval;
         UnixTimestamp::from_secs(floored_timestamp)
             + config.bucket_interval()
-            + config.debounce_delay()
+            + config.initial_delay()
```
iker-barriocanal, line 491: This is changing from 10 -> 30, is this intended?

iambriccardo: Yes, it will change only for processing relays since we had `debounce_delay` set to 30 for pop relays.

PR #3749, opened 2024-06-20
feat(spans): Correctly emit negative outcomes for rate limited transactions that have nested spans

### LINKED ISSUE #3705: Produce outcomes for spans in transactions
On the fast path (i.e. during request handling), we reject transactions if their quota is zero. For new plans that are based on spans rather than transactions, we would also like to produce an outcome for the number of rejected spans (contained spans + 1). This is not straight-forward because the transaction has not been parsed at the point where we reject it.

We could extend `MinimalEvent` with a list of empty objects (`spans`) just to count the number of spans to produce outcomes for: https://github.com/getsentry/relay/blob/e549ce1270f683f3ee86f4ebe5f6a6d0c95d161a/relay-server/src/endpoints/common.rs#L154-L160

Note: it is not necessary to derive a rate limiting decision for the transaction itself just from span quota, because sentry can / will send an exhausted transaction quota as well.

iambriccardo:
This PR correctly emits negative outcomes for rate-limited transactions with nested spans by also emitting negative outcomes for the nested spans themselves.

Closes: https://github.com/getsentry/relay/issues/3705

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 **Internal**:
 
 - Aggregate metrics before rate limiting. ([#3746](https://github.com/getsentry/relay/pull/3746))
-- Correctly emit negative outcomes for nested spans. ([#3749](https://github.com/getsentry/relay/pull/3749))
+- Report outcomes for spans when transactions are rate limited. ([#3749](https://github.com/getsentry/relay/pull/3749))
 
 ## 24.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,12 @@
 **Bug Fixes**:
 
 - Fixes metrics dropped due to missing project state. ([#3553](https://github.com/getsentry/relay/issues/3553))
+- Report outcomes for spans when transactions are rate limited. ([#3749](https://github.com/getsentry/relay/pull/3749))
 
 
 **Internal**:
 
 - Aggregate metrics before rate limiting. ([#3746](https://github.com/getsentry/relay/pull/3746))
-- Report outcomes for spans when transactions are rate limited. ([#3749](https://github.com/getsentry/relay/pull/3749))
 
 ## 24.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 **Internal**:
 
 - Aggregate metrics before rate limiting. ([#3746](https://github.com/getsentry/relay/pull/3746))
+- Correctly emit negative outcomes for nested spans. ([#3749](https://github.com/getsentry/relay/pull/3749))
 
 ## 24.6.0
 
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -1238,8 +1238,15 @@ impl EnvelopeProcessorService {
 
         let scoping = state.managed_envelope.scoping();
         let (enforcement, limits) = metric!(timer(RelayTimers::EventProcessingRateLimiting), {
-            envelope_limiter.enforce(state.managed_envelope.envelope_mut(), &scoping)?
+            envelope_limiter.compute(state.managed_envelope.envelope_mut(), &scoping)?
         });
+
+        if enforcement.event_active() {
+            state.remove_event();
+            debug_assert!(state.envelope().is_empty());
+        }
+
+        enforcement.apply_with_outcomes(&mut state.managed_envelope);
 
         // Use the same rate limits as used for the envelope on the metrics.
         // Those rate limits should not be checked for expiry or similar to ensure a consistent
@@ -1252,17 +1259,6 @@ impl EnvelopeProcessorService {
                 .project_cache
                 .send(UpdateRateLimits::new(scoping.project_key, limits));
         }
-
-        if enforcement.event_active() {
-            state.remove_event();
-            debug_assert!(state.envelope().is_empty());
-        }
-
-        enforcement.track_outcomes(
-            state.envelope(),
-            &state.managed_envelope.scoping(),
-            self.inner.addrs.outcome_aggregator.clone(),
-        );
 
         Ok(())
     }
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -1240,13 +1240,13 @@ impl EnvelopeProcessorService {
         let (enforcement, limits) = metric!(timer(RelayTimers::EventProcessingRateLimiting), {
             envelope_limiter.compute(state.managed_envelope.envelope_mut(), &scoping)?
         });
+        let event_active = enforcement.event_active();
+        enforcement.apply_with_outcomes(&mut state.managed_envelope);
 
-        if enforcement.event_active() {
+        if event_active {
             state.remove_event();
             debug_assert!(state.envelope().is_empty());
         }
-
-        enforcement.apply_with_outcomes(&mut state.managed_envelope);
 
         // Use the same rate limits as used for the envelope on the metrics.
         // Those rate limits should not be checked for expiry or similar to ensure a consistent
--- a/relay-server/src/services/project.rs
+++ b/relay-server/src/services/project.rs
@@ -1071,7 +1071,9 @@ impl Project {
                 .items()
                 .find(|i| *i.ty() == ItemType::Transaction && !i.spans_extracted())
                 .and_then(|i| serde_json::from_slice::<PartialEvent>(&i.payload()).ok())
-                .map_or(0, |p| p.spans.0)
+                // We do + 1, since we count the transaction itself because it will be extracted
+                // as a span and counted during the slow path of rate limiting.
+                .map_or(0, |p| p.spans.0 + 1)
         } else {
             0
         };
--- a/relay-server/src/services/project.rs
+++ b/relay-server/src/services/project.rs
@@ -1032,7 +1032,7 @@ impl Project {
     ///   should be accepted or discarded
     ///
     /// IMPORTANT: If the [`ProjectState`] is invalid, the `check_request` will be skipped and only
-    /// rate limites will be validated. This function **must not** be called in the main processing
+    /// rate limits will be validated. This function **must not** be called in the main processing
     /// pipeline.
     pub fn check_envelope(
         &mut self,
--- a/relay-server/src/services/project.rs
+++ b/relay-server/src/services/project.rs
@@ -933,14 +933,17 @@ impl Project {
             Ok(current_limits.check_with_quotas(quotas, item_scoping))
         });
 
-        let check_nested_spans = state
-            .as_ref()
-            .is_some_and(|s| s.has_feature(Feature::ExtractSpansFromEvent));
+        let (mut enforcement, mut rate_limits) =
+            envelope_limiter.enforce(envelope.envelope_mut(), &scoping)?;
 
         #[derive(Debug, Deserialize)]
         struct PartialEvent {
             spans: SeqCount,
         }
+
+        let check_nested_spans = state
+            .as_ref()
+            .is_some_and(|s| s.has_feature(Feature::ExtractSpansFromEvent));
 
         // We compute the number of nested spans before rate limiting since we will remove the
         // transaction item in case it is rate limited.
@@ -957,14 +960,11 @@ impl Project {
             0
         };
 
-        let (enforcement, mut rate_limits) =
-            envelope_limiter.enforce(envelope.envelope_mut(), &scoping)?;
-
         if spans > 0 {
-            track_nested_spans_outcomes(&envelope, &enforcement, spans);
+            add_nested_span_enforcements(&mut enforcement, spans);
         }
 
-        enforcement.track_outcomes(envelope.envelope(), &scoping, outcome_aggregator);
+        enforcement.apply_with_outcomes(envelope.envelope_mut(), &scoping, outcome_aggregator);
 
         envelope.update();
 
@@ -1064,34 +1064,24 @@ impl Project {
     }
 }
 
-/// Tracks the negative outcomes for the nested spans inside a transaction.
+/// Adds category limits for the nested spans inside a transaction.
 ///
 /// On the fast path of rate limiting, we do not have nested spans of a transaction extracted
 /// as top-level spans, thus if we limited a transaction, we want to count and emit negative
 /// outcomes for each of the spans nested inside that transaction.
-fn track_nested_spans_outcomes(
-    envelope: &ManagedEnvelope,
-    enforcement: &Enforcement,
-    spans: usize,
-) {
+fn add_nested_span_enforcements(enforcement: &mut Enforcement, spans: usize) {
     if !enforcement.event_active() {
         return;
     }
 
     if enforcement.event.is_active() {
-        envelope.track_outcome(
-            Outcome::RateLimited(enforcement.event.reason_code.clone()),
-            DataCategory::Span,
-            spans,
-        );
+        enforcement.spans = enforcement.event.clone_for(DataCategory::Span, spans);
     }
 
     if enforcement.event_indexed.is_active() {
-        envelope.track_outcome(
-            Outcome::RateLimited(enforcement.event_indexed.reason_code.clone()),
-            DataCategory::SpanIndexed,
-            spans,
-        );
+        enforcement.spans_indexed = enforcement
+            .event_indexed
+            .clone_for(DataCategory::SpanIndexed, spans);
     }
 }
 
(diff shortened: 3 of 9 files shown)


relay-server/src/utils/rate_limits.rs
```diff
@@ -373,6 +432,9 @@ impl Enforcement {
             profile_chunks,
         } = self;
 
+        let nested_spans = event.clone_for(DataCategory::Span, spans_length);
```
iambriccardo: In case `spans_length` is `0`, we will have both categories filtered out downstream since they won't be `active`.

relay-server/src/services/project.rs
```diff
@@ -1053,9 +1054,37 @@ impl Project {
...
+        // transaction item in case it is rate limited.
+        let spans = if check_nested_spans {
+            envelope
+                .envelope()
+                .items()
+                .find(|i| *i.ty() == ItemType::Transaction && !i.spans_extracted())
+                .and_then(|i| serde_json::from_slice::<PartialEvent>(&i.payload()).ok())
+                .map_or(0, |p| p.spans.0)
+        } else {
+            0
```
iambriccardo: Not very nice, but I wanted to avoid unnecessary work in case the flag is not set.

relay-server/src/services/project.rs
```diff
@@ -932,9 +933,39 @@ impl Project {
...
+                // as a span and counted during the slow path of rate limiting.
+                .map_or(0, |p| p.spans.0 + 1)
+        } else {
+            0
+        };
+
         let (enforcement, mut rate_limits) =
             envelope_limiter.enforce(envelope.envelope_mut(), &scoping)?;
+
+        if check_nested_spans {
```
Dav1dde, line 944:
```suggestion
        if spans > 0 {
```

tests/integration/test_spans.py
```diff
@@ -1718,9 +1718,8 @@ def summarize_outcomes():
 
 
 @pytest.mark.parametrize(
-    "category,raises_rate_limited",
+    "category,hits_fast_path",
```
Dav1dde: I prefer the `raises_rate_limit`.

iambriccardo: Yeah, will change it back. I have changed it because of a test refactor but I rolled it back.

relay-server/src/services/project.rs
iambriccardo: Good catch!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 **Internal**:
 
 - Aggregate metrics before rate limiting. ([#3746](https://github.com/getsentry/relay/pull/3746))
+- Correctly emit negative outcomes for nested spans. ([#3749](https://github.com/getsentry/relay/pull/3749))
```
jjbayer: Could you move this to**Bug Fixes**?
```suggestion
- Report outcomes for spans when transactions are rate limited. ([#3749](https://github.com/getsentry/relay/pull/3749))
```

relay-server/src/utils/rate_limits.rs
```diff
@@ -316,29 +316,29 @@ impl Default for CategoryLimit {
...
+    pub spans: CategoryLimit,
     /// The rate limit for the indexed span category.
-    spans_indexed: CategoryLimit,
+    pub spans_indexed: CategoryLimit,
     /// The combined rate limit for user-reports.
-    user_reports_v2: CategoryLimit,
+    pub user_reports_v2: CategoryLimit,
     /// The combined profile chunk item rate limit.
-    profile_chunks: CategoryLimit,
+    pub profile_chunks: CategoryLimit,
```
jjbayer, line 343: Can we make `track_nested_spans_outcomes` a method instead?

Dav1dde (approved):

relay-server/src/services/project.rs
```diff
@@ -932,9 +933,39 @@ impl Project {
...
+            track_nested_spans_outcomes(&envelope, &enforcement, spans);
```
jjbayer: I think we should integrate the new logic a little bit more with the `Enforcement`, to prevent having multiple code paths that do similar things. It's a good call to not inject the span count into the envelope summary, but it could still be part of the `Enforcement`. That is:

1. Pass a flag into `enforce` to conditionally check for spans in the transaction item and set `Enforcement::spans`.
2. `Enforcement::track_outcomes` automatically does the right thing

relay-server/src/utils/rate_limits.rs
Dav1dde: I don't think that's a good idea, this stuff is heavily tied to rate limiting and we specifically don't want the span count stuff be tied to it.

But maybe it should, should a span quota of 0 block transactions from being ingested?

relay-server/src/services/project.rs
Dav1dde: We could possibly pass the managed envelope and some options to the `track_outcomes` and do it there, I preferred this at the moment, because `track_outcomes` already does work correctly for rate limits with the envelope. Modifying it will affect both cached/non-cached rate limits.

Having `track_outcomes` inspect the (managed) envelope seems like a footgun to me, since it is potentially built from completely different values than what the envelope contains (`assume_event`).

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 **Internal**:
 
 - Aggregate metrics before rate limiting. ([#3746](https://github.com/getsentry/relay/pull/3746))
+- Report outcomes for spans when transactions are rate limited. ([#3749](https://github.com/getsentry/relay/pull/3749))
```
jjbayer: Could you please move this to `**Bug Fixes**`?

relay-server/src/services/project.rs
```diff
@@ -932,9 +933,39 @@ impl Project {
...
+                .map_or(0, |p| p.spans.0 + 1)
+        } else {
+            0
+        };
+
+        if spans > 0 {
+            add_nested_span_enforcements(&mut enforcement, spans);
+        }
+
+        enforcement.apply_with_outcomes(envelope.envelope_mut(), &scoping, outcome_aggregator);
```
jjbayer: nit: Could you move all of this into a helper function? Maybe that helper function can then inline add_nested_span_enforcements.

relay-server/src/utils/rate_limits.rs
```diff
@@ -508,7 +569,6 @@ where
         summary.event_category = self.event_category.or(summary.event_category);
 
         let (enforcement, rate_limits) = self.execute(&summary, scoping)?;
-        envelope.retain_items(|item| self.retain_item(item, &enforcement));
```
jjbayer, line 511: We should rename this function to something like `compute` to clarify that it does not enforce anything.

relay-server/src/utils/rate_limits.rs
```diff
@@ -403,6 +403,67 @@ impl Enforcement {
             })
     }
 
+    /// Applies the [`Enforcement`] on the [`Envelope`] by removing all items that were rate limited
+    /// and emits outcomes for each rate limited category.
+    pub fn apply_with_outcomes(
```
jjbayer: We should make `track_outcomes` private to make sure that all users call `apply_with_outcomes` instead.

jjbayer (changes requested):

relay-server/src/utils/rate_limits.rs
```diff
@@ -403,6 +403,67 @@ impl Enforcement {
...
+        self,
+        envelope: &mut Envelope,
+        scoping: &Scoping,
+        outcome_aggregator: Addr<TrackOutcome>,
```
Dav1dde: This could take the managed envelope instead and track using the managed envelope.

relay-server/src/services/project.rs
```diff
@@ -932,9 +933,39 @@ impl Project {
...
+                .and_then(|i| serde_json::from_slice::<PartialEvent>(&i.payload()).ok())
+                // We do + 1, since we count the transaction itself because it will be extracted
+                // as a span and counted during the slow path of rate limiting.
+                .map_or(0, |p| p.spans.0 + 1)
+        } else {
+            0
+        };
+
+        if spans > 0 {
+            add_nested_span_enforcements(&mut enforcement, spans);
```
Dav1dde: This function can now also do the conditional parsing. You had that before where the helper function did the parsing of the spans and was just conditionally called based on the feature flag.

relay-server/src/utils/rate_limits.rs
iambriccardo: I was thinking about it, I will have to adjust the code in the processor rate limit checks.

iambriccardo: I totally missed the other callsite because I had the cfg flag disabled.

relay-server/src/services/processor.rs
```diff
@@ -1253,17 +1260,6 @@ impl EnvelopeProcessorService {
...
-        if enforcement.event_active() {
-            state.remove_event();
-            debug_assert!(state.envelope().is_empty());
-        }
-
-        enforcement.track_outcomes(
-            state.envelope(),
-            &state.managed_envelope.scoping(),
-            self.inner.addrs.outcome_aggregator.clone(),
-        );
```
iambriccardo, line 1265: It's now done inside `apply_with_outcomes`

relay-server/src/services/processor.rs
```diff
@@ -1253,17 +1260,6 @@ impl EnvelopeProcessorService {
                 .send(UpdateRateLimits::new(scoping.project_key, limits));
         }
 
-        if enforcement.event_active() {
-            state.remove_event();
-            debug_assert!(state.envelope().is_empty());
-        }
```
iambriccardo, line 1259: Moved up, since it requires enforcement which is consumed by the `apply_with_outcomes`. I verified that the ordering doesn't seem to impact the end result.

relay-server/src/utils/rate_limits.rs
```diff
@@ -256,7 +256,8 @@ impl EnvelopeSummary {
 
 /// Rate limiting information for a data category.
 #[derive(Debug)]
-struct CategoryLimit {
+#[cfg_attr(test, derive(Clone))]
```
iambriccardo, line 259: I wanted to have the `clone()` in tests. Either we implement manually the `Clone` in the tests or we go this route. Idk which would be preferred. The manual definition might be more verbose.

relay-server/src/utils/rate_limits.rs
```diff
@@ -1043,30 +1005,49 @@ mod tests {
         }
     }
 
+    fn enforce_and_apply(
```
iambriccardo, line 1008: Generalized all the code into this function, since I wanted tests to be leaner since a lot of repetition was there previously and some tests were also passing without calling the `apply_with_outcomes` so to avoid that risk in the future, I decided to have everything into one.

relay-server/src/utils/rate_limits.rs
```diff
@@ -348,16 +350,7 @@ impl Enforcement {
...
     /// Helper for `track_outcomes`.
-    fn get_outcomes(
-        self,
-        envelope: &Envelope,
-        scoping: &Scoping,
-    ) -> impl Iterator<Item = TrackOutcome> {
-        let timestamp = relay_common::time::instant_to_date_time(envelope.meta().start_time());
-        let scoping = *scoping;
-        let event_id = envelope.event_id();
-        let remote_addr = envelope.meta().remote_addr();
```
iambriccardo, line 359: I made sure the fields were all the same as the ones set by the `track_outcome` of the `ManagedEnvelope`.

relay-server/src/utils/rate_limits.rs
```diff
@@ -390,30 +383,79 @@ impl Enforcement {
...
-                category: limit.category,
-                // XXX: on the limiter we have quantity of usize, but in the protocol
-                // and data store we're limited to u32.
-                quantity: limit.quantity as u32,
+            .map(move |limit| {
+                (
+                    Outcome::RateLimited(limit.reason_code),
+                    limit.category,
+                    limit.quantity,
+                )
```
iambriccardo, line 391: Felt like there was no need for a struct

relay-server/src/services/project.rs
```diff
@@ -932,9 +932,24 @@ impl Project {
...
+            .as_ref()
+            .is_some_and(|s| s.has_feature(Feature::ExtractSpansFromEvent));
+
+        // If we can extract spans from the event, we want to try and count the number of nested
+        // spans to correctly emit negative outcomes in case the transaction itself is dropped.
+        if check_nested_spans {
+            let spans_count = count_nested_spans(&envelope);
+            if spans_count > 0 {
+                add_nested_span_enforcements(&mut enforcement, spans_count);
+            }
```
Dav1dde: I'd call `count_nested_spans` from within `add_nested_span_enforcements` (which may need a new name then).

The high level control flow is `if feature_enabled { sync_spans_to_enforcement() }`, how to count the spans ins the concern of the sync function.

relay-server/src/services/project.rs
```diff
@@ -1033,6 +1048,47 @@ impl Project {
...
+/// Counts the nested spans inside an [`Envelope`] for each item of type
+/// transaction that didn't have spans extracted before.
+fn count_nested_spans(envelope: &ManagedEnvelope) -> usize {
+    #[derive(Debug, Deserialize)]
+    struct PartialEvent {
+        spans: SeqCount,
+    }
+
+    // We compute the number of nested spans before rate limiting since we will remove the
+    // transaction item in case it is rate limited.
```
Dav1dde: Outdated.

relay-server/src/services/project.rs
```diff
@@ -1033,6 +1048,47 @@ impl Project {
     }
 }
 
+/// Counts the nested spans inside an [`Envelope`] for each item of type
+/// transaction that didn't have spans extracted before.
```
Dav1dde: Technically it's the first. Practically envelopes aren't allowed to have multiple transactions.

relay-server/src/services/project.rs
```diff
@@ -1033,6 +1045,49 @@ impl Project {
...
+    }
+
+    envelope
+        .envelope()
+        .items()
+        .find(|i| *i.ty() == ItemType::Transaction && !i.spans_extracted())
+        .and_then(|i| serde_json::from_slice::<PartialEvent>(&i.payload()).ok())
+        // We do + 1, since we count the transaction itself because it will be extracted
+        // as a span and counted during the slow path of rate limiting.
+        .map_or(0, |p| p.spans.0 + 1)
```
jjbayer: nit

```suggestion
        .find(|item| *item.ty() == ItemType::Transaction && !item.spans_extracted())
        .and_then(|item| serde_json::from_slice::<PartialEvent>(&item.payload()).ok())
        // We do + 1, since we count the transaction itself because it will be extracted
        // as a span and counted during the slow path of rate limiting.
        .map_or(0, |event| event.spans.0 + 1)
```

relay-server/src/services/project.rs
```diff
@@ -1326,4 +1390,90 @@ mod tests {
...
+            (DataCategory::Span, 3),
+            (DataCategory::SpanIndexed, 3),
+        ];
+
+        for (expected_category, expected_quantity) in expected {
+            let outcome = outcome_aggregator_rx.blocking_recv().unwrap();
+            assert_eq!(outcome.category, expected_category);
+            assert_eq!(outcome.quantity, expected_quantity);
+        }
+    }
```
jjbayer, line 1478: No python, nice!

relay-server/src/utils/rate_limits.rs
jjbayer: Fine with me!

relay-server/src/utils/managed_envelope.rs
```diff
@@ -354,6 +354,10 @@ impl ManagedEnvelope {
         });
     }
 
+    pub fn track_outcome_object(&self, outcome: TrackOutcome) {
+        self.outcome_aggregator.send(outcome);
+    }
+
```
```suggestion
```

jjbayer (approved):

relay-server/src/utils/managed_envelope.rs
iambriccardo: Oh yeah good catch!

relay-server/src/utils/rate_limits.rs
```diff
@@ -466,40 +508,7 @@ where
     /// - Rate limits declare all active rate limits, regardless of whether they have been applied
     ///   to items in the envelope. This excludes rate limits applied to required attachments, since
     ///   clients are allowed to continue sending them.
-    ///
-    /// # Example
```
Dav1dde, line 470: Can we keep these docs, they seem to be still relevant? We just don't drop the items anymore directly.

Dav1dde (approved):

relay-server/src/utils/rate_limits.rs
iambriccardo: I wanted to rewrite them in the new function. Will do in a follow up.

PR #3753, opened 2024-06-20
feat(server): Extrapolate sampled metrics numbers

jan-auer:
Extrapolates numbers tracked as metrics after extraction from spans.
This allows to view the original order of magnitude before client
sampling and should remove jumps resulting from changes in sample rates.

Extrapolation is part of metric extraction in Relay and controlled via a
project config field. In addition to that, there is a global threshold
for the maximum duplication factor for expansion of distribution
metrics:

Extrapolation is controlled via `MetricExtractionConfig::extrapolate`.
Metrics from the namespaces listed in this field will be extrapolated
unless explicitly excluded. The default is empty.

```json
"metricExtraction": {
    "version": 1,
    "extrapolate": {
        "include": [
    	    "?:transactions/*",
    	    "?:spans/*",
    	    "?:custom/*"
        ],
        "exclude": [
    	    "c:spans/usage@none",
    	    "c:transactions/usage@none",
    	    "c:transactions/count_per_root_project@none"
        ]
    }
}
```

The option to control the maximum duplication factor is called
`sentry-metrics.extrapolation.duplication-limit`, declared in
`genericmetrics.yaml`, and is exposed to Relay via
`GlobalConfig::options`. It is a numeric value of type `u32` in the
range $[0, 2^{32})$. Relay can assume a non-zero default value when this
option is missing.

```yaml
# The maximum duplication factor used to extrapolate distribution metrics from sampled data.
#
# This applies as long as Relay duplicates distribution values to extrapolate. The default is
# `0`, which disables extrapolation of distributions completely. This option does not apply to
# any other metric types.
sentry-metrics.extrapolation.duplication-limit: 100
```

### DIFF
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -302,7 +302,7 @@ pub struct MetricExtractionConfig {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub tags: Vec<TagMapping>,
 
-    /// TODO(ja): Doc
+    /// Extrapolation for metrics extracted from sampled data.
     #[serde(default, skip_serializing_if = "ExtrapolationConfig::is_empty")]
     pub extrapolate: ExtrapolationConfig,
 
@@ -643,14 +643,14 @@ pub fn convert_conditional_tagging(project_config: &mut ProjectConfig) {
     }
 }
 
-/// TODO(ja): Doc
+/// Configuration for metric extrapolation from sampled data.
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct ExtrapolationConfig {
-    /// TODO(ja): Doc
+    /// A list of MRI glob patterns to include in extrapolation.
     #[serde(default)]
     pub include: Vec<LazyGlob>,
 
-    /// TODO(ja): Doc
+    /// A list of MRI glob patterns to exclude from extrapolation, overriding inclusion.
     #[serde(default)]
     pub exclude: Vec<LazyGlob>,
 }
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -303,7 +303,7 @@ pub struct MetricExtractionConfig {
     pub tags: Vec<TagMapping>,
 
     /// TODO(ja): Doc
-    #[serde(default, skip_serializing_if = "ExtrapolateConfig::is_empty")]
+    #[serde(default, skip_serializing_if = "ExtrapolationConfig::is_empty")]
     pub extrapolate: ExtrapolationConfig,
 
     /// This config has been extended with fields from `conditional_tagging`.
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -645,6 +645,7 @@ pub fn convert_conditional_tagging(project_config: &mut ProjectConfig) {
 
 /// Configuration for metric extrapolation from sampled data.
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct ExtrapolationConfig {
     /// A list of MRI glob patterns to include in extrapolation.
     #[serde(default)]
--- a/relay-dynamic-config/src/metrics.rs
+++ b/relay-dynamic-config/src/metrics.rs
@@ -302,6 +302,10 @@ pub struct MetricExtractionConfig {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub tags: Vec<TagMapping>,
 
+    /// TODO(ja): Doc
+    #[serde(default, skip_serializing_if = "ExtrapolateConfig::is_empty")]
+    pub extrapolate: ExtrapolateConfig,
+
     /// This config has been extended with fields from `conditional_tagging`.
     ///
     /// At the moment, Relay will parse `conditional_tagging` rules and insert them into the `tags`
@@ -339,6 +343,7 @@ impl MetricExtractionConfig {
             global_groups: BTreeMap::new(),
             metrics: Default::default(),
             tags: Default::default(),
+            extrapolate: Default::default(),
             _conditional_tags_extended: false,
             _span_metrics_extended: false,
         }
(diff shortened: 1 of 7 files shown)


relay-dynamic-config/src/metrics.rs
```diff
@@ -302,6 +302,10 @@ pub struct MetricExtractionConfig {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub tags: Vec<TagMapping>,
 
+    /// TODO(ja): Doc
+    #[serde(default, skip_serializing_if = "ExtrapolateConfig::is_empty")]
+    pub extrapolate: ExtrapolateConfig,
```
Dav1dde: I don't think this config should exist. Instead we should be making the extrapolation a boolean/enum on the `MetricSpec`.

It is error prone, requires careful sync with metric extraction versions and is very easy to just get wrong.
Also it applies to metrics which were not even extracted using the config this is scoped to.
Dropping this in the future and replacing it with an attribute on the `MetricSpec` will be a heavy dance with backwards compatibility for a feature we could have just never introduced in the first place. 

And we can do this in multiple steps, if we don't wanna do the extrapolation right now in the generic metrics extraction, we can build an index of all MRIs using `aho-corasick` once. Although probably better to just do it in the generic metrics extraction.

For transaction metrics, I am in favor of moving it as much as possible to generic metrics extraction, if our only special case remaining is the `count_per_root_project` metric, this should all work out fine. 
If we don't want to do this now, we can just pull the extrapolation into the transaction metric extraction and extrapolate there one by one.

Heck even an easier version is to maintain a list of transaction metrics that should alwyas be extrapolated within the source code, right next to the metrics we have there.

relay-server/src/services/processor.rs
```diff
@@ -651,38 +677,80 @@ impl ProcessingExtractedMetrics {
...
-        } = self.0;
+        let extrapolate = match self.project.config().metric_extraction {
+            ErrorBoundary::Ok(ref config) if !config.extrapolate.is_empty() => &config.extrapolate,
+            _ => return,
+        };
 
-        if !project_metrics.is_empty() {
-            project_cache.send(AddMetricBuckets::internal(project_key, project_metrics));
-        }
+        let max_duplication = self.global.options.extrapolation_duplication_limit as usize;
```
Dav1dde: Could define `extrapolation_duplication_limit` already as an usize in the config.

relay-dynamic-config/src/metrics.rs
```diff
@@ -638,6 +643,38 @@ pub fn convert_conditional_tagging(project_config: &mut ProjectConfig) {
     }
 }
 
+/// TODO(ja): Doc
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct ExtrapolateConfig {
```
Dav1dde:
```suggestion
pub struct ExtrapolationConfig {
```

Using the noun here matches what we do elsewhere e.g. `Processing*` not `Process*`

relay-dynamic-config/src/metrics.rs
Dav1dde: Also need to bump metric extraction version for this.

Dav1dde (commented):
Looks good to me, I am just really not happy about the allow/deny list of metrics and I think we can do better, see the comment attached to the `extrapolate` field.

relay-dynamic-config/src/metrics.rs
jan-auer: Makes sense, we'll do this once we're a bit more confident in this config.

jan-auer: This is a summary of our offline conversation: Ultimately, it's hard to predict where we will move with extrapolation. For the time being, we want it to apply _exclusively_ to extracted metrics, however we view it as an additional concern on top of metrics extraction. Therefore, we structure the config in a way where it is easy to implement extrapolation separately, move it into other places, or remove it in future.

relay-server/src/services/processor.rs
```diff
@@ -651,38 +677,80 @@ impl ProcessingExtractedMetrics {
...
+            }
+
+            match bucket.value {
+                BucketValue::Counter(ref mut counter) => {
+                    *counter = counter.saturating_mul(factor);
+                }
+                BucketValue::Distribution(ref mut dist) => {
+                    *dist = std::mem::take(dist)
+                        .into_iter()
+                        .flat_map(|f| std::iter::repeat(f).take(duplication))
```
jjbayer, line 717: Would be good to add a code comment somewhere that mentions that we want to move the repetition to the kafka producer (or close to the kafka producer) to save memory and minimize the rounding error.

Ideally Relay supports weighted buckets in the full pipeline, and only converts the weights in the last moment, so that it is easy to switch once storage supports weights.

relay-server/src/services/processor.rs
```diff
@@ -1188,13 +1257,19 @@ impl EnvelopeProcessorService {
             reservoir.set_redis(org_id, redis_pool);
         }
 
+        let extracted_metrics = ProcessingExtractedMetrics::new(
+            project_state.clone(),
+            global_config,
+            managed_envelope.envelope().dsc(), // TODO(ja): this is fine
```
```suggestion
            managed_envelope.envelope().dsc(),

**Review by jjbayer — APPROVED** [review]

File: relay-dynamic-config/src/metrics.rs [hunk 6]
```diff
@@ -638,6 +643,39 @@ pub fn convert_conditional_tagging(project_config: &mut ProjectConfig) {
...
+    /// patterns.
+    pub fn matches(&self, mri: &str) -> bool {
+        !self
+            .exclude
+            .iter()
+            .any(|glob| glob.compiled().is_match(mri))
+            && self
+                .include
+                .iter()
+                .any(|glob| glob.compiled().is_match(mri))
```
Dav1dde, line 675: nit: Could change the order here, if it's not included in the include we don't have to check excludes.

```suggestion
        self
            .include
            .iter()
            .any(|glob| glob.compiled().is_match(mri))
            && !self
                .exclude
                .iter()
                .any(|glob| glob.compiled().is_match(mri))
```

Dav1dde (approved):

relay-dynamic-config/src/metrics.rs
jan-auer: @jjbayer had suggested the opposite during implementation, hence the order. `exclude` is likely to have fewer and the cheaper globs, so we can execute those first.

relay-server/src/services/processor.rs
jan-auer: Good catch, will write a better comment, actually.

PR #3754, opened 2024-06-20
ref(normalization): Merge the two event validation steps

iker-barriocanal:
This PR merges the two existing validation functions, expected to be run in a specific order, into a single validation function. The requirements that caused the split no longer exist.

For an easier review:
- The first commit, `857d579cce7d8d4c8c02078e4ce2ccd05eb825d2`, contains the functional changes.
- The second commit, `c5f5a52f54227f0eb789bd5400156b5bc0a0fddf`, moves code around without functional changes.

#skip-changelog

### DIFF
--- a/relay-event-normalization/src/validation.rs
+++ b/relay-event-normalization/src/validation.rs
@@ -33,7 +33,7 @@ pub struct EventValidationConfig {
     ///
     /// Transactions that finish outside this range are invalid. The check is
     /// skipped if no range is provided.
-    pub timestamp_range: Option<Range<UnixTimestamp>>,
+    pub transaction_timestamp_range: Option<Range<UnixTimestamp>>,
 
     /// Controls whether the event has been validated before, in which case disables validation.
     ///
@@ -144,7 +144,7 @@ fn validate_transaction_timestamps(
         transaction_event.timestamp.value(),
     ) {
         (Some(start), Some(end)) => {
-            validate_timestamps(start, end, config.timestamp_range.as_ref())?;
+            validate_timestamps(start, end, config.transaction_timestamp_range.as_ref())?;
             Ok(())
         }
         (_, None) => Err(ProcessingAction::InvalidTransaction(
@@ -366,7 +366,7 @@ mod tests {
             validate_event(
                 &mut event,
                 &EventValidationConfig {
-                    timestamp_range: Some(UnixTimestamp::now()..UnixTimestamp::now()),
+                    transaction_timestamp_range: Some(UnixTimestamp::now()..UnixTimestamp::now()),
                     is_validated: false,
                     ..Default::default()
                 }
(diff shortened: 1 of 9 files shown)


relay-event-normalization/src/validation.rs
```diff
@@ -395,31 +350,27 @@ mod tests {
         })
     }
 
-    #[test]
-    fn test_discards_when_missing_timestamp() {
```
iker-barriocanal, line 399: I removed this test because it's not useful. The current flow in master looks like this:

1. `validate_event_timestamp`, through `normalize_timestamps`, adds a default `timestamp` if there isn't one. This function is executed before `validate_transaction`.
2. `validate_transaction` returns an error if there isn't an end timestamp.

Originally, these two functions could be called independently so the test validated a use case. However, now that these two are merged into `validate_event` and the above functions aren't exposed so they can't be called independently, the test is no longer useful (and could even be misleading into showing a use case that doesn't exist).

`test_timestamp_added_if_missing` validates the new behavior.

relay-event-normalization/src/validation.rs
```diff
@@ -11,9 +11,24 @@ use relay_protocol::{Annotated, ErrorKind, Meta};
...
+
+    /// The maximum amount of seconds an event can be dated in the past.
+    ///
+    /// If the event's timestamp is older, the received timestamp is assumed.
+    pub max_secs_in_past: Option<i64>,
+
+    /// The maximum amount of seconds an event can be predated into the future.
+    ///
+    /// If the event's timestamp lies further into the future, the received timestamp is assumed.
+    pub max_secs_in_future: Option<i64>,
```
jjbayer, line 30: This PR surfaces an inconsistency:

1. For timestamp correction we use two integers but for transactions we use a range.
2. We correct timestamps that are very old (30 days), but then we still dismiss transactions when they are older than 5 days. Should we correct transaction timestamps to 5 days instead?

relay-event-normalization/src/validation.rs
```diff
@@ -22,51 +37,107 @@ pub struct TransactionValidationConfig {
 
     /// Controls whether the event has been validated before, in which case disables validation.
     ///
-    /// By default, `is_validated` is disabled and transaction validation is run.
+    /// By default, `is_validated` is disabled and event validation is run.
```
jjbayer, line 40: We should rename `timestamp_range` to `transaction_timestamp_range` because it does not apply to error events.

relay-event-normalization/src/validation.rs
jjbayer: Treating this as resolved: 1. needs optional timestamps, 2. is a business decision.

jjbayer (approved):
We can consider my first comment as resolved, please take a look at my second comment before merging.

PR #3755, opened 2024-06-21
feat(statsd): Emit tokio runtime metrics via statsd

Dav1dde:
#skip-changelog

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
--- a/relay-server/src/services/stats.rs
+++ b/relay-server/src/services/stats.rs
@@ -1,0 +1,108 @@
+use std::sync::Arc;
+use std::time::Duration;
+
+use relay_config::{Config, RelayMode};
+use relay_statsd::metric;
+use relay_system::{Addr, Service};
+use tokio::time::interval;
+
+use crate::services::upstream::{IsNetworkOutage, UpstreamRelay};
+use crate::statsd::{RelayGauges, TokioGauges};
+
+/// Relay Stats Service.
+///
+/// Service which collects stats periodically and emits them via statsd.
+pub struct RelayStats {
+    config: Arc<Config>,
+    upstream_relay: Addr<UpstreamRelay>,
+}
+
+impl RelayStats {
+    pub fn new(config: Arc<Config>, upstream_relay: Addr<UpstreamRelay>) -> Self {
+        Self {
+            config,
+            upstream_relay,
+        }
+    }
+
+    async fn tokio_metrics(&self) {
+        let m = tokio::runtime::Handle::current().metrics();
+
+        metric!(gauge(TokioGauges::ActiveTasksCount) = m.active_tasks_count() as u64);
+        metric!(gauge(TokioGauges::BlockingQueueDepth) = m.blocking_queue_depth() as u64);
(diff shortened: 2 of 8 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
```
jjbayer, line 2: I hope this means "API unstable", not "unreliable"?

relay-server/src/services/stats.rs
```diff
@@ -0,0 +1,108 @@
...
+        metric!(gauge(TokioGauges::BudgetForcedYieldCount) = m.budget_forced_yield_count());
+        metric!(gauge(TokioGauges::NumBlockingThreads) = m.num_blocking_threads() as u64);
+        metric!(gauge(TokioGauges::NumIdleBlockingThreads) = m.num_idle_blocking_threads() as u64);
+        metric!(gauge(TokioGauges::NumWorkers) = m.num_workers() as u64);
+        for worker in 0..m.num_workers() {
+            let worker_name = worker.to_string();
+            metric!(
+                gauge(TokioGauges::WorkerLocalQueueDepth) =
+                    m.worker_local_queue_depth(worker) as u64,
+                worker = &worker_name,
```
jjbayer, line 41: Are we OK with the cardinality this produces (`num_pods * num_workers`)?

relay-server/src/services/stats.rs
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+    }
+}
+
+impl Service for RelayStats {
+    type Interface = ();
+
+    fn spawn_handler(self, _rx: relay_system::Receiver<Self::Interface>) {
+        let mut ticker = interval(Duration::from_secs(5));
```
jjbayer: I feel like we should make this configurable from the start, if only to have a killswitch to disable it (i.e. interpret `0` as disabled).

.cargo/config.toml
Dav1dde: That's what it means (API unstable). A bunch of it is already stable but not all of it.

relay-server/src/services/stats.rs
Dav1dde: What's the alternative? Can we emit a custom gauge with min/max/avg?

iker-barriocanal: The number of workers is static, so I think it should be fine.

relay-server/src/statsd.rs
```diff
@@ -48,6 +50,95 @@ impl GaugeMetric for RelayGauges {
...
+    /// Exposes [`RuntimeMetrics::num_blocking_threads`].
+    NumBlockingThreads,
+    /// Exposes [`RuntimeMetrics::num_idle_blocking_threads`].
+    NumIdleBlockingThreads,
+    /// Exposes [`RuntimeMetrics::num_workers`].
+    NumWorkers,
+    /// Exposes [`RuntimeMetrics::worker_local_queue_depth`].
+    ///
+    /// This metric is tagged with:
+    /// - `worker`: the worker id
```
iker-barriocanal: nit (same with below):
```suggestion
    /// - `worker`: the worker id.
```

iker-barriocanal (approved):
Are we trying to debug an issue with these metrics? Could we elaborate the PR description?

jjbayer (approved):
Looking forward to having these metrics!

Please make the interval configurable (and killswitchable), rest LGTM.

PR #3762, opened 2024-06-25
feat(web-vitals): add support for mobile browsers

KevinL10:
Adds support for mobile browsers to start ingesting web vitals. 

Continuation of https://github.com/getsentry/sentry/pull/73298/ which adds the scoring percentile configuration for mobile browsers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 **Internal**:
 
 - Aggregate metrics before rate limiting. ([#3746](https://github.com/getsentry/relay/pull/3746))
+- Add web vitals support for mobile browsers. ([#3762](https://github.com/getsentry/relay/pull/3762))
 
 ## 24.6.0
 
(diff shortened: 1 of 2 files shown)


Dav1dde:
I wonder if we should normalize the mobile browsers instead 🤔 .

KevinL10:
We could normalize `browser.name` but we still need a way to distinguish mobile browsers from desktop browsers since they have different scoring configurations. If we add a new tag it might be confusing for people who use span search and are already filtering by specific `browser.name` types like `Chrome Mobile`.

edwardgou-sentry (approved):
I think these browser names are set by relay by checking the user agent?
Normalizing these browser name sounds like it could be a different pr.

I think this change looks fine to unblock testing mobile perf scores for now.

iker-barriocanal (approved):

iker-barriocanal:
> I think these browser names are set by relay by checking the user agent?

I'm ok with merging as-is, but a test to validate that would be great.

Dav1dde (approved):

PR #3767, opened 2024-06-26
ref(profiles): Process profiles before metric extraction

Dav1dde:
Invalid profiles with a valid header which are ultimately discarded as invalid would still tag the transaction usage metric with `has_profile`, even though the profile was dropped.

- Collapses `process::transfer_id` into `process::filter`
- Moves profile expansion/processing before metrics extraction
- Adds an integration test for this case

### DIFF
--- a/relay-server/src/services/processor.rs
+++ b/relay-server/src/services/processor.rs
@@ -1632,7 +1632,8 @@ impl EnvelopeProcessorService {
             &self.inner.global_config.current(),
         )?;
 
-        profile::filter(state);
+        let profile_id = profile::filter(state);
+        profile::transfer_id(state, profile_id);
 
         if_processing!(self.inner.config, {
             attachment::create_placeholders(state);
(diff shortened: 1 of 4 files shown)


jjbayer (approved):

relay-server/src/services/processor.rs
```diff
@@ -1657,15 +1656,16 @@ impl EnvelopeProcessorService {
...
-            self.extract_transaction_metrics(state, SamplingDecision::Drop, profile_id)?;
-
             let keep_profiles = dynamic_sampling::forward_unsampled_profiles(state, &global_config);
-
             // Process profiles before dropping the transaction, if necessary.
-            if keep_profiles {
-                profile::process(state, &self.inner.config);
-            }
+            // Before metric extraction to make sure the profile count is reflected correctly.
+            let profile_id = match keep_profiles {
```
iker-barriocanal, line 1663: Is this easier to read than `if..else`?

iker-barriocanal (approved):

relay-server/src/services/processor.rs
jjbayer: One line less!

Dav1dde: One line less and personally for me it reads nicer.

PR #3768, opened 2024-06-27
feat(spans): Restrict extraction of span time metrics

jjbayer:
Only extract `span.duration`, `span.exclusive_time` and `span.exclusive_time_light` if it is needed for a common or an addon insights module.

We wanted to show the span.duration in the ddm UI for all spans, but this need disappeared, so we can restrict it again.

### DIFF
--- a/relay-event-normalization/src/normalize/span/tag_extraction.rs
+++ b/relay-event-normalization/src/normalize/span/tag_extraction.rs
@@ -2430,4 +2430,20 @@ LIMIT 1
             "admin@sentry.io"
         );
     }
+
+    #[test]
+    fn long_descriptions_are_truncated() {
+        let json = r#"{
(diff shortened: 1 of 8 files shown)


iker-barriocanal (approved):

Dav1dde (approved):

relay-event-normalization/src/normalize/span/tag_extraction.rs
```diff
@@ -2430,4 +2430,20 @@ LIMIT 1
...
+        }"#;
+
+        let span = Annotated::<Span>::from_json(json)
+            .unwrap()
+            .into_value()
+            .unwrap();
+
+        let tags = extract_tags(&span, 200, None, None, false, None);
+        assert_eq!(tags[&SpanTagKey::Description], "SELECT column FROM table1 WHERE another_col = %s AND yet_another_col = something_very_longggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggg*");
+    }
```
jjbayer, line 2448: Drive-by improvement: This was an integration test, but it does not need to be.

PR #3769, opened 2024-06-27
feat(metrics): Add ability to compute metrics summary and attach it to events

### LINKED ISSUE #73152: emit metrics summaries from relay when we have extraction rules
**Problem**: We are not able to show samples in the UI anymore today with metrics extracted from span attributes. We need to keep using metric summaries until the span indexing is done and we can switch over.

iambriccardo:
This PR introduces functionality to compute metric summaries from a slice of `Bucket`s, adhering to the semantics currently used in our SDKs.

The motivation for this change stems from our decision to discontinue metric emission from SDKs, instead extracting them in Relay through specified configurations. Without this update, the current Relay implementation would compromise the product experience, as it cannot extract metric summaries from event-derived metrics. This PR addresses and resolves this limitation.

Currently, metrics summaries are computed only for metrics belonging to the `custom` namespace.

Closes: https://github.com/getsentry/sentry/issues/73152


relay-event-schema/src/protocol/metrics_summary.rs
```diff
@@ -12,10 +13,39 @@ pub type MetricSummaryMapping = Object<Array<MetricSummary>>;
...
+
+    /// Merges a [`MetricsSummary`] into itself.
+    ///
+    /// The merge operation concatenates tags of the same metric even if they are identical for the
+    /// sake of simplicity. This doesn't cause issues for us, under the assumption that the two
+    /// summaries do not have overlapping metric names (which is the case for us since summaries
+    /// computed by the SDKs have metric names that are different from the ones we generate in
+    /// Relay).
+    ///
+    /// Do not use this function to merge any arbitrary [`MetricsSummary`]s.
```
iambriccardo: I am not happy about this comment, open for suggestions.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,432 @@
...
+
+        MetricsSummary(metrics_summary)
+    }
+}
+
+/// Computes the [`MetricsSummary`] from a slice of [`Bucket`]s.
+pub fn compute(buckets: &[Bucket]) -> Option<MetricsSummary> {
+    if buckets.is_empty() {
+        return None;
+    }
```
iambriccardo: Small optimization to avoid creating an aggregator with nothing inside.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,432 @@
...
+                x.push(Annotated::new(metric_summary));
+            }
+        }
+
+        MetricsSummary(metrics_summary)
+    }
+}
+
+/// Computes the [`MetricsSummary`] from a slice of [`Bucket`]s.
+pub fn compute(buckets: &[Bucket]) -> Option<MetricsSummary> {
```
jan-auer: Let's add a form of option or feature flag that controls whether this code should actually run. A global flag would be absolutely sufficient, though it would be useful to have the ability to control this per organization/project. Choose based on what's more straight forward to integrate.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,432 @@
...
+        for bucket in buckets {
+            aggregator.add(bucket);
+        }
+
+        aggregator
+    }
+
+    /// Adds a [`Bucket`] into the [`MetricsSummaryAggregator`].
+    pub fn add(&mut self, bucket: &Bucket) {
+        let key = MetricsSummaryBucketKey {
```
jan-auer, line 156: @shellmayr do we need summaries for built-in span metrics, too? If not, we should constrain this to metrics in the `custom` namespace. 

@iambriccardo ideally we could make this configurable in some form so that we have more flexibility in case we need to change this. However, if we know with certainty that we need this for only `custom` metrics, then you can hard-code this at an appropriate point and leave a code comment to explain the reasoning.

shellmayr: From what I can tell based on the code of the /metrics/samples/ endpoint, the samples are retrieved using the indexed spans and metrics summaries. Which of them is used depends on whether certain attributes on the span are indexed or not, but I am not yet sure what this means for built-in span metrics. I've reached out to @Zylphrex to see if there is more context on these mechanics - will follow up as soon as I know more.

jan-auer: This means that we won't need summaries for built-in span metrics, so we can add an exclusion. Preferably through a config option.

Zylphrex: Metric summaries are only used for custom metrics. Built-in metrics like transaction duration, span duration, and metrics derived from measurements can all be queried directly using other datasets.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,446 @@
...
+        return None;
+    }
+
+    // For now, we only want metrics summaries to be extracted for custom metrics.
+    let filtered_buckets = buckets
+        .iter()
+        .filter(|b| matches!(b.name.namespace(), MetricNamespace::Custom));
+
+    let aggregator = MetricsSummaryAggregator::from_buckets(filtered_buckets);
+    if aggregator.is_empty() {
```
iambriccardo: We do the check in the aggregator directly so that we avoid collecting the filtered buckets into a vector just for the sake of determining the size before calling `from_buckets`.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -46,38 +46,61 @@ impl Extractable for Span {
...
+    compute_metrics_summaries_sample_rate: Option<f32>,
 ) -> Vec<Bucket> {
-    let mut metrics = generic::extract_metrics(event, config);
+    let compute_metrics_summaries_sample_rate =
+        compute_metrics_summaries_sample_rate.unwrap_or(1.0);
 
-    // If spans were already extracted for an event,
-    // we rely on span processing to extract metrics.
+    let (mut metrics, metrics_summary_spec) = generic::extract_and_summarize_metrics(event, config);
+    if sample(compute_metrics_summaries_sample_rate) {
```
Dav1dde: I think we should use a rollout rate here so we get metric summaries consistently per org.

See `is_rolled_out` fn.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -46,38 +46,61 @@ impl Extractable for Span {
...
     max_tag_value_size: usize,
     output: &mut Vec<Bucket>,
+    compute_metrics_summaries_sample_rate: f32,
 ) {
     relay_statsd::metric!(timer(RelayTimers::EventProcessingSpanMetricsExtraction), {
         if let Some(transaction_span) = extract_transaction_span(event, max_tag_value_size) {
-            output.extend(generic::extract_metrics(&transaction_span, config));
+            let (metrics, metrics_summary_spec) =
+                generic::extract_and_summarize_metrics(&transaction_span, config);
+            if sample(compute_metrics_summaries_sample_rate) {
```
Dav1dde: This is a different decision than the check before, you'll have very inconsistent summaries like this. Either use `is_rolled_out` or pass it as a binary decision (boolean) and evaluating the decision only once.

relay-server/src/metrics_extraction/event.rs
```diff
@@ -46,38 +46,61 @@ impl Extractable for Span {
...
 
-        if let Some(spans) = event.spans.value() {
+        if let Some(spans) = event.spans.value_mut() {
             for annotated_span in spans {
-                if let Some(span) = annotated_span.value() {
-                    output.extend(generic::extract_metrics(span, config));
+                if let Some(span) = annotated_span.value_mut() {
+                    let (metrics, metrics_summary_spec) =
+                        generic::extract_and_summarize_metrics(span, config);
+                    if sample(compute_metrics_summaries_sample_rate) {
```
Dav1dde: dito about re-evaluation of sampling decision

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Add web vitals support for mobile browsers. ([#3762](https://github.com/getsentry/relay/pull/3762))
 - Accept profiler_id in the profile context. ([#3714](https://github.com/getsentry/relay/pull/3714))
 - Support extrapolation of metrics extracted from sampled data, as long as the sample rate is set in the DynamicSamplingContext. ([#3753](https://github.com/getsentry/relay/pull/3753))
+- Add ability to compute metrics summary on the extracted metrics. ([#3769](https://github.com/getsentry/relay/pull/3769))
```
```suggestion
- Compute metrics summary on the extracted custom metrics. ([#3769](https://github.com/getsentry/relay/pull/3769))
```

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+    /// The sum of all values reported in the bucket.
+    sum: Option<FiniteF64>,
+    /// The number of times this bucket was updated with a new value.
+    count: u64,
+}
+
+impl MetricsSummaryBucketValue {
+    /// Builds a [`MetricsSummaryBucketValue`] from a [`MetricSummary`] by aggregating the data into
+    /// a gauge like summary.
+    fn from_metric_summary(summary: &MetricSummary) -> MetricsSummaryBucketValue {
```
Dav1dde: This can just be a `From` impl

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+                .and_then(|min| FiniteF64::try_from(*min).ok()),
+            max: summary
+                .max
+                .value()
+                .and_then(|max| FiniteF64::try_from(*max).ok()),
+            sum: summary
+                .sum
+                .value()
+                .and_then(|sum| FiniteF64::try_from(*sum).ok()),
+            count: summary.count.value().cloned().unwrap_or(0),
```
Dav1dde:
```suggestion
            count: summary.count.value().copied().unwrap_or(0),
```

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+                .value()
+                .and_then(|min| FiniteF64::try_from(*min).ok()),
+            max: summary
+                .max
+                .value()
+                .and_then(|max| FiniteF64::try_from(*max).ok()),
+            sum: summary
+                .sum
+                .value()
+                .and_then(|sum| FiniteF64::try_from(*sum).ok()),
```
Dav1dde:
```suggestion
            min: summary.min.value().and_then(|&min| min.try_into().ok()),
            max: summary.max.value().and_then(|&max| max.try_into().ok()),
            sum: summary.sum.value().and_then(|&sum| sum.try_into().ok()),
```

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+            min: Some(gauge.min),
+            max: Some(gauge.max),
+            sum: Some(gauge.sum),
+            count: gauge.count,
+        }
+    }
+
+    /// Merges two [`MetricsSummaryBucketValue`]s together by mutating `self` and consuming
+    /// `other`.
+    fn merge(&mut self, other: MetricsSummaryBucketValue) {
```
Dav1dde: I'd instead implement `std::ops::Add` and make the value `Copy`, immutability ftw.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+        self.count += other.count;
+    }
+}
+
+/// metrics_summary_spec that tracks all the buckets containing the summaries for each
+/// [`MetricsSummaryBucketKey`].
+///
+/// The need for an metrics_summary_spec arises from the fact that we want to compute metrics summaries
+/// generically on any slice of [`Bucket`]s meaning that we need to handle cases in which
+/// the same metrics as identified by the [`MetricsSummaryBucketKey`] have to be merged.
```
Dav1dde: This doesn't explain what this is, requirements will change or the struct will be used elsewhere -> that comment does not apply anymore.

The comment should describe what this struct does and represents, not why it was created.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+pub struct MetricsSummarySpec {
```
Dav1dde: I'd just call this `MetricsSummary` and if you want to refer to the other one use an explicit path `event::MetricsSummary`. Might as well use modules when we have them, for me less confusing and get's around this inventing of new names.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+    buckets: BTreeMap<MetricsSummaryBucketKey, MetricsSummaryBucketValue>,
+}
+
+impl MetricsSummarySpec {
+    fn new() -> MetricsSummarySpec {
```
Dav1dde: Can be public, also should impl Default, surprised clippy doesn't complain about that.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+    /// Merges a [`MetricsSummaryBucketKey`] and [`MetricsSummaryBucketValue`] in the spec.
+    fn merge(&mut self, key: MetricsSummaryBucketKey, value: MetricsSummaryBucketValue) {
+        match self.buckets.entry(key) {
+            Entry::Occupied(mut entry) => {
+                entry.get_mut().merge(value);
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(value);
+            }
+        }
```
Dav1dde: We can just inline this. Make the value implement `Default` and `AddAssign`.

```rs
self.buckets.entry(key).or_default() += value;
```

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+                let mut tags = BTreeMap::new();
+                if let Some(inner_tags) = summary.tags.value() {
+                    for (tag_key, tag_value) in inner_tags.iter() {
+                        let Some(tag_value) = tag_value.value() else {
+                            continue;
+                        };
+
+                        tags.insert(tag_key.clone(), tag_value.clone());
+                    }
+                }
```
Dav1dde: Can be functional:

```suggestion
                let tags = summary
                    .tags
                    .value()
                    .iter()
                    .flat_map(|t| t.iter())
                    .filter_map(|(k, v)| Some((k.clone(), v.value()?.clone())))
                    .collect();
```

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(value);
+            }
+        }
+    }
+
+    /// Applies the [`MetricsSummarySpec`] on a receiving [`Annotated<MetricsSummary>`].
+    ///
+    /// In case the [`MetricsSummarySpec`] is empty, no mutation will take place.
```
Dav1dde: This is an implementation detail, not necessary for the docs.

Applying an no summaries to some summaries should always yield the same original summaries regardless.

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+                    .map_or(Annotated::empty(), |m| Annotated::new(m.to_f64())),
+                max: value
+                    .max
+                    .map_or(Annotated::empty(), |m| Annotated::new(m.to_f64())),
+                sum: value
+                    .sum
+                    .map_or(Annotated::empty(), |m| Annotated::new(m.to_f64())),
+                count: Annotated::new(value.count),
+                tags: Annotated::new(tags),
+            };
```
Dav1dde:
```suggestion
            let min = value.min.map_or(Annotated::empty(), |m| m.to_f64().into());
            let max = value.max.map_or(Annotated::empty(), |m| m.to_f64().into());
            let sum = value.sum.map_or(Annotated::empty(), |m| m.to_f64().into());
            let metric_summary = MetricSummary {
                min,
                max,
                sum,
                count: Annotated::new(value.count),
                tags: Annotated::new(tags),
            };
```

Not sure, feels like this is more readable?

relay-server/src/metrics_extraction/metrics_summary.rs
```diff
@@ -0,0 +1,511 @@
...
+            if let Some(x) = entry.value_mut() {
+                x.push(Annotated::new(metric_summary));
+            }
+        }
+
+        Annotated::new(MetricsSummary(metrics_summary))
+    }
+}
+
+/// Computes the [`MetricsSummary`] from a slice of [`Bucket`]s.
```
Dav1dde: This comment should contain that this also filters the buckets down to the custom namespace, this is the reason people should use this function instead of using the `MetricsSummary(Spec)` directly.

relay-server/src/metrics_extraction/metrics_summary.rs
Dav1dde: I'd move this into the utils package, keep the `compute` function here though.

relay-server/src/metrics_extraction/event.rs
iambriccardo: Good point, I didn't consider the different possibilities of sampling.

jan-auer (approved):

